PR #3, opened 2014-06-02
[FIX] remove raised error if obsolete fields are in old views

legalsylvain:
Do not raise an error if obsolete fields are presents in old views.

### DIFF
--- a/openerp/osv/orm.py
+++ b/openerp/osv/orm.py
@@ -1577,7 +1577,7 @@ class BaseModel(object):
         if error_msgs:
             # OpenUpgrade: do not raise for obsolete fields
             # raise except_orm('ValidateError', '\n'.join(error_msgs))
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Thanks! I remember in 7.0, we logged the validation errors at debug level, with a disclaimer that the error was probably ok during a migration. I would like to keep this if possible.

bwrsandman:
:+1:

legalsylvain:
@StefanRijnhart  Not sure. See : http://bazaar.launchpad.net/~openupgrade-committers/openupgrade-server/7.0/view/head:/openerp/osv/orm.py#L1949
For V7.
But ok I'll add a log.

StefanRijnhart:
Looks good now, except that you don't need to add 'pass' if you have any other line in your exception handling. Can you remove this line?

StefanRijnhart:
Thanks! :+1:

PR #5, opened 2014-06-02
[ADD] analysis work file for module report, new module nothing to do

bguillot:
Migrate the module report.
This is a new module so nothing has to be done. 
It doesn't mix with report_webkit (nothing in analysis of report_webkit)


legalsylvain:
:+1:

StefanRijnhart:
Thanks! :+1:

PR #6, opened 2014-06-02
migrate report_webkit, nothing has change except a new menu

bguillot:
Migrate report_webkit nothing has change except a new menu


legalsylvain:
:+1:

StefanRijnhart:
Webkit is going to be removed from the core, so I would rather not merge this because it would cause a conflict at that point

:-1:

bwrsandman:
As `report_webkit` is expected to be removed in 8.0 and nothing is to be done, I will now close this PR.

PR #8, opened 2014-06-02
[ADD] analysis work file for module resource, nothing has to be done to ...

bguillot:
Changes in the module resource don't imply migration scripts.

I just add the analysis_work file.


StefanRijnhart:
Agreed. Creating the leave ids will have to be taken care of in the holidays module.

PR #13, opened 2014-06-02
Add OpenUpgrade specific functions for message_last_post fields

bwrsandman:
Add set_message_last_post() which fills message_last_post values

Signed-off-by: Sandy Carter sandy.carter@savoirfairelinux.com


StefanRijnhart:
Tnx! Can you search with a context {'active_test': False} so that we get the inactive records as well, and please add the (nicely documented!) file to the doc/source/API.rst for the autodoc.

bwrsandman:
Done!

PR #14, opened 2014-06-02
Add pre and post migration scripts for mail

bwrsandman:
~~Depends on #13~~


addons/mail/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,68 @@
...
+      * reply_to
+    Get the mail.message from the mail.mail object and transfer the data that
+    way.
+    """
+    mail_mail_pool = pool['mail.mail']
+    legacy_server_id = openupgrade.get_legacy_name('mail_server_id')
+    legacy_reply_to = openupgrade.get_legacy_name('reply_to')
+    mail_ids = mail_mail_pool.search(
+        cr, uid, [])
+    for mail_obj in mail_mail_pool.browse(cr, uid, mail_ids):
```
hbrunn: why don't you simply update from a join over mail_message and mail_mail? The way you do it currently is very expensive I think and can really cause trouble for a lot of mails/messages

bwrsandman: Will do that.
Thanks

bwrsandman:
@hbrunn Replaced loop with SQL query for greater efficiency.

hbrunn:
:+1:

addons/mail/migrations/8.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
...
+mail         / mail.message.subtype     / hidden (boolean)              : NEW
+mail         / mail.message.subtype     / sequence (integer)            : NEW
+mail         / res.users                / display_groups_suggestions (boolean): NEW
+mail         / mail.message             / record_name (char)            : not a function anymore
+mail         / mail.thread              / message_last_post (datetime)  : NEW
+## Nothing to do
+
+mail         / mail.group               / message_last_post (datetime)  : NEW
+mail         / res.partner              / message_last_post (datetime)  : NEW
+## TODO something with last_post in openupgrade80
```
arthru, line 15: This TODO looks like there is work needed, but you treated it in post_migration script. Could you change that message ?

bwrsandman:
@arthru Good catch!
That's fixed now.

paulius-sladkevicius:
:+1:

addons/mail/migrations/8.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
+---Fields in module 'mail'---
+mail         / mail.alias               / alias_contact (selection)     : NEW required: required, selection_keys: ['everyone', 'followers', 'partners'], req_default: everyone
+mail         / mail.alias               / alias_parent_model_id (many2one): NEW relation: ir.model
+mail         / mail.alias               / alias_parent_thread_id (integer): NEW 
+mail         / mail.mail                / recipient_ids (many2many)     : NEW relation: res.partner
```
legalsylvain, line 5: [need information, code review : test and review]
Is it normal if alias_parent_model_id (and alias_parent_thread_id) are not initialized ?
I tested realizing diff between 7.0 database migrated to 8.0 and 8.0 just-installed and the data of those 2 columns are not the same.
(in migrated database the two fields are empty)

hbrunn, line 5: I wouldn't know how to fill this columns during migration. I see that only used in hr_reruitment and crm, this modules will have to update their data accordingly.

bwrsandman:
Moving target branch

PR #15, opened 2014-06-02
Base (Work in progress / stable)

legalsylvain:


StefanRijnhart:
Thanks, looks good. Will merge this now to enable other work in the same file.

PR #16, opened 2014-06-02
Honor oldname

hbrunn:
Don't forget to regenerate the records on the target database to make this work


StefanRijnhart:
Tnx! I'll regenerate the analysis

PR #20, opened 2014-06-02
Event addon migration (nothing to do)

paulius-sladkevicius:
Field renaming should be handled by ORM

related: pull request [#16](https://github.com/OpenUpgrade/OpenUpgrade/pull/16)

### DIFF
--- a/addons/event/migrations/8.0.0.1/user_notes.txt
+++ b/addons/event/migrations/8.0.0.1/user_notes.txt
@@ -1,0 +1,1 @@
+Nothing to do for event, because fields are using 'oldname' attribute.


bwrsandman:
I have a few issues here, there is no `analysis_work` file, even if the field renaming is handled by the ORM, there is work to be done with at least:

```
event        / event.event              / message_last_post (datetime)  : NEW 
```

For that, you can use the following:

```
from openerp.openupgrade import openupgrade_80
openupgrade_80.set_message_last_post(cr, uid, pool, 'event.event')
```

paulius-sladkevicius:
Thanks to pointing that, yesterday I've missed message_last_post helper.

PR #21, opened 2014-06-02
Added note for base_action_rule

paulius-sladkevicius:


bwrsandman:
How is `selection_keys` being filled? Is there a default value?
Other than that, it looks good.

Just update `openerp/openupgrade/doc/source/modules70-80.rst` to reflect the coverage of this module.
You will need to rebase with upstream/master

```
git remote add upstream https://github.com/OpenUpgrade/OpenUpgrade.git
git pull --rebase upstream master
```

paulius-sladkevicius:
@bwrsandman there is no default value and question is which we should take.
From my view, we could use 'on_create_or_write', then additionally it's needed to clean field ['filter_pre_id', 'trg_date_id', 'trg_date_range', 'trg_date_range_type'], because of onchange_kind() function.

bwrsandman:
@komsas do that if it seems right.
You could even call `onchange_kind()`, I think.

paulius-sladkevicius:
Okay, I'll close this pull request and create a new, because it's not relevant now. Thanks for help.

PR #22, opened 2014-06-02
Added note for hr addon (Nothing to do)

paulius-sladkevicius:


hbrunn:
agreed :+1:

arthru:
In fact, message_last_post on hr.employee and hr.job could be initialized. See https://github.com/savoirfairelinux/OpenUpgrade/blob/mail/addons/mail/migrations/8.0.1.0/post-migration.py#L49 for an example.

paulius-sladkevicius:
@arthru you're right, thanks to pointing that.

PR #23, opened 2014-06-02
Add function add_xmlid to OpenUpgrade API

alexis-via:


openerp/openupgrade/openupgrade.py
```diff
@@ -180,6 +181,38 @@ def rename_xmlids(cr, xmlids_spec):
...
+    in OpenERP 6 but the user had already created this missing currency
+    by hand in it's 5.0 database. In order to avoid having 2 identical
+    currencies (which is in fact blocked by an sql_constraint), you have to
+    add the entry in ir_model_data before the upgrade.
+    """
+    # Check if it doesn't already exists
+    cr.execute(
+        "SELECT id, res_id, noupdate FROM ir_model_data WHERE module=%s AND "
+        "name=%s AND module=%s",
+        (module, xmlid, model, res_id))
```
StefanRijnhart: You have 2x module=%s, the last one looks like it should be model=%s. You can leave out the res_id argument.

openerp/openupgrade/openupgrade.py
```diff
@@ -180,6 +181,38 @@ def rename_xmlids(cr, xmlids_spec):
...
+    already_exists = cr.fetchone()
+    if already_exists:
+        if already_exists[1] != res_id or already_exists[2] != noupdate:
```
StefanRijnhart: Maybe don't update an existing entry, or make it conditional with an extra function argument 'update', defaulting to False. I would not expect an unconditional update like this.

openerp/openupgrade/openupgrade.py
```diff
@@ -180,6 +181,38 @@ def rename_xmlids(cr, xmlids_spec):
                      "WHERE module = %s and name = %s")
             logged_query(cr, query, tuple(new.split('.') + old.split('.')))
 
+def add_xmlid(cr, module, xmlid, model, res_id, noupdate):
```
StefanRijnhart: Could make noupdate a keyword argument defaulting to False

openerp/openupgrade/openupgrade.py
```diff
@@ -180,6 +181,38 @@ def rename_xmlids(cr, xmlids_spec):
...
+        if already_exists[1] != res_id or already_exists[2] != noupdate:
+            cr.execute(
+                "UPDATE ir_model_data SET res_id=%d AND noupdate=%s WHERE "
+                "id=%d", (res_id, noupdate, already_exists[0]))
+    else:
+        cr.execute(
+            "INSERT INTO ir_model_data (create_uid, create_date, "
+            "write_uid, write_date, date_init, date_update, noupdate, "
+            "name, module, model, res_id) "
+            "VALUES (%d, (now() at time zone 'UTC'), %d, "
```
StefanRijnhart: From the psycopg docs: the variables placeholder must always be a %s, even if a different placeholder (such as a %d for integers or %f for floats) may look more appropriate

StefanRijnhart:
Thanks BTW, looks good in spite of my nags!

bwrsandman:
Looks good.
Just a note: for the inserts and database manipulations, you might want to consider using `logged_query()` for better debugging.

openerp/openupgrade/openupgrade.py
```diff
@@ -180,6 +181,36 @@ def rename_xmlids(cr, xmlids_spec):
...
+    # Check if the XMLID doesn't already exists
+    cr.execute(
+        "SELECT id FROM ir_model_data WHERE module=%s AND name=%s "
+        "AND model=%s",
+        (module, xmlid, model))
+    already_exists = cr.fetchone()
+    if already_exists:
+        return False
+    else:
+        cr.execute(
```
bwrsandman: `logged_query()` would be better here so we can trace SQL calls

Example:

```
logged_query(cr, """
INSERT INTO ir_model_data (create_uid, create_date,
write_uid, write_date, date_init, date_update, noupdate,
name, module, model, res_id)
VALUES (%s, (now() at time zone 'UTC'), %s,
(now() at time zone 'UTC'), (now() at time zone 'UTC'),
(now() at time zone 'UTC'), %s, %s, %s, %s, %s)
""", (SUPERUSER_ID, SUPERUSER_ID, noupdate,
       xmlid, module, model, res_id))
```

StefanRijnhart:
Thanks for the changes. I think not having logged_query is fine for one off inserts.
:+1:

StefanRijnhart:
Just for the record, after a short discussion we agreed to have logged_query in such cases as well so :-1: until then

paulius-sladkevicius:
:+1:

bwrsandman:
Looks good, just a few pep8 errors.
You can check them on the link in "Details"

StefanRijnhart:
Tnx, pep-bot is green ;-) :+1:

PR #9, opened 2014-06-02
(WIP) Product v8 migration

gurneyalex:
This is work in progress, but it is running. 
Feedback welcome.

PR #29, opened 2014-06-03
Event addon migration

paulius-sladkevicius:
Related #20


bwrsandman:
Forgot to say in #20:
Just update `openerp/openupgrade/doc/source/modules70-80.rst` to reflect the coverage of the module.
You will need to rebase with upstream/master

```
git remote add upstream https://github.com/OpenUpgrade/OpenUpgrade.git
git pull --rebase upstream master
```

paulius-sladkevicius:
@bwrsandman I've added it

bwrsandman:
@komsas Thank you!
:+1:

addons/event/migrations/8.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,51 @@
+---Fields in module 'event'---
+event        / event.event              / date_tz (selection)           : NEW selection_keys: function
+event        / event.event              / description (html)            : NEW
+event        / event.event              / main_speaker_id (many2one)    : DEL relation: res.partner
+event        / event.event              / message_last_post (datetime)  : NEW
+# Used set_message_last_post()
+
+event        / event.event              / note (text)                   : DEL
+# Fields argument 'oldname' used, now points to 'description' field
+
```
StefanRijnhart, line 12: Please convert to html using openerp.tools.mail.plaintext2html

paulius-sladkevicius, line 12: @StefanRijnhart  Could you explain why it should be done?
If this manner everybody should do that? e.g. in the #9

paulius-sladkevicius, line 12: Please ignore last questions.

paulius-sladkevicius:
Added generic function that can convert text value to HTML.

bwrsandman:
@komsas nice function, but wouldn't it be more useful to put it in `openerp.openupgrade.openupgrade80`?

paulius-sladkevicius:
Firstly we should decide if it's worth to put there. Needs to check if there more places where it could be used.

bwrsandman:
@StefanRijnhart @hbrunn 

> wouldn't it be more useful to put `convert_field_to_html()` in openerp.openupgrade.openupgrade80?

legalsylvain:
Agree with @bwrsandman remarks.

StefanRijnhart:
As it is already  refactored out, we might as well have it in the API. I would make that openupgrade.py (not openupgrade_80.py) because this is not particularly 8.0 specific.

paulius-sladkevicius:
@StefanRijnhart in such case I'll try to move it to API

paulius-sladkevicius:
Sorry, looks by mistake I've merged 8.0 while resolving conflict, but this pull request points to master. I'll prepare a new clean pull request to 8.0.

PR #30, opened 2014-06-03
migrate.py flake8 compliant

hbrunn:


bwrsandman:
:+1: for pep8
The `migrate.py` script should have the executable flag set.
The shebang could be updated to the following for better portability:

```
#!/usr/bin/env python2
```

It will make sure it is run in the correct environment and use the correct version of python.
Using `python2` vs `python` is debated as there is a bug in some older virtualenv versions (such as in debian stable) so you can use `#!/usr/bin/env python` instead

hbrunn:
thanks, both of them will go to another pull request due soon

StefanRijnhart:
:+1:

PR #31, opened 2014-06-03
[ADD] add argument no_version for migrate decorator to force loading ...

bguillot:
In version 8.0, during a migration, if a new module has to be installed (due to new dependency for instance), pre- and post- migration files are not loaded anymore. 

It can be usefull to pass sometimes in those files, for instance to not create a new ir_model_data that have been move to the new module (that will be installed).

This pull request revert this change (so pre- and post- migration files will be taken into account as before v 8.0) and adds a new argument 'no_version' in the decorator 'migrate'.
If no_version is True, it will pass threw the method wrap in the decorator. By default it's False to keep the "official" behaviour.

### DIFF
--- a/openerp/modules/migration.py
+++ b/openerp/modules/migration.py
@@ -90,7 +90,7 @@ class MigrationManager(object):
             'post': '[%s>]',
         }
 
-        if not (hasattr(pkg, 'update') or pkg.state == 'to upgrade') or pkg.installed_version is None:
+        if not (hasattr(pkg, 'update') or pkg.state == 'to upgrade'):
             return
 
         def convert_version(version):
--- a/openerp/modules/migration.py
+++ b/openerp/modules/migration.py
@@ -89,7 +89,11 @@ class MigrationManager(object):
             'pre': '[>%s]',
             'post': '[%s>]',
         }
-
+        # In openupgrade, remove 'or pkg.installed_version is None'
+        # We want to always pass in pre and post migration files and use a new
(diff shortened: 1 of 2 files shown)


bwrsandman:
:+1: 
LGTM

openerp/modules/migration.py
```diff
@@ -90,7 +90,7 @@ def migrate_module(self, pkg, stage):
             'post': '[%s>]',
         }
 
-        if not (hasattr(pkg, 'update') or pkg.state == 'to upgrade') or pkg.installed_version is None:
+        if not (hasattr(pkg, 'update') or pkg.state == 'to upgrade'):
             return
```
StefanRijnhart, line 98: Please add a comment mentioning OpenUpgrade and the reason for this change

openerp/openupgrade/openupgrade.py
```diff
@@ -485,7 +485,7 @@ def message(cr, module, table, column,
     logger.warn(prefix + message, *argslist, **kwargs)
 
 
-def migrate():
+def migrate(no_version=False):
     """
```
StefanRijnhart, line 489: Please update the docstring with the meaning of this parameter.

StefanRijnhart:
Thanks, very useful! Only a couple of nags in the code.

bguillot:
Thanks for the review, I added the explanaitions in the code.

bwrsandman:
:+1:

PR #35, opened 2014-06-03
hr addon migration + add float_to_integer general openupgrade function

florian-dacosta:


openerp/openupgrade/openupgrade.py
```diff
@@ -458,6 +459,24 @@ def m2o_to_m2m(cr, model, table, field, source_field):
...
+    """
+    Change column type from float to integer. It will just
+    truncate the float value (It won't round it)
+
+    :param table: The table
+    :param field: The field name for which we want to change the type
+
+    .. versionadded:: 8.0
+    """
+    cr.execute('ALTER TABLE %(table)s '
```
bwrsandman: Can you use `logged_query()` for this, please.

addons/hr/migrations/8.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+hr           / hr.department            / jobs_ids (one2many)           : NEW relation: hr.job
+hr           / hr.employee              / message_ids (one2many)        : NEW relation: mail.message
+## OK new fields -> nothing to do
+hr           / hr.employee              / message_last_post (datetime)  : NEW 
+# Used set_message_last_post()
+hr           / hr.job                   / no_of_recruitment (float)     : type is now 'integer' ('float')
+hr           / hr.job                   / no_of_employee (float)        : type is now 'integer' ('float')
+hr           / hr.job                   / expected_employees (float)    : type is now 'integer' ('float')
+### Transform float to integer
+hr           / hr.job                   / message_last_post (datetime)  : NEW
```
bwrsandman: Nitpicking, but you technically this should be after L5

addons/hr/migrations/8.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+# Used set_message_last_post()
+hr           / hr.job                   / no_of_recruitment (float)     : type is now 'integer' ('float')
+hr           / hr.job                   / no_of_employee (float)        : type is now 'integer' ('float')
+hr           / hr.job                   / expected_employees (float)    : type is now 'integer' ('float')
+### Transform float to integer
+hr           / hr.job                   / message_last_post (datetime)  : NEW 
+hr           / hr.job                   / no_of_hired_employee (integer): NEW 
+hr           / hr.job                   / write_date (datetime)         : NEW 
+hr           / res.users                / display_employees_suggestions (boolean): NEW 
+## OK new fields -> nothing to do
```
bwrsandman, line 15: Are there ways to ascertain what these three fields should have as a value during the migration?
I haven't checked, but it seems like `no_of_hired_employee` would be a sequence?
`write_date` could probably be calculated
If `display_employees_suggestions` has a default value, it might be good to blanket set it to that value for all res.users, or False since it's a boolean.

bwrsandman:
Other than the few issues I pointed out, pep8 issues, this looks solid.
:+1:

addons/hr/migrations/8.0.1.1/openupgrade_analysis_work.txt
florian-dacosta, line 15: Thans for the feedback, 
<code>no_of_hired_employee</code> does not seem to be a sequence
The <code>write_date</code> field is still good after migration
About <code>display_employees_suggestions</code>, I tested and the default value is automatically set, so I guess there is no need to do anything else?

arthru:
:+1:

openerp/openupgrade/openupgrade.py
```diff
@@ -458,6 +459,26 @@ def m2o_to_m2m(cr, model, table, field, source_field):
         model.write(cr, SUPERUSER_ID, row[0], {field: [(4, row[1])]})
 
 
+def float_to_integer(cr, table, field):
```
hbrunn, line 462: I think we can make this more generic:
def alter_column_type(cr, table, field, type)

openerp/openupgrade/openupgrade.py
```diff
@@ -458,6 +459,26 @@ def m2o_to_m2m(cr, model, table, field, source_field):
...
+    :param table: The table
+    :param field: The field name for which we want to change the type
+
+    .. versionadded:: 8.0
+    """
+    logged_query(
+        cr,
+        "ALTER TABLE %(table)s "
+        "ALTER COLUMN %(field)s "
+        "TYPE integer" % {
```
hbrunn, line 476: and put the type from above here. then we can use this function for all conversions postgres can handle

StefanRijnhart, line 476: I think OpenERP already takes advantage of most of the type conversions that Postgres handles, so maybe we should not make this generic before we find another exception?

florian-dacosta, line 476: I agree there shouldn't be a lot of other cases, but I don't mind change it if needed.

paulius-sladkevicius:
:+1:

hbrunn:
okay, :+1:

PR #37, opened 2014-06-03
Product v8 migration

guewen:
Migrate products!


StefanRijnhart:
Amazing work, thanks! :+1:

addons/product/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,162 @@
...
+               #
+
+               ]
+    for sql in queries:
+        execute(cr, sql)
+
+
+def copy_fields(cr, pool):
+    product_tmpl= pool['product.template']
+    # copy the active field from product to template
```
StefanRijnhart, line 58: Oh, flake8 warns about the missing space here in line 57

guewen:
Yep, missed this one. Should be better now.

addons/mrp/migrations/8.0.1.1/post-migrate.py
```diff
@@ -0,0 +1,18 @@
+from openerp.openupgrade import openupgrade
```
bwrsandman, line 1: This file should be called `post-migration.py`
Also, there are pep8 issues in this script.

addons/stock_account/migrations/8.0.1.1/post-migrate.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
```
bwrsandman, line 1: File should be called `post-migration.py`

bwrsandman:
A few files have the wrong name. Have they been running?

StefanRijnhart:
Everything called [pre|post]-*.py will run. Non-blocking IMHO, we can clean up and rename according to some convention later on.

bwrsandman:
In that case, the .travis.yml file needs to reflect this. 
This PR has pep8 issues which weren't caught.

StefanRijnhart:
All is well on pep-bot after changing the file pattern :+1:

addons/mrp/migrations/8.0.1.1/post-migrate.py
```diff
@@ -0,0 +1,18 @@
...
+from openerp import pooler, SUPERUSER_ID
+
+def move_fields(cr, pool):
```
bwrsandman, line 4: This function is incomplete:
- not closed bracket
- unused variables

guewen:
@komsas the CI actually returned a green result because it was not considering a file in mrp, that's why I didn't caught the issues. This is the only file having issues, I will make a new PR (note that mrp still needs to  be migrated).

PR #45, opened 2014-06-03
Hr expense migration

arthru:


addons/hr_expense/migrations/8.0.1.0/post_migration.py
```diff
@@ -0,0 +1,70 @@
...
+
+def update_hr_expense_ok(cr, pool):
+    field_name = openupgrade.get_legacy_name('hr_expense_ok')
+    template_obj = pool.get('product.template')
+
+    openupgrade.logged_query(
+        cr,
+        """SELECT product_templ_id
+        FROM product_product
+        WHERE %s = 't';""",
```
hbrunn: doesn't this escape field_name?

addons/hr_expense/migrations/8.0.1.0/post_migration.py
```diff
@@ -0,0 +1,70 @@
...
+    template_obj.write(cr, SUPERUSER_ID, template_ids, {'hr_expense_ok': True})
+    for template_id in template_ids:
+        openupgrade.logged_query(
+            cr,
+            """SELECT DISTINCT t.id
+            FROM product_template t
+            LEFT JOIN product_product p1 ON t.id = p1.product_tmpl_id
+            LEFT JOIN product_product p2 ON t.id = p2.product_tmpl_id
+            WHERE p1.%s = 't'
+            AND p2.%s = 'f';""",
```
hbrunn: same here?

bwrsandman:
:+1:

addons/hr_expense/migrations/8.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,49 @@
+---Fields in module 'hr_expense'---
+
+hr_expense   / hr.expense.expense       / message_last_post (datetime)  : NEW 
+# Initialized field
+
+hr_expense   / hr.expense.expense       / state (selection)             : selection_keys is now '['accepted', 'cancelled', 'confirm', 'done', 'draft', 'paid']' ('['accepted', 'cancelled', 'confirm', 'done', 'draft']')
```
legalsylvain, line 6: For the new state 'paid' it would be interesting to check if the account.moves are reconciled or not in order to split hr_expense 'done' into 'done' or 'paid' state. :-1:

addons/hr_expense/migrations/8.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,49 @@
...
+# Nothing to do
+
+hr_expense   / hr.expense.expense       / voucher_id (many2one)         : DEL relation: account.voucher
+# Kept as legacy field
+
+hr_expense   / hr.expense.expense       / website_message_ids (one2many): NEW relation: mail.message
+# nothing to do
+
+hr_expense   / hr.expense.report        / voucher_id (many2one)         : DEL relation: account.voucher
+# Kept as legacy field
```
legalsylvain: It is usefull to keep a field on a osv.osv model with _auto = False ?
The view are created by init function. 

@StefanRijnhart : Why don't you think if we change the script that generates analysis.txt files, to avoid to log that case. (_auto = False).

hbrunn: there's a pull request to do exactly that

addons/hr_expense/migrations/8.0.1.0/pre_migration.py
```diff
@@ -0,0 +1,35 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.openupgrade import openupgrade
+
+
+column_renames = {
+    'hr_expense_expense': [('voucher_id', None)],
+    'hr_expense_report': [('voucher_id', None)],
```
StefanRijnhart: As Sylvain noted, hr_expense_report does not have a table but a view instead so this line will raise an exception. You can just remove it.

addons/hr_expense/migrations/8.0.1.0/openupgrade_analysis_work.txt
StefanRijnhart, line 6: Agreed :-1:

arthru, line 6: It seems the analyze was wrong : the state paid exists in both [OpenERP](https://github.com/odoo/odoo/blob/7.0/addons/hr_expense/hr_expense.py#L89) and [Odoo](https://github.com/odoo/odoo/blob/master/addons/hr_expense/hr_expense.py#L93)

StefanRijnhart, line 6: I see. It was added to OpenERP 7.0 some time after the release of OpenERP. I should probably update OpenUpgrade 7.0 to rerun the analysis. Unfortunately, I am having problems running it right now after the last refactoring :-( Thanks for investigating this. :+1: for the pull request.

legalsylvain:
Thanks. :+1:

PR #47, opened 2014-06-03
[REF] openupgrade_records: sorting key can be a tuple

florentx:
remove useless string formatting.

No change for the boolean field because: "False" < "True" and False < True


hbrunn:
:+1:

bwrsandman:
:+1:

PR #50, opened 2014-06-03
Simplify migration calling, fix breakpoint use

bwrsandman:
Don't write to a temp file
Use `load_module` instead of `load_source`

This allows the use of breakpoints in IDEs and in pdb

Signed-off-by: Sandy Carter sandy.carter@savoirfairelinux.com

### DIFF
--- a/openerp/modules/migration.py
+++ b/openerp/modules/migration.py
@@ -157,17 +157,14 @@ class MigrationManager(object):
                     name, ext = os.path.splitext(os.path.basename(pyfile))
                     if ext.lower() != '.py':
                         continue
-                    mod = fp = fp2 = None
+                    # OpenUpgrade edit start:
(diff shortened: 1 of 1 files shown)


hbrunn:
:+1:

the reason they do it that clumsily is that back in the days we had modules in zip files. That's not an issue for openupgrade

bwrsandman:
@hbrunn The real issue was `imp.load_source()`, I found out later.
The function hides the context frame of python, so breakpoints are lost.

StefanRijnhart:
Really good, thanks! As this is core code, could you mark this code area with a comment mentioning OpenUpgrade andd the reason it is changed?

legalsylvain:
Agree with @StefanRijnhart remarks. 

:+1: otherwise.

bwrsandman:
Comments added

PR #52, opened 2014-06-03
Fix pep8 with account_voucher migration

bwrsandman:


StefanRijnhart:
Thanks! Not waiting for the flake8 report on this one.

StefanRijnhart:
That sounded very ironic but it was meant for another pull request (adding an image)

PR #55, opened 2014-06-03
base_action_rule migration

paulius-sladkevicius:
related #21


arthru:
:+1:

addons/base_action_rule/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+##############################################################################
+
+from openerp.openupgrade import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    cr.execute(
+        "UPDATE base_action_rule SET kind = 'on_create_or_write', "
+        "filter_pre_id = null, trg_date_id = null, trg_date_range = null, "
```
StefanRijnhart: The value of this new field should depend on whether trg_date_id is filled. If yes -> on_time, if no -> on_create_or_write.

paulius-sladkevicius: @StefanRijnhart thanks, I've updated it.

legalsylvain:
Some Pep8 problems.

paulius-sladkevicius:
@legalsylvain something wrong with Travis, because I've tested it locally with falke8, no warning arises.

addons/base_action_rule/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,35 @@
...
+        "trg_date_range_type = null where trg_date_id = null"
+    )
+    cr.exceute(
```
bguillot: there is a typo : cr.execute() instead of cr.exceute().

About that should it be openupgrade.logged_query() instead of cr.execute() ? (I'm not sure)

paulius-sladkevicius: @bguillot  Thanks, looks that github editor is not best way to write code.

legalsylvain:
agree with @bguillot on logged_query use. 
:+1: otherwise.

paulius-sladkevicius:
@legalsylvain @StefanRijnhart @hbrunn if there is any rule when logged_query should be used?

arthru:
And don't forget to update status in openerp/openupgrade/doc/source/modules70-80.rst

legalsylvain:
Not sure, but I think that :
logged_query must be use when you do a modification on database by queries during upgrade. (insert / update / delete);
it's unnecessary for select requests;

addons/base_action_rule/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,35 @@
...
+
+from openerp.openupgrade import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.logged_query(cr,
+        "UPDATE base_action_rule SET kind = 'on_create_or_write', "
+        "filter_pre_id = null, trg_date_id = null, trg_date_range = null, "
+        "trg_date_range_type = null where trg_date_id = null"
```
StefanRijnhart: filter_pre_id should probably not be cleared. This is used in update triggers.

addons/base_action_rule/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    openupgrade.logged_query(cr,
+        "UPDATE base_action_rule SET kind = 'on_time', "
+        "filter_pre_id = null where trg_date_id != null"
+    )
```
StefanRijnhart, line 37: -- edit --: I see that the data in 7.0 is not kept consistent, so you can ignore the comment below

Just like filter_pre_id, you may want to clear filter_id because it is not used in time triggers.

If the data in 7.0 is consistant, maybe you don't need to clear any columns? Just set the 'kind' based on the value of trg_date_id.

bwrsandman:
:+1:

addons/base_action_rule/migrations/8.0.1.0/post-migration.py
paulius-sladkevicius: @StefanRijnhart [onchange_kind()](https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/base_action_rule/base_action_rule.py#L107) clear field if on_create_or_write or on_time is used.

StefanRijnhart: Yes, but we are not changing the kind. We are deriving the kind from the existing data and want to keep filter_pre_id if kind == 'on_create_or_write'.

addons/base_action_rule/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,37 @@
...
+
+from openerp.openupgrade import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.logged_query(
+        cr,
+        "UPDATE base_action_rule SET kind = 'on_create_or_write', "
+        "filter_pre_id = null, trg_date_id = null, trg_date_range = null, "
```
StefanRijnhart: My comment is now hidden after the code update so I'll repeat it here. Not sure if that is the best thing to do, so sorry for the spam.

filter_pre_id should not be cleared. This is used in update triggers.

paulius-sladkevicius: @StefanRijnhart so you say that it's a odoo bug? Because they clean it in [on_change method](https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/base_action_rule/base_action_rule.py#L110)

StefanRijnhart: Yes, I think the first condition in Odoo should be 'if kind == 'on_create'.

StefanRijnhart:
Thanks! :+1: now

PR #56, opened 2014-06-03
[ADD] l10n_es migration script.

pedrobaeza:
It deletes old account chart template data to allow a clean installation of new account chart template structure.


StefanRijnhart:
Are there any changes? I don't see an analysis file, which usually means that nothing has changed between the old and new version.

pedrobaeza:
That is because new l10n_es was merged last night:

https://github.com/odoo/odoo/pull/329

Regards.

StefanRijnhart:
Thanks, I will refresh the analysis tomorrow.

pedrobaeza:
Stefan, I'm wondering if this effort is going to be sterile due to the postponed date of v8 debut. What do you think? A lot of things can change...

StefanRijnhart:
No, Quentin told me that it was like a release candidate, with a freeze of the database scheme. So the number of changes still affecting OpenUpgrade should be quite small. And we can check the diff of the analysis files with earlier versions to tell us what those changes are.

pedrobaeza:
OK, what a relief!

paulius-sladkevicius:
:+1:

bwrsandman:
:+1:

StefanRijnhart:
Wait, don't merge yet...

StefanRijnhart:
I see that you remove templates, a sequence and a todo action. There are problems with this:

Templates: these are noupdate=0 in 7.0 and will be removed automatically.
Todo action: same.
Sequence: removing a sequence seems unwise.

We really need the updated analysis to process any other changes.

:-1:

pedrobaeza:
Hi, Stefan, the problem is that due to the way Odoo updates values and the ondelete="set null" clause, it provokes an error when you try to update the module. That's why I cut to the chase removing all entries corresponding to l10n_es.

Regards.

StefanRijnhart:
If you remove all template entries, but keep the sequence and the todo action, would you still get an error?
I see in the latest analysis file that the module switches out account types. Do you need to migrate the type of existing accounts?

pedrobaeza:
Hi, Stefan, sorry for the late reply. As there is the duality between account chart templates and installed account charts, this script takes the most conservative approach: remove templates for avoiding errors, but keep the other things (account types included, because there is no exact mapping between old and new codes). Installed charts are not affected, and new companies can use the new chart. Old ones still have the chance to be updated with account_chart_update.

pedrobaeza:
@StefanRijnhart, can we merge this then?

StefanRijnhart:
I see, the Spanish localization has its template definitions in noupdate stanzas. So removing them is appropriate. To prevent further confusion, please add a note saying that this is the reason you remove the records manually, so that people do not apply this practice on localizations with template definitions which are not noupdate.
Removing a sequence still seems like a very bad idea. Note that I don't see any sequence being defined in l10n_es 7.0, so you might as well remove this and prevent yourself from setting a bad example.
Looking at the account types, it seems you can still convert quite a few that can be mapped directly, just not all. Having all these duplicate account types is annoying when configuring a new journal account.

pedrobaeza:
Superseeded by #199, because I have lost previous branch. @StefanRijnhart, please continue your review there.

PR #59, opened 2014-06-03
Fix #58 : fix import in email_template's pre_migration script

### LINKED ISSUE #58: Import problem in email_template migration
As pointed by @komsas  in #12 , an import in pre-migration.py is wrong. It should be 

```
from openerp.openupgrade import openupgrade
```

instead of

```
from openupgrade import openupgrade
```

arthru:
This fixes #58

### DIFF
--- a/addons/email_template/migrations/8.0.1.1/pre-migration.py
+++ b/addons/email_template/migrations/8.0.1.1/pre-migration.py
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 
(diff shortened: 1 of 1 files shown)


legalsylvain:
:+1:  Thanks.

bwrsandman:
:+1:

StefanRijnhart:
Also included in #37 but I'm guessing that does not hurt. :+1:

PR #62, opened 2014-06-04
[FIX] Syntax error and pep8, add work file

StefanRijnhart:


legalsylvain:
:+1:  Thanks.

addons/mrp/migrations/8.0.1.1/post-migrate.py
```diff
@@ -1,18 +1,35 @@
...
-                       WHERE product_template.id=product_product.product_tmpl_id)
-    """ % openupgrade.get_legacy_name('produce_delay'),
+
+def move_fields(cr):
+    openupgrade.logged_query(
+        """UPDATE product_product
+           SET produce_delay=(SELECT pt.%s
+           FROM product_template
+           WHERE product_template.id=product_product.product_tmpl_id)
+    """ % openupgrade.get_legacy_name('produce_delay'))
```
hbrunn, line 30: better update on a self join for performance

bwrsandman:
Sorry, I was trying to fix that concurrently.
:+1: to this!!!

StefanRijnhart:
Attempted Holger's suggestion of the join query.

StefanRijnhart:
Why does the diff show the original contributution of the product module? Was that change reverted somehow?

legalsylvain:
@StefanRijnhart : Maybe you could try some commands like : 
git rebase --random --iamlucky

More information on : http://git-man-page-generator.lokaltog.net/

Regards.

bwrsandman:
@StefanRijnhart yes it was since it was breaking all other PRs and with this PR on the way, it's not like we were going to lose the commits.
That said, I won't try any of that git black magic again...
The commits are in 8.0 and they have the same SHAs, so you won't have any conflicts.
:+1:

StefanRijnhart:
OK thanks. That is not something which you can do in bzr, you see.

StefanRijnhart:
@legalsylvain LOL but only because I know what that is ;-)

StefanRijnhart:
I guess https://github.com/OpenUpgrade/OpenUpgrade/pull/73/files makes this request redundant (even if I was first :-P )

PR #64, opened 2014-06-04
Add pre and post migration scripts for mail

bwrsandman:
Reopened #14 PR against 8.0 instead of master


addons/mail/migrations/8.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
+---Fields in module 'mail'---
+mail         / mail.alias               / alias_contact (selection)     : NEW required: required, selection_keys: ['everyone', 'followers', 'partners'], req_default: everyone
+mail         / mail.alias               / alias_parent_model_id (many2one): NEW relation: ir.model
+mail         / mail.alias               / alias_parent_thread_id (integer): NEW
```
StefanRijnhart, line 4: So the new alias_parent\* fields enable a button 'open_parent_document' from the alias view. This allows quick navigation from a task alias to the associated project. I guess the values for these fields will have to be filled in their respective modules. It looks like the create method of the models show us with which values to fill them. I find these fields in

crm/res_config.py
crm/sales_team.py
hr_recruitment/hr_recruitment.py
hr_recruitment/res_config.py
mail/mail_group.py
mail/res_users.py
project/project.py

See the project migration that I now adapted: 0ab9e6961b

You should do similarly for the mail.group and res.users model, because their aliases are created in this module.

bwrsandman:
Rebased, but still waiting on code from #80

hbrunn:
superseded by #116

PR #65, opened 2014-06-04
[ADD] Migration scripts for the project addon

StefanRijnhart:


addons/project/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+    legacy_priority = openupgrade.get_legacy_name('priority')
+    for old, new in [(4, 0), (3, 0), (2, 1), (1, 2), (0, 2)]:
+        openupgrade.logged_query(
+            cr, """
+            UPDATE project_task SET priority = %s
+            WHERE """ + legacy_priority + " = %s",
+            (new, old))
+
+
+def update_alias_parent(cr, registry):
```
bwrsandman: Could we move this to openupgrade80.py and make it more general?
https://github.com/OpenUpgrade/OpenUpgrade/pull/64#discussion_r13404465

StefanRijnhart: Hi Sandy, I refactored the method that is shown in the highlighted code in this github thread. Unfortunately, I now realize that is not what you mean. It is simply context code, and you meant to comment on the very last line, which is the start of the update_alias_parent method :-( But the mapping of priorities is actually featured in 5 modules in 8.0, and I made the method simply generic so I hope it can stay.

About updating alias parent settings, I have my doubts. The settings self are a bit obscure, and they vary from model to model. Let's wait until other models are adapted to have a clear view of the common code.

bwrsandman: How do you suggest I proceed with mail, another model?

StefanRijnhart: Good question. Refactored anyway, hoping that this does not raise more questions than it answers.

bwrsandman:
:+1:

paulius-sladkevicius:
:+1:

addons/project/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,48 @@
...
+        cr,
+        openupgrade.get_legacy_name('priority'),
+        'priority',
+        [(4, 0), (3, 0), (2, 1), (1, 2), (0, 2)],
+        table='project_task', write='sql')
+
+    openupgrade_80.update_aliases(
+        cr, registry, 'project.project',
+        set_parent_thread_id=True,
+        default_id_key='project_id')
```
red15, line 41: Should be defaults_id_key='project_id'

openerp/openupgrade/openupgrade_80.py
```diff
@@ -71,3 +74,37 @@ def set_message_last_post(cr, uid, pool, models):
...
+    """
+    model_id = registry['ir.model'].search(
+        cr, SUPERUSER_ID, [('model', '=', model_name)])[0]
+    vals = {'alias_parent_model_id': model_id}
+    if defaults_id_key and alias_defaults is None:
+        alias_defaults = {}
+    res_ids = registry[model_name].search(
+        cr, SUPERUSER_ID, [], context={'active_test': False})
+    for res in registry[model_name].search(
+            cr, SUPERUSER_ID, res_ids):
```
red15, line 103: When searching for res_ids it gets a list of [1,2,3].
Right now it will try to search on this directly.
Needs to change to 

```
cr, SUPERUSER_ID, [('id', 'in', res_ids)]
```

openerp/openupgrade/openupgrade.py
```diff
@@ -478,6 +483,60 @@ def float_to_integer(cr, table, field):
...
+            'new': new,
+        }
+        if write == 'sql':
+            query = """UPDATE %(table)s
+                       SET %(target)s = %%(new)s
+                       WHERE %(source)s = %%(old)s""" % values
+        else:
+            query = """SELECT id FROM %(table)s
+                       WHERE %(source)s = %%(old)s""" % values
+        logged_query(cr, query, values)
```
red15, line 532: When write type == 'sql' the values passed to logged_query should actually remain empty, the query has already been mogrified.

I suggest you actually also use the cr.mogrify("""UPDATE ...""", values) function after first filling in the table/target/source (but replace the %%(new)s with regular %%s or something ?)

red15, line 532: This is why I originally made the issue #79 but it can thus be ignored if we can remerge this branch after fixing it.

openerp/openupgrade/openupgrade_80.py
red15, line 103: I was incorrect, the correct fix here is to call browse (instead of search on line 102)

bwrsandman:
@red15 I can reset to the commit before the merge, but this PR cannot be reopened.
I recommend making a new PR with the fixes.
If @StefanRijnhart wants to reopen this PR, I can take out the merge.

red15:
@bwrsandman I've requested a pull into @StefanRijnhart 's branch, once that gets merged he should also be able to request a new pull based on that branch.

PR #53, opened 2014-06-03
[ADD] rescue the logo from the lp repository

florentx:
It fixes a broken link in the documentation.

... until we change its color

PR #68, opened 2014-06-04
[FIX] Restore image in original location

StefanRijnhart:
Refer to #53


pedrobaeza:
I have checked "openerp/openupgrade/doc/source/index.rst", and path looks good, so :+1:

florentx:
:+1:  (my mistake)

PR #70, opened 2014-06-05
Event addon migration

paulius-sladkevicius:
Related #29
Remake to point to 8.0 branch


addons/event/migrations/8.0.0.1/post-migration.py
```diff
@@ -0,0 +1,32 @@
...
+from openerp.openupgrade.openupgrade import migrate, convert_field_to_html
+from openerp.openupgrade.openupgrade_80 import set_message_last_post
+
+
+@migrate()
+def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
+    set_message_last_post(
+        cr, SUPERUSER_ID, pool, ['event.event', 'event.registration'])
+    convert_field_to_html(cr, 'event_event', 'note', 'description')
```
StefanRijnhart: The custom is to rename the abandoned column name ('note' in this case) to its legacy name in the pre-script, then refer to this name in the post-script. While this is only necessary when the column type changes while keeping its name, I think it would be good to do this for any abandoned column that is touched by OpenUpgrade.

paulius-sladkevicius: @StefanRijnhart I thought if fields 'oldname' func parameter was used we don't need to make it legacy by OpenUpgrade, so I don't see point to create additional legacy field when we don't touch e.g. 'note' field. Otherwise we should do that for all fields that are using 'oldname'.

StefanRijnhart: If oldname is used, it is even worse because then you will not have the column anymore when you try to access it in this code....

bwrsandman:
~~This PR will need to be rebased.~~

StefanRijnhart:
:+1:

legalsylvain:
:question:
Hi, 
I don't use this module, so maybe my question is not relevant.
In 7.0: there is a main_speaker_id (res.partner) : 'DEL';
in 8.0: there is a organizer_id (res.partner) : 'NEW;
OK, a speaker is not an organizer but maybe it could be better to rename main_speaker_id to organizer_id, to avoid to loose data for the users that use that module, even if the semantic of the fields has changed.

What do you think ?

(Otherwise: LGTM)

paulius-sladkevicius:
@legalsylvain I think it will only confuse user, if somebody will need to keep main_speaker_id then better to inherit event.event and add such field.

paulius-sladkevicius:
Could somebody help to close this PR?
Sorry I can blame myself, because I'm not active now.

legalsylvain:
Ow. sorry @komsas , I forgot this review. Thanks for the explanation. and :+1:

bwrsandman:
Needs to be rebased again

pedrobaeza:
:+1:

paulius-sladkevicius:
@bwrsandman where I need to rebase it? Can you give more details?

arthru:
@komsas it seems that there are conflicts between your branch and the official one.

You simply need to git pull --rebase from the official branch and then git push (possibly with --force).

paulius-sladkevicius:
@arthru thanks, I didn't notice in first place that there are conflicts. Travis green text gave me green light ;-)

PR #71, opened 2014-06-05
[ADD] Migration of the procurement references on the sale order line

StefanRijnhart:
I'm taking on sale, but it contains a change that needs migration in sale_stock. Requesting a pull of this atomic change to prevent conflicts when someone wants to work on sale_stock.

### DIFF
--- a/addons/sale_stock/migrations/8.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/sale_stock/migrations/8.0.1.0/openupgrade_analysis_work.txt
@@ -1,4 +1,12 @@
 ---Fields in module 'sale_stock'---
+# This module has not yet been migrated, except the one change below,
+# in preparation of the migration of the sale module
+
+# Now a one2many in sale. Migrated in this module.
+sale_stock   / sale.order.line          / procurement_id (many2one)     : DEL relation: procurement.order
+
+# TODO: process all other changes below
+
 sale_stock   / sale.order               / invoice_quantity (False)      : DEL selection_keys: ['order', 'procurement'], mode: modify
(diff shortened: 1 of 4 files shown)


StefanRijnhart:
Additionally, this change contains a renaming of the m2o_to_m2m API method to m2o_to_x2m because it works just as well in the case of migration to a one2many. Keeping the old function name as a reference for backwards compatibility.

legalsylvain:
:+1: LGTM.
Maybe you could mark some "TODO" notes in "openupgrade_analysis_work.txt" or a "work in progress" in the file "migration_70-80.rst" to prevent people believe that 'sale_stock' is done.

Regards.

paulius-sladkevicius:
:+1: 
I don't think it's worth to pull in progress to "migration_70-80.rst", it will confuse that somebody is working on it.

bwrsandman:
@StefanRijnhart Can we get an answer to this?
Do we need a work in progress note?

StefanRijnhart:
It should be clear from the 'work' file that only one aspect has been handled, right?

bwrsandman:
Quickly looking at it, I can say that it's not immediately clear.

StefanRijnhart:
OK, WIP as requested

legalsylvain:
@komsas No because if there is no people in the wiki page that means that nobody is working on that.
(https://github.com/OpenUpgrade/OpenUpgrade/wiki/Pending-Work)
Don't you think it is sufficient ?

StefanRijnhart:
I consider this change a part of the sale module really so I don't think we should mark the module coverage. But I will mark the work file.

StefanRijnhart:
Resubmit

pedrobaeza:
:+1:  LGTM

PR #72, opened 2014-06-05
[To review] 'Sales_team' migration.

legalsylvain:
**For code review**, this is Ok. 
**For functionnal test**
- the migrated database is unstable for many objects because to test 'sales_team' migration, we have to install in 7.0 'crm', whom migration is not covered yet;


bguillot:
Looks good, thanks !
:+1:

StefanRijnhart:
Looks good, but this module is  not installed by default automatically in all cases even if autinstall = True because of an interesting dependency called web_kanban_sparkling. This is a dependency of stock and mass_mailing only, which means that sales_team is only installed if either one of these modules is installed as well. I think it should be installed by default but I am afraid that there is no clear way to currently do that. If we don't cover this, then people will install their upgraded database on Odoo and lose their sales teams at the very first instance.

The cleanest way is to select the module for installation is probably in the base module's pre-script. In pseudo code: if state of the crm module is 'installed', set state of module sales_team to 'to install'.

legalsylvain:
@StefanRijnhart 

```
If we don't cover this, then people will install their upgraded database on Odoo 
and lose their sales teams at the very first instance.
```

I don't think so.
in V7, sales team (crm.case.section) model is defined in crm module.
in V8, sales team model is defined in 'sales_team' module and overloaded in 'crm';

So, in v7, if users have sales_team datas, they have 'crm', installed, 
in v8, 'sales_team' module will be installed by regular update progress.

I tested the migration with crm installed and some crm_case_section datas and my data are not lost.

Am I wrong ?

Regards.

StefanRijnhart:
Sorry, I was confused. Nothing will get lost because sales_team is a dependency of crm. :+1: for the pull request then.

PR #75, opened 2014-06-05
[8.0] 'point_of_sale' migration

legalsylvain:
Do not review for the moment.


kirca:
Hi,

Just to give a heads up that I'll start working on this. If someone is also doing the same let me know so that we don't duplicate work. Thanks.

kirca:
Hi @legalsylvain,

I've made a PR (https://github.com/legalsylvain/OpenUpgrade/pull/1) which finishes the POS migration, hopefully :).

Please take a look at it when possible. Thanks.

kirca:
Just a ping since this has been waiting for review some time now.

addons/point_of_sale/migrations/8.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,210 @@
...
+            pc.iface_cashdrawer or pc.iface_payment_terminal
+            or pc.iface_electronic_scale or pc.iface_print_via_proxy
+        ):
+            pc.write({'proxy_ip': 'http://localhost:8069'})
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    pool = RegistryManager.get(cr.dbname)
+    openupgrade.move_field_m2o(
```
pedrobaeza, line 187: You can encapsulate these calls to `move_field_m2o` for clarity in a method. You can even iterate through a list of the fields and make the call to the method with that variable as argument.

pedrobaeza:
Just a little comment, but seems :+1:

addons/point_of_sale/migrations/8.0.1.0.1/post-migration.py
legalsylvain, line 187: Yes It could be great. Thanks for your remark.
But in another PR, because move_field_m2o is used in many modules, and it's not the scope of this PR.

StefanRijnhart:
:+1:, thanks!

PR #81, opened 2014-06-19
[UPD] 8.0 analysis without _auto=False

StefanRijnhart:
This analysis ignores models with  _auto=False. There is a puzzling false positive stating that fields from mail.thread moved to website_mail, but that is because mail.thread (an AbstractModel from the mail module) is accidentally overriden in website_mail as a regular Model). I dislike the fact that there is now no clear overview of the changes in abstract models such as mail.thread, but the changes are in every model that inherits from it so I guess that will have to do.


bwrsandman:
This is a huge diff and it's code I am not familiar with.
Though I give my :+1: , keep that in mind.

legalsylvain:
:+1: 
Nice, all the bad analysis ('x.report' models) has been droped.
Thanks a lot.

pedrobaeza:
LGTM, but I still don't see l10n_es analysis file.

For the rest, :+1: 

Regards.

StefanRijnhart:
Correct, Pedro. I first want to pull the changes in the analysis due to a change in the analysis itself separately. At this very moment I am preparing another pull request to reflect the recent changes in OpenERP itself, such as l10n_es. Thanks for the review.

PR #83, opened 2014-06-20
[FIX] Generate missing XML only analysis files

StefanRijnhart:
First commit fixes a bug that prevented analysis files to be generated for modules with only changes in the set of XML ids (no changes in the database scheme).

The second commit adds these files, and rehashes the contents of the general analysis file containing information on dropped modules. Should not affect the work so far, or cause a lot of work anyway. Now, the l10n_es analysis is included, but again, this is still based on Odoo git of 3rd of June, without the latest refactoring by Pedro. Once this is merged, I can merge upstream Odoo and refresh all modules with the small number of recent, actual changes.

An anomality in the analysis added here is that l10n_be is now present, and listed as 'not installed in 8.0' in the general analysis file. That is only because l10n_be is incompatible with the analysis process because this process requires the modules to be reïnstalled on the analysis database. The l10n_be module breaks hard when being installed twice on the same database, so it has been set to uninstallable in OpenUpgrade 8.0 for now. Seems we had a different way of dealing with this in 7.0.


openerp/addons/openupgrade_records/model/analysis_wizard.py
```diff
@@ -126,14 +124,19 @@ def write_file(module, version, contents,
...
         res_xml = compare.compare_xml_sets(
             remote_xml_records, local_xml_records)
 
+        affected_modules = list(
+            set(
+                [
+                    record['module'] for record in
+                    remote_records + local_records +
+                    remote_xml_records + local_xml_records
+                    ]))
```
bwrsandman: Putting on my pythonista hat here:
You don't need to put a list generator in the set here, you can put a set generator here directly, it will consume less memory and perform less computation.

```
set([i for i in x]) -> set(i for i in x)
```

`affected_modules` seems to have an identity crisis here: lists turned into a list, turned into a set, turned into a list.
Poor guy.

bwrsandman:
Seems trivially correct.
:+1:

openerp/addons/openupgrade_records/model/analysis_wizard.py
StefanRijnhart: Removed the inner list generator so it's now just that poor man's uniqueness operator list(set()).

legalsylvain:
:-1: :+1: (means "needs information")

I'm not sure that remove analysis file when there are only xml_ids changes is correct. 
Theorical sample:
from V7 to V8, nothing change except a rename of an xml_id. So in the analysis file, it will be written only : 
old_xml_id: DEL
new_xml_id: NEW

So in that case, a call the function openupgrade.rename_xmlids() is necessary. It isn't ?

Regards.

StefanRijnhart:
@legalsylvain That is not what is happening here. There was a bug in the analysis process. If a module only contained changes in the set of XML ids, there was no analysis file. This commit re-adds them. So we totally agree!

-EDIT- In fact it was even worse: the analysis would also be missing if there were only schema changes but judging by a quick glance at the diff, that seems not to have happened, at least in this edition.

legalsylvain:
Oh. I'm confused.

In that case. :+1:

I really have to execute: `git improve --level=english --target=myself`

bwrsandman:
> :-1::+1: (means "needs information")

How about :question: (: question :)?

When did I become the emoji guy? :disappointed:
http://www.emoji-cheat-sheet.com/

hbrunn:
:+1:

PR #87, opened 2014-07-21
Fix errors on project_issue migration

dreispt:
Fixed errors resulting in exceptions when trying to upgrade my database.

Notably, I can't find the `openupgrade.map_values()`  function anywhere in the source code. Maybe it was missing from some commit?

### DIFF
--- a/addons/project_issue/migrations/8.0.1.0/pre-migration.py
+++ b/addons/project_issue/migrations/8.0.1.0/pre-migration.py
@@ -21,7 +21,7 @@
 
 from openerp.openupgrade import openupgrade
 
-column_renames = {'project_issue': ['priority']}
+#column_renames = {'project_issue': ['priority']}
 
 xmlid_renames = [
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
map_values is added in #80, which was reverted but has now been waiting for approval for a long time. I think that means you can uncomment the lines that deal with the priority field, right?

addons/project_issue/migrations/8.0.1.0/pre-migration.py
```diff
@@ -21,7 +21,7 @@
 
 from openerp.openupgrade import openupgrade
 
-column_renames = {'project_issue': ['priority']}
+#column_renames = {'project_issue': ['priority']}
```
StefanRijnhart, line 25: Should be {'project_issue': [('priority', None)]}. Sorry for my mistakes!

hbrunn:
this is done in #128 now

Issue #84, opened 2014-07-06
ONESTEiN Contribution


diellenne:
In order to accelerate the development and increase the coverage of OpenUpgrade v8 in the next few days our team will start to contribute to the project: in particular, we would like to complete as soon as possible the following modules:
- account
- delivery
- mail
- mrp
- portal
- portal_sale
- procurement
- product
- purchase
- report_intrastat
- sale
- sale_stock
- stock

Concerning the modules that are already in progress, we would like to hear from everyone who is actually working on these in order to share the missing work; for the others I will update the wikipage with our planning as soon as possible.

legalsylvain:
Hi, 

Glad to see that you're interested by this project. Yes, the wiki is not very up to date. Com' on guys, plz update the list with the current state! (@arthru, ccomb, c2c, @StefanRijnhart, ...)

Do you have some questions about how OpenUpgrade works, or how to contribute ?
Don't hesitate to ask.

About the list of your module you want to cover, the most important work is in 'stock & co' for the 7.0 -> 8.0 migration. Plz, see this post #82  and contact Alexandre if you want more information about the changes.

Regards.

diellenne:
Thank you,

about OpenUpgrade we know how to use it. We started migrating some small modules and we felt quite comfortable with the process.

About the contribution, as ONESTEiN we are now starting to be involved in the effective contribution to the Odoo ecosystem; for now we forked the OpenUpgrade project and we are going to take it as a basis to start working on the missing modules; so we will make our pull-requests to merge the code in the 'official' OpenUpgrade from our fork. For all the information related to the progress and the technical discussions we will refer directly to the pages of the main project. Do you think that might be a good way to work? 

Regards,
D

diellenne:
Hi,

I updated the 'planned migrations' wikipage with the modules that we are going to migrate by september 2014.

@legalsylvain: I saw that for some of these you had already planned the migration for 2015 ... I reinserted them in the table assigned to us but, if you want to collaborate, we would be glad to work with you in the implementation of the scripts. 

Regards.

plaes:
@diellenne Will you have a public branch of your efforts? I would really like to cooperate with `mrp` and `stock` modules.

diellenne:
@plaes I'm glad that you want to cooperate with us to improve OpenUpgrade faster. We landed on GitHub as onseteinbv just from a little so we are trying to organize our repositories in the best way to share our work.
[...]

PR #91, opened 2014-08-05
RFC: Add basic Odoo resource xml checker

plaes:
This script compares module resource xml files against data in Odoo
database and prints out various diagnostics.

Currently only supports `ir_actions_act_window` data objects.

### DIFF
--- a/scripts/audit_module_xml.py
+++ b/scripts/audit_module_xml.py
@@ -1,0 +1,137 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+"""
+    audit_module_xml
+    ~~~~~~~~~~~~~~~~
+
+    Compares Odoo module xml data files against Odoo database content
+    providing various diagnostic output.
+
+    Usage::
+
+        audit_module_xml.py path-to-odoo-module db-connection-uri
+
+    Where db-connection-uri has following form::
+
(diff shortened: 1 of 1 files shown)


plaes:
Fixed the pep8 issues and push-forced.

StefanRijnhart:
Hi @plaes, thanks! Can you tell me what problem is solved by this (you mention 'broken xml stuff' in #84, but what do you mean by this?)?

plaes:
@StefanRijnhart, this was a helper utility that helped me to find some broken data (and figure out how odoo itself works) in the `ir_actions_act_window` objects - things that caused errors after upgrade. One example is this: https://github.com/onesteinbv/OpenUpgrade/commit/ac321bdc71c71ffb3ef0872b8bf567436ad8bb65

The `domain` issue has since been fixed upstream:
- https://github.com/odoo/odoo/commit/bfb29230b78197aee98b694645787e6ee807418e

StefanRijnhart:
OK, I see. Looks like you could have used the existing scripts/compare_noupdate_xml_records.py to compare the records. Your use case is one of the reasons that it was developed. You use it to compare xml data from two versions of the same module and it prints a file that you can load in your migration script. It also lists updates to noupdate xml entries.

PR #92, opened 2014-08-06
Added migration of the notify_email field.

ajite:


pedrobaeza:
Hi @ajite, thanks for the migration script. 

Is this migration completed? Or is only a partial one?

We also modify documentation when we make a change to reflect these changes in two places:
- We create a file called `openupgrade_analysis_WORK.txt` that is a copy of the similar one on migration folder. In it, we annotate our "exploration" result. You can see an example here: https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/openupgrade_analysis_WORK.txt
- We modifiy the documentation that it's compiled on the web to reflect new module status. This is located here:
  https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/openerp/openupgrade/doc/source/modules70-80.rst

ajite:
Hi @pedrobaeza, It's a partial migration. I will migrate some more tomorrow and I will update the documentation. Thank you for your advices.

addons/mail/migrations/8.0.1.0/openupgrade_analysis_WORK.txt
```diff
@@ -0,0 +1,54 @@
...
+mail         / mail.mail                / mail_server_id (many2one)     : DEL relation: ir.mail_server
+mail         / mail.message             / mail_server_id (many2one)     : NEW relation: ir.mail_server
+# OK mail_server_id moved up in mail_message
+
+mail         / mail.mail                / reply_to (char)               : DEL 
+mail         / mail.message             / reply_to (char)               : NEW 
+# OK reply_to moved up in mail_message.
+
+
+mail         / mail.message             / record_name (char)            : not a function anymore
```
ajite, line 25: The old function was a stored char.

addons/mail/migrations/8.0.1.0/openupgrade_analysis_WORK.txt
```diff
@@ -0,0 +1,54 @@
...
+mail         / mail.alias               / alias_contact (selection)     : NEW required: required, selection_keys: ['everyone', 'followers', 'partners'], req_default: everyone
+mail         / mail.alias               / alias_parent_model_id (many2one): NEW relation: ir.model
+mail         / mail.alias               / alias_parent_thread_id (integer): NEW
+# Ok. Nothing to do.
+
+
+mail         / mail.group               / message_last_post (datetime)  : NEW 
+# This new value is going to be computed in the post migration script.
+
+mail         / mail.mail                / email_from (char)             : DEL
```
ajite, line 11: The email_from field in 7.0 existed in both mail.mail and mail.message. In both case it was a char.
The visible email_from in the Odoo 8.0 mail.mail form is the one inherited from mail.message.
I believe that we should just drop it.

pedrobaeza, line 11: You can add this explanation on the file itself for anyone reading it to understand that there's nothing to do.

addons/mail/migrations/8.0.1.0/openupgrade_analysis_WORK.txt
pedrobaeza, line 25: OK. Nothing to do also.

addons/mail/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,53 @@
...
+    ]
+    for sql in queries:
+        execute(cr, sql)
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
+    move_fields(cr, pool)
+    openupgrade.move_field_m2o(
```
pedrobaeza, line 43: If you do this in post-migration, I think the column gets lost in standard update process. If it so (please check it), you will have to rename column to None on pre-migration, and do this operation calling to openupgrade_get_legacy_name (approximated name of the function, I don't remember exactly) for getting the new column name.

addons/mail/migrations/8.0.1.0/openupgrade_analysis_WORK.txt
```diff
@@ -0,0 +1,54 @@
...
+
+mail         / mail.message             / record_name (char)            : not a function anymore
+mail         / mail.message.subtype     / hidden (boolean)              : NEW 
+mail         / mail.message.subtype     / sequence (integer)            : NEW 
+mail         / mail.thread              / message_follower_ids (many2many): module is now 'website_mail' ('mail')
+mail         / mail.thread              / message_ids (one2many)        : module is now 'website_mail' ('mail')
+mail         / mail.thread              / message_is_follower (boolean) : module is now 'website_mail' ('mail')
+mail         / mail.thread              / message_summary (text)        : module is now 'website_mail' ('mail')
+mail         / mail.thread              / message_unread (boolean)      : module is now 'website_mail' ('mail')
+# Ok. Nothing to do.
```
pedrobaeza, line 33: This is needed to check if in the standard process, columns are removed in `mail` module update and added again updating `website_mail` module.

bwrsandman:
Duplicate of  #64

Make sure you're not overlooking things.

There are some functions missing.

plaes:
@bwrsandman What's left todo in #64?

Also running into some errors with this:

```
2014-08-19 07:26:59,904 20362 DEBUG v7xxx_migrated OpenUpgrade: Running
        UPDATE res_partner
        SET notify_email = 'always'
        WHERE notify_email != 'none'

2014-08-19 07:26:59,904 20362 DEBUG v7xxx_migrated OpenUpgrade: 28 rows affected
2014-08-19 07:26:59,907 20362 ERROR v7xxx_migrated openerp.sql_db: Programming error: syntax error at or near ")"
LINE 1: ...ble.mail_message_id=new_table.id    LIMIT 1)  WHERE id in ()
                                                                      ^
, in query  UPDATE mail_message as new_table SET reply_to=(    SELECT old_table.reply_to    FROM mail_mail as old_table    WHERE old_table.mail_message_id=new_table.id    LIMIT 1)  WHERE id in %s
2014-08-19 07:26:59,907 20362 ERROR v7xxx_migrated OpenUpgrade: mail: error in migration script /home/plaes/code/OpenUpgrade/8.0/addons/mail/migrations/8.0.1.0/post-migration.py: syntax error at or near ")"
LINE 1: ...ble.mail_message_id=new_table.id    LIMIT 1)  WHERE id in ()
```

bwrsandman:
`alias_parent_*` fields need to be migrated in a way similar to https://github.com/OpenUpgrade/OpenUpgrade/commit/0ab9e6961b1581e7772d973b137a189b2e77fdc3

ajite:
Hi all,

Thank you for your comments. I will fix it as soon as I can.

@bwrsandman : For the missing function fields they are declared as stored char function in V7 and normal char field in V8. I believe that the migration should handle them automatically since in the end they are both declared as a char in the database. I will double check them.

hbrunn:
I'll close this PR as #116 does the same and is writeable by more users

PR #93, opened 2014-08-12
Stock v8 migration

diellenne:
Introducing migration scripts for 'procurement' and 'stock'


diellenne:
Incorrect pull request. Please ignore it.

PR #95, opened 2014-08-16
product: Fix invalid migration for product_supplierinfo

plaes:
This one was annoying. The column was simply renamed.

### DIFF
--- a/addons/product/migrations/8.0.1.1/openupgrade_analysis_WORK.txt
+++ b/addons/product/migrations/8.0.1.1/openupgrade_analysis_WORK.txt
@@ -42,7 +42,7 @@ product      / product.product          / attribute_value_ids (many2many): NEW r
 
 product      / product.supplierinfo     / product_id (many2one)         : DEL relation: product.template, required: required
 product      / product.supplierinfo     / product_tmpl_id (many2one)    : NEW relation: product.template, required: required
-# OK seller_ids moved from product.product to product.template
+# OK renamed
 
 product      / product.template         / active (boolean)              : NEW 
 # OK: set to True if at least one variant is active
--- a/addons/product/migrations/8.0.1.1/post-migration.py
+++ b/addons/product/migrations/8.0.1.1/post-migration.py
@@ -30,18 +30,6 @@ def load_data(cr):
                           mode='init')
 
 
-def move_fields(cr, pool):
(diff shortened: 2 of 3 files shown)


plaes:
Also, while I'm here - during migration there is one extra record created in the products with the data taken from product_data.xml (see below). How can this be disabled?

```
<record id="product_product_consultant" model="product.product">
    <field name="list_price">75.0</field>
    <field name="standard_price">30.0</field>
    <field name="uom_id" ref="product.product_uom_hour"/>
    <field name="uom_po_id" ref="product.product_uom_hour"/>
    <field name="name">Service</field>
    <field name="categ_id" ref="product.product_category_all"/>
    <field name="type">service</field>
    <field name="sale_ok" eval="False"/>
</record>
```

legalsylvain:
Approve. Thanks for the fix. :+1:

legalsylvain:
@all: Can some one else review that PR ?

@plaes: I don't understand your problem about the new product created. If a new data is created, it means that it is needed by OpenERP. If I understand correctly, your PoV is that is an new annoying demo data ? That'is not the aim of OpenUpgrade to disable that data.

If you want to disable this data, you can do it by UI after Upgrade, or by a module, 

```
<record id="product.product_product_consultant" model="product.product">
    <field name="active">False</field>
</record>
```

pedrobaeza:
The fix is not correct, because you don't have the equivalency product_id = template_id. This is only true when you don't have variants. Why the old code doesn't work?

legalsylvain:
I don't think so @pedrobaeza. It's just a renamed field.
See the analysis file :

```
product / product.supplierinfo / product_id (many2one) : DEL relation: product.template, required
product / product.supplierinfo / product_tmpl_id (many2one) : NEW relation: product.template, required
```

Both field are **product.template**.
the old field was bad named. It wasn't a product_id. New field is just good named.

Regards.

plaes:
Yeah, old column was referencing `product_template.id` as does the new one. Also, breakage was detected by postgresql too as NULL constraints were not fulfilled.

pedrobaeza:
Sorry, I read bad the analysis file. :+1: 

@plaes, about deleting record, you can do it loading an XML with this:

```
<delete model="ir.actions.report.xml" search="[('id', '=', ref('module.xml_id')]"/>
```

```
```

PR #96, opened 2014-08-20
Rename models in ir_model_data

bwrsandman:
When renaming models, we also want to rename them in ir_model_data.

This prevents errors in rename_xmlids such as:

> External ID conflict, document_page_procedure already refers to a `wiki.wiki` record, you can't define a `document.page` record with this ID.

Please also see 7.0 patch here: https://code.launchpad.net/~savoirfairelinux/openupgrade-server/rename_ir_model_data/+merge/231599

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -184,6 +184,8 @@ def rename_models(cr, model_spec):
                    'WHERE model = %s', (new, old,))
         cr.execute('UPDATE ir_model_fields SET relation = %s '
                    'WHERE relation = %s', (new, old,))
+        cr.execute('UPDATE ir_model_data SET model = %s '
+                   'WHERE model = %s', (new, old,))
     # TODO: signal where the model occurs in references to ir_model
 
 


rvalyi:
Hello @bwrsandman 
That looks good but I wonder what happen if we want to just move a model into another module. Do we have something for that? Because that would also mean touching the ir_model_data, but changing the name of it rather than model. Should we try to consider this case in this method? have it entirely elsewhere?

bwrsandman:
@rvalyi It shouldn't be affected by this patch, this is models, not modules.
The migration script which calls this must assert that both modules are installed in your case

rvalyi:
Hum OK I understand that this code works indeed.
I think we have nothing currently to move a model from a module to another, I was just wondering if that should be dealt with inside this method (with more params and logic). But anyway, let's leave that for later.

so :+1: for that patch already.

pedrobaeza:
I think that must be another function, not as an expansion of this one.

For this PR, :+1: 

Regards.

legalsylvain:
Thanks!
:+1:

bwrsandman:
Thank you for merging! Could I bother you reviewers to approve the MP on launchpad for v7 too? Link in the original description.

PR #101, opened 2014-09-22
select remote modules for install

ferdiga:
mark all uninstalled modules for installation - based on the modules installed on a remote db

how can I remove the openupgrade_analysis.txt from the pull request ?


StefanRijnhart:
Thanks! I can see this being useful. Just remove the extra lines in the analysis file and commit that, it will show up with no change here, I suppose.

Can you use map module names using renamed_modules in openupgrade_records/lib/apriory.py, and remove your initials in l.124?

ferdiga:
fixed

StefanRijnhart:
LGTM :+1:

legalsylvain:
LGTM :+1:

PR #103, opened 2014-09-23
Merge of Odoo 8.0 upstream of 20140923

StefanRijnhart:
Merged Odoo 8.0 and resolved trivial conflicts in
     - README.md
     - addons/l10n_be/**openerp**.py
     - modules/loading.py

Resolved conflicts in orm.py by porting the changes in http://bazaar.launchpad.net/~openupgrade-committers/openupgrade-server/8.0/revision/5144/openerp/osv/orm.py to openerp/models.py, as a lot of code in orm.py was refactored out to this file.


legalsylvain:
Hi @StefanRijnhart and thanks to work on sync OpenUpgrade and last version of Odoo.
I don't know exactly how to review that PR. 
When I click on your last commit (https://github.com/StefanRijnhart/OpenUpgrade/commit/8ba452fc19382eb292bf1761e084ba2145e01b03), I see all the changes and not only your work on conflict.

What is the best way to review that kind of PR ?

(You mention conflicted files in your comment but is there other way ?)
Regards.

pedrobaeza:
Indeed it's a very difficult PR to review, but we can concentrate on last commit, where Stefan resolved conflicts. It's also a big diff, but I see no problems in the way he has solved conflicts, so :+1: 

@StefanRijnhart, are we going to keep in sync OpenUpgrade the same way as OCB?

StefanRijnhart:
@legalsylvain, I know this is not reviewable. You might just trust me on this one (I am pretty confident about the changes) and if a couple of people are aware of this commit, then that should be enough to point in the right direction should problems arise after all. If you really want to verify what I've done, you can reproduce the merge and use the diff tool between the result of that and a separate checkout of this branch to get a clear view of the conflicts and how I resolved them. But honestly, I would not bother in this case.
@pedrobaeza thanks for the review! Keeping up with upstream has no benefit to OpenUpgrade as it is only concerned about changes in the data model, and Odoo 8.0 should be stable now. So I don't think that would be necessary.

pedrobaeza:
@StefanRijnhart, you are right about DB layout, but in the past we have found that some code has bugs, and it affects when you make any operation on OpenUpgrade, and more due to some records are modified via ORM instead of SQL. That's why I ask.

StefanRijnhart:
@pedrobaeza: That could happen (can't think of any from the past right now), but if they do now (what with the new API and all) I would like them to be analysed properly and communicated clearly throughout the OpenUpgrade community so that previously migrated databases can be fixed. I would not like it if the results of the upgrade process would differ from day to day. It would make support on the public forums more difficult as well.
I'd be curious to know what other people think about this. Obviously, if more people like to track Odoo actively, it would be easy to set it up using @hbrunn's sync script.

rvalyi:
Well, again, I don't know OpenUpgrade a lot, but I think we could be smarter on this by adopting a different file layout where we would be able to mount OpenUpgrade in the path and it would be able to pull the migrations files from an alternative location.

That way OpenUpgrade would be independent from the Odoo git tree (though versions will need to be sync'ed) and much easier to upgrade.

It's also likely that in the future, as OpenERP SA will transition to a SaaS model, monthly upgrades will get more importance, and having such structure might give a chance of not lagging so much. Eventually some non critical projects may follow the monthly changes and a year later OpenUpgrade would have the job already largely done instead of tackling all changes of the year at once and all Odoo bugs because nobody serious tested the code during 1 year. Just thinking loud...

StefanRijnhart:
@rvalyi no, because OpenUpgrade modifies the core at quite a few points in the database registration and the module loading methods.

rvalyi:
@StefanRijnhart 

1) couldn't these changes be a patch then? May be it would be easier to maintain such small patch that maintain merging with upstream, no?

2) We could also think OCB could introduce the required hooks (just the hooks, not the features) so that  OpenUpgrade would plug its extensions in a non invasive way. That would help a little bit as OCB is already maintained against the current stable branch. But as for the other branches than stable that wouldn't help much as we would have roughly the same sync' issue.

3) a Third idea would be these hooks (again not the features, just the hooks) would be a specific commit in a branch of the stable branch that would be merged, in the stable, branch or in OCB, or eventually in a newer branch.

What do you think about these two options?

StefanRijnhart:
@rvalyi I don't see how separating the code bases would help maintaining the project except for reviews such as these. In the integrated approach, the integrity of the scripts vs. the upstream code base is guaranteed which is important especially with shorter upgrade cycles. Separating the migration scripts from the code base in order to perform a migration over multiple iterations might not work as the upstream API might change in a non backwards compatible manner. Migrating over multiple iterations is however supported by the migrate.py script, so I don't see the functional gap here.

rvalyi:
@StefanRijnhart not sure it will change your mind, but when you say "separating the code bases", as for the main OpenUpgrade project, I propose to keep the OpenUpgrade methods and script together in the same repo.

In fact, there is only two things I propose to move out:

1) a patch introducing the required hooks so that OpenUpgrade can plug into the server core. that could be in the form of 1), 2) or 3). I should say I have no idea how large is that patch, so that's just an idea so far. Hopefully this patch would be minimalist.

2) not carry odoo inside OpenUpgrade. That is yes, splitting scripts from implementations. But may be it could be done in a way that: a given branch/revision of OpenUpgrade always explicitly targets a given revision of Odoo.
And may be that could be even made explicit with some sha1 ref in OpenUpgrade or pointing to Odoo with a submodule revision.
A given revision of OpenUprade would also make explicit it's using a specific patch version of the core.

I'm not asking for change, just thinking, of course I trust you to have a much better vision of what is better.

What I imagine though is by doing this the OpenUpgrade codebase would be 2 orders of magnitude smaller or so and suddenly that would become possible to compare OpenUpgrade for 8.0 stable with say OpenUpgrade for SaaS/8.0+n or SaaS/9.0-n that typically somebody will have played with and get these efforts integrated.

StefanRijnhart:
@rvalyi let's come back to this if and when work starts on the subrelease iteration cycle, and see what is needed then.

vaab:
Hi, may I share some experiments with you ?

Being a GIT afficionado for a few year, and using openupgrade code as basis for a prototype, we use to maintain commits upon odoo's code. This is like maintaining patches, except GIT makes it really easy and natural: we avoid merge and use only rebasing. This helps having easy and clear reviews, and as openupgrade keeps a low footprint on odoo's code, there is usually really few effort spent to maintain the commit(s). (We actually have 120+ patches like this)

We didn't have much time yet to explore, but we'd like to integrate completely openupgrade to odoo's code (this means, being able to use it in production code), and also to have it run the full test framework. We do have already a prototype that works well on these both side. We actually also use transaction for loading/install/migration that remove intermediate states of databases. But we didn't look too much yet into really leveraging the migration code.

We could keep a directory for each module version in "migration" folder, including long past version from last major openerp version (5, 6, 6.1, 7.0, 8.0). Any given version of our planned prototype should be able to incrementally migrate any previous database content to the current model structure gracefully. So we do not make special cases for major versions and are ready for sub-release iteration cycle. 

I don't know if these information are of interests for your discussion. If things are unclear or you think we have missed important points, please voice your concerns, as we are still building/maintaining this prototype and are aware that we probably don't grasp as much as you do in your area of expertise.

hbrunn:
@vaab Even though I think your proposal sounds charming, I don't see how it saves effort like this one when we have conflicts between odoo changes and OpenUpgrade? Also with rebasing, we'll have to resolve those conflicts manually and this will be hidden in the rewritten commits, right?

As for multiple migrations: If you want to run multiple migrations at once, you'd have to change the way OpenUpgrade runs the migrations. Currently, it does the same as odoo: Upgrade a module for which you find migrations scripts fully. That would mean that if you drop migration scripts for 6.1, 7.0 and 8.0 into the same migration folder and run a version 8.0 server, you get the base module upgraded to 8.0 first and then have migrations 6.1, 7.0 and 8.0 for the other modules. This is going to go wrong, as every migration relies on base (or other dependencies) being at the version we migrate to. So we'd have to introduce even more changes to the migration logic and probably would end up with some code that has nothing to do with odoo any more but is just a framework to run migration scripts. I doubt that simplifies maintenance.

Are you aware that the 6.1 and 7.0 version of this exist on launchpad and do you know the script that binds everything together to also make a multi version upgrade a oneliner? https://doc.therp.nl/openupgrade/migrate.py.html

hbrunn:
As for this PR: I reconstructed the changes and arrive at the same code, so I'm going to merge this.

openerp/models.py's diff attached for reference:

```
diff --git a/openerp/models.py b/openerp/models.py
index 4f9fc28..d9a90bf 100644
--- a/openerp/models.py
+++ b/openerp/models.py
@@ -1034,6 +1022,8 @@ class BaseModel(object):

         position = 0
         try:
+            # use savepoints for openupgrade instead of transactions
+            cr.execute('SAVEPOINT convert_records')
             for res_id, xml_id, res, info in self._convert_records(cr, uid,
                             self._extract_records(cr, uid, fields, datas,
                                                   context=context, log=log),
@@ -1051,8 +1041,9 @@ class BaseModel(object):
                     if context.get('defer_parent_store_computation'):
                         self._parent_store_compute(cr)
                     cr.commit()
+            cr.execute('RELEASE SAVEPOINT convert_records')
         except Exception, e:
-            cr.rollback()
+            cr.execute('ROLLBACK TO SAVEPOINT convert_records')
             return -1, {}, 'Line %d : %s' % (position + 1, tools.ustr(e)), ''

         if context.get('defer_parent_store_computation'):
@@ -2626,11 +2617,14 @@ class BaseModel(object):
                                     self._set_default_value_on_column(cr, k, context=context)
                                 # add the NOT NULL constraint
                                 try:
+                                    # use savepoints for openupgrade instead of transactions
+                                    cr.execute('SAVEPOINT add_constraint');
                                     cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" SET NOT NULL' % (self._table, k), log_exceptions=False)
-                                    cr.commit()
+                                    cr.execute('RELEASE SAVEPOINT add_constraint');
                                     _schema.debug("Table '%s': column '%s': added NOT NULL constraint",
                                         self._table, k)
                                 except Exception:
+                                    cr.execute('ROLLBACK TO SAVEPOINT add_constraint');
                                     msg = "Table '%s': unable to set a NOT NULL constraint on column '%s' !\n"\
                                         "If you want to have it, you should update the records and execute manually:\n"\
                                         "ALTER TABLE %s ALTER COLUMN %s SET NOT NULL"
@@ -2703,11 +2697,14 @@ class BaseModel(object):
                                 cr.execute('CREATE INDEX "%s_%s_index" ON "%s" ("%s")' % (self._table, k, self._table, k))
                             if f.required:
                                 try:
-                                    cr.commit()
+                                    #use savepoints for openupgrade instead of transactions
+                                    cr.execute('SAVEPOINT add_constraint');
                                     cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" SET NOT NULL' % (self._table, k))
                                     _schema.debug("Table '%s': column '%s': added a NOT NULL constraint",
                                         self._table, k)
+                                    cr.execute('RELEASE SAVEPOINT add_constraint');
                                 except Exception:
+                                    cr.execute('ROLLBACK TO SAVEPOINT add_constraint');
                                     msg = "WARNING: unable to set column %s of table %s not null !\n"\
                                         "Try to re-run: openerp-server --update=module\n"\
                                         "If it doesn't work, update records and execute manually:\n"\
@@ -2901,12 +2898,14 @@ class BaseModel(object):
             sql_actions.sort(key=lambda x: x['order'])
             for sql_action in [action for action in sql_actions if action['execute']]:
                 try:
+                    # use savepoints for openupgrade instead of transactions
+                    cr.execute('SAVEPOINT add_constraint2');
                     cr.execute(sql_action['query'])
-                    cr.commit()
+                    cr.execute('RELEASE SAVEPOINT add_constraint2');
                     _schema.debug(sql_action['msg_ok'])
                 except:
                     _schema.warning(sql_action['msg_err'])
-                    cr.rollback()
+                    cr.execute('ROLLBACK TO SAVEPOINT add_constraint2');


     def _execute_sql(self, cr):
```

vaab:
@hbrunn You asked: "Also with rebasing, we'll have to resolve those conflicts manually and this will be hidden in the rewritten commits, right?"

You are right. Rebasing means maintaining a pile of patch: this means that new patches superseeds older ones. You can of course keep older ones by keeping a reference to older patch stack (with a tag for instance).  If there are no conflict, commits are cleanly moved upon the new tree, no commits are added. 

I would be happy to know what info is loosed or not practical to access in your perspective. Merging to me, is making history a mess without any useful value.... But I don't want to start over the Merge vs Rebase argument all over, it has been done numerous time on the net, and I guess this is a matter of conviction. With bazaar and mercurial there wasn't really a choice, because the tools are too clumsy to be efficient at rebasing or branching. In git, rebasing and branching are not costly at all. As the discussion seemed to wander around the notion of having a distinct and clear way to keep track of the modification that openupgrade makes to odoo's code. I can't see better way than maintaining a pile of patches. But that's only my point of view.

As for migration issues, I will refrain of speaking of code that we are in process of writing. But I'll consider a success to avoid having to pass through your "oneliner" that hides the downloads of several openerp+openupgrade branches (all of them having different code and bugs) with dependencies that will come more difficult to gather, that must be run with a growing number of specific steps that seem arbitrarily chosen (what about saas-X ? what is the exact 8.0 for you ?). I'm having trouble to understand how sustainable this can be.

PR #109, opened 2014-10-13
[FIX] SQL query results in invalid reference

mariuszmizgier:
With post-migrate.py in mrp module there is a query, which uses aliased table without setting this alias. This commit will fix this.

### DIFF
--- a/addons/mrp/migrations/8.0.1.1/post-migrate.py
+++ b/addons/mrp/migrations/8.0.1.1/post-migrate.py
@@ -9,8 +9,8 @@ def move_fields(cr, pool):
         UPDATE product_product
         SET produce_delay = (
             SELECT pt.%s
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I prefer not to put abbreviations on table names, so my chance is:

```
SELECT product_template.%s
FROM product_template
WHERE product_template.id = product_product.product_tmpl_id)
```

but the patch is valid. Thanks for it.

Regards.

paulius-sladkevicius:
LGMT :+1:

hbrunn:
:+1:

pankk:
If I checked correctly, this part of the script is no longer necessary as the produce_delay field is moved back to product_template in the following commit: 82664ba614339bbe04e9ea653d38a5056ac11425

PR #111, opened 2014-10-21
[FIX] ignore view errors during upgrade

hbrunn:


pedrobaeza:
This was in previous versions of OpenUpgrade. Has this been deactivated in the switch to Odoo?

hbrunn:
Not on purpose, but it didn't make it through the automatic transition as this is new code (from the point of view of 7.0)

StefanRijnhart:
My bad for quick and dirty merge of Odoo 8.0 stable recently. Thanks! :+1:

legalsylvain:
Thanks. :+1:

PR #116, opened 2014-10-28
8.0 'mail' (onesteinbv)

legalsylvain:
Port 'mail' modification realized by OneStein BV. (branch onesteinbv:8.0-os)


hbrunn:
[retracted thumbs up]

hbrunn:
As noted in https://github.com/OpenUpgrade/OpenUpgrade/pull/64#discussion_r13404465, we need to migrate the alias parent fields

hbrunn:
I'll provide a general function for this that the other migration scripts can use

hbrunn:
This function already exists, I needed to rebase this on current 8.0 to make use of it

hbrunn:
:+1:

addons/mail/migrations/8.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+
+from openerp.openupgrade import openupgrade
+
+column_renames = {
+    'res_partner': [
+        ('notification_email_send', 'notify_email'),
+    ],
+    'mail_mail': [
+        ('email_from', None),
+
```
pedrobaeza: Why this empty line?

addons/mail/migrations/8.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.openupgrade import openupgrade
+
+column_renames = {
+    'res_partner': [
+        ('notification_email_send', 'notify_email'),
```
pedrobaeza: You should rename to None, and then use legacy name in map_values call.

addons/mail/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+UPDATE mail_message
+SET %s = %s, %s = %s
+FROM mail_message AS a JOIN mail_mail AS b ON a.id = b.mail_message_id
+""" % ('mail_server_id', legacy_server_id, 'reply_to', legacy_reply_to, ))
+
+
+def map_notify_email(cr):
+    openupgrade.map_values(
+        cr,
+        'notify_email',
```
pedrobaeza, line 51: As pointed by @christophlsa in #162, this function gives an error. But I have to say that the method is correct and you have to change this call: the check is intended to prevent to overwrite a column with unintended consequences, and also to be conservative, letting in legacy column the old values. Imagine this case: [('all', 'always'), ('always', 'none')]. The first mapping will set all values to always, and the second one will map all values (old 'all' values, and current 'always' values), so this in not what we want.

addons/mail/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
+
+    map_notify_email(cr)
+    mail_mail_to_mail_message_migration(cr, uid, pool)
+    openupgrade_80.set_message_last_post(cr, uid, pool, ['res.partner', 'mail.group'])
+    openupgrade.set_defaults(cr, pool, default_spec, force=False)
```
pedrobaeza: Isn't this done by standard migration machinery?

hbrunn: it's a new field, so yes

pedrobaeza: But this new field doesn't take the default as initial value? When I have created new fields with default value developing modules, system has created the columns with this initial value.

hbrunn: sorry, that's what I meant: You're right, this line (and the declaration of the default_spec dictionary above) is not necessary. For the record: The use case for openupgrade.set_defaults usually is an existing field that becomes required, so we want to set defaults for records that don't have a value set. Otherwise, the not null constraint can't be set up

pedrobaeza: OK, thanks for the clarification. We can remove this line then and commented properly on work file to not get confused people that uses this script as a pattern for creating another.

StefanRijnhart:
Thanks for the review, guys. I took this up and committed your suggestions. Looks fine by me now. :+1:

pedrobaeza:
:+1:  for me too

I merge.

PR #119, opened 2014-10-28
8.0 'portal' (onesteinbv)

legalsylvain:
Port 'portal' modification realized by OneStein BV. (branch onesteinbv:8.0-os)


addons/portal/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+
+def migrate_group_users(cr, uid, pool):
+    '''
+    Migrate groups portal and anonymous
+    :param cr: cursor
+    '''
+    data_obj = pool['ir.model.data']
+    old_group_portal = data_obj.get_object_reference(cr, uid, 'portal', 'group_portal')[1]
+    new_group_portal = data_obj.get_object_reference(cr, uid, 'base', 'group_portal')[1]
+    if old_group_portal and new_group_portal:
```
christophlsa: Instead of the if you should catch the ValueError exception which is thrown if this template doesn't exists or you get: "ValueError: External ID not found in the system: portal.group_portal"

addons/portal/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+        cr.execute("""update res_groups_users_rel set gid=%s where gid=%s""",
+                   (new_group_portal, old_group_portal,))
+        cr.execute("""DELETE FROM res_groups WHERE id = %s""", (old_group_portal,))
+
+    old_group_anonymous = data_obj.get_object_reference(cr, uid, 'portal', 'group_anonymous')[1]
+    new_group_public = data_obj.get_object_reference(cr, uid, 'base', 'group_public')[1]
+
+    if old_group_anonymous and new_group_public:
```
christophlsa: As some lines in top: Instead of the if you should catch the ValueError exception which is thrown if this template doesn't exists.

StefanRijnhart:
Revised and tested. Ready for review.

pedrobaeza:
Where is `portal_data.xml`?

StefanRijnhart:
It's included in the core module. It only sets 'is_portal' to True for the portal group now living in base.

pedrobaeza:
OK, thanks.

:+1:

legalsylvain:
:+1: Thanks !

pedrobaeza:
I merge. 

We are completing 8.0 migration scripts!!

PR #118, opened 2014-10-28
8.0 add 'map_values' function (onesteinbv)

legalsylvain:
Port 'core' modification of openupgrade realized by OneStein BV. (branch onesteinbv:8.0-os)

PR #122, opened 2014-10-28
8.0 'project_issue' fix priority field migration (onesteinbv)

legalsylvain:
Port 'project_issue' modification realized by OneStein BV. (branch onesteinbv:8.0-os)

### DIFF
--- a/addons/project_issue/migrations/8.0.1.0/post-migration.py
+++ b/addons/project_issue/migrations/8.0.1.0/post-migration.py
@@ -3,7 +3,7 @@
 #
 # Odoo, a suite of business apps
 # This module Copyright (C) 2014 Therp BV (<http://therp.nl>).
-#    @author: Onestein <www.onestein.nl>
+#    @contributor: Onestein <www.onestein.nl>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
--- a/addons/project_issue/migrations/8.0.1.0/post-migration.py
+++ b/addons/project_issue/migrations/8.0.1.0/post-migration.py
@@ -3,6 +3,7 @@
 #
 # Odoo, a suite of business apps
 # This module Copyright (C) 2014 Therp BV (<http://therp.nl>).
+#    @author: Onestein <www.onestein.nl>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


addons/project_issue/migrations/8.0.1.0/post-migration.py
```diff
@@ -3,6 +3,7 @@
 #
 # Odoo, a suite of business apps
 # This module Copyright (C) 2014 Therp BV (<http://therp.nl>).
+#    @author: Onestein <www.onestein.nl>
```
pedrobaeza: Onestein is not the author. Only a contributor. Original author is Therp.

addons/project_issue/migrations/8.0.1.0/post-migration.py
```diff
@@ -19,21 +20,22 @@
...
+@openupgrade.migrate()
 def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
 
     openupgrade.map_values(
         cr,
         openupgrade.get_legacy_name('priority'),
         'priority',
-        [('5', '0'), ('4', '0'), ('3', '1'), ('2', '2'), ('1', '2')],
+        [('4', '0'), ('3', '0'), ('2', '1'), ('1', '2'), ('0', '2')],
```
pedrobaeza, line 35: These are old values:
`('1', 'Highest'), ('2', 'High'), ('3', 'Normal'), ('4', 'Low'), ('5', 'Lowest')`
and these are the new ones:
`('0', 'Low'), ('1', 'Normal'), ('2', 'High')`
so I don't think we should change this. Previous mapping is correct.

pedrobaeza:
I put that this depends on #118 for the record.

addons/project_issue/migrations/8.0.1.0/post-migration.py
legalsylvain: Fixed.

addons/project_issue/migrations/8.0.1.0/pre-migration.py
```diff
@@ -3,6 +3,7 @@
 #
 # Odoo, a suite of business apps
 # This module Copyright (C) 2014 Therp BV (<http://therp.nl>).
+#    @author: Onestein <www.onestein.nl>
```
pedrobaeza, line 6: This must be also contributor.

addons/project_issue/migrations/8.0.1.0/post-migration.py
```diff
@@ -19,21 +20,22 @@
 #
 ##############################################################################
 
-from openerp import SUPERUSER_ID as uid
+from openerp import pooler, SUPERUSER_ID as uid
```
hbrunn, line 23: better do `from openerp.modules.registry import RegistryManager`...

addons/project_issue/migrations/8.0.1.0/post-migration.py
```diff
@@ -19,21 +20,22 @@
...
 from openerp.openupgrade import openupgrade, openupgrade_80
 
 
-@openupgrade.migrate
+@openupgrade.migrate()
 def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
```
hbrunn, line 29: and use `pool = RegistryManager.get(cr.dbname)` here in order to avoid deprecation warnings

red15:
:+1: 
Looks like I just did some duplicate work in #128 arriving at the same solutions.
We should get this merged quickly to prevent more duplicate work.

pedrobaeza:
I close this one as #128 has been merged.

PR #123, opened 2014-10-28
8.0 'purchase' improve (onesteinbv)

legalsylvain:
Port 'purchase' modification realized by OneStein BV. (branch onesteinbv:8.0-os)


StefanRijnhart:
LGTM, thanks! :+1:

addons/purchase/migrations/8.0.1.1/openupgrade_analysis.txt
```diff
@@ -1,28 +1,50 @@
 ---Fields in module 'purchase'---
```
pedrobaeza, line 1: Please make these changes on a file `openupgrade_analysis_work.txt` or similar, because this will be overwritten in next analysis update.

addons/purchase/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+        product_ids = []
+        cr.execute("SELECT id FROM product_template WHERE %s = %%s" % procure_method_legacy, ('buy',))
+        for res in cr.fetchall():
+            product_ids.append(res[0])
+
+        template_obj.write(cr, uid, product_ids, {'route_ids': [(4, mto_route_id)]})
+
+
+def migrate_procurement_order(cr):
+    '''
```
pedrobaeza: Please use double quotes for docstring (http://legacy.python.org/dev/peps/pep-0257/#id16)

addons/purchase/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+    res = cr.fetchall()
+
+    for po_id, purchase_id in res:
+        procurement = procurement_obj.browse(cr, uid, po_id)
+        cr.execute("""SELECT id FROM purchase_order_line WHERE order_id = %s AND product_qty = %s AND
+                    product_id = %s""", (purchase_id, procurement.product_qty, procurement.product_id.id))
+        res2 = cr.fetchone()
+        pol_id = res2 and res2[0] or False
+        if pol_id:
+            procurement_obj.write(cr, uid, po_id, {'purchase_line_id': pol_id})
```
pedrobaeza, line 126: We should made this by SQL to avoid possible triggers on write.

addons/purchase/migrations/8.0.1.1/openupgrade_analysis.txt
```diff
@@ -1,28 +1,50 @@
...
+
+purchase     / purchase.order           / bid_date (date)               : NEW
+purchase     / purchase.order           / bid_validity (date)           : NEW
 purchase     / purchase.order           / currency_id (many2one)        : not a function anymore
 purchase     / purchase.order           / date_order (date)             : type is now 'datetime' ('date')
 purchase     / purchase.order           / incoterm_id (many2one)        : NEW relation: stock.incoterms
-purchase     / purchase.order           / message_last_post (datetime)  : NEW 
+# NOTHING TO DO
+
+purchase     / purchase.order           / currency_id (many2one)        : not a function anymore
```
pedrobaeza: This was a related field not stored, so you must fill it with a proper value (you can get it from associated pricelist) because the field is required.

addons/purchase/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+    '''
+    Procurements of products: change the supply_method for the matching route
+    Buy -> Buy Rule
+    :param cr: Database cursor
+    '''
+    pool = pooler.get_pool(cr.dbname)
+    route_obj = pool['stock.location.route']
+    template_obj = pool['product.template']
+
+    mto_route_id = route_obj.search(cr, uid, [('name', 'like', 'Buy')])
```
pedrobaeza: Why don't you use xml-id ref? It's one more search, but you avoid problems if the rule name changes.

addons/purchase/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+    mto_route_id = mto_route_id and mto_route_id[0] or False
+
+    procure_method_legacy = openupgrade.get_legacy_name('supply_method')
+    if mto_route_id:
+        product_ids = []
+        cr.execute("SELECT id FROM product_template WHERE %s = %%s" % procure_method_legacy, ('buy',))
+        for res in cr.fetchall():
```
pedrobaeza: You can make it with a list comprehension:
`product_ids = [res[0] for res in cr.fetchall()]`

pedrobaeza:
Apart from the comments inline, one last thing: please update documentation properly to reflect that this migration script is done.

addons/purchase/migrations/8.0.1.1/post-migration.py
pankk: So we would use the id route_warehouse0_buy in this case, right? That would be the right way to do this as you say.

pedrobaeza: Yeah, that's it.

addons/purchase/migrations/8.0.1.1/openupgrade_analysis.txt
legalsylvain, line 1: Done.

legalsylvain:
all suggestion done, except:
- migration of purchase.order.currency_id;
- do a SQL request to write purchase_line_id;

pedrobaeza:
Do you want me to do it in a PR against your branch?

legalsylvain:
With pleasure !

pedrobaeza:
OK, I'll try to do it tomorrow.

legalsylvain:
Ok. i gave you write access on my project, to avoid loosing time, forking repository and branches.

addons/purchase/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,134 @@
...
+    res = cr.fetchone()
+    wkf_activity = res and res[0]
+
+    # Get the not shipped moves linked to any Purchase Order
+    cr.execute(
+        """SELECT purchase_line_id
+        FROM stock_move
+        WHERE purchase_line_id is not null
+        AND state != %s""",
+        ('done'))
```
christophlsa: For one-element-tuples you have to add a comma: ('done',))

Without this comma you get: ValueError: SQL query parameters should be a tuple, list or dict; got 'done'

addons/purchase/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,134 @@
...
+    template_obj = pool['product.template']
+
+    mto_route_id = pool['ir.model.data'].get_object_reference(
+        cr, uid, 'purchase', 'route_warehouse0_buy')[1]
+
+    procure_method_legacy = openupgrade.get_legacy_name('supply_method')
+    if mto_route_id:
+        product_ids = []
+        cr.execute("""SELECT id FROM product_template WHERE %s = %%s""" % (
+            procure_method_legacy, ('buy',)))
```
christophlsa: You have to move one of the closing bracket to the left: procure_method_legacy), ('buy',))

StefanRijnhart:
I made some improvements, and checked the changes to noupdate-xml: they only involve the order confirmation email template, so we can leave it. As far as I am concerned we can merge this and pick up any remaining issues afterwards. So :+1:

legalsylvain:
Hi ! 
If there are some remaining issues, maybe we could mark this module as 'work in progress', instead of 'done' : https://github.com/OpenUpgrade/OpenUpgrade/pull/123/files#diff-24e64b4df10ed12ace5e7e0346436923R383
What do you think ?

StefanRijnhart:
Oh, what are these issues?

legalsylvain:
I don't know : you said "As far as I am concerned we can merge this and pick up any remaining issues afterwards". Sorry If I misunderstood your remark.

pedrobaeza:
OK for me too. :+1:

pedrobaeza:
I merge then.

StefanRijnhart:
Oh, sorry for the misunderstanding. I don't see any issues yet, so I think we can leave it at 'done'.

PR #121, opened 2014-10-28
8.0 'procurement (dev)

legalsylvain:
Port 'procurement' modification based on OneStein BV. (branch onesteinbv:8.0-os)

PR #125, opened 2014-10-28
8.0 'stock' (dev)

legalsylvain:
Port 'stock' modification realized by OneStein BV. (branch onesteinbv:8.0-os)


red15:
Getting error with `openupgrade_legacy_8_0_procurement_id` not existing in model `stock.warehouse.orderpoint`.

I think you might need to add the 'procurement_id' column rename:

```
--- a/addons/stock/migrations/8.0.1.1/pre-migration.py
+++ b/addons/stock/migrations/8.0.1.1/pre-migration.py
@@ -64,6 +64,7 @@ column_renames = {
     ],
     'stock_warehouse_orderpoint': [
         ('product_uom', None),
+        ('procurement_id', None),
     ],
 }
```

red15:
Also missing `product.template / procure_method`

hbrunn:
@red15 this depends on #121, there both renames are done

red15:
@hbrunn well spotted, I was trying this branch alone to avoid mixing responsibilities. It's a good thing this huge work they have done is split up. Unfortunately this means you can get missing things like this.

legalsylvain:
@red15 :
Procurement module depends on 'stock' module. So there is no problem. When the 'stock' migration will be reviewed. (and it has to be reviewed first), so the migration of the procurement module will work.

Regards.

legalsylvain:
Oh wait, I spoke too soon.

In this case, we maybe have to move part of the 'procurement' (#121) migration code into this PR.

A migration of stock without procurement module installed has to work.

hbrunn:
@legalsylvain it's the other way around, right? stock depends on procurement, so in my opinion, that's fine

legalsylvain:
It depends ! We are both right. I checked V7 and you checked V8.
- V7: 'procurement' depends on 'stock':
  https://github.com/OCA/OCB/blob/7.0/addons/procurement/__openerp__.py#L29
- V8: 'stock' depends on 'procurement':
  https://github.com/OCA/OCB/blob/8.0/addons/stock/__openerp__.py#L55

-> basically, if procurement is not installed on V7, We have to install it ... and maybe before the migration. Could we imagine that OpenUpgrade block migration if procurement is not installed (and stock is). It will be the most simple. I'm not sure...

@StefanRijnhart, @hbrunn, @all, what do you think about that point ?

Regards.

StefanRijnhart:
Well spotted! Procurement should be installed automatically on databases with stock installed in 7.0 without any work from our side. But the stock migration script should then not assume the availabilty of any legacy tables or columns from the procurement module (haven't checked if it does).

hbrunn:
it does, that's the problem @legalsylvain raised earlier. Here my understanding:

| stock | procurement | problem? |
| --- | --- | --- |
| yes | yes | no |
| yes | no | yes |
| no | yes | no (shouldn't happen) |
| no | no | no (trivial) |

So to handle the second case, all references to legacy procurement columns should first check if they exist in the fist place.

legalsylvain:
I'm lost.
Why " But the stock migration script should then not assume the availabilty of any legacy tables or columns from the procurement module (haven't checked if it does).", if we block migration of database with 'stock' installed and 'procurement' uninstalled.
In that case, we can assume that second case will not occure.
And so, use procurement column in stock migration...

(I'm not sure that it is the good solution, but for me the most simple prima facie.)

StefanRijnhart:
The stock migration scripts should simply check if the procurement legacy table and/or columns that it wants to use exist, and apply a strategy for when they don't.

legalsylvain:
Agree, it's the most modular solution.

hbrunn:
my latest commit should do the trick

christophlsa:
I have some errors according to changing colums which have null values. Am I right that the workflow of migration is the following?:
1. pre-migration
2. normal odoo module update
3. post-migration

When this is right then I think some methods from post-migration have to be moved to pre-migration. Or have I only rerun the update script and it's ok (is this a good workflow?)? I looked in the tables in database and these null values are filled after post-migration (too late for the module update).

Here are my errors:

```
2014-12-12 09:32:31,017 23335 ERROR testdb openerp.sql_db: bad query: ALTER TABLE "stock_picking" ALTER COLUMN "picking_type_id" SET NOT NULL
IntegrityError: column "picking_type_id" contains null values

2014-12-12 09:32:31,018 23335 WARNING testdb openerp.models: WARNING: unable to set column picking_type_id of table stock_picking not null !
Try to re-run: openerp-server --update=module
If it doesn't work, update records and execute manually:
ALTER TABLE stock_picking ALTER COLUMN picking_type_id SET NOT NULL
  File "/var/tmp/openupgrade/8.0/server/openerp/models.py", line 2702, in _auto_init
    cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" SET NOT NULL' % (self._table, k))
  File "/var/tmp/openupgrade/8.0/server/openerp/sql_db.py", line 158, in wrapper
IntegrityError: column "picking_type_id" contains null values

2014-12-12 09:32:36,946 23335 ERROR testdb openerp.sql_db: bad query: ALTER TABLE "stock_warehouse" ALTER COLUMN "code" SET NOT NULL
IntegrityError: column "code" contains null values

2014-12-12 09:32:36,947 23335 WARNING testdb openerp.models: WARNING: unable to set column code of table stock_warehouse not null !
Try to re-run: openerp-server --update=module
If it doesn't work, update records and execute manually:
ALTER TABLE stock_warehouse ALTER COLUMN code SET NOT NULL
  File "/var/tmp/openupgrade/8.0/server/openerp/models.py", line 2702, in _auto_init
    cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" SET NOT NULL' % (self._table, k))
  File "/var/tmp/openupgrade/8.0/server/openerp/sql_db.py", line 158, in wrapper
IntegrityError: column "code" contains null values

2014-12-12 09:32:36,991 23335 ERROR testdb openerp.sql_db: bad query: ALTER TABLE "stock_warehouse" ALTER COLUMN "view_location_id" SET NOT NULL
IntegrityError: column "view_location_id" contains null values

2014-12-12 09:32:36,991 23335 WARNING testdb openerp.models: WARNING: unable to set column view_location_id of table stock_warehouse not null !
Try to re-run: openerp-server --update=module
If it doesn't work, update records and execute manually:
ALTER TABLE stock_warehouse ALTER COLUMN view_location_id SET NOT NULL
  File "/var/tmp/openupgrade/8.0/server/openerp/models.py", line 2702, in _auto_init
    cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" SET NOT NULL' % (self._table, k))
  File "/var/tmp/openupgrade/8.0/server/openerp/sql_db.py", line 158, in wrapper
IntegrityError: column "view_location_id" contains null values
```

hbrunn:
The workflow is as you describe, so the scripts should take care of fixing those values. If this is possible at all, there are some concepts that are simply new and there's no way to have a sensible default value.

You should btw always run the migration on a fresh database. Many of the migration scripts will break when being run multiple times

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,598 @@
...
+    """Reprocess stock moves in state done to fill stock.quant."""
+    stock_move_obj = pool['stock.move']
+
+    done_move_ids = stock_move_obj.search(cr, uid, [('state', '=', 'done')])
+    stock_move_obj.write(cr, uid, done_move_ids, {'state': 'draft'})
+    # Process moves using action_done.
+    stock_move_obj.action_done(cr, uid, done_move_ids, context=None)
+
+
+def delete_stock_account_journal(cr):
```
hbrunn: This should be an xmlid rename instead of deleting the journal. We'll soon provide a fix for that

StefanRijnhart:
I've pushed some improvements and set back to work in progress to indicate that it needs more work IMHO.

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,526 @@
...
+    # Fill picking_type_id required field
+    for type, type_id in (('in', in_id), ('out', out_id), ('internal', int_id)):
+        cr.execute("""UPDATE stock_picking SET picking_type_id = %%s WHERE %s = %%s""" % type_legacy, (type_id, type,))
+
+    # state key auto -> waiting
+    cr.execute("""UPDATE stock_picking SET state = %s WHERE state = %s""", ('waiting', 'auto',))
+
+
+def _migrate_stock_warehouse(cr, id):
+    """Warehouse adaptation to the new functionality. Sequences, Picking types, Rules, Paths..
```
StefanRijnhart: Note to self: check if this wad of copied code can simply be triggered by calling one or more methods of the warehouse model. Or better, reuse existing sequences and optionally keep sharing them between warehouses if that was the case in 7.0. NB sequence and sequence type codes 'stock.picking' (for internal pickings) were renamed in OCB to 'stock.picking.internal'. Match both when retrieving existing sequences.

legalsylvain:
Hi, I'd like to work on 'stock' migration. @StefanRijnhart, you put a WIP label, but you commited a lot after. 
Is this PR ready to review ? Is there some work to do ?
Thanks.

StefanRijnhart:
So the status AFAIK is: seems complete for some scenarios, but may be dubious for other, untested scenarios. Think multiwarehouse, multicompany, nontrivial chained picking steps. Other than that, I have no concrete todos in mind except that due to the complexity of this migration, a full revalidation of the work would be useful.

legalsylvain:
All right, thanks a lot.
I'll work on that in the next weeks, depending of a pending lead.
Regards.

legalsylvain:
All right, thanks a lot.
I'll work on that in the next weeks, depending of a pending lead.
Regards.

thomaspaulb:
I got the following error while running this script:

```
File "/var/tmp/openupgrade/8.0/addons/stock/migrations/8.0.1.1/post-migration.py", line 561, in migrate_product_supply_method
procure_method_legacy, ('make_to_order',)))
TypeError: not all arguments converted during string formatting
```

To solve it I needed to make the following change in post_migration.py:

```
cr.execute("SELECT id FROM product_template WHERE %s = %%s" % (
procure_method_legacy, ('make_to_order',)))
```

To

```
cr.execute("SELECT id FROM product_template WHERE %s = '%s'" % (
procure_method_legacy, 'make_to_order')
```

hbrunn:
please provide a PR against this branch

thomaspaulb:
https://github.com/OpenUpgrade-dev/OpenUpgrade/pull/2

addons/stock/migrations/8.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,164 @@
...
+        'stock_account.property_stock_account_input_categ'),
+    ('stock.property_stock_account_input_prd',
+        'stock_account.property_stock_account_input_prd'),
+    ('stock.property_stock_account_output_categ',
+        'stock_account.property_stock_account_output_categ'),
+    ('stock.property_stock_account_output_prd',
+        'stock_account.property_stock_account_output_prd'),
+    ('stock.property_stock_journal', 'stock_account.property_stock_journal'),
+    ('stock.stock_journal', 'stock_account.stock_journal'),
+]
```
njeudy, line 89: Think xmlid moved to stock_account module should be rename in pre-migration script of stock_account and not here.
stock_account is not a depends of stock ..

legalsylvain, line 89: I'm not sure.
Upgrade of 'stock' will delete xml_ids that disappeared. (except noupdate xml_id)
When process will update 'stock_account', the object has been deleted.

Regards.

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,714 @@
...
+    cr.execute(
+        """
+        UPDATE procurement_order AS po
+        SET partner_dest_id = sm.partner_id
+        FROM stock_move AS sm
+        WHERE po.move_dest_id = sm.id
+        """)
+    company_obj = registry['res.company']
+    warehouse_obj = registry['stock.warehouse']
+    procurement_obj = registry['stock.warehouse']
```
legalsylvain: :question: I guess it should be procurement_obj = registry['procurement.order'] ?
My current log (#l685), indicate :
stock.warehouse.write() with unknown fields: warehouse_id

StefanRijnhart: Yes, I'm afraid so ;-)

pedrobaeza:
I have face one error with this:

```
2015-03-23 03:25:04,195 29802 ERROR nayar_migrated OpenUpgrade: stock: error in migration script /opt/odoo/v8/OpenUpgrade/addons/stock/migrations/8.0.1.1/post-migration.py: maximum recursion depth exceeded while calling a Python object
2015-03-23 03:25:04,249 29802 ERROR nayar_migrated OpenUpgrade: maximum recursion depth exceeded while calling a Python object
  File "/opt/odoo/v8/OpenUpgrade/openerp/openupgrade/openupgrade.py", line 807, in wrapped_function
    func(cr, version)
  File "/opt/odoo/v8/OpenUpgrade/addons/stock/migrations/8.0.1.1/post-migration.py", line 703, in migrate
    migrate_stock_qty(cr, registry)
  File "/opt/odoo/v8/OpenUpgrade/addons/stock/migrations/8.0.1.1/post-migration.py", line 616, in migrate_stock_qty
    stock_move_obj.action_done(cr, uid, done_move_ids, context=None)
  File "/opt/odoo/v8/OpenUpgrade/openerp/api.py", line 241, in wrapper
  File "/opt/odoo/v8/OpenUpgrade/addons/stock/stock.py", line 2402, in action_done
    self.write(cr, uid, ids, {'state': 'done', 'date': time.strftime(DEFAULT_SERVER_DATETIME_FORMAT)}, context=context)
  File "/opt/odoo/v8/OpenUpgrade/openerp/api.py", line 241, in wrapper
  File "/opt/odoo/v8/OpenUpgrade/addons/stock/stock.py", line 1968, in write
    self.write(cr, uid, [move.move_dest_id.id], propagated_changes_dict, context=context)
  File "/opt/odoo/v8/OpenUpgrade/openerp/api.py", line 241, in wrapper
```

I'm checking

addons/stock/migrations/8.0.1.1/post-migration.py
pedrobaeza: Fix this part

legalsylvain:
Hi, 
I'm facing a problem, working on stock migration on a middle sized database. The DB has ~170.000 stock moves in a 'done' state. 
The call of the function 'migrate_stock_qty' to generate quants take a long time : > 10 hours, and it didn't finished. 2/3 of stock moves managed (I added some debug log in action_done function).
1/ @StefanRijnhart, @pedrobaeza, @other Did you faced the same problem ? Maybe there is some bug in my DB but I don't think so.
2/ Do you think it could be possible to realize a SQL request to create quants instead of the current trick (set all stock moves in draft state and call action_done).
3/ other ORM ideas to reduce the duration ? 

Thanks for your answer.

pedrobaeza:
@legalsylvain, I also faced a problem that I have put before. I'm working on a SQL in OpenUpgrade API method, but frozen for now until next week.

legalsylvain:
Yes, I saw it, but in my case I don't have recursion problem. How many stock move do you have and how long does this process (before your error) ?
Other point : in action_done function, the last call is : 

```
self.write(cr, uid, ids, {'state': 'done', 'date': time.strftime(DEFAULT_SERVER_DATETIME_FORMAT)}, context=context)
```

I don't think that using current server date is a good point. 
Maybe we could work together on SQL alternative ?

pedrobaeza:
I was in a similar number of records, and it took about 20 h, so it's good to have an SQL alternative. One of the bottleneck is also m2o_to_x2m function, but there's a lot of work to do.

Probably I can't work this week on this, but if you want to start...

legalsylvain:
Hi @all,
I was talking with @pedrobaeza. I know that what I'll propose you is not a very regular process but I have a problem with this PR. It is opened since October, 28 2014.
I think that it is too big to release. In other hand, it globally works. So, to avoid to block openupgrade process, I propose you to realize the Merge of this PR, and after to realize some little PR, more easy to review to improve duration (my case), or to fix some very specific case unmanaged for the time being. 
We could mark stock as 'done (experimental)' in documentation until more reviewers consider it as mature. 
How do you think about that ?

Regards.

paulius-sladkevicius:
I agree with @legalsylvain, this is open too long and looks already used to migrate db's, experimental sounds good. Could be it will make more attraction to make migrations and push fixes based on this great work then trying to work on this PR.

legalsylvain:
@hbrunn , @StefanRijnhart , @pedrobaeza  @all are you against this point of view.
If not, I'll merge this PR at the end of the week, setting this module as 'done (experimental)' and work on a little PR to improve quants creation. (reduce script duration)

thomaspaulb:
I already used it and for me it worked, there are probably still some glitches somewhere but they can be fixed whenever they pop up I think?

StefanRijnhart:
Yes, merging with experimental or alpha status is fine.

pedrobaeza:
I merge then and put it as experimental

PR #141, opened 2014-11-28
[FIX] ensure an email for admin to have valid message_posts during

hbrunn:
migration


pedrobaeza:
I understand then that `mail.catchall.domain` is used in all parts when mail is needed? If not, you should write that value to admin email, or take also the default one.

hbrunn:
It's either an email address or the alias: https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/mail/mail_message.py#L171 - and the alias domain comes from this parameter: https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/mail/mail_alias.py#L62

pedrobaeza:
Thanks for the explanation!

:+1:

StefanRijnhart:
Can you add a log entry when the email is set?

hbrunn:
done

StefanRijnhart:
Grand, thanks. :+1:

PR #144, opened 2014-11-28
[ADD] drop some views that prevent changing field types

hbrunn:


pedrobaeza:
:+1:

sbidoul:
I've had issues with the report_document_user view. Can you drop it too?

:+1:

hbrunn:
@sbidoul I think that should be done in the prescript of the module creating this view (which would be document)

sbidoul:
@hbrunn agreed

pedrobaeza:
@sbidoul I have made for you on #154 :wink:

Issue #137, opened 2014-11-25
Migrate rest of OpenUpgrade branches to GitHub


pedrobaeza:
We can use the same script that Odoo uses for merging the three projects and respect the history to migrate to Github entire project.

What do you think?

bwrsandman:
:+1: 
- [x] 6.0
- [x] 6.1
- [x] 7.0
- [ ] master?

hbrunn:
yes, that would be more than welcome. It's quite a bit of work though, so whoever volunteers, go ahead. Then we can also get rid of the convoluted code in migrate.py to support bzr and git

hbrunn:
...but please also to 6.0 then too

pedrobaeza:
Yeah, I was doing 6.0 tonight :wink: 

So, all is done now! Please check it and tell me if you any problem.

hbrunn:
thanks a lot @pedrobaeza ! I use those branches now in #139 and did a quick test, seems to work fine

StefanRijnhart:
Hi @pedrobaeza,

did you simply do a bzr import of the lp:openupgrade-\* projects? That means we lose the connection with the Odoo git projects. I believe the correct way is to port the diff between launchpad Odoo and launchpad OpenUpgrade onto a branch of GitHub Odoo like I did for 5.0

pedrobaeza:
More or less: I take Odoo script (https://github.com/odoo/odoo/wiki/GitHub-Transition#migration-from-bazaar) and adapt for OpenUpgrade. I haven't done that because it's a lot of more work that I think it doesn't worth the while:
- OpenUpgrade "Odoo/OpenERP" is frozen when the mature release was out for each version.
- In Github, we can still merge branches, make cherry-pickings, etc.

But if you think there is any advantage on doing that way, please point me to a little walkthrough for doing it and I'll do it.

StefanRijnhart:
What I did was branch both lp:openupgrade-server/5.0 and its corresponding bzr revision of lp:openobject-server/5.0. That allowed me to take the diff between these and apply to a GH branch of Odoo 5.0. For later versions, you would have to do the same for the addons. You could make separate commits, such as:
- changes to core files
- openupgrade_records module
- migration scripts

for easier review.

pedrobaeza:
[...]

PR #147, opened 2014-12-03
[IMP] speedup commercial field sync

hbrunn:


pedrobaeza:
:+1:

hbrunn:
wait for #137

pedrobaeza:
You have to rebase due to the change of branch

hbrunn:
rebasing done

StefanRijnhart:
You are replacing a call from _fields_sync with _commercial_sync_from_company. It looks like this way, addresses are not propagated.

Then, if you would solve that by calling update _address manually, there seems to be a fringe case in which there is a partner structure A -> B -> C where B and C both would want to use A's address. If C comes before B in the loop, then C syncs B's empty address, after which B syncs A's address, leaving C's empty. Not sure what the real impact of this would be though.

StefanRijnhart:
Hope I'm not misusing the Question tag when I am looking for a replacement of Launchpad's 'Needs information' review.

pedrobaeza:
As you can see here: https://github.com/OpenUpgrade/OpenUpgrade/blob/7.0/openerp/addons/base/res/res_partner.py#L467, that method makes commercial fields sync + address sync if "Use parent address" is marked. As this field is not set in migration, the result is the same.

pedrobaeza:
@hbrunn, please rebase again.

StefanRijnhart:
Thanks for the insight! That makes it a lot easier to accept this. :+1: after rebase.

hbrunn:
rebase done

PR #149, opened 2014-12-03
[ADD] Migration for l10n_es_account_invoice_sequence

pedrobaeza:
This is only for spanish installations, but for the rest of the people, it doesn't hurt.

The rest of the diff it's by my IDE that auto-fixes some PEP8, so it also doesn't hurt.


hbrunn:
:+1:

PR #154, opened 2014-12-06
[IMP] document: Remove annoying SQL view that prevents some operations

pedrobaeza:
Following #144, this mades the same for document module

PR #159, opened 2014-12-12
[IMP] resource: remove sql view created by hr_holidays that prevents column changings

christophlsa:
I had the error:

```
2014-12-12 09:23:35,921 23335 ERROR testdb openerp.sql_db: bad query: ALTER TABLE "resource_resource" ALTER COLUMN "name" TYPE VARCHAR
DETAIL:  rule _RETURN on view hr_holidays_remaining_leaves_user depends on column "name"
```

This change drops the view to be able alter the column (copied and adapted from #154).

### DIFF
--- a/addons/resource/migrations/8.0.1.1/pre-migration.py
+++ b/addons/resource/migrations/8.0.1.1/pre-migration.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


christophlsa:
Oh wait. For me the pre migration script is not called (no line in logfile and same error). Do you know why it is not called?

christophlsa:
Ok was my fault. Commit is fixed in branch/pullrequest.

pedrobaeza:
:+1:

legalsylvain:
:+1:

hbrunn:
:+1:

PR #160, opened 2014-12-12
[IMP] analytic: remove sql view that prevents column changings

christophlsa:
Small fix for error (similar to #159):

```
2014-12-12 09:25:21,861 23335 ERROR testdb openerp.sql_db: bad query: ALTER TABLE "account_analytic_line" ALTER COLUMN "name" TYPE VARCHAR
DETAIL:  rule _RETURN on view analytic_entries_report depends on column "name"
```


pedrobaeza:
Thanks :+1:

legalsylvain:
:+1:

hbrunn:
:+1:

PR #161, opened 2014-12-12
[IMP] base: remove sql view that prevents column changings

christophlsa:
Small fix for error (similar to #159):

```
2014-12-12 09:22:42,129 23335 ERROR testdb openerp.sql_db: bad query: ALTER TABLE "ir_attachment" ALTER COLUMN "datas_fname" TYPE VARCHAR
DETAIL:  rule _RETURN on view report_document_user depends on column "datas_fname"
```


pedrobaeza:
:+1:

hbrunn:
:+1:

StefanRijnhart:
Thanks! :+1:

Issue #151, opened 2014-12-04
account bank statement migration issues


pankk:
Hi,

For the account bank statement migrations we are facing some issues. I was wondering if anyone could to share their findings on the following issues:
The bank statements when in state open, when migrated, cannot be reconciled.
Also statements which are canceled after migrating cannot be used anymore.

pedrobaeza:
You are not indicating which version of OpenUpgrade you are using. It seems that the migration is not completed for that models (account.bank.statement and childrens).

pankk:
Thanks for the reply. I'm using version 8.0 of OpenUpgrade.

pedrobaeza:
All the work on statement models seems to be done (https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/account/migrations/8.0.1.1/openupgrade_analysis_work.txt), and nothing related to states and so on, so I don't what's happening.. Maybe a custom module?

pankk:
Fixed in PR #167. Simply made a legacy column for account_bank_statement.account_id. See PR for more information.

PR #167, opened 2014-12-15
[FIX]Do not migrate account_id to allow reconciliation of existing bank ...

pankk:
...statements.

This resolves issue #151. When account_id from account.bank.statement.line is migrated from 7 to 8, the reconciliation workflow in 8 uses the account_id to reconcile automatically, skipping the step where it is possible to make reconciliation adjustments. This also prevented closed bank statements from being adjusted.
Renaming the account_id field allows us to use the new reconciliation process for existing bank statements.

The help argument of account_bank_statement_line.account_id further clarifies this: "This technical field can be used at the statement line creation/import time in order to avoid the reconciliation process on it later on. The statement line will simply create a counterpart on this account".

### DIFF
--- a/addons/account/migrations/8.0.1.1/pre-migration.py
+++ b/addons/account/migrations/8.0.1.1/pre-migration.py
@@ -25,6 +25,7 @@ column_renames = {
     'account_bank_statement_line': [
         ('analytic_account_id', None),
         ('type', None),
+        ('account_id', None),
     ]
 }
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Makes sense. :+1:

hbrunn:
:+1:

pedrobaeza:
:+1:

PR #171, opened 2014-12-16
8.0 fix purchase

gfcapalbo:
[fix] added mandatory comma at the end of a single element tuple.


StefanRijnhart:
Did you mean to pull into the dev branch?

hbrunn:
same here for #123 - I'm not sure if PRs against the OpenUpgrade-dev branches show up here, so I think it's a good idea to PR against OpenUpgrade and fast-forward the original PR in question

StefanRijnhart:
Well, the submitter is not in the dev team so he cannot do that. I think pr's on the dev branches is fine. I think as a subscriber I should get notifications right?

PR #174, opened 2014-12-17
[FIX] get pool object, flake8

hbrunn:


StefanRijnhart:
Thanks. I'll try to get travis up for 7.0 because I think we started to rely on that. Fasttracking this blocking bug.

PR #176, opened 2014-12-17
[FIX] Some modules are not marked for upgrade with -u all

StefanRijnhart:
In standard Odoo, '--update all' just means: '--update base + upward (installed) dependencies'.
This breaks the chain when only new (glue) modules are encountered. E.g. purchase in 8.0 depends
on stock_account and report, both of which are new. They may be installed, but purchase as an
upward dependency is not selected for upgrade. Therefore, explicitely select all installed
modules for upgrading in OpenUpgrade.

### DIFF
--- a/openerp/modules/loading.py
+++ b/openerp/modules/loading.py
@@ -357,7 +357,18 @@ def load_modules(db, force_demo=False, status=None, update_module=False):
 
             mods = [k for k in tools.config['update'] if tools.config['update'][k]]
             if mods:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Thanks for the patch. :+1:

Does it apply to previous versions?

StefanRijnhart:
Thanks for the review, quick as ever ;-) No, I think this is the first time we encountered this, because an existing module only depends on new modules in the new version.

pedrobaeza:
OK, good to know.

Thanks again.

guewen:
:+1:

hbrunn:
:+1:

PR #177, opened 2014-12-19
[UPG] rename mrp_jit -> procurement_jit

StefanRijnhart:


rvalyi:
:+1: finally some simple stuff to review :-)

pedrobaeza:
Don't you need to rename the module also in base module? As far as I know, with this, you are only renaming it for analysis purpose, but not really rename it in a migration.

hbrunn:
@pedrobaeza not since https://github.com/OpenUpgrade/OpenUpgrade/commit/a58e37604f418dc40d73961f55438f4798515b93
:+1:

pedrobaeza:
Ok, thanks for the clarification.

:+1:

PR #180, opened 2014-12-23
[FIX] fix wrong filter in search view (account/sale)

pankk:
the field section_id is created in addon sale, but used in the account
reporting views.  This commit moves the search view definition
in the correct file.

This fixes a TypeError when clicking on a + sign in the account.invoice.report.graph view, similar to what is reported at: https://www.odoo.com/nl_NL/forum/help-1/question/error-with-customer-invoice-graph-view-70665

Note that this is a bugfix from Odoo 8.0. I am not entirely clear on what the strategy is to keep up with Odoo 8.0 changes, but from https://github.com/OpenUpgrade/OpenUpgrade/pull/103#issuecomment-56507805 it appears that we're applying fixes from Odoo 8.0 like this instead of making big merges?

### DIFF
--- a/addons/account/report/account_invoice_report_view.xml
+++ b/addons/account/report/account_invoice_report_view.xml
@@ -73,7 +73,6 @@
                 <group expand="1" string="Group By">
                     <filter string="Partner" name="partner_id" context="{'group_by':'partner_id','residual_visible':True}"/>
                     <filter string="Salesperson" name='user' context="{'group_by':'user_id'}"/>
-                    <filter string="Sales Team" domain="[]" context="{'group_by':'section_id'}" groups="base.group_multi_salesteams"/>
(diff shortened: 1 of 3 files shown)


christophlsa:
Is this necessary for the migration? My opinion is that for testing and production use the real odoo repository should be used. So we don't need this fix I think.

pedrobaeza:
Christoph is right, this branch is only intended for migration, not for keeping as production environment, so patches are not proposed here. Sometimes, we merge upstream updates to keep it more or less on date, but it's not mandatory, because DB layout is kept across an entire version.

Thanks anyway.

I close the PR.

StefanRijnhart:
Right. We don't maintain OpenUpgrade for this purpose. Please propose to Odoo/8.0 and perhaps to OCB/8.0.

PR #181, opened 2014-12-26
[IMP] Remove openerplib dependency if openupgrade_records is not used (any regular migration)

pedrobaeza:


StefanRijnhart:
Thanks! Can you just import it where it is used (in only one place, I believe)? Then you don't have to catch the exception either.

pedrobaeza:
OK, done!

StefanRijnhart:
Thanks! :+1:. Fasttracking this trivial change.

PR #182, opened 2014-12-27
[FIX] Match purchase lines and procurements by move_dest_id first (and faster)

StefanRijnhart:
[IMP] Warn about unmatched files
[ADD] Populate purchase procurement's move_ids
[RFR] Rename procurement purchase_id field in the right module


pedrobaeza:
Thanks. :+1: 

I will check if there's any procurement that doesn't match by purchase_line_id. Had you got any case? Can we make it the other way in that case?

Regards.

StefanRijnhart:
Not sure what you mean. I do check for dangling procurements. We need the same thing for MRP. Will get around to that soon.

pedrobaeza:
As you have included a logger warn if there are procurements that cannot be matched by purchase_line_id, I told about letting the old method for matching these ones.

hbrunn:
:+1:

pedrobaeza:
I merge this old one.

PR #185, opened 2014-12-30
Remove hr_holidays_meeting_id_fkey constrain to avoid migration issues.

mvaled:
- The constraint will be reset by the ORM later.
- Not doing it may abort the migration.

PR #187, opened 2014-12-31
Remove hr_holidays_meeting_id_fkey constrain to avoid migration issues.

mvaled:
- The constraint will be reset by the ORM later.
- Not doing it may abort it altogether.

This replaces PR #185.  See my next-to-last comment in that for rationale.


pedrobaeza:
:+1: again

StefanRijnhart:
Thanks for the follow up! :+1:

PR #190, opened 2015-01-01
[ADD] Files lost in the 6.1 migration to git

StefanRijnhart:
Add some files lost in the 6.1 migration to git. Some directories were missing from the doc directory. Instead of re-adding them, remove doc altogether as it is maintained on the latest version of OpenUpgrade.


pedrobaeza:
I have checked and some files are indeed missing. But why are you removing entire doc folder? Is for only maintaining it on last version? I see that OK, but I wonder if we can keep here `modules60-61.rst` with a possibility to link in 8.0 this file, for each possible PR against this branch that modifies also the corresponding documentation. Maybe a hard include link via Github blob URLs?

StefanRijnhart:
Looks like we would need to look to git submodules to do this properly, but it seems a bit overengineered and complicated for that. I don't really care that much about keeping an obsolete version of the docs there. The person to merge the PR can simply edit the coverage file online on GitHub in the current version of the docs.

pedrobaeza:
:+1: for the PR

What about making an RST link?

StefanRijnhart:
You mean, include an RST document that includes a file from https://doc.therp.nl/openupgrade? Note that I actually replace the documentation by a small readme that links to the online documentation: https://github.com/StefanRijnhart/OpenUpgrade/blob/6.1-git_missing_from_lib/openerp/openupgrade/doc/README.txt. Is that not enough?

pedrobaeza:
No, I mean a link in the 8.0 documentation, that it's compiled and show in https://doc.therp.nl/openupgrade/, to this file `modules60-61.rst` here in 6.1 branch, vía an rst link to https://github.com/OpenUpgrade/OpenUpgrade/blob/6.1/openerp/openupgrade/doc/source/modules60-61.rst. And the same for the rest of the branches. This way, we can maintain module coverage in the corresponding branch. Do you understand me?

StefanRijnhart:
Sounds fine. Any idea on how to include a file from an online location in the RST index?

pedrobaeza:
Let me investigate about that, but I have no control nor knowledge about the compiling to readthedocs, so my tests are going to be partial.

StefanRijnhart:
Good luck. I just spend some time on it and I can't seem to figure it out...

pedrobaeza:
You're right: submodules implies a heavy weight, because you have to treat with full repos, and `.. include::` doesn't allow external links.

We have to resigned for now...

hbrunn:
:+1:

PR #194, opened 2015-01-03
Fixed rename_xmlids to log error message instead of crashing (7.0)

### LINKED ISSUE #192: Error message is not shown, if dots are missing from rename_xmlids parameter xmlids_spec
Impacted versions:
- 7.0 and 8.0

Steps to reproduce:
1. create a new pre-migration.py for any module
2. call openupgrade.rename_xmlids(cr, [('my_field', 'my_field_old')]) in the migrate method

Current behavior:
- OpenUpgrade crashes with error message "IndexError: tuple index out of range"

Expected behavior:
- OpenUpgrade logs an error message "'Cannot rename XMLID my_field to my_field_old: need the module reference to be specified in the IDs"

mlaitinen:
Correct error message is now shown when rename_xmlids parameter is missing dots. Fixes #192 for branch 7.0

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -175,7 +175,7 @@ def rename_xmlids(cr, xmlids_spec):
     although they were still being defined in their respective module).
     """
     for (old, new) in xmlids_spec:
-        if not old.split('.') or not new.split('.'):
+        if '.' not in old or '.' not in new:
             logger.error(
                 'Cannot rename XMLID %s to %s: need the module '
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

StefanRijnhart:
Thanks! :+1:

StefanRijnhart:
Could you port this to 8.0 as well? Then we are sure to keep this fix alive in forthcoming versions.

StefanRijnhart:
Oh, you already did that in #193!

PR #195, opened 2015-01-03
[ADD] l10n_es module renamings

pedrobaeza:
These are some modules from Spanish localization that were renamed during 5.0 > 6.0 change.

This doesn't hurt any other migration.


StefanRijnhart:
:+1:

legalsylvain:
:+1:

PR #197, opened 2015-01-04
[FIX] update_module_names method should also consider module dependencies

pedrobaeza:
When migrating from 5.0 to 6.0 with renamed modules, if some modules are renamed, the dependencies from these renamed modules stay unchanged, and there are some warnings about the inexistence of them.

I don't know if this also applies to other versions of OpenUpgrade.

The rest of the diff is the automatic formatting of my IDE, that it's useful for PEP8 compliant.

### DIFF
--- a/bin/openupgrade/openupgrade.py
+++ b/bin/openupgrade/openupgrade.py
@@ -47,7 +47,7 @@ __all__ = [
     'rename_models',
     'rename_xmlids',
     'get_legacy_name',
-]    
+]
 
 def load_data(cr, module_name, filename, idref=None, mode='init'):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I see that from 7.0, this SQL is already included (https://github.com/OpenUpgrade/OpenUpgrade/blob/7.0/openerp/openupgrade/openupgrade.py#L384), so I supposed that it's a bug not backported. I'm going to make also the 6.1 PR.

StefanRijnhart:
:+1:

hbrunn:
:+1:

PR #199, opened 2015-01-05
[ADD] l10n_es migration script.

pedrobaeza:
- Deletes old account chart template data to allow a clean installation of new account chart template structure
- Map removed account types to new ones


StefanRijnhart:
Looks good, even though I keep doubting about the need to remove update data. Should be harmless though. Only thing left is the two default values (model: ir.values) which are defined in the 7.0 version. Do they need removal as well? You do remove the records they point to.

pedrobaeza:
Which ir.values do you refer? There's only new ones.

StefanRijnhart:
In 7.0:

account_chart_assoc.xml:        &lt;record id="l10nES_chart_template_assoc_value" model="ir.values">
account_chart_pymes.xml:        &lt;record id="l10nES_chart_template_pymes_value" model="ir.values">
account_chart.xml:        &lt;record id="l10nES_chart_template_value" model="ir.values">

pedrobaeza:
Ah, OK, that's old references, and they are harmless. I don't think it deserves to be treated here. databe_cleanup can do the job.

StefanRijnhart:
OK, fine. :+1:

pedrobaeza:
Please proceed then with the merge.

PR #200, opened 2015-01-06
[IMP] don't error out when trying to drop nonexisting constraints

hbrunn:
which for example happens often for migrated databases...


StefanRijnhart:
Interesting. :+1: for the code, but does this signal a problem with the upgrade process? When does this occur exactly, on databases previously migrated in a previous edition of openupgrade, or on picking up the migration after an exception in a migration script?

hbrunn:
In my case, it's a 6.1 -> 7.0 -> 8.0 migration, and one of those constraints didn't exist in 6.1 but was added in 7.0. Given the constraint couldn't be set up due to some data mismatch, it will break in the 8.0 migration

StefanRijnhart:
Of course, thanks for the explanation. :+1: again.

PR #204, opened 2015-01-07
[PRT] Backport improvements to the records module from 8.0

StefanRijnhart:


.gitignore
```diff
@@ -0,0 +1,2 @@
+*.pyc
```
pedrobaeza, line 1: Don't we want a more ellaborated gitignore like the one in other repos?

bin/addons/openupgrade_records/model/comparison_config.py
```diff
@@ -23,11 +23,25 @@
...
+    from openerp.addons.openupgrade_records.lib import apriori
+except ImportError:
+    from openupgrade_records.lib import apriori
+
+try:
+    from openerp.osv.orm import Model, except_orm
+    from openerp.osv import fields
+    from openerp.tools.translate import _
+except ImportError:
+    from osv.osv import osv as Model, except_osv as except_orm
```
pedrobaeza, line 39: We know that this is going to fall always in this part for v5, but good strategy to keep all versions with the same code :wink:

pedrobaeza:
The rest LGTM

.gitignore
StefanRijnhart, line 1: I'm not that interested. As it contains 'lib', which is a directory in this module, it has so far caused more problems than it has solved.

bin/addons/openupgrade_records/model/comparison_config.py
StefanRijnhart, line 39: Exactly. That allowed me to simply copy the code over from 8.0 and it worked (if I remember correctly).

pedrobaeza:
:+1:

pedrobaeza:
Should we proceed with the merge with this old one?

StefanRijnhart:
Yes, that should be good.

PR #206, opened 2015-01-07
[ADD] migration of sale module

StefanRijnhart:


christophlsa:
Seems to work. Can you explain why we need the "noupdate_changes.xml"? Thanks. :+1:

StefanRijnhart:
These are changes in data marked with noupdate="1", which as the attribute implies is not updated upon upgrade of the module. I created this file with scripts/compare_noupdate_xml_records.py from the 7.0 distribution (I still need to re-add it to 8.0). Some entries could qualify removal because they could overwrite meaningful customizations but not in this case.

addons/sale/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,62 @@
...
+    date field to datetime."""
+    cr.execute(
+        """
+        SELECT distinct(rp.tz)
+        FROM sale_order so, res_users ru, res_partner rp
+        WHERE rp.tz IS NOT NULL
+            AND so.user_id=ru.id
+            AND ru.partner_id=rp.id
+        """)
+    for timezone in [row[0] for row in cr.fetchall()]:
```
hbrunn: `for timezone, in cr.fetchall()` saves you looping the list twice

christophlsa:
Thanks for the explanation. :)

I think you forgot to change the xmlids of these XML records:
- sale_crm.account_invoice_form
- sale_crm.account_invoice_groupby_inherit
- sale_crm.account_invoice_tree

Or is it not necessary for views?

StefanRijnhart:
No, views don't need migration. They will simply be recreated with another id. This is not a problem, because their ids are not hardcoded anyware (or should not be).

addons/sale/migrations/8.0.1.0/post-migration.py
StefanRijnhart: Thanks, took your suggestion.

hbrunn:
:+1:

addons/sale/migrations/8.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,67 @@
...
+    ('sale_crm.mt_salesteam_order_sent', 'sale.mt_salesteam_order_sent'),
+    ]
+
+
+def preserve_date_order(cr):
+    """ Copy date_order column for access to the original data after migration
+    from date to datetime type."""
+    cr.execute(
+        """
+        ALTER TABLE sale_order
```
pedrobaeza, line 57: I think you can make a simple column rename, because in post migration, you're getting values from legacy column, not from the new.

StefanRijnhart, line 57: But I need a default because I don't necessarily cover all rows (e.g. for users without a timezone).

pedrobaeza, line 57: OK, I see. Thanks for the clarification.

pedrobaeza:
:+1:

May I proceed with the merge?

StefanRijnhart:
Yes please!

PR #211, opened 2015-01-08
analysis of sale_crm - Nothing to do

christophlsa:
I looked over the sale_crm addon. At my opinion there is nothing to do. Some fields and views are moved to other addons. Two XML records (not views) are moved to the sale addon and already handled in Pull Request #206.


StefanRijnhart:
Thanks. Did you consider trying to retrieve the campaign, medium and source from any related opportunities?

legalsylvain:
Hi, 
Yes for campaign_id and medium_id, it could be possible.

for (v8) sale_order.**campaign**_id (crm.tracking.campaign) 
-> from (V7) crm_lead.type_id (old model : crm.case.resource.type) 

for (v8) sale_order.**medium**_id (crm.tracking.medium)
-> from (V7) crm_lead.channel_id (old model : crm.case.channel)

But for (v8) sale_order.**source**_id (crm.tracking.source)
This model seems to be new. It isn't ? Or did I miss something ? @StefanRijnhart 

@christophlsa : did you planned to update your work ? If not, I can help on that point.

Except that new fields, the rest of the analysis looks good to me. Thanks.

Regards.

christophlsa:
Sorry I have atm not that much time for that. Help would be nice.

StefanRijnhart:
@legalsylvain medium_id is the old channel_id. Table and model names have changed as well though. Check the crm migration script.

legalsylvain:
My apologize : in my last comment I did a bad copy past. (I fixed)
My question is about the **source_id** field.

Regards.

StefanRijnhart:
Yes, source_id is new. Oh, I see: nothing to migrate there indeed. So it's just the campaign and the medium then. Thanks for noting!

legalsylvain:
Hi,
I don't see any fiable relation between sale_order and crm_lead.
The only field I see is sale_order.origin that is a string and cannot be used.
Did I miss something ?
@StefanRijnhart How did you thought to retrieve the information ?

Regards.

StefanRijnhart:
You are right, crm case is only linked to the sale order through the origin as _('Opportunity: %s') % str(case.id). But you could still use these numeric ids to resolve the sale's crm case, with a double check on the partner being the same.

pedrobaeza:
Any advance on this?

legalsylvain:
Not from my part. i understand what Stephan says, but for me, this solution is not very "deterministic" because origin is a free text, that user can modify. 
Morever, the field can be "Oportunity : XXX" in english but "Opportunité : XXX" in french, and maybe "XXX : blablabla" dépending of translation process.

StefanRijnhart:
Take any origin ending with : + numeric. Check if numeric is an id with the same partner as the sale. Can't go wrong.

pedrobaeza:
Is this still a WIP?

legalsylvain:
No this is an obsolete PR.
I realized what @StefanRijnhart asked in a new PR : #246.
I let this PR opened while #246 is not  reviewed. 
Regards.

pedrobaeza:
OK, I close this then, because the link remains although closed.

PR #219, opened 2015-01-22
[IMP] get user's partner_id from employee record if available

hbrunn:
For installations with hr installed, we need a different strategy
as the employee's data probably already lives in one of the addresses
linked with address_home_id and address_id. In this case, we simply
link the address' partner if it exists.

### DIFF
--- a/openerp/addons/base/migrations/7.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/7.0.1.3/pre-migration.py
@@ -178,11 +178,6 @@ def create_users_partner(cr):
     to track the partners we create here. These are not necessarily
     the same, especially if you have a custom module implementing
     similar behaviour for 6.1.
-
-    For installations with hr installed, we need a different strategy
-    as the employee's data probably already lives in one of the addresses
-    linked with address_home_id and address_id. In this case, we simply
-    link the address' partner if it exists.
     """
     if not openupgrade.column_exists(
             cr, 'res_users', 'partner_id'):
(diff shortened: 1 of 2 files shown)


openerp/addons/base/migrations/7.0.1.3/pre-migration.py
```diff
@@ -202,6 +207,31 @@ def create_users_partner(cr):
...
+                cr, 'resource_resource', 'user_id'):
+        # if hr is installed, get partners from there if possible
+        cr.execute(
+            '''UPDATE res_users SET partner_id=u.partner_id
+            FROM (
+                SELECT r.user_id, a.partner_id
+                FROM hr_employee e
+                JOIN resource_resource r ON e.resource_id=r.id
+                JOIN res_partner_address a ON COALESCE(
+                    e.address_id, e.address_home_id)=a.id
```
StefanRijnhart: Can you reverse the order? In a small sample of databases I noticed that the address_id was always the company address. Maybe exclude the company addresses explicitely anyway.

hbrunn: I think the latter is better, because in general we don't want the user's home address.

How about something like 

```
COALESCE(e.address_id, e.address_home_id)=a.id WHERE r.user_id IS NOT NULL AND a.partner_id IS NOT NULL AND a.partner_id <> ${base.main_partner}
```

?

hbrunn:
I moved all of this to hr's migration, this way we get the partner we want

StefanRijnhart:
Thanks, looks good now :+1:

pedrobaeza:
:+1:

PR #222, opened 2015-02-02
8.0 [ADD] l10n fr

legalsylvain:
Hi all, 
Working on l10n_fr migration, I realized that the analysis file was not up to date. (the analysis was wrong).
I so updated the analysis file following this procedure : 
https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/openerp%2Fopenupgrade%2Fdoc%2Fsource%2Fanalyse.rst

I conclude that there is nothing to do.

@StefanRijnhart (or other): 
I had a pretty weird log when I realized the "Generate Records" Step: (on both instance).

```
Cannot find xml_id l10n_fr.fiscal_position_template_intraeub2b
Cannot find xml_id l10n_fr.fiscal_position_template_intraeub2b
Cannot find xml_id l10n_fr.fiscal_position_template_import_export
Cannot find xml_id l10n_fr.fiscal_position_template_import_export
```

Do you have an idea ?

Maybe due to this unusual line : https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons%2Fl10n_fr%2Ffr_fiscal_templates.xml#L8

(Non blocking point, I checked manually in both versions and the fiscal position are unchanged).


pedrobaeza:
Seems OK for me. :+1:

StefanRijnhart:
Thanks! :+1:

PR #223, opened 2015-02-03
8.0 [ADD] sale_order_dates

legalsylvain:
Trivial PR for 'sale_order_dates' module.


pedrobaeza:
:+1:

StefanRijnhart:
Please migrate the date values to datetimes that reflect the original date in the user's timezone. See how I did that for the sale order date https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/addons/sale/migrations/8.0.1.0/post-migration.py#L27

pedrobaeza:
@StefanRijnhart, why do you need to convert to user tz in these cases? As the previous data is a date (in example, 2014-12-31), you will have in its conversion to datetime the following: 2014-12-31 00:00:00. Converting into timezones, you will get something like 2014-12-30 22:00:00 for GMT-2, and I think that it's not intended. Am I alright or I'm mistaken in something?

StefanRijnhart:
If a sale order was dated on 03-02-2015 in 7.0 and you don't compensate for the timezone, the UTC datetime in 8.0 will be 03-02-2015 00:00 which will show up in my timezone as 02-02-2015 23:00 which is a **different day**. So I modified the sale order to show up as the correct date with time 00:00 in the salesman's timezone.

pedrobaeza:
Thanks for the explanation, @StefanRijnhart. Yeah, I think about it on the contrary side, but you're right. Do you think this deserves an OpenUpgrade API method? I think so.

StefanRijnhart:
Yes, an API method could be useful!

legalsylvain:
I writted it. But untested for the moment. I'll push it tonight.

pedrobaeza:
Cool! Waiting the corresponding PR.

legalsylvain:
Hi, 
Thanks for you remark, @StefanRijnhart, I had not thought of timezone.
I refactored the functions written in 'sale' module.
- API & 'sale' commit tested;
- 'sale_order_dates' commit untested, as it depends of non migrated modules;

Thanks for your review.

openerp/openupgrade/openupgrade.py
```diff
@@ -135,6 +137,31 @@ def load_data(cr, module_name, filename, idref=None, mode='init'):
 table_exists = openupgrade_tools.table_exists
 
 
+def copy_columns(cr, column_spec):
+    """
+    Copy table columns. Typically called in the pre script.
+
+    :param column_spec: a hash with table keys, with lists of tuples as \
+    values. Tuples consist of (old_name, new_name, type).  Use None for \
```
pedrobaeza: It could be better if we just take the column type from the old_name.

openerp/openupgrade/openupgrade.py
```diff
@@ -135,6 +137,31 @@ def load_data(cr, module_name, filename, idref=None, mode='init'):
 table_exists = openupgrade_tools.table_exists
 
 
+def copy_columns(cr, column_spec):
+    """
+    Copy table columns. Typically called in the pre script.
+
+    :param column_spec: a hash with table keys, with lists of tuples as \
```
pedrobaeza, line 144: For docstring, you don't need `\`. You just need to indent a little more (two spaces is the usual), the second line.

openerp/openupgrade/doc/source/modules70-80.rst
```diff
@@ -410,7 +410,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |sale_mrp                           | Nothing to do                     |
 +-----------------------------------+-----------------------------------+
-|sale_order_dates                   |                                   |
+|sale_order_dates                   | Nothing to do                     |
```
pedrobaeza: Change this to 'Done'

openerp/openupgrade/openupgrade.py
legalsylvain: In that case yes, but we can imagine that the type of the new field is not exactly the same as the old one. 
Sample : 

```
column_copies = {
    'sale_order': [('date_order', 'new_field', 'datetime'),],
}
```

So I prefer to keep as it for more modularity.

pedrobaeza: When the copy is done, the type is still the same as far as I know (it is still date), because it is executed on pre-script. Anyways, we can have it as an optional parameter, that if not filled, the type is infered, and if set, the type is overwritten.

legalsylvain: Yes, my remarks wasn't for that case, but for a theoritical case.
Anyway, I made type field optional.
Regards.

pedrobaeza:
Thank you very much.

:+1:

StefanRijnhart:
Thanks, great work! One nag: you need to fix the indentation of your docstrings:

/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.copy_columns:3: ERROR: Unexpected indentation.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.copy_columns:6: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.move_field_m2o:9: ERROR: Unexpected indentation.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.move_field_m2o:10: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.move_field_m2o:23: WARNING: Field list ends without a blank line; unexpected unindent.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.deactivate_workflow_transitions:8: WARNING: Field list ends without a blank line; unexpected unindent.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.reactivate_workflow_transitions:5: WARNING: Field list ends without a blank line; unexpected unindent.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.date_to_datetime_tz:7: ERROR: Unexpected indentation.
/home/oerintupg80/odoo/openerp/openupgrade/openupgrade.py:docstring of openupgrade.date_to_datetime_tz:8: WARNING: Block quote ends without a blank line; unexpected unindent.

BTW. I had trouble building the documentation locally. For your testing purposes: I was able to get it building by installing sphinx and autodoc in my virtualenv, and adding another path to openerp/openupgrade/doc/source/conf.py:

-sys.path += [os.path.abspath('../../'), os.path.abspath('../../../')]
+sys.path += [os.path.abspath('../../'), os.path.abspath('../../../'), os.path.abspath('../../../../')]

I'll investigate that further at some point.

legalsylvain:
Hi, what is the command / the tools to check if the docstring is well formated ?

StefanRijnhart:
The output is generated while building the documentation. Go to openerp/openupgrade/doc and do 'make html'.

legalsylvain:
Hi, 
I tried to fix all docstring warning / error for openupgrade.py file. (I'm really not a docstring writter expert, so my apologise if it's not the exact way to write a docstring)
There is other bugs in openugrapde_XX.py  but I consider that it is out of the scope of that PR. 
Regards.

pedrobaeza:
According specification, you can put multi-line param doc indenting second and succesive lines, but it seems that docstring doesn't like it...

pedrobaeza:
All the questions have been honoured, so I merge this PR.

PR #229, opened 2015-02-06
[ADD] rename of web_widget_classes

hbrunn:


pedrobaeza:
Is this something that you carry from 6.1 or from 7.0?

hbrunn:
The module exists for 6.1 as web_widget_classes, and in my PR for 8.0 you asked for the rename. If ever there's a 7.0 version, it will also be called by the new name, so I think we should do the rename when going from 6.1 to 7

StefanRijnhart:
Trivial ;-) :+1:

pedrobaeza:
:+1: I merge then

PR #231, opened 2015-02-09
[ADD] crm_todo: Nothing to do

pedrobaeza:
It's curious that the analysis file mentions the field lead_id as removed, where the field remains unaltered. I have removed this line.


pedrobaeza:
Please see #232 for the update in the documentation.

StefanRijnhart:
I was thinking: leave the analysis file intact, and mark the line in the work file with 'false positive'.

Can't really explain this either. There was no entry in openupgrade_records for this field in an old reference database that I created for the generation of the analysis files, but if I recreate the reference database the record is there. I will try to recreate the analysis and I expect that the false positive will be gone then.

pedrobaeza:
Done.

About analysis, maybe it was a bad version that now  it doesn't happen again.

StefanRijnhart:
Tnx! :+1:

PR #234, opened 2015-02-11
[IMP/FIX] m2o_to_m2m: Use SQL for performance and for avoiding side effects

pedrobaeza:
When you make an ORM write, all the corresponding triggers are launched (compute stores, constraints, resolve references fields...), causing sometimes that the migration cannot continue to an error in one of this.

This is not needed for any migration script, so using SQL is faster and safer for this operation.

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -430,14 +430,27 @@ def m2o_to_m2m(cr, model, table, field, source_field):
 
     .. versionadded:: 7.0
     """
+    if not model._columns.get(field):
+        raise osv.except_osv(
+            "Error", "m2o_to_m2m: field %s doesn't exist in model %s" % (
+                field, model._name))
+    if not isinstance(model._columns[field], osv.fields.many2many):
+        raise osv.except_osv(
+            "Error", "m2o_to_m2m: field %s of model %s is not a many2many "
+                     "one" % (field, model._name))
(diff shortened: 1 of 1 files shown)


hbrunn:
:+1:

pedrobaeza:
Can anyone more review this?

legalsylvain:
:+1: Thanks.
Could you port this improvement in V8 ? 
Regards;

pedrobaeza:
Port to 8.0 made on #256

openerp/openupgrade/openupgrade.py
```diff
@@ -430,14 +430,27 @@ def m2o_to_m2m(cr, model, table, field, source_field):
...
                    'field': source_field,
                    })
+    rel, id1, id2 = osv.fields.many2many._sql_names(model._columns[field],
+                                                    model)
     for row in cr.fetchall():
-        model.write(cr, SUPERUSER_ID, row[0], {field: [(4, row[1])]})
+        cr.execute(
+            "INSERT INTO %s (%s, %s) "
+            "VALUES (%s, %s)" %
+            (rel, id1, id2, row[0], row[1]))
```
hbrunn, line 453: also here, values should be pg parameters, and only table/column names string formatted

pedrobaeza, line 453: Here you have: https://github.com/OpenUpgrade/OpenUpgrade/pull/258

PR #236, opened 2015-02-11
[FIX/IMP] API method set_partner_id_from_partner_address_id switched to SQL

pedrobaeza:
As in #234, using ORM write makes this call very slow and also error prone due to the checks that are performed on the write.

This switches the write to SQL to avoid this.

### DIFF
--- a/openerp/openupgrade/openupgrade_70.py
+++ b/openerp/openupgrade/openupgrade_70.py
@@ -23,8 +23,6 @@
 # This module provides simple tools for openupgrade migration, specific for the
 # 6.1 -> 7.0.
 
-from openerp import SUPERUSER_ID
-
 
 def set_partner_id_from_partner_address_id(
         cr, pool, model_name, partner_field, address_field, table=None):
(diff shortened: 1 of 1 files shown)


openerp/openupgrade/openupgrade_70.py
```diff
@@ -48,7 +46,10 @@ def set_partner_id_from_partner_address_id(
              res_partner_address as address
         WHERE address.id = target.%s""" % (table, address_field))
     for row in cr.fetchall():
-        model.write(cr, SUPERUSER_ID, row[0], {partner_field: row[1]})
+        cr.execute("UPDATE %s "
+                   "SET %s=%s "
+                   "WHERE id=%s" %
+                   (table, partner_field, row[1], row[0]))
```
hbrunn: please only put table and partner_field in the string formatting part, and row[0,1] as sql parameter

pedrobaeza: Well, it's going to be an akward string with +. As these are integer parameters, they don't need to be specially parsed. Are there injection risk?

hbrunn: I thought about
`cr.execute("UPDATE %s SET %s=%%s WHERE id=%%s" % (table, partner_field), (row[1], row[0]))`

I don't see any injection risks, but who am I to tell?

[EDIT] corrected param tuple

pedrobaeza:
@hbrunn, changed as you suggested.

hbrunn:
thanks, :+1: 
travis error unrelated

pedrobaeza:
@hbrunn, according https://github.com/OpenUpgrade-dev/OpenUpgrade/pull/2, is this going to work?

pedrobaeza:
I see from @StefanRijnhart's comments where is the problem with the other sentence, so this is right.

Anyone more can review?

StefanRijnhart:
:+1:, thanks.

pedrobaeza:
Port to 8.0 made in #255

PR #237, opened 2015-02-11
[FIX] project: act_window 'action_view_task' context not changed

pedrobaeza:
Comment in XML is self-explanatory

PR #238, opened 2015-02-12
[IMP] Detect removed tags in update data

StefanRijnhart:
Detect problems like #237

### DIFF
--- a/scripts/compare_noupdate_xml_records.py
+++ b/scripts/compare_noupdate_xml_records.py
@@ -124,10 +124,19 @@ def get_records(addon_dir):
     return records_update, records_noupdate
 
 
+def check_blacklist(record, key):
+    """ Known deleted fields. Does not scale. """
+    if (key == "./field[@name='type']" and
+            record.get('model') == 'ir.ui.view'):
+        return False
+    return True
+
+
 def main(argv=None):
     """
     Attempt to represent the differences in data records flagged with
     'noupdate' between to different versions of the same OpenERP module.
+    Also detects removed tags from 'update' records.
 
     Print out a complete XML data file that can be loaded in a post-migration
     script using openupgrade::load_xml().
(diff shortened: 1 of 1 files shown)


scripts/compare_noupdate_xml_records.py
```diff
@@ -185,20 +196,29 @@ def main(argv=None):
...
+            for key in record_new_dict.keys():
+                if not record_old.xpath(key):
+                    element.append(deepcopy(record_new_dict[key]))
 
         if len(element):
             data.append(element)
 
+    data.append(etree.Comment("Changes in noupdate data"))
+    for xml_id, record_new in new_noupdate.items():
+        compare_record(xml_id, record_new)
```
pedrobaeza, line 216: I don't get it. Here you are only reassigning old_record, but not doing anything with it.

StefanRijnhart, line 216: I'm not assigning old_record, but new_record and I'm passing this to an inline method that adds to the list of elements which is scoped under the main method.

pedrobaeza, line 216: But in `compare_record` method record_new is not handled...

StefanRijnhart, line 216: Are we talking about the same compare_record? In the inline function that I am talking about, record_new is a function argument which is referred to several times.

pedrobaeza, line 216: Sorry, I was confused about the diff and the indentation, which is very difficult to follow. I have seen full code and now it's clear.

pedrobaeza:
:+1:

StefanRijnhart:
I can imagine! I made minimal changes for this fix but I should probably refactor it next. Thanks for the review.

hbrunn:
:+1:

PR #239, opened 2015-02-13
[ADD] rename audittrail module

hbrunn:


pedrobaeza:
As the module is not based in the same structure, is there any migration script planned for the module itself?

About this, :+1:

hbrunn:
That happens in auditlog's init_hook - there I check if the audittrail tables exist and do some on the fly migration

pedrobaeza:
Cool! These kind of things is something that in the past hasn't been taken into account, and it's also important to assure the migration path.

Thanks.

hbrunn:
I retract this one, because the init part is not called when updating

pedrobaeza:
Why not a regular migration script that call the same method?

hbrunn:
Then I'd have to change the version number of the new auditlog module to something bigger than 1.0. Possible of course, but I'd like to leve the current PR in peace so that it can be merged without further complications.

pedrobaeza:
Why not simply 8.0.1.0? Migration script will be launched from 7.0.1.0.

hbrunn:
thanks Predro, after years of developing migration scripts I should know that. Will do as you propose

pedrobaeza:
This can be already be merged as it is safe enough and the scope is only limited to those who have audittrail module installed.

PR #244, opened 2015-02-20
[ADD] clear base.act_ir_actions_todo_form's domain

hbrunn:
cf https://github.com/odoo/odoo/blob/5.0/bin/addons/base/ir/ir.xml#L1364, https://github.com/odoo/odoo/blob/6.0/bin/addons/base/ir/ir.xml#L1919


pedrobaeza:
Thanks for the clear reference.

:+1:

StefanRijnhart:
:+1:

PR #248, opened 2015-03-04
Fix move field m2o binary

legalsylvain:
Hi, 
I try to fix a minor bug in move_field_m2o function of openupgrade framework that occures if we try to migrate binary_field with undefined data. (In my case, move of image field from product.product to product.template).

Kind regards.

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -865,12 +865,7 @@ def move_field_m2o(
                 field_old_model, table_old_model, m2o_field_old_model, ko_id))
         cr.execute(query)
         if binary_field:
-            vals = []
-            for x in cr.fetchall():
-                if x[0]:
-                    vals.append(x[0][:])
-                else:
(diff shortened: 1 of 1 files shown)


openerp/openupgrade/openupgrade.py
```diff
@@ -861,7 +865,12 @@ def default_func(cr, pool, id, vals):
                 field_old_model, table_old_model, m2o_field_old_model, ko_id))
         cr.execute(query)
         if binary_field:
-            vals = [str(x[0][:]) for x in cr.fetchall()]
```
pedrobaeza, line 864: Better to put:

```
vals = [str(x[0][:]) if x[0] else False for x in cr.fetchall()]
```

legalsylvain, line 864: Thanks. Done.

pedrobaeza:
:+1:

StefanRijnhart:
Thanks! :+1:

PR #249, opened 2015-03-04
[IMP] product_template.standard_price. create global properties; add logger;

legalsylvain:
Hi all, 

This PR is WIP but I would like to have your point of View. Using OpenUpgrade on a medium sized database, I find that  migrate product module take a while. My configuration is about 15 companies and 10.000 products.
The actual process create one property by product and by company. See post-migration.py script function 'create_properties'. So about 150.000 properties in my case ! My PoV : 
- It's unnecessary ; 
- It is database pollution ; 
- It take a big while ; 

As most of my products have a company_id defined (as some other implementation I guess), I don't understand the interest to create a property for a product of company A in company B, but maybe I missed something.

So we have three possibilities for each product : 
- Solution A/ create a property only for a company
- Solution B/ Create a global property (without company defined) ; 
- Solution C/ create a property for each company ; (the current behaviour

I propose to realize a PR with the following behaviour :
**Case 1 : product have a company A defined**
-> solution A;

**Case 2 : product doesn't have a company defined**
-> solution B;

How Odoo 8 works : 
- Create a company B in an instance and add company B right access to demo user ; 
- With demo user, Create a product in company B with company B defined and set a standard_price --> the property is only created for company B ; 
- With demo user, Create a product in company B WITHOUT company defined and set a standard_price --> the property is only created for company B ; (weird behaviour, IMHO) ; 
- With demo user, Create a product in company B WITHOUT company defined and set a standard_price ; change the created property and remove company B on the property (make it global) ; change after the property again in the product view. There is now two properties : 
  --> there is a global property that affect all other company except the company B ;
  --> there is a property for the company B ;

Waiting for your PoV. 

Kind regards.

### DIFF
--- a/addons/product/migrations/8.0.1.1/post-migration.py
+++ b/addons/product/migrations/8.0.1.1/post-migration.py
@@ -61,20 +61,23 @@ def create_properties(cr, pool):
     Write using the ORM so the prices will be written as properties.
     """
     template_obj = pool['product.template']
-    company_obj = pool['res.company']
-    company_ids = company_obj.search(cr, SUPERUSER_ID, [])
     sql = ("SELECT id, %s FROM product_template" %
            openupgrade.get_legacy_name('standard_price'))
     cr.execute(sql)
     logger.info(
         "Creating product_template.standard_price properties"
-        "for %d products." % (cr.rowcount))
+        " for %d products." % (cr.rowcount))
     for template_id, std_price in cr.fetchall():
-        for company_id in company_ids:
-            ctx = {'force_company': company_id}
-            template_obj.write(cr, SUPERUSER_ID, [template_id],
-                               {'standard_price': std_price},
-                               context=ctx)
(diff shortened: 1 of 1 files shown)


legalsylvain:
@StefanRijnhart @hbrunn @pedrobaeza, @all any point of view about that point ?
I really think that creating x properties, x = company \* product, is not a good approch.
Regards.

hbrunn:
@legalsylvain thanks for bringing this up. I think the proper thing to do is to create one global property per product in all cases. Why? That mimics 7.0's behavior, and given that we migrate from a 7.0 database, we should keep things the same if there's no need to do otherwise

legalsylvain:
@hbrunn, thanks ! it makes sense.
So : Currently implemented in my PR : solution B in all case.
Regards.

pedrobaeza:
:+1:

StefanRijnhart:
Thanks! :+1:

PR #250, opened 2015-03-06
[ADD] Function to fallback on workflow state

bwrsandman:
Replace state which could have been remove from all instances with a fallback state

This could happen in the case where a workflow is simplified and one state is removed.

This will get these items out of that removed state in the pre-upgrade and put them in a defined fallback.

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -312,12 +312,17 @@ def update_workflow_workitems(cr, pool, ref_spec_actions):
     """Find all the workflow items from the target state to set them to
     the wanted state.
 
-    Remove a workflow action and move those to fallback state
-    (use in pre-migration)
+    When a workflow action is removed, from model, the objects whose states
+    are in these actions need to be set to another to be able to continue the
+    workflow properly.
+
+    Run in pre-migration
 
     :param ref_spec_actions: list of tuples with couple of workflow.action's
-        external ids. The first id is replaced to the second.
+        external ids. The first id is replaced with the second.
     :return: None
+
+    .. versionadded:: 7.0
     """
     workflow_workitems = pool['workflow.workitem']
     ir_model_data_model = pool['ir.model.data']
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -324,10 +324,14 @@ def update_workflow_workitems(cr, pool, ref_spec_actions):
 
     for (target_external_id, wanted_external_id) in ref_spec_actions:
         target_activity_id = ir_model_data_model.get_object(
-            cr, SUPERUSER_ID, target_external_id
+            cr, SUPERUSER_ID,
+            target_external_id.split(".")[0],
+            target_external_id.split(".")[1],
         ).id
         wanted_activity_id = ir_model_data_model.get_object(
-            cr, SUPERUSER_ID, wanted_external_id
+            cr, SUPERUSER_ID,
+            wanted_external_id.split(".")[0],
+            wanted_external_id.split(".")[1],
         ).id
 
         ids = workflow_workitems.search(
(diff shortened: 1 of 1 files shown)


legalsylvain:
Great feature ! 
Thanks.
LGTM :+1: 

2 things: 
- Could you add .. versionadded:: 8.0 in the comment ?
- Could you add a logger to indicate to the openupgrade user the change of state ? Something like : 
  XXX state changes for object.name (from target_activity.name to dest_activity.name)
  or something like that ...

bwrsandman:
@legalsylvain this works on 7.0 as well (not tested on 6.1), should the versionadded be 7.0?

legalsylvain:
Yes, I guess.

openerp/openupgrade/openupgrade.py
```diff
@@ -307,6 +308,36 @@ def drop_columns(cr, column_spec):
...
 
+def update_workflow_workitems(cr, pool, ref_spec_actions):
+    """Find all the workflow items from the target state to set them to
+    the wanted state.
+
+    Remove a workflow action and move those to fallback state
+    (use in pre-migration)
+
+    :param ref_spec_actions: list of tuples with couple of workflow.action's
+        external ids. The first id is replaced to the second.
```
pedrobaeza: s/replaced to/replaced with/

openerp/openupgrade/openupgrade.py
```diff
@@ -307,6 +308,36 @@ def drop_columns(cr, column_spec):
                         table, column)
 
 
+def update_workflow_workitems(cr, pool, ref_spec_actions):
+    """Find all the workflow items from the target state to set them to
+    the wanted state.
+
+    Remove a workflow action and move those to fallback state
```
pedrobaeza: Can you explain a bit more this use case?

pedrobaeza:
If this function is not used in any migration script in v7, it's not needed to include it on that version.

openerp/openupgrade/openupgrade.py
```diff
@@ -307,6 +308,36 @@ def drop_columns(cr, column_spec):
...
+    """
+    workflow_workitems = pool['workflow.workitem']
+    ir_model_data_model = pool['ir.model.data']
+
+    for (target_external_id, wanted_external_id) in ref_spec_actions:
+        target_activity_id = ir_model_data_model.get_object(
+            cr, SUPERUSER_ID, target_external_id
+        ).id
+        wanted_activity_id = ir_model_data_model.get_object(
+            cr, SUPERUSER_ID, wanted_external_id
```
bwrsandman: needs a split on the "." here

bwrsandman:
> If this function is not used in any migration script in v7, it's not needed to include it on that version.

It's not being used in 8.0 either.

OpenUpgrade is a collection of functions which help migration. They can be used in other cases than just Odoo and OCA modules.

I am using this function in a private module for a script from 7.0.A.X to 7.0.B.Y, I put it here because it looks to be useful.

If you feel like it doesn't belong, you can close the PR. This is just a contribution of what I think is very useful.

bwrsandman:
Added logger which looks like

```
openupgrade: Moving 1 items in the removed workflow action (sent) to a fallback action (draft).
```

Is this clear enough? I can add more info, but I want to avoid doing too many reads.

legalsylvain:
Hi, Thanks ! I thought add [ids] list. To give the possibility to admin user to check after the migration, the behaviour of a given object and to check if the workflow is correct and if nothing is broken.
What do you think ? (not a blocking point for me, just a nice to have feature).

bwrsandman:
Of course!

hbrunn:
:+1:

StefanRijnhart:
Thanks! It will have to be .. versionadded:: 7.0 though, because it is not included in 7.0 yet. This functions as a hint in the API doc to developers looking to use these methods, and they have to know in which version which method became available.

StefanRijnhart:
Sorry, missed #252. :+1: then.

PR #251, opened 2015-03-09
[ADD] set task's stage from state if not set

hbrunn:
This becomes more important for 8.0 where state on project.task is dropped

### DIFF
--- a/addons/project/migrations/7.0.1.1/post-migration.py
+++ b/addons/project/migrations/7.0.1.1/post-migration.py
@@ -65,6 +65,27 @@ def createProjectAliases(cr, pool):
             """, (alias_id, id))
 
 
+def setStageFromState(cr):
+    cr.execute(
+        '''with task_types as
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

addons/project/migrations/7.0.1.1/post-migration.py
```diff
@@ -65,6 +65,27 @@ def createProjectAliases(cr, pool):
             """, (alias_id, id))
 
 
+def setStageFromState(cr):
```
StefanRijnhart: From pep8: Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability.
That would be set_stage_from_state()

StefanRijnhart:
Otherwise fine!

hbrunn:
@StefanRijnhart done

StefanRijnhart:
:+1:, thanks for the change.

PR #252, opened 2015-03-09
[7.0] Update workflow workitems

bwrsandman:
Same as #250

Replace state which could have been remove from all instances with a fallback state

This could happen in the case where a workflow is simplified and one state is removed.

This will get these items out of that removed state in the pre-upgrade and put them in a defined fallback.


hbrunn:
:+1:

StefanRijnhart:
Same as #252 :+1:

PR #255, opened 2015-03-12
[FIX/IMP] API method set_partner_id_from_partner_address_id switched to SQL

pedrobaeza:
Same as #236

### DIFF
--- a/openerp/openupgrade/openupgrade_70.py
+++ b/openerp/openupgrade/openupgrade_70.py
@@ -23,8 +23,7 @@
 # This module provides simple tools for openupgrade migration, specific for
 # the 6.1 -> 7.0 migration. It is kept in later editions to keep all the API
 # docs in the latest release.
-
-from openerp import SUPERUSER_ID
+import openupgrade
 
 
 def set_partner_id_from_partner_address_id(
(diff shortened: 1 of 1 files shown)


legalsylvain:
Thanks.

hbrunn:
:+1:

StefanRijnhart:
Please make it a single query. I think you can use UPDATE...FROM.

pedrobaeza:
I haven't touched that part for this PR.

StefanRijnhart:
No, but it makes no sense to loop over the results of an sql join to do an sql update. You should make it a single query.

pedrobaeza:
@StefanRijnhart , I have joined both queries. Please review it.

pedrobaeza:
Used the same SQL succesfully on 7.0 (where it applies).

pedrobaeza:
Can anyone review this?

openerp/openupgrade/openupgrade_70.py
```diff
@@ -42,14 +41,18 @@ def set_partner_id_from_partner_address_id(
...
+        "UPDATE %s "
+        "SET %s=src.partner_id "
+        "FROM ("
+        "    SELECT target.id as id, "
+        "           address.openupgrade_7_migrated_to_partner_id as partner_id"
+        "    FROM %s as target, res_partner_address as address "
+        "    WHERE address.id = target.%s"
+        ") src "
+        "WHERE id=src.id" %
+        (table, partner_field, table, address_field))
```
StefanRijnhart: You can eliminate the subquery using

```
UPDATE %{table}s
SET %{partner_field}s=address.openupgrade_7_migrated_to_partner_id
FROM res_partner_address address
WHERE %{table}s.%{address_field}=address.id
```

pedrobaeza:
@StefanRijnhart, thanks for the tip. UPDATE syntax is very confusing for me.

When you validate this, I'll apply it to 7.0 also.

StefanRijnhart:
:+1:, thanks!

PR #257, opened 2015-03-12
[FIX] don't try to fix nonexisting partner_id

hbrunn:
We move crm_meeting.partner_id out of the way in https://github.com/OpenUpgrade/OpenUpgrade/blob/7.0/addons/crm/migrations/7.0.1.0/pre-migration.py#L44. This problem was revealed by #236, before the ORM happily ignored the nonexisting field.


pedrobaeza:
Yeah, I always commented this line without going deeper in the problem. Thanks for investigating it.

:+1:

StefanRijnhart:
I see. Partner_id was set aside to be migrated to m2m partner_ids in base_calendar (which has not been done yet). :+1:

PR #258, opened 2015-03-13
[IMP] m2o_to_m2m: Join SQL queries in one

pedrobaeza:
Improvement according https://github.com/OpenUpgrade/OpenUpgrade/pull/234/files#r26291892

PR #261, opened 2015-03-20
[IMP] m2m_to_x2m: Make operations with SQL

pedrobaeza:


StefanRijnhart:
Same as #258, so :+1:

openerp/openupgrade/openupgrade.py
```diff
@@ -590,19 +590,14 @@ def m2o_to_x2m(cr, model, table, field, source_field):
...
-                   'field': source_field,
-                   })
     rel, id1, id2 = orm.fields.many2many._sql_names(model._columns[field],
                                                     model)
-    for row in cr.fetchall():
-        cr.execute(
-            "INSERT INTO %s (%s, %s) "
-            "VALUES (%%s, %%s)" %
-            (rel, id1, id2), (row[0], row[1]))
+    cr.execute(
```
bwrsandman: This might be a good place to use `logged_query`. What do you think?

pedrobaeza:
@bwrsandman, done also

pedrobaeza:
I have restarted the branch with correct code from the revert I have made on https://github.com/OpenUpgrade/OpenUpgrade/pull/263.

@StefanRijnhart, @hbrunn, @legalsylvain, @bwrsandman  I have a doubt with this method: when converting from many2one to one2many, with old method that this PR changes and with current mine, you can lose data in case of multiple references to the same row. What do you think we have to do?
- Warn as I did that this can be possible.
- Limit to the use case when source many2one is unique.
- Make a verbatim copy of each record that has duplicated references.

legalsylvain:
It seems that is the same problem that for the function 'move_field_m2o'. (field from product move to template). 
In that function, we choose to select the value the most present.

> - Warn as I did that this can be possible.

Could be great;

> - Limit to the use case when source many2one is unique.

I'm not in favor of this solution, because it will block the migration.

pedrobaeza:
Can anyone review this?

openerp/openupgrade/openupgrade.py
```diff
@@ -571,28 +573,85 @@ def get_legacy_name(original_name):
...
+    in your post-migrate script.
+
+    :param model: The target model registry object
+    :param table: The source table
+    :param field: The field name of the target model
+    :param source_field: the many2one column on the source table.
+
+    .. versionadded:: 7.0
+    """
+    return m2o_to_x2m(cr, model, table, field, source_field)
```
bwrsandman, line 656: Could you explain why this function was turned into a def instead of refering to the new one? I don't see anything in your commit message that makes this change seem relevant.

pedrobaeza, line 656: This is only for convenience in the documentation, to keep versionadded correctly, and to explain explicitely the use of this function for this kind of method.

StefanRijnhart, line 656: Seems alright to me. Anyone else thinking that we should take the opportunity to realign the argument order of the new method from 

_target_ model, _source_ table, _target_ field, _source_ field

to 

 _source_ table, _source_ field, _target_ model, _target_ field

pedrobaeza, line 656: Well, it can be, but we will loose compatibility with migration scripts external to this project.

StefanRijnhart, line 656: Oh I see now. I had mistakenly thought that this adds the x2m method but it already existed.

StefanRijnhart:
:+1:

openerp/openupgrade/openupgrade.py
bwrsandman, line 656: Cool. If that's the case it would be very useful to add the [deprecated option](http://sphinx-doc.org/markup/para.html#directive-deprecated)

pedrobaeza, line 656: @bwrsandman, done

openerp/openupgrade/openupgrade.py
```diff
@@ -571,28 +573,86 @@ def get_legacy_name(original_name):
...
+    script to retain the column's old value, then call m2o_to_m2m
+    in your post-migrate script.
+
+    :param model: The target model registry object
+    :param table: The source table
+    :param field: The field name of the target model
+    :param source_field: the many2one column on the source table.
+
+    .. versionadded:: 7.0
+    .. deprecated:: 8.0
```
StefanRijnhart, line 653: Thanks for adding this! Can you refer to the replacing method to use instead?

```
.. deprecated:: 8.0
   Use :func:`m2o_to_x2m` instead.
```

(you should try out GitHub online editing for this, you'll be done in 5 secs)

pedrobaeza, line 653: Done!

StefanRijnhart, line 653: Thanks! @bwrsandman is it good for you too now?

bwrsandman, line 653: :+1:

bwrsandman:
:+1: Thanks @pedrobaeza !

PR #262, opened 2015-03-21
[FIX] account: Check for partner existence on method migrate_payment_term

pedrobaeza:
There can be properties that belong to non-existant partners, so this must be taken into account.


pedrobaeza:
Please review this.

StefanRijnhart:
:+1:

pedrobaeza:
I think this can be merged as it's safe

PR #267, opened 2015-03-25
[ADD] nothing to do for account_analytic_analysis

hbrunn:


pedrobaeza:
We should cover the migration from module _account_analytic_analysis_recurring_ (https://github.com/OCA/project-service/tree/7.0/account_analytic_analysis_recurring), at least putting it as uninstalled it fields are the same. Don't you think?

hbrunn:
I didn't know of this module. Then probably someone knowing it should take over from here...

pedrobaeza:
It's a backport of this future, so it shouldn't mean any data migration, but put as uninstalled.

legalsylvain:
:+1: for this PR.

legalsylvain:
@pedrobaeza, I guess that it should be handled by account_analytic_analysis_recurring migration. Don't you think ? (I mean, people using this module in V7 should write a migration script for V8) don't you ?

pedrobaeza:
I don't think so. It's just like _base_contact_ migration handled by base migration on v7. You can then remove the ir_module_module record, and make any possible adjustment (minimal).

StefanRijnhart:
Thanks for the suggestion, @pedrobaeza. Although out of scope for this project at a bare minimum, If it is a proper backport, there should be very little to do. Just glancing over the data model, this seems indeed to be the case. Just changing the module for the "account_analytic_cron_for_invoice" XML ID should suffice. @hbrunn, care to include this here?

legalsylvain:
@hbrunn : what is your point of view, regarding @StefanRijnhart and @pedrobaeza remarks ?

hbrunn:
@legalsylvain I think @pedrobaeza and @StefanRijnhart are right. I added a commit to take care of this point.

Note that also in this case, #303 is crucial in order not to lose data

legalsylvain:
:+1:

StefanRijnhart:
Thanks! :+1:

PR #273, opened 2015-04-04
[FIX] Analysis wizard: Rename field "write" to "write_files"

alhashash:
To prevent error when calling the orm "write()" method on the wizard model as it calls the "write" field descriptor instead which raises `AttributeError: 'openupgrade.analysis.wizard' object has no attribute '_ids'`

### DIFF
--- a/openerp/addons/openupgrade_records/model/analysis_wizard.py
+++ b/openerp/addons/openupgrade_records/model/analysis_wizard.py
@@ -55,14 +55,14 @@ class openupgrade_analysis_wizard(TransientModel):
             [('init', 'Init'), ('ready', 'Ready')], 'State',
             readonly=True),
         'log': fields.text('Log'),
-        'write': fields.boolean(
+        'write_files': fields.boolean(
             'Write files',
             help='Write analysis files to the module directories'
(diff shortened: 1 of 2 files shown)


openerp/addons/openupgrade_records/model/analysis_wizard.py
```diff
@@ -55,14 +55,14 @@ class openupgrade_analysis_wizard(TransientModel):
...
-        'write': fields.boolean(
+        'write_files': fields.boolean(
             'Write files',
             help='Write analysis files to the module directories'
             ),
         }
     _defaults = {
         'state': lambda *a: 'init',
-        'write': lambda *a: True,
+        'write_files': lambda *a: True,
```
pedrobaeza: You can put only `'write_files':True,`

alhashash: Sure, for both fields, but I just made the minimum amount of change to fix the issue.

pedrobaeza:
:+1: 

Thanks.

legalsylvain:
:+1: Thanks

PR #277, opened 2015-04-09
[FIX] stock: Closes #276: deferred method stock installed detection

### LINKED ISSUE #276: Programming error: relation "stock_move" does not exist
in the the defered80.py the following update will always run

https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/openerp/openupgrade/deferred_80.py#L192

because the if condition check if there is at least one record returned
https://github.com/OpenUpgrade/OpenUpgrade/blob/8.0/openerp/openupgrade/deferred_80.py#L190

and hence there will always be one because the query on which the result is tested make a count so it will at least return one record with 0 even if stock is not installed

cr.execute(
        "SELECT count(*) FROM ir_module_module WHERE name='stock' "
        "AND state='installed'")

a possible fix should be to remove the count or to change the if condition

pedrobaeza:


openerp/openupgrade/deferred_80.py
```diff
@@ -99,7 +99,7 @@ def migrate_procurement_order_method(cr, pool):
         WHERE name = 'stock' AND state = 'installed'
         AND latest_version = '8.0.1.1'
         """)
-    if not cr.fetchone():
```
houssine78, line 102: this wasn't giving any error as the query return the id

openerp/openupgrade/deferred_80.py
```diff
@@ -187,7 +187,7 @@ def migrate_stock_move_warehouse(cr):
     cr.execute(
         "SELECT count(*) FROM ir_module_module WHERE name='stock' "
         "AND state='installed'")
-    if not cr.fetchone():  # No stock
+    if not cr.fetchone()[0]:  # No stock
```
houssine78: this solve the error i've already tested it"SELECT \* FROM ir_module_module WHERE name='stock' "
         "AND state='installed'")

pedrobaeza:
@houssine78, now it's fixed.

houssine78:
thanks @pedrobaeza  !!

legalsylvain:
:+1:

PR #280, opened 2015-04-15
8.0 stock improve fields creation

legalsylvain:
I propose some script to make faster stock migration.
For some fields of stock_picking or stock_move tables, the migration will take a while, using default ORM method and a trivial SQL request can make the fields creation faster.
Sample: (same DB / same server) : 
- Compute of stock_picking.priority : 20 min -> 10 sec;
- compute of stock_move.product_qty : ~2h -> 5 min;

Thanks for your review. CC @pedrobaeza.

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -75,7 +75,7 @@ def migrate_move_inventory(cr, registry):
         FROM {stock_inventory_move_rel} rel
         WHERE sm.id = rel.move_id
         """.format(stock_inventory_move_rel=openupgrade.get_legacy_name(
-        'stock_inventory_move_rel')))
+            'stock_inventory_move_rel')))
 
     openupgrade.logged_query(
         cr,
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I have compared the function methods in https://github.com/OCA/OCB/blob/8.0/addons/stock/stock.py#L679 and https://github.com/OCA/OCB/blob/8.0/addons/stock/stock.py#L1616 and  with correspondant SQL's and seems to be correct.

Very smart method to speed up!

Just a PEP8 remark:

```
./addons/stock/migrations/8.0.1.1/post-migration.py:78:9: E122 continuation line missing indentation or outdented
```

legalsylvain:
Hum, quite annoying, when I ran flake8 ./ : 
- I had an error "E126 continuation line over-indented for hanging indent" when Travis was Green,
- I had no error when travis said "E122 continuation line missing indentation or outdented";
  A little sudo pip install flake8 --upgrade fixed my bug.

@pedrobaeza thanks for your review !
Other point : I try to reduce quant creation, but it doesn't seems possible by SQL. Any help welcome.

paulius-sladkevicius:
:+1: will do migration on Monday, I'll give feedback if any issues raise.

pedrobaeza:
Any problem with this improvement?

legalsylvain:
@komsas : any news after your tests ?

paulius-sladkevicius:
@legalsylvain didn't have any issues, people are using db and no bad behavior noticed. :+1:

pedrobaeza:
Merging then!

Thanks.

PR #281, opened 2015-04-21
[FIX] don't assign address_home_id to users

hbrunn:
[ADD] amend newly created user partner from employee if it exists


hbrunn:
The rationale for not assigning address_home_id is that you don't want the user being connect to an employee's home address (that would mean getting mails to your home email etc.)

pedrobaeza:
:+1:

StefanRijnhart:
:+1:

PR #284, opened 2015-04-29
[FIX] use uppercase where necessary

hbrunn:
This fixes incorrect casing introduced by #261


pedrobaeza:
Sorry for that, I though it was fixed in the branch (I pushed something, but maybe it was to the wrong target).

This should be merged inmediately to avoid the error, so I fast-track.

PR #286, opened 2015-05-04
[IMP] when using an employee's work address, take

hbrunn:
care that this partner has the correct timezone and language


pedrobaeza:
:+1:

StefanRijnhart:
:+1:

PR #287, opened 2015-05-05
[FIX] use pickings to determine sale orders for

hbrunn:
procurement groups to create. Not all so's trigger a procurement, but
all of them trigger a picking.

### DIFF
--- a/addons/sale_stock/migrations/8.0.1.0/post-migration.py
+++ b/addons/sale_stock/migrations/8.0.1.0/post-migration.py
@@ -49,26 +49,23 @@ def set_procurement_groups(cr):
     the only way that sale orders and pickings are linked.
     """
 
-    # Create a procurement group for every sale order with at least one
-    # procurement, and link the procurement group to the sale order.
+    # Create a procurement group for every sale order pointed to by a picking,
+    # and link the procurement group to the sale order.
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Seems OK :+1:

addons/sale_stock/migrations/8.0.1.0/post-migration.py
```diff
@@ -49,26 +49,23 @@ def set_procurement_groups(cr):
...
+        SELECT DISTINCT so.id
             FROM sale_order so
-            WHERE (
-                SELECT COUNT(*) FROM sale_order_line sol
-                WHERE order_id = so.id
-                AND {procurement_id} IS NOT NULL
-                ) > 0
+            JOIN stock_picking sp
+            ON
+            so.id=sp.{sale_id}
```
StefanRijnhart: How about

SELECT DISTINCT sale_id FROM stock_picking WHERE sale_id IS NOT NULL

hbrunn:
@StefanRijnhart yes, why not? Thanks!

StefanRijnhart:
:+1:

pedrobaeza:
Better solution indeed. I merge to make a migration just now...

PR #290, opened 2015-05-08
[FIX] don't pass the same id multiple times to write

hbrunn:
this can happen if some of the fields are equal


pedrobaeza:
:+1:

StefanRijnhart:
Thanks! :+1:

PR #294, opened 2015-05-10
[FIX] mail: Incorrect mail_message post migration UPDATE statement

alhashash:
The mail_message table is wrongly included in the FROM clause. This led to a huge cross join and even if it is successfully completed, the data would not be correct.

In a database with about 280k mail_message rows and 970 mail_mail rows, the query always consumes all the memory (6GB) until the OOM killer takes down the PG backend. I'm wondering how this error was never caught before?!!


paulius-sladkevicius:
:+1:

pedrobaeza:
This was previously mentioned on https://github.com/OpenUpgrade/OpenUpgrade/issues/270, but it wasn't proposed any PR to solve it, so thank you very much for doing it.

:+1:

StefanRijnhart:
Thanks! :+1:

PR #295, opened 2015-05-11
[FIX] Migrate more than a single record

StefanRijnhart:


StefanRijnhart:
We noticed our sale_line_id fields on procurement_order not being populated properly anymore. 

```
odoo80_tst=> UPDATE procurement_order t                    
            SET sale_line_id=src.id
            FROM (
                  SELECT s.id, s.openupgrade_legacy_8_0_procurement_id
                  FROM procurement_order as target, sale_order_line as s
                  WHERE s.openupgrade_legacy_8_0_procurement_id=target.id
                  LIMIT 1
            ) src                                               
            WHERE t.id=src.openupgrade_legacy_8_0_procurement_id
;
UPDATE 1
odoo80_tst=> UPDATE procurement_order t                    
            SET sale_line_id=src.id
            FROM (
                  SELECT s.id, s.openupgrade_legacy_8_0_procurement_id
                  FROM procurement_order as target, sale_order_line as s
                  WHERE s.openupgrade_legacy_8_0_procurement_id=target.id
                              ) src
            WHERE t.id=src.openupgrade_legacy_8_0_procurement_id
;
UPDATE 4543

```

pedrobaeza:
I explain here (https://github.com/OpenUpgrade/OpenUpgrade/pull/261#issuecomment-84632802) why I limit that, and I think this patch is updating incorrectly, but correct me if I'm wrong.

StefanRijnhart:
I don't understand, @pedrobaeza. Without the fix, only one of the 4543 procurements that was linked to a sale order line in the database was still linked to a sale order line after the migration. It really messes things up ;-)

pedrobaeza:
The original problem was that more than one value for each record to update can be found when dealing with one2many, and I got that error in my tests, but you're right that my original SQL is actually limiting the number of updated records.

Don't you face with the other error I comment?

StefanRijnhart:
This only leads to data loss (as does using LIMIT in the right place, if there is one), not an error:

```
odoo80_tst2=> update sale_order_line set openupgrade_legacy_8_0_procurement_id = 4756;
UPDATE 13312
odoo80_tst2=> update procurement_order set sale_line_id = NULL;
UPDATE 5217
odoo80_tst2=>             UPDATE procurement_order t
            SET sale_line_id=src.id
            FROM (
                  SELECT s.id, s.openupgrade_legacy_8_0_procurement_id
                  FROM procurement_order as target, sale_order_line as s
                  WHERE s.openupgrade_legacy_8_0_procurement_id=target.id
                              ) src
            WHERE t.id=src.openupgrade_legacy_8_0_procurement_id
;
UPDATE 1
odoo80_tst2=> select id,sale_line_id from procurement_order where sale_line_id is not null;
  id  | sale_line_id 
------+--------------
 4756 |          959
(1 row)
```

StefanRijnhart:
I've eliminated the subselect now as well.

pedrobaeza:
OK, I think it's correct now. Thanks for the investigation.

:+1:

hbrunn:
:+1:

pedrobaeza:
OK, I merge and I'll test it in a migration I have to make.

PR #297, opened 2015-05-14
[ADD] Migration of stock_account

StefanRijnhart:
Some of which some was already present, but not called, and some was migrated in stock already.
[RFR] Don't rename procurement order's move_id to move_dest_id. Instead, use procurement's old move_id to populate stock_move's procurement_id
    and populate procurement's move_dest_id from the old move_id's move_dest_id.


addons/stock_account/migrations/8.0.1.1/post-migrate.py
```diff
@@ -42,7 +44,65 @@ def create_properties(cr, pool):
...
+        """
+        UPDATE procurement_order po
+        SET invoice_state = sm.invoice_state
+        FROM stock_move sm
+        WHERE sm.procurement_id = po.id
+    """)
+
+
+def inventory_period_id(cr, pool):
+    period_obj = pool['account.period']
```
pedrobaeza, line 83: Document this method putting something like "Fill account period in inventories."

addons/stock_account/migrations/8.0.1.1/post-migrate.py
```diff
@@ -42,7 +44,65 @@ def create_properties(cr, pool):
...
 
-@openupgrade.migrate()
+def propagate_invoice_state(cr):
+    """ Invoice state is now propagated from sale order to procurement to
+    stock move and picking. We trace it back from the picking and update
+    stock moves and procurements. """
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE stock_move sm
```
pedrobaeza, line 59: Can this be simplified like that?:

```
UPDATE stock_move sm
SET invoice_state = sp.invoice_state
FROM stock_picking sp
WHERE sm.picking_id = sp.id
AND sm.invoice_state != sp.invoice_state
```

pedrobaeza:
Two little remarks. The rest LGTM

addons/stock_account/migrations/8.0.1.1/post-migrate.py
StefanRijnhart, line 83: Done!

addons/stock_account/migrations/8.0.1.1/post-migrate.py
StefanRijnhart, line 59: That was my response when I got this query from my colleague @NL66278. We discussed and found out using the postgresql function 'explain analyse' that this subquery variant was 10 to 20 times as fast as the 'update...from' statement.

I know I have been going round myself requesting in reviews that subqueries be eliminated based on the assumption that joins are usually quicker. Turns out I was wrong.

I do agree that the join variant is more readable, and I find these queries easier to write, which is the reason that the other queries in this PR are still join based. None of these particular queries formed a bottleneck in the migrations that we performed. But it seemed silly not to use a more performant query if it is available.

pedrobaeza, line 59: Yeah, seems curious. Reading about that, PostgreSQL use the same optimizer for this clause than with INNER JOIN, so it should be in a good shape. Have you tried with a large set of data?

Anyways, please put a comment about what you told me for others that come to this code.

StefanRijnhart, line 59: This was with only a couple of thousand move lines. But the performance should be inherent to the database layout. Comment done!

pedrobaeza:
Please check PEP8 remarks:

```
./addons/stock_account/migrations/8.0.1.1/post-migrate.py:1:3: E231 missing whitespace after ';'
./addons/stock_account/migrations/8.0.1.1/post-migrate.py:1:3: E702 multiple statements on one line (semicolon)
./addons/stock_account/migrations/8.0.1.1/post-migrate.py:1:8: E231 missing whitespace after ';'
./addons/stock_account/migrations/8.0.1.1/post-migrate.py:1:8: E702 multiple statements on one line (semicolon)
./addons/stock_account/migrations/8.0.1.1/post-migrate.py:1:11: E261 at least two spaces before inline comment
./addons/stock_account/migrations/8.0.1.1/post-migrate.py:1:11: E901 SyntaxError: invalid syntax
```

StefanRijnhart:
Ugh. This happens when I use Emacs shortcuts before it is properly initialized...

pedrobaeza:
:+1: 

Thanks

addons/stock_account/migrations/8.0.1.1/post-migrate.py
```diff
@@ -42,7 +44,74 @@ def create_properties(cr, pool):
...
+    """
+    Run upon a fresh installation because this is a new glue module between
+    stock and account in Odoo 8.0.
+    Check for ourselves if the migration script applies by checking for a
+    specific legacy column that is migrated here and then removed.
+    """
+    if not openupgrade.column_exists(
+            cr, 'stock_inventory',
+            openupgrade.get_legacy_name('date_done')):
+        return
```
legalsylvain, line 113: :question: 
I don't understand in which case this script if usefull. If stock and account are installed, stock_account will be installed,  upgrade scripts will be executed one time and module will be marked as installed.

StefanRijnhart, line 113: This is in case someone installs the module on a fresh database (which happens in the analysis process for instance). Usually, the scripts are skipped because ''version'' is not None but because this is a new glue module we need a different heuristic.

legalsylvain, line 113: Thanks for your explanation.

legalsylvain:
:+1: (code review / no test.)

PR #300, opened 2015-05-15
Update the module's own external id after renaming it

bwrsandman:
Modules have an external id of the form `base.module_*`
This makes sure they are also renamed when the module name changes


StefanRijnhart:
:+1:, thanks!

pedrobaeza:
:+1:

pedrobaeza:
I merge due to the obvious of the patch.

PR #302, opened 2015-05-20
[FIX] cleanup a view from account_report_company

hbrunn:


pedrobaeza:
:+1:

StefanRijnhart:
Can you also clean up the dangling reference in ir_model_data?

hbrunn:
@StefanRijnhart shouldn't OpenUpgrade remove that automatically after the migration of account?

StefanRijnhart:
The uninstallation of account_report_company does that, yes. As such, it is an inconsistency but a harmless one at that. I guess I can live with that. :+1:

legalsylvain:
:+1: Thanks.

PR #303, opened 2015-05-20
[FIX] move xmlids of moved fields

hbrunn:
```
        # OpenUpgrade edit start: In rare cases, an old module defined a field
        # on a model that is not defined in another module earlier in the
        # chain of inheritance. Then we need to assign the ir.model.fields'
        # xmlid to this other module, otherwise the column would be dropped
        # when uninstalling the first module.
        # An example is res.partner#display_name defined in 7.0 by
        # account_report_company, but now the field belongs to the base
        # module
        # Given that we arrive here in order of inheritance, we simply check
        # if the field's xmlid belongs to a module already loaded, and if not,
        # drop the record to let the following standard code create it for the
        # correct module.
```

### DIFF
--- a/openerp/models.py
+++ b/openerp/models.py
@@ -385,9 +385,34 @@ class BaseModel(object):
                     (name_id, context['module'], 'ir.model', model_id)
                 )
 
-        cr.execute("SELECT * FROM ir_model_fields WHERE model=%s", (self._name,))
+        # OpenUpgrade edit start: In rare cases, an old module defined a field
+        # on a model that is not defined in another module earlier in the
+        # chain of inheritance. Then we need to assign the ir.model.fields'
+        # xmlid to this other module, otherwise the column would be dropped
+        # when uninstalling the first module.
+        # An example is res.partner#display_name defined in 7.0 by
+        # account_report_company, but now the field belongs to the base
+        # module
+        # Given that we arrive here in order of inheritance, we simply check
+        # if the field's xmlid belongs to a module already loaded, and if not,
+        # drop the record to let the following standard code create it for the
+        # correct module.
+        cr.execute(
+            "SELECT f.*, d.module, d.id as xmlid_id "
+            "FROM ir_model_fields f LEFT JOIN ir_model_data d "
+            "ON f.id=d.res_id and d.model='ir.model.fields' WHERE f.model=%s",
+            (self._name,))
+        # OpenUpgrade edit end
         cols = {}
         for rec in cr.dictfetchall():
+            # OpenUpgrade start:
+            if rec['name'] in self._columns.keys() and\
+                    rec['module'] != context.get('module') and\
+                    rec['module'] not in self.pool._init_modules:
+                cr.execute("DELETE FROM ir_model_fields WHERE id=%(id)s", rec)
+                cr.execute("DELETE FROM ir_model_data WHERE id=%(xmlid_id)s", rec)
+                continue
+            # OpenUpgrade end
             cols[rec['name']] = rec
 
         ir_model_fields_obj = self.pool.get('ir.model.fields')
--- a/openerp/models.py
+++ b/openerp/models.py
@@ -406,12 +406,18 @@ class BaseModel(object):
         cols = {}
         for rec in cr.dictfetchall():
             # OpenUpgrade start:
-            if rec['name'] in self._columns.keys() and\
+            if 'module' in context and\
+                    rec['name'] in self._columns.keys() and\
                     rec['module'] != context.get('module') and\
                     rec['module'] not in self.pool._init_modules:
-                cr.execute("DELETE FROM ir_model_fields WHERE id=%(id)s", rec)
-                cr.execute("DELETE FROM ir_model_data WHERE id=%(xmlid_id)s", rec)
-                continue
+                _logger.info(
+                    'Moving XMLID for ir.model.fields record of %s#%s '
+                    'from %s to %s',
+                    self._name, rec['name'], rec['module'], context['module'])
+                cr.execute(
+                    "UPDATE ir_model_data SET module=%(module)s "
+                    "WHERE id=%(xmlid_id)s",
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Thanks, this should prevent some weird column dropping in the long run!

If a field was formerly referenced in a technical module such as mass_update, this causes some slight data loss. If it is not too much trouble to actually replace the module in ir_model_data instead of dropping the row and the field itself, I would prefer that.

As a log voyeur, I would be pleased if you could log these XML ID replacements with level DEBUG.

hbrunn:
@StefanRijnhart thanks, an update is much better. And now you can marvel at log entries like

```
openerp.models: Moving XMLID for ir.model.fields record of res.partner#display_name from account_report_company to base
```

StefanRijnhart:
Thanks! :+1:

openerp/models.py
```diff
@@ -385,9 +385,40 @@ def _field_create(self, cr, context=None):
...
+        # on a model that is not defined in another module earlier in the
+        # chain of inheritance. Then we need to assign the ir.model.fields'
+        # xmlid to this other module, otherwise the column would be dropped
+        # when uninstalling the first module.
+        # An example is res.partner#display_name defined in 7.0 by
+        # account_report_company, but now the field belongs to the base
+        # module
+        # Given that we arrive here in order of inheritance, we simply check
+        # if the field's xmlid belongs to a module already loaded, and if not,
+        # drop the record to let the following standard code create it for the
```
pedrobaeza: update documentation

hbrunn: ?

pedrobaeza: s/and if not, drop the record/and if not, update the record with the new module name....

hbrunn: @pedrobaeza thanks! sorry for not getting it in the first place, the day was long...

pedrobaeza:
:+1:

mvaled:
It seems I didn't get affected by this:  Comparing the res_partner display_name column of my before and after migration DB shows data.  Actually there's one more row in the after DB:  The public user:

```
$ diff -u0 <(psql -d before -Ac "select id, display_name from res_partner order by id") <(psql -d after -Ac "select id, display_name from res_partner order by id")
--- /dev/fd/63  2015-05-25 08:57:00.616124886 -0400
+++ /dev/fd/62  2015-05-25 08:57:00.616124886 -0400
@@ -1971 +1971,2 @@
-(1969 rows)
+3051|Public user
+(1970 rows)
```

StefanRijnhart:
@mvaled yes, but have you tried uninstalling account_report_company on a test copy of your database?

mvaled:
@StefanRinhart No, I haven't...  The before DB had it installed:

```
$ psql -d before -c "select name from ir_module_module where state='installed' and name like 'account%'"
      name          
------------------------
 account
 account_voucher
 account_cancel
 account_accountant
 account_chart
 account_report_company
(6 rows)
```

But the after DB does not:

```
$ psql -d after -c "select name from ir_module_module where state='installed' and name like 'account%'"
    name        
--------------------
 account
 account_voucher
 account_cancel
 account_accountant
 account_chart
(5 rows)
```

In the after DB, this module is marked to be upgraded:

```
$ psql -d after -c "select name,state from ir_module_module where name like 'account%'"
               name                |    state    
-----------------------------------+-------------
 account                           | installed
 account_voucher                   | installed
 account_analytic_analysis         | uninstalled
 account_budget                    | uninstalled
 account_bank_statement_extensions | uninstalled
 account_check_writing             | uninstalled
 account_analytic_plans            | uninstalled
 account_test                      | uninstalled
 account_analytic_default          | uninstalled
 account_payment                   | uninstalled
 account_asset                     | uninstalled
 account_cancel                    | installed
 account_accountant                | installed
 account_chart                     | installed
 account_followup                  | uninstalled
 account_sequence                  | uninstalled
 account_anglo_saxon               | uninstalled
 account_report_company            | to upgrade
(18 rows)
```

In the current DB the module is actually uninstalled and we have not noticed any loss.  Comparing the res_partner shows a lot of movement in res_partner, but that's normal in our case.

PD: The before DB contains the last state our OpenERP 7 was, the after DB contains the state it was just after the OpenUpgrade migration was made, and the current DB, well is a daily dump of the working DB.

StefanRijnhart:
@mvaled that is the whole point of this fix. Uninstall the module and you lose the display_name column in res_partner because the field is registered in ir_model_data as belonging to this module. This was the case in 7.0, but in 8.0 it was moved to the base module.

Juicy detail: I got my hands on a SAAS database that lived through the official migration to 8.0 and it suffered from the same problem.

mvaled:
But then, how I didn't get the `display_name` lost?  Or did I misunderstand the issue?

Having recently migrated a production DB to Odoo using OpenUpgrade, I'm keeping an eye over this kind of merges, issues, etc., to make sure I didn't miss something.

StefanRijnhart:
Because you have not yet uninstalled the module. It is left dangling in the 'to upgrade' state because the module is not available on the 8.0. Go to the module in the interface, click 'Cancel upgrade', click 'Uninstall' -> Boom ;-)

mvaled:
Ah! I missed the uninstallable state in my current DB:

```
$ psql -d mercurio -c "select name, state from ir_module_module where name like 'account_report%'"
      name          |     state     
------------------------+---------------
 account_report_company | uninstallable
(1 row)
```

So I need to review the ir_model_fields and update accordingly.

mvaled:
Well, actually the "uninstallable"  state is my current copy of the DB.  I've just logged into the production and noticed that it is "uninstalled"  in the actual DB:

```
$ sudo -u openerp -H psql -d mercurio -c "select name, state from ir_module_module where name like 'account_report%'"
          name          |    state    
------------------------+-------------
 account_report_company | uninstalled
(1 row)
```

I'll still inspect the `ir_model_fields` more closely.

PR #305, opened 2015-05-22
[ADD] rename account_report_alt

hbrunn:
https://github.com/Therp/Therp-Addons/blob/6.1/account_report_alt/__openerp__.py#L31
and
https://github.com/OCA/account-financial-reporting/blob/7.0/account_financial_report_horizontal/__openerp__.py


pedrobaeza:
:+1: This is very straight, so I fast-track it.

PR #306, opened 2015-05-24
[IMP] base: Make contact names understandable

pedrobaeza:
Because `/` is not very readable, if you are not used to put name on addresses on 6.1, use similar schema than 6.1 address name_get: put street and city. I omitted country and added zip because I think that is more useful.

### DIFF
--- a/openerp/addons/base/migrations/7.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/7.0.1.3/post-migration.py
@@ -202,14 +202,25 @@ This can be the case if an additional module installed on your database changes
             row_cleaned = [val or False for val in row]
             address = dict(zip(fields, row_cleaned))
             partner_vals = address.copy()
+            # Try to compose an understandable name
+            if address['name']:
+                name = address['name']
+            else:
+                name = address['street'] or ''
+                if address['city']:
+                    name += ' - %s' % address['city']
+                if address['zip']:
+                    name += " (%s)" % address['zip']
+                if not name:
+                    name = '/'
             partner_defaults = {
                 # list of values that we should not overwrite
                 # in existing partners
                 'customer': False,
                 'is_company': address['type'] != 'contact',
                 'type': address['type'],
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Not sure. '/' is something of an Odoo convention and a clear marker for missing data (i.e you can search really easily for such records, which is not true for the substitution in this change).

pedrobaeza:
But here we want a usable system, which was in 6.1, although you don't put the address name. Imagine to force people to fill 5000 addresses with the '/'. It's more natural to fill it with valuable information.

Does it matter that this information comes from address name or from the rest of the data?

legalsylvain:
> Imagine to force people to fill 5000 addresses with the '/'.

:+1: I had this problem after my migration 6.1 -> 7.0.
From another side, I understand @StefanRijnhart remarks. I don't know...

pedrobaeza:
Can you review this, please?

StefanRijnhart:
Not convinced. Having a representation of the address as the partner's name does not strike me as being much more usable in the system than '/'. :-1:

pedrobaeza:
What do you suggest then?

StefanRijnhart:
Leave it as it is on our side. Based on the '/' marker, partners with this name can simply be exported and amended in a spreadsheet, then reimported.

pedrobaeza:
Well, I don't like that approach, but if you are not going to switch, I'll keep this patch for my own.

StefanRijnhart:
Any other reviewer to weigh in on this issue?

legalsylvain:
@rvalyi : any PoV about that PR ? (I ask as you analyse a lot partner changes between 6.1 & 7.0)

legalsylvain:
@pedrobaeza : As this PR doesn't receive a lot of acceptation, are you agree for closing it ? (It has been proposed 6 month ago).

pedrobaeza:
OK, but I'm using it on all my migrations. I'll continue catching the corresponding diff.

PR #307, opened 2015-05-24
[FIX] base: Propagate also customer and supplier flags from partner to address

pedrobaeza:
Propagate also these fields, because although at the beginning it feels that don't have addresses in the list with the default filter, this becomes quickly inconsistently when you create new partners + contacts. Also, you are not able to select these addresses on sales/purchase orders due to not having the corresponding field marked.


StefanRijnhart:
:+1:, this is consistent with the semantic change regarding contacts in Odoo 7.0.

legalsylvain:
:+1: LGTM.

hbrunn:
:+1:

PR #309, opened 2015-05-28
[ADD] rename some of the banking modules

hbrunn:
With https://github.com/OCA/bank-statement-import/pull/21, this restores a lot of functionality from 7.0


openerp/addons/openupgrade_records/lib/apriori.py
```diff
@@ -9,6 +9,11 @@
     'invoice_validation_wkfl': 'account_invoice_validation_workflow',
     # OCA/server-tools
     'audittrail': 'auditlog',
+    # OCA/bank-statement-import
+    'account_banking': 'account_bank_statement_import',
+    'account_banking_camt': 'bank_statement_parse_camt',
```
pedrobaeza, line 14: Where are these modules? Not in bank-statement-import...

hbrunn, line 14: right, that's still in PR state: https://github.com/OCA/bank-statement-import/pull/15 - the problem is that all those PRs are stuck because we want to wait how Laurent's PR to 9.0 is integrated, as those modules are supposed to be as close to 9.0 as possible...

pedrobaeza, line 14: OK, we can move on anyway with this PR

pedrobaeza:
:+1:

pedrobaeza:
Can we go ahead with this PR?

pedrobaeza:
As this renaming doesn't affect other processes, I merge it to clean up the list of pending PRs.

PR #310, opened 2015-05-28
[FIX] call the decorator

hbrunn:
who wrote this script?!? ;-)

Please fasttrack as this breaks the migration


legalsylvain:
:+1: 
;-)

PR #312, opened 2015-05-29
[FIX] base: Handle correctly migration of base_contact

pedrobaeza:
This modification handles correctly the migration of _base_contact_ module, making an update instead of inserting new records.

In v6.1, base_contact module has this structure:

```
res.partner
|
|-> res.partner.address -|
|-> res.partner.address -|
                         |-> res.partner.contact
res.partner              |
|                        |
|-> res.partner.address -|
```

And res.partner.contact contains information that is shared to linked res.partner.address via related fields.

In v7, now that we remove res.partner.contact, we should copy res.partner.contact information on each linked address, but not creating a new res.partner for each res.partner.contact.

### DIFF
--- a/openerp/addons/base/migrations/7.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/7.0.1.3/post-migration.py
@@ -61,71 +61,61 @@ def migrate_company(cr):
 
 def migrate_base_contact(cr):
     """
-    Move entries of res_partner_contact into res_partner
+    In v6.1, base_contact module has this structure:
+
+    res.partner
+    |
+    |-> res.partner.address -|
+    |-> res.partner.address -|
+                             |-> res.partner.contact
+    res.partner              |
+    |                        |
+    |-> res.partner.address -|
+
+    And res.partner.contact contains information that is shared to linked
+    res.partner.address via related fields.
+
+    In v7, now that we remove res.partner.contact, we should copy
+    res.partner.contact information on each linked address (that has been
+    converted also to a res.partner record), but not create a new res.partner
+    for each res.partner.contact.
     """
     cr.execute(
         "SELECT * FROM ir_module_module "
         "WHERE name = 'base_contact' and state = 'to remove';")
     if not cr.fetchall():
         return
-    fields = [
-        'create_date',
-        'name',
-        'website',
-        'image',
-        'active',
-        'comment',
-        'title',
-        'phone',
-        'country_id',
-        'email',
-        'birthdate',
-        'lang',
-        'parent_id',
-    ]
-    # Add lang from lang_id
+    # Add non-existing columns that can be used with partner_lastname module
+    cr.execute("ALTER TABLE res_partner "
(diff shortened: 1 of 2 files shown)


hbrunn:
isn't this going to run havoc on databases without base_contact installed? Shouldn't the script be in the migration of this module?

pedrobaeza:
This module doesn't exist in v7 (not in the same form), and the idea is to have base + base_contact on v6.1 and only base in v7.

Anyways, there's a check to avoid this to be run if base_contact is not installed: https://github.com/OpenUpgrade/OpenUpgrade/blob/7.0/openerp/addons/base/migrations/7.0.1.3/post-migration.py#L69

pedrobaeza:
And here is where the module is marked to be removed: https://github.com/OpenUpgrade/OpenUpgrade/blob/7.0/openerp/addons/base/migrations/7.0.1.3/pre-migration.py#L216

pedrobaeza:
Can you review, please?

pedrobaeza:
Any news about this one?

openerp/addons/base/migrations/7.0.1.3/post-migration.py
```diff
@@ -60,72 +60,43 @@ def migrate_company(cr):
...
-        'country_id',
-        'email',
-        'birthdate',
-        'lang',
-        'parent_id',
-    ]
-    # Add lang from lang_id
+    # Add non-existing columns that can be used with partner_lastname module
+    cr.execute("ALTER TABLE res_partner "
+               "ADD COLUMN firstname character varying;")
```
StefanRijnhart, line 102: Maybe check if these columns don't already exist?

openerp/addons/base/migrations/7.0.1.3/post-migration.py
```diff
@@ -60,72 +60,43 @@ def migrate_company(cr):
 
 
 def migrate_base_contact(cr):
-    """
-    Move entries of res_partner_contact into res_partner
-    """
+    """Move data of res_partner_contact into res_partner."""
```
StefanRijnhart: Can you document what this method does now into more detail?

openerp/addons/base/migrations/7.0.1.3/post-migration.py
pedrobaeza, line 102: I don't think they're going to exist, because all the possible references to last and first name are on contact or res.partner.address.

StefanRijnhart, line 102: Of course, thanks!

pedrobaeza:
@StefanRijnhart, method documented

StefanRijnhart:
Thanks! :+1:

pedrobaeza:
Can we merge this?

pedrobaeza:
Anyone more can check this?

paulius-sladkevicius:
Did code review, looks good :+1:

pedrobaeza:
Thanks. I merge due to 2 approvals and more than 5 days.

PR #316, opened 2015-06-02
[IMP] use same auto_install rules as OCB

hbrunn:
I think we shouldn't autoinstall modules that OCB doesn't autoinstall


pedrobaeza:
Me too :+1:

StefanRijnhart:
Tnx :+1:

PR #317, opened 2015-06-05

bwrsandman:


StefanRijnhart:
Hi @bwrsandman, thank you for your work on this.

You have moved all OpenUpgrade-specific python files to the external lib, but this does not make sense for 
- openupgrade_loading.py
- openupgrade_deferred80.py
- openupgrade_log.py

These files do not make generic methods available that can be called from the migration scripts.

Maybe these files should be kept here and removed from libopenupgrade?

pedrobaeza:
I have some questions more of organization: where do we now maintain the code for the extracted lib? Is the lib updated with the last changes we made?

bwrsandman:
@pedrobaeza sorry, I should have linked: https://github.com/OCA/openupgradelib 
@StefanRijnhart could you make the pull request on the lib and change the version to 0.1.2 as I am on vacation and I shouldn't even be here ;)

pedrobaeza:
Why not include the lib also in the OpenUpgrade organization (github.com/OpenUpgrade/openupgradelib)?

legalsylvain:
+1 to set openupgradelib into the same organization that openugprade. (To manage only one team, etc.).

I don't remember that OpenUpgrade is a project maintained officialy by the OCA.
But it could be ! (or maybe I missed some decision).
We can talk about that point in the next board meeting, if you want ?

pedrobaeza:
There was some statement last year in the Odoo days, but I'm not sure that oficially it was under the umbrella of OCA.

StefanRijnhart:
It should be uncontroversial after the announcement of Fabien at the Odoo Experience that they don't consider OpenUpgrade evil anymore, so +1 for adopting OpenUpgrade under the OCA. I would not mind being in the project steering committee for this one ;-)

pedrobaeza:
Count with me also for the PSC

bwrsandman:
@StefanRijnhart Bumped the lib version to 0.1.2 https://github.com/OCA/openupgradelib/commit/d1c83f979e6a21b24014f425a41947f2a0836f7b and rewrote the history to take out those files.

I, then, fixed this PR.

bwrsandman:
@StefanRijnhart ping

pedrobaeza:
@StefanRijnhart, can we work on this to finally use the lib? Because I want to propose some improvements, and it makes sense to make it only on the lib.

requirements.txt
```diff
@@ -39,3 +39,4 @@ vatnumber==1.2
 vobject==0.6.6
 wsgiref==0.1.2
 xlwt==0.7.5
+openupgradelib==0.1.2
```
pedrobaeza, line 42: This should be >= to allow new features/enhancements

bwrsandman:
@pedrobaeza done at #341

PR #319, opened 2015-06-13
[FIX] stock: Recreate correctly data of transferred moves

pedrobaeza:
With current script approach, no stock.pack.operation is recreated, losing information at stock.picking level (you can't print picking report with serial number), and with no link with moves for reversing operations.

This PR amends this, creating the equivalent stock.pack.operation, and calling after picking do_transfer method that creates the quant and makes the links.

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -3,6 +3,8 @@
 #
 #    Copyright (C) 2014 ONESTEin B.V.
 #              (C) 2014 Therp B.V.
+#              (C) 2015 Pedro M. Baeza
+#                       Antiun Ingenería, S.L.
 #    Snippets from odoo/addons/stock/stock.py (C) 2014 Odoo S.A.
 #
 #    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@legalsylvain can you check this?

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -652,15 +654,39 @@ def migrate_procurement_order(cr, registry):
...
+        INSERT INTO stock_pack_operation
+        (picking_id, product_id, product_uom_id, product_qty, qty_done,
+         lot_id, date, location_id, location_dest_id, processed)
+        SELECT
+            picking_id, product_id, product_uom, product_uom_qty,
+            product_uom_qty, %s, date, location_id, location_dest_id, 'true'
+        FROM stock_move
+        WHERE
+            id IN %%s
+            AND picking_id IS NOT NULL
```
clonedagain, line 676: What about the stock moves which are not attached to pickings ?

pedrobaeza, line 676: You're right. I think the correct approach is to recreate all quants processing all stock moves ordering by id (or maybe date), ignoring their source. I close this one and make another PR across the day.

PR #321, opened 2015-06-17
[FIX #288] migrate.py assign all external repositories as a module and not all of them are right written modules.

### LINKED ISSUE #288: from v7 to v8 the Report module not found.
I tried to update from v7 to v8, but migration script dead with the following error:

```
$ ./migrate.py --config=/.../etc/openerp-server.conf --database=ctmil_v7_0 --run-migrations=8.0 -B ../.. -A l10n_ar_migrate.py
$ tail -f migration.log
  File "/home/crocha/Projects/Update/8.0/server/openerp/service/server.py", line 909, in preload_registries
  File "/home/crocha/Projects/Update/8.0/server/openerp/modules/registry.py", line 366, in new
  File "/home/crocha/Projects/Update/8.0/server/openerp/modules/loading.py", line 369, in load_modules
    modobj.button_upgrade(cr, SUPERUSER_ID, ids)
  File "/home/crocha/Projects/Update/8.0/server/openerp/api.py", line 241, in wrapper
  File "/home/crocha/Projects/Update/8.0/server/openerp/addons/base/module/module.py", line 564, in button_upgrade
    raise orm.except_orm(_('Error'), _('You try to upgrade a module that depends on the module: %s.\nBut this module is not available in your system.') % (dep.name,))
except_orm: (u'Error', u'\ufeffIntenta actualizar un m\xf3dulo que depende del m\xf3dulo: report.\nPero este m\xf3dulo no est\xe1 disponible en su sistema.')
```

The first module with problem is "sale". Removing dependencies to "report" continue the migration, when other module with "report" dependency appear.

I think the main problem is because migration script no update the list of modules before load modules.

csrocha:
[FIX #288] migrate.py assign all external repositories as directories of modules, but some modules are in the root of the repository. Then odoo ignore the real module, and all subdirectories are assigned as new modules or shadowing the real ones. This fix use a common directory where all repositories are stored and find in them all **openerp**.py script to detect real modules. Detected modules will link it to the main addon directory.

### DIFF
--- a/scripts/migrate.py
+++ b/scripts/migrate.py
@@ -263,15 +263,14 @@ for version in options.migrations.split(','):
             if isinstance(addon_config, dict)\
             else {'url': addon_config}
         addon_config_type = addon_config.get('type', 'bzr')
-        if os.path.exists(os.path.join(options.branch_dir, version, name)):
+        target = os.path.join(options.branch_dir, version, 'rep_addons', name)
+        if os.path.exists(target):
             if addon_config_type == 'link':
                 continue
             elif addon_config_type == 'bzr':
                 cmd_revno = bzrlib.builtins.cmd_revno()
                 cmd_revno.outf = StringIO.StringIO()
-                cmd_revno.run(location=os.path.join(options.branch_dir,
-                                                    version,
-                                                    name))
+                cmd_revno.run(location=target)
                 print 'updating %s rev%s' % (
                     os.path.join(version, name),
                     cmd_revno.outf.getvalue().strip())
@@ -279,34 +278,28 @@ for version in options.migrations.split(','):
                 cmd_update.outf = StringIO.StringIO()
                 cmd_update.outf.encoding = 'utf8'
                 cmd_update.run(
-                    dir=os.path.join(options.branch_dir, version, name))
+                    dir=target)
                 if hasattr(cmd_update, '_operation'):
                     cmd_update.cleanup_now()
                 print 'now at rev' + cmd_revno.outf.getvalue().strip()
             elif addon_config_type == 'git':
                 os.system('cd %(location)s; git pull origin %(branch)s' % {
                     'branch': addon_config.get('branch', 'master'),
-                    'location': os.path.join(options.branch_dir,
-                                             version,
-                                             name),
+                    'location': target,
                 })
             else:
                 raise Exception('Unknown type %s' % addon_config_type)
         else:
             if addon_config_type == 'link':
-                print 'linking %s to %s' % (addon_config['url'],
-                                            os.path.join(options.branch_dir,
-                                                         version,
-                                                         name))
-                os.symlink(addon_config['url'],
-                           os.path.join(options.branch_dir, version, name))
+                print 'linking %s to %s' % (addon_config['url'], target)
+                os.symlink(addon_config['url'], target)
             elif addon_config_type == 'bzr':
                 print 'getting ' + addon_config['url']
                 cmd_checkout = bzrlib.builtins.cmd_checkout()
                 cmd_checkout.outf = StringIO.StringIO()
                 cmd_checkout.run(
                     addon_config['url'],
-                    os.path.join(options.branch_dir, version, name),
+                    target,
                     lightweight=True)
             elif addon_config_type == 'git':
                 print 'getting ' + addon_config['url']
@@ -315,12 +308,20 @@ for version in options.migrations.split(','):
                           {
                               'branch': addon_config.get('branch', 'master'),
                               'url': addon_config['url'],
-                              'target': os.path.join(options.branch_dir,
-                                                     version,
-                                                     name),
+                              'target': target,
                           })
             else:
                 raise Exception('Unknown type %s' % addon_config_type)
+
+        realtarget = os.path.realpath(target)
+        for root, dirs, files in os.walk(realtarget):
+            for file in files:
+                if file == "__openerp__.py":
+                    linkpath = os.path.join(options.branch_dir,
+                                            version, 'addons', name)
+                    if os.path.exists(linkpath):
+                        os.remove(linkpath)
+                    os.symlink(root, linkpath)
 
 db_name = conn_parms['database']
 if not options.inplace:
@@ -338,17 +339,10 @@ for version in options.migrations.split(','):
         'options',
         'addons_path',
         ','.join(
-            [os.path.join(options.branch_dir,
+            [os.path.realpath(os.path.join(options.branch_dir,
                           version,
-                          'server',
-                          migrations[version]['server']['addons_dir'])] +
-            [os.path.join(options.branch_dir,
-                          version,
-                          name,
-                          addon_conf.get('addons_dir', '')
-                          if isinstance(addon_conf, dict) else '')
-             for (name, addon_conf)
-             in migrations[version]['addons'].iteritems()]))
+                          'addons'))] +
+            []))
     config.set(
         'options',
         'root_path',
--- a/scripts/migrate.py
+++ b/scripts/migrate.py
@@ -340,8 +340,8 @@ for version in options.migrations.split(','):
         'addons_path',
         ','.join(
             [os.path.realpath(os.path.join(options.branch_dir,
-                          version,
-                          'addons'))] +
+                                           version,
+                                           'addons'))] +
             []))
     config.set(
         'options',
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Asking @hbrunn if he thinks this should be supported as he does most of the work on this script.

scripts/migrate.py
```diff
@@ -315,13 +308,21 @@ def deep_update(dict1, dict2):
...
 
+        realtarget = os.path.realpath(target)
+        for root, dirs, files in os.walk(realtarget):
+            for file in files:
+                if file == "__openerp__.py":
+                    linkpath = os.path.join(options.branch_dir,
+                                            version, 'addons', name)
+                    if os.path.exists(linkpath):
+                        os.remove(linkpath)
+                    os.symlink(root, linkpath)
```
StefanRijnhart: You may use 'break' here, I believe.

StefanRijnhart:
Ping @hbrunn

scripts/migrate.py
csrocha: Yes, exists a best optimized code. I will work on it, then commit it to this  pull request.

StefanRijnhart:
No reply from original author @hbrunn in more than two months, unfortunately. Personally, I think that repositories that don't include the module name level are non standard and unwise because the module name is essential and should not be left to something as arbitrary such as the name of the clone. So :-1: for making provisions for this non standard practice.

legalsylvain:
Thanks for your contribution but I agree with @StefanRijnhart. Setting repositories paths is not a good thing at all, and should not be promoted. :-1: 

I close this PR for the time being.

csrocha:
@StefanRijnhart , @legalsylvain  let me explain more what this patch do:

If you have two repository with the following directories:

Repository 1
- dir1
- dir2
- dir3

Repository 2
- dir4
- dir5
- dir6
- dir1

Here we have two directories dir1

One is a module in the Repository 1, but in Repository 2 this directory is documentation.

Some times (because depend how sort the repositories in the configuration) the original migrate.py could assign the dir1 of Repository2 as a module, and forgot dir1 of Repository1. If this happen Odoo will not recognize it.

The only way to be sure to check which of both is the real one is search for the _ _ openerp _ _.py file.

legalsylvain:
It's a non standard deployment. So you can keep this patch but I really see it useless for OpenUpgrade users.

I don't see any reason why you set documentation in such folder. 

could you point me your whole configuration ?
Do you have public repository ?

kind regards.

csrocha:
@legalsylvain Ok! It's not the right way to make deployments, but if happens OpenUpgrade is not useful. Why must be useless for alien modules just why is not OCA deployment compatible? Most of our deployments share modules developed by others and need to be upgrade them too. Not all is OCA in upgrade process.

My example is in a client customization, where not all modules are in the root of the repository, but in a "addons" directory. And has a directory for documentation, and it's name is "document".

StefanRijnhart:
@csrocha Sorry I misunderstood the purpose of this branch so thank you for clarifying. Nevertheless, repositories that are not valid addons paths themselves because they have directory names that compete with valid modules are non standard and changes to accomodate them are out of scope for this project. This does not mean that you cannot use OpenUpgrade for your project: You can either maintain these changes in your own branch and use your own version of migrate.py, or you can set up OpenUpgrade manually with its own addons paths (not using migrate.py).

Arguably, having directories with the same name compete with proper addons in the Odoo addons path is a bug in Odoo itself. Would it perhaps be possible and better to propose a fix there?

csrocha:
Ok. I will not insist about this patch. But If it's your choose, please be clear in the description about the "migrate.py" script. It's not support addons repositories without OCA organization which it's descripted in (who knows?)

About a bug in Odoo for the search of addons, for me this not a bug, just a decision. For me the bug belong to "migrate.py", and again, except if you describe this desicion in the OpenUpgrade documentation.

StefanRijnhart:
Yes, I agree that the documentation of adding addons is quite sparse for
this script.

pedrobaeza:
@csrocha, you're making an affirmation that is not correct: OpenUpgrade is fully compatible with any OpenERP/Odoo module, being OCA or not. The problem is you're messing the core mechanism to recognize addons introducing an alien folder - documentation -, not a module outside of OCA. If so, make your own automation script or make like me, use directly the migration preparing my config file, and starting the server with -u all parameter.

csrocha:
@pedrobaeza , I'm not talking about "modules", I'm talking about "repositories". Yes, OpenUpgrade support any Odoo addons, but not any organization of "repository". All this is about the issue #288 . I didn't found in the documentation (https://doc.therp.nl/openupgrade/migrate.py.html) any reference about repository organization. And, in the description about "-A" option in "--help", have no any reference about it.

```
$ python OpenUpgrade/scripts/migrate.py --help
Usage: migrate.py [options]

Migrate script for the impatient or lazy. Makes a copy of your database,
downloads the files necessary to migrate it as requested and runs the
migration on the copy (so your original database will not be touched). While
the migration is running only errors are shown, for a detailed log see
${branch-dir}/migration.log

Options:
  -h, --help            show this help message and exit
  -C CONFIG, --config=CONFIG
                        current openerp config (required)
  -D DATABASE, --database=DATABASE
                        current openerp database (required if not given in
                        config)
  -B BRANCH_DIR, --branch-dir=BRANCH_DIR
                        the directory to download openupgrade-server code to
                        [/var/tmp/openupgrade]
  -R MIGRATIONS, --run-migrations=MIGRATIONS
                        comma separated list of migrations to run, ie.
                        "6.0,6.1,7.0,8.0" (required)
  -A ADD, --add=ADD     load a python module that declares a dict 'migrations'
                        which is merged with the one of this script (see the
                        source for details). You also can pass a string that
                        evaluates to a dict. For the banking addons, pass
                        "{'6.1': {'addons': {'banking': 'lp:banking-
                        addons/6.1'}}}"
  -I, --inplace         don't copy database before attempting upgrade
                        (dangerous)
  -F FORCE_DEPS, --force-deps=FORCE_DEPS
                        force dependencies from a dict of the form
                        "{'module_name': ['new_dependency1',
                        'new_dependency2']}"
```

pedrobaeza:
Well, as said before by my colleague, you should propose a patch to core to ignore that folders instead of making a complicated selective copy of the downloaded repositories.

csrocha:
@pedrobaeza , the complicated selective copy of downloaded repositories is because the way migration.py script download the repositories.

migration.py download all external repositories in "os.path.join(options.branch_dir, version, name)"

This directory, for 9.0, is by default is: /var/tmp/openupgrade/9.0/the_git_repository_name . The result is, for example:

```
/var/tmp/openupgrade/9.0/odoo
/var/tmp/openupgrade/9.0/l10n-spain
/var/tmp/openupgrade/9.0/connector
/var/tmp/openupgrade/9.0/department
```

and others. Then add all this directories to the addon_path in the odoo server configuration file. Then you have three posible repositories organizations:
1. **OCA like**: Addons directories in the root of the repository. Works ok.
2. **Pure one addon**: It's one addon in the root of the repository. Doesn't work.
3. **Internal addons**: Where some directories in the repository exists addons. Works if you use the _addons_dir_ attribute in the migration directory, but it only works for one addon (https://github.com/OCA/OpenUpgrade/blob/8.0/scripts/migrate.py#L345)

The problem with the **Pure one addon** is if you add the directory to the **addon_path** Odoo understand all directories as an addon (report, model, migration, i18n, test, etc...) For the same case, for **Internal addons** you could have empty addons, because the real ones are hidding in other directory.

My request just found any addon (search for **openerp**.py) in all directories except "odoo" and link (NOT COPY) the addon directory to the server addon directory. My request is complex, ok, but solve any repository organization. Links are used by apache and other applications to "enable" and "disable" items (configurations, sites, etc...).

For me, change the Odoo core to ignore this directories is not the solution. The solution is keep clean the addons directory of aliens. If you want to change the Odoo core... then you must dive in the [module.py](https://github.com/odoo/odoo/blob/52484b0132ee3e0063dd67cc3c2eb434f99f5db8/openerp/modules/module.py#L341) file, but change it is more risky than modify the migration script. The risk about how simple is it to test and the number of people will affect a bug there.

pedrobaeza:
Thanks for the detail explanation. Well, I don't have a strong position on this with your rationale, because I'm not using the script. I prepare myself the addons directory (with links) to whatever I need, being one module repos, OCA-like repos or any other.

It can be an improvement indeed, so maybe @StefanRijnhart and @legalsylvain can rethink about the disapproval.

hbrunn:
@csrocha clone all your one-module-per-repo-repositories in `$somedir/my_modules` and then say 

```
migrate.py [...] -A "{'8.0': {
    'addons': {
        'my_modules': {'type': 'link', 'url': '$somedir/my_modules'},
    }
}}"
```

if you have a proposal how to update the documentation about that, please

PR #322, opened 2015-06-19
[IMP] account: Synchronize partner_id on account invoice lines

pedrobaeza:
I have ended up in several migrations with the same foreign key error in the account invoice line, so this little method avoid the issue. Anyways, in the best case, this SQL runs with no modifications.

### DIFF
--- a/addons/account/migrations/7.0.1.1/pre-migration.py
+++ b/addons/account/migrations/7.0.1.1/pre-migration.py
@@ -63,8 +63,26 @@ def fix_move_line_currency(cr):
         """)
 
 
+def sanitize_account_invoice_lines(cr):
+    """This method prevents that partner_id information on account invoice
+    lines is desynchronized with account invoice, and avoid possible foreign
+    key errors. This can be due to https://github.com/odoo/odoo/issues/6942 -
+    Stored related fields without recompute triggers are never recomputed.
+    """
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE account_invoice_line ail
+        SET partner_id=ai.partner_id
+        FROM account_invoice ai
+        WHERE ail.invoice_id=ai.id
+        AND ail.partner_id != ai.partner_id;
+        """)
+
+
 @openupgrade.migrate()
 def migrate(cr, version):
     openupgrade.rename_columns(cr, column_renames)
     openupgrade.rename_xmlids(cr, xmlid_renames)
     fix_move_line_currency(cr)
+    sanitize_account_invoice_lines(cr)


StefanRijnhart:
Harmless indeed, so :+1:. Any idea what caused this inconsistency?

pedrobaeza:
I think it's a bug of the old ORM dealing with related fields, but I didn't investigate further.

StefanRijnhart:
I think it's this bug: https://github.com/odoo/odoo/pull/6774. Stored related fields without recompute triggers are never recomputed. This is bigger than just this particular case. Maybe something for the database_cleanup module. Can you add this as a comment to the method?

pedrobaeza:
Uhm, not sure that it's that one, because that happens on v8 with new API, not in 6.1/6.0.

StefanRijnhart:
It was noted on the thread that this was already the case in older versions.

pedrobaeza:
Ahm, OK, I didn't see it. I'll document then the method. Anyways, I have only seen problems with this object, not another ones.

pedrobaeza:
Added the comment on the method. _database_cleanup_ can include this method, but you will face the problem before you can get the occasion to run it.

pedrobaeza:
Can you merge this if there's no opposition?

StefanRijnhart:
Do we need to be more strict about this now that this is an OCA project?

pedrobaeza:
Well, I don't think so. I have only faced this in this concrete case and a possible generic method can be time consuming in developing and execution.

StefanRijnhart:
No, I mean: merging without the minimum of two approvals

pedrobaeza:
Ah, OK, well, this is not a mass people repo, so maybe we can set a maximum delay to act, but please tell your opinion.

pedrobaeza:
Can anyone check this?

paulius-sladkevicius:
:+1:

pedrobaeza:
Thanks

StefanRijnhart:
And another OCA rule is not to merge your own work, of course (next time).

pedrobaeza:
@StefanRijnhart, I don't know why are you now so sensible now with these rules, but if you are so upset, I can stop and let these PRs die in the time. I don't have any problem.

StefanRijnhart:
Less well reviewed changes can be harmful. _You_ should be setting an example as a board member. And if there is a problem with not enough reviewers, than we should fix _or_ change the rules to adapt to that situation instead.

StefanRijnhart:
BTW I have always been strict with rules on OCA projects. It is just that on OpenUpgrade we did not have to abide by the OCA rules because it was not an OCA project.

pedrobaeza:
This PR is all except harmful. I'm giving my example, working in all branches, all repos, all reviews, so please don't tell me how to be a board member, OK? I'm strict enough. I'm not merging something that it's not tested. Everyone can make mistakes. I've made mistakes, you've made mistakes, and it's at one commit of distance to be amended. Just that.

StefanRijnhart:
That basically means that I cannot ask anyone else anymore not to merge their own changes because the board members do it too.

StefanRijnhart:
BTW if you think I am being too rigid with the rules, that is a personal trait I must admit so we don't have to decide on this issue here between the two of us ;-)

pedrobaeza:
BTW, I'm not asking that rule to anyone because I have never understood it and I find it absurd. It had any sense in the Launchpad times because it requires some manual operations and it can be screwed by a non-experienced person, but now that we only need to push a button...

We already have 2 approvals (that I have been waiting patiently for), the code is well tested by my part, so I have merged. What is the problem?

I have asked in other occasion to remove that rule and you silently ignored me...

StefanRijnhart:
Sorry, I had not understood that you asked me about that rule before if you did. So maybe it is good to discuss it in a larger context?

pedrobaeza:
Yeah, of course we can discuss the convenience of the rule. I'm writing to the contributors list.

pedrobaeza:
And apologize me if I sounded rude, but I'm a bit frustrated of my PRs not being treated, specially in this branch. And as I want to avoid this sensation to other collaborators, that's why I sometimes "bend" the rules. You know that there are other ways to bend these rules, like asking collaborators to blindly put a thumb up or similar, and I prefer to avoid that: if I merge something, it's up to me. Doesn't make that more sense?

StefanRijnhart:
Yes, I understand it is hard, especially on the less populated projects such as this one. I'll try and do some promotion of the project after the holiday season.

PR #323, opened 2015-06-19
[IMP+FIX] purchase: Avoid partners with no company in payment term search + SQL speed improvement

pedrobaeza:
Current code for assigning a payment term in the purchase order has a flaw: it relies on the company of the purchase order partner for getting the associated payment term, but you can have a partner with no company that is used for several companies. In this case, the script fails. Using the company of the purchase order is a more logical option for getting this property. 

I have also taken the occasion to improve this method  using a single SQL expression that it's damn faster.

### DIFF
--- a/addons/purchase/migrations/7.0.1.1/post-migration.py
+++ b/addons/purchase/migrations/7.0.1.1/post-migration.py
@@ -20,6 +20,10 @@
 ##############################################################################
 from openerp import pooler, SUPERUSER_ID
 from openerp.openupgrade import openupgrade, openupgrade_70
+import logging
+
+
+_logger = logging.getLogger(__name__)
 
 
 def migrate_purchase_order_addresses(cr, pool):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can anyone check this?

addons/purchase/migrations/7.0.1.1/post-migration.py
```diff
@@ -49,36 +53,22 @@ def migrate_purchase_order_line_names(cr, pool):
...
+    _logger.info("Setting payment term on purchase orders...")
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE purchase_order
+        SET payment_term_id =
+            substring(ir_property.value_reference from 22)::integer
+        FROM ir_property
+        WHERE ir_property.name = 'property_supplier_payment_term'
+        AND ir_property.company_id = purchase_order.company_id
```
paulius-sladkevicius: What if company is not set, what theoretical can happen on the ir_property?

pedrobaeza: That can't happen, as company_id is required on purchase_order.

paulius-sladkevicius:
Looks good! :+1:

addons/purchase/migrations/7.0.1.1/post-migration.py
hbrunn: @pedrobaeza I think what @komsas means is what if the property is a global one, that is, no company set in ir_property?
I think this must be `(ir_property.company_id = purchase_order.company_id or ir_property.company_id is null)`

addons/purchase/migrations/7.0.1.1/post-migration.py
```diff
@@ -49,36 +53,22 @@ def migrate_purchase_order_line_names(cr, pool):
...
+        AND ir_property.res_id =
+            'res.partner,' || purchase_order.partner_id::text
```
hbrunn: this is similar as above, also support global payment terms with `ir_property.res_id is null`

addons/purchase/migrations/7.0.1.1/post-migration.py
StefanRijnhart: @hbrunn does this actually give precedence to the property with the company if there is one? Otherwise the query needs to be modified accordingly.

pedrobaeza:
I have finally found a way to do via SQL the proper handling of ir_property priority. Please check and give the final push to this PR.

StefanRijnhart:
:+1: this looks good but I'd like @hbrunn to doublecheck.

pedrobaeza:
@hbrunn, can you check this one to finally merge it?

hbrunn:
:+1:

PR #324, opened 2015-06-26
[FIX] keep paid state

hbrunn:
adapt state+workflow for invoiced expenses without move

### DIFF
--- a/addons/hr_expense/migrations/7.0.1.0/post-migration.py
+++ b/addons/hr_expense/migrations/7.0.1.0/post-migration.py
@@ -26,8 +26,6 @@ def migrate_hr_expense_account_move(cr, pool):
     """
     Fill the field account_move_id with account_invoice information.
     """
-    invoice_obj = pool.get('account.invoice')
-    expense_obj = pool.get('hr.expense.expense')
     cr.execute("""
         UPDATE hr_expense_expense e
         SET account_move_id=i.move_id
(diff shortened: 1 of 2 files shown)


hbrunn:
Please don't invest much time in this right now, I'm still running a migration with this. But the gist is the following:
- the current migration wrongly assumed there's no paid state in 7.0 for expenses: https://github.com/odoo/odoo/blob/7.0/addons/hr_expense/hr_expense.py#L89
- the interesting part is, that this state is set circumventing the workflow mechanism: https://github.com/odoo/odoo/blob/7.0/addons/hr_expense/hr_expense.py#L474-L479

(possibly this is a historical thing and when the migration was written in the first place, those assumptions were true. I think I remember something like that)

So what I attempt with this code is:
- if state was 'paid', keep that, only afjust workflow state to act_done
- if state was 'invoiced', set state to 'done' and workflow state to act_done
- if state was 'invoiced' and the invoice had no moves, set state to 'accepted' and workflow state to act_accepted (in order to be able to make those payments later)

This way, we hopefully get consitent states/workflows for all possible earlier combinations.

pedrobaeza:
Tell me when it will be ready to check it

hbrunn:
@pedrobaeza now my migration is done and it looks like it did what it was supposed to do. Thanks!

pedrobaeza:
The changes seem OK, but I don't imagine now a case where the move is missing (but manual screw...)

Anyways :+1:

hbrunn:
@pedrobaeza thanks for the review! The move can be missing if an expense was in state 'invoiced' in 6.1, but the invoice was in state 'draft'

pedrobaeza:
OK, I see now

pedrobaeza:
Can anyone review this to go on?

paulius-sladkevicius:
did brief code review looks :+1:

pedrobaeza:
Thanks, merging!

PR #333, opened 2015-07-09
Fix duplicate removing script

ghost:
When the script is applied over a many2one field, the records will not be sorted properly
if some of them have null values.


pedrobaeza:
:+1:

bwrsandman:
@dufresnedavid This code is [being moved](https://github.com/OCA/OpenUpgrade/pull/317) to https://github.com/OCA/OpenUpgradeLib

Do you mind making the same PR there?

guewen:
:+1:

StefanRijnhart:
Thanks, but needs to be proposed onto openupgradelib instead.

PR #335, opened 2015-07-09
New logo suggestion for better visbility and impruved user attraction

blaggacao:
Have a look and if you like it consider "modernizing" the logo.

My motivation is to help with OpenUpgrade Adoption and the current logo is not so attractive, so many persons would probably discard this high quality project because of a "shaby" image. I think we do not want this?

No offence at all, btw. Just want to make things better :smiley:


blaggacao:
There is the original in svg (InkScape), consider yourself copyright owner.

legalsylvain:
:+1: thanks.

bwrsandman:
Can referred to by the rst?

pedrobaeza:
Thanks for the rebranding suggestion. Instead `update`, it should be `migration`. Also OpenERP word cannot be left out of the text, because from 5.0 to 7.0, this is the name for the product.

About branding itself, maybe the purple brand color looks good on "Open".

blaggacao:
I suppose, starting with a logo has multiple and tacit effects, which can drop in by and by. That's why this is my starting point. So all other questions of "migration" are as well important, but possibly a mine field.. So let's start with the simple and (generally) visible things...

From the Docs, I can deduct, that OpenUpgrade aims for "Simplicity" and is starting to tackle "Peace of Mind" as general solution attributes. I thing, clear lines, decent colors and simplistic logo does underpin exactly that.

blaggacao:
Sry @pedrobaeza I did completely misunderstand you, will change it asap in "migration"...
However, for a logo a tagline is already heavy stuff, I did it in order to be able to integrate elegantly the oca logo. Usually a logo has no tagline... So I'm pronouncing against direct reference to the old name "OpenERP" in this "logo"-tagline.. I wont be indexed anyhow ;-) Whaereas in text it can apear with the reference to OpenERP, obviously...

pedrobaeza:
OK, agree, then you should remove also Odoo, as OCA already include Odoo. The OCA migration library sounds very good.

blaggacao:
It looks quite nice already, especially with the black background... ;)
OCA as a token stands for an association, so syntactically it would say, that it is the "Migratory Library of the OCA-Organization". But it wouldn't say what is it for, without having the additional information that the OCA is something that has to do with Odoo. 

Obviously to insiders everything is crystalclear as mountain water. But that's the inside-out perspective. I wanted the logo to be a bridge to the outside-in perspective, though.

pedrobaeza:
I would like to omit Odoo word instead for avoiding copyright issues. As you have said, we can later explain in documentation for what is intended.

blaggacao:
alright, it has the advantage that its more readable also...

StefanRijnhart:
Thanks, I like it! :+1:

tafaRU:
Cool! :+1:

rafaelbn:
Guaranteed! :+1:

PR #338, opened 2015-07-31
[FIX] move categories from crm.case.categ to project.category

hbrunn:


pedrobaeza:
Don't you need to link new categories in issues?

hbrunn:
yes, I'm just testing the code that fixes this. Well spotted!

hbrunn:
@pedrobaeza done

addons/project_issue/migrations/7.0.1.0/post-migration.py
```diff
@@ -19,7 +19,25 @@
...
+def migrate_categories(cr, pool):
+    category_ids = pool['crm.case.categ'].search(
+        cr, SUPERUSER_ID, [('object_id.model', '=', 'project.issue')])
+    for category in pool['crm.case.categ'].browse(
+            cr, SUPERUSER_ID, category_ids):
+        new_category_id = pool['project.category'].create(
+            cr, SUPERUSER_ID, {
+                'name': category.name,
+            })
+        cr.execute('select id from project_issue where categ_id=%s',
```
pedrobaeza: This method is not valid, because you can overwrite other categories that are already changed. Think this:
- old categ_id=1. new categ_id=6. You change all records to categ_id=6.
- old categ_id=6. new categ_id=7. You change all records of old categ_id 1 and 6.

hbrunn: pay attention to the extra 's' in the write below. But things go wrong here anyways, because this write triggers recalculation of all function fields, and we don't want that. I'm currently busy rewriting this in sql

addons/project_issue/migrations/7.0.1.0/post-migration.py
```diff
@@ -19,7 +19,31 @@
...
+    for category_id, new_category_id in crm_category2project_category\
+            .iteritems():
+        cr.execute(
+            'insert into %s (%s, %s) '
+            'select id, %%s from project_issue where categ_id=%%s' %
+            pool['project.issue']._columns['categ_ids']._sql_names(
+                pool['project.issue']),
+            (new_category_id, category_id,))
+
+    pool['crm.case.categ'].unlink(cr, SUPERUSER_ID, category_ids)
```
StefanRijnhart: Please don't clean up these resources, even if they are unused in the core outside of this module. You don't know if there is any custom module that builds upon this functionality and still needs to perform its own migration at this point in the process.

pedrobaeza: And a new column `openupgrade_created_project_category` or something similar with a reference to the new ID would help in that cases (in the same way as the reference for the created partner from the address: https://github.com/OCA/OpenUpgrade/blob/7.0/openerp/addons/base/migrations/7.0.1.3/pre-migration.py#L185)

hbrunn:
@pedrobaeza @StefanRijnhart thanks, done

StefanRijnhart:
:+1:, thanks for the update.

pedrobaeza:
:+1:

PR #339, opened 2015-07-31
[FIX] actually get the state from matching analytic account

hbrunn:
the statement before didn't hook the join to the updated table, resulting in all states being copied from one account


pedrobaeza:
:+1: I fast-track because I faced this problem

PR #340, opened 2015-08-05
[IMP] set_defaults: Don't use ORM by default. Fixes #328

### LINKED ISSUE #328: set_defaults write values with ORM instead of SQL
I'm having very slow performance when migrating a DB with a lot of partners and I ended up finding that this method is the guilty of that.

@StefanRijnhart, I think you wrote this one. Is there any reason to use ORM in it instead of SQL?

pedrobaeza:
When merged, I will cherry-pick for 8.0

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -318,7 +318,7 @@ def warn_possible_dataloss(cr, pool, old_module, fields):
                     field['field'], field['new_module'], row[0])
 
 
-def set_defaults(cr, pool, default_spec, force=False):
+def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
     """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
We will also gain performance for sure (as the ORM write was done record by record).

paulius-sladkevicius:
Great improvement! :+1:

hbrunn:
:+1:

pedrobaeza:
Thanks for the merging, although we have to concentrate now on merging https://github.com/OCA/OpenUpgrade/pull/342 and its version variants to avoid duplicating effort, and I will need to submit again this improvement to openupgradelib.

pedrobaeza:
PR against openupgradelib made on https://github.com/OCA/openupgradelib/pull/4

PR #341, opened 2015-08-11

bwrsandman:
8.0 version


StefanRijnhart:
Gotten round to try and test this but when I add openupgradelib to my buildout.cfg, the buildout fails with

```
File "/tmp/easy_install-nqdGcY/openupgradelib-1.0.0/setup.py", line 5, in <module>
    import re
ImportError: No module named pip.req
An error occurred when trying to install openupgradelib 1.0.0. Look above this message for any errors that were output by easy_install.
anybox.recipe.odoo.base: Could not find or install '1.0.0'.  Original exception zc.buildout.UserError says: Couldn't install: openupgradelib 1.0.0
While:
  Installing odoo.
Error: Couldn't install: openupgradelib 1.0.0
```

Any clues?

bwrsandman:
You need pip as a dependency. I have removed that dependency for version 1.1.0 on the develop branch.

StefanRijnhart:
Thanks. I can successfully migrate a database with lots of modules using this code and the develop branch of openupgradelib in my buildout.

I did not get all the deprecation warnings that I expected, in fact I only got one and I think its message is slightly off.

```
Importing openerp.openupgrade.openerp.openupgrade.openupgrade_80 is deprecated. Use from openupgradelib import openupgrade
```

I think it should say to import openupgrade_80, not openupgrade.

If this is difficult to fix, I think we can live with that.

bwrsandman:
Not too much trouble. I also improved the message and fixed `openerp.openupgrade.openerp.openupgrade.openupgrade_80` to be the relative path

bwrsandman:
As for the lack of messages, this could be due to scripts simply importing `from openerp import openupgrade`, my last commit should fix that.

pedrobaeza:
I cannot get it working. I get:

```
No handlers could be found for logger "OpenUpgrade"
  File "/opt/odoo/v8/OpenUpgrade/odoo.py", line 160, in <module>
    main()
  File "/opt/odoo/v8/OpenUpgrade/odoo.py", line 156, in main
    import openerp
  File "/opt/odoo/v8/OpenUpgrade/openerp/__init__.py", line 76, in <module>
    import modules
  File "/opt/odoo/v8/OpenUpgrade/openerp/modules/__init__.py", line 27, in <module>
    from . import db, graph, loading, migration, module, registry
  File "/opt/odoo/v8/OpenUpgrade/openerp/modules/graph.py", line 32, in <module>
    import openerp.osv as osv
  File "/opt/odoo/v8/OpenUpgrade/openerp/osv/__init__.py", line 22, in <module>
    import osv
  File "/opt/odoo/v8/OpenUpgrade/openerp/osv/osv.py", line 23, in <module>
    from .orm import Model, TransientModel, AbstractModel
  File "/opt/odoo/v8/OpenUpgrade/openerp/osv/orm.py", line 5, in <module>
    from ..models import (
  File "/opt/odoo/v8/OpenUpgrade/openerp/models.py", line 61, in <module>
    from . import api
  File "/opt/odoo/v8/OpenUpgrade/openerp/api.py", line 71, in <module>
    from openerp.tools import frozendict, classproperty
  File "/opt/odoo/v8/OpenUpgrade/openerp/tools/__init__.py", line 27, in <module>
    from convert import *
  File "/opt/odoo/v8/OpenUpgrade/openerp/tools/convert.py", line 71, in <module>
    from openerp.openupgrade import openupgrade_log
  File "/opt/odoo/v8/OpenUpgrade/openerp/openupgrade/__init__.py", line 12, in <module>
    locals()[_i] = __import__("openupgradelib.%s" % _i)
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 26, in <module>
    from openerp import release, tools, SUPERUSER_ID
ImportError: cannot import name tools
```

I don't have Odoo installed, but I have always run it from the git source path without problems, and I would want to avoid making the install.

bwrsandman:
@pedrobaeza What command did you run? is it this:

```
from openerp import openupgrade
```

Seems that the openupgradelib doesn't have odoo in it's PATH. Can you confirm?

pedrobaeza:
No, it's running the migration itself. I run the command:

```
<path_to_openupgrade_git_source>/openerp-server -c <config_file> -u all --stop-after-init
```

That's the way I make the migration

bwrsandman:
OK, yeah I see... `from openerp.openupgrade import openupgrade_log`

I should remove 929af14, then.

bwrsandman:
@pedrobaeza try it now.

pedrobaeza:
Now it's working, but I don't get any deprecation warning. I only get:

```
No handlers could be found for logger "OpenUpgrade"
```

For me it's not a big problem I have to say.

bwrsandman:
That just means you won't get the warning. I am not sure how to go around that one. @StefanRijnhart Any ideas?

StefanRijnhart:
Suggesting https://github.com/bwrsandman/OpenUpgrade/pull/1/files to show a depreciation warning once for every imported file from openerp/openupgrade.

bwrsandman:
Done

StefanRijnhart:
Thanks, :+1:

pedrobaeza:
I have tried and it's working, but I still don't get any warning, but it's a minor detail, so :+1: 

So I think we can merge now, but first we need to push last openupgradelib to pypi. @bwrsandman, how is this operation done?

I suppose that we need also to upgrade docs/migrate.py script, doesn't it?

bwrsandman:
The command is `python setup.py sdist bdist upload`
You need to either be a maintainer of the package or have the OCA pypi user's password to do that.

pedrobaeza:
Thanks for the clarification, Sandy.

I let the "honor" to @StefanRijnhart to proceed with the last steps :wink:

StefanRijnhart:
@pedrobaeza, warnings are shown if you set

```
export PYTHONWARNINGS="default"
```

@bwrsandman Thanks again for your work on this.

bwrsandman:
\o/

PR #342, opened 2015-08-11

bwrsandman:
7.0 version of #341 


pedrobaeza:
@bwrsandman, can you please rebase? Or may I superseed the branch?

pedrobaeza:
Finished a 7.0 migration with this PR properly rebased and the changes in https://github.com/OCA/openupgradelib/pull/11, so :+1:

bwrsandman:
@pedrobaeza done

pedrobaeza:
:+1: again

pedrobaeza:
@StefanRijnhart, can you merge this one now that openupgradelib contains the fix for supporting this version? You should push to pypi that version although.

StefanRijnhart:
@pedrobaeza have you tested a migration with this branch?

StefanRijnhart:
Oh, I see you have, 28 days ago ;-)

StefanRijnhart:
@pedrobaeza I have made a release commit and tagged it 1.1.0, pushing it to develop and syncing develop to master. However, Pypi does not allow me to upload the release. I think it is because I am not an owner on the project. Only board members are owners of the project, so I think you will have to do it.

pedrobaeza:
OK, I have finally got it to push it, so now library is in final version. We should eventually grant access to a new openupgradelib-maintainers role for doing these things, although it can't be a group, but a login+password thing.

PR #343, opened 2015-08-11

bwrsandman:
6.1 version of #341


pedrobaeza:
For this version, there's this error:

```
    from openerp.tools.mail import plaintext2html
ImportError: No module named mail
```

Because this module is available from 7.0. We should protect this import with a `try: ... except ImportError:`

pedrobaeza:
:+1: when https://github.com/OCA/openupgradelib/pull/11 will be merged.

pedrobaeza:
Finishing migration, I found one last problem that has been fixed on https://github.com/bwrsandman/OpenUpgrade/pull/2. @bwrsandman, please merge, or if you really want to quit from all these duties, please tell me and I will superseed it with a new PR.

bwrsandman:
@pedrobaeza merged

pedrobaeza:
:+1: Thanks

pedrobaeza:
I also tried this one with the patched openupgradelib, so this one is ready to merge.

StefanRijnhart:
Thanks @pedrobaeza for testing!

PR #344, opened 2015-08-11

bwrsandman:
6.0 version


pedrobaeza:
I haven't tested this version yet in a real migration, but it seems to be correct.

StefanRijnhart:
:+1: as tested in other versions.

PR #349, opened 2015-08-17
[FIX] set default with use_orm=True

### LINKED ISSUE #347: Error migration 6.1->7.0: project addons
```
2015-08-15 00:22:28,193 84 ERROR openerp.sql_db: Programming error: column "use_tasks" of relation "project_project" does not exist
```

The migration fails in this line addons/project/migrations/7.0.1.1/post-migration.py#97

```
    openupgrade.set_defaults(
        cr, pool, {'project.project': [('use_tasks', None)]})
```

I have checked the official branch odoo/7.0 and this field use_tasks have a default of True but it is not defined in _columns.
Is it correct to remove that line?

hbrunn:
fixes #347

### DIFF
--- a/openerp/openupgrade/openupgrade.py
+++ b/openerp/openupgrade/openupgrade.py
@@ -346,12 +346,33 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
                 # Iterating over ids here as a workaround for lp:1131653
                 obj.write(cr, SUPERUSER_ID, [res_id], {field: value})
         else:
-            cr.execute(
-                """
-                UPDATE %s
-                SET %s = %%s
-                WHERE id IN %%s
-                """ % (obj._table, field), (value, tuple(ids)))
+            query, params = "UPDATE %s SET %s = %%s WHERE id IN %%s" % (
+                obj._table, field), (value, tuple(ids))
+            # handle fields inherited from somewhere else
+            if field not in obj._columns:
+                query, params = None, None
+                for table in obj._inherits:
+                    if obj._inherit_fields[field][0] != table:
+                        continue
+                    col = obj._inherits[table]
+                    # this is blatantly stolen and adapted from
+                    # https://github.com/OCA/OCB/blob/def7db0b93e45eda7b51b3b61
+                    # bae1e975d07968b/openerp/osv/orm.py#L4307
+                    nids = []
+                    for sub_ids in cr.split_for_in_conditions(ids):
+                        cr.execute(
+                            'SELECT DISTINCT %s FROM %s WHERE id IN %%s' % (
+                                col, obj._table), (sub_ids,))
+                        nids.extend(x for x, in cr.fetchall())
+                    query, params = "UPDATE %s SET %s = %%s WHERE id IN %%s" %\
+                        (pool[table]._table, field), (value, tuple(nids))
+            if not query:
+                raise Exception("Can't set default for %s on %s!",
+                                field, obj._name)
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Can you fix the set_defaults function instead so that it tests if the field is an inherited field instead, and modify the related table accordingly? This could save us some problems in the future.

hbrunn:
@pedrobaeza @StefanRijnhart here you go

openerp/openupgrade/openupgrade.py
```diff
@@ -346,12 +346,33 @@ def write_value(ids, field, value):
...
+                obj._table, field), (value, tuple(ids))
+            # handle fields inherited from somewhere else
+            if field not in obj._columns:
+                query, params = None, None
+                for table in obj._inherits:
+                    if obj._inherit_fields[field][0] != table:
+                        continue
+                    col = obj._inherits[table]
+                    # this is blatantly stolen and adapted from
+                    # https://github.com/OCA/OCB/blob/def7db0b93e45eda7b51b3b61
```
pedrobaeza, line 359: Can you replace this with an shortened URL? I have made it one with goo.gl for example: https://goo.gl/aj84Nf

hbrunn, line 359: I'd prefer not. Another dependency we've got to trust exists in 20 years, but the full hash + line number should be stable for a while

pedrobaeza:
Thanks for the effort of coping with this. We should make the PR against openupgradelib and push https://github.com/OCA/OpenUpgrade/pull/342 to be merged instead of making this patch.

I'm also thinking that instead of adding that bunch of code, maybe we can just switch to ORM mode when we detect that the field doesn't exist in the table, and we avoid to have some code that can change over versions. What do you think about it?

openerp/openupgrade/openupgrade.py
pedrobaeza, line 359: OK, you're right. Then we can put the line all together and put the noqa flag to avoid the check in that concrete line. Anyways, please check first my other comment changing the approach.

guewen, line 359: What is the reason to put it in one line? I think it is perfect like that.

hbrunn:
@pedrobaeza this code is not so very version specific: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L3927 - I'd keep it like that

openerp/openupgrade/openupgrade.py
```diff
@@ -346,12 +346,32 @@ def write_value(ids, field, value):
...
-                UPDATE %s
-                SET %s = %%s
-                WHERE id IN %%s
-                """ % (obj._table, field), (value, tuple(ids)))
+            query, params = "UPDATE %s SET %s = %%s WHERE id IN %%s" % (
+                obj._table, field), (value, tuple(ids))
+            # handle fields inherited from somewhere else
+            if field not in obj._columns:
+                query, params = None, None
+                for table in obj._inherits:
```
StefanRijnhart: Can you rename the misnamed variable _table_ (from the upstream code, I saw) to _model_?

StefanRijnhart:
@pedrobaeza is right. This code should be proposed to openupgradelib instead.

pedrobaeza:
I have made the proposal for you: https://github.com/OCA/openupgradelib/pull/19, so I close this one.

StefanRijnhart:
Merged in OpenUpgradelib. Thanks @hbrunn for proposing!

PR #351, opened 2015-08-17
[FIX] prefer the product's code to variant price

hbrunn:
Current code breaks if two variants have the same price, now we've got a chance if the variants have different codes


StefanRijnhart:
Makes sense, :+1:. CI failure looks unrelated.

hbrunn:
@StefanRijnhart it was, but I forgot to push the fix for it

pedrobaeza:
:+1:

PR #352, opened 2015-08-17
[FIX] drop old foreign key contraint

hbrunn:
Leaving it will cause trouble in the 8.0 migration of `hr_holidays`


pedrobaeza:
:+1:

StefanRijnhart:
:+1:, thanks

PR #353, opened 2015-08-18
[add] poweremail to email_template migration

hbrunn:
This is code @StefanRijnhart wrote a while ago, but he hesitated to publish it here because it might impair an eventual migrated version of poweremail. I think by now we can safely assume there's not going to be a public poweremail for 7.0, and make people happy who have to migrate old poweremail setups to email.template


pedrobaeza:
Code review :+1: 

I'll make a migration with this addon in a few weeks, so I'll try also in live soon.

StefanRijnhart:
:+1: as it has been used in production

pedrobaeza:
All goes well in my test, so merging.

PR #354, opened 2015-08-19
[FIX] Recursive import between openupgrade and osv

StefanRijnhart:
Trying to fix the documentation I encountered this recursive import again, which is exactly the reason that the table_exists function was isolated in its own module.


pedrobaeza:
:+1:

hbrunn:
:+1:

PR #357, opened 2015-08-24
[ADD] sale_layout migration

hbrunn:
Given that 6.1's sale_layout is not congruent, we can't migrate the layout exactly. What I do here is to simply create groups for every title line, and if there happens to be a separator, subtotal or line break within it, set the appropriate flag.


addons/sale_layout/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,110 @@
...
+                }[layout_type]
+                current_section.write({field: True})
+                if not line.product_id:
+                    line.unlink()
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
```
pedrobaeza, line 104: Good trick for using new API!

addons/sale_layout/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,110 @@
...
+            if layout_type == 'title':
+                current_section = env['sale_layout.category'].create({
+                    'name': line.name,
+                    'subtotal': False,
+                    'separator': False,
+                    'pagebreak': False,
+                    'sequence': sequence,
+                })
+                sequence += 1
+                if not line.product_id and line.state in ('draft', 'cancel'):
```
pedrobaeza, line 45: Why removing these lines?

addons/sale_layout/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,110 @@
...
+                continue
+            line.write({'sale_layout_cat_id': current_section.id})
+            if layout_type in ('subtotal', 'line', 'break'):
+                field = {
+                    'subtotal': 'subtotal',
+                    'line': 'separator',
+                    'break': 'pagebreak',
+                }[layout_type]
+                current_section.write({field: True})
+                if not line.product_id and line.state in ('draft', 'cancel'):
```
pedrobaeza, line 58: Why these 2 times?

addons/sale_layout/migrations/8.0.1.0/post-migration.py
hbrunn, line 45: in 6.1, you would use a line to only say "I'm a page break" etc. So if they don't do anything useful (checked by product_id), it's better to remove them because they'd mess up the layout.

addons/sale_layout/migrations/8.0.1.0/post-migration.py
hbrunn, line 58: the branch above is about titles, this one about the rest

addons/sale_layout/migrations/8.0.1.0/post-migration.py
pedrobaeza, line 45: OK, thanks for the clarification

addons/sale_layout/migrations/8.0.1.0/post-migration.py
pedrobaeza, line 58: Why not do it then in the upper part only one time before the if?

hbrunn, line 58: it would have to be conditional there too, so I don't think that makes the code easier to overlook

pedrobaeza, line 58: OK then

pedrobaeza:
:+1:

StefanRijnhart:
Can you maybe document somewhere that the module was missing in 7.0, but has returned in 8.0?

pedrobaeza:
@hbrunn, please attend @StefanRijnhart's comment and I'll merge.

StefanRijnhart:
Superseded by https://github.com/OCA/OpenUpgrade/pull/389

PR #358, opened 2015-08-24
[UPD] documentation

hbrunn:


pedrobaeza:
Checked manually that it's the same model file as there is no `migrations` folder, so :+1:

yvaucher:
:+1:

legalsylvain:
:+1:

PR #363, opened 2015-08-30
[FIX] stock migration script

pedrobaeza:
Final attempt of fixing stock migration. Still in progress

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -192,7 +192,6 @@ def migrate_stock_picking(cr, registry):
                 'one found (%s) to determine the picking types for this '
                 'company\'s pickings. Please verify this setting.',
                 company.name, warehouse.name)
-
         # Fill picking_type_id required field
         for picking_type, type_id in (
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This can be considered now tested and ready to be reviewed

sebalix:
I'm quite new to the new way to manage stock in v8.0, but is there any drawback to make the migration 7.0 -> 8.0 without this PR? Or is it only to preserve a consistent history?

parthivgls:
:+1:

pedrobaeza:
Well, basic operation of creating quants is already covered, but stock.pack.operation isn't, which can lead to inconsistencies in pickings pre and post-migration. Also the serial number is now assigned in quant creation for this pickings, instead of needing to visit again all the created quants.

sebalix:
Thank you, so I will do the migration with your branch while it's not merged (I'm not familiar enough with this subject to upvote this PR, others will do surely!)

pedrobaeza:
You can do that, check that all was correct, and then put your thumb up here.

sebalix:
Tested and all seems correct at a first glance, so :+1:

PR #364, opened 2015-09-03
[DOC] Run Analysis details / patch workflow doc stubs

blaggacao:
After having gone through this harsh learning curve, this is how I would have understood it upfront, if someone had explained it to me that way.

### DIFF
--- a/openerp/openupgrade/doc/source/analyse.rst
+++ b/openerp/openupgrade/doc/source/analyse.rst
@@ -1,25 +1,25 @@
 Run your own analysis
 =====================
 
-* First you need to prepare your python environment by installing the appropiate release of [openupgradelib](https://github.com/OCA/openupgradelib/releases)
+* First you need to prepare your python environment by installing the appropriate release of openupgradelib_.
+.. _openupgradelib: https://github.com/OCA/openupgradelib/releases
 
-* In order to analyze the difference of two codebases you need to specially prepare your different codebases by applying this OpenUpgrade patch onto your odoo code-base. (@ProjectLeader: Please provide pathces towards OCA/OCB and standard Odoo for different versions) This will add some folders of the OpenUpgrade tool suite and modify some odoo core files in order to work smoothly with OpenUpgrade.
+* In order to analyse the difference of two codebases you need to specially prepare your different codebases by applying this OpenUpgrade patch onto your Odoo code-base. (Note@ProjectLeader: Please provide patches towards OCA/OCB and standard Odoo for different versions) This will add some folders of the OpenUpgrade tool suite and modify some Odoo core files in order to work smoothly with OpenUpgrade.
 
 * Then create empty databases, even without demo data: one on top of the new code, the other on top of the old code.
 
 * Then install the modules you want to take into account on both databases. Also, install the module *openupgrade_records* on both databases.
 
-* If you have obvious changes in module or model names and you do not want your analyze output to be cluttered, include this mapping in the ['apripri.py'](https://github.com/OCA/OpenUpgrade/tree/8.0/openerp/addons/openupgrade_records/lib)
+* If you have obvious changes in module or model names and you do not want your analyse output to be cluttered, include this mapping in the ['apripri.py'](https://github.com/OCA/OpenUpgrade/tree/8.0/openerp/addons/openupgrade_records/lib)
 
-* On both instances: from the development menu, select the *Generate Records*
-  option.
+* On both instances: from the development menu, select the *Generate Records* option.
 
-* On the new (target) database, from the OpenUpgrade Development menu, set the server connection to your old (source) database in the "Comparision Config" menu. Test the connection with the appropiate button.
+* On the new (target) database, from the OpenUpgrade Development menu, set the server connection to your old (source) database in the "Comparison Config" menu. Test the connection with the appropriate button.
 
-* Make sure, that the odoo user has writing rights in the folders of your target database's source code.
+* Make sure, that the Odoo user has writing rights in the folders of your target database's source code.
 
-* In the target instance, in the *Comparison Config* form, click on *Perform Analysis*. This will create 'openupgrade_analisis.txt' files in the (newly created) migrations subfolders of each of your installed modules on your target source code. It will also creeate a 'openupgrade_general_log.txt' in the appropiate 'migrations' subfolder of the 'base' module.
+* In the target instance, in the *Comparison Config* form, click on *Perform Analysis*. This will create 'openupgrade_analisis.txt' files in the (newly created) migrations subfolders of each of your installed modules on your target source code. It will also create an 'openupgrade_general_log.txt' in the appropriate 'migrations' subfolder of the 'base' module.
 
-Read on in the docs on how to create migration scrips from thos analyze files.
+Read on in the docs on how to create migration scrips from those analyse files.
 
-Note that in many of the operations above you may get a client timeout or a concurrent access error even if the operation completes successfully. You should be able to assertain a succesful operation by verifying that all modules involved are in an installed state and the analysis files in the module directories have an appropriate modification time.
+Note that in many of the operations above you may get a client timeout or a concurrent access error even if the operation completes successfully. You should be able to ascertain a successful operation by verifying that all modules involved are in an installed state and the analysis files in the module directories have an appropriate modification time.
--- a/openerp/openupgrade/doc/source/analyse.rst
+++ b/openerp/openupgrade/doc/source/analyse.rst
@@ -1,29 +1,25 @@
-How to run your own analysis
-============================
-If you do need to run your own analysis, you need to perform the following
-steps (the awkward processing of the server log file is now obsolete).
+Run your own analysis
+=====================
 
-* Set up two OpenUpgrade servers of subsequent Odoo releases
+* First you need to prepare your python environment by installing the appropiate release of [openupgradelib](https://github.com/OCA/openupgradelib/releases)
 
-* On both instances, install a database without demo data and
-  install the *openupgrade_records* module, which is included in the
-  OpenUpgrade server distribution. This will add a menu
-  *OpenUpgrade Development* to the Administration menu.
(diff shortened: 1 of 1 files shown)


openerp/openupgrade/doc/source/analyse.rst
```diff
@@ -1,29 +1,25 @@
-How to run your own analysis
-============================
-If you do need to run your own analysis, you need to perform the following
-steps (the awkward processing of the server log file is now obsolete).
+Run your own analysis
+=====================
 
-* Set up two OpenUpgrade servers of subsequent Odoo releases
+* First you need to prepare your python environment by installing the appropiate release of [openupgradelib](https://github.com/OCA/openupgradelib/releases)
```
dreispt: s/python/Python

openerp/openupgrade/doc/source/analyse.rst
```diff
@@ -1,29 +1,25 @@
...
 
-* On both instances, install a database without demo data and
-  install the *openupgrade_records* module, which is included in the
-  OpenUpgrade server distribution. This will add a menu
-  *OpenUpgrade Development* to the Administration menu.
+* In order to analyze the difference of two codebases you need to specially prepare your different codebases by applying this OpenUpgrade patch onto your odoo code-base. (@ProjectLeader: Please provide pathces towards OCA/OCB and standard Odoo for different versions) This will add some folders of the OpenUpgrade tool suite and modify some odoo core files in order to work smoothly with OpenUpgrade.
```
dreispt: It's good to make the explanation clearer, but removing the bullets makes it less readable

dreispt:
In general, removing the bullets makes it less readable.
Keeping topic bullets followed by a paragraph expanding on them would be wiser.

blaggacao:
@dreispt Would you be able to provide patches against Odoo, OCA, OCB (at least for version 8.0?)
`git format-patch master --stdout > fix_empty_poster.patch`
`git apply --stat fix_empty_poster.patch`

But this would need to be curated (!), as there is no common ancestor in this repository (for whatever reason)...

StefanRijnhart:
Maybe it is me, but I don't think this is particularly clearer.

openerp/openupgrade/doc/source/analyse.rst
```diff
@@ -1,29 +1,27 @@
...
-should be able to assertain a succesful operation by verifying that all
-modules involved are in an installed state and the analysis files in the
-module directories have an appropriate modification time.
+Run your own analysis
+=====================
+
+1. First you need to prepare your Python environment by installing the appropriate release of openupgradelib_.
+
+2. In order to analyse the difference of two codebases you need to specially prepare your different codebases by applying this OpenUpgrade patch onto your Odoo code-base. (Note@ProjectLeader: Please provide patches towards OCA/OCB and standard Odoo for different versions) This will add some folders of the OpenUpgrade tool suite and modify some Odoo core files in order to work smoothly with OpenUpgrade.
+
```
StefanRijnhart, line 7: Sorry, OpenUpgrade is not a set of patches, it is a distribution.

blaggacao, line 7: Jep, that's how things are. I dared to document, how I think how things should be. For the following reason: 
- OpenUpgrade has no common ancestor with any other code base (in terms of git fork), so without a curated patch I'm not able to figure out reliably the changes applied by OpenUpgrade to the source code across the whole code base. (or at least, the connecting commit is not separately documented)
- OpenUpgrade works fine on a standard codebase (eg. Odoo 8.0), but it gets a bit harder, when you have custom modules, you would need to copy them to the OpenUpgrade Server. (If you've just one or two, that's still manageable, but if those are many, it gets harder)
- OpenUpgrade is especially hard to handle, if you altered the source code directly (few sane people do, but when they do, they might have good reasons for). (eg. planning to maintain a patch instead of evening out a bug in an own module with all it's extra difficulties etcpp)

Do I understand thinks wrong, if I'm thinking that something like this is very viable to do: http://www.ralfebert.de/git/patches/ (I hope you might be able to understand)

I'll prepare a layout shortly in my github repo order to show more clearly, what I mean.

blaggacao:
@StefanRijnhart It's a newcomer perspective. Probably at large the main target group of the documentation. Honestly, it did not understand it very well before. However, those are my own words. There is no guarantee, that other newcomer have a similar kind of thinking, nor, indeed, that they can follow this line of thought. I tried, however, to be more specific, overall.

openerp/openupgrade/doc/source/analyse.rst
nhomar, line 7: 2015-09-04 22:24 GMT-05:00 David Arnold notifications@github.com:

> > +2. In order to analyse the difference of two codebases you need to
> > specially prepare your different codebases by applying this OpenUpgrade
> > patch onto your Odoo code-base. (Note@ProjectLeader: Please provide
> > patches towards OCA/OCB and standard Odoo for different versions) This will
> > add some folders of the OpenUpgrade tool suite and modify some Odoo core
> > files in order to work smoothly with OpenUpgrade.

OpenUpgrade is a "Use One Time" distribution, You do not need to apply any
patch on any place, simply run and export the migrated db.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

blaggacao, line 7: I understand that. Nevertheless based on the my actual hypothesis (which might be understandably wrong), a patch would do the job better. What's the problem about that suggestion? If there is a better solution in sight, it is worth considering duly. OpenUpgrade is such a great tool! It's worth it :smile: 

Maybe there are misunderstandings: A patch (="diff applied") would ennoble every (copy of a) codebase to be a one time OpenUpgrade hero.

The actual process is just not very user friendly, that's all to say about this point :smile: 
(At least the doc claims user-friendliness: https://github.com/OCA/OpenUpgrade/blame/8.0/openerp/openupgrade/doc/source/intro.rst#L32)

blaggacao:
Hi, Stefan uploading of 2 entire repos took a while... :) This is for you to check, if this is a possible way to organize things:
https://github.com/DevCOco/OpenUpgrade
- odoo-8.0 : odoo codebase
- oca-8.0 : oca codebase
- odoo-8.0-openupgraded : OpenUgrade server (as found in the 8.0 branch)

Now the nice thing we can do easily is:
https://github.com/DevCOco/OpenUpgrade/compare/odoo-8.0...DevCOco:odoo-8.0-openupgraded
Download the patch: 
https://github.com/DevCOco/OpenUpgrade/compare/odoo-8.0...DevCOco:odoo-8.0-openupgraded.patch
or
https://github.com/DevCOco/OpenUpgrade/compare/odoo-8.0...DevCOco:odoo-8.0-openupgraded.diff

Like this I can easily make a codebase an OpenUpgrade Server, for example temporarily within a docker environment (read-only filesystem)... That would just be hillarious...

A possible workflow would be: `git rebase` the openupgraded branch onto the odoo main branch every now and then. Not sure if a separate ocb/openupgraded branch would need to be readily avialbale...

openerp/openupgrade/doc/source/analyse.rst
StefanRijnhart, line 7: @blaggacao In principle, it should be very easy of course to pull a patch from the diff between OpenUpgrade x.x and Odoo x.x (bar the occasional conflict with recent changes in Odoo). It just adds to the maintenance of OpenUpgrade. May I suggest that if you are interested in the availability of the project in this patch format, that you volunteer as the maintainer of the patches?

blaggacao, line 7: OK, let's see how a concrete workflow would be...
based off a curated rebase in order to get history in the right order.
1. git rebase openupgrade on respective odoo/ocb branch on each release of openupgrade

Then update the documentation accordingly.

I think I can do that... Next steps? Process of obtaining maintainers rights?

StefanRijnhart:
So maybe the point is that OpenUpgrade can simply be merged into Odoo already as long as there are no conflicts?

blaggacao:
Yes, but I think we would need a one time effort to determine a common ancestor between the actual Odoo branch and the openupgrade branch. Rebasing openupgrade from time to time would give me then a very clean diff and I would be able to understand openupgrade in no time...

pedrobaeza:
Currently this is the used technique for upcoming versions, but due to the migration from Launchpad to Github, existing branches can't be deducted that way, so sorry for that. As your PR is totally based in this principle, I close it.

blaggacao:
Does this mean, there is little interest in improving documentation, what should I do next? Nothing? Cannot interpret...

pedrobaeza:
No, it means that the documentation should be improved with proper things. The changes you proposed are not the way to do it. The correct way is the one that is already present.

blaggacao:
Ok, so what to do next? Open a new pull request without the proposed workflow stubs?
Or do you consider the actual documentation sufficient. As come clear from the issue description, I had a hard time understanding and diving into the topic based on the actual available docs...

pedrobaeza:
No, documentation can always be improved, but not rewriting it and putting something that is not the process at all. Basic steps are there. You can complete them with further information.

blaggacao:
Thanks. Honestly, I don't understand what exactly I'm expected to do next. However, I understand, what not to do. Are you fine with leaving thinks like they are and drop the discussion?

pedrobaeza:
OK, but I don't really get what do you want to do with OpenUpgrade that makes you try that ugly steps that you have defined in this PR.

blaggacao:
If you are interested in further reasoning set up a channel on gitter (gitter.com/OCA/OpenUpgrade/), and invite me to join, so I might be able to elaborate...

pedrobaeza:
Well, not for now. My curiosity is not so big :stuck_out_tongue_winking_eye:

gurneyalex:
@blaggacao note that the OCA already has an official IRC chan and official mailing lists. Please don't set up alternate non official channels of communication, especially using the OCA name and use the existing ones. 

Alternatively, if you feel a new dedicated chan is required, please request it by sending an email to "contribute at odoo-community.org" and the request will be considered. 

Thanks

blaggacao:
@gurneyalex OK, thanks for the guidance, will do that eventually, because IRC feels kind of like 2G Cel Network... :smile: Users get very demanding those days.. :smile_cat:

PR #365, opened 2015-09-03
[DOC] Migration Script Development v0.5

blaggacao:
Based on going through the learning curve. Thanks.

### DIFF
--- a/openerp/openupgrade/doc/source/development.rst
+++ b/openerp/openupgrade/doc/source/development.rst
@@ -10,7 +10,8 @@ Migration Scrips consist of two main files in every module's migration directory
 - pre-migration.py
 - post-migration.py
 
-What they do is derived from their names, you should check the API Documentation of the openupgradelib for details of standard methods available.
+What they do is derived from their names, you should check the API Documentation of the openupgradelib_ for details of standard methods available.
+.. _openupgradelib: https://github.com/OCA/openupgradelib/releases
 
 The general purpose of those scripts is to manipulate directly the database entries of the source database *copy* and change them in a way so that they are compatible with the target database source code. Note, that when running Odoo's '--update-all' flag on your *target source code* but on top of your *source database*, Odoo's internal updating mechanism will use your migrating scripts to manipulate your database in a way, so that it can be start up properly on the underlying (target) source code. If this succeeds, chances are, that your scripts are working, if not, you have to go over the scripts and make a new test run on a fresh copy (@Reviewer: or is a falling migration reverted?) of your source database.
 
--- a/openerp/openupgrade/doc/source/development.rst
+++ b/openerp/openupgrade/doc/source/development.rst
@@ -6,19 +6,18 @@ Migration script development
 
    devfaq
 
-Migration Scrips consist of two main files in every module's migration directory located in the target sourcecode:
+Migration Scrips consist of two main files in every module's migration directory located in the target source code:
 - pre-migration.py
 - post-migration.py
 
 What they do is derived from their names, you should check the API Documentation of the openupgradelib for details of standard methods available.
 
-The general purpose of those scirpts is to manipulate directly the database entries of the source database *copy* and change them in a way so that they are compatible with the target database sourcecode. Note, that when runnung Odoo's '--update-all' flag on your *target sourcecode* but on top of your *source database*, Odoo's internal updating mechanism will use your migrating scripts to manipulate your database in a way, so that it can be start up properly on the underlying (target) sourcecode. If this succeeds, chances are, that your scripts are working, if not, you have to go over the scripts and make a new testrun on a freah copy (@Reviewer: or is a faling migration reverted?) of your source database.
+The general purpose of those scripts is to manipulate directly the database entries of the source database *copy* and change them in a way so that they are compatible with the target database source code. Note, that when running Odoo's '--update-all' flag on your *target source code* but on top of your *source database*, Odoo's internal updating mechanism will use your migrating scripts to manipulate your database in a way, so that it can be start up properly on the underlying (target) source code. If this succeeds, chances are, that your scripts are working, if not, you have to go over the scripts and make a new test run on a fresh copy (@Reviewer: or is a falling migration reverted?) of your source database.
 
 General recommendations about 'pre-migration.py' and 'post-migration.py'
 ---------------------------------------------------------
 tbd by power-users
 
-Note on deffered upgrades
+Note on deferred upgrades
 -------------------------
-In some cases, you need to deferr some operations until the rest of the migration of the individual modules has terminated. The deferred_VERSION.py https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade is responsible for that. Note, that OpenUpgrade cannot see if migrations was sucessfull so it executes this script every time regardless of sucess of the migration. Therefore this script must be robust against repeting execution (include checks!)
-
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_VERSION.py https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade is responsible for that. Note, that OpenUpgrade cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
--- a/openerp/openupgrade/doc/source/development.rst
+++ b/openerp/openupgrade/doc/source/development.rst
@@ -1,24 +1,34 @@
+=========================================================
 Migration script development
-++++++++++++++++++++++++++++
+=========================================================
 
 .. toctree::
    :maxdepth: 2
 
    devfaq
 
-Migration Scrips consist of two main files in every module's migration directory located in the target source code:
-- pre-migration.py
-- post-migration.py
+Migration Scrips consist of two (or more) main files in every module's migration directory located in the target source code:
 
-What they do is derived from their names, you should check the API Documentation of the openupgradelib_ for details of standard methods available.
-.. _openupgradelib: https://github.com/OCA/openupgradelib/releases
+* pre-migration.py (or pre-anythingyouwant.py)
+* post-migration.py (or post-anythingyouwant.py)
 
-The general purpose of those scripts is to manipulate directly the database entries of the source database *copy* and change them in a way so that they are compatible with the target database source code. Note, that when running Odoo's '--update-all' flag on your *target source code* but on top of your *source database*, Odoo's internal updating mechanism will use your migrating scripts to manipulate your database in a way, so that it can be start up properly on the underlying (target) source code. If this succeeds, chances are, that your scripts are working, if not, you have to go over the scripts and make a new test run on a fresh copy (@Reviewer: or is a falling migration reverted?) of your source database.
+What they do is derived from their names, you should check the API Documentation_ of the openupgradelib_ for details of library methods that are readily available for your migration task.
+
+The general purpose of those scripts is to manipulate directly the database entries of the source database in order to change them in a way so that they are compatible with the target database source code. To make that happen smoothly, is your task.
+When you run (TODO DOC HOWTO RUN A CUSTOM MIGRATION) a migration, Odoo's *internal* updating mechanism will be used (OpenUpgrade is providing sugar code for the migration task only). (@Experts: Under what circumstances a failing migration is reverted?)
 
 General recommendations about 'pre-migration.py' and 'post-migration.py'
----------------------------------------------------------
-tbd by power-users
+------------------------------------------------------------------------
+tbd by an expert
 
-Note on deferred upgrades
--------------------------
-In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_VERSION.py https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade is responsible for that. Note, that OpenUpgrade cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+Special Case: Module Renames and Module Merges
+----------------------------------------------
+tbd by an expert
+
+Special Case: deferred upgrades
+-------------------------------
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_VERSION.py_ is responsible for that. Note, that OpenUpgrade cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+
+.. _Documentation: https://doc.therp.nl/openupgrade/API.html
+.. _openupgradelib: https://github.com/OCA/openupgradelib/releases
+.. _deferred_VERSION.py: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade
--- a/openerp/openupgrade/doc/source/development.rst
+++ b/openerp/openupgrade/doc/source/development.rst
@@ -34,7 +34,7 @@ Special Case: deferred upgrades
 In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
 
 .. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
-.. _Documentation: https://doc.therp.nl/openupgrade/API.html
+.. _Documentation: API.rst
 .. _openupgradelib: https://github.com/OCA/openupgradelib/releases
 .. _deferred_80.py: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade
 .. apriori.py_: https://github.com/OCA/OpenUpgrade/blob/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/openupgrade_records/lib/apriori.py
--- a/openerp/openupgrade/doc/source/development.rst
+++ b/openerp/openupgrade/doc/source/development.rst
@@ -25,7 +25,7 @@ TODO
 Special Case: Module Renames and Module Merges
 ----------------------------------------------
 In a normal case, you should be able to include your module name changes into the apriori.py_ which is automatically loaded at this_ point. However, in the case you only want to migrate a specific module, things are slightly more complicated (doc improvements via PR welcome):
- 1. In order for a module to be recognized for update, it must be formally installed. In targe source code, you need to fool the server by copying the newmodule and change it's name to old folder (which presumably is not present any more in your target source code). Additioanly, delete the migrations directory in the copied folder.
+ 1. In order for a module to be recognized for update, it must be formally installed. In targe source code, you need to fool the server by copying the new module and change it's name to old folder (which presumably is not present any more in your target source code). Additioanlly, delete the migrations directory in the copied folder.
  2. You then need to establish a dependency of this old name, depending on the new name. You can do that in the `[openupgrade]` stanza of the odoo config file with the `autoinstall` option.
  3. In the migration scripts, which live in your newname folder, put the those_ lines (pre-script).
 
@@ -34,10 +34,10 @@ Special Case: deferred upgrades
 In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
 
 .. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
-.. _Documentation: API.rst
+.. :doc:`Documentation<API>`
 .. _openupgradelib: https://github.com/OCA/openupgradelib/releases
 .. _deferred_80.py: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade
-.. apriori.py_: https://github.com/OCA/OpenUpgrade/blob/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/openupgrade_records/lib/apriori.py
+.. _apriori.py: https://github.com/OCA/OpenUpgrade/blob/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/openupgrade_records/lib/apriori.py
 .. _this: https://github.com/OCA/OpenUpgrade/blame/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/base/migrations/8.0.1.3/pre-migration.py#L37-L39
 .. _point: https://github.com/OCA/OpenUpgrade/blame/0a4814e0bf849e2b1f96b998f7271ae52e5cd8c5/openerp/modules/loading.py#L427-L429
 .. _those: https://github.com/OCA/OpenUpgrade/blame/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/base/migrations/8.0.1.3/pre-migration.py#L37-L39
--- a/openerp/openupgrade/doc/source/development.rst
+++ b/openerp/openupgrade/doc/source/development.rst
@@ -15,20 +15,29 @@ Migration Scrips consist of two (or more) main files in every module's migration
 What they do is derived from their names, you should check the API Documentation_ of the openupgradelib_ for details of library methods that are readily available for your migration task.
 
 The general purpose of those scripts is to manipulate directly the database entries of the source database in order to change them in a way so that they are compatible with the target database source code. To make that happen smoothly, is your task.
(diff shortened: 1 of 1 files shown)


blaggacao:
What happens in the case that I rename one module from let's say:

- 'old-code' : 'abc'
- 'old-code' : 'def'
merge to become more or less:
- 'new-code' : 'xyz'

How one needs to proceed?

openerp/openupgrade/doc/source/development.rst
```diff
@@ -5,3 +5,20 @@ Migration script development
    :maxdepth: 2
 
    devfaq
+
+Migration Scrips consist of two main files in every module's migration directory located in the target sourcecode:
+- pre-migration.py
+- post-migration.py
+
+What they do is derived from their names, you should check the API Documentation of the openupgradelib for details of standard methods available.
```
dreispt: It would be nice to add a link

openerp/openupgrade/doc/source/development.rst
```diff
@@ -5,3 +5,20 @@ Migration script development
...
+
+The general purpose of those scirpts is to manipulate directly the database entries of the source database *copy* and change them in a way so that they are compatible with the target database sourcecode. Note, that when runnung Odoo's '--update-all' flag on your *target sourcecode* but on top of your *source database*, Odoo's internal updating mechanism will use your migrating scripts to manipulate your database in a way, so that it can be start up properly on the underlying (target) sourcecode. If this succeeds, chances are, that your scripts are working, if not, you have to go over the scripts and make a new testrun on a freah copy (@Reviewer: or is a faling migration reverted?) of your source database.
```
dreispt: s/scirpts/scripts

dreispt: s/runnung/running

dreispt: s/freah/fresh

blaggacao:
@StefanRijnhart Hi Stefan, I included two sections of importance, that I'm not able to figure out, and less so explain in concequence. I wanted to ask, if you could take on the challenge to write the special case of renaming or merging modules?

I think the general recommendations for pre- and post- scripts can be shared across OpenUpgrade experts... @pedrobaeza @hbrunn ?

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,34 @@
...
+
+Migration Scrips consist of two (or more) main files in every module's migration directory located in the target source code:
+
+* pre-migration.py (or pre-anythingyouwant.py)
+* post-migration.py (or post-anythingyouwant.py)
+
+What they do is derived from their names, you should check the API Documentation_ of the openupgradelib_ for details of library methods that are readily available for your migration task.
+
+The general purpose of those scripts is to manipulate directly the database entries of the source database in order to change them in a way so that they are compatible with the target database source code. To make that happen smoothly, is your task.
+When you run (TODO DOC HOWTO RUN A CUSTOM MIGRATION) a migration, Odoo's *internal* updating mechanism will be used (OpenUpgrade is providing sugar code for the migration task only). (@Experts: Under what circumstances a failing migration is reverted?)
```
StefanRijnhart: How to run a migration is documented here: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database.
This is probably not yet documented sufficiently: unlike a plain Odoo upgrade, OpenUpgrade only commits after each individual module upgrade. If a migration script fails, the transaction in which that module migration is run will be rolled back. The migration process can then be restarted by running only with the -d argument for the database (i.e. without --update all). However, there are some technical subtleties that prevent a continued migration without manual intervention in the database.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,34 @@
...
+
+General recommendations about 'pre-migration.py' and 'post-migration.py'
+------------------------------------------------------------------------
+tbd by an expert
+
+Special Case: Module Renames and Module Merges
+----------------------------------------------
+tbd by an expert
```
StefanRijnhart: Renamed modules should be added to https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/openupgrade_records/lib/apriori.py. That way, it is handled in the analysis phase as well as at migration time.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,34 @@
...
+* post-migration.py (or post-anythingyouwant.py)
+
+What they do is derived from their names, you should check the API Documentation_ of the openupgradelib_ for details of library methods that are readily available for your migration task.
+
+The general purpose of those scripts is to manipulate directly the database entries of the source database in order to change them in a way so that they are compatible with the target database source code. To make that happen smoothly, is your task.
+When you run (TODO DOC HOWTO RUN A CUSTOM MIGRATION) a migration, Odoo's *internal* updating mechanism will be used (OpenUpgrade is providing sugar code for the migration task only). (@Experts: Under what circumstances a failing migration is reverted?)
+
+General recommendations about 'pre-migration.py' and 'post-migration.py'
+------------------------------------------------------------------------
+tbd by an expert
```
StefanRijnhart: There is some information here, but I'm sure it can be improved:

https://doc.therp.nl/openupgrade/format.html
https://doc.therp.nl/openupgrade/devfaq.html

openerp/openupgrade/doc/source/development.rst
StefanRijnhart: No specific support is present for merged modules. I'm not even sure if I have ever encountered a real world example of this.

openerp/openupgrade/doc/source/development.rst
blaggacao: Greate! Thanks! I'll try to figure out, how I could put this... Could you please point me to the bits of code, where this is being done?

openerp/openupgrade/doc/source/development.rst
blaggacao: OK, thanks. Then I encountered other problems, because for us, that did not work. A question for better understanding, so I can try document some internals: At what point Odoo's migration.py is passing on to touch OpenUpgrade code? (Without proper documentation and without clear patch information it's especially hard to figure out what actually happens, even by looking at the code.) I thought for example, that apriori.py is only touched at analysis time.
Merging -> Code Refactoring across versions.

openerp/openupgrade/doc/source/development.rst
blaggacao: I think what would be especially useful is to organize the mind of the reader in order to give him guidance on what sort of methods to put in which file...

openerp/openupgrade/doc/source/development.rst
blaggacao: Gathering already for the "how to run a migration": is the --update all absolutely necessary? Can this be an --update module_x as well? Can I performe the update from the odoo backend also? Would those options trigger the openupgrade routinge or the odoo one only? or none of them?

openerp/openupgrade/doc/source/development.rst
StefanRijnhart: https://github.com/odoo/odoo/blob/8.0/openerp/modules/migration.py

Yes I understand merging, just have not come across it in real life I think.

openerp/openupgrade/doc/source/development.rst
StefanRijnhart: Curious to know about your use case for migrating a single module using OpenUpgrade. We use the OpenUpgrade distribution to migrate database from one major version to another. However, when using OpenUpgradelib for minor upgrades in a regular Odoo instance, migrating only that module is fine of course.
What do you mean by backend? Running OpenUpgrade from the command line seems pretty backend to me.

openerp/openupgrade/doc/source/development.rst
blaggacao: Thanks, well hm, it is still very confusing, as the software does not behave as you said, in `apriori.py`, we have:  `'l10n_co_base': 'l10n_co_geo',` ` l10n_co_base` being the old name. A breakpoint in the openupgrade server `migrations.py` in the `migrate_module` method AFTER the `hasattr(pkg, 'update')` statment is never hit when launching with `--update l10n_co_geo`, neither so with `--update l10n_co_base`. So probably your statement about the `apriori.py` considerations at migrate time does have any implicit limitations? It seems that on one of the eligible modules, the script doesn't even enter the `migrate_module` method at all. It just returns without any error specific, nothing... Do you have an idea?

blaggacao: We feel it might have helped to document the alterations in `loading.py` which produces this awkward behavior and point out that nevertheless what you specify, always all *installed* modules are upgraded...

blaggacao: The problem seems to be the version checking?

blaggacao: It seems like in order to process update modules you need to combine the `--update` flag with the `--init` flag. :smile: 

Just out of curiosity, is line https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/modules/migration.py#L175 not confusing and redundant for the as it is already triggered in the try part in case of success?

StefanRijnhart: > It seems like in order to process update modules you need to combine the --update flag with the --init flag. :smile:

No, I don't think so.

blaggacao: Well, that was the only thing that worked. If you like we can analyze this further and improve the documentation. Please let me know, however, if you prefer to keep things as they ar, so not to spend time in vain..

blaggacao:
Should we continue with the intention to improve the docs or rather close this?

openerp/openupgrade/doc/source/development.rst
blaggacao: Sorry didn't see your comment here, we are actually further now, the use case is that we where working in data generation and module development simultaneously and now need to fix the mess.. ;) From backand I menat click on "Upgrade", but to my current underanding this should just work fine.. :)

openerp/openupgrade/doc/source/development.rst
blaggacao: Maybe I need to be clearer about our hypothesises:
We could not put the migration code in the old module, because then we get a version conflict in the `migrating.py`. We decided to respect this version check rather than manually fool it. But it seems from the code of `openupgrade.update_module_names`, that interpreting it right, this is not the expected way to do, as this handles a query on `ir_module_module` directly, which in our case, where the module needs to be initialized manually (!Update Modules list") beforehand would result in a double entry in `ir_module_module`. I suppose we are deviating here from the standard way to do it. But as said, we get a version conflict if we copy the migration scripts simply into the old module.

Another thing, which comes to my mind, would be to simply put `openupgrade.update_module_names` in the old module migrations folder, with manually crafted versions of course and then do the actual migrating in the new module folder in the canonical way in order to trigger that that module will be pulled...

I still think a thoughtful "standard way" of doing this that is well documented, is useful.

blaggacao: I feel like the `--init` trick is an alternative for some kind of deferred scripts https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade/deferred_80.py#L34

It is based on the fact, that in the `loading.py` `init` and `update` are handled in the same step. https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/modules/loading.py#L348

blaggacao:
Request for help: @StefanRijnhart @dreispt @hbrunn @bwrsandman @max3903 

In order to complete this pull request for doc improvement, I would like to elaborate, what's, finally, the correct steps to make a module name update?
We are still running around in the dark.

Is it something like this?
- after conducting analysis:
- in target src:
  - copy `new_module` folder and rename the copy to `old_module`
  - delete `new_module` folder
  - actualize module list (!)
  - put `update_module_names` in the migration scripts

This is how far we could get. If we can establish a standard and resilient way of doing this task, we might contribute some automation, but first, we need to know how to do this... :smile: 

Thanks in advance for maintaining our momentum on the topic by helping us as quickly as it might be convenient for you!

https://github.com/OCA/openupgradelib/pull/13 would then complement this renaming workflow by adding the possibility to merge a third module (which is a lot more complicated and would require additional docs once we shed light on this)

hbrunn:
If a module has changed names, you just put a mapping in https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/openupgrade_records/lib/apriori.py, that's it

blaggacao:
@hbrunn Thanks! It seems, this recommendation has further prerequisites. It didn't work in our case..
`apriori.py` is never called from the (modified) openupgrade `loading.py` so it's not clear to me how this should work... (isolated case of just one single module to migrate!)

hbrunn:
The pre_migrate step of the base module does it: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/base/migrations/8.0.1.3/pre-migration.py#L37 - So in your single module case, you're out of luck indeed.

In you case, you should create a module with the old name in the new version, have it depend in the new name and call the function linked above

blaggacao:
Oh thanks a lot! That's the missing piece to get things straight, once completely cleared, I will update the docs accordingly on this pull request.

blaggacao:
Oh thanks a lot! That's the missing piece to get things straight, once completely cleared, I will update the docs accordingly on this pull request.

As I understand you, please confirm:
A) Dependency on the newname is for the purpose that the new name will be loaded at all?
B) The dependency should or could (?) be set in the openupgrade stanza `autoinstall` of the config file?
C) The migration scripts reside in the newname?

blaggacao:
@hbrunn I feel, the `openupgrade.update_module_name` would break then as the new name would then already be initialized in the `ir_module_module`, wouldn't it?
This leads me to the feeling, that any module name change should be done in this step: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/modules/loading.py#L319

hbrunn:
hmm, right. then you'll have to adapt that function that it is robust against that

openerp/openupgrade/doc/source/development.rst
StefanRijnhart: How is migrating product prices to properties solved by --init? Plus, --init can be harmful in some cases.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+ 2. You then need to establish a dependency of this old name, depending on the new name. You can do that in the `[openupgrade]` stanza of the odoo config file with the `autoinstall` option.
+ 3. In the migration scripts, which live in your newname folder, put the those_ lines (pre-script).
+
+Special Case: deferred upgrades
+-------------------------------
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+
+.. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
+.. _Documentation: https://doc.therp.nl/openupgrade/API.html
```
StefanRijnhart: Please use internal RST references here.

blaggacao: Do you mean this empty file? https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade/doc/source/API.rst

So "_Documentation: API.rst" ?

StefanRijnhart: On 21-09-15 05:42, David Arnold wrote:

> Do you mean this empty file?
> https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade/doc/source/API.rst
> 
> So "_Documentation: API.rst" ?

Yes. It is not empty as it contains the directives to load the autodoc.

pedrobaeza:
I think this can be a first step in improving the documentation.

Thanks :+1:

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
+=========================================================
 Migration script development
-++++++++++++++++++++++++++++
+=========================================================
 
 .. toctree::
    :maxdepth: 2
 
    devfaq
+
```
StefanRijnhart, line 9: Index does not make any sense anymore. Please move your added lines to a separate file and refer to that in this index instead of adding it here.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+Migration Scrips consist of two (or more) main files in every module's migration directory located in the target source code:
```
StefanRijnhart, line 10: Can you reformat so that we get <80 char lines like in the rest of the docs?

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+Special Case: deferred upgrades
+-------------------------------
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+
+.. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
+.. _Documentation: API.rst
+.. _openupgradelib: https://github.com/OCA/openupgradelib/releases
+.. _deferred_80.py: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade
+.. apriori.py_: https://github.com/OCA/OpenUpgrade/blob/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/openupgrade_records/lib/apriori.py
+.. _this: https://github.com/OCA/OpenUpgrade/blame/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/base/migrations/8.0.1.3/pre-migration.py#L37-L39
```
StefanRijnhart, line 41: You probably want to refer to the blob, not to the blame in this link and the following links?

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+ 2. You then need to establish a dependency of this old name, depending on the new name. You can do that in the `[openupgrade]` stanza of the odoo config file with the `autoinstall` option.
+ 3. In the migration scripts, which live in your newname folder, put the those_ lines (pre-script).
+
+Special Case: deferred upgrades
+-------------------------------
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+
+.. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
+.. _Documentation: API.rst
```
StefanRijnhart: This link does indeed not work. Not sure how to do it as a footnote, but inline, this should work:

```
:doc:`Documentation<API>`
```

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+ 1. In order for a module to be recognized for update, it must be formally installed. In targe source code, you need to fool the server by copying the newmodule and change it's name to old folder (which presumably is not present any more in your target source code). Additioanly, delete the migrations directory in the copied folder.
+ 2. You then need to establish a dependency of this old name, depending on the new name. You can do that in the `[openupgrade]` stanza of the odoo config file with the `autoinstall` option.
+ 3. In the migration scripts, which live in your newname folder, put the those_ lines (pre-script).
+
+Special Case: deferred upgrades
+-------------------------------
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+
+.. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
```
StefanRijnhart, line 36: Another external link that needs to be refactored.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+
+General recommendations about 'pre-migration.py' and 'post-migration.py'
+------------------------------------------------------------------------
+TODO
+
+Special Case: Module Renames and Module Merges
+----------------------------------------------
+In a normal case, you should be able to include your module name changes into the apriori.py_ which is automatically loaded at this_ point. However, in the case you only want to migrate a specific module, things are slightly more complicated (doc improvements via PR welcome):
+ 1. In order for a module to be recognized for update, it must be formally installed. In targe source code, you need to fool the server by copying the newmodule and change it's name to old folder (which presumably is not present any more in your target source code). Additioanly, delete the migrations directory in the copied folder.
```
StefanRijnhart: Typo in _Additionally_, and missing space in _new module_.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+When you run a migration_, basically Odoo's *internal* updating mechanism will be used (OpenUpgrade is providing sugar code for the migration task only). 
+Expert Note: Unlike a plain Odoo upgrade, OpenUpgrade only commits after each individual module upgrade. If a migration script fails, the transaction in which that module migration is run will be rolled back. The migration process can then be restarted by running `odoo.py` only with the `-d` argument for the database (i.e. without `--update all`). However, there are some technical subtleties that prevent a continued migration without manual intervention in the database.
+
+General recommendations about 'pre-migration.py' and 'post-migration.py'
+------------------------------------------------------------------------
+TODO
+
+Special Case: Module Renames and Module Merges
+----------------------------------------------
+In a normal case, you should be able to include your module name changes into the apriori.py_ which is automatically loaded at this_ point. However, in the case you only want to migrate a specific module, things are slightly more complicated (doc improvements via PR welcome):
```
StefanRijnhart, line 27: Again, what was the rationale for migrating a specific module only? Sorry if I have missed this.

openerp/openupgrade/doc/source/development.rst
```diff
@@ -1,7 +1,43 @@
...
+
+Special Case: deferred upgrades
+-------------------------------
+In some cases, you need to defer some operations until the rest of the migration of the individual modules has terminated. The deferred_80.py_ is responsible for that. Note, that OpenUpgrade at that point_ cannot see if migrations was successful so it executes this script every time regardless of success of the migration. Therefore this script must be robust against repeating execution (include checks!)
+
+.. _migration: https://doc.therp.nl/openupgrade/intro.html#migrating-your-database
+.. _Documentation: API.rst
+.. _openupgradelib: https://github.com/OCA/openupgradelib/releases
+.. _deferred_80.py: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/openupgrade
+.. apriori.py_: https://github.com/OCA/OpenUpgrade/blob/e2726bff62cf5929f028fe3922f80a664f129179/openerp/addons/openupgrade_records/lib/apriori.py
```
StefanRijnhart: Underscore at the wrong end, so the reference is not found.

```
.. _apriory.py:
```

openerp/openupgrade/doc/source/development.rst
blaggacao, line 27: As I recall it, we did this when refactoring modules while already having data prepared in parallel.

openerp/openupgrade/doc/source/development.rst
blaggacao, line 9: I do not understand, what you mean, nor what to do exactly...

blaggacao:
I hope this is now mergable now. I'm sorry to need to retire otherwise from this PR. I cannot commit to keep stuff pending for that long time. Thanks for your understanding.

openerp/openupgrade/doc/source/development.rst
StefanRijnhart, line 27: As is evident in this part of the change, this is not covered by the OpenUpgrade distribution. Instead of getting out an OpenUpgrade branch and perform this elaborate instruction, I'd rather that people fire the necessary queries manually as described in the second part of https://gist.github.com/antespi/38978614e522b1a99563. You can leave this out or refer to this gist in a footnote if you want.

openerp/openupgrade/doc/source/development.rst
StefanRijnhart, line 9: This is the index of the document. But it does not refer to the page contents. Instead, it refers to another page and then the document contains other contents not listed in the index. I'm thinking the document layout make more sense if you would your text in a new file called guidelines.rst and add _guidelines_ under (or above) _devfaq_. Or you drop the toctree entirely and combine your text with the devfaq.rst file into this file, given that the faq is quite short still.

StefanRijnhart:
_You_'re giving up, after four people put a great deal of effort into this to help you get it merged? I find that disappointing.

blaggacao:
I'm really sorry for that. I hope someone else can probably pick it up and improve even further the docs. For me, the whole process is too complicated to keep on. I do not even know rst, and I cannot afford to learn it properly.

StefanRijnhart:
OK, closing this as per your comment. If there are any takers they can fork your branch and take it from here.

Issue #369, opened 2015-09-18
[6.1->7.0] Error if the 'warning' module is installed


sebalix:
The output:

```
ERROR bad query: INSERT INTO res_partner (name, active) VALUES(E'Administrator',true) RETURNING id
Traceback (most recent call last):
  File "/home/openerp/oerp/server/openerp/sql_db.py", line 226, in execute
    res = self._obj.execute(query, params)
IntegrityError: ERREUR:  une valeur NULL viole la contrainte NOT NULL de la colonne « purchase_warn »

2015-09-18 14:19:07,963 24769 ERROR test_61to70 openerp.sql_db: bad query: INSERT INTO res_partner (name, active) VALUES(E'Administrator',true) RETURNING id
Traceback (most recent call last):
  File "/home/openerp/oerp/server/openerp/sql_db.py", line 226, in execute
    res = self._obj.execute(query, params)
IntegrityError: ERREUR:  une valeur NULL viole la contrainte NOT NULL de la colonne « purchase_warn »
```

The `warning` module set some required fields on `res.partner` (`purchase_warn`, `picking_warn`, `sale_warn` and `invoice_warn`, with `no-message` as a default value for all of them), so the SQL query at https://github.com/OCA/OpenUpgrade/blob/7.0/openerp/addons/base/migrations/7.0.1.3/pre-migration.py#L180 fails.

pedrobaeza:
Uhm, this has difficult solution just right now, because we can't use ORM as this a pre-migration script, and I'm not sure if we can move that part to the post part. A tricky solution for your current migration is to add in the script the corresponding values or to remove the constraints in Postgres.

sebalix:
Yes, I'm working on a fix in the `pre-migration.py` script, I will make a PR right after.

StefanRijnhart:
Fixed by the related PR. Thanks!

PR #370, opened 2015-09-21
[FIX] Module 'base' - pre-migration.py - create_users_partner()

sebalix:
Change the INSERT INTO SQL query if the `warning` module is installed to satisfy required fields, fixing #369.

This is not the more robust way to do it for sure (don't know if there are other modules which add required fields on the `res_partner` table?), but as the `warning` module is a standard one, I think it's worth it.

### DIFF
--- a/openerp/addons/base/migrations/7.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/7.0.1.3/pre-migration.py
@@ -128,6 +128,13 @@ CREATE TABLE ir_model_relation (
             "ALTER TABLE ir_act_client INHERIT ir_actions")
 
 
+def is_module_installed(cr, module):
+    cr.execute(
+        "SELECT id FROM ir_module_module "
+        "WHERE name=%s and state='installed'", (module,))
+    return bool(cr.fetchone())
+
+
 def create_users_partner(cr):
     """
     Users now have an inherits on res.partner.
@@ -173,14 +180,24 @@ def create_users_partner(cr):
         "WHERE module='base' and name='user_root'")
     user_root = cr.fetchone()
     user_root_id = user_root and user_root[0] or 0
+    warning_installed = is_module_installed(cr, 'warning')
     cr.execute(
         "SELECT id, name, active FROM res_users "
         "WHERE partner_id IS NULL")
-    for row in cr.fetchall():
-        cr.execute(
+    sql_insert_partner = (
+        "INSERT INTO res_partner "
+        "(name, active) "
+        "VALUES(%s,%s) RETURNING id")
+    if warning_installed:
+        sql_insert_partner = (
             "INSERT INTO res_partner "
-            "(name, active) "
-            "VALUES(%s,%s) RETURNING id", row[1:])
+            "(name, active,picking_warn,sale_warn,"
+            "invoice_warn,purchase_warn) "
+            "VALUES("
+            "%s,%s,'no-message','no-message','no-message','no-message') "
+            "RETURNING id")
+    for row in cr.fetchall():
+        cr.execute(sql_insert_partner, row[1:])
         partner_id = cr.fetchone()[0]
         cr.execute(
             "UPDATE res_users "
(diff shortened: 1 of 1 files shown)


openerp/addons/base/migrations/7.0.1.3/pre-migration.py
```diff
@@ -176,11 +183,24 @@ def create_users_partner(cr):
...
-            "VALUES(%s,%s) RETURNING id", row[1:])
+        if warning_installed:
+            sql_insert_partner = (
+                "INSERT INTO res_partner "
+                "(name, active,picking_warn,sale_warn,"
+                "invoice_warn,purchase_warn) "
+                "VALUES(%s,%s, %s, %s, %s, %s) RETURNING id")
+            sql_args = (
+                row[0], row[1],
+                'no-message', 'no-message', 'no-message', 'no-message')
```
pedrobaeza: You don't need to pass 'no-message' as parameters. You can hard-code them into SQL. This way, you can save the SQL before the loop in a variable, and always use the same independently if the module is installed or not.

sebalix: That's right, I will do that.

sebalix: @pedrobaeza done

sebalix:
There was an error in the use of the database cursor, re-pushing.

openerp/addons/base/migrations/7.0.1.3/pre-migration.py
```diff
@@ -128,6 +128,13 @@ def update_base_sql(cr):
             "ALTER TABLE ir_act_client INHERIT ir_actions")
 
 
+def is_module_installed(cr, module):
+    cr.execute(
+        "SELECT id FROM ir_module_module "
+        "WHERE name=%s and state='installed'", (module,))
```
pedrobaeza: You should include also state 'to upgrade'. I'm thinking that this can be a good candidate for including as a method on openupgradelib. What do you think?

sebalix: I think I'm totally neophyte with OpenUpgrade, so... I can not disagree with you ;) You suggest me to make a PR on OCA/openupgradelib to add this method? Why not.

pedrobaeza: Yeah, although simple, we avoid to forget details like the 'to upgrade' state, or if the installed condition change avoid versions, we can abstract it with one call.

openerp/addons/base/migrations/7.0.1.3/pre-migration.py
```diff
@@ -128,6 +128,13 @@ def update_base_sql(cr):
             "ALTER TABLE ir_act_client INHERIT ir_actions")
 
 
+def is_module_installed(cr, module):
```
hbrunn: you can remove this as it's merges in https://github.com/OCA/openupgradelib/pull/20
don't forget to import it though

hbrunn:
I think we should take the time to do this properly:
- have a function that returns not null database fields for a given model (walk through postgres' metadata table and collect columns with not null constraints)
- use it here with a map of known fields and their default value, otherwise set 0, '', whatever according to the column's type

openerp/addons/base/migrations/7.0.1.3/pre-migration.py
sebalix: Done (and rebased too)

pedrobaeza:
:+1:, although @hbrunn's suggestions is the more general approach

StefanRijnhart:
I agree with @hbrunn in principle, but I don't expect that many changes in 7.0 so the advantage for future developments will be minimal. :+1: for this now.

pedrobaeza:
@hbrunn, may I merge then?

hbrunn:
@pedrobaeza I did it for you

PR #378, opened 2015-10-01
[FIX] don't crash if there are no done moves

hbrunn:


pedrobaeza:
I didn't think about the border case. :+1:

sebalix:
:+1:

PR #379, opened 2015-10-02
[FIX] Set l10n_be_intrastat to uninstallable, as it depends on l10n_be

StefanRijnhart:
which itself is set to uninstallable in OpenUpgrade because it messes
up the analysis.
This module was added after the last 8.0 analysis and it now makes the
analysis fail when comparing with 9.0

### DIFF
--- a/addons/l10n_be/__openerp__.py
+++ b/addons/l10n_be/__openerp__.py
@@ -81,7 +81,7 @@ Wizards provided by this module:
         'views/report_vatpartnerlisting.xml',
     ],
     'demo': [],
-    'installable': False,
+    'installable': True,
     'website': 'https://www.odoo.com/page/accounting',
 }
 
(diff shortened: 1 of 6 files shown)


StefanRijnhart:
The good news is that the l10n_be modules seem to have been fixed in 9.0 so they will be supported in the future

pedrobaeza:
:+1:

StefanRijnhart:
Modified this PR to make blacklisting of modules at analysis time explicit and a bit more elegant IMHO.

pedrobaeza:
Yeah, more elegant indeed. Finishing to improve this, maybe it's better to put the list in a separate list to ease the maintenance of this list across versions.

StefanRijnhart:
Sure. Where do you suggest we put this list?

pedrobaeza:
`openerp/addons/openupgrade_records/blacklist.py` or similar should be fine.

StefanRijnhart:
Done!

pedrobaeza:
Thanks!! :+1:

pedrobaeza:
Merging this also as it has been tested on the migration to v9.

PR #380, opened 2015-10-03
9.0 port framework

StefanRijnhart:


StefanRijnhart:
Here is a port of the OpenUpgrade analysis and migration framework on top of Odoo 9.0.

You may note:
- This change contains  module blacklisting instead of setting to uninstallable, as proposed to 8.0 in https://github.com/OCA/OpenUpgrade/pull/379
- A quirk has been added in the records module to allow reinstallation of the account module
- The l10n_be modules are now reinstallable, which means that they can be supported by OpenUpgrade (except that the analysis in 9.0 will treat it as a new module by necessity, and a proper analysis will only be possible from 9.0 to 10.0 on ).

This PR will allow an analysis comparison with 8.0, but there are indeed problems with comparing new API fields with old API fields. These will be treated in a follow up proposal.

pedrobaeza:
I made a code review and overall :+1:. I have made some little remarks.

StefanRijnhart:
To get an idea of the changes to the analysis code, this would be the basis of the next followup PR: https://github.com/StefanRijnhart/OpenUpgrade/compare/9.0-port_framework...StefanRijnhart:9.0-framework_new_api_fields

This gets me an analysis like the following: https://github.com/StefanRijnhart/OpenUpgrade/compare/9.0-framework_new_api_fields...StefanRijnhart:9.0-analysis.diff.

Work in progress, both.

pedrobaeza:
I think this can be merged to go on with following steps.

PR #382, opened 2015-10-12

StefanRijnhart:
Fixes confusion between related and computed fields (related fields are computed according to their related path).
Implements detection of fields becoming a property field (not seen in the wild, but still).
Prevent signalling new 'monetary' field as distinct from 'float' field.
Take 'default' attribute on new API fields into account.

### DIFF
--- a/openerp/addons/openupgrade_records/lib/compare.py
+++ b/openerp/addons/openupgrade_records/lib/compare.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2011 Therp BV (<http://therp.nl>).
+#    Copyright (C) 2011 Therp BV (<http://therp.nl>).
+#              (C) 2015 Opener B.V. (<https://opener.am>).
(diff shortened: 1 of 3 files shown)


openerp/addons/openupgrade_records/lib/compare.py
```diff
@@ -202,7 +216,8 @@ def match(match_fields, report_fields, warn=False):
         ]
     for column in old_records:
         # we do not care about removed function fields
-        if column['isfunction'] or column['field'] in IGNORE_FIELDS:
+        if (column['isfunction'] or column['isrelated'] or
```
pedrobaeza, line 219: This is a good candidate for being refactored to a method that can be called from the multiple places where you are making this change.

StefanRijnhart, line 219: Can I call on the rule of three here, and not do this now?

pedrobaeza, line 219: Of course, no problem. I told it as I see the same replace in several places, but indeed there are only 2 appearances.

pedrobaeza:
:+1:

PR #383, opened 2015-10-12

StefanRijnhart:
Backport of https://github.com/OCA/OpenUpgrade/pull/382, which is necessary to take advantage of the improved detection of these fields during the analysis of changes from 8.0 to 9.0.


pedrobaeza:
Same as in 9.0, so :+1:

pedrobaeza:
I'm going to merge these PRs, as they're very focused on the analysis, which is something that people don't usually work on, so it's difficult to get reviewed.

PR #384, opened 2015-10-12

StefanRijnhart:
Backport of https://github.com/OCA/OpenUpgrade/pull/382

### DIFF
--- a/openerp/addons/openupgrade_records/lib/compare.py
+++ b/openerp/addons/openupgrade_records/lib/compare.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2011 Therp BV (<http://therp.nl>).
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I don't think this problem was so old...

:+1:

StefanRijnhart:
Yes, it could be down there all the way back to 5.0 but I am not sure if it is worthwhile to backport it further down. I looked at 6.1 but the cherry pick wasn't exactly clean due to some refactorings in the OpenUpgrade patches between 6.1 and 7.0 so I decided to stop here for now. This is only a problem if someone wants to do an analysis between 6.1 and 7.0. Such an analysis will contain false message on all related fields, saying that they are not function fields anymore.

pedrobaeza:
Well, I don't think this kind of analysis is going to be done anymore. As the scripts and the analysis for standard modules is there, and a direct code analysis can make the rest, I don't think necessary. Maybe a mention on the module description can help, but you're making enough.

Off-topic: I migrate from 6.1 my last customer 2 weeks ago! (one that resists to go through version 7).

PR #395, opened 2015-11-08
[FIX] stock: Put correctly stock.move and stock.quant dates

pedrobaeza:
As the standard code overwrites date field when transferring, we need to keep the old date, and the created quants also need to take this date instead the date when the migration is done.

With this last patch, I think we can consider the stock migration as stable, having migrated already 4 DBs with plenty of stock stuff.


openerp/openupgrade/doc/source/modules70-80.rst
```diff
@@ -416,7 +416,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |share                              | Nothing to do                     |
 +-----------------------------------+-----------------------------------+
-|stock                              | Done (experimental)               |
+|stock                              | Done                              |
```
StefanRijnhart, line 419: Any experiences with multi-company + multi-warehouse? That was the aspect that I was unsure of when I added my bit to the stock migration script.

StefanRijnhart:
:+1: overall, thanks!

openerp/openupgrade/doc/source/modules70-80.rst
pedrobaeza, line 419: Uhm, not for multi-company, but with multi-warehouse I didn't have any special problem

StefanRijnhart, line 419: OK, thanks!

legalsylvain:
thanks !
:+1:

PR #397, opened 2015-11-14
[ADD] have migrate.py download its own copy of openupgradelib

hbrunn:
This is more in line with the script fetching everything specific to openupgrade


pedrobaeza:
Why do you need this?

hbrunn:
This guarantees that openupgradelib's version matches openupgrade's version. Given that `migrate.py` pull the latest sources of openupgrade, it might rely on features not yet available in the user's openupgradelib installed via pypi

pedrobaeza:
OK, I understand, because while we don't tag a commit, this won't be push to PyPi.

:+1:

StefanRijnhart:
:+1:

PR #398, opened 2015-11-15
[FIX] Cannot reinstall sale_timesheet and project modules

StefanRijnhart:
After a merge of upstream Odoo into OpenUpgrade 9.0, two modules could not be reinstalled anyore (reinstallation essential for the analysis process). It concerns a hack that creates XML ids for the same field that is present in two independent modules. This shared ownership of the data prevents the dropping of the column after deinstallation of the module that was installed first.

The hack consist of an XML record that forces the creation of the XML id in each module. As it is executed at installation time, and each XML id (module + name) is supposed to be unique, upon reinstallation the server crashes on uniqueness constraints in the ir_model_data table. This change disables the forced creation of the XML ids upon installation.

To prevent the original problem, the dropping of the column, the XML ids need to be added in the migration scripts of each module. That is something that this change does not yet implement though.

### DIFF
--- a/addons/project/project_data.xml
+++ b/addons/project/project_data.xml
@@ -23,14 +23,18 @@
 
 
         <!-- sale_timesheet and project define the same field without depending on each, which causes the field to be deleted when the module that created it
-        is deleted. To avoid this, we create xmlids manually for this field in both modules to prevent accidental deletion. To fix in saas-7 by moving the field definition-->
-        <record id="duplicate_field_xmlid" model="ir.model.data">
+        is deleted. To avoid this, we create xmlids manually for this field in both modules to prevent accidental deletion. To fix in saas-7 by moving the field definition
+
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why there's a Travis error on migration scripts on purchase?

StefanRijnhart:
Because we did not write it. It is a native Odoo script, apparently. I
had not noticed it before.

pedrobaeza:
Uhm, curious... What we do then? Refactor it to be PEP8 compliant? Remove it? The second option seems the best, as we can make the scripts already aware of the change made here.

StefanRijnhart:
I would leave it until the module is being covered by our own script, adding a noqa marker for the CI.

pedrobaeza:
OK for me

StefanRijnhart:
Done!

pedrobaeza:
Thanks :+1:

mmalorni:
:+1:

ghost:
:+1:

PR #399, opened 2015-11-17
[ADD] Analysis

StefanRijnhart:


StefanRijnhart:
Note that because the diff is too long, the Files Changed tab in this PR is slightly misleading. At some point, GitHub will not show the contents of newly added files anymore, saying that they 'have not changed'.

dreispt:
:+1:

parthivgls:
:+1:

pedrobaeza:
Only one question: do you think it deserves to be included the l10n_\* analysis file?

StefanRijnhart:
Yes, certainly. The l10n analysis files are only interesting in very few corner cases, but it doesn't really hurt to include them now, even if they make the diff of this particular PR very long. It would be more of a pain for the odd l10n developer to have to generate the analysis themselves in case they need it. Also for the regular OpenUpgrade migration script submission process it is essential to be able to check easily if an l10n module does not unexpectedly contains any noupdate data that has to be taken care of.

pedrobaeza:
OK, you're right. Thanks for the huge work you always do in this project when a new version comes. I merge this PR for allowing to start the migration scripts. For those who wants to contribute in this part, please comment on https://github.com/OCA/OpenUpgrade/issues/400 for avoiding the duplication of efforts.

PR #402, opened 2015-11-19
[FIX] don't pass None to string.join

### LINKED ISSUE #401: setting PYTHONPATH fails in migration.py script when PYTHONPATH is undefined
In line 337 of `migration.py` we have:

``` python
os.environ['PYTHONPATH'] = ':'.join([
    openupgradelib, os.environ.get('PYTHONPATH')])
```

but that fails when PATHONPATH is not previously setted up in the os enviroment. Maybe ..?

``` python
os.environ['PYTHONPATH'] = ':'.join([
    openupgradelib, os.environ.get('PYTHONPATH', '')])
```

hbrunn:
fixes #401

@yucer slightly different from your solution, but to the same effect

### DIFF
--- a/scripts/migrate.py
+++ b/scripts/migrate.py
@@ -333,8 +333,8 @@ else:
             'url': 'https://github.com/OCA/openupgradelib',
             'target': openupgradelib,
             })
-os.environ['PYTHONPATH'] = ':'.join([
-    openupgradelib, os.environ.get('PYTHONPATH')])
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Any of both serves :+1:

mmalorni:
:+1:

yucer:
Great ! Thanks

Sorry for the late response. Then we can close my ticket.

PR #405, opened 2015-11-25
[FIX] patch tables for setup_models()

coleste:
Odoo does not boot unless some tables are manually updated on the migrated database. As discussed in #404 .

### DIFF
--- a/openerp/modules/registry.py
+++ b/openerp/modules/registry.py
@@ -13,6 +13,7 @@ import openerp
 from .. import SUPERUSER_ID
 from openerp.tools import assertion_report, lazy_property, classproperty, config, topological_sort
 from openerp.tools.lru import LRU
+from openerp.openupgrade import openupgrade_loading_90
 
 _logger = logging.getLogger(__name__)
 
@@ -66,8 +67,8 @@ class Registry(Mapping):
             _logger.warning("The option --unaccent was given but no unaccent() function was found in database.")
         self.has_unaccent = openerp.tools.config['unaccent'] and has_unaccent
 
-        # ir_model must be in 9.0 format before continuing
-        self.patch_migrate_9(cr)
+        #OpenUpgrade: ir_model must be in 9.0 format before continuing
+        openupgrade_loading_90.migrate_model_tables(cr)
         cr.close()
 
     #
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Seems OK and can be run multiple times without breaking, so :+1:

hbrunn:
isn't this too late? Why not in `registry.__init__`? This way, it's run at https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/modules/loading.py#L133

coleste:
It crashes on cr.execute('select model, transient from ir_model where state=%s', ('manual',)), 3 lines below. But yes, it could be moved right after the has_unaccent bit. Should I? Done.

@hbrunn is this better?

mmalorni:
:+1:

openerp/modules/registry.py
```diff
@@ -65,6 +65,9 @@ def __init__(self, db_name):
         if openerp.tools.config['unaccent'] and not has_unaccent:
             _logger.warning("The option --unaccent was given but no unaccent() function was found in database.")
         self.has_unaccent = openerp.tools.config['unaccent'] and has_unaccent
+
+        # ir_model must be in 9.0 format before continuing
```
StefanRijnhart: Can you add a marker #OpenUpgrade to this change?

openerp/modules/registry.py
```diff
@@ -172,6 +175,34 @@ def load(self, cr, module):
 
         return [self.models[m] for m in models_to_load]
 
+    def patch_migrate_9(self, cr):
```
StefanRijnhart: Can you put this in a new module odoo/openerp/openupgrade/openupgrade_loading_90.py?

StefanRijnhart:
Thanks, looks good! Just a request to isolate the code into the OpenUpgrade corner as much as possible.

StefanRijnhart:
Thanks for the update. Next time, please write a comment that you followed up the comments, so that the reviewers get notified by email.
:+1:

StefanRijnhart:
Fixed trivial flake8 issue in direct commit https://github.com/OCA/OpenUpgrade/commit/cbced5496c81f9d328da408f27751471e78aa73f

coleste:
I will, and thanks for the whitespace!

coleste:
@StefanRijnhart , I missed something and had to resubmit as #410 
I missed the commit()

PR #406, opened 2015-11-25
[WIP] 9.0 Project

ghost:


addons/project/migrations/9.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,38 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.openupgrade import openupgrade
```
pedrobaeza: Please use new library call `from openupgradelib import openugprade`

addons/project/migrations/9.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,38 @@
...
+from openerp.addons.openupgrade_records.lib import apriori
```
pedrobaeza: Why do you need this?

addons/project/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,62 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import logging
+from openerp import pooler, SUPERUSER_ID
+from openerp.openupgrade import openupgrade
```
pedrobaeza: Please use new library call `from openupgradelib import openugprade`

addons/project/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,62 @@
...
+logger = logging.getLogger('OpenUpgrade')
+
+# copied from pre-migration
+column_copies = {
```
pedrobaeza, line 27: This is not needed here

addons/project/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,62 @@
...
+        [('template', 'draft')],
+        table='project_project', write='sql')
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
+    map_priority(cr)
+    map_template_state(cr)
+    for table_name in column_spec.keys():
```
pedrobaeza: Where is `column_spec`?

addons/project/migrations/9.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,38 @@
...
+column_copies = {
+    'project.task': [
+        ('description', None, None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.copy_columns(cr, column_copies)
```
pedrobaeza, line 63: Why do you do this?

ghost:
@pedrobaeza Please keep in mind that's it a WIP! ;-)

Thanks for your comments, I will fix what needs to be fixed ASAP (which should be tomorrow). 

BTW, I'm still not able to test anything that I've done, as the "bootstrapping" fix has not yet been completed at this time.

pedrobaeza:
Yeah, no problem. I made an early review just for guiding you (or trying to).

addons/project/migrations/9.0.1.1/pre-migration.py
ghost: I copy-pasted code from another script. I'm cleaning it at this very moment!

addons/project/migrations/9.0.1.1/post-migration.py
ghost, line 27: You mean I can completely remove "column_copies" code block?

# copied from pre-migration

column_copies = {
    'project.task': [
        ('description', None, None),
    ],
}

addons/project/migrations/9.0.1.1/pre-migration.py
ghost, line 63: Because this field has changed its type from text to html (see openupgrade_analysis_work.txt).

And because looking at other scripts, it seems to be the way to do it. Now like I said before, I wasn't able to test this.

If there is a better way of doing this, I'd sure like to know as I'm pretty new to all this...

ghost:
I really appreciate (and need!) your guidance to do this, thanks!

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,162 @@
...
+project      / project.project          / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project      / project.project          / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+project      / project.project          / state (selection)             : selection_keys is now '['cancelled', 'close', 'draft', 'open', 'pending']' ('['cancelled', 'close', 'draft', 'open', 'pending', 'template']')
+# Done ---> 'template' is converted to 'draft'
+# TODO - does this behavior make sense on the functional level?
+
+project      / project.project          / user_id (many2one)            : NEW relation: res.users
+# TODO - does someone know where user_id comes from?
```
dreispt: This can't be right: `user_id`exists on 8.0 and is the project manager, but maybe I came from the `mail` module.
I think there's nothing to do.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,162 @@
...
+
+project      / project.project          / website_message_ids (one2many): NEW relation: mail.message
+# Field doesn't exist neither in v8 or in v9...weird! ---> Nothing to do
```
dreispt: Probably related to a `mail` module.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,162 @@
...
+project      / project.task             / kanban_state (selection)      : now required, default = normal
+# Field is now required and default is the same as before ---> Nothing to do
+
+project      / project.task             / message_follower_ids (many2many): not a function anymore
+project      / project.task             / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project      / project.task             / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+project      / project.task             / priority (selection)          : selection_keys is now '['0', '1']' ('['0', '1', '2']')
+# Done ---> '2' is converted to '1' and '1' is converted to '0'
```
dreispt: It should only convert 2 to 1: the logic is starred vs not starred. two stars means starred.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,162 @@
...
+# (legend: in v8 = [('0','Low'), ('1','Normal'), ('2','High')] and in v9 = ('0','Normal'), ('1','High'))
+
+project      / project.task             / reviewer_id (many2one)        : DEL relation: res.users
+# Field removed, no more task reviewers ---> Nothing to do
```
dreispt: Ouch! I'll need an OCA module to restore this...

pedrobaeza: Then the column should be renamed to None to allow to recover it when installing this new module.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,162 @@
...
+
+project      / project.task             / website_message_ids (one2many): NEW relation: mail.message
+# Field doesn't exist neither in v8 or in v9...weird! ---> Nothing to do
+
+project      / project.task             / work_ids (one2many)           : DEL relation: project.task.work
```
dreispt, line 68: AFAICS Task Work in now supported through Timesheets.
Migration script for this will be needed.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,162 @@
+---Fields in module 'project'---
+project      / project.project          / date (date)                   : NEW
+project      / project.project          / date_start (date)             : NEW
+project      / project.project          / label_tasks (char)            : NEW
+# New fields ---> Nothing to do
+
+project      / project.project          / members (many2many)           : DEL relation: res.users
+# Field removed ---> Nothing to do
```
dreispt: Project team is a relevant information; OCA should provide a module to keep this info.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
MicrocomAdmin: Agreed.

ghost: Agreed. Will fix it right now!

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
ghost: @pedrobaeza I will do that.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
ghost: Should we rename this column to None then, like @pedrobaeza suggested for column reviewer_id?

legalsylvain: Indeed, it is the easier way to keep this data.
new OCA module should have a migration folder to recover this data.

ghost: Acknowledged! Will do just that in a minute, then.

addons/project/migrations/9.0.1.1/post-migration.py
ghost: I renamed it.

addons/project/migrations/9.0.1.1/post-migration.py
ghost, line 27: Are you trying to say it should be done in pre-migration instead?

I don't understand why it's not needed here since I'm converting that field to html. Please provide a better solution if you know one, thanks!

addons/project/migrations/9.0.1.1/pre-migration.py
pedrobaeza, line 63: OK, I was just asking because I didn't know this method, but it seems correct.

addons/project/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,58 @@
...
+        [('template', 'draft')],
+        table='project_project', write='sql')
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    map_priority(cr)
+    map_template_state(cr)
+    for table_name in column_copies.keys():
+        for (old, new, field_type) in column_copies[table_name]:
```
pedrobaeza, line 67: Here you can save the loop if you want, as there is only one table and only one field with this case, and makes it more readable.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
mmalorni: We actually have a module that extends the stars to show 5 instead of 3. We might publish it for those that want to keep the 3 star priority system. We need to preserve the values in this field though. Should we build a module that modifies the upgrade script or should we store the values in another field in order to restore them after the upgrade?

pedrobaeza:
About the 5-start special-module, your technique should be:
- project pre-migration script: rename the column.
- project post-migration script: map to have the 3-start value.
- your_module post-migration script (including it in the addons path of your OpenUpgrade deployment): map with the same source column the 5-start value.

addons/project/migrations/9.0.1.1/post-migration.py
MicrocomAdmin: Agreed. I was and still am waiting to see if we'll introduce other column_copies before doing anything here.

ghost, line 67: Agreed. I was and still am waiting to see if we'll introduce other column_copies before doing anything here.

addons/project/migrations/9.0.1.1/pre-migration.py
ghost, line 63: :+1:

ghost:
Hi guys, I get these warnings when I'm running my Project migration script:

```
2015-12-10 16:21:20,384 8358 WARNING v8-proj2_migrated openerp.models.schema: Table `project_config_settings`: column `module_sale_service` has changed type (DB=bool, def=selection), data moved to column `module_sale_service_moved0`
2015-12-10 16:21:20,386 8358 WARNING v8-proj2_migrated openerp.models.schema: Table `project_config_settings`: column `module_pad` has changed type (DB=bool, def=selection), data moved to column `module_pad_moved0`
2015-12-10 16:21:20,389 8358 WARNING v8-proj2_migrated openerp.models.schema: Table `project_config_settings`: column `module_project_issue_sheet` has changed type (DB=bool, def=selection), data moved to column `module_project_issue_sheet_moved0`
2015-12-10 16:21:20,393 8358 WARNING v8-proj2_migrated openerp.models.schema: Table `project_config_settings`: column `group_time_work_estimation_tasks` has changed type (DB=bool, def=selection), data moved to column `group_time_work_estimation_tasks_moved0`
2015-12-10 16:21:20,394 8358 WARNING v8-proj2_migrated openerp.models.schema: Table `project_config_settings`: column `module_project_timesheet` has changed type (DB=bool, def=selection), data moved to column `module_project_timesheet_moved0`
```

Since this table is temporary ("transient"), is there an "OpenUpgrade approved way" of getting rid of these warnings? Should I even bother?

pedrobaeza:
Are that tables from transient models? In that case, you can just simply remove then in the pre-script if you want to avoid that warnings. If there are from regular models, you can rename then on pre and make any treatment you need on post-script.

ghost:
@pedrobaeza Thanks for your suggestion! That's what I did, I modified the pre-migration script so that the tables are removed outright.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
...
+
+project      / project.task             / priority (selection)          : selection_keys is now '['0', '1']' ('['0', '1', '2']')
+# Done ---> '2' is converted to '1' (converts only 2 to 1: the logic is starred vs not starred. two stars means starred)
+# (legend: in v8 = [('0','Low'), ('1','Normal'), ('2','High')] and in v9 = ('0','Normal'), ('1','High'))
+
+project      / project.task             / reviewer_id (many2one)        : DEL relation: res.users
+# Done ---> Field removed, no more task reviewers, so renamed column to None to allow recovery with a future OCA module
+
+project      / project.task             / website_message_ids (one2many): NEW relation: mail.message
+# Field doesn't exist neither in v8 or in v9...weird! ---> Nothing to do
```
StefanRijnhart, line 66: It is a new property of mail.thread, so shows up in every model that inherits this model. But nothing to do, indeed.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
...
+project      / project.project          / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project      / project.project          / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+project      / project.project          / state (selection)             : selection_keys is now '['cancelled', 'close', 'draft', 'open', 'pending']' ('['cancelled', 'close', 'draft', 'open', 'pending', 'template']')
+# Done ---> 'template' is converted to 'draft'
+# TODO - does this behavior make sense on the functional level?
+
+project      / project.project          / user_id (many2one)            : NEW relation: res.users
+# TODO - does someone know where user_id comes from? (probably related to Mail module)
```
StefanRijnhart: user_id is defined in project.py around 168. Meanwhile, it has been removed from account.analytic.account, so it needs to be renamed in that module and then copied into the project_project table. If I am thinking clear at this moment, it seems that in addition to the existing _move_field_m2o_ API method we could use a new "move_field_o2m" method, or you could write a short SQL query instead.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
+---Fields in module 'project'---
+project      / project.project          / date (date)                   : NEW
+project      / project.project          / date_start (date)             : NEW
+project      / project.project          / label_tasks (char)            : NEW
+# New fields ---> Nothing to do
+
+project      / project.project          / members (many2many)           : DEL relation: res.users
+# Done ---> Field removed, no more project team, so renamed column to None to allow recovery with a future OCA module
```
StefanRijnhart, line 8: Or check if team members are followers? https://github.com/odoo/odoo/issues/9030

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
...
+NEW project.task.type: project.project_stage_data_2
+DEL project.task.type: project.project_tt_analysis
+DEL project.task.type: project.project_tt_cancel
+DEL project.task.type: project.project_tt_deployment
+DEL project.task.type: project.project_tt_design
+DEL project.task.type: project.project_tt_development
+DEL project.task.type: project.project_tt_merge
+DEL project.task.type: project.project_tt_specification
+DEL project.task.type: project.project_tt_testing
+# TODO - project and task related stuff - analysis needed
```
StefanRijnhart, line 177: Default task entries apparently are no longer included in the module's data. In 8.0, they are added with noupdate=1, so they will not be automatically removed after the migration of the module -> Nothing to do. Task data entries together compose a sample project with some Odoo UI hints -> Nothing to do.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
...
+
+DEL res.groups: project.group_delegate_task
+# TODO - res.groups - has it moved in another module? - analysis needed
```
StefanRijnhart, line 180: Delegation has been 'simplified' in Odoo 9.0 ;-) https://www.odoo.com/groups/community-59/community-14009273 -> Nothing to do.

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
+---Fields in module 'project'---
+project      / project.project          / date (date)                   : NEW
+project      / project.project          / date_start (date)             : NEW
```
StefanRijnhart, line 3: Move these two fields from analytic account, like user_id

addons/project/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,187 @@
...
+# Done ---> Field removed, no more project team, so renamed column to None to allow recovery with a future OCA module
+
+project      / project.project          / message_follower_ids (many2many): not a function anymore
+project      / project.project          / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project      / project.project          / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+project      / project.project          / state (selection)             : selection_keys is now '['cancelled', 'close', 'draft', 'open', 'pending']' ('['cancelled', 'close', 'draft', 'open', 'pending', 'template']')
+# Done ---> 'template' is converted to 'draft'
+# TODO - does this behavior make sense on the functional level?
```
StefanRijnhart, line 17: Looks like template functionality has been removed. Set existing templates to active=False and/or state draft indeed.

StefanRijnhart:
@mel-microcom. Thank you so far. I just added some comments that I thought might be helpful.

mmalorni:
Hi, @StefanRijnhart , @pedrobaeza , 
mel-microcom has finished his internship with us and will probably not continue on the module. We have @coleste could take over. What is the best way to transfer the ownership. We own the branch on which the PR was done on our side. We might be able to gain access to his github user as well since it was dedicated to internal work. tks.

mmalorni:
Hi, @StefanRijnhart , @pedrobaeza ,
@coleste will take over with the project module and there should not be any issues. Should continue smoothly.

JordiBForgeFlow:
@coleste We'd be interested to take over with the  migration, if you're not working on this.

coleste:
@jbeficent please do, this was sufficient for our migration.

dreispt:
@jbeficent Can you open a new PR?

PR #407, opened 2015-11-25
[FIX] remove global bzrlib dependencies

coleste:
Since I don't have bzrlib installed, could someone validate from that side please?

### DIFF
--- a/scripts/migrate.py
+++ b/scripts/migrate.py
@@ -263,8 +263,6 @@ for version in options.migrations.split(','):
             version,
             ','.join(sorted([a for a in migrations])))
 
-bzrlib.plugin.load_plugins()
-bzrlib.trace.enable_default_logging()
 logfile = os.path.join(options.branch_dir, 'migration.log')
 
 if not os.path.exists(options.branch_dir):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

pedrobaeza:
Another option is to protect the import.

coleste:
You mean making the calls right after import bzrlib.builtins?

pedrobaeza:
No, putting `try:` ... `except ImportError:` block: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror

coleste:
Ok, but we don't want it to fail since it also can work with git. Bzr is not necessary.

Wait, ok - protect when we get in the bzr case itself...

pedrobaeza:
With that, the script doesn't fail. As bzr is marginally, the solution you provide is good enough also.

coleste:
yeah, all the migrations use git, so addon_config_type 'bzr' is a dead case now.

legalsylvain:
:+1: thanks.

mmalorni:
:+1:

PR #411, opened 2015-11-30
9.0 project issue

pedrobaeza:


addons/project_issue/migrations/9.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,35 @@
...
+#
+##############################################################################
+
+from openupgradelib import openupgrade
+
+
+# Column is kept even though it was removed in v9, as to allow recovery by a future OCA module
+column_renames = {
+    'project.issue': [
+        ('categ_ids', 'tag_ids'),
```
pedrobaeza, line 26: This is not enough. You have to convert old categories to new project tags.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
+---Fields in module 'project_issue'---
+project_issue / project.issue            / categ_ids (many2many)         : DEL relation: project.category
+project_issue / project.issue            / tag_ids (many2many)           : NEW relation: project.tags
+# TODO DONE
+
+project_issue / project.issue            / kanban_state (selection)      : now required, default = normal
+# Nothing to do
```
pedrobaeza, line 7: You have to call set_defaults to avoid possible empty values in the 8.0 DB: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L463

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+
+project_issue / project.issue            / message_follower_ids (many2many): not a function anymore
+project_issue / project.issue            / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project_issue / project.issue            / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
```
pedrobaeza: rephrase as "Done in mail migration script", so that this sentence serves in the future.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+
+project_issue / project.issue            / progress (float)              : module is now 'project_issue_sheet' ('project_issue')
+# Nothing to do
+
+project_issue / project.issue            / section_id (many2one)         : relation is now 'crm.team' ('crm.case.section')
+project_issue / project.issue            / section_id (many2one)         : was renamed to team_id [nothing to to]
+# Nothing to do
+
+project_issue / project.issue            / version_id (many2one)         : DEL relation: project.issue.version
+# Feature removed ---> Nothing to do
```
pedrobaeza, line 22: Rename it to `None` in case anyone wants to restore this field in a module.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+project_issue / project.issue            / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project_issue / project.issue            / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+project_issue / project.issue            / progress (float)              : module is now 'project_issue_sheet' ('project_issue')
+# Nothing to do
+
+project_issue / project.issue            / section_id (many2one)         : relation is now 'crm.team' ('crm.case.section')
+project_issue / project.issue            / section_id (many2one)         : was renamed to team_id [nothing to to]
+# Nothing to do
```
pedrobaeza, line 19: Here you have to convert sections to the new model.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+project_issue / project.issue            / message_follower_ids (many2many): not a function anymore
+project_issue / project.issue            / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+project_issue / project.issue            / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+project_issue / project.issue            / progress (float)              : module is now 'project_issue_sheet' ('project_issue')
+# Nothing to do
+
+project_issue / project.issue            / section_id (many2one)         : relation is now 'crm.team' ('crm.case.section')
+project_issue / project.issue            / section_id (many2one)         : was renamed to team_id [nothing to to]
```
pedrobaeza: Remove the sentence between brackets. In the analysis file I don't see that it was renamed. The line says that relation has changed.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+# Feature removed ---> Nothing to do
+
+project_issue / project.issue            / website_message_ids (one2many): NEW relation: mail.message
+# Field doesn't exist neither in v8 or in v9 (probably related to Mail module) ---> Nothing to do
+
+project_issue / project.project          / label_issues (char)           : NEW
+# Nothing to do
+
+project_issue / project.project          / project_escalation_id (many2one): DEL relation: project.project
+# Nothing to do
```
pedrobaeza, line 31: Also rename it to None for not losing it.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+
+NEW mail.message.subtype: project_issue.mt_project_issue_ready
+DEL mail.message.subtype: project_issue.mt_issue_assigned
+DEL mail.message.subtype: project_issue.mt_project_issue_assigned
+# Related to mail module changes ---> Nothing to do
+
+DEL project.issue.version: project_issue.type1
+DEL project.issue.version: project_issue.type2
+DEL project.issue.version: project_issue.type3
+# Feature removed ---> Nothing to do
```
pedrobaeza, line 70: If we preserve version_id, we should preserve this also, but I'm not sure if at the end it deserves to be preserved or is a dead feature.

pedrobaeza:
@vincentcoll, I have made this PR for you.

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
dreispt, line 70: IMO keep the db column just in case, but drop the XML records.

pedrobaeza, line 70: But if we drop the XML records, the cascade="set null" option will erase this column.

dreispt, line 70: I thought those were `ir.model.data` records -> just XML Id to Database Id translations.

pedrobaeza, line 70: I'm not sure if the records are dropped as well. @StefanRijnhart, can you clarify us this question?

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
vincentcoll, line 22: What do I need to rename to None?
This line: +# Feature removed ---> Nothing to do 
OR
These: +# Nothing to do

legalsylvain, line 22: the feature is removed, so there is nothing to do -> ok.
but maybe the feature was interesting, and will be redevelopped in an OCA / custom modules. 
In that case, rename the field by None will rename the column with a prefix / suffix like '__openupgrade__V8_V9' and so the data will not be lost and could be reused by a migration script of the future OCA / custom modules.

regards.

addons/project_issue/migrations/9.0.1.0/pre-migration.py
vincentcoll, line 26: @pedrobaeza Can't we just rename the table project_category_project_issue_rel to project_issue_project)tags_rel and its column names and update the res_id names?

vincentcoll:
@pedrobaeza For the conversion of categ_ids to tag_ids can't we just rename the table project_category_project_issue_rel to project_issue_project)tags_rel and its column names and update the res_id names?

pedrobaeza:
I don't think so, because the following reasons:
- The new table can contain already records.
- The sequence for the ID will be desynchronized.
- The new table can contain different columns from the old one
- other unexpected side effects

so the solution is to re-create these records in the new model (via ORM or SQL), filling/transforming the possible differences between both.

vincentcoll:
@pedrobaeza To save some time, can you point an example in modules in which a relation changed . Thanks

pedrobaeza:
This one is by far more complicated, but the idea is the same: create some records in a new table and link them in the new field.

https://github.com/OCA/OpenUpgrade/blob/8.0/addons/calendar/migrations/8.0.1.0/post-migration.py#L26-L114

vincentcoll:
@pedrobaeza Thank you for the example. Since the functional part of this module is exceeding my knowledge of Odoo, can you brief me on what changed exactly for this relation 'cause I don't know what I need to transform or if I can just migrate the data without any transformation. Regards,

addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
StefanRijnhart, line 70: Sorry, missed your mentioning at the time. To answer your question, the Odoo module loading mechanism will delete records if they had a (noupdate=False) record in ir_model_data before, which has been removed in the new version. But if you remove these references in the pre-script (or if you set noupdate to True), then the data should be preserved.

coleste:
FYI, this was enough for our migration to work. We're still open to cleanup and fix our PRs.

pedrobaeza:
cc @Tecnativa

PR #412, opened 2015-12-02
[UPD] merge upstream 2015 12 02

hbrunn:
We need this to support https://github.com/OCA/OCB/commit/8e1a5add38ddc6d19a38f7eff7c5645de4a4b1a4 which makes modules  incompatible. Everyone should know about this commit.

The same probably needs to be done for the 9.0 branch.


pedrobaeza:
I see no adjustment has been needed to be made, so :+1:

legalsylvain:
Some pep8 errors in some commits, and code coverage decreased...

joking aside, thanks :+1:

PR #418, opened 2015-12-11
[WIP] 9.0 product

ghost:
Very early analysis and migration scripts...lots of work left - help appreciated.

(sorry for the last identical merge request - wrong account)

### DIFF
--- a/addons/product/migrations/9.0.1.2/openupgrade_analysis_work.txt
+++ b/addons/product/migrations/9.0.1.2/openupgrade_analysis_work.txt
@@ -12,20 +12,26 @@ product      / product.packaging        / ul (many2one)                 : DEL re
 product      / product.packaging        / ul_container (many2one)       : DEL relation: product.ul
 product      / product.packaging        / ul_qty (integer)              : DEL 
 product      / product.packaging        / weight (float)                : DEL 
-# Seems almost all packaging's fields were removed, they can't seem to be found anywhere else ---> Nothing to do
+# Seems like almost all packaging's fields were removed, they can't seem to be found anywhere else ---> Nothing to do
 
 product      / product.price.history    / product_id (many2one)         : NEW relation: product.product, required: required
 product      / product.price.history    / product_template_id (many2one): DEL relation: product.template, required: required
 # Done ---> product_template_id converted to product_id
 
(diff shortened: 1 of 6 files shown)


ghost:
Hi guys,

Product is not an easy one to migrate! 

Help is needed in order to finish the analysis I started, especially on the functional level, as I am not very knowledgeable on that matter...

Thanks!

StefanRijnhart:
Can you ask in more concrete questions what you need?

StefanRijnhart:
I have added my insights in the analysis work file, proposing into your branch here: https://github.com/microcom/OpenUpgrade/pull/3/files. Hope that helps.

addons/product/migrations/9.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,183 @@
...
+# New field ---> Nothing to do
+
+product      / product.packaging        / code (char)                   : DEL
+product      / product.packaging        / ean (char)                    : DEL 
+product      / product.packaging        / name (text)                   : now required
+product      / product.packaging        / name (text)                   : type is now 'char' ('text')
+product      / product.packaging        / rows (integer)                : DEL required: required, req_default: 3
+product      / product.packaging        / ul (many2one)                 : DEL relation: product.ul, required: required, req_default: function
+product      / product.packaging        / ul_container (many2one)       : DEL relation: product.ul
+# TODO: remove NOT NULL constraint on these obsolete fields
```
JordiBForgeFlow, line 14: @StefanRijnhart Can you help me to clarify whether a specific action needs to take place to remove NOT NULL constraint on these fields? I have not seen any prior example in OpenUpgrate where a sentence such as ALTER TABLE table ALTER COLUMN column DROP NOT NULL would be used.

StefanRijnhart, line 14: OpenUpgrade, unlike Odoo does not drop obsolete columns at the end of a module upgrade. That means we have to remove these constraints to prevent problems when a new product is added in the migrated production environment (before using something like database_cleanup). You are right, I don't remember this coming up before.

We can probably do this globally in the exact same place where Odoo drops the column instead of on a per module basis.

coleste:
FYI, this was enough for our migration to work. We're still open to cleanup and fix our PRs.

PR #425, opened 2015-12-28
Stock quants' history_ids is not generated in a correct manner

andreiboyanov:
In Odoo 7 and earlier Serial Number Traceability is assured trough the direct relation `stock.production.lot` -> `stock.move`s. In Odoo 8 the relation is more complex - `stock.production.lot` -> `stock.quant` -> `stock.move`s. The last one is pointed by the `stock.quant`'s field `history_ids`.

Technically speaking - in Odoo 7 there is a many2one relation between stock_move and stock_production_lot. In Odoo 8 there is a many2one relation between `stock.quant` and `stock.production.lot` and one2many relation between `stock.quant' and`stock.move`

With the base migration scripts I have the following problem. 

Before the migration: 

```
Production Lot 1 -> Stock Move A
Production Lot 1 -> Stock Move B
Production Lot 2 -> Stock Move C
Production Lot 2 -> Stock Move D
```

After the migration these relations are wrong. Could be something like that:

```
Production Lot 1 -> Stock Quant A -> Stock Moves A & M
Production Lot 2 -> Stock Quant B -> Stock Moves C & N
```

I fixed this problem setting the stock_move's field `restrict_lot_id` before the process of quants creation.

Additionally you'll find an optimisation of the sql query in the method `migrate_stock_picking(cr, registry)` of the `post-migration.py` script

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -214,8 +214,6 @@ def migrate_stock_picking(cr, registry):
         ADD COLUMN %s INTEGER
         """ % openupgrade.get_legacy_name('move_id'))
     # Recreate stock.pack.operation (only for moves that belongs to a picking)
-    if not done_move_ids:
-        return
     openupgrade.logged_query(
         cr,
         """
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Hi, Andrei, thanks for your contribution. You should sign the CLA to allow your contribution gets to OCA: https://odoo-community.org/page/cla

Serial number migration is covered in this method: https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L710-L749, but your method seems cleaner. Can you please remove the method then to clean the code?

andreiboyanov:
Hi Pedro,

As I remember I signed my CLA right after the message of oca-clabot... How can I check that?

The method mentioned by you (https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L710-L749) is about lots which are not related to pickings. Mine was related with problems I had pickings related lots. 

If you confirm me that the migrate_stock_production_lot() method is not needed I will remove it. If you want I can test that myself...

StefanRijnhart:
I checked the CLA on https://odoo-community.org, and it's fine.

StefanRijnhart:
Thank you for your clear analysis and the fix. It looks like the fix is unrelated to migrate_stock_production_lot indeed.  :+1:

pedrobaeza:
Sorry, you're right, so :+1: by my part.

PR #436, opened 2016-02-14
payment.mode.type error

thomaspaulb:
Had an error with payment.mode.type model not existing, am not smart enough to produce a better fix than this but it worked


pedrobaeza:
This is not valid, because you need _account_payment_extension_ migration scripts. They're in this PR: https://github.com/OCA/account-payment/pull/35

PR #437, opened 2016-02-16
[IMP] procurement: Use SQL for post messages

hbrunn:
my first approach was nonsense. This is a rewrite to use sql instead of `message_post`. My current database has ~80000 procurements with messages/notes, it would take hours to do this via the ORM, and given we don't need most of the things `message_post` does, I think this suffices

### DIFF
--- a/addons/procurement/migrations/8.0.1.0/post-migration.py
+++ b/addons/procurement/migrations/8.0.1.0/post-migration.py
@@ -26,16 +26,29 @@ from openerp import pooler, SUPERUSER_ID
 def post_messages(cr, pool):
     """ The obsolete message and note fields on procurements are replaced
     by posting messages on the chatter. Posting existing messages here."""
-    procurement_obj = pool['procurement.order']
-    for field in ('note', 'message'):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I have changed the PR title to go accordingly. The body field should be in HTML, so please enclose the field into a `<p>` tag.

hbrunn:
thanks! For good measure, I also escape `<` now

pedrobaeza:
Should we scape also other characters (accentuated for example)? Isn't a method to make it?

hbrunn:
No sql method I'm aware of. But given Odoo is delivered as utf8, escaping shouldn't be necessary.

But `&apos;` still has to be escaped

StefanRijnhart:
I can see why you'd want this :+1:

hbrunn:
@pedrobaeza with & and < escaped by now, we should be safe html wise. Also rebased

pedrobaeza:
:+1:

legalsylvain:
thanks. :+1:

PR #438, opened 2016-02-17
[ADD] disable workflow transitions for stock.picking

hbrunn:
while replaying stock moves

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -21,7 +21,7 @@
 ##############################################################################
 
 import logging
-from openerp import api, SUPERUSER_ID
+from openerp import api, models, SUPERUSER_ID
 from openerp.openupgrade import openupgrade, openupgrade_80
 from openerp.modules.registry import RegistryManager
 from openerp import SUPERUSER_ID as uid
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1: We will gain speed with this

hbrunn:
@pedrobaeza I'm not so sure if this is a speed improvement, as the workflow engine will be pulled anyways, only all transition conditions evaluate to `False`. In my case, `stock.picking` has transition conditions set which can't be evaluated at this point, that's why I propose to deactivate them.
By assumption the pickings are already in the correct state, so this shouldn't be harmful.

Unfortunately, there's no way I'm aware of to just disable the workflow engine completely, in a context key or something.

pedrobaeza:
Uhm, I understand. Anyway, it's harmless as you say.

hbrunn:
so given my ~200.000 stock moves take hours to migrate, I went for a radical disabling of the workflow. For OpenUpgrade, I think this is okay, because there's no multithreading and no multi database issues. This way, it's a few hours less. What do you think?

pedrobaeza:
Well, you can do it if only for certain operations like this one, but yes, it's a valid approach for OpenUpgrade I think.

StefanRijnhart:
It's like you say, the goal of this method is to fill quants. Any side effects from rerunning the workflow may even be harmful. :+1:

PR #439, opened 2016-02-17
[IMP] Make the script fault tolerant when run on a DB missing the con…

pablocm-aserti:
…straint due to duplicates on meeting_category_rel

### DIFF
--- a/addons/calendar/migrations/8.0.1.0/post-migration.py
+++ b/addons/calendar/migrations/8.0.1.0/post-migration.py
@@ -79,12 +79,18 @@ def import_crm_meeting(cr):
     cr.execute(
         '''alter table meeting_category_rel
         drop constraint meeting_category_rel_event_id_fkey''')
-    # we need to disable this constraint temporarily because some new event_id
-    # might collide during the update with the old event_id of another row
-    # to be updated yet
     cr.execute(
-        '''alter table meeting_category_rel
-        drop constraint meeting_category_rel_event_id_type_id_key''')
+        '''select * from information_schema.table_constraints
+        where table_name = 'meeting_category_rel' and
+        constraint_name = \'meeting_category_rel_event_id_type_id_key\'''')
+    constraint_enabled = cr.fetchall()
+    if constraint_enabled:
+        # we need to disable this constraint temporarily because some new event_id
+        # might collide during the update with the old event_id of another row
+        # to be updated yet
+        cr.execute(
+            '''alter table meeting_category_rel
+            drop constraint meeting_category_rel_event_id_type_id_key''')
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Thanks for the improvement

:+1:

addons/calendar/migrations/8.0.1.0/post-migration.py
```diff
@@ -92,11 +98,12 @@ def import_crm_meeting(cr):
             openupgrade.get_legacy_name('crm_meeting_id'),
         )
     )
-    # restore the disabled unique constraint
-    cr.execute(
-        '''alter table meeting_category_rel
-        add constraint meeting_category_rel_event_id_type_id_key
-        UNIQUE(event_id, type_id)''')
+    if constraint_enabled:
```
StefanRijnhart, line 101: Can you try using

```
    with cr.rollback():
        cr.execute(...)
```

This should raise a warning when the constraint cannot be put in place (as a useful reminder to the person running the migration), while keeping the cursor alive.

pablocm-aserti, line 101: I avoided that solution on purpose because for me, it was causing a postgresql internal error. Not on the rollback itself, but the next query you tried to run, that despite it beign perfectly valid, it raised that excepcion causing the rest of the commands of that transaction to be ignored.

StefanRijnhart, line 101: Just to be sure, you tested this and in spite of the rollback, the cursor was still broken?

pablocm-aserti, line 101: It wasn't here, and now that I've checked it, not with

```
with cr.rollback():
        cr.execute(...)
```

but [here](https://github.com/OCA/l10n-spain/blob/deb2fb19364e584cc9162449c457f88f46d4d498/l10n_es_partner/migrations/8.0.1.3/pre-migration.py#L43), which is slightly different, but relies too in rollbacks.
The delete query in that line caused the internal error, while executing it manually before the rollback was fine.

StefanRijnhart, line 101: Sorry, I meant

```
    with cr.savepoint():
        cr.execute(...)
```

legalsylvain:
what is the state of this PR ?

StefanRijnhart:
Merging this. The question was about a trivial improvement.

StefanRijnhart:
Oh wait, it has conflicts. Let me have a look.

StefanRijnhart:
Obsoleted by https://github.com/OCA/OpenUpgrade/pull/452

PR #440, opened 2016-02-18
[ADD] also product uom's decimal precision was renamed

hbrunn:
Especially with this dp, we run into problems without this, as it is used in bom lines' product_qty which have a constraint that enforces product_qty > 0. Without this, we get the default precision (2 digits), which breaks if we have smaller values as they will be cast to 0.

Relevant code:
https://github.com/odoo/odoo/blob/6.1/addons/product/product_data.xml#L163
https://github.com/odoo/odoo/blob/7.0/addons/product/product_data.xml#L51


pedrobaeza:
:+1:

StefanRijnhart:
:+1:

PR #446, opened 2016-03-02
[FIX] don't choke if we have custom states and there are no records with state picking_except

hbrunn:
In this case, production orders have a custom state added, so `check_values_selection_field` fails, and given there are no production orders in state `picking_except`, migration fails on the update call.

I think it's not a good idea anyways to check for all known states, so I propose to remove the call to `check_values_selection_field` altogether and only check if we have ids.

Then what should be changed independently from accepting the above or not is copying the id list to a tuple several times.

### DIFF
--- a/addons/mrp/migrations/8.0.1.1/pre-migration.py
+++ b/addons/mrp/migrations/8.0.1.1/pre-migration.py
@@ -52,24 +52,12 @@ def check_production_state(cr):
     """Check if a record with a state that is no longer supported
     (picking_except) exists in mrp_production and adjust it (to draft).
     """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
OK, seems reasonable.

:+1:

hbrunn:
@pedrobaeza thanks, for your fast review. I changed the proposal, was we can simply run one statement to get what we want

pedrobaeza:
Yeah, even better, but about the removal of the check, it still seems reasonable :wink:

StefanRijnhart:
:+1:

PR #449, opened 2016-03-16
[ADD] crm_claim: Migration scripts

pedrobaeza:


JordiBForgeFlow:
:+1

addons/crm_claim/migrations/8.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,27 @@
...
+        <field name="sequence">1</field>
+    </record>
+    <record id="stage_claim3" model="crm.claim.stage">
+      <field name="fold"/>
+      <field name="case_refused"/>
+      <field name="state"/>
+    </record>
+    <record id="stage_claim2" model="crm.claim.stage">
+      <field name="state"/>
+    </record>
```
StefanRijnhart: state, fold, case_refused fields are actually obsolete fields, for which the values were removed from the noupdate data. Sorry the compare script is not that smart! I would remove them, because loading these entries will give warnings about unknown fields being written to these stages.

addons/crm_claim/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
+    uid = SUPERUSER_ID
+    openupgrade_80.set_message_last_post(cr, uid, pool, ['crm.claim'])
+    openupgrade.map_values(
+        cr, 'priority', openupgrade.get_legacy_name('priority'),
+        [('1', '0'), ('2', '1'), ('3', '2'), ('4', '2')], table='crm_claim')
```
StefanRijnhart: Mapping for original value '5' missing.

pedrobaeza:
@StefanRijnhart, fixed!

Thanks.

StefanRijnhart:
Thanks! :+1:

PR #450, opened 2016-03-16
[ADD] crm_helpdesk: Migration scripts

pedrobaeza:


JordiBForgeFlow:
:+1:

addons/crm_helpdesk/migrations/8.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    pool = pooler.get_pool(cr.dbname)
+    uid = SUPERUSER_ID
+    openupgrade_80.set_message_last_post(cr, uid, pool, ['crm.helpdesk'])
+    openupgrade.map_values(
+        cr, 'priority', openupgrade.get_legacy_name('priority'),
+        [('1', '0'), ('2', '1'), ('3', '2'), ('4', '2')], table='crm_helpdesk')
```
StefanRijnhart: not mapping value '5'?

pedrobaeza: I don't know how I saw both old and new values, but I did a totally wrong mapping. Now it should be good ('2' value remains unaltered, so that's why I don't map it).

StefanRijnhart:
:+1:

PR #452, opened 2016-03-16
[FIX] calendar: Don't halt on non existant constraint

pedrobaeza:
There can be occasions where the constraint doesn't exist, so with this, we prevent the hang on this error.


pedrobaeza:
cc @pablocm-aserti, @hbrunn, @StefanRijnhart

hbrunn:
`if exists` for constraints is new in 9.0 (postgres, not odoo) to my knowledge. I think it's okay to presuppose people have updated their server within the last 6 years, so :+1:

pablocm-aserti:
Much better this way, thanks! :+1:

JordiBForgeFlow:
:+1:

StefanRijnhart:
Thanks :+1:

PR #453, opened 2016-03-16
[IMP] account: Make partner_id consistent in invoices

pedrobaeza:
Both account.invoice and account.invoice.line contains a reference to the
partner_id. The partner_id field in the line is a related one, so it should
be automatically computed, but we have known through issues/not accepted
patches like https://github.com/odoo/odoo/pull/6774, that this doesn't
work correctly.

This desynchronization happens only on certain models, and
account.invoice.line is one of them. If the old linked record is removed,
there's also no restrict constraint, so we find a foreign key error
when updating.

This commit synchronize again all desynchronized records.

### DIFF
--- a/addons/account/migrations/8.0.1.1/pre-migration.py
+++ b/addons/account/migrations/8.0.1.1/pre-migration.py
@@ -1,25 +1,10 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Copyright (C) 2014 Akretion (http://www.akretion.com/)
-#    @author: Alexis de Lattre <alexis.delattre@akretion.com>
-#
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is harmless if no inconsistency found.

JordiBForgeFlow:
:+1: I agree. Related fields in previous problems would not recompute at all. When you changed the field in the relating model, the field in the related model is not updated. In v7 I've found it to happen in many models.

StefanRijnhart:
You're going to sneak this one into every versions aren't you ;-) https://github.com/OCA/OpenUpgrade/commit/a39029f3c

Fine by me :+1:

PR #454, opened 2016-03-17
[7.0][FIX] Rate for ZRZ violates decimal range for its field

mistotebe:
Same issue as https://github.com/odoo/odoo/pull/11383 (value of rate violates data model), affects upgrades for databases that do not have this record in ir.model.data.

### DIFF
--- a/openerp/addons/base/currency_data.xml
+++ b/openerp/addons/base/currency_data.xml
@@ -1413,11 +1413,6 @@
             <field name="rounding">0.01</field>
             <field name="accuracy">4</field>
         </record>
-        <record id="rateZRZ" model="res.currency.rate">
-            <field name="currency_id" ref="ZRZ" />
-            <field eval="time.strftime('%Y-01-01')" name="name"/>
-            <field name="rate">1148948.76</field>
-        </record>
         
         <record id="MZN" model="res.currency">
             <field name="name">MZN</field>


pedrobaeza:
This shouldn't be published here. If it's accepted in Odoo, we will merge all the new changes here, but not individual patches to the Odoo general files is accepted unless they are needed for OpenUpgrade and they can't be integrated on general Odoo.

mistotebe:
Odoo are not a very responsive bunch, especially for older releases. Also, OpenUpgrade does not seem to track OCB or Odoo at all, so this looks like the only way to get the change merged here.

How do upstream changes end up in this repo?

pedrobaeza:
By a manual merge operation proposed here as a PR. Check for example https://github.com/OCA/OpenUpgrade/pull/412, but we don't do it too much because for a general migration process is not usually needed.

mistotebe:
Yes, but this does not get fixed by anything else and databases without this xml_id will fail migration without manual interaction. And that's because of the actual data being broken - what this change proposes to fix.

mistotebe:
So again, how should I go about getting that change to land here?
- do I get Odoo to merge it and there it gets here how?
- OCA to merge it and then?
- some other way?

pedrobaeza:
I don't understand why it doesn't fail on regular DB creation, because that data is loaded by default, isn't it?

mistotebe:
That's the only thing I'm not sure about, unless it does but the initial load for `base` is less strict and works around/ignores that issue somehow.

mistotebe:
Oh, that setting was not honoured until later postgres versions, which explains why it's only an issue now.

pedrobaeza:
From what PG version? I remember to set a v7 DB recently in a 9.3 PostgreSQL version

mistotebe:
Hmm, yeah, that's not the case either. There is one database whose history goes back to Postgres 8.x and the rate is defined as a `numeric(12,6)` (which is what fails on this item) whereas even newest Odoo 9.0 with Postgres 9.5 does not set this and keeps it to `numeric` without restrictions. Maybe a psycopg with different behaviour?

pedrobaeza:
Definitively you should upgrade that PG version instead :wink:

mistotebe:
No, it's on 9.5 now, just that the database I mention in my comment was created with 8.x or older with OpenERP 5.0(?) and upgraded a few times since. And it's not the only one affected :)

It's possible we might have a change somewhere in the stack that will enforce the settings on the float field in the database, just never occurred to me to check since obviously it is the data that is at fault.

pedrobaeza:
OK, let's see what Odoo says in the PR, but this is out of the scope of this project.

mistotebe:
If the ORM/psycopg does not actually bother to set up the proper field definitions, then we might just keep it a local patch.

PR #455, opened 2016-03-17
[7.0][FIX] Was a reference to the wrong field

mistotebe:
There is a typo in the poweremail migration which fails the migration with a type mismatch.


pedrobaeza:
OK, I see, it should be `enforce_from_account` field, but it takes `use_sign` boolean field.

:+1: 

cc @StefanRijnhart

pedrobaeza:
If you rebase, you'll find a green branch.

mistotebe:
@pedrobaeza: it is as you said, thanks.

StefanRijnhart:
:+1: thanks

PR #456, opened 2016-03-17
[7.0][FIX] Merge user with existing partner if exists

mistotebe:
Databases might already hold partners for users, merge if that's the case.

### DIFF
--- a/openerp/addons/base/migrations/7.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/7.0.1.3/pre-migration.py
@@ -190,8 +190,14 @@ def create_users_partner(cr):
             "%s,%s,'no-message','no-message','no-message','no-message') "
             "RETURNING id")
     for row in cr.fetchall():
-        cr.execute(sql_insert_partner, row[1:])
-        partner_id = cr.fetchone()[0]
+        cr.execute(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Uhm, merge by name is not the best approach...

I remember @hbrunn made something for this.

hbrunn:
hmmm, not sure what you mean. Maybe #219?

But I also think this PR needs a lot more safeguards (match by email for example). And it should also exclude some special partners (base.main_partner for example).

The situation with duplicated partners is annoying, but also our 2k+ employees customer managed with the merge wizard. So every code that tries to match stuff must be super conservative in my opinion, and then I'm not sure how many matches will be found at all.

pedrobaeza:
That's it, @hbrunn. I remember something that involves users, but definitely is not the same as this.

For this, and having the possibility of merging partners with the OCA module, I would say :-1:

mistotebe:
This branch is really only here since we used it in a migration that we did and keep a branch for it for tracking purposes. I'm happy to make it more strict to check on email as well (if set on user) or if you're not inclined to accept it anyway, we'll just keep it as a local patch on top of OpenUpgrade.

StefanRijnhart:
Thanks for understanding! Agree with @pedrobaeza and @hbrunn above

PR #457, opened 2016-03-22
[8.0] mrp: Improvement on method

pedrobaeza:
This PR includes 2 commits:
- One for removing the renaming on the import on SUPERUSER_ID variable, which is a bad practice.
- Another for set the environment for new API, and method _update_stock_moves_ is rewritten with this new API. The reason is because I need to touch this method because it reads all the properties for production location, but as is a many2one reference field, it's possible that one of the records doesn't exist. With this code we prevent this possibility.


pedrobaeza:
cc @jbeficent

StefanRijnhart:
:+1:

hbrunn:
:+1:

PR #458, opened 2016-03-22
[FIX] Warings about using openupgradelib

m1k3lm:


addons/product/product.py
```diff
@@ -746,7 +746,10 @@ def _get_uom_id(self, cr, uid, *args):
     def _check_uom(self, cr, uid, ids, context=None):
         for supplier_info in self.browse(cr, uid, ids, context=context):
             if supplier_info.product_uom and supplier_info.product_uom.category_id.id <> supplier_info.product_id.uom_id.category_id.id:
-                return False
```
StefanRijnhart, line 749: Seems unrelated?

bin/addons/base/migrations/6.0.1.3/post-migration.py
```diff
@@ -302,7 +302,7 @@ def migrate(cr, version):
     pool = pooler.get_pool(cr.dbname)
     set_defaults_on_act_window(cr)
     set_defaults_on_ir_attachment(cr)
-    openupgrade.set_defaults(cr, pool, defaults)
+    openupgrade.set_defaults(cr, pool, defaults,False,True)
```
StefanRijnhart: Unrelated change?

bin/openupgrade/openupgrade_log.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-from openupgrade_tools import table_exists
+from openupgradelib import openupgrade_tools
```
StefanRijnhart: We don't need this higher level import, do we?

StefanRijnhart:
Thank you for your contributions to the project! Have you signed the OCA CLA, so that we can start adopting them?

bin/addons/base/migrations/6.0.1.3/post-migration.py
m1k3lm: Without the use_orm to True the process fails in with company_ids in res_user with:

"UPDATE res_users SET company_ids = ARRAY[1]"

Probably because the defaults function is not exactly the same it was when openupgrade class was not in openupgradelib. So it is related to the move from openupgrade class to opeenupgradelib.openupgrade.

StefanRijnhart: OK, good one! Can you keep pep8 spacing and use named arguments for these, e.g. 

```
openupgrade.set_defaults(cr, pool, defaults, force=False, use_orm=True)
```

addons/product/product.py
m1k3lm, line 749: The real problem here is in 

```
def _get_uom_id(self, cr, uid, *args):
    cr.execute('select id from product_uom order by id limit 1')
    res = cr.fetchone()
    return res and res[0] or False
```

That sets the same uom to all supplier_info and causes error if products have uom of different categories.  Didn't know how to fix _get_uom_id function so I needed to fix it here to have the migration done.

bin/openupgrade/openupgrade_log.py
m1k3lm: openupgrade_tools is present in bin/openupgrade this is needed to make sure the openupgrade_tools from openupgradelib is chosen.

bin/addons/base/migrations/6.0.1.3/post-migration.py
m1k3lm: done

m1k3lm:
Thank you very much for your review, @StefanRijnhart !

bin/openupgrade/openupgrade_log.py
StefanRijnhart: That is what the next line does, by referring to openupgradelib.openupgrade_tools explicitely. Line 2 above is void here, please remove.

addons/product/product.py
StefanRijnhart, line 749: OK, I think a better fix is possible and it's out of scope for this
change so please remove.

StefanRijnhart:
Thanks for the update. Still waiting for the CLA!

m1k3lm:
I've already sent the signed documento and got the response. What more should i do?

StefanRijnhart:
Indeed, I just checked your entry in the Odoo community database and I see that there was a typo in your Github handle, which is why the CLA-bot did not report back on this PR properly. All good now, thanks!

bin/openupgrade/openupgrade.py
```diff
@@ -32,4 +32,4 @@
 
 _modules = __import__(_new_name, globals(), locals(), ['*'])
 for _i in dir(_modules):
-    locals()[_i] = getattr(_modules, _i)
+    locals()[_i] = getattr(_modules, _i)
```
pedrobaeza: Why do you remove the last line here? Anyway, I think this is not needed anymore as you have replaced all the imports.

m1k3lm: Oh I see, might be I accidentally deleted the last newline char.

pedrobaeza:
:+1:

hbrunn:
taking https://github.com/OCA/OpenUpgrade/pull/458#issuecomment-204679867 as an implicit thumbs up, we have three: :+1:

m1k3lm:
Thanks all for your review!

PR #464, opened 2016-04-01
[ADD] documentation in 9.0 branch

hbrunn:


pedrobaeza:
Thanks for this work? :+1: 

Do you think we should remove the duplicated pages from old versions?

hbrunn:
I wrote the script with the idea in mind that it should be unnecessary to delete things, because the latest version of every file wins. After merging this I think we can move the module coverage files to the appropriate branch though

StefanRijnhart:
Thanks for working on this. Cloning every release is a bit slow. Did you consider downloading the raw coverage files from Github into the doc directory of 9.0 instead?

hbrunn:
The idea is that you clone only once, so I don't really think that's a problem. I have no problem changing the script to just fetch a couple of files from predefined urls after having cloned 9.0, if you really think we need this.

StefanRijnhart:
That is true. I was thinking about someone building the documentation locally but they are not going to use this script in the first place. :+1:

hbrunn:
cool! Now we use this on https://doc.therp.nl/openupgrade

StefanRijnhart:
Excellent, thanks for your work!

PR #466, opened 2016-04-04
[ENH] Add missing xml_id's for created stock.picking.type's.

NL66278:


pedrobaeza:
:+1:

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -483,6 +483,47 @@ def get_location_by_usage(usage):
...
+                logger.warn(
+                    "xml_id %s now points to res_id %d, no longer to %d.",
+                    xml_id, res_id, existing_id
+                )
+
+    model_data_obj = registry['ir.model.data']
+    main_warehouse = model_data_obj.search(
+        cr, uid, [('module', '=', 'stock'), ('name', '=', 'warehouse0')]
+    )
+    if not main_warehouse:
```
hbrunn: it's an error condition if this doesn't exist, as the record must be created during the update: https://github.com/OCA/OCB/blob/8.0/addons/stock/stock_data.xml#L147 (not not noupdate). So I think we can leave this check out

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -483,6 +483,47 @@ def get_location_by_usage(usage):
...
+        logger.error("Main warehouse not found in ir_model_data!")
+    else:
+        main_warehouse_id = main_warehouse[0].res_id
```
hbrunn: isn't `main_warehouse` a list of integers? (we're on v7 api)

hbrunn:
you can save most of your code by replacing the create calls with calls to ir.model.data#_update: https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_model.py#L989

hbrunn:
flake8 also has something to say:

```
./addons/stock/migrations/8.0.1.1/post-migration.py:511:37: F821 undefined name 'existing_id'
./addons/stock/migrations/8.0.1.1/post-migration.py:522:12: F821 undefined name 'warehouse_id'
```

addons/stock/migrations/8.0.1.1/post-migration.py
pedrobaeza: Or you can make it v8 API... with `with self.Environment.manage:...`

addons/stock/migrations/8.0.1.1/post-migration.py
NL66278: I had put it in by way of defensive programming, trying to pinpoint the real problem if the error occured. First I had an assert, but as most migration seems to be forgiving of errors, I changed it into a message.

addons/stock/migrations/8.0.1.1/post-migration.py
NL66278: Can you point to any example or documentation?

hbrunn: https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L691

NL66278: aha, I grepped all the code on self.Environment.manage, but it is api.Environment.manage. Thanx.

pedrobaeza: Sorry, I write it incorrectly. It's `api...` indeed

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -483,6 +483,50 @@ def get_location_by_usage(usage):
...
+                )
+
+    # TODO: environment might be used to put complete method in v8 api:
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        model_data_obj = env['ir.model.data']
+        main_warehouse = model_data_obj.search([
+            ('module', '=', 'stock'),
+            ('name', '=', 'warehouse0')
+        ], limit=1)
```
StefanRijnhart: replace with main_warehouse = env.ref('stock.warehouse0')

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -483,6 +483,50 @@ def get_location_by_usage(usage):
...
+        if not main_warehouse:
```
StefanRijnhart: This does not happen after an upgrade of the stock module

StefanRijnhart:
I don't understand the need for this code. The XMLIDs should be created by https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/stock_data.yml#L16-L18, which is run during the upgrade.

NL66278:
@StefanRijnhart The records are created, but the xmlid's are not created.

StefanRijnhart:
Strange. If this yaml data is loaded, then the references should be created. Have a look at a couple of lines down in the same file: https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/stock_data.yml#L23. Note that 'self' here refers to ir.model.data.

hbrunn:
look at https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/base/ir/ir_model.py#L977 - the records pointed to by the xmlids are only created in `post-migrate`, so `record` comes back empty (or it raises, but anyways) and the created xmlids are thrown away again. It _might_ be sufficient to call _update_dummy instead here.

StefanRijnhart:
Apparently not, because _update_dummy is called immediately afterwards in said yaml file.

hbrunn:
look again at the code I linked and reread my post

StefanRijnhart:
True. So we need this code, thanks for clarifying. Then that just leaves the nits.

hbrunn:
@NL66278 last nag from travis, then we can merge I think: https://travis-ci.org/OCA/OpenUpgrade/builds/122755423#L184

NL66278:
@hbrunn It seems travis is failing here. It complains about model_data_obj being assigned to, but not used. But it is used on the very next line to do the search!

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -483,6 +483,44 @@ def get_location_by_usage(usage):
...
+                model_data_record.write({'res_id': res_id})
+                logger.warn(
+                    "xml_id %s now points to res_id %d, no longer to %d.",
+                    xml_id, res_id, old_res_id
+                )
+
+    # TODO: environment might be used to put complete method in v8 api:
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        model_data_obj = env['ir.model.data']
```
pedrobaeza: You have to remove this one

NL66278: @pedrobaeza @hbrunn I see now, my own branch got out of synch with were I was looking. Commit done. Thanks.

hbrunn:
:+1:

StefanRijnhart:
Thanks for the update :+1:

hbrunn:
turns out you need the update mechanism I proposed in my first comment after all: XMLIDs not marked as touched will be deleted by the update mechanism, just creating the record is not enough. Call https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_model.py#L958 after creating you xml ids, that should suffice.

hbrunn:
@NL66278 can you add a fix?

NL66278:
Hello Holger and others,

I will try to add a fix.

Kind regards,

Ronald

Citeren Holger Brunn notifications@github.com:

PR #467, opened 2016-04-05
[WIP] 9.0 account

JordiBForgeFlow:
I have started to analyze what needs to migrate the accounting module. It is still WIP, but should be helpful as an assessment of the work to be done.

### DIFF
--- a/addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt
@@ -53,37 +53,56 @@ account      / account.analytic.line    / journal_id (many2one)         : DEL re
 account      / account.bank.statement   / account_id (many2one)         : module is now 'point_of_sale' ('account')
 # TODO: Will have to deal with it in the point_of_sale module
 account      / account.bank.statement   / cashbox_end_id (many2one)     : NEW relation: account.bank.statement.cashbox
+# TODO:
 account      / account.bank.statement   / cashbox_start_id (many2one)   : NEW relation: account.bank.statement.cashbox
+# TODO:
 account      / account.bank.statement   / closing_date (datetime)       : DEL 
 account      / account.bank.statement   / closing_details_ids (one2many): DEL relation: account.cashbox.line
 account      / account.bank.statement   / date_done (datetime)          : NEW 
+# TODO: post-migration script. Move from field 'closing_date'
 account      / account.bank.statement   / details_ids (one2many)        : DEL relation: account.cashbox.line
 account      / account.bank.statement   / message_follower_ids (many2many): not a function anymore
 account      / account.bank.statement   / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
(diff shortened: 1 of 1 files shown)


addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,935 @@
...
+# TODO: in post-migration, if the active field is False, set it to Trueg
+account      / account.account          / financial_report_ids (many2many): DEL relation: account.financial.report
+account      / account.account          / last_time_entries_checked (datetime): NEW 
+# The code in account module that uses this field is commented... So seems that it has no purpose (and no motivation to build a migration script)
+account      / account.account          / parent_id (many2one)          : DEL relation: account.account
+account      / account.account          / parent_left (integer)         : DEL 
+account      / account.account          / parent_right (integer)        : DEL 
+account      / account.account          / shortcut (char)               : DEL 
+account      / account.account          / tag_ids (many2many)           : NEW relation: account.account.tag
+# NOTHING TO DO: The migration of tags should be provided by the l10n account modules
```
pedrobaeza: Not really. Chart templates are one thing, and the other is the already installed plans. We should provide a generic method that converts parent accounts into tags, and tag leaf accounts with these tags.

JordiBForgeFlow: @pedrobaeza, but the module account now does not add any COA at all, not even the generic. The generic is provided by module l10n_generic_coa.

Now, a generic method that converts parent accounts into tags would be useful for one or more localizations, so I understand the justification for a generic method, that will not be called in the migration of account, but during the migration of the other modules.

pedrobaeza: You haven't understood me, Jordi. It has nothing to do with the generic CoA. I say generic method, not generic CoA. Think again about what I have said, and take into account that we provide a migration path, so we need to smooth the best we can the changes.

JordiBForgeFlow: @pedrobaeza I understand the need for this method, and agree with it. My point was that this method will be called by l10n account migration scripts, so that each time you migrate an account from 8 to 9 the tags are determined and populated, based on thd old hierachy.

pedrobaeza: No, it won't. We need a generic method that makes what I have said in account module. We can't pretend to create a specific method for each l10n_xx module. Even more, Odoo doesn't know (until v9) from which chart template you install from.

JordiBForgeFlow: Oh, ok. I understand now. Thanks for the feedback.

JordiBForgeFlow:
@hbrunn You might be interested in this.. Today we reviewed some parts with @pedrobaeza

hbrunn:
superseded by #488

PR #469, opened 2016-04-11
[IMP][STOCK] Improvements to the stock migration

JordiBForgeFlow:
This PR enhances the stock migration script in order to determine the correct picking_type_id in stock pickings, on a multi-warehouse setup.

The previous script was assigning to all pickings the picking type of the main warehouse.

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -25,6 +25,7 @@ from openerp import api, models, SUPERUSER_ID
 from openerp.openupgrade import openupgrade, openupgrade_80
 from openerp.modules.registry import RegistryManager
 from openerp import SUPERUSER_ID as uid
+from openerp.tools.float_utils import float_compare
 
 logger = logging.getLogger('OpenUpgrade.stock')
 default_spec = {
@@ -571,12 +572,19 @@ def _migrate_stock_warehouse(cr, registry, res_id):
 def migrate_stock_warehouses(cr, registry):
     """Migrate all the warehouses"""
     warehouse_obj = registry['stock.warehouse']
+    # Add a code to all warehouses that have no code
+    openupgrade.logged_query(
+        cr, """
+        UPDATE stock_warehouse SET code= 'WH' || to_char(id, 'FM999MI')
+        WHERE code IS NULL;
+        """)
+
     # Set code
     cr.execute("""select id, code from stock_warehouse order by id asc""")
     res = cr.fetchall()
-    for wh in res:
-        if not wh[1]:
-            warehouse_obj.write(cr, uid, wh[0], {'code': 'WH%s' % (wh[0])})
+    # for wh in res:
+    #     if not wh[1]:
+    #         warehouse_obj.write(cr, uid, wh[0], {'code': 'WH%s' % (wh[0])})
     # Migrate each warehouse
     for wh in res:
         _migrate_stock_warehouse(cr, registry, wh[0])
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As warehouse.view_location_id has been previously populated on `migrate_stock_warehouses` method, this code seems correct (a comment about this would be welcome in the code as a reminder). 

What I have doubts is about the correct population on this data. Can you check if it's correct in the current migration we are running?

StefanRijnhart:
Definitely an improvement! :+1:

JordiBForgeFlow:
@pedrobaeza @StefanRijnhart I am extending this PR with additional changes to the stock migration.
- Change strategy to add code to warehouses, so that it does not affect to the locations.
-  If a stock move is Waiting availability ('confirmed'), but the source location is 'supplier', 'inventory' or 'production', then set it as Available ('assigned').
- Changed the method to create quants, so that only the logic of quant creation is proposed. The previous approach of resetting the stock moves to draft was not correct in scenarios of perpetual inventory, because that would duplicate the accounting entries that the system creates when a stock move changes to 'done'. This change in the approach may also increase performance of the overall process

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -530,12 +531,19 @@ def get_location_by_usage(usage):
 def migrate_stock_warehouses(cr, registry):
     """Migrate all the warehouses"""
     warehouse_obj = registry['stock.warehouse']
+    # Add a code to all warehouses that have no code
+    openupgrade.logged_query(
+        cr, """
+        UPDATE stock_warehouse SET code= 'WH' || to_char(id, 'FM999MI')
```
pedrobaeza, line 577: Shouldn't it be a different code per warehouse?

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -708,6 +716,169 @@ def migrate_procurement_order(cr, registry):
             cr, uid, procurement_ids, {'warehouse_id': warehouse.id})
 
 
+def _move_assign(env, move):
```
pedrobaeza, line 759: This is the same code as in stock module. Why is going to be more efficient?

addons/stock/migrations/8.0.1.1/post-migration.py
StefanRijnhart, line 577: The _|| to_char..._ bit effectively appends the warehouse id to the code, so it is different per warehouse.

addons/stock/migrations/8.0.1.1/post-migration.py
```diff
@@ -502,12 +531,19 @@ def get_location_by_usage(usage):
...
+
     # Set code
     cr.execute("""select id, code from stock_warehouse order by id asc""")
     res = cr.fetchall()
-    for wh in res:
-        if not wh[1]:
-            warehouse_obj.write(cr, uid, wh[0], {'code': 'WH%s' % (wh[0])})
+    # for wh in res:
+    #     if not wh[1]:
+    #         warehouse_obj.write(cr, uid, wh[0], {'code': 'WH%s' % (wh[0])})
```
StefanRijnhart, line 586: You may remove these lines.

StefanRijnhart:
Did you try to call the method from the stock module directly? You should be able to filter out the correct base class from pool['stock.move'].__class__.__mro__, which is an iterable over the inherited classes. Take the class where __module__ starts with 'openerp.addons.stock.' (or just 'stock.', not sure). Then call action_done from this class with old api arguments.

StefanRijnhart:
Actually, you can just import the base class directly and call the method from there:

(Pdb) from openerp.addons.stock.stock import stock_move as stock_move_base
(Pdb) stock_move_base.action_done(self, cr, uid, ids, context=context)

JordiBForgeFlow:
@StefanRijnhart Actually they are not exactly the same code. The parts of the original code that would write to other data objects, such as stock moves and pickings, were explicity removed, so that the migration only updates quants.

addons/stock/migrations/8.0.1.1/post-migration.py
JordiBForgeFlow, line 759: @pedrobaeza @StefanRijnhart This change helped reduced processing time to 1/2. Can you please review from your end that the logic that has been added looks fine from your end?

StefanRijnhart:
OK, so we need to review more closely ;-). In the meantime, can you fix the lint errors in the migration script as reported by Travis?

StefanRijnhart:
Can you rebase to check the new tests?

StefanRijnhart:
Oh wait, we need to port the test to this branch first, I think.

hbrunn:
@StefanRijnhart working on it ;-)

hbrunn:
@jbeficent could you rebase this one too? Then we'll also get the tests for 8.0.

JordiBForgeFlow:
Done @hbrunn.

hbrunn:
yay, now we wait...

addons/stock/migrations/8.0.1.1/post-migration.py
kryskool, line 577: @StefanRijnhart 
Right, id field is used to generate unique code per warehouse
In the stock addons, i see that the code field is required, i think after this update we must alter the field to add NOT NULL like
ALTER TABLE stock_warehouse ALTER COLUMN code SET NOT NULL; 

regards,

StefanRijnhart:
The test failures seem to indicate that what are non fatal errors for OpenUpgrade still make the tests fail.

Oh, @jbeficent a merge of upstream branch is not a rebase. Try 'git rebase -i origin/8.0'

addons/stock/migrations/8.0.1.1/post-migration.py
StefanRijnhart, line 577: That happens automatically on the next upgrade (it is a good practice to upgrade all again when deploying the migrated database).

hbrunn:
tested, works and is a lot faster.

@jbeficent after fixing https://travis-ci.org/OCA/OpenUpgrade/builds/138921636#L265-L268, the branch should become green and we can merge

legalsylvain:
@jbeficent, I fixed flake8 errors that @hbrunn mentioned. Could you accept this PR ? 
https://github.com/Eficent/OpenUpgrade/pull/10

kind regards.

pedrobaeza:
@jbeficent, please check Travis and attend the remaining comment and I'll merge.

JordiBForgeFlow:
@pedrobaeza @hbrunn This one is ready to merge.

JordiBForgeFlow:
@pedrobaeza Can you merge this one?

PR #471, opened 2016-04-20
[FIX] Remove deprecated zsi.

NL66278:
Also remove import_sugarcrm, the one module that was using zsi.

zsi requires (for no reason by the way) PyXML, that no longer installs on
up to date debian/ubuntu, at least not in virtualenv. Anyway PyXML is quickly removed from distributions.

The removal of import_sugarcrm is no problem for migration, because the module does not exist in 7.0 any longer anyway.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,6 @@ setuptools.setup(
           'vobject',
           'werkzeug',
           'xlwt',
-          'zsi',
       ],
       extras_require = {
           'SSL' : ['pyopenssl'],


pedrobaeza:
I better prefer to just enclose the import sentence in a try to bypass the error.

NL66278:
@pedrobaeza The trouble is that I want the OpenUpgrade server to run from an Anybox buildout environment. Buildout tries to install eggs for all required python packages. But the zsi package no longer installs, because it depends on PyXML (that itself can not be installed because of an error in the way it inspects python.h). The failure to install zsi causes the buildout itself to fail.

pedrobaeza:
You can remove the dependency (in the requirements.txt), and put the try in the import and get the same result.

NL66278:
My excuses. This PR is from the wrong branch. This one needs to be pushed to OpenUpgrade server. Will replace this one with another. This will be along the lines proposed by @pedrobaeza  (that is: absolute minumum of changes).

NL66278:
I am getting confused. Trying to understand why the commits for project_mrp show up in my PR.

NL66278:
@pedrobaeza The fix should be right now. The actual imports of ZSI were already protected by a try except block. Although the actual except raises a new exception warning that python soap support should be installed, I think or rather hope  this will not be a problem when not installing the module.

StefanRijnhart:
:+1:

pedrobaeza:
:+1:

PR #473, opened 2016-04-21
[FIX] Prevent transaction lockups

### LINKED ISSUE #433: [9.0] migration death at product_view.xml
Migration process freezes after logging "loading product/product_view.xml". Process stops on its track, no exception or exit.

Problem occurs when validating the record `product_attribute_value_view_tree`. `BaseModel.fields_get()` iterates for `price_extra` but `field.get_description()` never returns. Debugger communication also stops when accessing `_description_digits` prior to this. My guess is Float.digits() never returns.

Problem is created when processing `product_security.xml`. A call to `res_group.write()` flushes the cache through `call_cache_clearing_methods()`. At this point `_description_digits` becomes inaccessible.

Is there a way to keep the cache or restore it?

StefanRijnhart:
... when Odoo tries to figure out about a  specific decimal precision
after these have been written when loading module
data (product module). These lockups occur because OpenUpgrade does not
commit after loading this data, and Odoo uses a separate cursor (most likely
from another connection cq. transaction) to query the precision when views
are validated. The solution is to use the migration cursor instead.

Fixes #433

### DIFF
--- a/openerp/fields.py
+++ b/openerp/fields.py
@@ -11,6 +11,7 @@ from types import NoneType
 import logging
 import pytz
 import xmlrpclib
+from psycopg2 import OperationalError
 
 from openerp.tools import float_round, frozendict, html_sanitize, ustr, OrderedSet
 from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as DATE_FORMAT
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
:+1: Tested on a migration of product to v9 including module 'product'. Works fine!

StefanRijnhart:
Thanks @jbeficent for testing. I made this multi-database safe and rebased into a single commit. Ready for further reviews.

pedrobaeza:
That sounds better! At first sight I didn't see where `dbname` variable was located, but it's correctly used.

Can you please change also the same call for old API: https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/osv/fields.py#L381 and https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/osv/fields.py#L1349?

StefanRijnhart:
@pedrobaeza I believe that is already included in this change!

pedrobaeza:
Ouch, I don't know why I didn't see it.

:+1:

PR #475, opened 2016-04-25
[FIX] Add fix to prevent created xml_id's from being deleted afterwards.

NL66278:
After the PR to add missing xml_id's it became clear we needed one additional step to prevent the created data from being thrown away at the end of the update process.


hbrunn:
:+1:

StefanRijnhart:
:+1:

PR #489, opened 2016-05-02
9.0 base

coleste:
Minimum fixes to get a migration running.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -54,7 +54,23 @@ column_renames = {
 }
 
 
+OBSOLETE_RULES = (
+    'multi_company_default_rule',
+    'res_currency_rule',
+)
+
+
+def remove_obsolete(cr):
+    openupgrade.logged_query(cr, """
+        delete from ir_rule rr
+        using ir_model_data d where rr.id=d.res_id
+        and d.model = 'ir.rule' and d.module = 'base'
+        and d.name in {}
+        """.format(OBSOLETE_RULES))
+
+
 @openupgrade.migrate()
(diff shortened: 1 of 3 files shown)


openerp/addons/base/migrations/9.0.1.3/pre-migration.py
```diff
@@ -54,7 +54,23 @@
...
+)
+
+
+def remove_obsolete(cr):
+    openupgrade.logged_query(cr, """
+        delete from ir_rule rr
+        using ir_model_data d where rr.id=d.res_id
+        and d.model = 'ir.rule' and d.module = 'base'
+        and d.name in {}
+        """.format(OBSOLETE_RULES))
```
hbrunn, line 69: @coleste why do you need to do this manually? Shouldn't Odoo clean that up at the end of the migration run?

coleste, line 69: @hbrunn I'm happy to see I'm not the only one surprised they weren't cleaned up.

Every change was done because the database would fail to run after migration. In this case, model multi_company.default is gone from 9.0.

StefanRijnhart:
Yes, sorry @coleste. That was a bug in the framework. Seems Odoo 9.0 raises in other places than 8.0 did so the port of the code was not enough to prevent these errors.

StefanRijnhart:
It was fixed by @hbrunn, and then he did a great job of removing these stanzas from your migration scripts so that they could be merged.

StefanRijnhart:
Oh, thanks for your contributions, of course :-)

coleste:
You're welcome. Nice to see those hacks were necessary because of a bug, and were enough of a pointer for @hbrunn to correct what was beyond my abilities.

PR #490, opened 2016-05-02
9.0 crm

coleste:
Minimum fixes to get a migration running.


hbrunn:
here (and in other PRs) you delete a lot of views. Why is this necessary? In theory, the upgrade mechanism shouldn't care about broken views until the very end of the migration, and then, obsolete views should be cleaned up. So it shouldn't be necessary to do most of what you do.

If this mechanism broken?

PR #497, opened 2016-05-06
[8.0] Fix migration script of module account

ghost:


hbrunn:
:+1:, verified it's now like this in 7.0: https://github.com/OCA/OCB/blob/7.0/addons/account/account_bank_statement.py#L567 but at the time of writing this script, it was probably this: https://github.com/OCA/OCB/commit/59e5b3e475e4037c480b83059e5bfd6ad29c5dff

pedrobaeza:
Isn't this a stable version where no DB layout change is made?!!

Thanks for the patch. :+1:

ghost:
@pedrobaeza I ported account and stock successfully, except for this small error, so I guess the branch is stable.

pedrobaeza:
I mean about stability in Odoo. Changing a field name in the middle of the version 8.0 doesn't mean that.

hbrunn:
@pedrobaeza mind you this is a change in 7.0, which was introduced quite early in its lifetime

pedrobaeza:
Ah, OK, I mixed the information. I proceed with the merge.

pedrobaeza:
Eh, wait... if now it's `statement_line_id`, why are we changing it to `statement_id`? I think I wouldn't merged the PR. The problem is an outdated 7.0. Am I wrong?

hbrunn:
you made me think, that hurts. But yes, I think you're right. Outdated v7, we need to revert this commit.

pedrobaeza:
Reverted. @dufresnedavid, please check what we commented.

ghost:
@pedrobaeza I don't understand your point. In both Odoo 7.0 and Odoo 8.0. The field is statement_id. There is no such thing as statement_line_id in any version of Odoo anymore. So why are we changing this ?

Also, this is the post-migration script, so why are we talking about v7 ?

pedrobaeza:
No, in 7.0 is the other column name: https://github.com/OCA/OCB/blob/7.0/addons/account/account_bank_statement.py#L567

ghost:
@pedrobaeza ok, I get it. Thanks!

PR #498, opened 2016-05-06
[ADD] membership migration

hbrunn:


addons/membership/migrations/8.0.0.1/post-migrate.py
```diff
@@ -0,0 +1,20 @@
...
+@openupgrade.migrate()
+def migrate(cr, version):
+    cr.execute(
+        '''update product_template set
+        membership = p.membership,
+        membership_date_from = p.membership_date_from,
+        membership_date_to = p.membership_date_to
+        from product_product p
+        where p.product_tmpl_id = product_template.id''')
+    cr.execute(
```
pedrobaeza: Better to let this job be done by database_cleanup?

StefanRijnhart:
:+1:, I prefer like @pedrobaeza not to delete data in scripts. However, risk of dataloss in this case is nihil IMHO.

hbrunn:
@pedrobaeza @StefanRijnhart agreed and removed

pedrobaeza:
:+1:

PR #500, opened 2016-05-17
7.0 fix partner performance

mourad-ehm:
Fix partner migration performance by bypassing ORM method and doing mass data update and insert.
This modification allows me to pass migration time of partners from ~ 4.5 hours to 15 mn (about 116000 partner).

### DIFF
--- a/openerp/addons/base/migrations/7.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/7.0.1.3/post-migration.py
@@ -334,6 +334,19 @@ This can be the case if an additional module installed on your database changes
         _insert_partners(cr, pool, partner_store_insert)
         _update_partners(cr, pool, partner_store_update)
 
+        # Update propagate_fields from partner parent
+        # TODO Use propagate_fields = [ 'lang', 'tz', 'customer', 'supplier',]
+        # to generate this query
+        openupgrade.logged_query(
+            cr, "\n"
+            "UPDATE res_partner part "
+            "SET lang = parent.lang, "
+            "tz = parent.tz,"
+            "customer = parent.customer,"
+            "supplier = parent.supplier"
+            "FROM res_partner parent "
+            "WHERE part.parent_id = parent.id"
+            " AND part.parent_id IS NOT NULL")
         # set_address_partner by mass update
         openupgrade.logged_query(
             cr, "\n"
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Interesting, thank you. Tentative :+1: (no test).

openerp/addons/base/migrations/7.0.1.3/post-migration.py
```diff
@@ -240,58 +264,107 @@ def create_partner(address_id, vals, defaults):
...
+        openupgrade.logged_query(
+            cr, "\n"
+            "UPDATE res_partner part "
+            "SET lang = parent.lang, "
+            "tz = parent.tz, "
+            "customer = parent.customer, "
+            "supplier = parent.supplier "
+            "FROM res_partner parent "
+            "WHERE part.parent_id = parent.id "
+            " AND part.parent_id IS NOT NULL")
```
hbrunn: We should only do this with partners we just inserted, right? Otherwise, those values will be reset for existing sub-partners from 6.1.
Also consider sorting partners by their nesting nevel before doing the update, this query will not propagate correctly in multi level hierarchies.

hbrunn:
I just ran this on a customer database and it works. But I think you need to address https://github.com/OCA/OpenUpgrade/pull/500/files#r66779069 before this can be merged.

openerp/addons/base/migrations/7.0.1.3/post-migration.py
mourad-ehm: OK for first part of the note. I'll fix it. 
But what about sorting of partner by their nesting level? If I understood, you suppose that we can have several level of companies ? Theorically it is possible but in the real use in openerp V6.1, only not company partner can have a parent. And partner parent is necessarily a company ! right ?

hbrunn: Don't make assumptions what people did to their installation. The data model allows any partner being parent of any other partner. More importantly: The code you replace supports it, so you should do the same.

But thinking again: As soon as you introduced the restriction from the first part of the comment, there can't be multi-level relations, so nothing to worry about.

hbrunn:
:+1:

StefanRijnhart:
Thanks for the follow up. @hbrunn thanks for the review.

PR #503, opened 2016-06-07
[FIX] updating every record once is enough

hbrunn:
the statement as it is causes n^2 updates, that does quite some pain for big databases (in my case, this went until the OOM killer kicked in for the postgres process)


pedrobaeza:
:+1:

StefanRijnhart:
OK, I confess. But it was three years ago ;-)
:+1:

PR #505, opened 2016-06-08
[8.0] Convert email server actions to email.template

ghost:
In 8.0, mail body of server actions are not stored directly on the action. The action is linked to a email.template. This script converts the messages to templates and links the action to the new created template.

### DIFF
--- a/addons/email_template/migrations/8.0.1.1/post-migration.py
+++ b/addons/email_template/migrations/8.0.1.1/post-migration.py
@@ -4,6 +4,7 @@
 
 import logging
 
+from psycopg2.extensions import AsIs
 from openerp.openupgrade import openupgrade
 
 from openerp import SUPERUSER_ID
@@ -43,19 +44,14 @@ def convert_action_mail_server_email(env):
 
     for action in actions:
 
+        subject_column = openupgrade.get_legacy_name('subject')
+
         env.cr.execute(
             """
-            SELECT openupgrade_legacy_email
-                , openupgrade_legacy_subject
-                , openupgrade_legacy_message
+            SELECT email, %s, message
             FROM ir_act_server
             WHERE id = %s
-            """, (
-                openupgrade.get_legacy_name('email'),
-                openupgrade.get_legacy_name('subject'),
-                openupgrade.get_legacy_name('message'),
-                action.id,
-            ))
+            """, (AsIs(subject_column), action.id, ))
 
         (email, subject, message) = env.cr.fetchone()
 
--- a/addons/email_template/migrations/8.0.1.1/post-migration.py
+++ b/addons/email_template/migrations/8.0.1.1/post-migration.py
@@ -50,7 +50,12 @@ def convert_action_mail_server_email(env):
                 , openupgrade_legacy_message
             FROM ir_act_server
             WHERE id = %s
-            """, (action.id, ))
+            """, (
+                openupgrade.get_legacy_name('email'),
+                openupgrade.get_legacy_name('subject'),
+                openupgrade.get_legacy_name('message'),
+                action.id,
+            ))
 
         (email, subject, message) = env.cr.fetchone()
 
--- a/addons/email_template/migrations/8.0.1.1/post-migration.py
+++ b/addons/email_template/migrations/8.0.1.1/post-migration.py
@@ -45,7 +45,9 @@ def convert_action_mail_server_email(env):
 
         env.cr.execute(
             """
-            SELECT email, old_subject, message
+            SELECT openupgrade_legacy_email
+                , openupgrade_legacy_subject
+                , openupgrade_legacy_message
             FROM ir_act_server
             WHERE id = %s
             """, (action.id, ))
--- a/addons/email_template/migrations/8.0.1.1/post-migration.py
+++ b/addons/email_template/migrations/8.0.1.1/post-migration.py
@@ -61,9 +61,3 @@ def convert_action_mail_server_email(env):
         })
 
         action.template_id = email_template.id
-
-    openupgrade.drop_columns(env.cr, [
-        ('ir_act_server', 'email'),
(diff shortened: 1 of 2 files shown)


addons/email_template/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+    convert_action_mail_server_email(env)
+
+
+def convert_action_mail_server_email(env):
+    def convert_body(body):
+        return body.replace("[[", "${").replace("]]", "}")
+
+    def convert_subject(subject):
+        subject = subject.strip()
+        if subject.startswith('object.'):
```
pedrobaeza, line 31: Shouldn't you replace any occurrence of `object` with `${` and find the end of the word to put `}`?

addons/email_template/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+    ])
+
+    _logger.info(
+        'Transfering existing ir.actions.server emails to '
+        'email templates.'
+    )
+
+    for action in actions:
+
+        env.cr.execute(
```
pedrobaeza, line 49: Use directly `action.email`, `action.old_subject` and `action.message`

addons/email_template/migrations/8.0.1.1/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+            'name': action.name,
+            'model_id': action.model_id.id,
+            'email_from': convert_subject(email),
+            'subject': convert_body(subject),
+            'body_html': convert_body(message),
+        })
+
+        action.template_id = email_template.id
+
+    openupgrade.drop_columns(env.cr, [
```
pedrobaeza: Let the columns drop to module database_cleanup. It's the OpenUpgrade technique to not drop anything that can be used for debugging later or for other custom migration scripts.

addons/email_template/migrations/8.0.1.1/post-migration.py
ghost, line 49: @pedrobaeza the fields don't exist, you can not retrieve those from the ORM.

addons/email_template/migrations/8.0.1.1/post-migration.py
ghost, line 31: @pedrobaeza I prefer doing it simple. For the database I am migrating, this covers most cases. Anyway, its not possible to have a 100% accurate solution. The details can be fixed manually.

addons/email_template/migrations/8.0.1.1/post-migration.py
ghost: @pedrobaeza ok, I don't mind removing this part.

addons/email_template/migrations/8.0.1.1/post-migration.py
pedrobaeza, line 49: Shouldn't you then rename the columns to None to mark them as "openupgrade handled" columns?

ghost, line 49: @pedrobaeza why would I do that, I don't get the use case ? I prefer keeping the old values as backup.

addons/email_template/migrations/8.0.1.1/post-migration.py
pedrobaeza, line 31: I think it's not too complicated (draft code):

```
final_subject = []
while subject:
    index = subject.find('object.')
    if index >= 0:
        index_end = subject.find(' ', index)
        if index_end:
            final_subject += subject[:index] + "${" + subject[index:index_end] + "}"
            subject = subject[index_end:]
        else:
            final_subject += subject[:index] + "${" + subject[index:] + "}"
            subject = ""
    else:
        final_subject += subject
        subject = ""


```

addons/email_template/migrations/8.0.1.1/post-migration.py
pedrobaeza, line 49: The values are preserved also, but columns are renamed to openupgrade_legacy.... It's a way to avoid that another module with the same field name waste old data.

ghost, line 49: I renamed the columns to openupgrade_legacy....

addons/email_template/migrations/8.0.1.1/post-migration.py
ghost, line 31: @pedrobaeza again, I don't see the use case. What I want is to cover 90% cases and the rest will be fixed manually. I have 19 templates to create. If three of them must be corrected manually, I don't care. Can we move on?

pedrobaeza, line 31: OK, but if I give you the code, it's not so hard to refine I think, but I won't block this.

addons/email_template/migrations/8.0.1.1/pre-migration.py
```diff
@@ -24,6 +24,9 @@
 column_renames = {
     'email_template': [
         ('email_recipients', 'partner_to'),
+    ],
+    'ir_act_server': [
+        ('subject', 'old_subject'),
```
StefanRijnhart: Use None here, and refer to the old subject column name in the post script with openupgrade.get_legacy_name('old_subject')

hbrunn: `openupgrade.get_legacy_name('subject')` to be precise

ghost:
@StefanRijnhart @hbrunn Fixed

hbrunn:
:+1:

StefanRijnhart:
Thanks!

ghost:
@pedrobaeza @hbrunn @StefanRijnhart Thanks for the reviews!

PR #506, opened 2016-06-09
[ADD] initial PoC for tests for OpenUpgrade

hbrunn:
this is a PoC for the scenario described in https://odoo-community.org/groups/contributors-15/contributors-41651?mode=date&date_begin=2016-06-01&date_end=2016-07-01, don't merge (yet)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,22 @@ sudo: false
 python:
   - "2.7"
 
+addons:
+    apt:
+        packages:
+            - expect-dev  # provides unbuffer utility
+            - python-lxml  # because pip installation is slow
+            - python-simplejson
+            - python-serial
+            - python-yaml
+    postgresql: "9.4"
+
+env:
+  - DB=openupgrade ODOO=./openerp-server OPENUPGRADE_TESTS=1
+
 # Need flake8 for pep8 testing
 install:
-    - pip install flake8
+    - pip install flake8 coveralls
 
 # Test with flake for:
 #   * migration scripts
@@ -20,3 +33,18 @@ script:
     - flake8 openerp/addons/openupgrade* --max-line-length=120 --filename=__init__.py --ignore=F401
     - flake8 openerp/addons/openupgrade* --max-line-length=120 --exclude=__init__.py
     - flake8 . --max-line-length=120 --filename=pre-*.py,post-*.py
+    - npm install -g less less-plugin-clean-css
+    - pip install -q -r requirements.txt
+# don't use pypi's openupgradelib, but the one from source to avoid choking
+# on unreleased features in openupgradelib
+    - pip uninstall -q --yes openupgradelib
+    - pip install -q git+https://github.com/OCA/openupgradelib.git@master
+    - createdb $DB
+    - wget -q -O- $(wget -q -O- https://doc.therp.nl/openupgrade/databases/for/$($ODOO --version | awk '{print $3}')) | pg_restore -d $DB --no-owner
+    - MODULES=base,$(sed -n '/^+========/,$p'  openerp/openupgrade/doc/source/modules80-90.rst | grep Done | sed -n '/^|/ s/^|\([0-9a-z_]*\) *|.*$/\1/g p' | paste -d, -s)
+    - psql $DB -c "update ir_module_module set state='uninstalled' where name not in ('$(echo $MODULES | sed -e "s/,/','/g")')"
+    - echo Testing modules $MODULES
+    - coverage run $ODOO --database=$DB --update=$MODULES --stop-after-init
+
+after_success:
+    coveralls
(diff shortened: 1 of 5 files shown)


pedrobaeza:
The problem I see is that we are only going to test:
- The correct run of the script.
- The checks we want to incorporate, based only in what we figure that we have to migrate

But we are not going to know exactly if the migration is totally correct. These tests can create the false sensation of correct migration even if we don't include any assert.

Anyway, having this clear in documentation, it's a good step to avoid some running issues.

hbrunn:
...and we can (and should) also include measuring coverage, and at the end, only merge stuff which doesn't decrease coverage

StefanRijnhart:
:star2: :star2: :star2:

Try

```
postgresql: "9.4"
```

hbrunn:
I'm trying to find the reason for https://travis-ci.org/OCA/OpenUpgrade/builds/136356675#L503, which is weird, because both https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/addons/base/migrations/9.0.1.3/post-migration.py#L119 and https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1140 look correct. If you have any ideas, let me know

hbrunn:
looks like https://github.com/OCA/openupgradelib/commit/e748410bdc0ff8c3281ede39693c32c2f8c074c8 is missing from the openupgradelib travis uses. Will have a look what's going on here.

pedrobaeza:
It seems there has no been another release in PyPi with latest changes of openupgradelib.

pedrobaeza:
I'm going to make another release.

hbrunn:
@pedrobaeza thanks, the new release surely makes sense. But for the tests, I'll rewrite `.travis.yml` to pull the git version of openupgradelib in order to have the latest fixes in there. Especially at the start of the version, we'll probably add stuff and then having to release basically every commit will become annoying.

pedrobaeza:
@hbrunn, OK, you're right about that point.

.travis.yml
```diff
@@ -35,6 +35,11 @@ script:
     - flake8 . --max-line-length=120 --filename=pre-*.py,post-*.py
     - npm install -g less less-plugin-clean-css
     - pip install -q -r requirements.txt
+# don't use pypi's openupgradelib, but the one from source to avoid choking
+# on unreleased features in openupgradelib
+    - pip uninstall openupgradelib
+    - git clone https://github.com/OCA/openupgradelib ~/openupgradelib
+    - pip install ~/openupgradelib
```
pedrobaeza: You can install directly with pip: http://stackoverflow.com/questions/20101834/pip-install-from-github-repo-branch

hbrunn:
@pedrobaeza @StefanRijnhart now it's the way I want it, please review

StefanRijnhart:
Awesome! And it's green. :+1: :+1: :+1:

hbrunn:
I see three thumbs up, so I'll merge

hbrunn:
kidding, but thanks for your enthusiasm ;-)

pedrobaeza:
Is the coverage information correct? How do you measure this coverage?

hbrunn:
@pedrobaeza it only looks at the migration scripts themselves: https://github.com/OCA/OpenUpgrade/pull/506/files#diff-6f1f859c6ca5c4cae7310924786cf661R7

legalsylvain:
Hi, 

Well first :100: :+1: for this work (*) ! This a good step forward to improve Openupgrade.

A suggestion :  
- What about generating DB dynamicaly, based on an OCB instance created each time ?

Pro : 
- we could install only "upgradable" module. If lunch is not ported, no need to install it. -> less time during upgrade / update process
- if something changes in OCB / 8.0, maybe the migration will no longer works, and we could detect it (theoritical case maybe ...)
- no need for Therp BV to host and maintain db https://doc.therp.nl/openupgrade/databases/for/9.0
- maybe we could quickly patch OCB / 8.0 branch if there a bug in a demo data that makes the upgrade fail

Cons : 
- more time to generate db
- other ?

And a question : I remember that for previous migration, it was not possible to run an openupgrade on a db with demo data but I don't remember why. Why it is now possible from 8.0 -> 9.0 ?

kind regard.

(*) 103 thumbs : time to merge... ;-)

pedrobaeza:
:+1: Merging then

hbrunn:
@pedrobaeza thanks for merging

@legalsylvain the time thing is crucial: We only have 50 minutes, then travis will kill the job. Now doing nearly nothing it costs 10 minutes, so I fear we'll get timing problems anyways with a growing amount of modules.

Therp hosting the DB images was now the simplest choice, sooner or later I'll ask the OCA to do so - also makes more sense because then more people have access to it. Or just on github, weird I didn't think of that before.

legalsylvain:
@hbrunn Thanks for the explanation. (and again for this improvement).

PR #509, opened 2016-06-14
[ADD] all in one 2016-06-14

hbrunn:
This is an umbrella PR for the existing 9.0 PRs, pushed to @OpenUpgrade-dev so that the regular contributors and simply add their stuff there. The hope is that this passes the tests and we can just merge it.


pedrobaeza:
@hbrunn, this is being PRed to the regular 9.0 branch. Is this what you intend?

hbrunn:
yes, I want first check if the tests work. Then I think we can just merge it. What I want to avoid are the long standing PRs for months, this frustrates contributors just as much as users.

pedrobaeza:
But I think that scripts are not ready for production yet. I prefer to rebase each PR and check independently.

hbrunn:
I think this is going to slow down development a lot

pedrobaeza:
As each module migration is independent from each other (most of the times), I don't think so, but let's see what others say.

hbrunn:
actually, nearly every module depends on account, mail, product or the like. Having those as lurking PRs complicates things needlessly in my opinion. What do @OCA/openupgrade-maintainers think?

StefanRijnhart:
I'm already losing oversight, so let's start with the bottom dependencies.

hbrunn:
@StefanRijnhart what is your proposal? My plan is to make this PR green, have you guys merge it, and from then on, we can merge fast because the basics are covered and new PRs will be green if they don't mess up.

StefanRijnhart:
OK, let's do it.

hbrunn:
yay! Just wait until travis finishes, then we should have 100% coverage

PR #510, opened 2016-06-14
[ADD] migration for sales_team

hbrunn:


hbrunn:
tests are broken because depending modules are not installed

addons/sales_team/migrations/9.0.1.0/pre-migration.py
```diff
@@ -1,72 +1,20 @@
...
-    'crm_case_section_salesteams_view_kanban',
-    'crm_case_section_view_form',
-    'crm_case_section_view_tree',
-    'res_user_form',
-    'view_sale_config_settings',
-    'view_users_form_preferences',
-)
-
-
-def recursive_delete(cr):
```
hbrunn, line 37: we don't need the whole function. Before #509, the upgrade would choke on all kinds of view problems, but with commit https://github.com/OCA/OpenUpgrade/commit/33ff049e18fa7b15892c8dd141f64e8e1fd53773, we don't need any manual cleanup, because the offending views will be dropped later (or changed during the upgrade)

StefanRijnhart:
? SyntaxError: Non-ASCII character '\xc2' in file /home/travis/build/OCA/OpenUpgrade/addons/sales_team/migrations/9.0.1.0/pre-migration.py on line 1, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

StefanRijnhart:
Ah, you did remove the coding tag when slashing the file with your big code hammer

hbrunn:
you gotta picture my like that when coding: http://st.depositphotos.com/1494934/4274/i/950/depositphotos_42748967-Strong-fat-man-holding-a.jpg _scnr_

should be fixed now

StefanRijnhart:
Thanks!

PR #512, opened 2016-06-14
[ADD] test migration

hbrunn:
The database creation script is still busy, don't merge just yet as the first test will fail anyways, and then we probably need some rework to make it work on demo data. In the PR for 9.0, I had to rename a couple of XMLIDs for the demo data, which is probably necessary here too.

You're more than welcome to help: Find demo data with renamed XMLIDs where some constraint will cause problems if some record is duplicated. Only in this case, we need the rename. Notorious candidates for this are the name field on categories or tags for example.


StefanRijnhart:
You beat me to it, but maybe you want my test? https://github.com/OCA/OpenUpgrade/pull/513/files

StefanRijnhart:
Set to work in progress to indicate no merging.

hbrunn:
@StefanRijnhart thanks, added your test and uploaded the test database, now let's wait

StefanRijnhart:
That's it, thanks.

PR #513, opened 2016-06-14
8.0 tests

StefanRijnhart:
Port of #506


StefanRijnhart:
Beaten with a whole three minutes by https://github.com/OCA/OpenUpgrade/pull/512

PR #520, opened 2016-06-16
[IMP] Lint tests

StefanRijnhart:


StefanRijnhart:
Included in #519 because of inherent dependency

PR #524, opened 2016-06-19
[FIX] don't pull openupgradelib in graph.py

hbrunn:
the problem with this is that with it, we need to import openerp before we can import something from openupgradelib. This happens to go right when running the server, but not for example when building the documentation.

### DIFF
--- a/openerp/modules/graph.py
+++ b/openerp/modules/graph.py
@@ -18,7 +18,6 @@ from openerp.tools.translate import _
 
 import zipfile
 import openerp.release as release
-from openupgradelib import openupgrade
 
 import re
 import base64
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't really see the point you want to arise, but this change seems no special problem, so :+1:

hbrunn:
@pedrobaeza before this change, you couldn't simply say `from openupgradelib import openupgrade` in some helper script. You need to know that this only works by importing openerp first, then the other import works. With my proposed change, order doesn't matter, as before.

pedrobaeza:
Thanks for the explanation, @hbrunn

StefanRijnhart:
Thanks. We should add building the documentation to the tests.

PR #525, opened 2016-06-20
[ADD] migrate l10n_nl

hbrunn:
All the other localizations will need the same thing, I guess?


addons/l10n_nl/migrations/8.0.1.5/noupdate_data.xml
```diff
@@ -0,0 +1,7 @@
+<openerp>
+    <data>
+        <record id="l10nnl_chart_template" model="account.chart.template">
```
pedrobaeza, line 3: l10n_nl_chart_template?

pedrobaeza:
Are chart templates noupdate? At least, l10n_es isn't.

addons/l10n_nl/migrations/8.0.1.5/noupdate_data.xml
hbrunn, line 3: no: https://github.com/OCA/OCB/blob/8.0/addons/l10n_nl/account_chart_netherlands.xml#L4096

hbrunn:
in 9.0, this one isn't either, but this is about 8.0. Thinking about that: Does that mean we need to lift the noupdate flag when migrating to 9.0? Probably, right?

addons/l10n_nl/migrations/8.0.1.5/noupdate_data.xml
pedrobaeza, line 3: OK

pedrobaeza:
Yes, that would be desirable to leave a correct state after migration.

StefanRijnhart:
So it should be in the analysis.

StefanRijnhart:
Or in the compare script.

hbrunn:
@StefanRijnhart 'should' as in 'I expect that to happen already' or 'should' as in 'someone should put it in there'?

StefanRijnhart:
Someone, sometime!

PR #528, opened 2016-06-20
[ADD][9.0] migration script for module 'account'

JordiBForgeFlow:


addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,112 @@
+# -*- coding: utf-8 -*-
+# © 2016 Serpent Consulting Services Pvt. Ltd.
+# © 2016 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
hbrunn, line 5: there already exists a file with this name, please rebase: https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/pre-migration.py

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,362 @@
...
+    cr.execute("""
+    UPDATE account_account_type SET include_initial_balance =  CASE
+    WHEN %(openupgrade)s = 'none' THEN False
+    ELSE True
+    END 
+    """%{'openupgrade' : openupgrade.get_legacy_name('close_method')})
+
+    # Set bank_statements_source to 'manual' 
+    cr.execute("""
+    UPDATE account_journal SET bank_statements_source = 'manual'
```
hbrunn, line 221: for things like this, you can use https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.set_defaults

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,362 @@
...
+    """)
+
+    # Value 'percentage_of_total' => 'percentage' 
+    cr.execute("""
+    UPDATE account_operation_template SET amount_type = 'percentage' WHERE amount_type = 'percentage_of_total'
```
hbrunn, line 226: and for those, https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.map_values

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -0,0 +1,362 @@
...
+    cr.execute("""
+    UPDATE account_chart_template SET company_id = r.id 
+    from res_company r
+    """)
+
+    for n in range(company_count):
+        company_id = list_of_companies[n]['id']
+        for m in range(chart_template_count):
+            chart_id = chart_ids[m]['id']
+            cr.execute("""
```
hbrunn: I don't think you should duplicate all the templates here. The important part is that we _assign_ some existing template to the company, and the tricky part is to assign the right one (based on the one initially used). I'll come up with a PR on yours soon

hbrunn:
please rebase on current 9.0 (this gives you tests for free and avoids conflicts)

JordiBForgeFlow:
Thanks @hbrunn Now rebased. Tests means that we should be able to run with a test database, and the unit tests will be run? That will be great!

hbrunn:
that's exactly what travis does now. We worked hard on this last week, but it seems to work nicely by now.

hbrunn:
@jbeficent I made a proposal for how to handle the account templates in https://github.com/Eficent/OpenUpgrade/pull/5

You'll have to fix the flake8 errors yourself, I'll look into the constraint the test runs on.

JordiBForgeFlow:
@hbrunn The idea to find the best possible matching template for the company looks good to me. As for assigning the parent tax as a tag, we discussed with @pedrobaeza about having a method that converts parent accounts in the Chart of Accounts into tags as well. A method that would look, for each account, at all it's parents, and convert to corresponding tags (creating or reusing existing).

hbrunn:
yes, that's the plan with `parent_id_to_tag` - there are several instance in 9.0 where we need something like this. I _think_ just adding `parent_id_to_tag(cr, 'account.account')` will be sufficient.

JordiBForgeFlow:
Yes 'parent_id_to_tag' would be used by both models. I see that the record will have the tags of all the parents up the hierarchy. In Spain the COA has multiple levels, and for reporting purposes an account tagged with each upper level is important.

hbrunn:
the function does not work recursively (yet), but that's simple to add. Best guided by a parameter so that we can turn it of (or on, not sure) as we need

hbrunn:
superseded by #540

PR #531, opened 2016-06-22
[ADD][9.0] migration script for module 'base'

JordiBForgeFlow:


openerp/addons/base/migrations/9.0.1.3/pre-migration.py
```diff
@@ -53,24 +53,35 @@
...
+    """)
+#    amount = cr.dictfetchone()
+#    double_validation_amount = str(amount['condition']).split(" ")[-1]
+
+    # delete purchase order workflow
+    openupgrade.delete_model_workflow(cr, 'purchase.order')
+#    transitions = openupgrade.deactivate_workflow_transitions(cr, 'purchase.order')
+#    openupgrade.reactivate_workflow_transitions(cr, transitions)
+
+    # Inherited Views that encountered errors while running the migration.
```
hbrunn, line 74: this is not necessary any more, remove all view related stuff

openerp/addons/base/migrations/9.0.1.3/pre-migration.py
```diff
@@ -53,24 +53,35 @@
...
+    cr.execute("""
+    select condition from wkf_transition wkf join ir_model_data ir 
+    on wkf.id = ir.res_id and ir.name = 'trans_confirmed_double_gt' 
+    and ir.module = 'purchase_double_validation'
+    """)
+#    amount = cr.dictfetchone()
+#    double_validation_amount = str(amount['condition']).split(" ")[-1]
+
+    # delete purchase order workflow
+    openupgrade.delete_model_workflow(cr, 'purchase.order')
```
hbrunn, line 70: why here and not in the migration of purchase?

openerp/addons/base/migrations/9.0.1.3/post-migration.py
```diff
@@ -72,6 +72,7 @@ def clear_inherit_id(cr):
         """)
 
 
+# updates to ir_ui_view will not clear inherit_id
```
hbrunn, line 75: how are those comments relevant?

openerp/addons/base/migrations/9.0.1.3/post-migration.py
```diff
@@ -83,6 +84,18 @@ def rename_your_company(cr):
...
 
+def update_res_currency(cr):
+    openupgrade.logged_query(cr, """
+        DELETE FROM ir_rule
+        WHERE model_id IN (
+            SELECT id from ir_model
+            WHERE model = 'res.currency'
+        )
+        AND domain_force like '%%company_id%%'
+        """)
```
hbrunn, line 95: why would we want to do that? Records with an XMLID will be deleted at the end of a successful migration anyways, and the migration itself runs as `SUPERUSER_ID`, so they won't get in the way

hbrunn:
I think that all the code in this branch is either not necessary or should be moved somewhere else, so I'll close this PR. Please reopen if I missed something.

StefanRijnhart:
Indeed. Also note the strange changes to openerp/addons/base/migrations/9.0.1.3/openupgrade_general_log.txt.

Issue #400, opened 2015-11-17
Migration scripts for v9


pedrobaeza:
This issue summarizes the development of the migration scripts for v9. Please write down a comment if you want to develop a migration script for one module, to avoid duplicating efforts between the contributors. I have extracted the module list to have it as a reference for tracking the progress:
- [x] base - Done by @coleste, @hbrunn 
- [x] account - Partial covered by @coleste, completed by @hbrunn
- [x] account_accountant - Nothing to do
- [x] account_analytic_default - Nothing to do
- [x] account_asset - Done by @hbrunn 
- [x] account_budget - Nothing to do
- [x] account_cancel - Nothing to do
- [x] account_chart (`del`) - By @cubells - #1070 
- [x] account_check_writing > account_check_printing - Done by @hbrunn
- [ ] account_tax_cash_basis
- [ ] account_tax_python
- [ ] account_test
- [x] account_voucher - Assigned to @jbeficent
- [x] analytic - Assigned to @hbrunn 
- [ ] anonymization
- [x] association - By @pedrobaeza - #911 
- [x] auth_crypt - Nothing to do
- [x] auth_ldap - Nothing to do
- [x] auth_oauth
- [x] auth_signup - Done by @hbrunn
- [x] barcodes - Nothing to do
- [x] base_action_rule - Nothing to do
- [ ] base_gengo
- [ ] base_geolocalize
- [x] base_iban - Nothing to do
- [x] base_import - Nothing to do
- [x] base_import_module - Nothing to do
- [x] base_setup - Nothing to do
- [x] base_vat
- [x] board - Nothing to do
- [x] bus - Done by @hbrunn
- [x] calendar - Nothing to do
- [x] claim_from_delivery - Done by @hbrunn
- [x] crm - Completed by @hbrunn , @coleste 
- [x] crm_claim - Done by @hbrunn 
- [ ] crm_partner_assign
- [ ] crm_project_issue
- [x] decimal_precision - Nothing to do
- [x] delivery - Done by @hbrunn 
- [x] document - Completed by @hbrunn 
- [x] event - By @cubells - #712 
- [x] event_sale
- [x] fetchmail - Nothing to do
- [x] fleet - By @cubells - #831 
- [x] gamification - By @cubells - #830 
- [x] gamification_sale_crm - By @cubells - #832 
- [x] google_account
- [x] google_calendar
- [x] google_drive
- [x] google_spreadsheet
- [x] hr - Done by @coleste, @hbrunn
[...]

PR #532, opened 2016-06-22
[ADD][9.0] migration script for module 'bus'

JordiBForgeFlow:


hbrunn:
Don't PR empty branches. Reopen when you have code to review.

hbrunn:
In case this is meant as 'Nothing to do for bus', I completely agree. In this case, file a PR that updates https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/doc/source/modules80-90.rst accordingly and mark the module as done in https://github.com/OCA/OpenUpgrade/issues/400 (all module PRs should also contain a change in the file mentioned above so that https://doc.therp.nl/openupgrade/modules80-90.html reflects the correct status)

PR #526, opened 2016-06-20
[ADD][9.0] migration script for module 'product'

JordiBForgeFlow:

PR #534, opened 2016-06-22
[ADD][9.0] migration script for module 'sale'

JordiBForgeFlow:


addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -1,36 +1,41 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenUpgrade module for Odoo
-#    @copyright 2014-Today: Odoo Community Association
-#    @author: Sylvain LE GAL <https://twitter.com/legalsylvain>
```
StefanRijnhart, line 6: Removing copyrights on an existing file is a death sin. Note that original author was @coleste who put in the wrong copyrights

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,199 @@
...
+
+sale         / sale.order               / order_policy (selection)      : DEL required: required, selection_keys: ['manual'], req_default: manual
+#Nothing To Do : Functionality removed now no longer needed.
+
+sale         / sale.order               / payment_term (many2one)       : was renamed to payment_term_id [nothing to to]
+#Nothing To Do : ORM will change the name automatically.
+
+sale         / sale.order               / section_id (many2one)         : DEL relation: crm.case.section
+sale         / sale.order               / team_id (many2one)            : NEW relation: crm.team
+#TODO : section_id is now team_id
```
StefanRijnhart, line 54: Remove TODO's if you think they are covered.

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,199 @@
...
+#Done : Created Service type product and assigned sale.order.line where product_id is null
+
+sale         / sale.order.line          / product_uos (many2one)        : DEL relation: product.uom
+#TODO : Moved to product_uos module (so saved it for later use.)
+
+sale         / sale.order.line          / product_uos_qty (float)       : DEL 
+#TODO : Moved to product_uos module (so saved it for later use.)
+
+sale         / sale.order.line          / qty_delivered (float)         : NEW 
+#Nothing To Do : New Functionality added.
```
StefanRijnhart, line 81: Migration needed. Unless you can show that this field is meaningless in existing orders after the migration, we need to try and retrieve the value for every line on every open sale order.

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -1,36 +1,41 @@
...
+    'sale_order': [
+        ('section_id', 'team_id'),
+    ],
+    'account_invoice': [
+        ('section_id', 'team_id'),
+    ],
     'sale_order_line_invoice_rel': [
         ('invoice_id', 'invoice_line_id'),
     ],
+    # These columns are moved to product_uos module so they are kept to be recovered later.
```
StefanRijnhart, line 19: So don't rename them. Check if they are used on recent orders, and in that case set product_uos to 'to install'.

serpentcs-dev1, line 19: I guess it is already covered in the product module pre-migration script: 
https://github.com/OCA/OpenUpgrade/pull/526/commits/b44dff76265257f16c4d3386a5ab55a5a6ecde5b#diff-0cf69fc4c7c5c5d1564d0493294d6471R81

StefanRijnhart, line 19: Yes, good find. So we just have to not rename them.

StefanRijnhart:
Superseded by #568

PR #540, opened 2016-06-23
[ADD] account

hbrunn:
this is #528 with a small fix to make if pass runbot hopefully


hbrunn:
@OCA/openupgrade-maintainers yay, it's green!

StefanRijnhart:
In order not to break the optimistic flow, can you please remove all the done TODOs so that we can easily grep for what is really left to do?

hbrunn:
@StefanRijnhart done

StefanRijnhart:
Thanks!

PR #546, opened 2016-06-29
[ADD] fix copyright notice

coleste:
Fixing bad copyright copypasta. Is this the right format?


StefanRijnhart:
Thanks. OCA prefers the short version now (as in https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py) but I can live with this.

PR #549, opened 2016-07-05
[ADD] finish project migration

hbrunn:


hbrunn:
I'm not sure if this migration should install project_timesheet if there are `project.task.work` records. If `project_timesheet` is installed, things will go right anyways (I'm working on this migration right now). But if not, this probably was a conscious choice, then installing this during migration is not what we want right?

StefanRijnhart:
Maybe install it iff there are timesheet modules installed?

hbrunn:
@StefanRijnhart such as https://github.com/OCA/OpenUpgrade/pull/554 ?

PR #551, opened 2016-07-05
[ADD] project_timesheet

hbrunn:


karlis-dreizis:
After a migration account.analytic.line's records no longer have sheet_id if the timesheet was in a closed/done state. Not sure where would be the best place for these statements, but I put them here.

```
diff --git a/addons/project_timesheet/migrations/9.0.1.0/post-migration.py b/addons/project_timesheet/migrations/9.0.1.0/post-migration.py
index 4521b01..41c853b 100644
--- a/addons/project_timesheet/migrations/9.0.1.0/post-migration.py
+++ b/addons/project_timesheet/migrations/9.0.1.0/post-migration.py
@@ -7,6 +7,12 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(cr, version):
     cr.execute(
+       'update account_analytic_line '
+       'set is_timesheet=True '
+       'from account_analytic_account a '
+       'where account_id=a.id and a.use_timesheets=True')
+
+    cr.execute(
         'update account_analytic_line '
         'set task_id=work.task_id '
         'from '
@@ -15,3 +21,12 @@ def migrate(cr, version):
         'join hr_analytic_timesheet t '
         'on w.hr_analytic_timesheet_id=t.id) work '
         'where work.line_id=account_analytic_line.id')
+
+    cr.execute(
+       'update account_analytic_line l '
+       'set sheet_id=('
+               'select id from hr_timesheet_sheet_sheet '
+               'where l."date" > date_from '
+                       'and l."date" < date_to '
+                       'and l.user_id = user_id) '
+       ' where is_timesheet=True')
```

hbrunn:
thanks @InfernalLV, but the statements belong into `hr_timesheet`'s and `hr_timesheet_sheet`'s migration scripts respectively. And with both, a join with `hr_analytic_timesheet` should give more reliable results.

PR #552, opened 2016-07-05
[ADD] nothing to do for document

hbrunn:


pedrobaeza:
Can you please create the openupgrade_analysis_work.txt as usual for this module? (although it contains only comments for nothing to do)

hbrunn:
that would be a bit pointless in my opinion. Let's discuss with @OCA/openupgrade-maintainers how we handle that. The first two nothing to do modules I see in the list don't have it either: https://github.com/OCA/OpenUpgrade/tree/9.0/addons/base_action_rule/migrations/9.0.1.0, https://github.com/OCA/OpenUpgrade/tree/9.0/addons/barcodes/migrations/9.0.2.0

So if we decide we want to have that, I'll add those files in the PR too.

StefanRijnhart:
No need, I think. :+1:

PR #553, opened 2016-07-05
[ADD] migration for crm

hbrunn:


JordiBForgeFlow:
Using this branch I cannot start odoo after the migration. I'm getting this error:

image

I can only proceed if I do the following statement:
delete from ir_ui_view v where name = 'im_chat assets'

Seems to me that the issue is related to the migration of module 'bus'.

StefanRijnhart:
Merging optimistically. @jbeficent cleanup of im_chat should be covered by a migration of the bus module. We can try to enter your statement in a preliminary migration script.

mmalorni:
We had issues with old views that would not get deleted automatically during the migration. We had to delete them manually but it was an error in the OpenUpgradeLib. @hbrunn I think did the fix and removed all the views cleanup from the migration scripts. Maybe this issue is related. https://github.com/OCA/OpenUpgrade/pull/489

PR #555, opened 2016-07-12
[MIG][9.0] runbot_relative: Update module from runbot_secure

lasley:
Module is being renamed in OCA/runbot-addons#87 so this PR is dependent on that. Wasn't sure when I was supposed to submit this in relation to the other PR, so I figured sooner rather than later.


pedrobaeza:
Well, this one is not installed in a regular DB, but it can be here as a reminder.

:+1:

StefanRijnhart:
Thanks!

PR #556, opened 2016-07-20
[ADD] stock migration

hbrunn:


StefanRijnhart:
How can the coverage drop be explained?

hbrunn:
no `stock.pack.operation` records in demo data I think: https://coveralls.io/builds/7086500/source?filename=addons%2Fstock%2Fmigrations%2F9.0.1.1%2Fpost-migration.py#L40

I'll add a test that adds some records there and runs the function, but not today

StefanRijnhart:
Thanks for the clarification.

PR #557, opened 2016-07-20
[ADD] Set lot params on picking type

StefanRijnhart:


StefanRijnhart:
OK, that was too soon...

StefanRijnhart:
Actually, you were right on time. I had rebased a post-hoc change seconds before you merged. I don't think you actually saw what you merged ;-)

PR #562, opened 2016-07-21
[ADD] inject some custom demo data in test database

hbrunn:
This is a proposal for a general approach to fix problems like https://github.com/OCA/OpenUpgrade/pull/556#issuecomment-233965167

To be done in the (n-1)st version, then we'd have to regenerate the test database (we should be able to automate that), and things work out. I'll create a PR soon that uses a database generated with this.

### DIFF
--- a/scripts/create_test_db
+++ b/scripts/create_test_db
@@ -9,11 +9,42 @@ VERSION=$($ODOO --version | awk '{ print $3 }')
 DBNAME="openupgrade-test-$VERSION-$(date +%F)"
 ADDONS=$(echo $(ls $ODOO_DIR/addons | grep --invert-match -E $IGNORE_ADDONS) | tr ' ' ,)
 DUMP=/tmp/openupgrade-test-$VERSION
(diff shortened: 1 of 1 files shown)


hbrunn:
#563 shows this works as expected

hbrunn:
@OCA/openupgrade-maintainers do you agree with replacing the current test database with the one generated by the modified script with this PR? This will give us 100% coverage builds again

dreispt:
Sounds good to me.
Wouldn't it be simpler to include an actual yml file in the scripts directory?

hbrunn:
@dreispt not really, because then the script has dependencies. Now, you can drop it wherever and it's going to work

pedrobaeza:
OK, it seems a bit ugly but you should be sure about no other better option

:+1:

dreispt:
OK, I think we can live with it, at least for now. :+1:

StefanRijnhart:
It's a start! :+1:

PR #564, opened 2016-07-21
[9.0] purchase requisition

JordiBForgeFlow:
This is the proposed migration script for the purchase_requisition module


hbrunn:
please also include an update to https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/doc/source/modules80-90.rst in your PRs

PR #568, opened 2016-07-25
[ADD] 9.0 sale

StefanRijnhart:


hbrunn:
good opportunity to provide extra demo data to test https://coveralls.io/builds/7157267/source?filename=addons%2Fsale%2Fmigrations%2F9.0.1.0%2Fpost-migration.py#L56. I give an example in #562

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,197 @@
+---Fields in module 'sale'---
+sale         / account.analytic.line    / so_line (many2one)            : NEW relation: sale.order.line
```
hbrunn, line 2: it might make sense to guess best matches, we should come quite far when matching the line's name and/or the product involved

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,197 @@
...
+
+sale         / sale.order.line          / product_id (many2one)         : now required
+#Done : Created Service type product and assigned sale.order.line where product_id is null
+
+sale         / sale.order.line          / product_uos (many2one)        : DEL relation: product.uom
+sale         / sale.order.line          / product_uos_qty (float)       : DEL 
+#Moved to product_uos module
+
+sale         / sale.order.line          / qty_delivered (float)         : NEW 
+#Nothing To Do : New Functionality added.
```
hbrunn, line 71: can we infer that from the order's state and/or pickings involved?

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -0,0 +1,77 @@
...
+        WHERE track_service IS NULL;
+    """)
+
+
+def set_dummy_product(env):
+    product = env['product.product'].create({
+        'name': 'Any product',
+        'type': 'service',
+        'order_policy': 'manual',
+    })
```
hbrunn, line 43: how about using https://github.com/OCA/OCB/blob/8.0/addons/product/product_data.xml#L248 if it exists?

hbrunn:
a few comments, but no reason not to merge

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
StefanRijnhart, line 2: This is used to create sale orders for services or expenses. I don't think that functionality exists in Odoo 8.0.

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
StefanRijnhart, line 71: Yes, to be done in sale stock.

addons/sale/migrations/9.0.1.0/post-migration.py
StefanRijnhart, line 43: No, because that product might have very different semantics than this dummy product.

addons/sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
StefanRijnhart, line 2: That is, it replaces the to_invoice marker on the analytic line. So we should indeed create sale orders for existing lines with this marker set, as the migration of hr_timesheet_invoice which is no longer present). I guess doing it in the sale module is as good as any location.

StefanRijnhart, line 2: I've added a reminder task in https://github.com/OCA/OpenUpgrade/issues/400

PR #570, opened 2016-07-26
[FIX][9.0] select on incoterm column fails if sale_stock is not insta…

JordiBForgeFlow:
When executing migration with module 'sale' installed, but module 'sale_stock' not installed the following error appears:
ProgrammingError: column "incoterm" does not exist
LINE 1: SELECT incoterm FROM sale_order WHERE incoterm IS NOT NULL

This PR fixes that.

### DIFF
--- a/addons/sale/migrations/9.0.1.0/post-migration.py
+++ b/addons/sale/migrations/9.0.1.0/post-migration.py
@@ -50,11 +50,15 @@ def set_dummy_product(env):
 def set_incoterm(env):
     """ Set the new Incoterm configuration option if incoterms
     are used at all in this database. """
+
+    if not openupgrade.is_module_installed(env.cr, 'sale_stock'):
+        return
     env.cr.execute("SELECT incoterm FROM sale_order "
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
cc @StefanRijnhart @hbrunn

hbrunn:
better use https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1230 and exit otherwise

JordiBForgeFlow:
@hbrunn Done. You can merge now.

hbrunn:
small nag: When a condition determines if a function does something at all (or for the same reasons, a loop body or the like), don't put the whole code in the scope of the condition, but say

``` python
if not condition:
    return
```

Especially with python, we want to save indents as much as possible.

Then also squash your commits to one

JordiBForgeFlow:
@hbrunn Changed as per your suggestion.

StefanRijnhart:
Thanks!

PR #574, opened 2016-07-28
[ADD] crm_claim migration

hbrunn:


JordiBForgeFlow:

StefanRijnhart:
Should we maybe branch off the crm.case.categ's that were used as categ_id on any claims as items of the new crm.claim.category model?

hbrunn:
@StefanRijnhart do you mean something else than https://github.com/OCA/OpenUpgrade/pull/574/files#diff-39a14eae8c8a7e993a88c6b386b3b968R12 ?

StefanRijnhart:
Thanks, I had not even looked at that. It just occurred to me that the XML ids needed some more work. I created https://github.com/OCA/OpenUpgrade/pull/595 for that.

PR #579, opened 2016-07-28
[UPD] docs and analysis_work for project_issue

hbrunn:


addons/project_issue/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -4,7 +4,7 @@ project_issue / project.issue            / tag_ids (many2many)           : NEW r
 # DONE
 
 project_issue / project.issue            / kanban_state (selection)      : now required, default = normal
-# Nothing to do
+# set default value
```
pedrobaeza, line 7: Where is the script for setting this default value?

hbrunn, line 7: in the existing code. The author just failed to update the work file: https://github.com/hbrunn/OpenUpgrade/blob/c8d4505e4f43b55ca0c2fd1775883bcd5a4f07d3/addons/project_issue/migrations/9.0.1.0/post-migration.py#L36

PR #581, opened 2016-07-28
[9.0][hr_timesheet]

JordiBForgeFlow:
Migration script for module hr_timesheet to v9


addons/hr_timesheet/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,39 @@
+---Fields in module 'hr_timesheet'---
+hr_timesheet / account.analytic.line    / is_timesheet (boolean)        : NEW 
+#TODO : Set all analytic line to true where its analytic account use_timesheet is true.
```
hbrunn: delete those when it's done (rationale: We don't want false positives when we grep for TODO in the work files)

addons/hr_timesheet/migrations/9.0.1.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
...
+        UPDATE account_analytic_line l
+        SET is_timesheet = a.use_timesheets
+        FROM account_analytic_account a
+        WHERE a.id = l.account_id""")
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    set_timesheet(cr)
+
```
hbrunn: https://travis-ci.org/OCA/OpenUpgrade/builds/148031969#L269

StefanRijnhart:
I think this logic is wrong. Follow-up PR here: https://github.com/OCA/OpenUpgrade/pull/587

PR #583, opened 2016-07-28
[ADD] account_asset migration

hbrunn:


JordiBForgeFlow:

PR #584, opened 2016-07-28
[ADD] nothing to do for account_check_printing

hbrunn:


JordiBForgeFlow:

StefanRijnhart:
Wondering if you considered renaming the old account_check_writing to this new module? There does not seem to be a shared data model, but they both cover the domain of check printing so it would be nice to have the new version available after the migration.

PR #587, opened 2016-07-29
[FIX] Set is_timesheet on analytic lines

StefanRijnhart:
Setting the is_timesheet marker only for analytic lines that were linked to an actual timesheet line in 8.0. The previous solution would assign the marker to all analytic lines for any project that uses timesheets, but that setting 'use_timesheet' on analytic accounts does not preclude that other analytic lines are present.

### DIFF
--- a/addons/hr_timesheet/migrations/9.0.1.0/tests/test_hr_timesheet.py
+++ b/addons/hr_timesheet/migrations/9.0.1.0/tests/test_hr_timesheet.py
@@ -1,0 +1,11 @@
+# coding: utf-8
+from openerp.tests.common import TransactionCase
+
+
+class TestHrTimesheet(TransactionCase):
+    def test_hr_timesheet(self):
+        self.env.cr.execute(
+            "SELECT line_id FROM hr_analytic_timesheet LIMIT 1")
+        line_id, = self.env.cr.fetchone()
+        line = self.env['account.analytic.line'].browse(line_id)
+        self.assertTrue(line.is_timesheet)
--- a/addons/hr_timesheet/migrations/9.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/9.0.1.0/post-migration.py
@@ -2,6 +2,7 @@
 # © 2015 Eficent Business and IT Consulting Services S.L. -
 # Jordi Ballester Alomar
 # © 2015 Serpent Consulting Services Pvt. Ltd. - Darshan Patel
+# © 2016 Opener B.V. - Stefan Rijnhart
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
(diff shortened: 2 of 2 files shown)


addons/hr_timesheet/migrations/9.0.1.0/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,11 @@
...
+from openerp.tests.common import TransactionCase
+
+
+class TestHrTimesheet(TransactionCase):
+    def test_hr_timesheet(self):
+        self.env.cr.execute(
+            "SELECT line_id FROM hr_analytic_timesheet LIMIT 1")
+        line_id, = self.env.cr.fetchone()
+        line = self.env['account.analytic.line'].browse(line_id)
+        self.assertTrue(line.is_timesheet)
```
pedrobaeza, line 11: You should test also a line that shouldn't have a related timesheet.

StefanRijnhart, line 11: Yes, I did that but there is no such line with standard demo data.

pedrobaeza, line 11: Should we inject them with @hbrunn's technique?

StefanRijnhart, line 11: It is injected in another branch

StefanRijnhart, line 11: .. I believe, which means it has to be proposed and merged separately and then the database has to be created manually and uploaded. After that, the test can be run.

I feel this method still needs some work to be usable....

pedrobaeza, line 11: Ok, let's continue with this one as is then

:+1:

hbrunn, line 11: @StefanRijnhart the complication is mainly caused by your refusal to add the database files to the repository. If they would live in the repo, all of this would be doable in one PR

StefanRijnhart, line 11: @hbrunn how would that work?

hbrunn, line 11: my original idea was:
1. you push the script
2. notice that the test database doesn't cover all code
3. push new commit adding the data in the database generation script
4. push another commit that adds the test database generated with the new version of the script

The different commits allow us to squash the database commits so that they only live once in the commit history, in case we want to save space and bandwidth

StefanRijnhart, line 11: Thanks for your expansion of the idea. I'm proposing to have the extra test data in the 9.0 branch here: https://github.com/OCA/OpenUpgrade/pull/624

PR #593, opened 2016-07-31
[ADD] reset window actions context (contains project_id)

hbrunn:


pedrobaeza:
I remember this bug because data is not reset when there are omitted values...

:+1:

StefanRijnhart:
This is usually detected by the compare xml script, but it seems that in this case the context was set in portal_project_issue (can't seem to find a context here : https://github.com/odoo/odoo/blob/8.0/addons/project_issue/project_issue_menu.xml#L8)

StefanRijnhart:
Ah, this is 7.0. That explains.

hbrunn:
yes, that's meant to get rid of https://github.com/odoo/odoo/blob/6.1/addons/project_issue/project_issue_menu.xml#L14

Issue #577, opened 2016-07-28
[9.0][account] Analytic line views not removed during migration


JordiBForgeFlow:
Impacted versions: 9.0

Steps to reproduce:
Migrate a demo database in v8 with modules 'analytic' and 'account' installed.

Current behavior:
The form, search and tree views for model 'account.analytic.line.tree' are not deleted. 

Depending on how Odoo decides to render the views, this may result in an error  such as:
_File "/home/jordi/Odoo9/openerp/addons/base/ir/ir_ui_view.py", line 463, in raise_view_error     raise AttributeError(message) AttributeError: Field `journal_id` does not exist  Error context: View `account.analytic.line.form` [view_id: 566, xml_id: n/a, model: account.analytic.line, parent_id: n/a]_

Expected behavior:
The following views should be deleted during the migration process, or during the database cleanup:
account.view_account_analytic_line_tree
account.view_account_analytic_line_form
account.view_account_analytic_line_filter

Video/Screenshot link (optional):
![image](https://cloud.githubusercontent.com/assets/7683926/17205497/a8f095ac-54ab-11e6-9ffd-b256d35def4a.png)

![image](https://cloud.githubusercontent.com/assets/7683926/17205502/ad0232cc-54ab-11e6-8b59-4806032a5b23.png)

hbrunn:
have a look at the log of a successful migration: https://travis-ci.org/OCA/OpenUpgrade/builds/147954920#L4416

At the end, odoo will delete all records pointed to by obsolete xmlids from migrated models. Check out how this looks in your migration, probably you'll find a line where odoo tries to do this, but there are other views inheriting from it or some other reason leads to the view not being deleted.

JordiBForgeFlow:
Looking at Travis I can see:
2016-07-28 07:14:48,586 24245 INFO openupgrade openerp.addons.base.ir.ir_model: Deleting 824@ir.ui.view
2016-07-28 07:14:48,589 24245 INFO openupgrade openerp.sql_db: bad query: delete from ir_ui_view where id IN (824)
2016-07-28 07:14:48,590 24245 WARNING openupgrade openerp.addons.base.ir.ir_model: Could not delete obsolete record with id: 824 of model ir.ui.view

Them the following error message at the end of the log:
Error context:
View `account.analytic.line.form`
[view_id: 824, xml_id: n/a, model: account.analytic.line, parent_id: n/a]
[...]

PR #594, opened 2016-07-31
[IMP] don't delete the xmlid of an undeletable record

hbrunn:
This is a demo to see if the strategy mentioned in #577 actually works

Marked as invalid because we should get something like this via a merge with upstream. Keep until either this is merged in upstream or we decide to use this anyways.

### DIFF
--- a/addons/analytic/migrations/9.0.1.1/tests/test_analytic.py
+++ b/addons/analytic/migrations/9.0.1.1/tests/test_analytic.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestAnalytic(TransactionCase):
+    def test_analytic(self):
+        self.assertTrue(
+            all(
+                self.env['ir.ui.view'].search([
+                    ('model', '=', 'account.analytic.line'),
+                ]).mapped('xml_id')
+            )
+        )
--- a/openerp/addons/base/ir/ir_model.py
+++ b/openerp/addons/base/ir/ir_model.py
@@ -1217,6 +1217,7 @@ class ir_model_data(osv.osv):
         to_unlink = []
         ids.sort()
         ids.reverse()
+        id_by_data = {}
         for data in self.browse(cr, uid, ids, context):
             model = data.model
             res_id = data.res_id
@@ -1224,6 +1225,7 @@ class ir_model_data(osv.osv):
             pair_to_unlink = (model, res_id)
             if pair_to_unlink not in to_unlink:
                 to_unlink.append(pair_to_unlink)
+                id_by_data[pair_to_unlink] = data['id']
 
             if model == 'workflow.activity':
                 # Special treatment for workflow activities: temporarily revert their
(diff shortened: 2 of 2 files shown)


StefanRijnhart:
It seems hard to tell from CI if this fixes the problem. Can you add a test to check that no view for analytic lines remain without a proper xml id?

hbrunn:
good idea, let's see. Logging says it doesn't delete the xmlid in https://travis-ci.org/OCA/OpenUpgrade/builds/148643765#L6532, but trust is good, control is better.

StefanRijnhart:
So it works! Can we merge this?

hbrunn:
let's give Odoo SA a week or so to contemplate on the PR I made. If they don't merge until then,let's merge this. @jbeficent meanwhile can put this branch in hin buildout to test it

JordiBForgeFlow:
It's not deleting the record. The old record still exists, without an external ID

hbrunn:
by now it also keeps the xmlid as the test shows. Then it can be removed by usual database_cleanup operations.

JordiBForgeFlow:
When I do the actual migration I can see the records there:
image

I am not sure why the test passes. Perhaps the test in module 'analytic' is run before the module 'account' is migrated?

See:
Test in module analytic is executed here:
https://travis-ci.org/OCA/OpenUpgrade/builds/148815247#L1848

Migration of account module starts here:
https://travis-ci.org/OCA/OpenUpgrade/builds/148815247#L2113

pedrobaeza:
How is the state of this?

pedrobaeza:
@hbrunn should this finally get merged?

hbrunn:
I never came around to try this with @jbeficent's example, so I'm doubtful. But for some corner cases, I do think we need this (or something similar)

pedrobaeza:
@jbeficent please approve / merge this if it's correct. I don't know enough the problem it tries to solve.

mvaled:
@StefanRijnhart today I ran another migration with this branch merged (see https://github.com/merchise-autrement/openupgrade/tree/merchise-predevelop-9.0).

When running Odoo 9 (https://github.com/merchise-autrement/odoo/tree/merchise-develop-9.0) againts the DB I still get the following:

```
QWebException: Could not find /account_analytic_analysis/static/src/css/analytic.css defined in bundle 'web.assets_backend'
```

The same happens for `edi` and `im_chat`.


The xmlids of those modules were deleted:

```
$ psql -d mercurio -c "select name from ir_model_data where module='account.analytic.analysis'"
 name 
------
(0 rows)
```

Those were the removed xmlids I though #594 was about, but after reading more carefully, obsolete modules seem to be still a WIP: https://github.com/OCA/OpenUpgrade/issues/577#issuecomment-235903507.

I see that odoo/odoo#12954 accepted the patch with odoo/odoo@14ac3a975f67488887889365fba2c410dadc1be3 (a variation) for Odoo 10.  So I'm OK to have this merged.  But we still need to decide how to proceed with obsolote modules.

StefanRijnhart:
On the upstream PR, there is talk of a bug (without making it clear what it is). And these commits are older than the one on the PR on which the bug is supposedly fixed. Does this change still contain the bug?

StefanRijnhart:
The bug mentioned on the upstream PR is the one fixed here: https://github.com/odoo/odoo/commit/8f6b5fcf9e3f10054253f3e074fa2d40baca67fc#diff-678bd201dfe1a0c69add630cf73e76ba. It does not apply to this change. So, +1 for having this merged.

StefanRijnhart (approved):

pedrobaeza:
But that one was done for 11.0, not 9.0. Does it matter?

StefanRijnhart:
@pedrobaeza indeed, the bug was introduced in the refactoring to the new API, so it is not in 9.0 and it is not related to the change in this PR.

pedrobaeza:
OK, merging then.

MiquelRForgeFlow:
Merging this

PR #596, opened 2016-08-01
[FIX] pass the actual result

hbrunn:


hbrunn:
yay, it's red!

StefanRijnhart:
Thanks!

PR #597, opened 2016-08-01
[ADD] auth_signup migration

hbrunn:
@OCA/openupgrade-maintainers shouldn't we move xmlids for all modules in the mail migration? Or better leave this to each module in order to be able to check if we need to do something or not? (in this case, the mail is changed slightly, but not in a relevant way, so I think it's a bad idea to overwrite the noupdate data)


StefanRijnhart:
Already handled when the template model is renamed as per https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L345?

hbrunn:
thanks, force pushed a commit saying so

PR #604, opened 2016-08-01
[FIX] a voucher's tax id can be null

hbrunn:


addons/account_voucher/migrations/9.0.1.0/pre-migration.py
```diff
@@ -3,8 +3,15 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
+column_copies = {
+    'account_voucher': [
+        ('voucher_type', None, None),
```
JordiBForgeFlow: Use 'type' instead of 'voucher_type', as this is how the field was named in v8. This will fix the current error in Travis.

hbrunn: omg, what a misser. thanks!

hbrunn:
travis will stay red until we fix https://travis-ci.org/OCA/OpenUpgrade/builds/149410226#L1610

JordiBForgeFlow:
See https://github.com/OCA/OpenUpgrade/pull/607

JordiBForgeFlow:
@pedrobaeza Can you merge this one?

PR #603, opened 2016-08-01
[ADD] hr_payroll_account migration, nothing to do for hr_payroll

hbrunn:

PR #605, opened 2016-08-02
9.0 eficent hr expense fixes

JordiBForgeFlow:
The total_amount (a computed, stored field) was recomputed via direct sql statement, incorrectly. Now recompute using odoo's recompute() method.

### DIFF
--- a/addons/hr_expense/migrations/9.0.2.0/post-migration.py
+++ b/addons/hr_expense/migrations/9.0.2.0/post-migration.py
@@ -51,7 +51,7 @@ def hr_expense(cr):
                 create_date, write_date, create_uid, write_uid,
                 product_uom_id, unit_amount, quantity, product_id)
                 SELECT company_id, currency_id, journal_id, employee_id,
-                state, date, account_move_id, name, bank_journal_id, 0.00.
+                state, date, account_move_id, name, bank_journal_id, 0.00,
                 'own_account',
                 (select analytic_account from hr_expense_line where id =
                 %(a)s),
--- a/addons/hr_expense/migrations/9.0.2.0/post-migration.py
+++ b/addons/hr_expense/migrations/9.0.2.0/post-migration.py
@@ -2,7 +2,7 @@
 # © 2016 Eficent Business and IT Consulting Services S.L.
 # © 2016 Serpent Consulting Services Pvt. Ltd.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-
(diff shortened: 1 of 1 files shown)


addons/hr_expense/migrations/9.0.2.0/post-migration.py
```diff
@@ -51,7 +51,7 @@ def hr_expense(cr):
                 create_date, write_date, create_uid, write_uid,
                 product_uom_id, unit_amount, quantity, product_id)
                 SELECT company_id, currency_id, journal_id, employee_id,
-                state, date, account_move_id, name, bank_journal_id, 0.00.
+                state, date, account_move_id, name, bank_journal_id, 0.00,
```
hbrunn, line 53: this one fell on my feet too: https://github.com/OCA/OpenUpgrade/pull/603/files#diff-c8cd1d437f6fef2bc83ae6d87cd9f753R54

addons/hr_expense/migrations/9.0.2.0/post-migration.py
```diff
@@ -65,14 +65,13 @@ def hr_expense(cr):
...
-        decimal(18,2)) from hr_expense e where e.id=h.id)
-    """)
+    expenses = env['hr.expense'].search([])
+    env.add_todo(env['hr.expense']._fields['total_amount'], expenses)
+    env['hr.expense'].recompute()
 
 
 @openupgrade.migrate()
 def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
```
hbrunn: how about using `use_env=True`? https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.migrate
Then it will pass an environment instead of a cursor

JordiBForgeFlow:
How would you do that?

Now it's:
@openupgrade.migrate()
def migrate(cr, version):

If I change to:
@openupgrade.migrate()
def migrate(env, use_env=True, version):

, won't work.

I have not seen any example in previous migrations using the use_env=True

hbrunn:
say

``` python
@openupgrade.migrate(use_env=True)
def migrate(env, version):
    # etc.
```

JordiBForgeFlow:
The travis error is not caused by this module, but by error in account_voucher. https://github.com/OCA/OpenUpgrade/pull/606 fixes it.

JordiBForgeFlow:
There's some issue with this PR. Travis is not running. I'm closing this one.

PR #607, opened 2016-08-03
[FIX][9.0] Admin user should access technical features

JordiBForgeFlow:
Fixes the error by which the admin user is not assigned by default to group "Technical features".
Seems that during the upgrade the changes to XMLID 'group_user' of model 'res.group' (https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/security/base_security.xml#L45) were not updated, even when these records are not supposed to be "noupdate=True".

So, we treat them as if they were noupdate=True.

Perhaps this only occurs with "implied_ids"?¿ Not sure. Anyway this quick and dirty solution works.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
+++ b/openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
@@ -1,0 +1,8 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+    <data>
+        <record model="res.groups" id="group_user">
+            <field name="implied_ids" eval="[(4, ref('group_no_one'))]"/>
+        </record>
+    </data>
+</openerp>
--- a/openerp/addons/base/migrations/9.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/post-migration.py
@@ -121,3 +121,4 @@ def migrate(cr, version):
     set_filter_active(cr)
     precalculate_checksum(cr)
     remove_obsolete_modules(cr, ('web_gantt', 'web_graph', 'web_tests'))
+    openupgrade.load_data(cr, 'base', 'migrations/9.0.1.3/noupdate_changes.xml')


openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+    <data>
+        <record model="res.groups" id="group_user">
+            <field name="implied_ids" eval="[(4, ref('group_no_one'))]"/>
```
hbrunn, line 5: this will turn on the technical setting for every user. What's the rationale here as opposed to assign this to groyp_system?

JordiBForgeFlow:
See https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/security/base_security.xml#L46
This group is only assigned in v9 to base.user_root

JordiBForgeFlow:
@pedrobaeza Can you merge this one? All checks passed.

JordiBForgeFlow:
cc @StefanRijnhart

elicoidal:
There are still some (valid) questions

openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
JordiBForgeFlow, line 5: @elicoidal If you refer to this question by @hbrunn, odoo 9 assigns group_no_one to everyone, as you can see here: https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/security/base_security.xml#L45

So all users have access to the group "Technical Features" in Odoo9 standard, by default. It's just that they have to activate the debugging mode in order to see them. I am not introducing any new element in the migration, just forcing that the group "group_user" behaves as it should on a migrated v9, as it would work for a plain vanilla v9.

JordiBForgeFlow:
@elicoidal @pedrobaeza We established an optimistic merge criteria at this stage. Meaning that if all checks passed on the PR, it can be merged. I need this merged because it's blocking me the progress to continue with other migration scripts. Lots of PR fail due to this PR not merged.

pedrobaeza:
@jbeficent, optimistic merge requires 2 things:
- All CIs passed
- No review comments unattended.

Said that, I agree with your reasoning and @hbrunn has been answered, so I merge.

JordiBForgeFlow:
Thanks!

StefanRijnhart:
Agree with the merge. Still puzzled why the code is not loaded using the standard mechanisms though.

PR #614, opened 2016-08-08
[FIX] Run init scripts for new dependencies

### LINKED ISSUE #613: [9.0] Init script of account_full_reconcile is not run
account_full_reconcile is a module that was added after the release of 9.0. It is installed automatically after a migration of a database containing 'account', because it is set to auto_install and depends on 'account' only. It includes its own post init script, but the init script is not run because auto-installed modules are installed in update mode and their (cached) state at that moment seems to be 'uninstalled' so this check fails: https://github.com/odoo/odoo/blob/9.0/openerp/modules/loading.py#L125

StefanRijnhart:
Fixes #613

### DIFF
--- a/openerp/openupgrade/openupgrade_loading.py
+++ b/openerp/openupgrade/openupgrade_loading.py
@@ -50,6 +50,7 @@ def add_module_dependencies(cr, module_list):
     if not module_list:
         return module_list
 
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
All checks passed. Merging.

PR #615, opened 2016-08-08
[FIX] Variable was rewritten in every loop iteration

StefanRijnhart:
[FIX] Variable was rewritten in every loop iteration, preventing proper recomputation.
[RFR] Style


JordiBForgeFlow:
Thanks @StefanRijnhart for spotting.

PR #618, opened 2016-08-09
[9.0][base] fixes to migration of manual fields

JordiBForgeFlow:
This PR fixes an error found during the migration of manually added x_\* fields.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -69,8 +69,14 @@ def remove_obsolete(cr):
         """.format(OBSOLETE_RULES))
 
 
+def pre_create_columns(cr):
+    openupgrade.logged_query(cr, """
+        alter table ir_model_fields add column compute text""")
+
+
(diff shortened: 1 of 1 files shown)


hbrunn:
please specify which error, and more important, why we can't wait until the ORM creates the column anyways during the update of the base module

JordiBForgeFlow:
As far as I can understand, method _setup_base from /openerp/modules/registry.py is called before the ORM is started:
https://github.com/odoo/odoo/blob/87d160a7423243ecd760fefe5ba22fb85c460cb6/openerp/modules/registry.py#L197

This method then calls to _add_manual_fields method:
https://github.com/odoo/odoo/blob/a1fbb3d3892dbc978c237e70e6b40315684643cb/openerp/models.py#L3020

And that method then expects the column 'compute' to exist:
https://github.com/odoo/odoo/blob/a1fbb3d3892dbc978c237e70e6b40315684643cb/openerp/models.py#L731

The column 'compute' has been introduced in v9.

hbrunn:
thanks, this important info could have been a comment in the code. Merging now, but feel free to add a comment there

Issue #502, opened 2016-06-06
OpenUpgrade buildout


hbrunn:
I'm currently toying with https://github.com/hbrunn/OpenUpgrade/tree/buildout hoping to streamline the way migrations go at our place. Would this be interesting as a branch for this repo? What I like about this approach is that it's a lot more foolproof than `migrate.py` and simpler to extend, provided you use buildouts anyways.

pedrobaeza:
Well, the buildout itself is by far more complicated that migrate.py I think... but it can be interesting. Let's see what others say.

StefanRijnhart:
Yes, definitely. I always bark when people request the project to make it 'easier' in their humble opinion to run OpenUpgrade, but if these efforts are contributed to the project it should host them.

pedrobaeza:
Closing this waiting for a PR. I would prefer anyway a Docker for migrating instead of buildout.

legalsylvain:
Hi @hbrunn. thanks for pointing this work. I'll try to take a look on that in a while. I used to work with buildout, for my deployment process.
kind regards.

hbrunn:
I'm using this for all our migrations now, and I'm quite happy with it

lord-enzo:
I can't make it work... `bin/buildout` gives me the following error

```
fatal: Couldn't find remote ref 9.0-base
While:
  Installing openupgrade9.
Unexpected end of command stream
```
[...]
`CalledProcessError: Command '['git', 'pull', '--no-edit', 'https://github.com/microcom/openupgrade.git', '9.0-base']' returned non-zero exit status 1`

BeAntares:
Hi @hbrunn Is it possible to migrate to version 10 using this buildout?
Thanks

hbrunn:
not as it is, but currently, we also don't have enough migration scripts done for 10 to migrate more than a very basic database anyways. When the migration project I'm working on currently is far enough so that it makes sense to run a migration on the real database, I'll probably update this branch.

ovnicraft:
Hi, i am working on migrations from 6.1 to 7.0,8.0,9.0,10.0,this discussion in which states lets openupgrade? can i use migrate.py or i must use buildot ?

pedrobaeza:
Up to 8.0, you can use migrate.py (anyway, I have never done it that way - just download OpenUpgrade and run server with `-u all`). From version 8, we are using our Docker system putting openupgrade as origin: https://github.com/Tecnativa/docker-odoo-base

PR #620, opened 2016-08-10
9.0 migrate.py update from 8.0

hbrunn:
I'm not going to support this script any more by the way because https://github.com/OCA/OpenUpgrade/issues/502 is so much more convenient for me. So you might also decide to delete all of this (and the references to it in the documentation)


pedrobaeza:
:+1:

Sekiltoyai:
From a user point of view, the modified version behaves correctly. Thanks !

StefanRijnhart:
Can we include your the buildout branch and deprecate the script?

PR #621, opened 2016-08-10
[IMP] resolves conflict when an account.tax records exists with type …

JordiBForgeFlow:
The tax type 'code' is not an option in the account module for v9. We need to assign a temporary 'dummy' value and install module account_tax_python. In post-migration we will restore the original value.


JordiBForgeFlow:
@hbrunn This one is ready to merge.

Issue #622, opened 2016-08-11
[9.0][account] Migration of fiscal year closing entries.


JordiBForgeFlow:
The journal entries generated by the Odoo standard fiscal year closing procedure are not closed in the previous FY and reopened in the next FY. Instead, the resulting balance of the previous FY is repeated in the FY. So, if you keep these journal entries in v9, and compute the BS dynamically, from the start of times until  the desired time, the result will be TOTALLY incorrect.

Companies that use the standard odoo FY closing procedure should, as a pre-migration manual adjustment,  move out of the special periods any journal entry used to record the intial balances of the company, created during the initial migration into Odoo.

Alternatively the pre-migration script should look for the initial year in the system, search for the first special period, and move it out of the special period, into the first general period (e.g. 01/20xx), assuming that this was the starting balance of the company.

PR #623, opened 2016-08-11
[IMP] migration of fiscal year opening journal entries

JordiBForgeFlow:
Implements the changes to account migration script to fix https://github.com/OCA/OpenUpgrade/issues/622


pedrobaeza:
As you know, this is not valid for Spain. You can include a condition to move (not delete) all the entries if the company country is Spain.

StefanRijnhart:
Can you rebase instead of merging to prevent these commits from showing up in the main branch history?
How are opening balances used in Spain? Can we detect this method regardless of the installed module?
Would you like to attempt to replace the subqueries by joins? Or I could do it after the optimistic merge.

JordiBForgeFlow:
@StefanRijnhart I need your help to rebase. I'm having trouble understanding how to squash the commits.

As for the opening balances, in spain a journal entry is created at the end of the year to set the P&L accounts to 0, and move the net profit to the balance sheet. There's really no way to detect this method, regardless of the installed module, at least that I know of.

As for replacing the subqueries by joins, please feel free to improve after an optimistic merge.

StefanRijnhart:
We can squash at merge time using the GitHub dropdown button next to 'Merge Pull Request'. You are right, we can optimize the code later. Also, in a further improvement this had better be done by company, as periods are company aware. :+1:

StefanRijnhart:
Merging this after too long of a delay to be optimistic

PR #624, opened 2016-08-14
[RFR] Load custom test data before test migration

StefanRijnhart:
Include test data for the test migration in the target version branch. Copy the test data outside the source tree. Switch branches to load the test data just-in-time. Switch back to run the migration.


StefanRijnhart:
@hbrunn what do you think? It needs https://github.com/odoo/odoo/pull/13146 and maybe a refactoring to allow for a directory with test data instead of a single file.

hbrunn:
this would replace the extra data in https://github.com/OCA/OpenUpgrade/blob/8.0/scripts/create_test_db#L18 ? Fine I think, but if we patch the server anyways, can't we make it load extra demo data before pre migration for every module? This way, we don't need to worry about dependencies and can place the file(s) in the appropriate `migrations` folder, thereby avoiding clutter. But let's only do this if the diff for this is similarly small as the patch you use here.

StefanRijnhart:
Yes, this serves to include the required test data in the same PR as the migration (test) scripts. The server cannot be patched to load the data before pre-migration, because the data needs to be loaded by the previous major release of Odoo (the travis code swaps the branches in order to load it). I was thinking of a test_data.d/ directory to avoid clutter, but not taking into account any dependencies because the reference database already has all modules available.

hbrunn:
sounds good. Ping me when you want this to be merged

StefanRijnhart:
@hrunn thanks. Let's start with this one https://github.com/OCA/OpenUpgrade/pull/627/files

StefanRijnhart:
@hbrunn you can now store the test data for the previous edition as $addon/migrations/$version/tests/*.yml.

hbrunn:
this looks great! Why do you need to rewind to 8 for loading the yaml data? Doesn't 9 do the same for loading concerns?

StefanRijnhart:
Well, given that we start out with an 8.0 database and we want the data in there before the migration, loading it using version 9.0 (by necessity after the migration of the database) kind of defeats the purpose!

hbrunn:
so I went back to this issue on my own volition (believe it or not) to retract my stupid question. Of course!

PR #627, opened 2016-08-16
[ADD] Allow to run test file without post tests

StefanRijnhart:
The OpenUpgrade test mechanism should be able to load test data more dynamically and incrementally. It uses --test-file + --test-commit to load these files. However, the post tests cannot be committed multiple times so I am proposing to skip these tests when only using --test-file without --test-enable.

I have also proposed this into Odoo as a general improvement as https://github.com/odoo/odoo/pull/13146. 

This change is a prerequisite for https://github.com/OCA/OpenUpgrade/pull/624, which allows Travis to apply test data for the migration of 8.0 to 9.0 that is included in the 9.0 distribution (so that test data can be added in the same PR as the migration scripts).

### DIFF
--- a/openerp/cli/server.py
+++ b/openerp/cli/server.py
@@ -134,8 +134,10 @@ def main(args):
         except openerp.service.db.DatabaseExists:
             pass
 
-    if config["test_file"]:
-        config["test_enable"] = True
+    # OpenUpgrade: don't run post tests if --test-enable is not specified
+    # explicitely (problematic when loading test_data with --test-commit)
+    # if config["test_file"]:
+    #     config["test_enable"] = True
 
     if config["translate_out"]:
         export_translation()
--- a/openerp/service/model.py
+++ b/openerp/service/model.py
@@ -113,7 +113,9 @@ def check(f):
         tries = 0
         while True:
             try:
-                if openerp.registry(dbname)._init and not openerp.tools.config['test_enable']:
+                # OpenUpgrade: test_file does not imply test_enable as it does in upstream Odoo
+                test_mode = openerp.tools.config['test_enable'] or openerp.tools.config['test_file']
+                if openerp.registry(dbname)._init and not test_mode:
                     raise openerp.exceptions.Warning('Currently, this database is not fully loaded and can not be used.')
                 return f(dbname, *args, **kwargs)
             except (OperationalError, QWebException) as e:


StefanRijnhart:
Please review, not merge optimistically.

openerp/service/model.py
```diff
@@ -113,7 +113,8 @@ def _(src):
         tries = 0
         while True:
             try:
-                if openerp.registry(dbname)._init and not openerp.tools.config['test_enable']:
+                test_mode = openerp.tools.config['test_enable'] or openerp.tools.config['test_file']
```
pedrobaeza, line 117: Should this be marked as OpenUpgrade specific change with a comment or something?

StefanRijnhart, line 117: Done!

openerp/cli/server.py
```diff
@@ -134,8 +134,10 @@ def main(args):
         except openerp.service.db.DatabaseExists:
             pass
 
-    if config["test_file"]:
-        config["test_enable"] = True
+    # OpenUpgrade: don't run post tests if --test-enable is not specified
+    # explicitely (problematic when when loading test_data with --test-commit)
```
pedrobaeza: Double when

StefanRijnhart: Fixed!

legalsylvain:
LGTM. Thanks for this improvement :+1:

StefanRijnhart:
@hbrunn can I have your review? This is the change that warrants a custom 8.0 branch in #624. With this change, we can load the testing data with the main branch.

hbrunn:
I think this will work just fine, but are we now stacking patch after patch to make the original idea of generating a test database work? Maybe we'll be better off to think of a different way of generating test databases. Don't have a better idea right now, so :+1:

StefanRijnhart:
No, this is the only patch needed. After this is merged (someone, please!) we can start including test data in the same branch as the migration script, which is then injected just-in-time by Travis along the lines of #624.

pedrobaeza (approved):

StefanRijnhart:
Thank you @pedrobaeza ;-) I will cherry-pick this onto the other branches if you don't mind.

pedrobaeza:
OK, but we aren't for now making tests in the rest.

StefanRijnhart:
Backported to 8.0, where we actually need it the most right now. For 10.0, it seems that test_file no longer implies test_enable. It may still raise _this database is not fully loaded and can not be used._ but we can fix it when it comes to that.

PR #633, opened 2016-09-09
[9.0] [base] Remove pre-calculate checksum

JordiBForgeFlow:
The 'checksum' field is not a required field. To pre-calculate checksums on all existing attachments is a very time consuming process. I have experienced that in databases with hundreds of thousands of attachments the migration lasts for more than 2 days just calculating the checksum, which is a totally unacceptable time delay for a migration.

We have migrated a database excluding the calculation of checksum, and the attachments can be correctly recovered.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/post-migration.py
@@ -21,7 +21,6 @@
 ##############################################################################
 
 import logging
-import base64
 from openupgradelib import openupgrade
 from openerp.modules.registry import RegistryManager
 from openerp import SUPERUSER_ID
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
I could not manage to squash the commits..... I think that this can be done in the merge PR using Github.

pedrobaeza:
OK, I don't see any use of this field.

:+1:

simahawk:
Sorry, I don't have the whole picture here but this note could be useful :)

Looks like the field is used for caching, see a comment in `_compute_checksum` into  `base/ir/ir_attachment.py` and it's used to generate a unique link that changes as the file changes, see `server/addons/web_editor/models/ir_attachment.py` ->
`'/web/image/%s?unique=%s' % (attach.id, attach.checksum)`.

If the computation has not been changed - as it looks to me - LGTM: :+1:

pedrobaeza:
Thanks for the tip on the usage, Simone!

PR #635, opened 2016-09-15
[ADD] product_images was renamed for 8

hbrunn:
cf https://github.com/OCA/product-attribute/tree/8.0/product_multi_image and https://github.com/OCA/product-attribute/tree/7.0/product_images


pedrobaeza:
Well, it's not exactly a rename, but a new module for the same feature, so I'm not sure of this one. The new module also depends on server-tools/base_multi_image.

hbrunn:
yes, exactly. I'm busy writing an init hook for `base_multi_image` that will move the `product_images` table with the necessary adjustments if it exists on installation time. With the hook and this identity, an upgrade from 7 to 8 should go automatically.

pedrobaeza:
OK then

:+1:

legalsylvain:
thanks. :+1:

PR #636, opened 2016-09-19
[FIX] adapt documentation to reality

hbrunn:
the config option doesn't have the `d`: https://github.com/OCA/OpenUpgrade/blob/a92ae71ef807ef58e9b6ac436135a531ecd49fac/openerp/modules/graph.py#L89


pedrobaeza:
:+1:

PR #638, opened 2016-09-27
[FIX] stock_account: Don't fail on period not found in inventory_period_id

pedrobaeza:
You can make inventories without having a "Force account period" set, and even you can use inventory without account and thus not having created accounting periods.

### DIFF
--- a/addons/stock_account/migrations/8.0.1.1/post-migrate.py
+++ b/addons/stock_account/migrations/8.0.1.1/post-migrate.py
@@ -22,6 +22,7 @@
 ##############################################################################
 
 from openerp.openupgrade import openupgrade
+from openerp.exceptions import RedirectWarning
 from openerp import pooler, SUPERUSER_ID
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
cc @Tecnativa

addons/stock_account/migrations/8.0.1.1/post-migrate.py
```diff
@@ -91,10 +91,15 @@ def inventory_period_id(cr, pool):
...
-            cr, SUPERUSER_ID, dt=date[:10], context={'company_id': company_id})
-        inventory_obj.write(
-            cr, SUPERUSER_ID, inv_id, {'period_id': period_ids[0]})
+        try:
+            period_ids = period_obj.find(
+                cr, SUPERUSER_ID, dt=date[:10],
+                context={'company_id': company_id})
+            inventory_obj.write(
+                cr, SUPERUSER_ID, inv_id, {'period_id': period_ids[0]})
+        except:
```
hbrunn: better catch only `openerp.exceptions.RedirectWarning`

StefanRijnhart:
@pedrobaeza can you catch the specific exception that Holger mentioned? Then it's :+1: for me.

pedrobaeza:
Done

StefanRijnhart:
Thanks! Merging because Holger's comment was honoured.

PR #645, opened 2016-10-15
9.0 account voucher

NL66278:
## This PR is to add the migration of account_voucher rows containing payments to account_payment.

### DIFF
--- a/addons/account_voucher/migrations/9.0.1.0/post-migration.py
+++ b/addons/account_voucher/migrations/9.0.1.0/post-migration.py
@@ -4,44 +4,10 @@
 from openupgradelib import openupgrade
 
 
-def get_ref(cr, module, xml_id):
-    """Get reference (res_id) from module and xml_id from ir_model_data."""
-    cr.execute("""
-        SELECT res_id
-        FROM ir_model_data
-        WHERE module = %s AND name = %s
-    """, (module, xml_id))
-    return cr.fetchone()[0]
-
-
-def migrate_voucher_types(cr):
-    """Migrate voucher types."""
-    # TODO: Should we really keep receipt and payment records? Might contain
-    # duplicate info with the sales and purchases for which they have been
-    # made, especially now that we migrate to account.payment.
-    openupgrade.map_values(
-        cr, openupgrade.get_legacy_name('type'), 'voucher_type',
-        [('receipt', 'sale'), ('payment', 'purchase')],
-        table='account_voucher')
-
-def create_voucher_line_tax_lines(cr):
-    """Migrate tax information on voucher lines to m2m relation."""
-    cr.execute(
-        "insert into account_tax_account_voucher_line_rel "
-        "(account_tax_id, account_voucher_line_id) "
-        "select v.tax_id, l.id "
-        "from account_voucher v "
-        "join account_voucher_line l on l.voucher_id=v.id "
-        "where v.tax_id is not null")
-
-def create_payments_from_vouchers(cr):
+def create_payments_from_vouchers(env):
     """Create account.payment rows from appropiate account.voucher rows."""
-    receipt_method = get_ref(
-        cr, 'account', 'account_payment_method_manual_in'
-    )
-    payment_method = get_ref(
-        cr, 'account', 'account_payment_method_manual_out'
-    )
+    receipt_method = env.ref('account.account_payment_method_manual_in')
+    payment_method = env.ref('account.account_payment_method_manual_out')
     # Keep original id, to later migrate references from other tables,
     # although it seems not te be needed for now.
     # Possible states of account.payment: draft, posted, sent, reconciled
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Check flake8 errors:

```
./addons/account_voucher/migrations/9.0.1.0/post-migration.py:27:1: E302 expected 2 blank lines, found 1
./addons/account_voucher/migrations/9.0.1.0/post-migration.py:37:1: E302 expected 2 blank lines, found 1
./addons/account_voucher/migrations/9.0.1.0/post-migration.py:88:9: E128 continuation line under-indented for visual indent
```

and can you please add some tests?

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -4,16 +4,102 @@
...
+        """SELECT COALESCE(MAX(id), 0) AS max_id FROM account_payment"""
+    )
+    max_id = cr.fetchone()[0] + 1
+    cr.execute(
+        """ALTER SEQUENCE account_payment_id_seq RESTART WITH %s""",
+        (max_id,)
+    )
+
+
+@openupgrade.migrate()
```
legalsylvain: Please use

```
@openupgrade.migrate(use_env=True)
def migrate(cr, version):
    cr = env.cr
```

sample : https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/post-migration.py#L351

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -4,16 +4,102 @@
...
         "insert into account_tax_account_voucher_line_rel "
         "(account_tax_id, account_voucher_line_id) "
         "select v.tax_id, l.id "
         "from account_voucher v "
         "join account_voucher_line l on l.voucher_id=v.id "
         "where v.tax_id is not null")
+
+def create_payments_from_vouchers(cr):
+    """Create account.payment rows from appropiate account.voucher rows."""
+    receipt_method = get_ref(
```
legalsylvain: Then your function get_ref is unnecessary. Just use : 

```
env.ref('account.account_payment_method_manual_in')
```

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -4,16 +4,102 @@
...
+    # Keep original id, to later migrate references from other tables,
+    # although it seems not te be needed for now.
+    # Possible states of account.payment: draft, posted, sent, reconciled
+    # Possible states of account_voucher: draft, cancel, proforma, posted
+    # "type" field in 8.0 renamed to "voucher_type" during migration.
+    # Field destination_journal_id is used for internal transfer payments.
+    # Not used here.
+    # As far as I can see there is no connection between invoices or moves
+    # in 8.0 taht we can or have to migrate to 9.0.
+    cr.execute("""
```
legalsylvain: why dont you use orm ? 

```
env['account.payment'].create({})
```

addons/account_voucher/migrations/9.0.1.0/pre-migration.py
```diff
@@ -10,8 +10,18 @@
 }
 
 
+def delete_payment_views(cr):
```
legalsylvain, line 13: Why is this function necessary. View error in openupgrade are just warning.

legalsylvain (commented):
Hi, 
Thanks for your contribution. Some technical remarks inline.
For payment migration, I don't know enough Odoo9 to review.
Regards.

addons/account_voucher/migrations/9.0.1.0/pre-migration.py
NL66278: This view was calling a failure during testing, because it referred to no longer existing fields, during upgrade of the newly migrated database.

addons/account_voucher/migrations/9.0.1.0/post-migration.py
NL66278: Will do.

addons/account_voucher/migrations/9.0.1.0/post-migration.py
NL66278: Thanx for pointing to this function.

addons/account_voucher/migrations/9.0.1.0/post-migration.py
NL66278: I think using plain SQL here is much faster, but above all more clear/clean/readable.

NL66278:
@pedrobaeza Is you have seen this PR is also included in the PR for account_check_printing. I did correct the flake8 errors. As for the tests, I agree with the desirability of unit tests in general, but I do not quit see how it would make a difference here. If I made some logical mistakes in how to convert account_voucher to account_payment, the tests I would think up would probably reflect the same mistakes.

pedrobaeza:
Yeah, you're probably right about some tests in general in OpenUpgrade, but sometimes having original records in one side and result records in others, and comparing them, you get several things:
- Awkward differences that you don't expect.
- Errors in the process itself.

But I'm not going to push you to develop the tests.

pedrobaeza (approved):

Issue #646, opened 2016-10-18
Rename of module account_check_writing to account_check_printing has no effect


NL66278:
Impacted versions: migration of 8.0 to 9.0

Steps to reproduce: Run migration script on a database with account_check_writing installed

Current behavior: Module account_check_writing (which no longer exists) will have status 'to upgrade'. module account_check_printing will be uninstalled

Expected behavior: Module account_check_printing having status installed and module account_check_writing deleted from list of modules (table ir_module_module). Migration scripts in account_check_printing should be called.

The rename of account_check_writing to account_check_printing is listed in openerp/addons/openupgrade_records/lib/apriori.py renamed_modules. But that does not seem to have any effect.

pedrobaeza:
For other renamings, I haven't got any problem, so I'm not sure if something went bad in your migration that unrolled the change or what...

NL66278:
@pedrobaeza Is this for 8.0 to 9.0? Strangely enough I did not find any logic to actually process the renaming at all. Only in the analysis. But I might have looked at the wrong places...

StefanRijnhart:
Closing, as fix is merged

PR #647, opened 2016-10-18
[ENH] Actually process module renames.

NL66278:
https://github.com/OCA/OpenUpgrade/issues/646#issuecomment-254475025

Current behavior before PR: Modules not renamed

Desired behavior after PR is merged: Modules renamed
## This PR fixes the rename problem for me. I understood that for others the module rename functionality works. If so, I would like to know how. Creating this PR to hopefully get some answers. And maybe it will be usefull for others with similar problems to mine.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -21,6 +21,7 @@
 ##############################################################################
 
 from openupgradelib import openupgrade
+from openerp.addons.openupgrade_records.lib import apriori
 
 
 column_copies = {
@@ -71,6 +72,9 @@ def remove_obsolete(cr):
 
 @openupgrade.migrate()
 def migrate(cr, version):
+    openupgrade.update_module_names(
+        cr, apriori.renamed_modules.iteritems()
+    )
     openupgrade.copy_columns(cr, column_copies)
     openupgrade.rename_columns(cr, column_renames)
     remove_obsolete(cr)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Is this code taken from 8.0? If so, I think the problem is that it's not marked with OpenUpgrade comment or similar to be detected, and it was forgotten in the migration.

hbrunn:
this seems missing in the base migration, in 8, this happens here: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/base/migrations/8.0.1.3/pre-migration.py#L39

NL66278:
In 9.0 the code called in 8.0 is no longer present.

hbrunn:
it lives in openupgradelib: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L703

hbrunn (approved):

NL66278:
@pedrobaeza @hbrunn I changed my PR to use the same logic as in 8.0. Tested this as well and it works (at least for me). Obvious difference is that there will be no longer a row in ir_module_module with the old module name and state 'uninstalled' (like before last commit in my PR). But renaming is much more thorough, because also other tables are updated from update_module_names that is called now.

pedrobaeza (approved):

pedrobaeza:
We should make the same for 10.0 in base module. @StefanRijnhart, please take into account if you're doing it.

StefanRijnhart:
Thanks. Maybe we should put it in the core somewhere instead of having to put in the base scripts every edition.

pedrobaeza:
Yeah, I have thought about that, but the problem is to not invoke the method each time you restart the migration process, which I don't see possible by now. Well, in fact, nothing will break, but executing a lot of update operations with 0 results each time doesn't seem to me the best method.

NL66278:
If it would be possible to do when the server collects the names of available modules, it would be possible to do this only for modules not yet renamed. Then for each module to be renamed update_module_names could be called with just a single tuple with old name and new name.

pedrobaeza:
Yeah, it can be a good solution.

StefanRijnhart:
Yes, but it's not the names of available modules, but the names of installed modules instead.

PR #648, opened 2016-10-18
9.0 account check printing

NL66278:
Description of the issue/feature this PR addresses: migrate amount in words for check from account_voucher to account_payment

Current behavior before PR: check_amount_in_words is empty in migrated payment

Desired behavior after PR is merged: check_amount_in_words in payment is filled from migrated account_voucher row, field amount_in_words.

This branch builds on the branch to migrate account_voucher data
## 

### DIFF
--- a/addons/account_check_printing/__openerp__.py
+++ b/addons/account_check_printing/__openerp__.py
@@ -10,7 +10,10 @@ It must be used as a dependency for modules that provide country-specific check 
 The check settings are located in the accounting journals configuration page.
     """,
     'website': 'https://www.odoo.com/page/accounting',
-    'depends' : ['account'],
+    'depends' : [
+        'account',
+        'account_voucher',  # Not for normal Odoo, but needed for migration
+    ],
     'data': [
         'data/check_printing.xml',
         'views/account_journal_dashboard_view.xml',
--- a/addons/account_check_printing/__openerp__.py
+++ b/addons/account_check_printing/__openerp__.py
@@ -8,6 +8,14 @@
 This module offers the basic functionalities to make payments by printing checks.
 It must be used as a dependency for modules that provide country-specific check templates.
 The check settings are located in the accounting journals configuration page.
+
+OpenUpgrade specific: we set account_voucher here as a dependency for
+account_check_printing. account_voucher was an actual dependency for
+account_check_writing in 8.0. Therefore, when upgrading this module, we can
+be sure account_voucher is also installed. But we have to make sure
+account_voucher is loaded before account_check_printing, because the migration
+of check information uses the results of account_voucher migration. To be
+specific: the account_payment records created by account_voucher migration.
     """,
     'website': 'https://www.odoo.com/page/accounting',
     'depends' : [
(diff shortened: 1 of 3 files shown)


addons/account_check_printing/__openerp__.py
```diff
@@ -10,7 +10,10 @@
 The check settings are located in the accounting journals configuration page.
     """,
     'website': 'https://www.odoo.com/page/accounting',
-    'depends' : ['account'],
+    'depends' : [
+        'account',
+        'account_voucher',  # Not for normal Odoo, but needed for migration
```
pedrobaeza, line 23: Does this forces the installation when you do `--update=all`? Maybe you need to update manually in base module the state to `to install` for this module when `account_check_printing` is installed

pedrobaeza (changes requested):
You have mixed commits from other PRs

JordiBForgeFlow:
The field 'number' in account_voucher v8 (https://github.com/odoo/odoo/blob/8.0/addons/account_check_writing/account_voucher.py#L41) should migrate to 'check_number' in account_payment in v9 (https://github.com/odoo/odoo/blob/9.0/addons/account_check_printing/account_payment.py#L15)

NL66278:
@pedrobaeza I deleted two spurious commits - needed before module rename PR was merged.

NL66278:
@jbeficent I got this for account_voucher.number in my limited test set. How to convert to an integer, like account_payment.number:

```
odocrmdem80_migrated=# select id, name, voucher_type, number from account_voucher;
 id |        name        | voucher_type |     number      
----+--------------------+--------------+-----------------
  7 |                    | payment      | CHKEU/2016/0001
  4 |                    | receipt      | BNK2/2016/0002
  1 |                    | receipt      | DKAS/2016/0001
  2 | Test sales voucher | sale         | SAJ/2016/007
  5 |                    | receipt      | BNK1/2016/0001
  3 |                    | receipt      | BNK2/2016/0001
  6 |                    | purchase     | 
(7 rows)
```

JordiBForgeFlow:
@NL66278 you're right..

addons/account_check_printing/__openerp__.py
NL66278: @pedrobaeza I think nothing is realy forced, because account_voucher was a requirement for account_check_writing in 8.0. So if account_check_printing, the renamed account_check_writing is installed, it is certain account_voucher is also already installed. This is just to ensure that account_voucher is loaded (and migrated) before account_check_printing, because the migration of account_check_printing updates account_payment, that is filled by the migration of account_voucher.

pedrobaeza: OK, thanks for the clarification. This text pasted as comment here would help a lot.

NL66278: @pedrobaeza I added the explanation to the module description.

pedrobaeza (approved):

pedrobaeza:
Oh, wait, you have missed to update doc to indicate that there's actually something done instead of "Nothing to do" text.

StefanRijnhart:
Thank you. Let's keep it optimistic. @pedrobaeza https://github.com/OCA/OpenUpgrade/pull/650

PR #651, opened 2016-10-25
[IMP] don't autoinstall nonexisting modules

hbrunn:
without this, OpenUpgrade will try to install email_template for example, and whatever other junk records exist from previous versions


pedrobaeza (changes requested):
It seems reasonable, as this is only in case of autoinstall modules. Please mark this with comments at the beginning and at the end to allow to identify it for later versions to be picked.

hbrunn:
@pedrobaeza this change is in the openupgrade folder, this is specific to openupgrade anyways

pedrobaeza (approved):
OK, sorry for not checking that

StefanRijnhart:
Thanks!

PR #652, opened 2016-11-02
[FIX] replace subselect by join

hbrunn:
the original fails if a move line has multiple vouchers (that was possible earlier). And getting rid of the subselects also gives us quite a performance boost


pedrobaeza (approved):

susport:
:+1

susport:
+1

susport:
👍

pedrobaeza:
@susport, use new approval system for counting on the statistics that will be available for OCA.

PR #653, opened 2016-11-02
[FIX] purchase: Fill currency_id with the related value from pricelist

pedrobaeza:
As currency_id was in v7 a non stored related field, when passing to v8,
we lose the associated value, and thus, the default one is used. With
this patch, we restore it later the value expected taking it from the
related pricelist.


pedrobaeza:
cc @Tecnativa

legalsylvain:
:+1: LGTM. Thanks.

PR #656, opened 2016-11-03
[ADD] rename account_payment to account_payment_order

hbrunn:
I'm not so sure about this one, what do you think?

### DIFF
--- a/openerp/addons/openupgrade_records/lib/apriori.py
+++ b/openerp/addons/openupgrade_records/lib/apriori.py
@@ -19,6 +19,8 @@ renamed_modules = {
     'runbot_secure': 'runbot_relative',
     # not exactly a module rename, but they do the same
     'account_check_writing': 'account_check_printing',
+    # same here, whoever want this will also need bank-payment
+    'account_payment': 'account_payment_order',
     # OCA/account-invoicing
     'account_refund_original': 'account_invoice_refund_link',
 }


StefanRijnhart:
you are mapping a standard module to an OCA module. Does not sound 100% correct to me. Maybe we should make this dependent on the presence of another OCA payment related module?

pedrobaeza:
I don't see a special problem with that while we have migration scripts in the _account_payment_order_. Both modules share most of the models, but the script is mandatory.

hbrunn:
yes, the migration script will be necessary. Setting this to work in progress meanwhile

hbrunn:
@StefanRijnhart it's a module that replaces lost functionality, so I don't think that's a big issue. Probably we'll encounter more of this the more vital stuff is moved from community to enterprise...

pedrobaeza:
Yeah, we have the same with contract module for example, or account_followup...

pedrobaeza:
FYI, we are working on this now.

pedrobaeza:
OK, this is the map:

- [x] account_payment > account_payment_order: This one
- [x] account_payment_order: https://github.com/OCA/bank-payment/pull/325
- [x] account_payment_mode: https://github.com/OCA/bank-payment/pull/320
- [x] account_banking_mandate: https://github.com/OCA/bank-payment/pull/321
- [x] account_payment_partner: https://github.com/OCA/bank-payment/pull/322
- [x] account_banking_pain_base: https://github.com/OCA/bank-payment/pull/324

hbrunn:
now that https://github.com/OCA/bank-payment/pull/320 is merged, we can merge this, right?

pedrobaeza:
There is still one PR pending, that I'm checking right now (OCA/bank-payment#325). I'll merge when finish.

pedrobaeza:
@hbrunn, we need to merge before https://github.com/OCA/openupgradelib/pull/63. Please check.

pedrobaeza:
All merged, so merging this one also.

PR #663, opened 2016-11-14
[9.0] fast reconcile

NL66278:
Migration of reconciliations is unworkable slow

Current behavior before PR: reconciliation migration for a database with a significant amount of data can take days to complete.

Desired behavior after PR is merged: Migration should take not more then a few hours and preferably considerable less.
--


hbrunn:
tests fail in https://travis-ci.org/OCA/OpenUpgrade/builds/175729408#L2814

pedrobaeza:
Is this the same as https://github.com/OCA/OpenUpgrade/pull/634?

NL66278:
@pedrobaeza It is an alternative approach to #634, relying more on pure SQL to achieve a more or less acceptable runtime for the migration.

pedrobaeza:
I think you shouldn't modify addon code, but migration script.

NL66278:
@pedrobaeza The problem with a migration script is that in 9.0 the functionality of reconciliation and full reconciliation has been split in two modules. The needed database tables and models are only complete after the load of module account_full_reconcile. But this module did not exist before 9.0, so a migration script would never run (as I found out the painfull way).

pedrobaeza:
No, it's very simple to do:

1) mark as "to install" the module full reconcile in this one
2) Put in the post-migration of the full reconcile module the scripts.

NL66278:
@pedrobaeza I think we have some miscommunication of sorts, or there is some magic that I don't know of. The module account_full_reconcile is an automatically installed module, ofcourse if its requirements are met. The only requirement is the account module, so why Odoo SA decided to make this a separate module is beyond my understanding. But account_full_reconcile not existing in 8.0 to my knowledge (and experience) will result in no migration being run, as there is no version to be migrated.

pedrobaeza:
OK, you're right about that part. I mixed the things to do. What you have to use instead of a post-migration script is a post-init-hook, like your partner @hbrunn has done in another PR.

NL66278:
@pedrobaeza  Odoo SA had done some migration stuff in the __init__.py which I simply replaced with the complete migration of reconciliation. But if you think it is better to move it to a post-init hook, I have no problem with that.

hbrunn:
the issue here is that we regularly merge upstream into openupgrade. So we need to keep changes to the upstream code to a minimum in order to let this process go smoothly. That's why indeed the post init hook is less prone to cause trouble later, because then only conflicts in the manifest could happen which would be simple to solve, while you have to think when solving a conflict involving "real" code

NL66278:
@pedrobaeza @hbrunn Actually I was confused and now even more. The _migrate_full_reconcile already is a post_init_hook. I understand this should be a method defined in __init__.py. Which it already is. So what change do you propose?

hbrunn:
right, first read code again, then comment. Nothing from my side, apologies for the noise

pedrobaeza:
The same for me. I see the changes, but not where. I think this module is really nominee to the Odoo modules hall of shame.

hbrunn:
I don't see code that recomputes the `reconciled` field, do you have to set this one too manually? The original code recomputes it, but that's forbiddingly expensive I expect: https://github.com/OCA/OpenUpgrade/pull/663/files#diff-37a531f02f2e57d4fa1bc4bfcb55404aL317 (can't look closely at the code because I'm in a somewhat suboptimal position for work at the moment)

hbrunn:
the tests fail in https://travis-ci.org/OCA/OpenUpgrade/builds/184268280#L3027 - can you check if this indicates a problem or if you have to move the test code to the full reconcile module?

NL66278:
The full_reconcile_id is set here. Only called if moves were already fully reconciled:
```
+        full_reconcile_id = (
+            record.full_reconcile and
+            create_full_reconcile(cr, record.reconcile_ref) or
+            False
+        )
```
Actually the code is slightly misleading as I read the original reconcile_id into record.full_reconcile_id, and then fill variablefull_reconcile_id with the id of the newly created row in account_full_reconcile. The original code copies the old value of reconcile_id straight to full_reconcile_id and then in the end resets the sequence linked to the id to the maximum value created.

NL66278:
As for the tests I have difficulty finding out what happens. test data is seemingly loaded somehow from the file account/views/account_unit_test.xml. Even though this file is not mentioned in ```__openerp__.py```. And what is test data doing in views// ?? Might well need to move, but still looking into these strange constructions.

hbrunn:
the full_reconcile_id is fine, but I meant the `reconciled` flag. This is used as marker for a line being reconciled in most of the code. Probably you also should recompute the other fields being recomputed in my link somehow.

Concerning the tests: Have a look at https://github.com/OCA/OpenUpgrade/blob/9.0/.travis.yml#L47 and https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/tests/test_account.py. So what happens during the tests is that we download a previously generated database of version n-1 (using https://github.com/OCA/OpenUpgrade/blob/9.0/scripts/create_test_db from (n-1)'s source branch), run a migration and then we've patched https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/modules/loading.py#L235 to run the tests defined in the migration folder. You can do the same locally to run and fix the tests for this migration.

NL66278:
reconciled should now be filled correctly

hbrunn:
unfortunately, we still get a lot of false negatives. Your code really needs to take care that

```sql
select count(l.id) from account_move_line l join account_account a on l.account_id=a.id and a.reconcile where reconcile_id is not null and not l.reconciled;
```

is `0` afterwards

hbrunn:
how does your code handle reconciliations involving more than two move lines? It seems your code tries to find pairs, but this will not always be possible. In https://github.com/OCA/OpenUpgrade/pull/663/files#diff-fa384c4c1891344a3d81ac40469855eeR97, I think you need to look at the sum of amount_residual of all your move lines. This now only sets lines to reconciled where you found a pair.

NL66278:
The code first tries to match pairs first, because they are most likely to belong together. After that the code tries to do all the other reconciliations. There must be something wrong in the logic of that part, but is is not that I only find pairs.

hbrunn:
then I guess not looking at the sum over all `amount_residual`s is the problem, but you miss a *lot* of reconciliations in the database I look at currently

addons/account/views/account_unit_test.xml
```diff
@@ -1,62 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
hbrunn: you shouldn't touch this file because it's odoo core

addons/account/migrations/9.0.1.1/tests/test_account.py
```diff
@@ -1,13 +0,0 @@
-# coding: utf-8
```
hbrunn: how about moving this test to account_full_reconcile instead of deleting it?

hbrunn (changes requested):

addons/account/migrations/9.0.1.1/tests/test_account.py
hbrunn: and the select statement from https://github.com/OCA/OpenUpgrade/pull/663#issuecomment-267907021 too I guess

addons/account/views/account_unit_test.xml
NL66278: I could leave it, but the file is not loaded anywhere, so just sowing confusion.

addons/account/migrations/9.0.1.1/tests/test_account.py
NL66278: I thought about moving the test. The problem is: account_full_reconciliation is not really migrated, because the module did not exist before. Then maybe the OpenUpgrade code should be changed to run migrations based on wether or not there is a migrations directory in a module, even for modules that did not exist in the previous version. As yet had no time to investigate this.

hbrunn: ah, another proposal: If you just do an sql update on `ir_module_module` setting state to `to upgrade` for the `full_reconcile` module, the full reconcile module will be treated as being updated. Then you can not touch the init hook at all, and this test will be run

addons/account/views/account_unit_test.xml
hbrunn: unnecessarily editing the upstream code gives us the chance of merge conflicts we have to sort out manually, first having to dig through the git history to find out why a specific change was made in the first place. So I highly doubt that's worth it

addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -165,6 +166,78 @@ def map_account_tax_template_type(cr):
...
+        "Currency"
+    )
+    create_field(
+        cr,
+        "account_move", "amount", "numeric",
+        "Amount"
+    )
+    create_field(
+        cr,
+        "account_move", "matched_percentage", "numeric",
```
hbrunn: note that everything you create here will keep the value `NULL` if you don't fill the value manually, I think this is the case with this column and some below too.
Also note that when you precreate a column, you're responsible for eventual indexes on the columns you create, if you don't create them too, the database might be slower than it should. But for the columns here, I don't think upstream sets indexes.

addons/account/migrations/9.0.1.1/tests/test_account.py
NL66278: I tried to update the state of the account_full_reconcile module to 'to upgrade' and latest_version to '8.0.1.0'  in the post_migration of account, but migrations are not run. Seems that somehow the latest_version is reset and therefore migration ignored...

pedrobaeza:
Why don't you simply call the script protecting the import in the account post-migration?

NL66278:
@pedrobaeza Maybe it is too early in the morning, but I do not understand what you mean with "the script protecting the import".

pedrobaeza:
Yeah, I mean:
```
try:
    from openerp.addons.account_full_reconcile import _migrate_full_reconcile
    _migrate_full_reconcile(...)
except ImportError:
    pass
```

NL66278:
Now I understand more or less what you mean. But what would be the advantage of this?

pedrobaeza:
As you said that you are not able to launch the pre-init hook although you mark as "to upgrade" the module, with this you assure that, but I don't know the full context of the problem.

hbrunn:
it's indeed not so simple unfortunately because at this moment, the table is already read and the data lives in the `graph` object. On https://github.com/OCA/OpenUpgrade/pull/718, there's a proposal how to deal with this. Let's see what CI tests say

JordiBForgeFlow:
@NL66278 @hbrunn I think that in this case it would make sense to add unit tests, as the complexity of this migration is considerable. What do you think?

hbrunn:
@jbeficent absolutely. @NL66278 will move the code to a migration script so that we don't need to touch core code, and then adding the tests is a second step there, this then integrates with OpenUpgrade's tests beautifully

hbrunn:
the current version of the code sets the correct reconciled flag, but amount_residual is 0 for all of my migrated lines

JordiBForgeFlow:
In my migration I see more than a double of unreconciled journal items, as
compared with v8. Need to analyze that in more detail.

addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -165,6 +166,210 @@ def map_account_tax_template_type(cr):
...
+            SELECT
+                aml.id as id,
+                rc.currency_id as company_currency_id,
+                (aml.debit - aml.credit) as balance,
+                aml.debit as debit,
+                aml.credit as credit,
+                aa.user_type as user_type_id
+            FROM account_move_line aml
+            JOIN res_company rc ON rc.id = aml.company_id
+            JOIN account_account aa ON aa.id = aml.account_id
```
hbrunn: do you really need the subquery to operate on `account_move_line` too? Why not just `...from res_company rc, account_account aa where account_move_line.company_id=rc.id and account_move_line.account_id=aa.id`?

addons/account_full_reconcile/__init__.py
```diff
@@ -1,108 +1,325 @@
...
+
+    # Define generator to loop over potentially huge amount of records in
+    # cursor
+    def result_iter(cursor, arraysize=1024):
+        'An iterator that uses fetchmany to keep memory usage down'
+        while True:
+            try:
+                results = cursor.fetchmany(arraysize)
+            except:
+                # Assume no more rows to fetch
```
hbrunn: assumption is the mother of... the [docs](http://initd.org/psycopg/docs/cursor.html#cursor.fetchmany) quite clearly state that a falsy result is returned when there are no more rows. So if there's an exception, something is going on you shouldn't ignore

addons/account/migrations/9.0.1.1/pre-migration.py
NL66278: I need a subquery as far as I understand how UPDATE FROM works that has one line for each of the lines to be updated. If that is not needed I am mistaken.

addons/account_full_reconcile/__init__.py
NL66278: Maybe not all implementations are the same. In the original ecample I followed there always was an exception when there were no more rows. As this did not function I added the test for a falsey result. And left the exception handling there just in case. I agree it might be better to at least log the fact that an exeption occured.

NL66278:
Migration is now run from migrations directory and fields that were not filled before should be filled now.

NL66278:
@hbrunn Requested changes have been done. Thinking about additional tests.

JordiBForgeFlow:
@NL66278 I tested with the latest commits, and I'm seeing that the current approach will not respect the reconciliations where the amounts reconciled were 0. These cannot be ignored. It is common to record a payment with amount 0 for a free of charge invoice. In my use case I'm seeing a discrepancy of 6500 unreconciled journal items comparing v8 and v9.

NL66278:
@jbeficent I noticed the problem with the 0 amounts. With our customer it is a quite rare occurrence (419 lines out of > 960.000). But debugging the problem right now.

JordiBForgeFlow:
Thanks!

NL66278:
@jbeficent Problem with reconciliation of zero debit and credit lines should be fixed now.

JordiBForgeFlow:
@NL66278 The invoices do not make any reference to the associated payments, and this is a big issue for users.

What should be:
image


With the actual migration this is missing:
image

See the code I added in PR https://github.com/OCA/OpenUpgrade/pull/634 for that:
https://github.com/OCA/OpenUpgrade/pull/634/files#diff-37a531f02f2e57d4fa1bc4bfcb55404aR533

https://github.com/OCA/OpenUpgrade/pull/634

NL66278:
@jbeficent I restored the link between invoices and payments.

JordiBForgeFlow:
Thanks!! I will try today.

JordiBForgeFlow:
The reconciliation of entries with 0 now works fine.
image


The payments are also displayed correctly in the invoice.
image

addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -165,8 +166,37 @@ def map_account_tax_template_type(cr):
...
+    filled with valid values by migration."""
+    from openerp.addons.account.models.account_move import \
+            AccountMove, AccountMoveLine
+    AccountMove._openupgrade_recompute_fields_blacklist = [
+        'currency_id',
+        'amount',
+        'matched_percentage',
+    ]
+    AccountMoveLine._openupgrade_recompute_fields_blacklist = [
+        'amount_residual',
```
JordiBForgeFlow, line 189: This is probably causing https://github.com/OCA/OpenUpgrade/issues/746

JordiBForgeFlow:
@NL66278 @hbrunn Please keep away other modules from this PR. It causes conflicts when using it with buildout.

JordiBForgeFlow:
@NL66278 Travis fails due to conflicting files.

addons/account/migrations/9.0.1.1/pre-migration.py
NL66278: amount_residual since some time is already filled in the post_migration of account_full_reconcile. So I do not think this can be the problem.

NL66278:
I squashed all commits and solved the problem that merges from other branches were included in this one (except for the blacklisting, which is intentional).

JordiBForgeFlow:
@NL66278 After migrating the invoices, I still see that the field 'Amount due' shown in the list view is wrong.

The field 'Amount due' in the invoice list view is 0:
image
But in the form view is != 0:
image

NL66278:
@jbeficent The treeview is showing the residual_signed field, the form the residual field. So apperently the second is correctly filled, the first not. Looking into it.

NL66278:
@jbeficent This was a regression due to changes in the way blacklisting field recomputation was implemented. Before the blacklist was reset too often, now not enough, meaning that the forced recomputation had no effect. Solved this by manually doing the reset.

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -370,6 +341,144 @@ def migrate_account_auto_fy_sequence(env):
...
+                    WHEN ai.type IN ('in_refund', 'out_refund')
+                    THEN ROUND(-ail.price_subtotal * er.effective_rate, 2)
+                    ELSE ROUND(ail.price_subtotal  * er.effective_rate, 2)
+                END
+            FROM account_invoice_line ail
+            JOIN account_invoice ai ON ail.invoice_id = ai.id
+            JOIN effective_rates er ON ai.id = er.invoice_id
+        )
+        UPDATE account_invoice_line
+        SET price_subtotal_signed = subquery.price_subtotal_signed
```
JordiBForgeFlow, line 463: @NL66278 This query did not work on my latest migration. Was resulting in 0.

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -370,6 +341,144 @@ def migrate_account_auto_fy_sequence(env):
...
+            GROUP BY ai.id, com_cur.currency_id
+        )
+        , invoice_rate_selection AS (
+            SELECT
+                ai.id, inv_cur.currency_id as invoice_currency_id,
+                max(inv_cur.name) as rate_date
+            FROM account_invoice ai
+            JOIN res_company com on ai.company_id = com.id
+            LEFT OUTER JOIN res_currency_rate inv_cur
+                ON COALESCE(ai.currency_id, com.currency_id) = inv_cur.id
```
JordiBForgeFlow: `inv_cur.id` is wrong. Replace with `inv_cur.currency_id`

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -370,6 +341,144 @@ def migrate_account_auto_fy_sequence(env):
...
+        cr,
+        """\
+        WITH company_rate_selection AS (
+            SELECT
+                ai.id, com_cur.currency_id as company_currency_id,
+                max(com_cur.name) as rate_date
+            FROM account_invoice ai
+            JOIN res_company com on ai.company_id = com.id
+            LEFT OUTER JOIN res_currency_rate com_cur
+                ON com.currency_id = com_cur.id
```
JordiBForgeFlow: `com_cur.id` is wrong. Replace with `com_cur.currency_id`

addons/account/migrations/9.0.1.1/post-migration.py
JordiBForgeFlow: This is the query that worked for me:

`WITH company_rate_selection AS (
            SELECT
                ai.id, com_cur.currency_id as company_currency_id,
                max(com_cur.name) as rate_date
            FROM account_invoice ai
            JOIN res_company com on ai.company_id = com.id
            LEFT OUTER JOIN res_currency_rate com_cur
                ON com.currency_id = com_cur.currency_id
                    AND ai.company_id =
                        COALESCE(com_cur.company_id, ai.company_id)
                    AND COALESCE(ai.date_invoice, date(ai.create_date)) >=
                        date(com_cur.name)
            GROUP BY ai.id, com_cur.currency_id
        )
        , invoice_rate_selection AS (
            SELECT
                ai.id, inv_cur.currency_id as invoice_currency_id,
                max(inv_cur.name) as rate_date
            FROM account_invoice ai
            JOIN res_company com on ai.company_id = com.id
            LEFT OUTER JOIN res_currency_rate inv_cur
                ON COALESCE(ai.currency_id, com.currency_id) = inv_cur.currency_id
                    AND ai.company_id =
                        COALESCE(inv_cur.company_id, ai.company_id)
                    AND COALESCE(ai.date_invoice, date(ai.create_date)) >=
                        date(inv_cur.name)
            GROUP BY ai.id, inv_cur.currency_id
        )
        , effective_rates AS (
            SELECT
                ai.id as invoice_id,
                com_cur.rate as company_currency_rate,
                inv_cur.rate as invoice_currency_rate,
                (inv_cur.rate / com_cur.rate) as effective_rate
            FROM account_invoice ai
            JOIN company_rate_selection crs ON ai.id = crs.id
            JOIN res_currency_rate com_cur
                ON crs.company_currency_id = com_cur.currency_id
                    AND crs.rate_date = com_cur.name
            JOIN invoice_rate_selection irs ON ai.id = irs.id
            JOIN res_currency_rate inv_cur
                ON irs.invoice_currency_id = inv_cur.currency_id
                    AND irs.rate_date = inv_cur.name)
        , subquery(id, price_subtotal, price_subtotal_signed) AS (
            SELECT
                ail.id,
                ail.price_subtotal,
                CASE
                    WHEN ai.type IN ('in_refund', 'out_refund')
                    THEN ROUND(-ail.price_subtotal * er.effective_rate, 2)
                    ELSE ROUND(ail.price_subtotal  * er.effective_rate, 2)
                END
            FROM account_invoice_line ail
            JOIN account_invoice ai ON ail.invoice_id = ai.id
            JOIN effective_rates er ON ai.id = er.invoice_id
        )
        UPDATE account_invoice_line
        SET price_subtotal_signed = subquery.price_subtotal_signed
        FROM subquery
        WHERE account_invoice_line.id = subquery.id`

pedrobaeza:
There is a strange error on Travis, but I think it's due to the PR mechanism. I'm merging as @hbrunn's are honored.

addons/account/migrations/9.0.1.1/post-migration.py
NL66278: @jordi, it is very difficult for me to see the difference with what I proposed (and is now merged), due to the different formatting. Could you make a new PR containing your improvements?

PR #668, opened 2016-11-21
[IMP] fix pylint's complaints

hbrunn:

PR #669, opened 2016-11-21
[ADD] multiply taxes by 100

hbrunn:


legalsylvain:
:+1: 
Thanks.

hbrunn:
will be green after #668 is merged

jjscarafia:
:+1:

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -417,6 +417,10 @@ def migrate(env, version):
     WHERE child_depend IS True
     """)
 
+    # In v9, percentages are expressed as percentage
```
pedrobaeza: Better to say: `In v9, percentages are expressed as hundred-based percentage, not one-based percentage`

susport (commented):
👍

PR #672, opened 2016-11-24
[MIG] procurement module migration scripts

cubells:

 * completed analysys
 * added analysis_work file


addons/procurement/migrations/9.0.1.0/openupgrade_analysis.txt
```diff
@@ -1,10 +1,16 @@
 ---Fields in module 'procurement'---
-procurement  / procurement.order        / message_follower_ids (many2many): not a function anymore
-procurement  / procurement.order        / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
-procurement  / procurement.order        / message_follower_ids (many2many): type is now 'one2many' ('many2many')
-procurement  / procurement.order        / product_uos (many2one)        : DEL relation: product.uom
-procurement  / procurement.order        / product_uos_qty (float)       : DEL 
-procurement  / procurement.order        / website_message_ids (one2many): NEW relation: mail.message
+procurement  / procurement.order        / message_follower_ids (many2many)      : not a function anymore
```
pedrobaeza: Why have you changed this file?

pedrobaeza:
You have to update documentation in https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/doc/source/modules80-90.rst

addons/procurement/migrations/9.0.1.0/openupgrade_analysis.txt
cubells: To add some analysis of fields/views that are missing in initial analysis.

pedrobaeza: Two things about this:

* Don't change lines although the indentation is not so good for not increasing the diff.
* Why they are not included? Maybe we should improve the analysis method. Anyway, don't include here that extra lines, but it _work file.

addons/procurement/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,50 @@
...
+procurement  / procurement.order        / message_partner_ids (many2many)       : NEW relation: res.partner
+# New field ---> Nothing to do
+
+procurement  / procurement.order        / message_summary (text)                : DEL
+# Nothing to do
+
+procurement  / procurement.order        / message_unread_counter (integer)      : NEW
+# New field ---> Nothing to do
+
+procurement  / procurement.order        / product_uos (many2one)                : DEL relation: product.uom
```
pedrobaeza, line 25: If this dissapears, maybe now `product_uos` and `product_uos_qty` of the sales orders are put directly in `product_uom` and `product_uom_qty`. If so, then you should include a script for copying "uos" values (in pre-migration) to "uom" when "uos" is present. Please check the behavior in v8 and v9 when you put sales units and tell me back.

cubells: Not.

product_uos and product_uos_qty dissapear:
https://github.com/OCA/OpenUpgrade/blob/9.0/addons/sale/sale.py#L557
and product_uom is setted with product_uom and product_uom_qty.


I'm right?

pedrobaeza: OK, correct, then nothing to do on this.

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -348,7 +348,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |pos_restaurant                     |                                   |
 +-----------------------------------+-----------------------------------+
-|procurement                        |                                   |
+|procurement                        | Done                              |
```
pedrobaeza: s/Done/Nothing to do

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #674, opened 2016-11-24
[MIG] survey module migration script

cubells:


cubells:
@Tecnativa

pedrobaeza (changes requested):
I have found that there's a change in noupdate data:

```
<openerp>
  <data>
    <record id="base.group_survey_manager" model="res.groups"><field name="name">Manager</field>
            <field name="category_id" ref="base.module_category_survey"/>
            </record>
    <record id="base.group_survey_user" model="res.groups"><field name="name">User</field>
            <field name="category_id" ref="base.module_category_survey"/>
        </record>
  </data>
</openerp>
```

Take a look on this to see how to load it: https://github.com/OCA/OpenUpgrade/blob/71c6abdd40561c22fdaf03fb3cfbca5508231f9a/addons/hr_contract/migrations/9.0.1.0/post-migration.py#L9

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -430,7 +430,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |subscription                       | Nothing to do                     |
 +-----------------------------------+-----------------------------------+
-|survey                             |                                   |
+|survey                             | Nothing to do                     |
```
pedrobaeza: s/Nothing to do/Done

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #675, opened 2016-11-24
[MIG] payment module migration script

cubells:
 *  Created pre-migration.py

@Tecnativa


addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+payment      / payment.acquirer         / pending_msg (html)            : NEW
+# Nothing to do
+
+payment      / payment.acquirer         / registration_view_template_id (many2one): NEW relation: ir.ui.view
+# Nothing to do
+
+payment      / payment.acquirer         / sequence (integer)            : NEW
+# Nothing to do
+
+payment      / payment.acquirer         / validation (selection)        : DEL selection_keys: ['automatic', 'manual']
```
pedrobaeza: If the column is renamed, where is the line with the `NEW`entry? You should join both lines here in the work file.

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+# Nothing to do
+
+payment      / payment.method           / active (boolean)              : NEW
+# Nothing to do
+
+payment      / payment.method           / name (char)                   : NEW
+# Nothing to do
+
+payment      / payment.method           / partner_id (many2one)         : NEW relation: res.partner, required: required
+# Nothing to do
```
pedrobaeza: As this is a required field, you need to fill this with something.

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+
+payment      / payment.method           / payment_ids (one2many)        : NEW relation: payment.transaction
+# Nothing to do
```
pedrobaeza: What is the inverse of this one2many? Was it already present in v8? If not, we should try to fill it heuristically. If yes, please add here apart from `Nothing to do` an explanation like `Inverse field <x> was already present on v8`

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+
+payment      / payment.transaction      / callback_eval (char)          : NEW
+# Nothing to do
+
+payment      / payment.transaction      / date_create (datetime)        : DEL required: required, req_default: function
+# Column renamed -> done
```
pedrobaeza: Renamed to what?

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+payment      / payment.transaction      / message_ids (one2many)        : DEL relation: mail.message
+payment      / payment.transaction      / message_last_post (datetime)  : DEL
+# Nothing to do
+
+payment      / payment.transaction      / partner_lang (char)           : selection_keys is now 'function' ('False')
+payment      / payment.transaction      / partner_lang (char)           : type is now 'selection' ('char')
+# Nothing to do (function returns the user lang)
+
+payment      / payment.transaction      / partner_reference (char)      : DEL
+# Column renamed -> done
```
pedrobaeza: Renamed to what?

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+
+NEW ir.model.access: payment.payment_method_all
+NEW ir.model.access: payment.payment_method_portal
+NEW ir.model.access: payment.payment_method_system
+NEW ir.model.access: payment.payment_method_user
+# ir.model.access removed ---> Nothing to do
+
+NEW ir.rule: payment.payment_method_salesman_rule
+NEW ir.rule: payment.payment_method_user_rule
+# Rule removed in v9 ---> Nothing to do
```
pedrobaeza: s/removed/added

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+NEW ir.actions.act_window: payment.action_payment_tx_ids
+NEW ir.actions.act_window: payment.payment_method_action
+NEW ir.actions.act_window: payment.payment_transaction_action_child
+# ir.actions.* ---> Nothing to do
+
+NEW ir.model.access: payment.payment_method_all
+NEW ir.model.access: payment.payment_method_portal
+NEW ir.model.access: payment.payment_method_system
+NEW ir.model.access: payment.payment_method_user
+# ir.model.access removed ---> Nothing to do
```
pedrobaeza: s/removed/added

addons/payment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+# Nothing to do
+
+payment      / payment.acquirer         / registration_view_template_id (many2one): NEW relation: ir.ui.view
+# Nothing to do
+
+payment      / payment.acquirer         / sequence (integer)            : NEW
+# Nothing to do
+
+payment      / payment.acquirer         / validation (selection)        : DEL selection_keys: ['automatic', 'manual']
+# Column renamed -> done
```
pedrobaeza: If the column is renamed, where is the line with the `NEW`entry? You should join both lines here in the work file.

pedrobaeza: OK, I get it. You are renaming to preserve the columns in case of being needed. Then please put the text `Column renamed to None to be preserved` or something similar.

pedrobaeza (changes requested):
Some things to fix

pedrobaeza (approved):

pedrobaeza (changes requested):
Sorry, it's not correct. Please see Travis: https://travis-ci.org/OCA/OpenUpgrade/builds/178852352#L3131

pedrobaeza (approved):

PR #676, opened 2016-11-24
[MIG] website_sale module migration script

cubells:

@Tecnativa


addons/website_sale/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,84 @@
...
+# Nothing to do
+
+website_sale / product.template         / rating_ids (one2many)         : NEW relation: rating.rating
+# Nothing to do
+
+website_sale / res.country.group        / website_pricelist_ids (many2many): NEW relation: website_pricelist
+# Nothing to do
+
+website_sale / res.partner              / last_website_so_id (many2one) : NEW relation: sale.order
+# Nothing to do
```
pedrobaeza: We can try to deduce this looking for a sales order that comes from website from this partner.

cubells: A possible solution.

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -524,7 +524,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |website_rating_project_issue       |                                   |
 +-----------------------------------+-----------------------------------+
-|website_sale                       |                                   |
+|website_sale                       | Nothing to do                     |
```
pedrobaeza: This should be `Done`

addons/website_sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2016 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+from openerp import api, SUPERUSER_ID
+
+
+def set_last_sale_on_partner(env):
+    partners = env['res.partner'].search([])
```
pedrobaeza: This is going to be awfully slow. You can use an SQL to make this:
```
UPDATE res_partner rp
SET last_website_so_id = so.id
FROM sale_order so
WHERE so.id = (
   SELECT max(id) FROM sale_order
   WHERE partner_id = rp.id
)
AND rp.last_website_so_id IS NULL;
```
Note that `website_order_line` is a one2many field that behaves the same of normal order_line (so all sales orders are website applicable).

pedrobaeza (changes requested):

addons/website_sale/migrations/9.0.1.0/post-migration.py
cubells: But how can I know which sales are normal sale and which sales are web sales?

cubells: Ok. I read your message now.

pedrobaeza (approved):

PR #677, opened 2016-11-24
9.0 [MIG] mass_mailing migration scripts

cubells:
 * Added pre and post migration fields

@Tecnativa


addons/mass_mailing/migrations/9.0.2.0/openupgrade_analysis.txt
```diff
@@ -1,30 +1,76 @@
 ---Fields in module 'mass_mailing'---
 mass_mailing / link.tracker             / mass_mailing_campaign_id (many2one): NEW relation: mail.mass_mailing.campaign
+# Nothing to do
```
pedrobaeza: You should do this on a copy.

addons/mass_mailing/migrations/9.0.2.0/openupgrade_analysis.txt
```diff
@@ -1,30 +1,76 @@
...
+# Done ---> 'test' is converted to 'draft'
+
+mass_mailing / mail.mass_mailing.campaign / _inherits (False)             : NEW
+# Nothing to do
+
 mass_mailing / mail.mass_mailing.campaign / campaign_id (many2one)        : NEW relation: utm.campaign, required: required, req_default: function
+# Nothing to do
+
 mass_mailing / mail.mass_mailing.campaign / category_ids (many2many)      : DEL relation: mail.mass_mailing.category
+# Nothing to do
```
pedrobaeza: You should preserve this column and rename the table for any possible inherited module that uses it.

pedrobaeza (changes requested):

addons/mass_mailing/migrations/9.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
...
+
+column_copies = {
+    'mail_mass_mailing': [
+        ('state', None, None),
+    ],
+}
+
+column_renames = {
+    'mail_mass_mailing_campaign': [
+        ('category_ids', None),
```
pedrobaeza: As this is a many2many, you can't rename a column, but rename the table to avoid the removal.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #679, opened 2016-11-26
[ADD] rename product category properties

hbrunn:


pedrobaeza (approved):

legalsylvain:
Hi @hbrunn, 

I don't figure out why you need to add this script ?

The definition of the field manages this renaming. [Ref](https://github.com/hbrunn/OpenUpgrade/blob/5af5cf455297ffca1bf07dc1a72c6c227214ab01/addons/stock_account/product.py#L256)

I guess that's why the analysis file marks the two fields with 'nothing to do'. [Ref](
https://github.com/hbrunn/OpenUpgrade/blob/5af5cf455297ffca1bf07dc1a72c6c227214ab01/addons/stock_account/migrations/9.0.1.1/openupgrade_analysis.txt#L5)

hbrunn:
those are properties, and there's no code (at least in odoo 9 community) that picks up those hints for properties and does something useful with it. This is then a false negative in the analysis script which also needs fixing.

legalsylvain:
Thanks for this explanation ! I didn't know. 
So :+1:

legalsylvain (approved):

PR #680, opened 2016-11-27
[FIX] don't claim there's nothing to do for properties with an oldname

hbrunn:
for details, see https://github.com/OCA/OpenUpgrade/pull/679#issuecomment-263115071


pedrobaeza (approved):
Maybe we should also send a patch to Odoo to fix this oldname behavior in properties...

legalsylvain (approved):

PR #681, opened 2016-11-27
[FIX] don't claim there's nothing to do for properties with an oldname

hbrunn:
cf https://github.com/OCA/OpenUpgrade/pull/680


pedrobaeza (approved):

legalsylvain:
:+1:

PR #682, opened 2016-11-28
[MIG] hr_recruitment module migration script

cubells:
* Added analysis files
 * Created pre and post files


legalsylvain:
Hi  @cubells Thanks for your work. Could you add in your PR the change of the module coverage file. (set to 'Done' hr_recrutment line.)
https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/doc/source/modules80-90.rst

kind regards.

cubells:
Done @legalsylvain

addons/hr_recruitment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,113 @@
+---Fields in module 'hr_recruitment'---
+hr_recruitment / hr.applicant             / attachment_ids (one2many)     : NEW relation: ir.attachment
+# Nothing to do
+
+hr_recruitment / hr.applicant             / availability (integer)        : type is now 'date' ('integer')
+# Unable to calculate date form a integer --> Nothing to do
```
pedrobaeza: You should take field `create_date` and add this number of days to get the new date.

addons/hr_recruitment/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,113 @@
...
+hr_recruitment / hr.applicant             / message_follower_ids (many2many): not a function anymore
+hr_recruitment / hr.applicant             / message_follower_ids (many2many): relation is now 'mail.followers' ('res.partner')
+hr_recruitment / hr.applicant             / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+hr_recruitment / hr.applicant             / priority (selection)          : selection_keys is now '['0', '1', '2', '3']' ('['0', '1', '2', '3', '4']')
+# Done: Adapt to new range
+
+hr_recruitment / hr.applicant             / source_id (many2one)          : relation is now 'utm.source' ('hr.recruitment.source')
+# Nothing to do
```
pedrobaeza: You need to create an utm.source for each of the existing hr.recruitment.source, as now the object inherits (via delegation) of this object, and fill the corresponding field:
````
cr.execute('INSERT INTO utm_source ... RETURNING id')
new_id = cr.fetchone()
cr.execute('UPDATE hr_applicant SET source_id = %s WHERE id = %s', (new_id, old_id))
```

pedrobaeza (changes requested):

addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
```diff
@@ -2,6 +2,27 @@
...
+    cr.execute(
+        """
+            SELECT id, create_date, %s
+            FROM hr_applicant
+        """ % (openupgrade.get_legacy_name('availability'),)
+    )
+    for applicant in cr.fetchall():
+        new_date = fields.Date.from_string(applicant[1])
+        if applicant[2]:
+            new_date += timedelta(days=applicant[2])
```
pedrobaeza: You must convert this again: `fields.Date.to_string(new_date)`

addons/hr_recruitment/migrations/9.0.1.0/pre-migration.py
```diff
@@ -5,9 +5,31 @@
...
 
 column_renames = {
-    'hr_applicant': [('priority', None)]}
+    'hr_applicant': [
+        ('priority', None),
+        ('availability', None),
+    ]}
+
+
+def migrate_applicant_source(cr):
```
pedrobaeza: You should do this on post-migration, as utm.source doesn't have yet the new structure.

addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
```diff
@@ -0,0 +1,58 @@
...
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.load_data(cr, 'hr_recruitment',
+                          'migrations/9.0.1.0/noupdate_changes.xml')
+    openupgrade.map_values(
+        cr,
+        openupgrade.get_legacy_name('priority'),
+        'priority',
+        [('4', '3'), ('3', '1'), ('2', '0'), ('1', '0')],
```
pedrobaeza: You must change the map_values order, as this is got iterating over values, so in your current map, you will replace `4` value by `3`, and next iteration, the existing `3` and the new `3` will be replaced by `1`. You must reverse the mapping.

And another thing, why there's no `2` in the new mapping?

cubells: I do that mapping:
4 is excellent and is mapped to excellent.
3 is good and is mapped to good.
All the others are mapped to normal.
Very good is not mapped.

pedrobaeza: But right now three of the old values map to the new 0 (0, 1 and 2). Is that correct?

cubells: yes, It is.

Average, Below Average and Bad are mapped to Normal because there is not another similar value.

addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
```diff
@@ -22,7 +22,7 @@ def update_applicant_availability(cr):
             cr,
             """
             UPDATE hr_applicant SET availability = %s
-            WHERE id = %s
+            WHERE id = '%s'
```
pedrobaeza: You need to do this because you're not using psycopg capability to convert types. You have to do this:
```
WHERE id = %s
""", (new_date, applicant[0]))
```
This will also prevent SQL injection possibilities: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#no-sql-injection

cubells: It's true.
thanks @pedrobaeza

addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
```diff
@@ -0,0 +1,58 @@
...
+def migrate_applicant_source(cr):
+    cr.execute(
+        """
+            SELECT id
+            FROM hr_recruitment_source
+        """)
+    for old_id in cr.fetchall():
+        cr.execute("INSERT INTO utm_source (name) "
+                   "SELECT name "
+                   "FROM hr_recruitment_source "
```
pedrobaeza: You have to rename table to avoid the removal of the significant fields in the update process

addons/hr_recruitment/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
```
susport, line 2: Remove labels openerp data by odoo in v9.0

pedrobaeza: This is an automatic file generated by https://github.com/OCA/OpenUpgrade/blob/9.0/scripts/compare_noupdate_xml_records.py, so we can leave it as is for this version and change the script for next one.

addons/hr_recruitment/migrations/9.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,21 @@
...
+
+table_renames = [
+    ('hr_recruitment_source', None),
+]
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.rename_columns(cr, column_renames)
+    openupgrade.rename_tables(cr, table_renames)
```
pedrobaeza: Line at the end

addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
pedrobaeza: Here you have to access to the migrated table name, not the new one.

pedrobaeza (approved):

PR #692, opened 2016-12-06
[FIX] graph: Fix AttributeError

mvaled:
Fixes the following error:

```
  ...
  File "openupgrade/openerp/modules/graph.py", line 93, in add_modules
    forced_deps = tools.safe_eval.safe_eval(forced_deps)
AttributeError: 'function' object has no attribute 'safe_eval'
```

### DIFF
--- a/openerp/modules/graph.py
+++ b/openerp/modules/graph.py
@@ -90,7 +90,7 @@ class Graph(dict):
         forced_deps = tools.config.get_misc('openupgrade',
                                             'force_deps_' + release.version,
                                             forced_deps)
-        forced_deps = tools.safe_eval.safe_eval(forced_deps)
+        forced_deps = eval(forced_deps)
 
         for module in module_list:
             # This will raise an exception if no/unreadable descriptor file.


mvaled:
I have sent the OCA CLA by email. Do I need to reopen this PR?

pedrobaeza:
No, you don't need to do anything on the PR for the CLA, but I'm not sure the issue is real, as the import and the call are correct.

mvaled:
> No, you don't need to do anything on the PR for the CLA, but I'm not
> sure the issue is real, as the import and the call are correct.

Hi @pedrobaeza,

The issue is real in the sense that it happened to me.  The issue, I
think, comes from the following:

- `tools/__init__.py` does not import the module `safe_eval`.

- `tools/__init__.py` imports * from module `convert`.

- module `tools/convert.py` defines:

  ```
  from safe_eval import safe_eval as s_eval
  safe_eval = lambda expr, ctx={}: s_eval(expr, ctx, nocopy=True)
  ```

So the `safe_eval` we get from openupgrade's `tools` module is the
function defined in `convert.py`.  That's why I get the Traceback
complaining about a function not having the `safe_eval` attribute.

Best regards,
Manuel.

mvaled:
Hi @pedrobaeza,

Double checking: It seems the change in convert comes from our own changes.  So I'm closing this.

In fact the change came from Odoo's commit 12634e1 and the merge in 447179c.  Not sure why this is not merged.  So this issue came from me, merging Odoo's 9.0 branch prematurely.  This change will be needed when merging the Odoo's 9.0 branch into 9.0.

mvaled:
I see that the previous changes around the lines I'm changing were done by @StefanRijnhart in 3426b638e.  @StefanRijnhart would you think this change will make any sense after we merge again from Odoo's 9.0?

StefanRijnhart:
We might not merge again from 9.0 because we don't expect the data model to have changed. But if we do, it should become relevant. To prevent surprises at that time, could you cherry pick the change that lead to the incompatibility and change the call to eval accordingly? Could you make sure that 'eval' is not used as an alias anymore, because it implies unsafe eval? Rather something like s_eval or safe_eval itself.

mvaled:
Well, in this module there's single mention of 'eval'.

```
$ grep eval openerp/modules/graph.py
from openerp.tools.safe_eval import safe_eval as eval
        forced_deps = eval(forced_deps)
```
I didn't find any import of `eval` from the `graph.py` module.

I had to make this merge to include several changes we've done to the core.  I'm just beginning to asses the feasibility to upgrade a large installation, so it's early for me to see this is 100% bug-free.  But I don't see any downsides so far.

hbrunn:
this change uses the unsaef eval, and https://github.com/OCA/OpenUpgrade/pull/692#issuecomment-265250321 looks like there's some patches upstream used. Quite out of scope here.

PR #694, opened 2016-12-09
[9.0][ADD] point_of_sale: Add script for migrate from v8

sergio-teruel:
Migration script for Point_of_sale

--

### DIFF
--- a/addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
+++ b/addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openupgradelib import openupgrade
+
+column_defaults = {
+    'pos.config': [
+        ('barcode_nomenclature_id', None)
+    ],
+    'pos.order': [
+        ('session_id', None)
+    ],
+}
+
+
+def update_barcodes_nomenclatures(env):
+    # Create new barcode_nomenclature for each pos_config and create rules
+    # for old fields
+    cr = env.cr
+    pos_configs = env['pos.config'].search([])
+    for pos_config in pos_configs:
+        barcode_nomenclature = env['barcode.nomenclature'].create({
+            'name': pos_config.name,
+            'upc_ean_conv': 'always',
+            'rule_ids': 1,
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+                    field_name + 'Barcode', barcode_nomenclature.id, 50,
+                    field_name, 'any', value
+                )
+            )
+        pos_config.barcode_nomenclature_id = barcode_nomenclature.id
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    registry = RegistryManager.get(env.cr.dbname)
```
hbrunn: an `env` already knows its registry: https://github.com/OCA/OCB/blob/9.0/openerp/api.py#L753

addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+    ],
+}
+
+
+def update_barcodes_nomenclatures(env):
+    # Create new barcode_nomenclature for each pos_config and create rules
+    # for old fields
+    cr = env.cr
+    pos_configs = env['pos.config'].search([])
+    for pos_config in pos_configs:
```
hbrunn, line 22: all of this could be a lot more efficient with a `select...into` statement, but given we don't expect a lot of pos configs, we don't really care I guess

hbrunn (changes requested):
a few nags, but fine overall

addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
pedrobaeza: This is exactly the case I have commented about old `pool` variables we need in https://github.com/OCA/openupgradelib/issues/65

addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
pedrobaeza: As the operations are a bit complicated, we prefer to keep it as is, as the result SQL would be very difficult to read and maintain.

pedrobaeza:
Comments fixed or replied, @hbrunn

pedrobaeza:
@hbrunn, any objection on merging this?

hbrunn (approved):

hbrunn:
no, go ahead

PR #697, opened 2016-12-10
[IMP] delivery: Alternative migration

pedrobaeza:
Renaming delivery_grid to delivery_carrier and taking this table as the main object, makes the operation cleaner, as the number of fields to copy, the numbers of records to insert and so on are less than with old approach.

This also allows to keep old delivery_carrier table for other modules that can take advantage of it, like *delivery_multi_destination*

### DIFF
--- a/addons/delivery/migrations/9.0.1.0/post-migration.py
+++ b/addons/delivery/migrations/9.0.1.0/post-migration.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 # © 2016 Therp BV <http://therp.nl>
+# Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 from openupgradelib import openupgrade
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@hbrunn, I ping you as you were the original author of the script. You might be interested also in https://github.com/OCA/delivery-carrier/pull/113

hbrunn:
I can't find where you deal with the zip codes, what am I missing?

pedrobaeza:
That's the good thing: you don't have to deal with them. They are already in the delivery_grid table, that has been renamed to delivery_carrier, so nothing to do.

hbrunn (approved):

hbrunn:
for the migration I work on, this works. What are you going to add?

pedrobaeza:
I haven't tried yet on a real environment, so if you confirm it works, let's merge.

hbrunn:
Hey Pedro, I think this is not cool. I only post stuff here that has worked already in some migration. Do I understand correctly you only had test data?

pedrobaeza:
Yeah, sorry, I think you were willing me to merge.  My tests are in a demo DB indeed, and as OpenUpgrade has optimistic merging if everything green, I merged. I waited for your approval as this is a change on an existing method, and I thought I had it...

hbrunn:
understood. no harm done. I still think that https://github.com/OCA/server-tools/pull/653 is undervalued

pedrobaeza:
I don't get you: why do you refer here OCA/server-tools#653? I love that module anyway :wink:

hbrunn:
it's totally unrelated, I only was fishing for compliments

hbrunn:
...and according to all the people who old-style thumbed-up it, we could merge it btw...

pedrobaeza:
hehe, you have lots of them from my team and me, although not still explicit!

pedrobaeza:
Yeah, please wait for the LGPL switch PR. Some website customers requires to be LGPL on the website layer, and it's a pity to not be able to use this cool module.

PR #702, opened 2016-12-13
[ADD] email_template: Merged in mail

pedrobaeza:
With this, we avoid one dangled module when finishing the migration, and this helps also for relocating xml-ids, translations, and so on.


pedrobaeza:
This one is OK but it depends on openupgradelib now.

pedrobaeza:
Dependency is now merged, and everything green, so I merge.

PR #704, opened 2016-12-16
[8.0] File Attachments do not show in interface

ghost:
An attachment has been saved to a shipment form but is not available after the upgrade to odoo V8

Current behavior before PR:
attachment is not available in v8 in a shipment form

Attachment is available

--

To find the correct attachment files the id and table name are searched. This causes missing attachment becaut the tables stock.picking.in and stock.picking.out are merged into stock.picking

### DIFF
--- a/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
@@ -47,3 +47,17 @@ def migrate(cr, version):
         cr, 'ir_ui_view', 'type', [
             'calendar', 'diagram', 'form', 'gantt', 'graph', 'kanban',
             'qweb', 'search', 'tree'])
+    
+    #The tables stock.picking.in and stock.picking.out are merged into stock.picking
+    openupgrade.logged_query(
+        cr, """
+        UPDATE ir_attachment
+        SET res_model = 'stock.picking'
+        WHERE res_model = 'stock.picking.out';
+        """)
+    openupgrade.logged_query(
+        cr, """
+        UPDATE ir_attachment
+        SET res_model = 'stock.picking'
+        WHERE res_model = 'stock.picking.in';
+        """)
--- a/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
@@ -48,7 +48,8 @@ def migrate(cr, version):
             'calendar', 'diagram', 'form', 'gantt', 'graph', 'kanban',
             'qweb', 'search', 'tree'])
     
-    #The tables stock.picking.in and stock.picking.out are merged into stock.picking
+    # The tables stock.picking.in and stock.picking.out are merged into 
+    # stock.picking
     openupgrade.logged_query(
         cr, """
         UPDATE ir_attachment
--- a/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
@@ -53,11 +53,6 @@ def migrate(cr, version):
         cr, """
         UPDATE ir_attachment
         SET res_model = 'stock.picking'
-        WHERE res_model = 'stock.picking.out';
-        """)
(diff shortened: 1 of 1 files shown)


hbrunn:
[edit] stupid question.

looks good!

openerp/addons/base/migrations/8.0.1.3/pre-migration.py
```diff
@@ -47,3 +47,17 @@ def migrate(cr, version):
         cr, 'ir_ui_view', 'type', [
             'calendar', 'diagram', 'form', 'gantt', 'graph', 'kanban',
             'qweb', 'search', 'tree'])
+    
+    #The tables stock.picking.in and stock.picking.out are merged into stock.picking
+    openupgrade.logged_query(
+        cr, """
+        UPDATE ir_attachment
+        SET res_model = 'stock.picking'
+        WHERE res_model = 'stock.picking.out';
```
hbrunn: make this one query

openerp/addons/base/migrations/8.0.1.3/pre-migration.py
```diff
@@ -47,3 +47,12 @@ def migrate(cr, version):
         cr, 'ir_ui_view', 'type', [
             'calendar', 'diagram', 'form', 'gantt', 'graph', 'kanban',
             'qweb', 'search', 'tree'])
+    
+    #The tables stock.picking.in and stock.picking.out are merged into stock.picking
```
pedrobaeza: This comment should be have a space between the `#` and the first word and also check 80 cols limit.

openerp/addons/base/migrations/8.0.1.3/pre-migration.py
```diff
@@ -47,3 +47,12 @@ def migrate(cr, version):
...
+    openupgrade.logged_query(
+        cr, """
+        UPDATE ir_attachment
+        SET res_model = 'stock.picking'
+        WHERE res_model = 'stock.picking.in' OR
```
pedrobaeza: Better to use:
```
WHERE res_model in ('stock.picking.in', 'stock.picking.out')
```

pedrobaeza:
@hbrunn @StefanRijnhart, don't you feel the need for an `update_model_names` method in openupgradelib that handles these things?:

* Attachments
* Reference fields entries
* Other references through base objects

pedrobaeza (approved):
Squash on merge

hbrunn:
this need was felt a long time ago: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L362

it's just that this specific migration script doesn't seem to use it

pedrobaeza:
Wow, but that something that we have to use in all places!! We don't know when we are going to have a reference field pointing to our model, so this one should be the default renaming method.

pedrobaeza:
Anyway, I'm seeing that the reference fields are not covered by this method, and also ir.ui.view and other base models that uses model name as a text column.

hbrunn:
reading the code, we don't rename anything in the stock migration. @M2tec better use aforementioned function and apply it  to both models in https://github.com/M2tec/OpenUpgrade/blob/50c02fb3cfed1cdca29e5ecf9bcd0690f21f8d42/addons/stock/migrations/8.0.1.1/pre-migration.py#L178

pedrobaeza:
The problem this time is going to be with duplicated field entries, as there are fields with the same name in the three tables.

ghost:
Ok I will look into the function when I have time

ghost:
If you insert this in stock pre-migrations script:
openupgrade.rename_models(cr, [('stock.picking.in','stock.picking')])

you get the following error:

IntegrityError: duplicate key value violates unique constraint "ir_model_obj_name_uniq"
DETAIL:  Key (model)=(stock.picking) already exists.

hbrunn (approved):
yes, that's what @pedrobaeza wrote before. Leave it as it is

StefanRijnhart (approved):
Thank you!

ghost:
Ok cool, first patch in opensource software :). Thanks guys

pedrobaeza:
I hope this is the first of many to come :wink:

PR #706, opened 2016-12-22
8.0 Lost file attachments for products

ghost:
An attachment has been saved to a product but is not available after the upgrade to odoo v8

Current behavior before PR:
Attachment is not available in v8 in a product form

Attachment is available

To find the correct attachment files the id and table name are searched. This causes missing attachment because table names have changed
V7: product form -> product.product
product variant (unsupported module?)
V8: product form -> product.template
product variant form-> product.template

I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pra

### DIFF
--- a/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
@@ -56,3 +56,13 @@ def migrate(cr, version):
         SET res_model = 'stock.picking'
         WHERE res_model in ('stock.picking.in', 'stock.picking.out');
         """)
+    
+    # Product.template is used for non variant product in v7 this was
+    # product.product
+    openupgrade.logged_query(
+        cr, """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It seems you use attachments a lot, hehe

product.template table still exists on version 8. You are moving the attachment to the product variant level. Or the problem is that the template IDs change?

ghost:
Yes we use it for product documents / design drawings. Maybe I am the only one :). 

No I do not move the products to the variant level. A normal product creates an attachment with 
model_id = product.template in table ir_attachment in v8. This used to be product.product in v7. 

I don't think it will be very useful to create product documents on the variant level. This would be very difficult to keep up to date due to the large amounts.

pedrobaeza:
This PR is incorrect and has been reverted in f4ff1d62de1ad8da16eaa948bfc083fa7bbbe24c

PR #708, opened 2016-12-24
[ADD] marketing_campaign: Migration script

pedrobaeza:


pedrobaeza:
Optimistically merging

Rathnakreeswaran:
marketing_campaign module is not available in odoo 11. because it is moved to enterprise edition.so which module will be the alternative for it @pedrobaeza sir? whether i have to uninstall that module from the odoo 11 migrated database?

pedrobaeza:
There's none. Please don't hijack threads with your questions.

PR #710, opened 2016-12-26
[ADD] speedup sale migration by prepopulating computed fields

hbrunn:

Current behavior before PR:

--

### DIFF
--- a/addons/sale/migrations/9.0.1.0/post-migration.py
+++ b/addons/sale/migrations/9.0.1.0/post-migration.py
@@ -55,22 +55,6 @@ def set_crm_team_message_types(env):
             (4, env.ref('sale.mt_salesteam_invoice_created').id)]})
 
 
-def recalculate_prepopulated_fields(env):
-    """ We prepopulated only lines with uom=product_uom, trigger a recompute
-    for the rest here"""
-    env.cr.execute(
-        """select l.id
-        from sale_order_line l
-        join sale_order_line_invoice_rel r on r.order_line_id=l.id
-        join account_invoice_line il on r.invoice_line_id=il.id
-        where l.product_uom<>il.uom_id
-        """)
-    env['sale.order.line'].invalidate_cache(
-        fnames=['invoice_status', 'qty_invoiced', 'qty_to_invoice'],
-        ids=[x for x, in env.cr.fetchall()]
-    )
-
-
 @openupgrade.migrate()
 def migrate(cr, version):
     env = api.Environment(cr, SUPERUSER_ID, {})
@@ -80,4 +64,3 @@ def migrate(cr, version):
     set_crm_team_message_types(env)
     openupgrade.load_data(
         cr, 'sale', 'migrations/9.0.1.0/noupdate_changes.xml')
-    recalculate_prepopulated_fields(env)
--- a/addons/sale/migrations/9.0.1.0/post-migration.py
+++ b/addons/sale/migrations/9.0.1.0/post-migration.py
@@ -55,6 +55,22 @@ def set_crm_team_message_types(env):
             (4, env.ref('sale.mt_salesteam_invoice_created').id)]})
 
 
+def recalculate_prepopulated_fields(env):
+    """ We prepopulated only lines with uom=product_uom, trigger a recompute
+    for the rest here"""
+    env.cr.execute(
+        """select l.id
+        from sale_order_line l
+        join sale_order_line_invoice_rel r on r.order_line_id=l.id
+        join account_invoice_line il on r.invoice_line_id=il.id
+        where l.product_uom<>il.uom_id
+        """)
+    env['sale.order.line'].invalidate_cache(
+        fnames=['invoice_status', 'qty_invoiced', 'qty_to_invoice'],
+        ids=[x for x, in env.cr.fetchall()]
+    )
+
+
 @openupgrade.migrate()
 def migrate(cr, version):
     env = api.Environment(cr, SUPERUSER_ID, {})
@@ -64,3 +80,4 @@ def migrate(cr, version):
     set_crm_team_message_types(env)
     openupgrade.load_data(
         cr, 'sale', 'migrations/9.0.1.0/noupdate_changes.xml')
+    recalculate_prepopulated_fields(env)
--- a/addons/sale/migrations/9.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/9.0.1.0/pre-migration.py
@@ -65,8 +65,9 @@ def prepopulate_fields(cr):
         with line2qty_invoiced(id, qty_invoiced) as (
             select l.id, sum(
                 case
-                    when i.type='out_invoice' then il.quantity
-                    when i.type='out_refund' then -il.quantity
+                    when i.state != 'cancel' and i.type='out_invoice' then il.quantity
+                    when i.state != 'cancel' and i.type='out_refund' then -il.quantity
+                    else 0
                 end * uom_invoice.factor / uom_sale.factor
             ) s
             from sale_order_line l
--- a/addons/sale/migrations/9.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/9.0.1.0/pre-migration.py
@@ -85,12 +85,7 @@ def prepopulate_fields(cr):
     cr.execute(
         """\
         update sale_order_line
-        set qty_to_invoice=case
-            when pt.invoice_policy = 'order' then
-                product_uom_qty - coalesce(qty_invoiced, 0)
-            else
-                qty_delivered - coalesce(qty_invoiced, 0)
-        end
+        set qty_to_invoice=product_uom_qty - coalesce(qty_invoiced, 0)
         from product_product p
         join product_template pt on p.product_tmpl_id=pt.id
         where sale_order_line.product_id=p.id
--- a/addons/sale/migrations/9.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/9.0.1.0/pre-migration.py
@@ -85,8 +85,15 @@ def prepopulate_fields(cr):
     cr.execute(
         """\
         update sale_order_line
-        set qty_to_invoice=product_uom_qty - qty_invoiced
-        where qty_invoiced is not null
+        set qty_to_invoice=case
+            when pt.invoice_policy = 'order' then
+                product_uom_qty - coalesce(qty_invoiced, 0)
+            else
+                qty_delivered - coalesce(qty_invoiced, 0)
+        end
+        from product_product p
+        join product_template pt on p.product_tmpl_id=pt.id
+        where sale_order_line.product_id=p.id
         """)
     cr.execute('alter table sale_order_line add column currency_id integer')
     cr.execute(
@@ -101,7 +108,7 @@ def prepopulate_fields(cr):
         """\
         update sale_order_line
         set invoice_status=(
-            case when qty_to_invoice<=0 then 'invoiced'
+            case when coalesce(qty_to_invoice, 0)<=0 then 'invoiced'
             else 'to invoice' end
         )
         where state in ('sale', 'done')
--- a/addons/sale/migrations/9.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/9.0.1.0/pre-migration.py
@@ -56,9 +56,79 @@ def map_order_state(cr):
         WHERE sol.order_id = so.id""")
 
 
+def prepopulate_fields(cr):
+    """ Recomputing a fields will be very expensive via the ORM, so do it
+    here for fields where the computation is trivial"""
+    cr.execute('alter table sale_order_line add column qty_invoiced numeric')
+    cr.execute("update sale_order_line set qty_invoiced=0")
+    cr.execute(
+        """\
+        with line2qty_invoiced(id, qty_invoiced) as (
+            select l.id, sum(
+                case
+                when i.type='out_invoice' then il.quantity
+                when i.type='out_refund' then -il.quantity end
+            ) s
+            from sale_order_line l
+            join sale_order_line_invoice_rel r on r.order_line_id=l.id
+            join account_invoice_line il on r.invoice_line_id=il.id
+            join account_invoice i on i.id=il.invoice_id
+            where l.product_uom=il.uom_id group by l.id
+        )
+        update sale_order_line
+        set qty_invoiced=line2qty_invoiced.qty_invoiced
+        from line2qty_invoiced where line2qty_invoiced.id=sale_order_line.id
+        """)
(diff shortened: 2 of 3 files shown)


addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,6 +55,22 @@ def set_crm_team_message_types(env):
...
+    env.cr.execute(
+        """select l.id
+        from sale_order_line l
+        join sale_order_line_invoice_rel r on r.order_line_id=l.id
+        join account_invoice_line il on r.invoice_line_id=il.id
+        where l.product_uom<>il.uom_id
+        """)
+    env['sale.order.line'].invalidate_cache(
+        fnames=['invoice_status', 'qty_invoiced', 'qty_to_invoice'],
+        ids=[x for x, in env.cr.fetchall()]
```
pedrobaeza: Shouldn't this be `ids[x[0] for x in ...]`?

hbrunn: no, look closely for the comma

pedrobaeza: Ah, a bit weird syntax, but OK, you're right

hbrunn: the weird syntax is actually slightly faster - compare

`python -m timeit '[x for x, _ in [(y, y) for y in range(100000)]]'`

and

`python -m timeit '[x[0] for x in [(y, y) for y in range(100000)]]'`

pedrobaeza: Noted for future list comprehension that I make.

hbrunn: this can also be helpful in a normal loop when you iterate over a huge result set, but do quite little per iteration. Then

`for a, b, c, d in cr.fetchall()`

performs better than

`for row in cr.fetchall()`

if within the loop, all values of `row` are used

pedrobaeza: Thanks for the info. The problem is if the list if very huge or you don't know all possible values (in a `select *` statement), but obviously that cases can be always contained.

NL66278:
It might speed things up even more if you first add all the fields to a table and then fill the values with one statement. Now you have to go several times to all or most of the sale orders and lines.

hbrunn:
yes, good point.

But at the moment I'm busy with a general speedup of recomputations of which I promise myself much more performance gains, because something is fishy here. It runs currently for testing purposes, afterwards I'll propose this to upstream, OCB and here https://github.com/hbrunn/OCB/commit/25e5f9526f182d18dd37a4fbcea07f638bea2a92

There are two problems with current recomputation: The time used grows nonlinear, that's a problem with my 300k sale order lines. Additionally, what makes this hard to debug is that there's some nondeterminism going on, two times recomputing the same field on the same rows yields a different execution path (or the same, that's up to chance).

NL66278:
Apart from the general rewrite of recomputation you might also look at this PR: https://github.com/OCA/OpenUpgrade/pull/735
Allows to just stop the expensive recomputations, while having Odoo create the fields themselves. Consequence is that the actual data_migration must move to the post_migration

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,83 @@ def map_order_state(cr):
...
+        )
+        update sale_order_line
+        set qty_invoiced=line2qty_invoiced.qty_invoiced
+        from line2qty_invoiced where line2qty_invoiced.id=sale_order_line.id
+        """)
+    cr.execute('alter table sale_order_line add column qty_to_invoice numeric')
+    cr.execute(
+        """\
+        update sale_order_line
+        set qty_to_invoice=product_uom_qty - qty_invoiced
```
JordiBForgeFlow: @hbrunn This sql sentence is not correct. It does not replicate the same conditions as https://github.com/odoo/odoo/blob/9.0/addons/sale/sale.py#L516

For example, in products where the invoice policy is set based on delivered quantities the quantity to invoice is to be based on the quantity delivered - quantity invoiced.

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,83 @@ def map_order_state(cr):
...
+        from sale_order o join product_pricelist p on o.pricelist_id=p.id
+        where l.order_id=o.id
+        """)
+    cr.execute('alter table sale_order_line add column invoice_status varchar')
+    cr.execute("update sale_order_line set invoice_status='no'")
+    cr.execute(
+        """\
+        update sale_order_line
+        set invoice_status=(
+            case when qty_to_invoice<=0 then 'invoiced'
```
JordiBForgeFlow: replace with `case when coalesce(qty_to_invoice,0)<=0 then 'invoiced'`

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,83 @@ def map_order_state(cr):
...
+        update sale_order_line
+        set qty_invoiced=line2qty_invoiced.qty_invoiced
+        from line2qty_invoiced where line2qty_invoiced.id=sale_order_line.id
+        """)
+    cr.execute('alter table sale_order_line add column qty_to_invoice numeric')
+    cr.execute(
+        """\
+        update sale_order_line
+        set qty_to_invoice=product_uom_qty - qty_invoiced
+        where qty_invoiced is not null
```
JordiBForgeFlow: This where condition is not correct. Instead use:
```
update sale_order_line
set qty_to_invoice=product_uom_qty - coalesce(qty_invoiced,0)
```

This issue has been giving me a hard time.

hbrunn:
@jbeficent thanks for you input, done. Next time, just create a PR

pedrobaeza:
What is the status of this?

hbrunn:
I'm waiting for @jbeficent to test the last commit where I integrated his points

pedrobaeza:
@jbeficent can you review? Travis is red anyway.

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,90 @@ def map_order_state(cr):
...
+        update sale_order_line
+        set qty_invoiced=line2qty_invoiced.qty_invoiced
+        from line2qty_invoiced where line2qty_invoiced.id=sale_order_line.id
+        """)
+    cr.execute('alter table sale_order_line add column qty_to_invoice numeric')
+    cr.execute(
+        """\
+        update sale_order_line
+        set qty_to_invoice=case
+            when pt.invoice_policy = 'order' then
```
omar7r: invoice_policy field in product.template doesn't exists in pre-migration.py because it's a new field in 9.0 version If you move this query to post-migrations it works

pedrobaeza:
@jbeficent please feedback about this PR.

@hbrunn check last comment and red status

hbrunn:
@pedrobaeza last comment defeats the purpose of the pr, the code is fixed but travis is acting up

pedrobaeza:
Again the TRAVIS_COMMIT error, so I close and reopen the PR for forcing a correct build.

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,85 @@ def map_order_state(cr):
         WHERE sol.order_id = so.id""")
 
 
+def prepopulate_fields(cr):
+    """ Recomputing a fields will be very expensive via the ORM, so do it
+    here for fields where the computation is trivial"""
+    cr.execute('alter table sale_order_line add column qty_invoiced numeric')
```
pedrobaeza, line 62: ADD `set default=0`

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,85 @@ def map_order_state(cr):
...
+            join account_invoice i on i.id=il.invoice_id
+            join product_uom uom_invoice on il.uom_id=uom_invoice.id
+            join product_uom uom_sale on l.product_uom=uom_sale.id
+            group by l.id
+        )
+        update sale_order_line
+        set qty_invoiced=line2qty_invoiced.qty_invoiced
+        from line2qty_invoiced where line2qty_invoiced.id=sale_order_line.id
+        """)
+    cr.execute('alter table sale_order_line add column qty_to_invoice numeric')
```
pedrobaeza, line 85: ADD `set default=product_uom_qty`

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,85 @@ def map_order_state(cr):
...
+        """)
+    cr.execute('alter table sale_order_line add column currency_id integer')
+    cr.execute(
+        """\
+        update sale_order_line l set currency_id=p.currency_id
+        from sale_order o join product_pricelist p on o.pricelist_id=p.id
+        where l.order_id=o.id
+        """)
+    cr.execute('alter table sale_order_line add column invoice_status varchar')
+    cr.execute("update sale_order_line set invoice_status='no'")
```
pedrobaeza, line 102: Instead of this, just add in the previous statement `set default='no'`

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -56,9 +56,85 @@ def map_order_state(cr):
...
+        """\
+        update sale_order_line
+        set invoice_status=(
+            case when coalesce(qty_to_invoice, 0)<=0 then 'invoiced'
+            else 'to invoice' end
+        )
+        where state in ('sale', 'done')
+        """)
+    cr.execute('alter table sale_order add column invoice_status varchar')
+    cr.execute("update sale_order set invoice_status='no'")
```
pedrobaeza, line 113: The same (`set default='no'`)

pedrobaeza (changes requested):

mvaled:
Should we merge this?

mvaled:
I'm testing this today.  I'll be reporting later (tomorrow perhaps)

mvaled:
TL;DR It seems this is incorrect.

### Experiment

I took a DB with Odoo 8, and clone it twice:

```
createdb -T o8 o9base
createdb -T o8 o9speedy
```

Then I migrated the `o9base` without this patch (https://github.com/merchise-autrement/openupgrade/tree/merchise-develop-9.0) and the `o9speedy` with the patch (https://github.com/merchise-autrement/OpenUpgrade/tree/merchise-9.0-sale-speedup).

However comparing the resulting `sale_order_line` and `sale_order` gives a lot of differences:

For instance, the field `sale.order.line.qty_invoiced` is different in almost all lines:

```
$ diff -u <(psql -d o9speedy -Ac 'select id, qty_invoiced from sale_order_line where order by id;') <(psql -d o9base -Ac 'select id, qty_invoiced from sale_order_line order by id;') | wc -l
191365
```

The first thing I notice is that where the `o9base` get 0, `o9speedy` gets NULL.  Maybe this is not a problem.  Because we can always do `UPDATE sale_order_line SET qty_invoiced=0 WHERE qty_invoiced IS NULL`.  Let's do it:

```
$ psql -d o9speedy -c "UPDATE sale_order_line SET qty_invoiced=0 WHERE qty_invoiced IS NULL"
UPDATE 62646
```

However, we still have a lot of differences:

```
$ diff -u <(psql -d o9speedy -Ac 'select id, qty_invoiced from sale_order_line order by id;') <(psql -d o9base -Ac 'select id, qty_invoiced from sale_order_line order by id;') | wc -l
82575
```

Let's see some records that differ:

```
-38|15.0000000000000000
-39|1.00000000000000000000
+38|0
+39|0
```

For the record 38, the speedier version computes 15 while the normal version computes 0.  Other fields show as much differences as well.  I'll try to look at these differences and see if I can spot the issue.

mvaled:
I pushed 8625c1029ad5e5851a3f1e82f0a7fe01b8a14546 to fix the qty_invoiced amount.

hbrunn:
can you PR this to this PR's branch? Then I'll merge this in this PR. If you want to speed up getting this merged, you can check @pedrobaeza's comments to. I won't have time to address them any time soon.

mvaled:
I already did: https://github.com/hbrunn/OpenUpgrade/pull/4

mvaled:
I'll be addressing @pedrobaeza's comments soon.

mvaled:
See PR https://github.com/hbrunn/OpenUpgrade/pull/5

addons/sale/migrations/9.0.1.0/pre-migration.py
mvaled: About this PG says: cannot use column references in default expression

mvaled: From PG doc:

> DEFAULT default_expr

> The DEFAULT clause assigns a default data value for the column whose column definition it appears within. The value is any *variable-free expression* (subqueries and cross-references to other columns in the current table are not allowed). The data type of the default expression must match the data type of the column.

> The default expression will be used in any insert operation that does not specify a value for the column. If there is no default for a column, then the default is null.

pedrobaeza: OK, I didn't know. Then we should still keep here 2 sentences

mvaled:
I created another PR to the original branch.  It addresses @pedrobaeza's
requests.  One of the requested defaults cannot be set by SQL.

mvaled:
Still working on this.  The field `qty_to_invoice` shows differences.

mvaled:
Hum. The `qty_to_invoice` depends on the computation `qty_delivered`.  However `qty_delivered` is not computed in this addon but in `sale_analytic`.

addons/sale/migrations/9.0.1.0/pre-migration.py
mvaled: I'm working on https://github.com/hbrunn/OpenUpgrade/pull/5.  I hope next week this can be merged.

pedrobaeza:
Any news on this?

mvaled:
I've been unable to keep working on this for the past weeks.  We're heavily busy migrating ~50 addons to Odoo 9 and then to Odoo 10.  After they all migrated to 9, I could try another round of DB migration and see if I could work on this.

pedrobaeza:
OK, let me know. I don't see this as critical because times are more or less manageable in my cases, but maybe with a lot of sale orders it's not the same.

pedrobaeza:
Closing this due to errors on the migration. If anyone wants to continue this experiment, can take this code and go on with the needed changes, proposing the new alternative.

PR #712, opened 2016-12-26
[MIG] event: Migration scripts

cubells:
* Analysis work
* Pre-migration script
* Post-migration script


pedrobaeza:
Why have you used 8.0 commits? In this case, you don't have to follow the migration guide: don't import commits corresponding to previous version.

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,127 @@
...
+
+event        / event.event              / country_id (many2one)         : not a function anymore
+# Nothing to do as country_id is now a related field from address_id
+
+event        / event.event              / email_confirmation_id (many2one): DEL relation: email.template
+event        / event.event              / email_registration_id (many2one): DEL relation: email.template
+
+
+event        / event.event              / event_logo (html)             : NEW
+# Noting to do
```
pedrobaeza: s/Noting/Nothing

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,127 @@
...
+# Nothing to do
+
+event        / event.event              / color (integer)               : NEW
+# Nothing to do
+
+event        / event.event              / country_id (many2one)         : not a function anymore
+# Nothing to do as country_id is now a related field from address_id
+
+event        / event.event              / email_confirmation_id (many2one): DEL relation: email.template
+event        / event.event              / email_registration_id (many2one): DEL relation: email.template
```
pedrobaeza: This worries me, as we have customers using specific templates for each event. How Odoo v9 handles this? Anyway, at least we need to keep the columns if an specific module restoring this function comes.

addons/event/migrations/9.0.0.1/post-migration.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2016 Tecnativa - Vicent Cubells <vicent.cubells@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def update_seats_availability(cr):
+    cr.execute("""
+                UPDATE event_event
```
pedrobaeza: You don't need to indent so much these lines. One indent level more than previous line is enough.

addons/event/migrations/9.0.0.1/post-migration.py
```diff
@@ -0,0 +1,22 @@
...
+                SET seats_availability = 'limited'
+                WHERE seats_max IS NOT NULL
```
pedrobaeza: seats_max is an integer, so the value is NULL or 0 when you don't set it?

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,127 @@
...
+event        / event.event              / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+event        / event.event              / seats_availability (selection): NEW required: required, selection_keys: ['limited', 'unlimited'], req_default: unlimited
+# Updated from seats_max
+
+event        / event.event              / type (many2one)               : was renamed to event_type_id [nothing to to]
+# Nothing to do
+
+event        / event.mail               / event_id (many2one)           : NEW relation: event.event, required: required
```
pedrobaeza, line 37: Is this the replacement of the two dissapeared fields with templates? Without looking the code or runbot, it seems this is a way to program reminders and steps to send emails in certain stages.

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,127 @@
...
+event        / event.registration       / message_follower_ids (many2many): type is now 'one2many' ('many2many')
+# Concerns module Mail (has to be done in another migration script specifically for that module) ---> Nothing to do
+
+event        / event.registration       / nb_register (integer)         : DEL required: required, req_default: 1
+# Nothing to do: new relation
+
+event        / event.registration       / user_id (many2one)            : DEL relation: res.users
+# Nothing to do: new relation
+
+event        / event.type               / default_email_event (many2one): DEL relation: email.template
```
pedrobaeza, line 84: How do we put now default event.mail?

addons/event/migrations/9.0.0.1/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+    ],
+    'res_partner': [
+        ('speaker', None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.rename_columns(cr, column_renames)
```
pedrobaeza: Line at the end

pedrobaeza (changes requested):

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
cubells: Columns are renamed.

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
cubells: No, it isn't. The field contains the template of the mail that will be automatically sent. It isn't the same that email_confirmation_id and email_registration_id fields.

addons/event/migrations/9.0.0.1/openupgrade_analysis_work.txt
cubells: It seems this feature don't exist in version 9.0.

pedrobaeza:
This is still pending the creation of mail records.

pedrobaeza:
cc @Tecnativa

cubells:
cc @pedrobaeza

pedrobaeza (approved):

PR #715, opened 2016-12-27
[IMP] base: Merge obsolote modules

pedrobaeza:
`remove_obsolete_modules` method in post-migration is not correctly working, and
it's not as clean as this formula due to the new openupgradelib feature of merging
modules.

I have removed it and use the new method, including as well other OCA modules that
are obsolete now in version 9.


pedrobaeza:
@StefanRijnhart, I think you implemented previous method. What do you think?

pedrobaeza:
Any feedback for this?

jjscarafia:
I like the new approach that takes advantages of new merge functionality, :+1: for me but sill would like to hear @StefanRijnhart feedback

StefanRijnhart (approved):

PR #717, opened 2017-01-02
[ADD] try to avoid using addresses with names as company address

hbrunn:
the rationale here is that addresses with names might be contact persons, while we can be quite sure an address without a name has something to do with the company itself.


pedrobaeza (approved):
It depends a lot on the use case of the customer itself (it depends on their habits), but it can be logical this way.

The ideal one should be to compare fields to see if they are the same and then merge in that case.

pedrobaeza:
Merging this one anyway.

PR #730, opened 2017-01-11
[FIX] sales_team migration error if crm is not installed

### LINKED ISSUE #729: [8.0] sales_team error (relation "crm_case_section" does not exist)
Impacted versions: 8.0 (target upgrade version)

Steps to reproduce:
* install a fresh V7.0 database, -i point_of_sale,purchase. (i guess that less module is sufficient, but not tested)
* upgrade to V8.0

Current behavior:
* During the migration, sales_team (new 8.0 module is installed by [dependencies](https://github.com/OCA/OpenUpgrade/blob/8.0/addons/sales_team/__openerp__.py#L32)).
* sales_team migration is executed.
* an error is raised :

        error in migration script OpenUpgrade/addons/sales_team/migrations/8.0.1.0/pre-migration.py: 
        relation "crm_case_section" does not exist
        LINE 1: UPDATE crm_case_section SET code = '_DM' WHERE code = 'DM'

Expected behavior:
No error.

Analyse & Solution:
Part of the crm migration should be executed early. That's why part of the code is in sales_team. But migration should include test to know if crm is installed.

legalsylvain:
FIX #729

--

### DIFF
--- a/addons/sales_team/migrations/8.0.1.0/post-migration.py
+++ b/addons/sales_team/migrations/8.0.1.0/post-migration.py
@@ -20,12 +20,19 @@
 #
 ##############################################################################
 
+import logging
 from openerp import pooler, SUPERUSER_ID
 from openerp.openupgrade import openupgrade, openupgrade_80
+
+logger = logging.getLogger('OpenUpgrade.sales_team')
 
 
 @openupgrade.migrate(no_version=True)
 def migrate(cr, version):
+    if not openupgrade.is_module_installed(cr, 'crm'):
+        logger.info(
+            "sales_team post-migration skipped, because crm is not installed")
+        return
     pool = pooler.get_pool(cr.dbname)
     uid = SUPERUSER_ID
     openupgrade_80.set_message_last_post(
--- a/addons/sales_team/migrations/8.0.1.0/pre-migration.py
+++ b/addons/sales_team/migrations/8.0.1.0/pre-migration.py
@@ -20,8 +20,10 @@
 #
 ##############################################################################
 
+import logging
 from openerp.openupgrade import openupgrade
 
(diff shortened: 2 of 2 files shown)


addons/sales_team/migrations/8.0.1.0/post-migration.py
```diff
@@ -20,12 +20,19 @@
...
 ##############################################################################
 
+import logging
 from openerp import pooler, SUPERUSER_ID
 from openerp.openupgrade import openupgrade, openupgrade_80
 
+logger = logging.getLogger('OpenUpgrade.sales_team')
+
 
 @openupgrade.migrate(no_version=True)
```
pedrobaeza, line 30: Better to remove `no_version` and then the scripts are not executed on module install (which is your case). As you were the original author of the module, maybe you had a reason for putting that.

legalsylvain: Hi @pedrobaeza. 
Thanks a lot for your review.

- sales_team doesn't exist in V7.0.
- this module will be installed on V8.0, by dependencies.
- this migration script should be executed if crm is installed on V7.0.
- this migration script should not be executed if crm is not installed on V7.0.

so no_version should stay, but crm test is necessary.

What do you think ?

pedrobaeza: OK, understood, but then move this to crm migration scripts...

legalsylvain: No. it's too late in crm. crm.case.section moved from crm module (V7.0) to sales_team module (V8.0).

https://github.com/OCA/OpenUpgrade/blob/8.0/addons/sales_team/sales_team.py#L28

kind regards.

pedrobaeza:
Ok then

PR #731, opened 2017-01-11
[9.0][point_of_sale] various fixes

JordiBForgeFlow:

Various errors appearing during the migration of module 'point_of_sale', at post-migration stage.

Current behavior before PR:
Errors stop the migration.

Migration of point_of_sale occurs without disruption

--

### DIFF
--- a/addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
+++ b/addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
@@ -12,6 +12,14 @@ column_defaults = {
         ('session_id', None)
     ],
 }
+barcode_rule_types_map = {
+    openupgrade.get_legacy_name('barcode_weight'): 'weight',
+    openupgrade.get_legacy_name('barcode_price'): 'price',
+    openupgrade.get_legacy_name('barcode_discount'): 'discount',
+    openupgrade.get_legacy_name('barcode_customer'): 'client',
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
@pedrobaeza @sergio-teruel

sergio-teruel (approved):

addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py
```diff
@@ -12,6 +12,14 @@
         ('session_id', None)
     ],
 }
+barcode_rule_types_map = {
+    'openupgrade_legacy_9_0_barcode_weight': 'weight',
```
pedrobaeza: Use `openupgrade.get_legacy_name('field')` instead

pedrobaeza:
Please squash commits and don't merge until you are sure this is the definitive solution

NL66278:
:+1:  Tested the changes and solved also my problems with migrating the OpenUpgrade test database from 8.0 to 9.0. Please take the remarks of @pedrobaeza in account, and then it will be ready to merge IMHO.

StefanRijnhart:
After changing the hardcoded field names to a call to get_legacy_name, we can squash the commits when merging.

NL66278:
:+1: Seems ready to merge now...

PR #732, opened 2017-01-11
[FIX] project: Create analytic lines from task works if needed

pedrobaeza:
If the module project_timesheet is not installed, we need to convert
project.task.work elements, or we will lose this history.

This is done inserting new records in account_analytic_line, and
adding one extra column work_id in case other modules need to
add more elements for each converted line.


pedrobaeza:
cc @Tecnativa

pedrobaeza:
Merging optimistically, as there's no way to test this with the current test toolset

PR #733, opened 2017-01-11
[WIP][8.0] sql quants creation

### LINKED ISSUE #724: [8.0] stock migration very slow
Impacted versions: 8.0

Steps to reproduce:
Using a Database with 2.000.000 stock moves, trying to openupgrade it from 7.0 to 8.0. During script migration, in the [quant generation script](https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L934), the process is very slow. 
10.000 stock move managed in 25 minutes. So, full quants generation will take 5.000 minutes. (3,5 days). That is quite annoying. 

The server is dedicated for the migration with a not so bad hardware : 
- RAM : 32 GO
- HD : SSD
- Processor : [Intel Xeon E3-1245 v5](https://www.cpubenchmark.net/cpu.php?cpu=Intel+Xeon+E3-1245+v5+%40+3.50GHz&id=2674) @ 3,5 GHz
* Cpu Benchmark Note : 10.350. (but just one is used for Odoo). so the real note is more ~2.000
* No of Cores: 4 (2 logical cores per physical)

What I tried 
* Set a higher value for shared_buffers value in the ./postgresql.conf file. (128 MB -> 10 GO) --> Bad result. (longer process)
* Set a higher value for limit_memory_hard et limit_memory_soft. (~600MB -> 10 GO) --> no result.

Odoo settings : (quite simple)
- multi company. (~25)
- one warehouse per company
- one internal location by warehouse
- no product lots management

~I was asking me if it's possible to create only one global quant per product in this configuration. I'll loose maybe some features based on history... In a generic way, I could add extra keys in the update process like -u all --stock-migration-simple-or-whatever, and make some conditionnal migration script.~
--> Dummy remarks. Talked with @sebastienbeau. This solution will not work because quants should exist in external customers location too.

Do you have some ideas for optimizations ? I'm ready to work on that but I'm not a quant expert, so any advice is welcome.
CC : @pedrobaeza, @StefanRijnhart. 

Did you faced duration problems with databases with similar size ?

legalsylvain:
This PR try to fix  #724.

### Current behavior before PR:
quants creation is quite slow with current algorigthm. (including huge recent work made by @jbeficent 
some info : 
- 250.000 Stock move -> 1 days;
- 2.000.000 stock move ->  3,5 days ; (other system)

### Implementation
based on [Vauxoo work](https://github.com/Vauxoo/sql_migration), this PR replace quants creation by ORM by quants creation by SQL request. 
Because it's quite experimental, an extra key should be added in the config file to activate SQL requests : 

        sql_quants_creation = True

### State of this PR
- test on small database with some quants : Seems OK. (see below)
- test on database with 200k move : pending.

Any PoV about this PR are welcome. 
CC :  @josemoralesp, @sebastienbeau, @pedrobaeza, @hbrunn, @jbeficent. 

#### Sample :
List of stock moves
stock_move

Quants generated by the current code
orm_quants

Quants generated by this PR
sql_quants

Note the incoming date seems to be correct with this script. (for the being, the incoming date of the quant seems to be the date of the migration)

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -963,6 +963,14 @@ def sql_migrate_stock_qty(cr, registry):
             moves_qty))
     f = tools.file_open('stock/migrations/8.0.1.1/quants_creation.sql')
     cr.execute(f.read())
+    error_qty = cr.fetchone()
+    if error_qty:
+        logger.error(
+            "%d / %d stock moves failed. Please see the table"
+            " 'stock_quants_openupgrade_8_log' for more details." % (
+                error_qty, moves_qty))
+    else:
+        logger.info("%d done moves processed by SQL." % (moves_qty))
 
 
 def migrate_stock_production_lot(cr, registry):
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -21,7 +21,7 @@
 ##############################################################################
 
 import logging
-from openerp import api, SUPERUSER_ID
+from openerp import api, SUPERUSER_ID, tools
 from openerp.openupgrade import openupgrade, openupgrade_80
 from openerp.modules.registry import RegistryManager
 from openerp import SUPERUSER_ID as uid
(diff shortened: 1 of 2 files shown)


legalsylvain:
I just tested this PR with a DB with 250k stock moves of an akretion customer.  
- 44 minutes. (~95 moves / second).

I let @sebastienbeau share his point of view about quality of the data, as he made previously regular upgrade.

sebastienbeau:
Hi I have done some check on the database migrated here are the result

## Context:
My customer only use 1 unit of mesure and do not have prodlot number

## Result:

* The quantity of product per location are the same.
I have some little difference (3 differences) in some sub location of stock but no big deal (more over we drop that sub-location after the migration). Maybe I case that you have a lot of sub-location of stock you may obtain some difference.

* The number of quant are different (but not a big difference):
I have manually analyzed some difference / issue

### With the fast version
- Issue on inventory : 2 inventory have generated one quant for one product indeed of 2 quants

### With the slow version
- quant in_date are wrong

### On both
Some time quant are splitted or merge but have an acceptable historic for example

1: I have received 1500 product to stock A
2: I move 100 to stock B
3: I move 400 to stock B
4: I sell 142 product to many customer (small picking)

At this end I have 358 product in stock B

Slow solution have generated 2 quants
- 14 PCE with historic on the move of 100PCE
- 344 PCE with historic on the move of 400PCE

Fast solution have generated one quand of 358 based on historic of 400PCE.

Sometime is the oposite more move for the fast solution but in general the fast solution generate less quant

## To analyse my result I have run the following request (export on csv) and use meld

* SELECT count(id), sum(qty) FROM stock_quant;
* SELECT location_id, count(id), sum(qty) FROM stock_quant GROUP BY location_id;
* SELECT product_id, location_id, count(id), sum(qty) FROM stock_quant GROUP BY product_id, location_id order by product_id, location_id;

To give you an idea of qty and quant on the database the result of "SELECT count(id), sum(qty) FROM stock_quant" is

|         | nbr of quant   |       qty|
|---|---|---|
|fast   |  251567      |  1100000685305 |
|slow  |   251585      |  1100000685305 |

pedrobaeza:
@legalsylvain this PR is suppose to work correctly?

pedrobaeza:
Any news on this?

samy-m-dev:
We try that solution with a database of 600k stock moves, and the script is running for more than 2 days now. 

We add a RAISE LOG after each iteration of moves to see what happens. And we notice that the script was very fast at the beginning around 100k moves processed in ~15 minutes (so the estimation of 250k move in 44 minutes seems correct), but then everything start to slow down and now still running for days.

We try on our local computer (Windows) and also on a server with 4 CPU and 64GB of RAM, the results are the same at the beginning very fast and then very slow.

**With the ORM version:**  ~36 hours
**With the SQL version:**  >2 days and still running

@legalsylvain @sebastienbeau Do you have any idea what goes wrong? Or maybe there's a way to optimize it?

I also notice the script create quants for 'done' moves, but that PR doesn't take care of 'assign' moves,

We would like to upgrade our database from 6.0 to 10.0 in a window of 48 hours, but that part of the upgrade make it a challenge. Ultimately we don't really need to keep total history of quants as long as stock quantities are correct, maybe you have an idea for a fast dirty solution?

pedrobaeza:
There are no news for this on a while, so I'm closing this. If anyone wants to continue the work, he/she can take this work and fix the remaining problems.

PR #736, opened 2017-01-13
[FIX] website: Set as updateable website main entry

pedrobaeza:
In Odoo v8, the website main entry is flagged as noupdate=1, and
this causes that changes made in v9 (action and web_icon), are not
updated and thus is hidden in migrated instance.

cc @Tecnativa


pedrobaeza:
Merging optimistically

PR #738, opened 2017-01-17
[FIX] delete the renamed channel rule

hbrunn:
the problem is that this rule won't be touched by the migration because it's noupdate, and it will cause access errors when trying to use mail channels afterwards.
Alternative would be to rename the xmlid to the new one, and clean out the noupdate flag, but I don't see the advantage of this. Manual changes will be gone in any case.

### DIFF
--- a/addons/mail/migrations/9.0.1.0/post-migration.py
+++ b/addons/mail/migrations/9.0.1.0/post-migration.py
@@ -41,8 +41,3 @@ def migrate(cr, version):
         '(mail_message_id, res_partner_id) '
         'select distinct message_id, partner_id from mail_notification '
         'where starred')
-    # with this, the rule and the xmlid will be cleaned up at the end of
-    # the migration
-    cr.execute(
-        'update ir_model_data set noupdate=False where '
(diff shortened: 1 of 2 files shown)


hbrunn:
cf https://github.com/OCA/OCB/blob/8.0/addons/mail/security/mail_security.xml#L6 and https://github.com/OCA/OCB/blob/9.0/addons/mail/security/mail_security.xml#L6

pedrobaeza:
If you are sure Odoo clean the record with this change, then :+1:

hbrunn:
we should see this in the log when travis finishes, works at least locally

pedrobaeza:
OK, cool, another thing I didn't know.

hbrunn:
hmm, doesn't work. Trying again doing the same in pre migration

hbrunn:
...will have to look at this properly some time later. But for the record: You want to delete/deactivate this rule after your migration, otherwise most chat related functions won't work in your migrated database

hbrunn:
superseded by #803

PR #740, opened 2017-01-20
Preserve data if crm_lead_lost_reason is installed

cubells:
crm_lead_lost_reason module features are included in crm module.

These changes preserve module data if user has that module installed.

cc @Tecnativa

### DIFF
--- a/addons/crm/migrations/9.0.1.0/pre-migration.py
+++ b/addons/crm/migrations/9.0.1.0/pre-migration.py
@@ -89,3 +89,6 @@ def migrate(cr, version):
         [('4', '3')], table='crm_lead',
     )
     cr.execute("update crm_lead set type='opportunity' where type is null")
+    # Needed for crm_lead_lost_reason migration
+    cr.execute("ALTER TABLE IF EXISTS crm_lead_lost_reason "
+               "RENAME TO crm_lost_reason")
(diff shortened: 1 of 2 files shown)


addons/crm/migrations/9.0.1.0/pre-migration.py
```diff
@@ -19,6 +19,7 @@
     ],
     'crm_lead': [
         ('section_id', 'team_id'),
+        ('lost_reason_id', 'lost_reason'),
```
pedrobaeza: Don't include this here. You are already doing this in the SQL

addons/crm/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -1,11 +1,11 @@
 ---Fields in module 'crm'---
```
pedrobaeza: Why are you touching this file?

cubells: Undone.

pedrobaeza: No, changes are still here.

addons/crm/migrations/9.0.1.0/pre-migration.py
```diff
@@ -89,3 +89,5 @@ def migrate(cr, version):
         [('4', '3')], table='crm_lead',
     )
     cr.execute("update crm_lead set type='opportunity' where type is null")
+    cr.execute("ALTER TABLE IF EXISTS crm_lead_lost_reason "
```
pedrobaeza: Include a comment indicating that this is for migrating module ....

StefanRijnhart (approved):

pedrobaeza (approved):

PR #745, opened 2017-01-21
[9.0][FIX] *: Handle attachment=True property

pedrobaeza:
It depends on:
- [x] https://github.com/OCA/openupgradelib/pull/66

This commit patches all the existing migrated modules for fixing the binaries on fields with attachment=True property set in v9.

NOTE: There's still some modules pending, like fleet or gamification, with this kind of fields, but as there's not yet migration script for them, it's better to not add anything for not creating false sensation that the module is migrated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ script:
 # don't use pypi's openupgradelib, but the one from source to avoid choking
 # on unreleased features in openupgradelib
     - pip uninstall -q --yes openupgradelib
-    - pip install -q git+https://github.com/OCA/openupgradelib.git@master
+    - pip install -q git+https://github.com/tecnativa/openupgradelib.git@9.0-convert_binary_field_to_attachment
     - createdb $DB
     - wget -q -O- https://github.com/OCA/OpenUpgrade/releases/download/databases/8.0.psql | pg_restore -d $DB --no-owner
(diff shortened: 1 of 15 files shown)


pedrobaeza:
cc @Tecnativa

StefanRijnhart:
Can't we do this in the base module for all binary fields in the database, instead of having to specify the fields per module?

pedrobaeza:
No, it can't be done. It was my first line of investigation, but there's no way to detect on base which fields are going to have the `attachment=True` attribute set.

pedrobaeza:
I need to make some changes to make possible to run in pre-migration. I'll do it later.

JordiBForgeFlow:
What do you mean in the last commit? Can we test the migration now using these PR's?

pedrobaeza:
This is still not working. Let me work on it. 

Last commit is for having correct tests in Travis until openupgradelib PR is merged.

pedrobaeza:
OK, this one is ready to be reviewed

JordiBForgeFlow (approved):
:+1: Tested. Works as expected. Summary of results below.

__Countries__:
image

__Partners__:
image

__Products__:
image

__Live Chat__:
image

__Employees__:
image

__Point of Sale__:
image


__Website__:
image

JordiBForgeFlow:
https://github.com/OCA/OpenUpgrade/pull/745/files#diff-354f30a63fb0907d4ad57269548329e3R45 will need to be reverted once https://github.com/OCA/openupgradelib/pull/66 is merged

StefanRijnhart:
Thank you!

PR #747, opened 2017-01-23
[9.0][product] migrate pricelist_partnerinfo

JordiBForgeFlow:
The data stored in pricelist_partnerinfo is not being migrated to v9. This PR migrates it into the product_supplierinfo

v8:
image

v9 (after migration):
image

--

### DIFF
--- a/addons/product/migrations/9.0.1.2/post-migration.py
+++ b/addons/product/migrations/9.0.1.2/post-migration.py
@@ -193,9 +193,9 @@ def update_product_supplierinfo(env):
         psi.sequence, ppi.min_quantity, ppi.price,  psi.product_tmpl_id,
         psi.delay, psi.company_id
         FROM product_supplierinfo AS psi
-        INNER JOIN %s AS ppi
+        INNER JOIN pricelist_partnerinfo AS ppi
         ON ppi.suppinfo_id = psi.id
-    """ % openupgrade.get_legacy_name('pricelist_partnerinfo'))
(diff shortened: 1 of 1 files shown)


addons/product/migrations/9.0.1.2/pre-migration.py
```diff
@@ -6,6 +6,9 @@
 
 from openupgradelib import openupgrade
 
+table_renames = [
```
pedrobaeza: I think you don't need to rename the table, as it's not dropped during upgrade process.

pedrobaeza (changes requested):
Please add a test for this

StefanRijnhart:
I regret that you replace the existing supplierinfo instead of reusing the existing ones and only add new ones if necessary. This will untraceably remove any custom data that are stored on the supplierinfo. Would you consider refactoring to preserve the existing ones?

StefanRijnhart:
@pedrobaeza adding tests is not really feasible yet. I'm currently looking at reviving #624 though.

pedrobaeza:
OK, I understand that there isn't any demo data for this.

pedrobaeza:
Now that there's the possibility of adding demo data, I think this should be completed.

pedrobaeza:
@jbeficent please complete this

StefanRijnhart:
I'm continuing this here: https://github.com/OCA/OpenUpgrade/pull/1169

PR #752, opened 2017-02-01
[FIX] crm_phonecall migration

cubells:
These changes allows preserve data for crm_phonecall module migration:
https://github.com/OCA/crm/pull/131

cc @Tecnativa

### DIFF
--- a/addons/crm/migrations/9.0.1.0/pre-migration.py
+++ b/addons/crm/migrations/9.0.1.0/pre-migration.py
@@ -25,6 +25,9 @@ column_renames = {
     'section_stage_rel': [
         ('section_id', 'team_id'),
     ],
+    'calendar_event': [
+        ('phonecall_id', None),
+    ],
 }
 
 column_copys = {


addons/crm/migrations/9.0.1.0/pre-migration.py
```diff
@@ -25,6 +25,13 @@
     'section_stage_rel': [
         ('section_id', 'team_id'),
     ],
+    'calendar_event': [
+        ('phonecall_id', ''),
```
pedrobaeza: Second element should be `None`

addons/crm/migrations/9.0.1.0/pre-migration.py
```diff
@@ -33,6 +40,16 @@
     ],
 }
 
+xmlid_renames = [
```
pedrobaeza: This should be done on crm_phonecall itself, not here. Result: if you don't have crm_phonecall, XML-IDs are removed at the end. If you have it, it doesn't crush.

PR #754, opened 2017-02-02
mass_mailing_sending_queue module migration

cubells:
mass_mailing_sending_queue module features are now in mass_mailing module.


pedrobaeza (approved):
Yeah, same feature. The only possible drawback is that you have a mailing not fully completed when migrating. But who are going to do that!!??

PR #766, opened 2017-02-17
[9.0][stock]recompute pack operations for pickings in progress

### LINKED ISSUE #765: [9.0] operations not created for pickings where qty has been reserved but not transferred
Impacted versions: 9.0

Steps to reproduce:
In 8.0
* Install module 'stock'
* Create a delivery order, and ensure availability to transfer, but do not transfer yet

image
* Migrate to v9

Current behavior:
After the migration the  operations are not created:
image

Only after the user presses the button 'Recompute' in the picking will it re-create the operations.
image

Expected behavior:

The migration script should create all operations, so that the target state does not require any further manual action from the user.

JordiBForgeFlow:
Fixes #765
--

### DIFF
--- a/addons/stock/migrations/9.0.1.1/post-migration.py
+++ b/addons/stock/migrations/9.0.1.1/post-migration.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 # @copyright 2016-Today: Odoo Community Association, Therp BV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
 from psycopg2.extensions import AsIs
 from openerp import SUPERUSER_ID, api
 from openupgradelib import openupgrade
+logger = logging.getLogger('OpenUpgrade')
 
 
 def _migrate_tracking(cr):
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/9.0.1.1/post-migration.py
```diff
@@ -50,15 +52,35 @@ def _migrate_pack_operation(env):
...
+    """)
+
+    env.cr.execute("""
+        UPDATE stock_move sm
+        SET location_id = sp.location_id,
+        location_dest_id = sp.location_dest_id
+        FROM stock_picking sp
+        WHERE sp.id = sm.picking_id
+        AND sm.state NOT IN ('done', 'cancel')
+    """)
```
StefanRijnhart, line 77: It looks like you copy the locations from the stock moves to the pickings, and then back again from the pickings to the stock moves. Am I missing something?

hbrunn: I come to the same conclusion. @jbeficent can you elaborate?

pedrobaeza:
@jbeficent any news about this?

pedrobaeza:
No news, so closing.

PR #767, opened 2017-02-18
[FIX] precompute sheet_id in analytic lines

JordiBForgeFlow:
This PR will precompute the field sheet_id in analytic lines, to avoid the recomputation made by odoo, that turns out to produce errors.

--

### DIFF
--- a/addons/hr_timesheet_sheet/migrations/9.0.1.0/pre-migration.py
+++ b/addons/hr_timesheet_sheet/migrations/9.0.1.0/pre-migration.py
@@ -6,8 +6,33 @@
 from openupgradelib import openupgrade
 
 
+def prepopulate_fields(cr):
+
+    cr.execute("""SELECT column_name
+    FROM information_schema.columns
+    WHERE table_name='account_analytic_line' AND
+    column_name='sheet_id'""")
+
+    if not cr.fetchone():
+        cr.execute(
+            """
+            ALTER TABLE account_analytic_line ADD COLUMN sheet_id
+            integer;
+            COMMENT ON COLUMN account_analytic_line.sheet_id IS
+            'Sheet';
+            """)
+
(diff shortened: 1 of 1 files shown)


hbrunn (changes requested):
I'd suggest using the mechanism @NL66278 employs in https://github.com/OCA/OpenUpgrade/pull/663/files#diff-050c291be549d31c2482669f8281e619R169 - this way, you don't miss creating the index on `sheet` id for example: https://github.com/OCA/OCB/blob/9.0/addons/hr_timesheet_sheet/hr_timesheet_sheet.py#L331

JordiBForgeFlow:
@hbrunn Thanks. But then the method `_openupgrade_recompute_fields_blacklist` should be proposed in a separate PR, not associated with the PR specific to accounting, right? cc @NL66278

NL66278:
@jbeficent @hbrunn There actually is a seprate PR for this: https://github.com/OCA/OpenUpgrade/pull/735

I found an issue with the above mentioned PR that blacklists are not always removed automatically. That is why in the reconciliation PR I removed them manually. As I found no obvious way to register globally what blacklists have registered, I got stuck at how to remove them automatically. That is why this PR is still WIP.

JordiBForgeFlow:
@hbrunn @NL66278 As it seems that the PR is not concluded, why not manually creating the index, same as https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L2616?

hbrunn (approved):

hbrunn:
fine with me for now

NL66278:
:+1:  If index is also created as already suggested by @jbeficent

JordiBForgeFlow:
Index is now correctly created:

v8:
image


v9:
image

@hbrunn  can you merge?

PR #768, opened 2017-02-22
[FIX] only delete move lines and modes from centralized journals

### LINKED ISSUE #759: deleting moves of special periods
In https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/pre-migration.py#L86, we delete all moves and lines of special periods which are not in the first special period. I think this is wrong, as we lose all financial corrections made when closing a fiscal year.

@jbeficent could you elaborate your reasoning for deleting the other move lines and do you agree we rather need something like (pseudocode)

```python
for period in special_periods:
    next_normal_period = get_normal_period_after(period)
    move_all_lines_from_period_to_period(period, next_normal_period)
```

the change is simple enough to do, but before submitting I wanted to make sure there's no wrong reasoning on my side. I believe with the current migration, we get inconsistent databases as soon as there's a write-off or some other correction when closing a fiscal year.

If the above is true, I'll also provide a script that generates the missing moves and lines from an existing 8.0 database.

hbrunn:
fixes #759

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -122,14 +122,17 @@ def remove_account_moves_from_special_periods(cr):
             """, (first_nsp_id, tuple(move_ids)))
 
     openupgrade.logged_query(cr, """
-        DELETE FROM account_move_line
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
:+1:

PR #772, opened 2017-03-02
[FIX] rename renamed xmlids for account types

hbrunn:
we missed renaming the xmlids of account types, this way it looks to the user like they were duplicated and it's a pita to clean this up manually.

I only do renames where the equivalences were obvious to me, so if I missed something, I'd appreciate a pointer


JordiBForgeFlow:
:+1:

pedrobaeza (approved):

PR #776, opened 2017-03-10
Updated 8.0-9.0 module list (auth_ldap, rating*)

mlaitinen:
Marked auth_ldap as 'No change'. Also marked `rating`* modules 'Nothing to do' as they are new in Odoo 9.0.

--


openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -50,7 +50,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |auth_crypt                         | Nothing to do                     |
 +-----------------------------------+-----------------------------------+
-|auth_ldap                          |                                   |
+|auth_ldap                          | No change                         |
```
pedrobaeza: Why this text is different? Put also "Nothing to do".

mlaitinen: It says on the top of [the file](https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/doc/source/modules80-90.rst) that
> 'No change' : No change between 8.0 and 9.0;
> 'Nothing to do' : Analyse has been done, and there is nothing to do;

If it's a redundant status, I suggest we remove it.

pedrobaeza: You're right, although there's a connotation: if there's no analysis file, then "No change" is the correct, but we have used "Nothing to do" for both things, so better to unify it.

PR #777, opened 2017-03-10
[9.0][account_check_printing] migrate the payment method for check printing

JordiBForgeFlow:
The payments that were created in 8.0 associated to journals that were related to check printing should have as payment method 'allow_check_writing'. Now they are being set to 'manual'

image

### DIFF
--- a/addons/account_check_printing/migrations/9.0.1.0/post-migration.py
+++ b/addons/account_check_printing/migrations/9.0.1.0/post-migration.py
@@ -23,7 +23,8 @@ def update_payments_from_vouchers(env):
     check_method = env.ref(
         'account_check_printing.account_payment_method_check')
 
-    if openupgrade.column_exists(cr, 'account_journal', 'allow_check_writing'):
+    if openupgrade.column_exists(
+            env.cr, 'account_journal', 'allow_check_writing'):
         env.cr.execute(
             """\
             UPDATE account_payment ap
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
cc @NL66278 Do you agree with this one?

addons/account_check_printing/migrations/9.0.1.0/post-migration.py
```diff
@@ -20,6 +20,28 @@ def update_payments_from_vouchers(env):
         WHERE ap.id = av.id
         """
     )
+    check_method = env.ref(
+        'account_check_printing.account_payment_method_check')
+
+    env.cr.execute("""SELECT column_name
```
StefanRijnhart: Next time, you can use https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.column_exists

StefanRijnhart (approved):

NL66278 (approved):
:+1: LGTM

addons/account_check_printing/migrations/9.0.1.0/post-migration.py
```diff
@@ -20,6 +20,24 @@ def update_payments_from_vouchers(env):
         WHERE ap.id = av.id
         """
     )
+    check_method = env.ref(
+        'account_check_printing.account_payment_method_check')
+
+    if openupgrade.column_exists(cr, 'account_journal', 'allow_check_writing'):
```
NL66278: cr should be env.cr

pedrobaeza: Right, I was waiting Travis confirmation.  I'll change it later.

PR #779, opened 2017-03-20
[9.0][purchase] double validation

JordiBForgeFlow:
In v8 the module 'purchase_double_validation' allows you to change the PO approval workflow and set a limit by which PO's will change to state 'Waiting Approval'.

As of v9 this logic has been integrated into the 'purchase' module.

Current behavior before PR:
If you migrate a DB where module 'purchase_double_validation' is installed, the migrated database assumes that you are not doing a double validation.

If you migrate a DB where module 'purchase_double_validation' is installed, the migrated database assumes that you are doing a double validation, and keeps the same validation threshold.

--

### DIFF
--- a/addons/purchase/migrations/9.0.1.2/pre-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/pre-migration.py
@@ -73,14 +73,8 @@ def pre_create_po_double_validation(cr, env):
     po_double_validation_amount = 5000.0
     if double_validation_transition:
         po_double_validation = 'two_step'
-        condition = False
-        if double_validation_transition.condition.find('>=') >= 0:
-            condition = '>='
-        elif double_validation_transition.condition.find('>=') >= 0:
-            condition = '<'
-        if condition:
-            po_double_validation_amount = (
-                double_validation_transition.condition.split(condition, 1)[1])
+        po_double_validation_amount = (
+            double_validation_transition.condition.split('>=', 1)[1])
     else:
         po_double_validation = 'one_step'
 
--- a/addons/purchase/migrations/9.0.1.2/pre-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/pre-migration.py
@@ -46,7 +46,7 @@ def pre_create_po_double_validation(cr, env):
         cr.execute(
             """
             ALTER TABLE res_company ADD COLUMN po_double_validation
-            boolean;
+            character varying;
             COMMENT ON COLUMN res_company.po_double_validation IS
             'Levels of Approvals';
             """)
--- a/addons/purchase/migrations/9.0.1.2/pre-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/pre-migration.py
@@ -65,7 +65,7 @@ def pre_create_po_double_validation(cr, env):
             'Double validation amount';
             """)
 
-    double_validation_transition = env['ir.model.data'].ref(
+    double_validation_transition = env.ref(
         'purchase_double_validation.trans_confirmed_double_gt')
     po_double_validation_amount = 5000.0
     if double_validation_transition:
--- a/addons/purchase/migrations/9.0.1.2/pre-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/pre-migration.py
@@ -31,7 +31,66 @@ def map_order_state(cr):
         WHERE l.order_id = o.id""")
 
 
-@openupgrade.migrate()
-def migrate(cr, version):
+def pre_create_po_double_validation(cr, env):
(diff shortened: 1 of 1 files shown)


addons/purchase/migrations/9.0.1.2/pre-migration.py
```diff
@@ -31,7 +31,63 @@ def map_order_state(cr):
...
+            po_double_validation_amount = %s
+    """ % (po_double_validation, po_double_validation_amount))
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    cr = env.cr
     openupgrade.copy_columns(cr, column_copies)
     map_order_state(cr)
+    pre_create_po_double_validation(cr, env)
```
pedrobaeza: No need to pass both cr and env.

addons/purchase/migrations/9.0.1.2/pre-migration.py
```diff
@@ -31,7 +31,63 @@ def map_order_state(cr):
...
 
-@openupgrade.migrate()
-def migrate(cr, version):
+def pre_create_po_double_validation(cr, env):
+    """ In v8 when you install the module 'purchase_double_validation'
+    the po workflow is altered to put conditions to change the po status to
+    waiting approval if a condition is met. In v9 this is handled without
+    workflow."""
+
+    cr.execute("""SELECT column_name
```
pedrobaeza: You should use openupgradelib's function to check if a module in installed.

pedrobaeza (changes requested):
You need also to execute update_module_names with merge option to avoid dangling modules at the end of the process.

addons/purchase/migrations/9.0.1.2/pre-migration.py
JordiBForgeFlow: I do not really need to check if the module exists. Just by checking if the workflow step added by the module 'purchase_double_validation' exists is enough.

pedrobaeza: It's only for convenience. It's faster and shorter.

addons/purchase/migrations/9.0.1.2/pre-migration.py
JordiBForgeFlow: Fixed that

addons/purchase/migrations/9.0.1.2/pre-migration.py
JordiBForgeFlow: But I will still need to access to env.ref('purchase_double_validation.trans_confirmed_double_gt') and do everything else. So is it really necessary to check if the module has been installed?

addons/purchase/migrations/9.0.1.2/pre-migration.py
```diff
@@ -31,7 +31,63 @@ def map_order_state(cr):
...
 
-@openupgrade.migrate()
-def migrate(cr, version):
+def pre_create_po_double_validation(env):
+    """ In v8 when you install the module 'purchase_double_validation'
+    the po workflow is altered to put conditions to change the po status to
+    waiting approval if a condition is met. In v9 this is handled without
+    workflow."""
+
+    env.cr.execute("""SELECT column_name
```
pedrobaeza, line 40: You have also a function for seeing if the column exists.

addons/purchase/migrations/9.0.1.2/pre-migration.py
pedrobaeza: OK, it was a suggestion for saving some lines. If not, no matter.

addons/purchase/migrations/9.0.1.2/pre-migration.py
```diff
@@ -31,7 +31,63 @@ def map_order_state(cr):
...
+        po_double_validation_amount = (
+            double_validation_transition.condition.split('>=', 1)[1])
+    else:
+        po_double_validation = 'one_step'
+
+    env.cr.execute("""
+        UPDATE res_company
+        SET po_double_validation = '%s',
+            po_double_validation_amount = %s
+    """ % (po_double_validation, po_double_validation_amount))
```
StefanRijnhart, line 85: Please use the safe variable substitution
```
        SET po_double_validation = %s,
            po_double_validation_amount = %s
    """, (po_double_validation, float(po_double_validation_amount)))
```

pedrobaeza:
@jbeficent are you going to continue with this PR?

pedrobaeza:
No news, so I close. Feel free to reopen if you are going to continue with this.

rousseldenis:
I confirm this.

The most annoying after migration is that configuration view is broken as purchase_double_validation's view has not been deleted

PR #782, opened 2017-03-22
[IMP] Add openerp-client-lib to requirements.txt

ateijelo:
If the environment is created from requirements.txt, this library
wasn't installed and it prevented the OpenUpgrade Records addon
from being installed

Current behavior before PR:
OpenUpgrade Records failed to install.

Creating an environment from requirements.txt allows installation
of OpenUpgrade Records.

--

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,3 +41,4 @@ wsgiref==0.1.2
 XlsxWriter==0.7.7
 xlwt==0.7.5
 openupgradelib>=1.2.0
+openerp-client-lib==1.1.2


pedrobaeza:
openerp-client-lib is not a requirement for OpenUpgrade, but for module openupgrade_records. You should declare in its manifest. the external_dependencies key

ateijelo:
Hi, I set openerp-client-lib version exactly to 1.1.2 as `pip freeze` showed me. Should it be >= or anything different?
Also, how strictly should one adhere to the PR format of Description, Current Behavior, Desired Behavior. Sometimes if feels overkill for PRs as small as this.

pedrobaeza:
Please read again my comment. It's not a question of version, but a question of not requiring something that is not needed for running a regular OpenUpgrade migration procedure. You only need that for getting analysis file.

ateijelo:
Sorry, I wrote my second comment before I saw yours. I understand now, and I see that openupgrade_records' manifest already declares those external dependencies. I'll close the PR.

PR #784, opened 2017-03-28
10.0 doc migration manager

StefanRijnhart:
Clarify the use of the Odoo migration manager and the migration stages.
Re-add makefile from Openupgrade 8.0 to build documentation locally.
Remove the deferred migration step obsoleted by the 'end' migration stage introduced in Odoo 9.0


StefanRijnhart:
@hbrunn could you switch https://doc.therp.nl/openupgrade to version 10?

legalsylvain (approved):
Thanks !

hbrunn:
thanks for the extra explanations. The documentation script pulls the repo's default branch, I think we should keep it like that and just change the default branch, do you agree?

PR #789, opened 2017-04-04
[ADD] base

hbrunn:
if this works, I propose to change the default branch to 10, this way also getting the correct documentation on https://doc.therp.nl/openupgrade


StefanRijnhart (approved):
I would merge this optimistically, but you requested a merge from @pedrobaeza.

StefanRijnhart:
Loading changes in noupdate data here: https://github.com/hbrunn/OpenUpgrade/pull/1

hbrunn:
@StefanRijnhart this is not how I understand the requested reviews feature, so your merge will be just as welcome as @pedrobaeza's to me

StefanRijnhart:
I guess the meaning is in the eyes of the beholder. You meant just to notify @pedrobaeza then?

Anyway, merging now!

odoo/addons/base/migrations/10.0.1.3/openupgrade_analysis_work.txt
```diff
@@ -1,26 +1,26 @@
...
+# sales_team_dashboard => kanban
 
 base         / res.company              / bank_ids (one2many)           : NEW relation: res.partner.bank
 base         / res.company              / sequence (integer)            : NEW
 # NOTHING TO DO
 
 base         / res.currency             / symbol (char)                 : now required
-# TODO: pre-migration: Use the value of name if symbol is blank
+# Use the value of name if symbol is blank
 base         / res.partner              / birthdate (char)              : DEL
```
pedrobaeza, line 24: It's a good practice to rename these fields to `None` in case they are needed by other module. database_cleanup can later purge the column if you want.

PR #790, opened 2017-04-05
[ADD] mail

hbrunn:


addons/mail/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+    # notifications are unread by definition
+    cr.execute(
+        'update mail_message_res_partner_needaction_rel set is_read=False'
+    )
+    # set email properties
+    cr.execute(
+        """update mail_message_res_partner_needaction_rel rel
+        set is_email=True, email_status=case
+            when m.state in ('sent', 'exception') then m.state
+            when m.state='outgoing' then 'ready'
```
StefanRijnhart: Looks like a redundant check, given the next line?

StefanRijnhart:
Loading noupdate data here: https://github.com/hbrunn/OpenUpgrade/pull/2

StefanRijnhart:
What,  no env? https://github.com/hbrunn/OpenUpgrade/pull/3
But really, you might want to refactor to use env everywhere (or specify use_env=False) because we planned for the decorator to return an env by default in version 10.
-- edit: included in the PR --

hbrunn:
I thought we only wanted to default to an env in 11? I'd like to stick to the cr version as soon as it's the default and I only use cursors anyways
If we want to change the default in openupgradelib, let's do that now

StefanRijnhart:
Ah, that explains. Docstring says 'Starting from version 10, this is the default', but the code says > 10.

StefanRijnhart:
Let's do it now https://github.com/OCA/openupgradelib/pull/69/files

StefanRijnhart:
While you are here, could you confirm that this line is redundant? https://github.com/OCA/OpenUpgrade/pull/790#discussion-diff-111407848R24

addons/mail/migrations/10.0.1.0/post-migration.py
hbrunn: I agree, but don't have my ssh keys at hand

odoo/openupgrade/doc/source/modules90-100.rst
```diff
@@ -72,7 +72,7 @@ Status :
 +-----------------------------------+-----------------------------------+
 |base_import_module                 |                                   |
 +-----------------------------------+-----------------------------------+
-|base_setup                         |                                   |
+|base_setup                         | Nothing to do                     |
```
pedrobaeza: Don't change in this PR other modules doc

pedrobaeza (approved):
Curious mechanism that Odoo has used with `_table = 'mail_message_res_partner_needaction_rel'` for keeping the same many2many table, but it makes things simpler.

odoo/openupgrade/doc/source/modules90-100.rst
hbrunn: this was my judgement of the dependencies, but your PR says otherwise

pedrobaeza: Well, it can be barely consider as nothing to do, as this is only have side-effects sometimes, but I prefer to stay safe with the previous known issues. Anyway, my comment was for isolating this kind of changes from PR with the scripts of one module. We can make a PR for these changes, or directly a commit including this documentation (as it doesn't affect Travis status and thus doesn't require even optimistic merging).

addons/mail/migrations/10.0.1.0/post-migration.py
pedrobaeza: Please attend this comment and we are good to go with this PR.

hbrunn: done

PR #792, opened 2017-04-13
10.0 merge upstream Odoo 20170413

StefanRijnhart:
Testing this branch here: https://github.com/OCA/OpenUpgrade/pull/793
-- edit: tests passed --


StefanRijnhart:
Conflict in migration.py solved as follows:
```
--- a/odoo/modules/migration.py
+++ b/odoo/modules/migration.py
@@@ -69,13 -74,11 +74,16 @@@ class MigrationManager(object)
          stageformat = {
              'pre': '[>%s]',
              'post': '[%s>]',
+             'end': '[$%s]',
          }
-         # In openupgrade, remove 'or pkg.installed_version is None'
+         state = pkg.state if stage in ('pre', 'post') else getattr(pkg, 'load_state', None)
+ 
 -        if not (hasattr(pkg, 'update') or state == 'to upgrade') or state == 'to install':
++        # In openupgrade, also run migration scripts upon installation.
 +        # We want to always pass in pre and post migration files and use a new
 +        # argument in the migrate decorator (explained in the docstring)
 +        # to decide if we want to do something if a new module is installed
 +        # during the migration.
-         if not (hasattr(pkg, 'update') or pkg.state == 'to upgrade'):
++        if not (hasattr(pkg, 'update') or state in ('to upgrade', 'to install')):
              return
  
          def convert_version(version):
@@@ -116,12 -116,7 +121,12 @@@
              lst.sort()
              return lst
  
 +        def mergedict(a, b):
 +            a = a.copy()
 +            a.update(b)
 +            return a
 +
-         parsed_installed_version = parse_version(pkg.installed_version or '')
+         parsed_installed_version = parse_version(getattr(pkg, 'load_version', pkg.installed_version) or '')
          current_version = parse_version(convert_version(pkg.data['version']))
  
          versions = _get_migration_versions(pkg)
```

PR #798, opened 2017-04-21
[RFR] Readd new modules and mark the differences in module sets

StefanRijnhart:
Follow up to the discussion from https://github.com/OCA/OpenUpgrade/pull/795 about whether the module coverage list should feature the old modules from the previous release, the new modules from the new release, or both.

image


pedrobaeza:
This joins the best of both worlds, so I see this correct. Maybe now modules that are the same looks not correctly indented in the table. There are 3 possible solutions for this:

* Put the icon in a new column.
* Put a blank image with the same size of the other images.
* Put the new/old image after the module name.

odoo/openupgrade/doc/source/modules90-100.rst
```diff
@@ -14,6 +14,12 @@ Status :
 
 * 'Done' : Analyse has been done & work has been merged into trunk project ;
 
+Modules that are new in this release are marked with |new|. Modules that are
```
pedrobaeza, line 17: Are these icons to be correctly rendered as the definition is after?

StefanRijnhart: Yes, no problem.

StefanRijnhart:
The placement is a matter of taste. None of the solutions are pretty. I tried adding the icon after the name, but I liked this better. This looks better when you look at the list in a browser tha it looks here in this screenshot, actually.

pedrobaeza:
OK, if you prefer this way, it's not the most important thing.

pedrobaeza:
There's still the question of renamed modules. I think here we have _hr_equipment_ > _maintenance_.

hbrunn:
for the renamed modules, I'd suggest a different PR. Merging this one

PR #800, opened 2017-05-01
Fix for older versions of openupgrade where default isn't env

gfcapalbo:

explictly writing use_env for those instances still running openupgradelib without  https://github.com/OCA/openupgradelib/pull/69

### DIFF
--- a/addons/base_setup/migrations/10.0.1.0/pre-migration.py
+++ b/addons/base_setup/migrations/10.0.1.0/pre-migration.py
@@ -10,6 +10,6 @@ _xmlid_renames = [
 ]
 
 
-@openupgrade.migrate()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You need anyway commit https://github.com/OCA/openupgradelib/commit/e937345daf2df8c737d88ed1644308cbb9136104, that it's introduced later, so I don't see too much sense to not having latest openupgradelib.

gfcapalbo:
makes sense. closing this.

pedrobaeza:
I have created anyway a new release for latest version to being directly accessible via `pip install openupgradelib`

pedrobaeza:
https://github.com/OCA/openupgradelib/releases/tag/1.3.0

PR #802, opened 2017-05-02
Add migration for analytic

gfcapalbo:

Current behavior before PR:

--


addons/analytic/migrations/10.0.1.1/openupgrade_analysis.txt
```diff
@@ -7,3 +7,4 @@ DEL ir.actions.act_window: analytic.action_account_analytic_chart
 NEW ir.ui.view: analytic.account_analytic_tag_form_view
 NEW ir.ui.view: analytic.account_analytic_tag_tree_view
 DEL ir.ui.view: analytic.account_analytic_chart_view
+
```
pedrobaeza: Why this extra line?

addons/analytic/migrations/10.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+@openupgrade.migrate(use_env=False)
```
pedrobaeza: Don't change the default `use_env=True`, and make your calls with `env.cr` instead of `cr`, because if the environment is later needed, then we need to change all calls, as it happened in https://github.com/OCA/OpenUpgrade/pull/803/files#diff-cd4cccdd111034d7430f83844f9d6372

addons/analytic/migrations/10.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,17 @@
...
+def migrate(cr, version):
+    # copy columns good practice in pre-mig script format of colum_spec should
+    # be: { Table_name_in_db : [( old_column_name , new_column_name, type)]
+    # defaults of new columname are fetchable via method get_legacy_name 
+    openupgrade.copy_columns( cr, {
```
pedrobaeza: Remove the leading space

addons/analytic/migrations/10.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+@openupgrade.migrate(use_env=False)
+def migrate(cr, version):
+    # copy columns good practice in pre-mig script format of colum_spec should
```
pedrobaeza: Better to just rename the column to None. You don't need to copy values.

addons/analytic/migrations/10.0.1.1/tests/analytic_data_v10.yml
```diff
@@ -0,0 +1,19 @@
+-
```
pedrobaeza, line 1: I don't see the test that checks that these accounts are inactive.

pedrobaeza (changes requested):

addons/analytic/migrations/10.0.1.1/tests/analytic_data_v10.yml
gfcapalbo: https://github.com/OCA/OpenUpgrade/pull/802/commits/ba927fef16709a1171c1d00b13bcc0a4e8e89a25#diff-21238dc2fbb1a7124cb1ad70747c974eR5

addons/analytic/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+    openupgrade.map_values(
+        cr, openupgrade.get_legacy_name('account_type'),
+        'active',  [('closed' ,  False) , ('normal', True)],
+        table='account_analytic_account',
+        write='sql'
+    )
+
+@openupgrade.migrate(use_env=True)
+def migrate(cr, version):
+    cr = env.cr
```
hbrunn: https://travis-ci.org/OCA/OpenUpgrade/builds/229092256#L745

hbrunn (changes requested):

hbrunn (approved):

pedrobaeza (approved):
The error is due to crm module

addons/analytic/migrations/10.0.1.1/tests/test_analytic.py
```diff
@@ -0,0 +1,13 @@
+# coding: utf-8
+from openerp.tests.common import TransactionCase
+
+
+class TestAnalytic(TransactionCase)
```
pedrobaeza, line 5: I have just discovered that this is not being executed, or this would be marked as error, as it misses the final `:`

cc @hbrunn @gfcapalbo

pedrobaeza: I'm fixing the issues directly for getting a green build: https://travis-ci.org/OCA/OpenUpgrade/builds/232892182#L357

pedrobaeza: Partially fixed on https://github.com/OCA/OpenUpgrade/commit/993e0ecc8028aa20b8cd9e52b774732d15d59ef3. Switching to PR for trying to fix the other error: https://travis-ci.org/OCA/OpenUpgrade/builds/232906089#L428

PR #805, opened 2017-05-05
[9.0][FIX] Change the query that links the account move lines with the pay…

JordiBForgeFlow:
The prior approach was adding to the move lines of the invoice that was being paid the payment_id (Originator payment). This does not make sense. 

The field "Originator payment" in account move lines is intended to be used as the link that relates the move lines to the payment that originated them.

See: 
https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_payment.py#L511
https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_payment.py#L521

--


JordiBForgeFlow:
cc @NL66278 @pedrobaeza @hbrunn This is an important fix.

NL66278 (approved):
:+1: LGTM

PR #809, opened 2017-05-09
[DEL] no placeholders, files will be merged from older branches by

### LINKED ISSUE #807: Published documentation not correct
If you go to https://doc.therp.nl/openupgrade/modules80-90.html, there isn't any list. This should be due to an incorrect linking of all pages.

hbrunn:
build_openupgrade_docs
Fixes #807

@StefanRijnhart what was your rationale for 283ea545d15351b4aa33b84e5315a7f2a0e179dc? The build script copies the folder from every branch in ascending order, so the last version of some file wins. We need this to be able to update the documentation, but then we can't overwrite module coverage files because we need the latest version from the branch it originates from: https://github.com/OCA/OpenUpgrade/blob/10.0/scripts/build_openupgrade_docs#L26

### DIFF
--- a/odoo/openupgrade/doc/source/modules50-60.rst
+++ b/odoo/openupgrade/doc/source/modules50-60.rst
@@ -1,4 +1,0 @@
-Module coverage 5.0 -> 6.0
-==========================
-
-This is a placeholder for the aggregated documentation published at https://doc.therp.nl/openupgrade/modules50-60.html
--- a/odoo/openupgrade/doc/source/modules60-61.rst
+++ b/odoo/openupgrade/doc/source/modules60-61.rst
@@ -1,4 +1,0 @@
-Module coverage 6.0 -> 6.1
-==========================
-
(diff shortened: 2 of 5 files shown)


pedrobaeza (approved):
This closes #807

StefanRijnhart:
@hbrunn I'm sorry, I don't understand. I only replaced the contents of the files, didn't I? I did that so that a locally built documentation would not show an outdated module coverage for the older releases.

Are you saying that for example modules61-70 is only meant to exist in openupgrade/7.0? Because it exists in 8.0 and in 9.0 as well.

StefanRijnhart:
Naive suggestion to use wget to replace current logic: https://github.com/OCA/OpenUpgrade/pull/810

hbrunn:
we introduced this with 9, then we indeed had all files everywhere, but the idea was indeed to phase that out with upcoming versions. But I think your approach is better anyways, as soon as we agree on the implementation details. Will close this then.

PR #810, opened 2017-05-09
[RFR] Use wget instead of git checkouts

StefanRijnhart:
for more reliable fetching of older module coverage.

### DIFF
--- a/scripts/build_openupgrade_docs
+++ b/scripts/build_openupgrade_docs
@@ -1,36 +1,25 @@
 #!/bin/sh
 
-DOC_BUILD_BASE_DIR=/var/tmp/openupgrade
-DOC_PARTS="6.0|bin/openupgrade/doc
-6.1|openerp/openupgrade/doc
-7.0|openerp/openupgrade/doc
-8.0|openerp/openupgrade/doc
-9.0|openerp/openupgrade/doc
-10.0|odoo/openupgrade/doc"
-OUTPUT_DIR=${OUTPUT_DIR:-$DOC_BUILD_BASE_DIR/doc}
-DOC_BUILD_DIR=$(mktemp -d)
-REPO_URL=https://github.com/OCA/OpenUpgrade.git
-OPENUPGRADELIB=https://github.com/OCA/openupgradelib.git
+DOC_BUILD_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../odoo/openupgrade/doc && pwd )"
+BASE_URL=https://github.com/OCA/OpenUpgrade/raw/
+DOC_PARTS="7.0/openerp/openupgrade/doc/source/modules50-60.rst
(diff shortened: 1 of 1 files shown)


scripts/build_openupgrade_docs
```diff
@@ -1,36 +1,25 @@
...
-6.1|openerp/openupgrade/doc
-7.0|openerp/openupgrade/doc
-8.0|openerp/openupgrade/doc
-9.0|openerp/openupgrade/doc
-10.0|odoo/openupgrade/doc"
-OUTPUT_DIR=${OUTPUT_DIR:-$DOC_BUILD_BASE_DIR/doc}
-DOC_BUILD_DIR=$(mktemp -d)
-REPO_URL=https://github.com/OCA/OpenUpgrade.git
-OPENUPGRADELIB=https://github.com/OCA/openupgradelib.git
+DOC_BUILD_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../odoo/openupgrade/doc && pwd )"
```
hbrunn: this combines badly with the shebang above. Better say `$(realpath $(dirname $0)/../odoo/openupgrade/doc)`, this should run on basically any shell and only needs coreutils installed. It's a weird system if that's not the case.

But I don't really see the point of keeping the build dir, what's your rationale here?

hbrunn: if you really want it like that, make this the default, so that I can override it from the outside for the build on Therp's servers

hbrunn (changes requested):
what I miss here is cloning the latest version and updating it. The beauty of the previous version was that you could just run it and it will take care of everything, I think that's not possible now?

But yes, that's much better than always cloning everything.

scripts/build_openupgrade_docs
StefanRijnhart: Fixed the bashism (thanks!) and made it a default value.

I think that this way it's easiest to build the full documentation locally, without having to go through another checkout.

hbrunn: but the way the script looks currently, there's no checkout of openupgrade at all, or am i blind?

StefanRijnhart:
Just 'git reset --hard; git clean -f; git pull' would update it to the latest version. Could you live with creating a wrapper that does just this (and checkout the repo in the first place if it does not yet exist)?

scripts/build_openupgrade_docs
StefanRijnhart: True. Because this file is included in the openupgrade distribution, it is not unreasonable to assume that we are currently in a checkout already.

hbrunn (approved):
I was hoping for a drop in replacement, but I do already live with a lot of things I don't like, so one more doesn't matter too much ;-)

StefanRijnhart:
Oh dear, sorry to add to your agony ;-)

pedrobaeza:
Merging this for having correct documentation.

PR #811, opened 2017-05-10
[MIG] account_auto_fy_sequence module: update prefix sequences

cubells:
It is needed if module account_auto_fy_sequence is installed.

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -174,6 +174,17 @@ def map_account_tax_template_type(cr):
         table='account_tax_template', write='sql')
 
 
+def upgrade_old_sequences_format(cr):
+    """ Maybe old sequences prefix have format: %(fy)s and new format is 
+    %(range_year)s instead. We have to update format in all sequences. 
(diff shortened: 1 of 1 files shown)


addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -174,6 +174,17 @@ def map_account_tax_template_type(cr):
...
 
 
+def upgrade_old_sequences_format(cr):
+    """ Maybe old sequences prefix have format: %(fy)s and new format is 
+    %(range_year)s instead. We have to update format in all sequences. 
+    Especially if module account_auto_fy_sequence is installed."""
+    openupgrade.logged_query(cr, """
+        UPDATE ir_sequence 
+        SET prefix = replace(prefix, '(fy)', '(range_year)') 
+        WHERE prefix LIKE '%(fy)%';
```
pedrobaeza, line 184: `%` is a wildcard in PG, so you have to mask them

cubells:
I close this pr because it's already implemented.

pedrobaeza:
For the records, it was done here: https://github.com/OCA/OpenUpgrade/commit/b19f359b82b3da506e0ce16544b4ff395f44262f

Issue #812, opened 2017-05-10
Issues migrating hr_evaluation 7.0->8.0


bwrsandman:
# Description
Duplicate of #214

# Impacted versions
7.0 -> 8.0

# Steps to reproduce
* Install `hr_evaluation` on a vanilla 7.0 database
* In Human Resources Create an Interview Request only filling `survey_id`

Issue is not present with only survey module installed

# Current behavior

Upgrade fails with:
```
2017-05-10 16:49:12,943 11081 CRITICAL survey7 openerp.service.server: Failed to initialize database `survey7`.
Traceback (most recent call last):
  File "/home/odoo/openupgrade/openerp/service/server.py", line 929, in preload_registries
    registry = RegistryManager.new(dbname, update_module=update_module)
  File "/home/odoo/openupgrade/openerp/modules/registry.py", line 370, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/home/odoo/openupgrade/openerp/modules/loading.py", line 420, in load_modules
    force, status, report, loaded_modules, update_module, upg_registry)
  File "/home/odoo/openupgrade/openerp/modules/loading.py", line 312, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks, upg_registry=upg_registry)
  File "/home/odoo/openupgrade/openerp/modules/loading.py", line 179, in load_module_graph
    init_module_models(cr, package.name, models)
  File "/home/odoo/openupgrade/openerp/modules/module.py", line 293, in init_module_models
    obj._auto_end(cr, {'module': module_name})
  File "/home/odoo/openupgrade/openerp/api.py", line 268, in wrapper
    return old_api(self, *args, **kwargs)
  File "/home/odoo/openupgrade/openerp/models.py", line 2751, in _auto_end
    cr.execute('ALTER TABLE "%s" ADD FOREIGN KEY ("%s") REFERENCES "%s" ON DELETE %s' % (t, k, r, d))
  File "/home/odoo/openupgrade/openerp/sql_db.py", line 158, in wrapper
    return f(self, *args, **kwargs)
  File "/home/odoo/openupgrade/openerp/sql_db.py", line 234, in execute
    res = self._obj.execute(query, params)
IntegrityError: insert or update on table "hr_evaluation_interview" violates foreign key constraint "hr_evaluation_interview_request_id_fkey"
[...]

PR #813, opened 2017-05-10
[8.0] Migrate hr_evaluation_interview

bwrsandman:
Fix critical due to null request_id
Set temporary request_id in pre migration
Set request_id to hr_evaluation.appraisal_form in post-migration

Fixes: #812 

--


addons/hr_evaluation/migrations/8.0.0.1/pre-migration.py
```diff
@@ -0,0 +1,44 @@
...
+    for interview_id in cr.fetchall():
+        ret = user_input_obj.create(cr, SUPERUSER_ID, {
+            'survey_id': 1,
+            'type': 'link',
+        }, context={})
+        execute(
+            cr,
+            "UPDATE hr_evaluation_interview\n"
+            "SET request_id = %d WHERE id = %d\n" %
+            (ret, interview_id[0])
```
pedrobaeza, line 42: Please avoid SQL injections here passing values as arguments.

pedrobaeza (changes requested):

addons/hr_evaluation/migrations/8.0.0.1/pre-migration.py
bwrsandman: It has been a while... is this the proper syntax?
```
         execute(
            cr,
            "UPDATE hr_evaluation_interview\n"
            "SET request_id = %d WHERE id = %d\n",
            (ret, interview_id[0])
```

pedrobaeza: Yeah, that's it!

pedrobaeza:
Closing as old one.

PR #817, opened 2017-05-12
[9.0][account_voucher] Fixes migration of memo field in payments from vouchers

JordiBForgeFlow:
in 8.0 the field 'memo' of the voucher is stored in the field 'name'. In v9 the field 'memo' is stored in the field 'communication'. Prior to this the memos in payments in v9 showed the content in field 'comment' of the voucher, which often contained something meaningless like 'write-off'.


pedrobaeza:
Why do you take now first name instead of comment?

pedrobaeza:
OK with the explanation, mergin.

PR #819, opened 2017-05-15
[10.0][product]

JordiBForgeFlow:
Adds migration scripts for the 'product' module.

--


pedrobaeza (changes requested):
Preserve dropped fields (renaming them to None) if someone is using them on other modules:

* 'sequence' on product.category
* 'state' on product.template

JordiBForgeFlow:
@pedrobaeza Thanks. As a general rule in all migrations we I will always preserve the fields that are dropped. This has certainly proven to be extremely useful.

JordiBForgeFlow:
@pedrobaeza This is now done.

NL66278 (approved):
:+1:  Tested this in database with demo data and worked

pedrobaeza (approved):

PR #824, opened 2017-05-16
[ADD] web_tour: Nothing to do

pedrobaeza:
* Updated documentation
* Analysis work file


NL66278:
Just a question, is there not, albeit a remote, change that there are tours existing without name?

pedrobaeza:
This one is a new module, so all the data will enter in the module installation, and thus that can't be possible.

NL66278 (approved):
:+1: LGTM

PR #828, opened 2017-05-16
[ADD] crm: Migration scripts

pedrobaeza:
* Updated documentation
* Analysis work file
* Pre and post migration scripts
* Tests


pedrobaeza:
@hbrunn can you help me to determine the root of the problem?

hbrunn:
@pedrobaeza I'd hope so, but I feel quite clueless myself currently. Any chance the reference doesn't return what you expect it to return? (a missing id or the like?) Next idea would be to check if the registry gets botched somewhere on the way, did you inspect `self.env['crm.lead']._fields['team_id']` if members like `_setup_done` contain what they should? Chances are we simply run our tests too early during initialization.

StefanRijnhart:
@pedrobaeza I took the liberty to commit what should be the fix of your test https://github.com/OCA/OpenUpgrade/pull/828/commits/0941fbc11662c2cde1d15a5f1ed5106c7b1492a0

pedrobaeza:
Travis is still red but it's due to sale_stock dependency without merging stock one. I merge and fix finally on stock migration.

@StefanRijnhart, sorry for letting this happens. I'm fixing it ASAP.

StefanRijnhart:
@pedrobaeza Thank you for understanding, and thank you and the other developers for the great work that you are currently doing!

PR #829, opened 2017-05-16
[FIX] analytic: Test data

pedrobaeza:
YAML data seems to be not correct


pedrobaeza:
OK, now we have the original error, that should be fixed with crm module. I self-merge this for fixing (almost) the branch.

PR #830, opened 2017-05-17
[MIG][9.0] gamification: add changes in noupdate data

cubells:
- Add noupdate_changes.xml file


pedrobaeza:
Include analysis work file

pedrobaeza (approved):

addons/gamification/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,28 @@
...
+---XML records in module 'gamification'---
+DEL email.template: gamification.email_template_badge_received
+DEL email.template: gamification.email_template_goal_reminder
+DEL email.template: gamification.simple_report_template
+DEL ir.module.category: gamification.module_goal_category
+NEW mail.template: gamification.email_template_badge_received
+NEW mail.template: gamification.email_template_goal_reminder
+NEW mail.template: gamification.simple_report_template
+NEW res.groups: base.group_hr_manager
+DEL res.groups: gamification.group_goal_manager
```
pedrobaeza, line 27: Although not the case, next time check when there are DEL XML records, if they are noupdate, as in this case, records are not automatically purged, and put explicitly that on the work file to know that you have take that into account.

PR #831, opened 2017-05-17
[MIG][9.0] fleet: add migration scripts

cubells:


pedrobaeza (approved):
There isn't noupdate records also, so :+1:

PR #832, opened 2017-05-17
[MIG][9.0] gamification_sale_crm: add migration scripts

cubells:
- add changes file in noupdate data


addons/gamification_sale_crm/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,4 @@
+---Fields in module 'gamification_sale_crm'---
+---XML records in module 'gamification_sale_crm'---
+DEL gamification.goal.definition: gamification_sale_crm.definition_crm_nbr_call
```
pedrobaeza, line 3: This is a noupdate record, and it depends on a model that doesn't exist now (crm.phonecall), so you need to remove it on the script.

pedrobaeza (changes requested):
One comment

cubells:
I think it's done

pedrobaeza (approved):

PR #837, opened 2017-05-19
[10.0][Stock] Added migration scripts for stock module

pepetreshere:
Description of the issue/feature this PR addresses: Added migration scripts to v10 for stock module

Current behavior before PR: Module stock fail migration

Desired behavior after PR is merged: Module stock migrated successfully

--


lfreeke:
@jbeficent can you have a look at this?

addons/stock/migrations/10.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -3,19 +3,22 @@ stock        / barcode.rule             / type (False)                  : NEW se
 # NOTHING TO DO
 
 stock        / procurement.rule         / picking_type_id (many2one)    : now required
-# TODO: pre-migration. Default to some value, guessing by looking at source
+# TODO: post-migration. Default to some value, guessing by looking at source
 # and destination locations, and warehouse.
+# Done!
```
pedrobaeza: Remove the TODO and put `Done: explanation of what you have done`

addons/stock/migrations/10.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# © 2017 Trescloud <http://trescloud.com>
```
pedrobaeza: Use `copyright` full word

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+# © 2017 Trescloud <http://trescloud.com>
```
pedrobaeza: Use `copyright` full word

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
```
pedrobaeza: Put this at the end of the file by general convention

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+    update_picking_type_id(env)
+    update_ordered_qty(env)
+    populate_stock_scrap(env)
+    
+def update_picking_type_id(env):
+    '''
```
pedrobaeza: According PEP257, method documentation goes with triple double quota, not simple, and first line start in the same one as the initial quotas.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+    Updates picking_type_id, defaulting to some value, guessing by looking at source
+    and destination locations, and warehouse.  
+    :param env: enviroment variable (self)
+    '''
+
```
pedrobaeza: Don't let blank lines inside a method. They're useless and are less readable.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+    procurement_rules_to_set = env['procurement.rule'].search([])
+    
+    for procurement_rule in procurement_rules_to_set:
+        if not procurement_rule.picking_type_id:
+            
+            picking_type_id = False
+            env.cr.execute(
+                """
+                SELECT id from stock_picking_type 
+                WHERE warehouse_id = %s AND default_location_dest_id = %s AND default_location_src_id = %s""" % (
```
pedrobaeza: Don't overpass 80 chars limit

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+            if not picking_type_id: #fallback values when everything else fails
+                if procurement_rule.action == 'buy':
+                    #location_src_id is not considered as is not mandatory in this case
+                    xml_record = env.ref("stock.picking_type_in")
+                    if xml_record:
+                        picking_type_id = xml_record.id
+                elif procurement_rule.action == 'move':
+                    if procurement_rule.location_id and procurement_rule.location_id.usage == 'customer':
+                        #special case when is a delivarable to a customer location
+                        xml_record = env.ref("stock.picking_type_out")
```
pedrobaeza: Pre-load this references. Don't load them on each loop.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+def update_picking_type_id(env):
+    '''
+    Updates picking_type_id, defaulting to some value, guessing by looking at source
+    and destination locations, and warehouse.  
+    :param env: enviroment variable (self)
+    '''
+    
+    procurement_rules_to_set = env['procurement.rule'].search([])
+    
+    for procurement_rule in procurement_rules_to_set:
```
pedrobaeza: You can use ORM API here instead of dealing with the ugly SQL.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+                        picking_type_id = xml_record.id
+        
+            procurement_rule.write({'picking_type_id': picking_type_id})
+
+def update_ordered_qty(env):
+    '''
+    Set as default the value of field ordered_qty from the value of field product_uom_qty  
+    :param env: enviroment variable (self)
+    '''
+    #stock        / stock.move               / ordered_qty (float)           : NEW
```
pedrobaeza: Don't be too verbose. This line can be removed.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+    # TODO: post-migration. Default to product_uom_qty
+    env.cr.execute(
+        """
```
pedrobaeza: Use one quote when the query is very simple.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+    
+    #stock        / stock.pack.operation     / ordered_qty (float)           : NEW
+    # TODO: post-migration. Default to product_uom_qty
+    env.cr.execute(
+        """
+        UPDATE stock_pack_operation SET ordered_qty = product_qty
+        """)
+    
+def populate_stock_scrap(env):
+    '''
```
pedrobaeza: You have forgotten to put Done on the work analysis for this.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        SELECT id from stock_location 
+        WHERE scrap_location is True
+        """
+        )
+    scrap_location_ids = env.cr.fetchone()
+    
+    #do not call stock_scrap.create as it will create a duplicated stock move, use SQL instead
+    #field package_id not set as no value is defined    
+    env.cr.execute(
```
pedrobaeza: Again, using ORM can be simple and the same in speed (more or less).

addons/stock/migrations/10.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,32 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    
+    #field route_sequence
+    openupgrade.copy_columns(
+        env.cr, {
```
pedrobaeza: Use variables to store the list of column renames, copies, and so on, the same as in https://github.com/OCA/OpenUpgrade/pull/839/files#diff-3cdc4348ce357d41a6908e42cadbca5aR8

Group also all the copies in one.

pedrobaeza (changes requested):
Good work being your first submission to OpenUpgrade. Please follow my remarks.

pepetreshere:
Thanks for your remarks @pedrobaeza!, we will work on this by next monday

addons/stock/migrations/10.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -69,8 +68,7 @@ stock        / stock.scrap              / product_uom_id (many2one)     : NEW re
 stock        / stock.scrap              / scrap_location_id (many2one)  : NEW relation: stock.location
 stock        / stock.scrap              / scrap_qty (float)             : NEW required: required, req_default: function
 stock        / stock.scrap              / state (selection)             : NEW selection_keys: ['done', 'draft']
-# TODO: Create one record for each move that had as destionation a scrap
-# location
+# Done: pre-migration: Create one record for each move that had as destionation a scrap location
```
MiquelRForgeFlow: is post-migration

addons/stock/migrations/10.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -29,17 +29,16 @@ stock        / stock.inventory.line     / state (char)                  : type i
 # NOTHING TO DO
 
 stock        / stock.location.path      / auto (selection)              : selection_keys is now '['manual', 'transparent']' ('['auto', 'manual', 'transparent']')
-# TODO: pre-migration. copy column
-# TODO: post-migration. Move 'auto' to some other value
+# Done: pre-migration. copy column and then replace value "auto" with value "manual"
```
MiquelRForgeFlow: the replacement should be done in post-migration

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+def update_picking_type_id(env):
+    """Updates picking_type_id, defaulting to some value, guessing by looking at source
+    and destination locations, and warehouse.  
+    :param env: enviroment variable (self)
+    """
+    #load xml data to be used for filling in missing info
+    xml_stock_picking_type_internal = env.ref("stock.picking_type_internal")
+    xml_stock_picking_type_out = env.ref("stock.picking_type_out")
+    xml_stock_picking_type_in = env.ref("stock.picking_type_in")
+    xml_stock_picking_type_manufacturing = env.ref("mrp.picking_type_manufacturing")
```
MiquelRForgeFlow: this should be removed. the 'manufacture' case should be into the migration of mrp module

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+    #verify for each procument rule to set
+    procurement_rules_to_set = env['procurement.rule'].search([])
+    for procurement_rule in procurement_rules_to_set:
+        if not procurement_rule.picking_type_id:
+            picking_type_id = False
+            env.cr.execute(
+                '''
```
MiquelRForgeFlow: ''' → """

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+                SELECT id from stock_picking_type 
+                WHERE warehouse_id = %s AND 
+                      default_location_dest_id = %s AND 
+                      default_location_src_id = %s''' % (
+                    procurement_rule.warehouse_id, 
+                    procurement_rule.location_id,
+                    procurement_rule.location_src_id,
+                )
+            )
+            picking_type_ids = cr.fetchone()
```
MiquelRForgeFlow: → <code>env.cr.fetchone()</code>

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+            if picking_type_ids:
+                picking_type_id = picking_type_ids[0]
+            if not picking_type_id: #fallback values when everything else fails
```
MiquelRForgeFlow: comments should be in a separated line, and also there should be a space between <code>#</code> and the words

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+    """Updates picking_type_id, defaulting to some value, guessing by looking at source
+    and destination locations, and warehouse.  
+    :param env: enviroment variable (self)
+    """
+    #load xml data to be used for filling in missing info
+    xml_stock_picking_type_internal = env.ref("stock.picking_type_internal")
+    xml_stock_picking_type_out = env.ref("stock.picking_type_out")
+    xml_stock_picking_type_in = env.ref("stock.picking_type_in")
+    xml_stock_picking_type_manufacturing = env.ref("mrp.picking_type_manufacturing")
+    #verify for each procument rule to set
```
MiquelRForgeFlow: → procurement

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+
+def update_ordered_qty(env):
+    """ Set the value of new field ordered_qty as:
+      - for stock moves the value of field product_uom_qty
+      - for stock_pack_operation the value of product_qty  
+    :param env: enviroment variable (self)
+    """
+    env.cr.execute(
+        '''
+        UPDATE stock_move SET ordered_qty = product_uom_qty
```
MiquelRForgeFlow: the query should be compacted in one line

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    env.cr.execute(
+        '''
+        SELECT id from stock_location 
+        WHERE scrap_location is True
+        '''
+        )
+    scrap_location_ids = env.cr.fetchone()
+    #do not call stock_scrap.create as it will create a duplicated stock move, use SQL instead
+    #field package_id not set as no value is defined
```
MiquelRForgeFlow: field <code>package_id</code> can be set, with a witty query, if you relate the move with the stock_quant

addons/stock/migrations/10.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+        ('route_sequence', None, None),
+        ],
+    }
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):    
+    openupgrade.copy_columns(env.cr, column_copies)
+    openupgrade.float_to_integer(env.cr, 'stock_location_path', 'route_sequence')
+    openupgrade.float_to_integer(env.cr, 'procurement_rule', 'route_sequence')
+    env.cr.execute(
```
MiquelRForgeFlow: this replacement should be done in the post-migration

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+                        if xml_stock_picking_type_out:
+                            picking_type_id = xml_stock_picking_type_out.id
+                    else:
+                        if xml_stock_picking_type_internal:
+                            picking_type_id = xml_stock_picking_type_internal.id
+                elif procurement_rule.action == 'manufacture':
+                    if xml_stock_picking_type_manufacturing:
+                        picking_type_id = xml_stock_picking_type_manufacturing.id
+            procurement_rule.write({'picking_type_id': picking_type_id})
+
```
MiquelRForgeFlow: between the defs should be two empty lines

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,94 @@
...
+            SELECT date_expected,location_id,restrict_lot_id,id,name,origin,restrict_partner_id,picking_id,product_id,product_uom,location_dest_id,product_uom_qty,state
+            FROM stock_move
+            WHERE location_id IN %s AND product_uom_qty < 0.0 AND state = 'done' 
+                OR location_dest_id IN %s AND product_uom_qty >= 0.0 AND state = 'done'
+        ''',(scrap_location_ids,scrap_location_ids))
+    
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    update_picking_type_id(env)
+    update_ordered_qty(env)
```
MiquelRForgeFlow: pass the <code>cr</code> instead of the <code>env</code> if the def only use the <code>cr</code>

MiquelRForgeFlow (changes requested):
I just sent you a PR (https://github.com/TRESCLOUD/OpenUpgrade/pull/2) that includes all adjustments I requested here

pepetreshere:
Thanks @pedrobaeza and @mreficent for your remarks, I am still doing some tests today and I will likely push a few more commits.

pepetreshere:
@mreficent I have included some minor changes for the migration of stock module, and I think we can merge it now. 

I have included again the case for mrp.picking_type_manufacturing with a None option if it is missing (line 24 of post-migration.py), I believe it might be much easier to migrate that field in stock module rather than in mrp module (this change is not even detected in the analysis file of mrp). If you insist I can remove it again, what do you think?. Thanks in advance for your advise.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,131 @@
...
+        table='stock_location_path', write='sql')
+
+
+def update_picking_type_id(env):
+    """Updates picking_type_id, defaulting to some value, guessing by looking
+    at source and destination locations, and warehouse.  
+    :param env: environment variable (self)
+    """
+    # load xml data to be used for filling in missing info
+    xml_stock_picking_type_int = env.ref("stock.picking_type_internal",False)
```
MiquelRForgeFlow: put a space between the comma and the False, and also remember to not overpass 80 chars limit by line in all the code

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -0,0 +1,131 @@
...
+    openupgrade.map_values(
+        cr,
+        openupgrade.get_legacy_name('auto'), 'auto',
+        [('auto', 'manual')],
+        table='stock_location_path', write='sql')
+
+
+def update_picking_type_id(env):
+    """Updates picking_type_id, defaulting to some value, guessing by looking
+    at source and destination locations, and warehouse.
```
MiquelRForgeFlow: flake8 wants the whitespaces at the end of the line to be removed. there are 9 lines affected.

MiquelRForgeFlow (changes requested):

pepetreshere:
@pedrobaeza  thanks for your fixes, this error appears again in travis:
`2017-05-31 14:43:01,815 40924 ERROR openupgrade OpenUpgrade: External ID not found in the system: mrp.picking_type_manufacturing`

I think that to avoid this missing ID we should keep the the False option in the method [env.ref method](https://github.com/OCA/OpenUpgrade/pull/837/commits/cc65573525df09124d8afae699eefcf58d6fde87), or alternatively move this part to the migration script of mrp (which is not advisable from my point of view as detailed above).

So I have included a solution in [this PR](https://github.com/TRESCLOUD/OpenUpgrade/pull/3), please comment if you agree with the resolution as I don't want to overwrite one line of your last commit without your agreement.

pedrobaeza:
Oh, I was working on that and other missing things right now. Well, I remove that part and continue.

pepetreshere:
Hello @pedrobaeza, just to make sure I understood correctly. I understood that you will make the fix altogether with other fixes, so I have not merged my fix into this branch. If you like me to merge [my fix ](https://github.com/TRESCLOUD/OpenUpgrade/pull/3/files) please let me know.

pedrobaeza:
Yeah, I'm making them, but today and tomorrow there is the Odoo days in Spain (https://odoospain.odoo.com/event/jornadas-odoo-2017-2017-06-01-2017-06-02-5), so I'm quite busy. I will try to work tomorrow or on Monday.

pedrobaeza:
@StefanRijnhart the same behavior is happening again with Travis (infinite loop), so the reason should be another one. I'll check it later.

StefanRijnhart:
@pedrobaeza I see, thanks for the heads up. Why is the generic coa set to uninstallable?

pedrobaeza:
That was another try before throwing this error because it tries to create some invoices without account_id. I see that this is not the ideal solution anyway. I'll make other tries later.

hbrunn:
@pepetreshere please rebase your branch on current 10.0, then we should get a better picture from travis

pedrobaeza:
@hbrunn I'm the one that is handling with this PR for having a green branch, but you are decided to put me in troubles :stuck_out_tongue:

hbrunn:
sorry @pedrobaeza, I wasn't aware I'm interfering with your work here. Will cease operations for today then

pedrobaeza:
It was a joke, Holger. Most of the PRs shouldn't be a problem, but it's better to wait to have a green branch before continuing merging more possible problems (for example, we merged sale_stock before having stock module and thus we can't get a green branch until get this one correct).

hbrunn:
looks like stock's migration succeeds now and https://travis-ci.org/OCA/OpenUpgrade/builds/240091842#L4133 is the problem. Should we add a [force deps](https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/modules/graph.py#L88) config entry to make `base_action_rule` depend on all the modules the demo data refers to? Then we should be rid of this problem too.

pepetreshere:
Hello @hbrunn, done with the rebase of the branch, let me know if anything else is needed.

pepetreshere:
Hello @pedrobaeza and @hbrunn , I have tried to rebase but I find everything is up to date as shown in below image. In order to not block your progress I have added you guys as collaborators with write access to this branch (hope you don't mind) as I need your support to make this rebase.

capture 2017-06-02 at 18 01 58

hbrunn:
@pepetreshere thanks for the invitation, but @pedrobaeza already did it, that's why you don't see a difference. Now if we just fix https://github.com/OCA/OpenUpgrade/pull/837#issuecomment-306588914 too, this will be green, the whole branch will be green and we'll be super happy

pedrobaeza:
I'm on it, so I expect to get it through the day.

pedrobaeza:
FYI, all the PR closing/opening activity is because Travis don't set correctly $TRAVIS_COMMIT variable on rebase or forced push operations, so the test is not correct. This is the way I have found for triggering a correct test.

MiquelRForgeFlow (approved):
It's green.

pedrobaeza:
And finally we have it!! :tada: :tada: :tada:

@hbrunn what I have done instead of force-deps, which requires to search for all the declared action rules, is to simply not load non existing models. Done in 9314ff87188da31d1263b6c3bf7bb4f3d2485ac0.

@StefanRijnhart I had to set as not installable l10n_generic_coa and l10n_us, or an error comes which says that the account_id of an invoice can't be null. Maybe it's because account is not yet migrated? We can activate them on the PR to see, but for now this is the only solution I see for having a green branch.

Please give your final blessings them and we merge this for fixing the branch.

pedrobaeza (approved):

StefanRijnhart:
@pedrobaeza thanks for getting this green. Should we revert this hack and keep the red branch while working on the account migration?

StefanRijnhart:
This could then be merged, of course.

hbrunn:
I'd suggest to merge this without the fix and hurry up with migrating hr_contract: https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=model%3D%22base.action.rule%22+language%3Axml&type=Code

StefanRijnhart:
@hbrunn you write "I'd suggest to merge this without the fix" 

but I don't know what you mean exactly. To be precise myself, I think the change to base_action_rule is fine, but I would rather not set modules uninstallable.

hbrunn:
@StefanRijnhart I feel uncomfortable patching upstream code where it's not absolutely necessary, so I'd prefer not to have 9314ff87188da31d1263b6c3bf7bb4f3d2485ac0 in OpenUpgrade. You know from painful experience in #895 that every change we apply is a possible source of conflicts.

My proposal is: Revert 9314ff87188da31d1263b6c3bf7bb4f3d2485ac0, merge https://github.com/OCA/OpenUpgrade/pull/897, then this should be green too, and we're happy

StefanRijnhart:
@hbrunn I can agree with that, yes.

StefanRijnhart:
Wait, you want to keep setting modules to uninstallable?

pedrobaeza:
@hbrunn I don't think the change in base_action_rule will conflict in the future, but anyway, have you checked if there's any other non migrated module that declares action rules?

hbrunn:
@StefanRijnhart setting modules to uninstallable: Now that I looked at the code for the first time, no, I don't want that either
@pedrobaeza I think https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=model%3D%22base.action.rule%22+language%3Axml&type=Code makes reasonably sure that's the last module causing trouble

pedrobaeza:
We still have problems with website module that don't anticipate before. Should I fix it here or do you prefer to merge this and create another PR fixing website?

StefanRijnhart:
We can merge, if you remove the change of setting the coa modules uninstallable.

pedrobaeza:
Removing that commit, I won't still be able to get a green branch although I fix website part. Can't we let the commit and revert later when dealing with account migration?

pedrobaeza:
OK, finally I'm going to translate my greenifying efforts to the account migration PR, so I have removed the commit that makes uninstallable the l10n_* modules and merging this one although in red. Then, I'll wait for website PR and rebase everything to work on the final solution at account level.

JordiBForgeFlow:
:+1:

pepetreshere:
👍 :)

PR #839, opened 2017-05-23
[ADD][10.0] account_asset migration script

cubells:
cc @Tecnativa


pedrobaeza (approved):

addons/account_asset/migrations/10.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
...
+
+column_renames = {
+    'account_asset': [
+        ('asset_id', None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
```
pedrobaeza: This is env now by default and you have to use `env.cr` then

pedrobaeza (changes requested):

addons/account_asset/migrations/10.0.1.0/pre-migration.py
cubells: Do you prefer to use use_env=False or to use use_env=True?

pedrobaeza: I prefer the standard (which is use_env=True)

addons/account_asset/migrations/10.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+column_renames = {
+    'account_move': [
+        ('asset_id', None),
+    ],
+}
+
+
+@openupgrade.migrate(use_env=True)
```
pedrobaeza, line 15: No need to put `use_env=True`, as it's the default in v10, but it's not so important.

pedrobaeza (approved):

PR #841, opened 2017-05-25
Rename account_analytic_plan_required

cubells:
Module  account_analytic_plan_required is renamed to account_analytic_distribution_required

cc @Tecnativa

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -3,6 +3,8 @@ to help the matching process
 """
 
 renamed_modules = {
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza (changes requested):
One thing I'm thinking: this should be target to 9.0, as the module dissapears there. For now, it's not migrated to this version, but anyway you have to pass through this version and the effect is going to be the same. Please change it.

cubells:
Done: https://github.com/OCA/OpenUpgrade/pull/842

I'll close this pr.

PR #843, opened 2017-05-29
[10.0] account

NL66278:

Current behavior before PR:

--

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ script:
     - pip install -q -r requirements.txt
     # this crashes if there is no test data
     - if [ -s ../test_data90.yml ]; then ./openerp-server --database=$DB --test-file=`readlink -f ../test_data90.yml` --test-commit --stop-after-init; fi
+    # Line below may fail occasionaly due to Travis bug:
     - git reset -q --hard $TRAVIS_COMMIT
     # Install Python requirements of target release
     - pip install -q -r requirements.txt
--- a/addons/account/__manifest__.py
+++ b/addons/account/__manifest__.py
@@ -67,6 +67,7 @@ You could use this simplified accounting in case you work with an (external) acc
     ],
     'demo': [
         'demo/account_demo.xml',
+        'demo/openupgrade_account_demo.xml',
     ],
     'qweb': [
         "static/src/xml/account_reconciliation.xml",
--- a/addons/account/demo/account_invoice_demo.yml
+++ b/addons/account/demo/account_invoice_demo.yml
@@ -12,9 +12,11 @@
       - product_id: product.consu_delivery_02
         price_unit: 642.0
         quantity: 5
+        account_id: account.openupgrade_demo_sales_account
       - product_id: product.consu_delivery_03
         price_unit: 280.0
         quantity: 5.0
+        account_id: account.openupgrade_demo_sales_account
 -
   !python {model: account.invoice, id: demo_invoice_1}:
     self.action_invoice_open()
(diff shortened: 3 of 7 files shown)


pedrobaeza:
@hbrunn @StefanRijnhart  I have rebased this PR on latest 10.0 branch and we continue having the same error installing l10n_generic_coa.

@NL66278 are you going to continue working on this or should I take over the work?

NL66278:
@pedrobaeza I will be working on this on friday.

pedrobaeza:
OK, let me know any progress or question

NL66278:
Now fails on l10n_be:
2017-06-09 11:56:13,326 41064 INFO openupgrade odoo.sql_db: bad query: INSERT INTO "account_invoice_line" ("id", "account_id", "layout_category_sequence", "invoice_id", "price_unit", "sequence", "discount", "product_id", "quantity", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval('account_invoice_line_id_seq'), NULL, 0, 8, '642.00', 10, '0.00', 163, '5.000', 1, 1, (now() at time zone 'UTC'), (now() at time zone 'UTC')) RETURNING id

2017-06-09 11:56:13,326 41064 ERROR openupgrade odoo.tools.yaml_import: null value in column "account_id" violates not-null constraint

DETAIL:  Failing row contains (19, null, 2017-06-09 11:56:10.618323, 10, 642.00, null, 1, null, null, null, 1, null, null, null, 0.00, 2017-06-09 11:56:10.618323, null, null, 163, 8, 5.000, null, null, null, null, null, null, null, null, 0, null).

pedrobaeza:
That's the same problem as it was happening with l10n_generic_coa, complaining about an account_id null value. I suppose that this is because now l10n_be is installed instead of the generic one, but the problem stays the same, and can be because there's something to do in account migration itself.

NL66278:
@pedrobaeza Something incredible happens in l10n_be, They are loading demo data from the account module in the manifest file. Talk about bad practice.... I think the account.invoice specified in the yaml file will not be created (because already existing), and then the child account.invoice lines have nothing to point to.

pedrobaeza:
All localization modules do that, but I think it's not loaded twice. Please check if in account that file is loaded, which I guess it isn't

NL66278:
@pedrobaeza Indeed the problem is different, account_id on an invoice line is now required, but is not set. I still think it bad practice to load demo data from another module, but that is not what is causing the crash.

NL66278:
@pedrobaeza I have not really a good idea on how to solve this. Seems best to patch the yaml file to also create an account and then refer to that account in the invoice lines. Of course the yaml file would then be different from what we have in standard Odoo. Or would you see another solution?

NL66278:
Some Travis builds fail, because this bug: https://github.com/travis-ci/travis-ci/issues/7459

When we reset branch to $TRAVIS_COMMIT, this env var may contain an outdated commit. If commit still exists, there will be no error and you will wonder why new features do not work. If it no longer exists, everything later on wil bomb, because you will still be on 9.0.

NL66278:
And finally it is green. Now please review the real functionality. And yes I have no authority to put a review label on the PR myself.

NL66278:
@StefanRijnhart @hbrunn Had to patch demo data to get this green. I wonder what you think about this, could there have been a better solution?

StefanRijnhart:
Well, the fix looks harmless. If it stops us worrying about l10n_be insanity, well done!

StefanRijnhart (approved):

PR #847, opened 2017-05-29
[10.0][FIX] Call cursor in _drop_column()

astirpe:
When the `_drop_column()` method of `ir.model` is executed, an error is raised.

Current behavior before PR:
Cursor is invoked as "old api" stile and an error is raised.

The `_drop_column()` method is executed without any error.

--


pedrobaeza (approved):
I'm pretty sure I changed this last week, but I forgot any of the occurrences...

PR #850, opened 2017-05-29
[10.0][ADD] Updated 'openupgrade_analysis.txt' files.

reinka:
@pedrobaeza @jbeficent @StefanRijnhart @NL66278  After running openupgrade analysis again, we saw there have been some changes since the first openupgrade analysis back in Oct 2016. This was the motivation for checking and updating all related migration scripts if needed.

--


omar7r:
:+1:

pedrobaeza (approved):
Tentatively approved, but I would like to know @StefanRijnhart's opinion

NL66278 (changes requested):
It is clear that something is wrong with the analysis text. Many fields are now said to be in the hw_ecspos module, but that does not have any models at all.

The first change in this PR says currency_id is in hw_ecspos instead of in analytic. In reality it is in analytic still.

pedrobaeza:
I think we need to blacklist this module because it explodes the system for having an stripped version for PosBox

NL66278:
@pedrobaeza Actually the module is in odoo/addons/openupgrade_records/blacklist.py.

But the blacklist might me needed in more places then used now.

pedrobaeza:
Yeah, it seems so. Then I bet now simply to remove the folder manually for making the analysis.

pedrobaeza (changes requested):
I remove my approval meanwhile

StefanRijnhart:
I arrive at the following changes: https://github.com/OCA/OpenUpgrade/pull/871

NL66278:
I would recommend closing this in favor of Stefan's PR: https://github.com/OCA/OpenUpgrade/pull/871

hbrunn:
closing as this is superseded by #871

PR #856, opened 2017-05-30
[9.0] openupgradelib requirement

NL66278:
Current behavior before PR: Starting a pip installed OpenUpgrade server fails because it misses openupgradelib

Desired behavior after PR is merged: pip installed openupgrade server runs OOTB.

--

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,7 @@ jcconv==0.2.3
 lxml==3.4.1
 mock==1.0.1
 ofxparse==0.14
+openerp-client-lib==1.1.2
 passlib==1.6.2
 psutil==2.2.0
 psycogreen==1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -39,4 +39,4 @@ vatnumber==1.2
 vobject==0.6.6
 wsgiref==0.1.2
 xlwt==0.7.5
-openupgradelib>=1.1.0
+openupgradelib>=1.3.0
--- a/setup.py
+++ b/setup.py
@@ -144,6 +144,7 @@ setup(
         'mako',
         'mock',
         'ofxparse',
+        'openupgradelib>=1.3.0',
         'passlib',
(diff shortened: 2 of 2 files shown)


pedrobaeza (changes requested):
Can you switch to the same scheme in Travis (directly from GitHub) to always have latest version?

NL66278:
@pedrobaeza It seems trying to use a certain source version in setup.py instead of a version on pypi seems very convoluted. Would it not be better to always have a specific version? If a new release of openupgradelib is released, and its features used in future versions of openupgrade, those can have the higher version of openupgradelib specified.

pedrobaeza:
Well, you know that we always work with other migration deployments and this is going to be outdated for sure. That's why putting one time the latest version, we assure not to need change it more, but if it's not possible... Anyway, maybe requirements.txt can have the latest one and setup.py the other?

NL66278:
@pedrobaeza I do not see any way to install a master version of openupgradelib from the setup.py of Odoo. There is some things you can find on the internet about how to handle dependency-links, adding -dev suffixes, needing to specify --process-dependency-links --allow-all-external etc. But nothing works.

So I think it better to leave the PR as is.

pedrobaeza:
This PR contains several unrelated commits about marketing.

NL66278:
@pedrobaeza I removed the extra commit. And finally found out why these extra commits are occuring. When I want to squash commits I use to to git rebase -i HEAD\~5 (the actual number after HEAD\~ might be different). This most of the time contains some extra commits to see what I am doing. But apparently, because I gpg sign my commits, the extra commits get signed and therefore changed as well. The solution seems to be quite precise in what to rebase and make sure only the relevant commits belonging to the branch are included.

pedrobaeza:
Next time, use `git rebase -i origin/9.0` to get always the latest and I think you will solve also these problems.

pedrobaeza (approved):

PR #864, opened 2017-05-30
[10.0][MIG] base_vat: Nothing to do

reinka:
Add `Nothing to do` flag to documentation.

--


tafaRU:
@pedrobaeza or some other @OCA/openupgrade-maintainers, just out of curiosity: why is there not `migrations`folder here?

StefanRijnhart:
@tafaRU modules with no detected changes have no such directory.

reinka:
@StefanRijnhart There's inconsistencies though, e.g. https://github.com/OCA/OpenUpgrade/blob/10.0/addons/l10n_be_intrastat/migrations/10.0.1.0/openupgrade_analysis.txt and https://github.com/OCA/OpenUpgrade/blob/10.0/addons/l10n_be_invoice_bba/migrations/10.0.1.2/openupgrade_analysis.txt. These directories should not exist.

StefanRijnhart:
@reinka yes, indeed. Any idea what causes these files to be present even though there are no changes?

reinka:
@StefanRijnhart My best guess would be someone added them manually because they thought it important to keep a consistent module structure.

PR #869, opened 2017-05-30
10.0 delivery

rruebner:
Added migration scripts for v10 and set Done flag in documentation.
Checked for noupdate=1 records and everything was fine.

--


addons/delivery/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+# © 2017 bloopark systems (<http://bloopark.de>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=False)
+def migrate(cr, version):
+    # map old False value to new value 'rate', old True / new 'rate_and_ship'
+    # values are set per default
+    cr.execute("""
```
pedrobaeza, line 11: Next time, you can use here openupgradelib map_values method.

PR #875, opened 2017-05-31
[10.0][MIGR] hr_gamification: nothing to do

tafaRU:


addons/hr_gamification/migrations/10.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,9 @@
+---Fields in module 'hr_gamification'---
+hr_gamification / hr.employee              / direct_badge_ids (one2many)   : NEW relation: gamification.badge.user
+hr_gamification / res.users                / goal_ids (one2many)           : NEW relation: gamification.goal
+# NOTHING TO DO
+
+---XML records in module 'hr_gamification'---
+NEW ir.rule: gamification.goal_gamification_manager_visibility
+DEL ir.rule: hr_gamification.goal_officer_visibility
```
hbrunn, line 8: actually, it's not true that there's nothing to do here: This is a noupdate record (https://github.com/OCA/OCB/blob/9.0/addons/hr_gamification/security/gamification_security.xml#L3), so you'll have to delete it in order not to have this rules interfere with stuff

hbrunn:
@tafaRU can you submit another PR for this?

PR #892, opened 2017-06-05
Rename payment_mode on migration

cubells:
cc @Tecnativa


cubells:
It's not correct this pr.

I close.

PR #898, opened 2017-06-07
[IMP] also install modules where there was nothing to do

hbrunn:
we also should install modules that are considered nothing to do

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ script:
     # on unreleased features in openupgradelib
     - pip install --ignore-installed git+https://github.com/OCA/openupgradelib.git@master
     # select modules and perform the upgrade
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This should wait until we have a green build to see if everything is correct.

hbrunn:
the intention is to get a green build with this. Modules like `hr_contract` won't be installed otherwise, and we need this to make `base_action_rule` work...
The problem now is that we have a bunch of modules depending on `website` migrated, while `website` itself errors out. I'll add a migration for that, let's see how this goes. But merging this won't make things worse, so I suggest to merge this anyways instead of hacking other stuff to make runbot green

pedrobaeza:
Puf, this is getting really complicated...

Are you going to include here website migration or in a separate PR?

hbrunn:
separate (completely untested) PR: #899

pedrobaeza:
OK, merging this one, and reviewing the other.

PR #903, opened 2017-06-07
[FIX] compare_noupdate_xml_records: Don't fail when there's no id on <record>

pedrobaeza:
Example: https://github.com/OCA/OpenUpgrade/blob/43eacb6066ba96e01bc05de6e9dc25d7c67c3fd7/addons/website/data/data.xml#L110


StefanRijnhart (approved):
How strange!

MiquelRForgeFlow (approved):

PR #904, opened 2017-06-07
[FIX] function signature

hbrunn:
@pedrobaeza thanks for the trust, but I messed up


MiquelRForgeFlow (approved):

pedrobaeza (approved):
Ouch, this is one of the obvious things that you don't check...

PR #905, opened 2017-06-07
[FIX] website: delete group assignments for views

hbrunn:
let's wait for the test results this time

### DIFF
--- a/addons/website/migrations/10.0.1.0/pre-migration.py
+++ b/addons/website/migrations/10.0.1.0/pre-migration.py
@@ -6,7 +6,13 @@ from openupgradelib import openupgrade
 
 @openupgrade.migrate()
 def migrate(env, version):
+    # rename groups' xmlids
     openupgrade.rename_xmlids(env.cr, [
         ('base.group_website_designer', 'website.group_website_designer'),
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Better to wait, yes, but tentatively OK

pedrobaeza:
Now the error is on stock module (which has been merged meanwhile), and we also have an error on the documentation build. We can merge this and continue fixing the rest.

JordiBForgeFlow:
@pedrobaeza @hbrunn The problem with stock is that a function is being called to re-confirm an a demo inventory adjustment that was already created an completed in Odoo 9: https://github.com/OCA/OpenUpgrade/blob/10.0/addons/stock/data/stock_demo.xml#L133

I am not sure why a function defined under a noupdate=1 data tag would be triggered again. I would assume that updating the stock module in a demo database in normal circumstances would trigger this error as well.

pedrobaeza:
@jbeficent this branch doesn't contain the stock migration, so the error will be gone for sure when merged.

hbrunn:
so let's merge :-)

pedrobaeza:
@hbrunn please check the documentation build part while I check account one

PR #908, opened 2017-06-08
[10.0][MIG] Several website modules

MiquelRForgeFlow:
This PR marks as <code>#NothingToDo</code> the following modules:
* website_blog
* website_crm
* website_crm_partner_assign
* website_event
* website_event_sale (without analysis file)
* website_google_map (without analysis file)
* website_hr (without analysis file)
* website_mail
* website_membership (without analysis file)
* website_partner
* website_project


pedrobaeza:
Have you checked noupdate records with the script compare_noupdate_records.py?

JordiBForgeFlow:
@pedrobaeza @hbrunn We have never used the script `compare_noupdate_records.py` before. Any tips on how to run it? Perhaps it would be good to run it for all modules, and create an analysis file with the results, so that we can consistently analyze this output?

pedrobaeza:
The problem is that it has to be done manually. The use is very simple:
```
python compare_noupdate_records.py <previous_version_module_path> <current_version_module_path>
```
The script can be updated for taking a full addons path and drop the noupdate_records.xml files inside migrations folder indeed, but will require much work.

MiquelRForgeFlow:
Hey, are you refering to <code>compare_noupdate_xml_records.py
</code> instead of <code>compare_noupdate_records.py</code>, right?

pedrobaeza:
Yeah, I was talking out of memory. The name is pretty similar, hehe :smile:

MiquelRForgeFlow:
@pedrobaeza I just checked all the noupdate records. It was useful to realize that `website_project` was a new module  😉

Ready to be merged.

pedrobaeza (approved):

PR #913, opened 2017-06-13
[9.0][Upd] module coverage list

MiquelRForgeFlow:
I found very useful in v10 the image tag to know which modules are missing. So I thought to apply the same in v9.

The changes in this PR are not an exhaustive list of missing modules but it's a first approach.


pedrobaeza:
Please complete the list then, or we will think that it's complete.

MiquelRForgeFlow:
The list is now exhaustively completed 💪😎

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -15,542 +15,657 @@ Status :
...
++----------------------------------------+-----------------------------------+
+|account_budget                          | Nothing to do                     |
++----------------------------------------+-----------------------------------+
+|account_cancel                          | Nothing to do                     |
++----------------------------------------+-----------------------------------+
+| |del| account_chart                    |                                   |
++----------------------------------------+-----------------------------------+
+| |new| account_check_printing           | Done                              |
++----------------------------------------+-----------------------------------+
+| |del| account_check_writing            | Renamed [#account_check_writing]_ |
```
pedrobaeza: Shouldn't this be account_check_printing?

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -15,542 +15,657 @@ Status :
...
++----------------------------------------+-----------------------------------+
+|account_analytic_default                | Nothing to do                     |
++----------------------------------------+-----------------------------------+
+| |del| account_analytic_plans           |                                   |
++----------------------------------------+-----------------------------------+
+| |del| account_anglo_saxon              |                                   |
++----------------------------------------+-----------------------------------+
+|account_asset                           | Done                              |
++----------------------------------------+-----------------------------------+
+| |del| account_bank_statement_extensions|                                   |
```
pedrobaeza: As you are changing the entire file, better to give some more room to the columns, specially the second column that can contain references to renamings

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -15,542 +15,657 @@ Status :
...
++========================================+===================================+
+|account                                 | Done (tentative)                  |
++----------------------------------------+-----------------------------------+
+|account_accountant                      | Nothing to do                     |
++----------------------------------------+-----------------------------------+
+| |del| account_analytic_analysis        |                                   |
++----------------------------------------+-----------------------------------+
+|account_analytic_default                | Nothing to do                     |
++----------------------------------------+-----------------------------------+
+| |del| account_analytic_plans           |                                   |
```
pedrobaeza: There's an equivalent module and OpenUpgrade renames it to it in https://github.com/OCA/account-analytic/tree/9.0/account_analytic_distribution

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -15,542 +15,657 @@ Status :
...
++----------------------------------------+-----------------------------------+
+|l10n_pl                                 |                                   |
++----------------------------------------+-----------------------------------+
+|l10n_pt                                 |                                   |
++----------------------------------------+-----------------------------------+
+|l10n_ro                                 |                                   |
++----------------------------------------+-----------------------------------+
+|l10n_sa                                 |                                   |
++----------------------------------------+-----------------------------------+
+|l10n_sg                                 |
```
pedrobaeza: Equivalent module and renamed in OU: https://github.com/OCA/account-analytic/tree/9.0/purchase_analytic_distribution

pedrobaeza: Equivalent module and renamed in OU in https://github.com/OCA/account-analytic/tree/9.0/sale_analytic_distribution

pedrobaeza (changes requested):

openerp/openupgrade/doc/source/modules80-90.rst
MiquelRForgeFlow: writing vs printing?  😏

pedrobaeza: Yes, I see that this a foot note, but at first sight I saw renamed and the same name and thought that there was an error.

openerp/openupgrade/doc/source/modules80-90.rst
MiquelRForgeFlow: Who encounters a problem with the width of the second column is who should change the width, not me 😒

pedrobaeza: Yeah, but this means again another full diff. That's why I asking you this favor for not having such big diffs each time.

MiquelRForgeFlow: Ok, I encountered the problem 😂

MiquelRForgeFlow:
It's ready @pedrobaeza

openerp/openupgrade/doc/source/modules80-90.rst
MiquelRForgeFlow: The spacing in the second column is done :+1:

pedrobaeza (approved):
Thank you very much. Now the documentation shines much more brighter :wink:

pedrobaeza:
Travis error is as always due to incorrect `TRAVIS_COMMIT` environment variable, so it's safe to merge.

PR #914, opened 2017-06-13
[9.0] More account speed ups

mvaled:
Copy the `currency_id` in invoice lines from the parent invoice.   It's a related field anyways.  For DBs with many invoice lines this saves a lot of time.

Compute the voucher line `price_subtotal` via SQL, only when all the vouchers have no taxes.


pedrobaeza:
Please only one change per PR, so they can be discussed separately and speed up the review process.

mvaled:
Ok. I'll split them and open new PRs.

mvaled:
First PR in #920.

PR #917, opened 2017-06-15
[IMP] Add moved fields to the analysis of their new module

StefanRijnhart:


MiquelRForgeFlow (approved):
Nice :+1:

pedrobaeza (approved):
Thanks again, @StefanRijnhart 

@mreficent please review new analysis lines of the already migrated modules to see if they impact something on the migration scripts.

mreficent:
Ok

MiquelRForgeFlow:
Ok @pedrobaeza 

@StefanRijnhart , should'nt this change be applied also in previous versions?

MiquelRForgeFlow:
@pedrobaeza, I reviewed the new analysis lines (just the selection fields) and there is only a new impact. It is in `sale` module, one that is not migrated yet.

pedrobaeza:
OK, thanks.

PR #920, opened 2017-06-19
[9.0] Speed up migration of `currency_id` in invoice lines

mvaled:
This is a small change: Since the field `currency_id` of invoice lines is just a related field from the same field on the parent invoice, we can simply copy it.


pedrobaeza:
Closing and opening for forcing Travis rebuild to see if `$TRAVIS_COMMIT` is correct

pedrobaeza (approved):

PR #921, opened 2017-06-19
[9.0] Speed up computation of voucher lines price_subtotal when voucher has no taxes.

mvaled:
`price_subtotal` can be simply computed by `quantity * price_unit` when the voucher line has no taxes.  The ``tax_ids`` field is a new M2M relation that we simply copy from the parent voucher's ``tax_id``.  So we may see that if there's no voucher with a ``tax_id`` we can blacklist the `price_subtotal` and fill it with the simpler formula.

--

### DIFF
--- a/addons/account_voucher/migrations/9.0.1.0/pre-migration.py
+++ b/addons/account_voucher/migrations/9.0.1.0/pre-migration.py
@@ -2,6 +2,11 @@
 # © 2016 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+from openerp.addons.account_voucher.account_voucher import \
+    account_voucher_line
+
+account_voucher_line._openupgrade_recompute_fields_blacklist = []
+
 
 column_copies = {
     'account_voucher': [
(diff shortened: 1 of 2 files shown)


addons/account_voucher/migrations/9.0.1.0/pre-migration.py
```diff
@@ -26,3 +31,15 @@ def migrate(cr, version):
     cr.execute("update account_voucher_line set name='/' where name is null")
     openupgrade.copy_columns(cr, column_copies)
     delete_payment_views(cr)
+
+    cr.execute('SELECT count(*) FROM account_voucher WHERE tax_id IS NOT NULL')
+    taxed_vouchers = cr.fetchone()[0]
```
pedrobaeza, line 36: If there are no records, this will raise an error, isn't it?

mvaled: Since we're counting PG should always return a single tuple.  The following shows it:

```
tmp=# CREATE TABLE tmp(name VARCHAR);
CREATE TABLE
tmp=# select count(*) from tmp where name IS NOT NULL;
 count 
-------
     0
(1 row)

tmp=# select count(*) from tmp where name IS NULL;
 count 
-------
     0
(1 row)
```

mvaled: Nevertheless, we could be safer by:

```python
res = cr.fetchone()
taxed_vouchers = res[0] if res else -1  # no records, no vouchers, nothing to migrate
```

pedrobaeza: OK, got it. It's correct as is.

pedrobaeza:
Don't you need to blacklist the computation of the field?

mvaled:
@pedrobaeza Yes. I do blacklist it but *only* when all vouchers have no taxes.  If I blacklisted it unconditionally then the SQL in the post migration would have to "translate" the computation of `compute_all` of taxes.  I don't have a DB where vouchers have taxes applied, so I can't verify that case.

pedrobaeza (approved):

PR #923, opened 2017-06-20
[9.0][FIX]account voucher. Delete inherited views first

AaronHForgeFlow:
Description of the issue/feature this PR addresses: There may exist some custom modules that inherits from the account.voucher.payment views

Current behavior before PR: Tried to migrate a database with a custom module that inherits a view from account.voucher.payment. Constraint Error when executing the query that deletes the view because there are some other views that inherit the original one.

Desired behavior after PR is merged: Delete all the views for account.voucher.payment. First, the inherited ones and then the standard ones.

--

### DIFF
--- a/addons/account_voucher/migrations/9.0.1.0/pre-migration.py
+++ b/addons/account_voucher/migrations/9.0.1.0/pre-migration.py
@@ -15,6 +15,13 @@ def delete_payment_views(cr):
     cr.execute(
         """\
         DELETE FROM ir_ui_view
+        WHERE inherit_id in (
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):
Perfect! :+1:

mvaled:
I think this is special case of #594. Don't you think we should provide a general solution?

pedrobaeza:
I don't see why is this needed. Using correctly openupgradelib's `update_module_names` method with the `merge_modules=True` argument, you can lead perfectly with dangling views on a missing module, as Odoo update process remove them.

AaronHForgeFlow:
All right @pedrobaeza Closing this :disappointed_relieved:

PR #929, opened 2017-06-21
[10.0][ADD] sale_order_back2draft: The feature is now in core 'sale' module

pedrobaeza:


MiquelRForgeFlow:
I think it's fine, but a bit odd, because `sale_order_back2draft` is in v8 repository but not in v9.

pedrobaeza:
But it's in v10 where the feature is integrated. If someone wants to have the feature in v9, then he/she should migrate it. If not, I'm migrating from v8 > v10, so here it's where the module should be merged.

MiquelRForgeFlow (approved):

PR #932, opened 2017-06-25
[ADD] re-add logo

hbrunn:
For whatever reason, the current logo is empty: https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/openupgrade/doc/source/images/OpenUpgrade.png


MiquelRForgeFlow (approved):

pedrobaeza:
Travis fails as always due to incorrect re-checkout, so I merge.

PR #944, opened 2017-06-30
[MIG] hr_timesheet: Migration scripts

pedrobaeza:
* Checked noupdate=1 records
* Included analysis work file
* Included migration scripts
* Updated documentation


addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,50 @@
...
+        SET allow_timesheets = True
+        FROM account_analytic_account aaa
+        WHERE aaa.id = pp.analytic_account_id
+        AND aaa.%s
+        """ % openupgrade.get_legacy_name('use_timesheets'),
+    )
+
+
+def fill_analytic_line_project(env):
+    """Fill project with the linked one in the related task or issue."""
```
MiquelRForgeFlow: The migration with project_issue should occur as part of the migration of project_issue module. If a database does not have the column issue_id because that module does not exist, this will break the migration here.

Same with project_task. Does not belong here.

Also, in v9 the account_analytic_line does not have a field task_id defined in hr_timesheet module.

addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,50 @@
...
+        AND pi.project_id IS NOT NULL
+        AND aal.project_id IS NULL
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    migrate_allow_timesheets(env)
+    fill_analytic_line_project(env)
```
MiquelRForgeFlow: In my opinion we need to consider that in v9 the analytic lines were linked to analytic account. 
The basic migration process should consist of detecting the analytic lines with `is_timesheet = True`, then look for an existing project associated to the analytic account of the line, and use it. If no project exists, create one.

Because there are companies that used analytic accounts not connected to projects to report time (e.g. Office time), and in v10 this is no longer allowed. But we need to respect the time entered in the past on those analytic accounts.

This is very, very important.

MiquelRForgeFlow (changes requested):

addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
pedrobaeza: No in the case of task_id, as hr_timesheet depends on project and it's the one that adds these fields in the account.analytic.line (see https://github.com/odoo/odoo/blob/c61ae3a3ed65b7a40b470076c5dbbd881e791f78/addons/hr_timesheet/hr_timesheet.py#L10). For field issue_id, I have to check.

addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
pedrobaeza: You are mixing things. This is a complementary data. You always have the analytic account in the line. The only thing is to have also a related project_id.

JordiBForgeFlow: But in v10 you can no longer report time on timesheets for analytic accounts. So all the time logged against pure analytic accounts will no longer be accessible on the timesheet.

pedrobaeza: What do you propose then?

JordiBForgeFlow: we need to make sure that all analytic lines connected to timesheets in v9 are now connected to projects in v10. And wherever a project does not exist, we must create one an connect it to that analytic account.

pedrobaeza: OK, anyway both methods are needed: move flag to projects linked to analytic accounts, and create projects from analytic accounts with the flag marked that haven't got a project.

JordiBForgeFlow:
The use_timesheets is deprecated in v9, but the field was not removed. So
the script will be still useful

pedrobaeza:
Added the project creation

addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,74 @@
...
+        WHERE pi.id = aal.issue_id
+        AND pi.project_id IS NOT NULL
+        AND aal.project_id IS NULL
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    migrate_allow_timesheets(env)
```
MiquelRForgeFlow, line 73: the `migrate_missing_projects` method is missing

MiquelRForgeFlow (changes requested):

pedrobaeza:
@mreficent included method call.

pedrobaeza:
As I have fixed what @mreficent asked and everything is green, I merge as there's other PR pending from this one (hr_timesheet_sheet).

remi-filament:
I still have an issue with this module since I do not have project_issue module installed.
A check whether project_issue is installed should be performed before running the lines updating account_analytic_line from project_issue table.

pedrobaeza:
Why? Is there any error? Please put an issue or open a PR with the check.

remi-filament:
Yes I have an error that model does not exist, I will try to raise a formal issue if you like

PR #949, opened 2017-07-03
[9.0][IMP] account: Merge invoice journals

pedrobaeza:
In version 8, Odoo had 2 journal types for each invoice operation (sale and sale_refund, purchase and purchase_refund). In v9, now both are merged in one journal, and this eases the view of all the operations together.

If we use a migrated v8 DB, we still see both journals, but creating new refunds are made on the same journal than the normal operation, splitting the invoices in 2 journals.

With this, we move the old refunds to the normal journal if it can be determined (if there's only one journal).

For cases where the journal layout is more complex, you can pass arguments for mapping the journals, although it has to be manually hacked on each installation, or look for another automation.

cc @Tecnativa

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -524,11 +524,11 @@ def merge_invoice_journals(env, refund_journal_ids=None, journal_mapping=None):
             FROM account_journal
             WHERE openupgrade_legacy_9_0_type = %s
             """
-        args = [journal_type, ]
+        query_args = [journal_type, ]
         if refund_journal_ids:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@jbeficent @mreficent can you review this?

pedrobaeza:
Merging optimistically with green status.

gopona:
Shouldn't merge_invoice_journals() be called before map_journal_state() which upgrade the old states ?

pedrobaeza:
It doesn't matter, as we are using `openupgrade_legacy_9_0_type` as the source for knowing the previous journal type for getting the refund ones. Note that only if one refund journal is found, the merge is done.

PR #950, opened 2017-07-04
[FIX] account: Fill move taxes

### LINKED ISSUE #806: [9.0] [account] Migration of taxes in account move lines
The migration scripts currently do not consider the migration of the taxes. I had to to a post migration fix for a customer, but still needs to be adjusted in the migration script.

pedrobaeza:
Try to deduce taxes in account move lines from old tax codes.

Method followed:

Tax amounts
-----------

* If the tax code is present in only one tax as tax code, then we can fill tax_line_id with that tax.
* If there are more than one tax with that tax code, try to match tax lines with the name of the move line.

Base amounts
------------

* If the tax code is present in only one tax as base code, then we can add that tax to tax_ids.
* If there are more than one for base amounts, see tax codes for other lines in the same move and try to match entire tax. If matched exactly with one tax, we can have 2 cases:

   * The move line has debit/credit: then we just put the tax as base.
   * It's a dummy line introduced for adding another base tax code, due to the restriction of <v8 of only one tax code per line. We locate the first line in the same move with the same amount in debit or credit column that doesn't have this tax already as base tax.

* If still no success, then we make a last try searching only for active taxes to see if this way we match only one record.

Closes #806

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -343,6 +343,181 @@ def migrate_account_auto_fy_sequence(env):
     env.cr.execute(query.format('suffix'))
 
 
+def fill_move_taxes(env):
+    """Try to deduce taxes in account move lines from old tax codes.
+
+    Method followed:
+
+    Tax amounts
+    -----------
+
+    * If the tax code is present in only one tax as tax code, then we can fill
+      tax_line_id with that tax.
+    * If there are more than one tax with that tax code, try to match tax lines
+      with the name of the move line.
+
+    Base amounts
+    ------------
+
+    * If the tax code is present in only one tax as base code, then we can add
+      that tax to tax_ids.
+    * If there are more than one for base amounts, see tax codes for other
+      lines in the same move and try to match entire tax. If matched exactly
+      with one tax, we can have 2 cases:
+
+       * The move line has debit/credit: then we just put the tax as base.
+       * It's a dummy line introduced for adding another base tax code, due to
+         the restriction of <v8 of only one tax code per line. We locate the
+         first line in the same move with the same amount in debit or credit
+         column that doesn't have this tax already as base tax.
+
+    * If still no success, then we make a last try searching only for active
+      taxes to see if this way we match only one record.
+    """
+    env.cr.execute(
+        "SELECT tax_code_id FROM account_move_line GROUP BY tax_code_id"
+    )
+    move_line_obj = env['account.move.line']
+    for row in env.cr.fetchall():
+        tax_code_id = row[0]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Now this is finished (at least all the finished it can be). It took a long time to figure out all the possibilities, but seems to be covered, as with this PR I have checked all the `l10n_es` taxes matches (which they are a lot), so for simpler localizations it would work also.

If no review in some time, I will merge it any way, as this is of course better than not having any tax assigned on v9.

JordiBForgeFlow:
Ok!

hbrunn:
could you create a file `tests/whatever.yml` to inject some test data in order to have the code tested? travis loads them as in https://github.com/Tecnativa/OpenUpgrade/blob/9.0/.travis.yml#L42-L43

pedrobaeza:
Yeah, I know that, but there's too much time already consumed here, and you know that replicating a chart of accounts with taxes and propagating them to journal items is not easy. In this case, unit testing is not an actual possibility.

misern2:
Pedro,

I checked the fill_move_taxes function and I found out a problem calculating the base tax with moves comming from invoice lines with several taxes in the same invoice line.

For example,

The following move is created with an invoice having two different taxes in the same invoice line.
captura de pantalla 2017-08-25 a les 17 55 51

In version 8 we have 2 entries to the 621000 account due the restriction on <v8. In version 9 we need to add all the taxes to the 621000 account move line with balance<>0.

captura de pantalla 2017-08-25 a les 17 56 14

Once the migration is finished, in the 621000 move we only have one tax (21% IVA soportado). The "Retenciones 19%" is missing.
 

We have a little different aproach for taxes calculation:
Starting from invoices, we look for the related move and we update the move line according to invoice line taxes.
You can check it at https://github.com/QubiQ/OpenUpgrade/commit/fae3833e6b89191d7df70460bcd978827f24137e

If you think it is usefull, I can make a PR!.

Best Regards!.
marsal.

pedrobaeza:
Hi, Marçal, good to see you joining here. I'm already doing the operation you mention here: https://github.com/OCA/OpenUpgrade/pull/950/files#diff-37a531f02f2e57d4fa1bc4bfcb55404aR483 and it's working fine for my control DB.

JordiBForgeFlow (approved):

pedrobaeza:
Merging finally this one.

pedrobaeza:
hehe, curiously today I have found a little problem after merging this that it was noted by @misern2 but I didn't reproduce till now. I have directly commited  the solution in https://github.com/OCA/OpenUpgrade/commit/164c70410d825ff6fbf211d64d360a603388a23f as I have tested in the DB I have found the problem.

PR #954, opened 2017-07-05
[8.0] 'purchase' improve reception to invoice computation (stock picking field)

legalsylvain:

In V8, purchase module introduce a new computed & stored field reception_to_invoice. For the time being, the computation of this field is managed by the ORM.

In a OpenUpgrade regular migration, the computation is done for all the picking, even if the compute is irrelevant for out picking. (sale picking). Therefore, the computation of this field takes too much time unnecessarily.

Make the purchase upgrade migration faster.
- create the field is the pre-migration field, with default False value.
- recompute the field only for relevant picking. (pickings that have move with puchase_line_id defined, and invoice_method set to 'picking').

See the the compute method for more information : 
https://github.com/OCA/OpenUpgrade/blob/8.0/addons/purchase/stock.py#L200

BTW, I don't understand why this new field doesn't appear in the analysis file.

--

### DIFF
--- a/addons/purchase/migrations/8.0.1.1/post-migration.py
+++ b/addons/purchase/migrations/8.0.1.1/post-migration.py
@@ -22,6 +22,7 @@
 import logging
 from openerp.openupgrade import openupgrade, openupgrade_80
 from openerp import pooler, SUPERUSER_ID as uid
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Thank you! The field is not listed because new stored function fields are handled automatically by the ORM. But as you mention already, this is a good performance improvement.

PR #955, opened 2017-07-05
[10.0][FIX] openupgrade_records

MiquelRForgeFlow:
This PR fixes the buttons of the `openupgrade.comparison.config` view and their corresponding methods.

### DIFF
--- a/odoo/addons/openupgrade_records/models/comparison_config.py
+++ b/odoo/addons/openupgrade_records/models/comparison_config.py
@@ -45,12 +45,11 @@ class openupgrade_comparison_config(models.Model):
             connection = self.get_connection()
             user_model = connection.get_model("res.users")
             ids = user_model.search([("login", "=", "admin")])
-            user_info = user_model.read(ids[0], ["name"])
+            user_info = user_model.read([ids[0]], ["name"])[0]
         except Exception, e:
             raise UserError(
-                _("Connection failed."), unicode(e))
+                unicode(e))
         raise UserError(
-            _("Connection succesful."),
             _("%s is connected.") % user_info["name"])
 
     @api.multi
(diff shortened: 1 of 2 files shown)


odoo/addons/openupgrade_records/models/comparison_config.py
```diff
@@ -45,12 +45,11 @@ def test_connection(self):
             connection = self.get_connection()
             user_model = connection.get_model("res.users")
             ids = user_model.search([("login", "=", "admin")])
-            user_info = user_model.read(ids[0], ["name"])
+            user_info = user_model.read(ids, ["name"])[0]
```
pedrobaeza: Don't read all the ids. Just make:
```
user_info = user_model.read([ids[0]], ["name"])[0]
```
And why not browse?

odoo/addons/openupgrade_records/models/comparison_config.py
```diff
@@ -81,9 +80,10 @@ def install_modules(self):
 
         modules = []
         for module_id in remote_module_ids:
-            name = remote_module_obj.read(module_id, ["name"])
-            name = apriori.renamed_modules.get(name, name)
-            modules.append(name)
+            mod = remote_module_obj.read([module_id], ["name"])[0]
```
pedrobaeza, line 83: The same about browse

pedrobaeza (changes requested):

odoo/addons/openupgrade_records/models/comparison_config.py
MiquelRForgeFlow: I don't now why, but the `browse` fails in this two cases

odoo/addons/openupgrade_records/models/comparison_config.py
MiquelRForgeFlow: I think the `browse` fails because the models are not from `env['model']` but from `connection.get_model('model')`

pedrobaeza: Yeah, you're right. I didn't remember about XML-RPC connection.

pedrobaeza (approved):

PR #957, opened 2017-07-05
[10.0]Sale Migration missing tuples

flotho:
Implements tuple syntax

Current behavior before PR:

--

### DIFF
--- a/addons/sale/migrations/10.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/10.0.1.0/pre-migration.py
@@ -86,14 +86,14 @@ def warning_update_module_names_partial(cr):
         return
 
     # get moved model fields
-    moved_fields = ['sale_line_warn', 'sale_line_warn_msg', 'sale_warn',
-                    'sale_warn_msg']
+    moved_fields = tuple(['sale_line_warn', 'sale_line_warn_msg', 'sale_warn',
+                    'sale_warn_msg'])
     cr.execute("""
         SELECT id
         FROM ir_model_fields
         WHERE model IN ('res.partner', 'product.template') AND name in %s
     """, (moved_fields,))
-    field_ids = [r[0] for r in cr.fetchall()]
+    field_ids = tuple([r[0] for r in cr.fetchall()])
 
     # update ir_model_data, the subselect allows to avoid duplicated XML-IDs
     query = ("UPDATE ir_model_data SET module = %s "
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Next time, you can make directly the PR and put the issue here.

addons/sale/migrations/10.0.1.0/pre-migration.py
```diff
@@ -92,20 +92,20 @@ def warning_update_module_names_partial(cr):
         SELECT id
         FROM ir_model_fields
         WHERE model IN ('res.partner', 'product.template') AND name in %s
-    """, (moved_fields,))
+    """, (tuple(moved_fields),))
     field_ids = [r[0] for r in cr.fetchall()]
```
pedrobaeza: Better to make here `fields_ids = tuple([r[0]...` and you avoid to put it in all the places where it's used.

pedrobaeza (changes requested):
Please check if there's the same case on the rest of the script and the post-migration script

It was not detected by Travis as the module is not installed being removed. @hbrunn can we make something with this to recognize also |del| modules to be installed on version n-1?

addons/sale/migrations/10.0.1.0/pre-migration.py
hbrunn: this check happpens in https://github.com/OCA/OpenUpgrade/blob/10.0/.travis.yml#L64 - so either use one of the keywords this greps for deleted addons too or change the regex to output also deleted addons

flotho:
Hi @pedrobaeza, sorry to ask such an obvious thing but I'm not able to identify what part of code make this job failed :  https://travis-ci.org/OCA/OpenUpgrade/jobs/250747464#L387

BTW, I'm not sure to understand the #958, is it better to use tuple or change to ANY ?

Regards

pedrobaeza:
It's because:
```
./addons/sale/migrations/10.0.1.0/pre-migration.py:89:66: W291 trailing whitespace
```
You have an space at the end of the line. Configure your IDE to remove them automatically.

About the best query, `IN` or `= ANY`, the final conclusion is that it performs the same (IN is transformed to = ANY), so the only advantage of ANY is to not convert everything to tuples.

flotho:
@pedrobaeza , thanks I was expecting to have the error below the red line.

flotho:
Hi @pedrobaeza , @hbrunn , 

Is it possible to have either a merge on my PR or a positive review and merge on the #958 ?

regards

pedrobaeza (approved):

PR #958, opened 2017-07-06
[10.0] Replace SQL IN operator by = ANY to format query with lists

ecino:
Some migrations may fail using the IN operator of postgres with an array. I replace that with  = ANY.

Current behavior before PR:
Migration to version 10 of modules sale and stock may fail

Migration work

--

### DIFF
--- a/addons/sale/migrations/10.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/10.0.1.0/pre-migration.py
@@ -97,14 +97,14 @@ def warning_update_module_names_partial(cr):
 
     # update ir_model_data, the subselect allows to avoid duplicated XML-IDs
     query = ("UPDATE ir_model_data SET module = %s "
-             "WHERE module = %s AND res_id IN %s AND name NOT IN "
+             "WHERE module = %s AND res_id = ANY(%s) AND name NOT IN "
              "(SELECT name FROM ir_model_data WHERE module = %s)")
     openupgrade.logged_query(cr, query, (new_name, old_name, field_ids,
                                          new_name))
 
     # update ir_translation
     query = ("UPDATE ir_translation SET module = %s "
-             "WHERE module = %s AND res_id IN %s")
+             "WHERE module = %s AND res_id = ANY(%s)")
     openupgrade.logged_query(cr, query, (new_name, old_name, field_ids))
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Check #957 for an alternative and please use tuple instead as suggested there.

addons/stock/migrations/10.0.1.1/post-migration.py
```diff
@@ -21,9 +21,9 @@ def update_picking_type_id(env):
     :param env: environment variable (self)
     """
     # load xml data to be used for filling in missing info
-    xml_stock_picking_type_int = env.ref("stock.picking_type_internal")
-    xml_stock_picking_type_out = env.ref("stock.picking_type_out")
-    xml_stock_picking_type_in = env.ref("stock.picking_type_in")
+    xml_stock_picking_type_int = env.ref("stock.picking_type_internal", False)
```
pedrobaeza: If these refs are missing, then you have a serious problem... but the solution is not this, as later data will remain dangled.

ecino:
Is it really better to use tuples? Not sure

pedrobaeza:
OK, you are right. It performs the same as `IN` operator. Reference: https://stackoverflow.com/a/31192557

addons/stock/migrations/10.0.1.1/post-migration.py
ecino: You're probably right but I cannot find any reference to stock.picking_type_internal outside from demo data of module stock_picking_wave which is not installed on our database. Are you sure these ids should always be present?

pedrobaeza:
But there are other places where you need to make the same change. And there's also the question about the types.

addons/stock/migrations/10.0.1.1/post-migration.py
pedrobaeza: It's a bit tricky because they are YML data: https://github.com/odoo/odoo/blob/afefa1b7717b47f5d5c2c156be82d4aaed789bf7/addons/stock/data/stock_data.yml#L16

ecino: Ok I missed that one, thanks for pointing me to this. I will look then why I don't have this in my database. So should I close this request in favor of #957?

pedrobaeza:
No, if you complete the places where we need the `ANY`, and remove the other part, then we can merge this one as in the other PR @flotho has not answered yet.

ecino:
Ok, I will do the changes tomorrow.

ecino:
I made the changes and updated the pull request.

pedrobaeza:
Hi @ecino, I didn't see the changes as you made `git push -f`, which doesn't send a message, and didn't write any comment, so I have merged the other PR as it was fixed and it's working. Sorry for the misunderstanding.

ecino:
I made the comment (just above), but no worries.

pedrobaeza:
Ouch, I see it now. I don't know why I didn't receive it by e-mail (or maybe I just lost it). Sorry again.

PR #959, opened 2017-07-06
[10.0][MIG] point_of_sale

MiquelRForgeFlow:
- Checked noupdate=1 records
- Included migration scripts
- Updated documentation


addons/point_of_sale/migrations/10.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent - Miquel Raich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def map_pos_config_state_active(cr):
+    cr.execute(
```
pedrobaeza: Maybe you can use here `map_values` openupgradelib method

addons/point_of_sale/migrations/10.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,15 @@
...
+column_copies = {
+    'pos_config': [
+        ('state', None, None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, column_copies)
```
pedrobaeza: No need to copy the column, as the column is not dropped, so you can use it as is. We only copy columns, when we need to make changes over the same column (for example, if there are changes on the values of a selection field).

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #960, opened 2017-07-11
[9.0][UPD] Several renames and merges

MiquelRForgeFlow:
This PR adds several renames and merges.


openerp/addons/openupgrade_records/lib/apriori.py
```diff
@@ -46,6 +49,11 @@
     'hr_expense_analytic_plans': 'hr_expense_analytic_distribution',
     # OCA/project
     'project_work_time_control': 'project_timesheet_time_control',
+    # OCA/manufacture-reporting
+    'mrp_bom_structure_xls': 'mrp_bom_structure_xlsx',
+    'mrp_bom_structure_xls_level_1': 'mrp_bom_structure_xlsx_level_1',
+    # OCA/reporting-engine
+    'report_xls': 'report_xlsx',
```
pedrobaeza, line 56: This one can't be renamed this way: you can be using report_xls for other modules that haven't been converted to XLSX.

openerp/addons/base/migrations/9.0.1.3/pre-migration.py
```diff
@@ -59,6 +59,7 @@ def cleanup_modules(cr):
         cr, [
             ('contacts', 'mail'),
             ('marketing_crm', 'crm'),
+            ('email_template', 'mail'),  # mail_template class
```
pedrobaeza, line 62: This one is already done in https://github.com/OCA/OpenUpgrade/blob/9.0/addons/mail/migrations/9.0.1.0/pre-migration.py#L25. Maybe it's better to move it here, but remove it from the other place then.

openerp/addons/openupgrade_records/lib/apriori.py
MiquelRForgeFlow: I disagree. `report_xls` of `reporting-engine` is not installable in v9. We are doing thinks based on what OCA has now in the repositories, not based on what some individuals may do.

pedrobaeza: OK, I thought it was ported (you did the PR), then let's remove it definitively report_xls in OCA in both v9 and v10.

openerp/addons/base/migrations/9.0.1.3/pre-migration.py
MiquelRForgeFlow: Done

openerp/addons/openupgrade_records/lib/apriori.py
MiquelRForgeFlow: Yeah, could you remove them?

pedrobaeza: Well, I was thinking more in you for that task :stuck_out_tongue_closed_eyes:

MiquelRForgeFlow: 😒 ok..., I will do it now 😅

pedrobaeza: Thanks! I will be in the review side!

openerp/addons/base/migrations/9.0.1.3/pre-migration.py
MiquelRForgeFlow: Should I move the associated model and table renames scripts?

pedrobaeza: No, that should be in the corresponding module. This is only the initiation of the migration, for setting which modules to upgrade. When upgrading the mail module is when renamings should happen.

MiquelRForgeFlow: Ok, fine!

pedrobaeza (approved):

PR #962, opened 2017-07-11
[10.0][MIG] marketing

MiquelRForgeFlow:
- Checked noupdate=1 records
- Included migration scripts
- Updated documentation


pedrobaeza (changes requested):
You must update the `marketing` module name to `marketing_campaign` in base module with `merge_modules=True`option and then you won't need any script in marketing campaign.

pedrobaeza (approved):

PR #968, opened 2017-07-20
10.0 website sale

gfcapalbo:

Current behavior before PR:

--


pedrobaeza:
Please start the PR with a clean commit history (not 7 commits)

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,20 @@
...
+    cr = env.cr
+    sql = """select pricelist_id, website_id, selectable from 
+          website_pricelist_openupgrade_10"""
+    cr.execute(sql)
+    for pricelist_id , website_id, selectable in  cr.fetchall():
+        env['product.pricelist'].search([('id', '=', pricelist_id)]).write({
+            'website_id': website_id,
+            'selectable': selectable
+        })
+    # now that all pricelists have a their website the field pricelist_ids in
```
MiquelRForgeFlow: `#` now that all the pricelist_ids have their corresponding website_id and selectable
or something similar

MiquelRForgeFlow (changes requested):
in the product.pricelist table, you should fill the field <code>website_id</code> by <code>self.env['website'].search([], limit=1)</code> as default in the cases where it is empty

also please check travis, I dunno why it fails

pedrobaeza:
Where is the analysis work file?

addons/website_sale/migrations/10.0.1.0/post-migration.py
gfcapalbo: website_id has a default  self.env['website'].search([], limit=1)  ,but is not a required field.
So it should be empty sometimes?
I had started to code your proposal by calling _default_website(), but technically if the original db has no website ID so it should remain.
You can make null.

do you agree?

MiquelRForgeFlow:
Why my commit of sale_crm is here? 😵

gfcapalbo:
@mreficent  rebase error, fixing

gfcapalbo:
rebased, added defaults,  linted, passing.

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+    cr.execute(sql)
+    pricelists = []
+    for pricelist_id, website_id, selectable in cr.fetchall():
+        pl_model.search([('id', '=', pricelist_id)]).write({
+            'website_id': website_id,
+            'selectable': selectable
+        })
+        pricelists.append(pricelist_id)
+    # all remaining pricelists will be assigned to default
+    for pricelist in pl_model.search([]):
```
MiquelRForgeFlow: `pl_model.search([('id', 'not in', pricelists)])` is simpler

MiquelRForgeFlow (changes requested):

gfcapalbo:
improved pricelist scanning and rebased  again

MiquelRForgeFlow (approved):

pedrobaeza (changes requested):
This still misses the analysis work file.

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+            'website_id': website_id,
+            'selectable': selectable
+        })
+        pricelists.append(pricelist_id)
+    # all remaining pricelists will be assigned to default
+    for pricelist in pl_model.search([('id', 'not in', pricelists)]):
+        pricelist.write({'website_id': pricelist._default_website().id})
+    # now that all pricelists have a their corresponding website_id
+    # and selectable drop support table
+    sql = "drop table website_pricelist_openupgrade_10"
```
pedrobaeza: Don't drop this table, as it can be used for other modules that inherit old website.pricelist model and needs to make a migration.

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    cr = env.cr
+    pl_model = env['product.pricelist']
+    sql = """select pricelist_id, website_id, selectable from
+          website_pricelist_openupgrade_10"""
+    cr.execute(sql)
+    pricelists = []
```
pedrobaeza: `pricelist_ids` please

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+    for pricelist_id, website_id, selectable in cr.fetchall():
+        pl_model.search([('id', '=', pricelist_id)]).write({
```
pedrobaeza: Don't make a search, but a browse.

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+            'website_id': website_id,
+            'selectable': selectable
```
pedrobaeza: end comma for reducing possible future diffs.

addons/website_sale/migrations/10.0.1.0/post-migration.py
pedrobaeza: I'm thinking that this is better done with an update query:
```
UPDATE product_pricelist pp
SET website_id = wp.id,
    selectable = wp.selectable
FROM website_pricelist_openupgrade_10 wp
WHERE wp.pricelist_id = pp.id
```
(and make it with openupgrade.logged_query)

pedrobaeza (changes requested):

addons/website_sale/migrations/10.0.1.0/post-migration.py
gfcapalbo: agreed , query,  logged.

gfcapalbo: this query does not consider those pricelists that need to be set at the current default website, so anyways we are just copying the old column so we can save the first loop with the same query , but we still need the second to update all those possible pricelist ids not in the website_pricelist_openupgrade_10 column with the default website.   addressed in 20c76159af016f73e7b271727ae9bba0cfb962ba  waiting for checks to pass.

gfcapalbo:
@pedrobaeza used your query for the mig of the non-default pricelists. 
also added some explanation on the analysis file

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+    cr = env.cr
+    pl_model = env['product.pricelist']
+    sql = """
+    UPDATE product_pricelist pp
+    SET website_id = wp.id,
+        selectable = wp.selectable
+        FROM website_pricelist_openupgrade_10 wp
+        WHERE wp.pricelist_id = pp.id
+    """
+    cr.execute(sql)
```
pedrobaeza: This is not done with `openupgrade.logged_query`.

addons/website_sale/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+    # all remaining pricelists will be assigned to default
+    sql = """select pricelist_id from
+          website_pricelist_openupgrade_10"""
+    cr.execute(sql)
+    pricelist_ids = cr.fetchall()
+    for pricelist in pl_model.search([('id', 'not in', pricelist_ids)]):
+        pricelist.write({'website_id': pricelist._default_website().id})
```
pedrobaeza, line 25: Cache the default website in a previous variable for avoiding successive calls to the method.

pedrobaeza (changes requested):

addons/website_sale/migrations/10.0.1.0/post-migration.py
gfcapalbo: I did this in case the default calculation became more complex in a near future (e.g. per-pricelists default websites) , as things stand, you are right, but wouldn't you agree to sacrifice performance to increase modularity in a one-time run like a migration?

pedrobaeza: OK

pedrobaeza (approved):

pedrobaeza:
Hey, I'm seeing a problem with Travis:
```
IntegrityError: insert or update on table "product_pricelist" violates foreign key constraint "product_pricelist_website_id_fkey"

DETAIL:  Key (website_id)=(3) is not present in table "website".
```

MiquelRForgeFlow:
ready to be merged :+1:

PR #970, opened 2017-07-26
[9.0][FIX] Merge report_xls into report_xlsx and delete from apriori renamed module

sergio-teruel:
cc @Tecnativa

--


sergio-teruel:
@pedrobaeza Can you take a look?

pedrobaeza (approved):
This is because you can have both `report_xls` and `report_xlsx` modules listed in your old v8 instance, and thus this gives error.

:+1: and merging when CIs finish.

MiquelRForgeFlow (approved):

pedrobaeza:
Travis problem is the typical one with the checkout problem, so I merge.

PR #971, opened 2017-07-26
[10.0][MIG] all payment_* modules

MiquelRForgeFlow:
- Updated documentation
- Included analysis work file
- Included migration scripts
- Checked noupdate=1 records


pedrobaeza (changes requested):
You have to rename both payment.acquirer XML-IDs for avoiding to have duplicate acquirer records on new version and incorrect references.

MiquelRForgeFlow:
It's seems Travis didn't like the renames.

pedrobaeza:
I see the problem: now the acquirers are declared on main `payment` module (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/payment/data/payment_acquirer_data.xml), so you have to include these renamings (all of them) in that module migration.

pedrobaeza (approved):

PR #972, opened 2017-07-26
[10.0][MIG] portal/portal_sale/portal_stock

MiquelRForgeFlow:
- Updated documentation
- Included analysis work file
- Included migration scripts
- Checked noupdate=1 records


addons/portal_sale/migrations/10.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,30 @@
...
+DEL ir.model.access: portal_sale.access_res_partner_category
+DEL ir.model.access: portal_sale.access_sale_order
+DEL ir.model.access: portal_sale.access_sale_order_line
+DEL ir.ui.menu: portal_sale.portal_invoices
+DEL ir.ui.menu: portal_sale.portal_quotations
+DEL ir.ui.menu: portal_sale.portal_sales_orders
+DEL ir.ui.view: portal_sale.invoice_form_portal
+DEL ir.ui.view: portal_sale.view_order_form_portal_sale
+DEL mail.template: portal_sale.email_template_edi_invoice
+DEL mail.template: portal_sale.email_template_edi_sale
```
pedrobaeza: v9 version of portal_sale overwrites the method `action_quotation_send` of the model  `sale.order`, changing the default template to be used for sending the mail to the one provided by this module. As now there's no such change (as the template itself disappears), we expect that the default template will be the same (and we can even customize this template to our needs, not the sale one).

Following this rationale, you have to provide a migration script here, that removes xml-id reference to `sale.email_template_edi_sale`, and rename XML-ID `portal_sale.email_template_edi_sale` to `sale.email_template_edi_sale`.

pedrobaeza: The same logic applies also for the invoice template.

pedrobaeza (changes requested):

addons/portal_sale/migrations/10.0.0.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Sorry, I don't get it  😳

pedrobaeza: Do:
```
env['ir.model.data'].search([
    ('module', '=', 'sale'),
    ('name', '=', 'email_template_edi_sale'),
]).unlink()
openupgrade.rename_xml_ids([
    ('sale_portal.email_template_edi_sale',
     'sale.email_template_edi_sale'),
])
```
and the same with invoice one.

MiquelRForgeFlow: Thanks

gfcapalbo:
@pedrobaeza thanks , missed the template.

pedrobaeza (approved):

PR #973, opened 2017-07-26
[10.0][MIG] im_livechat

MiquelRForgeFlow:
- Updated documentation
- Included analysis work file
- Included migration scripts
- Checked noupdate=1 records


addons/im_livechat/migrations/10.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Fields in module 'im_livechat'---
+---XML records in module 'im_livechat'---
+NEW ir.actions.act_window: im_livechat.im_livechat_canned_response_action
+NEW ir.actions.act_window.view: im_livechat.mail_channel_action_livechat_form
+NEW ir.actions.act_window.view: im_livechat.mail_channel_action_livechat_tree
+DEL ir.actions.client: im_livechat.ir_action_client_open_livechat_menu
+DEL ir.actions.todo: base.open_menu
+DEL ir.cron: im_livechat.ir_cron_remove_empty_session
```
pedrobaeza: You have to manually remove this cron in a migration script as this is flagged `noupdate=1`

addons/im_livechat/migrations/10.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Fields in module 'im_livechat'---
+---XML records in module 'im_livechat'---
+NEW ir.actions.act_window: im_livechat.im_livechat_canned_response_action
+NEW ir.actions.act_window.view: im_livechat.mail_channel_action_livechat_form
+NEW ir.actions.act_window.view: im_livechat.mail_channel_action_livechat_tree
+DEL ir.actions.client: im_livechat.ir_action_client_open_livechat_menu
+DEL ir.actions.todo: base.open_menu
```
pedrobaeza: You have to manually remove this todo in a migration script as this is flagged `noupdate=1`

addons/im_livechat/migrations/10.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Fields in module 'im_livechat'---
+---XML records in module 'im_livechat'---
+NEW ir.actions.act_window: im_livechat.im_livechat_canned_response_action
+NEW ir.actions.act_window.view: im_livechat.mail_channel_action_livechat_form
+NEW ir.actions.act_window.view: im_livechat.mail_channel_action_livechat_tree
+DEL ir.actions.client: im_livechat.ir_action_client_open_livechat_menu
```
pedrobaeza: You have to manually remove this action in a migration script as this is flagged `noupdate=1`

pedrobaeza (changes requested):

addons/im_livechat/migrations/10.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: I used the `compare_noupdate_xml_records` script and the result was nothing. The `ir.actions.*`  always are ignored. If you insist, please, show me an existing example of removing `ir.actions.*`

pedrobaeza: We have already talked about this. This is something that the script is not going to detect. It's simply: noupdate=1 records are not automatically removed if they disappear, and thus, all of these records will remain. Imagine a cron that is logging error each time it's executed because the method that is calling is missing. That's one of the cases of not removing these noupdate=1 records.

addons/im_livechat/migrations/10.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env['ir.model.data'].search([
```
pedrobaeza: You have to remove the records, not the XML-IDs:
```
xml_ids = [
    'im_livechat.ir_action_client_open_livechat_menu',
    'im_livechat.ir_cron_remove_empty_session',
]
for xml_id in xml_ids:
    self.env.ref(xml_id).unlink()
```
I have seen that the the record referenced from base module is not needed.

pedrobaeza (approved):

pedrobaeza:
@mreficent next time you make an amend/squash and `git push -f`, please throw a message in the PR as we are not notified about this push.

PR #976, opened 2017-07-26
[ADD] sale_documents_comments renamed

chienandalu:
Related https://github.com/OCA/sale-workflow/pull/509#issuecomment-317485138

cc @Tecnativa

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -96,6 +96,7 @@ def cleanup_modules(cr):
             ('sale_order_back2draft', 'sale'),
             ('partner_prepayment', 'sale_delivery_block'),
             ('sale_fiscal_position_update', 'sale'),
+            ('sale_documents_comments', 'sale_comment_propagation'),
             # from OCA/bank-payment
             ('account_payment_sale_stock', 'account_payment_sale'),
             # from OCA/website


openerp/addons/openupgrade_records/lib/apriori.py
```diff
@@ -29,6 +29,7 @@
     'marketing_security_group': 'mass_mailing_security_group',
     # odoomrp/odoomrp-wip
     'product_variants_no_automatic_creation': 'product_variant_configurator',
+    'sale_documents_comments': 'sale_comment_propagation',
```
MiquelRForgeFlow: the module https://github.com/OCA/sale-workflow/pull/509 is still named `sale_comment_propagation`, see last commit

MiquelRForgeFlow (changes requested):

mreficent:
I think this PR should be closed, at least while the referenced PR is named as it is now

openerp/addons/openupgrade_records/lib/apriori.py
pedrobaeza: We know that. This is for ensuring the migration path from odoomrp-wip module to OCA. If you have already 8.0 OCA module, this is harmless (well, not yet).

pedrobaeza (changes requested):
@chienandalu, you have to include the renaming in the base module where update_module_names is called with merge_modules=True option, as you can have both modules (old and new) in your module list.

openerp/addons/openupgrade_records/lib/apriori.py
pedrobaeza: It has been already done in other odoomrp-wip modules. See https://github.com/OCA/OpenUpgrade/blob/3243e327e903bd0b24db41f9cf922bc2cdb7fc20/openerp/addons/openupgrade_records/lib/apriori.py#L30

MiquelRForgeFlow: Oh, fine!

chienandalu:
Done @pedrobaeza

pedrobaeza:
But you haven't removed it from apriori.py...

chienandalu:
Sure, @pedrobaeza... :sweat_smile:  It's done

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
@chienandalu, why you put the depends in the head comment? It doesn't depend at all.

chienandalu:
@mreficent Yes, now it doesn't. Changed.

pedrobaeza (approved):

pedrobaeza:
It seems Travis has fixed its problem referencing $TRAVIS_COMMIT with forced pushed PRs...

PR #978, opened 2017-07-26
[9.0][FIX] account: Correct invoice line price_subtotal_signed

luismontalba:
Correction in calculation of price_subtotal_signed field in account.invoice.line model

--


pedrobaeza (approved):
This is the same as the bug I found and solved in https://github.com/OCA/OpenUpgrade/commit/bc41903c25ead6ed162581efb2cb2a2179382efa, so I merge it.

PR #985, opened 2017-08-01
[FIX] call function, not module

hbrunn:
my migration fails here, and I must admit I wonder how this worked before. Probably some weird loading order of the python modules issue

### DIFF
--- a/odoo/modules/graph.py
+++ b/odoo/modules/graph.py
@@ -9,6 +9,7 @@ import itertools
 import zipimport
 
 import odoo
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Uhm, previously it should work previously due to `import odoo.tools as tools`. I don't understand why it doesn't work now. Any possible `tools` variable override?

hbrunn:
well, it depends on the code path how (and in which order) the file is imported. This now seems to work either way

pedrobaeza:
OK, let's merge it as it's harmless any way.

PR #1001, opened 2017-08-04
[10.0][MIG] hr_expense

remi-filament:
hr_expense module migration for V10

Current behavior before PR:
Migration of hr_expense module not initialized

Migration of hr_expense module is performed

--


addons/hr_expense/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,114 @@
...
+    cr.execute(
+	'''UPDATE hr_expense_sheet
+	SET state='submit'
+	WHERE state='draft'
+	''')
+
+    # Set correct account corresponding to product subject to this expense 
+    cr.execute("SELECT distinct he.product_id, pp.product_tmpl_id from hr_expense he, product_product pp where he.product_id=pp.id")
+    product_ids = cr.fetchall()
+    for product in product_ids:
```
StefanRijnhart: This can go wrong in several ways. How about making this company aware by fetching the product's account through the ORM. If you browse the product with *force_company*=company.id in the context, you get the correct account automatically that you can write on the expenses. Update the expenses per product per company in one big sweep to prevent all too many lookups or write operations.

addons/hr_expense/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,114 @@
...
+    # Set accounting_date to date from account_analytic_line when it exists
+    cr.execute(
+	'''UPDATE hr_expense_sheet
+	SET accounting_date = subquery.date
+	FROM (
+            SELECT
+                am.id as id, am.date as date
+            FROM account_move am
+        ) AS subquery
+        WHERE hr_expense_sheet.account_move_id = subquery.id
```
StefanRijnhart: More readable, if not faster:
```
UPDATE hr_expense_sheet hes
SET accounting_date = am.date
FROM account_move am
WHERE hes.account_move_id = am.id;
```

addons/hr_expense/migrations/10.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,21 @@
...
+        ('department_id', None, None),
+        ('journal_id', None, None),
+    ],
+}
+
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, column_copies)
```
StefanRijnhart: Is this necessary? We like to preserve data in OpenUpgrade, but you are literally copying the data to the related sheet already in the post script.

addons/hr_expense/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,114 @@
...
+	'''UPDATE hr_expense
+	SET state='refused'
+	WHERE state='cancel'
+	''')
+  
+    # Set correct submit state for hr_expense_sheet since draft does not exist anymore for sheet
+    cr.execute(
+	'''UPDATE hr_expense_sheet
+	SET state='submit'
+	WHERE state='draft'
```
StefanRijnhart: Glancing over the code in the Odoo module, it seems that hr.expense records in draft state can stay that way, and don't need a sheet because submitting the expense will create the sheet: https://github.com/odoo/odoo/blob/10.0/addons/hr_expense/models/hr_expense.py#L103

addons/hr_expense/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,114 @@
...
+	    FROM hr_department
+	) AS subquery
+	WHERE department_id = subquery.id
+	''')
+
+    # Set correct reported state for hr_expense
+    cr.execute(
+	'''UPDATE hr_expense
+	SET state='reported'
+	WHERE state='submit' or state='approved'
```
StefanRijnhart: New state *approve* for old state *approved*, I think.

addons/hr_expense/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,114 @@
...
+	''')
+
+    # Set correct done state for hr_expense
+    cr.execute(
+	'''UPDATE hr_expense
+	SET state='done'
+	WHERE state='post' or state='done'
```
StefanRijnhart: But the *post* state is still possible, isn't it?

addons/hr_expense/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+hr_expense   / hr.expense.sheet         / responsible_id (many2one)     : NEW relation: res.users
+# Done - Responsible ID set by default to department_id manager
+
+hr_expense   / hr.expense               / state (selection)             : now a function
+hr_expense   / hr.expense               / state (selection)             : selection_keys is now '['done', 'draft', 'refused', 'reported']' ('['approve', 'cancel', 'done', 'draft', 'post', 'submit']')
+hr_expense   / hr.expense.sheet         / state (selection)             : NEW required: required, selection_keys: ['approve', 'cancel', 'done', 'post', 'submit'], req_default: function
+# Done - Update states to fit with new definition
+
+hr_expense   / hr.expense               / website_message_ids (one2many): DEL relation: mail.message
+# Not present if no website..., I do not know what should be done here...
```
StefanRijnhart: You can ignore this. You can write a comment # Nothing to do.

StefanRijnhart:
Hi @remi-filament, thank you for your contribution. I have some comments and I am looking forward to your follow-up. Did you send your contribution-license-agreement to the OCA already? Let me know if you have any questions about all that.

remi-filament:
Hi @StefanRijnhart , thank you for your comments, really useful !
I will try to answer them one by one.

addons/hr_expense/migrations/10.0.2.0/pre-migration.py
remi-filament: You are right, I was not sure whether the lines would not have disappeared before running post-migration, which is why I wanted to make sure they were not lost.
From your comment I understand that this is not the case so I will drop this pre-migration file.

addons/hr_expense/migrations/10.0.2.0/post-migration.py
remi-filament: Initially I tried to call a function to get the account associated to each product from the product addon (which retrieves it from ir_property table) but did not manage to figure out how to do that...
That would have been much cleaner I agree.

Regarding the specifics of your comment, I am not sure about what you propose since the account is not linked to the company, there are so many accounts per company. However the account is linked to the product template and therefore should be retrieved from there in my view...

remi-filament:
@StefanRijnhart I have reviewed your comments in details, and I think we need to clarify the new states that are now in use for hr_expense modules.
As a matter of fact, a new model has been created, called hr_expense_sheet and fields have been moved between hr_expense and hr_expense_sheet.
As I understood it, hr_expense covers each line of expense now when hr_expense_sheet covers the overall expense from employee (covering all the lines described in hr_expense).

Therefore the products and accounts are defined in hr_expense still when the accounting information are covered in hr_expense_sheet now.

The states have also changed accordingly. These were the states present before on hr_expense in V9:
- draft
- submit
- approve
- post
- done
- cancel

And we now have on hr_expense:
- draft --> expense is registered but not yet linked to a hr_expense_sheet (can be kept from former state = draft)
- reported --> expense has been submitted for approval / reimbursement (I have then mapped submit and approve to this new state)
- done --> expense has been treated in terms of accountability (whether it has been reimbursed or not is not managed anymore at hr_expense level but at hr_expense_sheet now) (I have then mapped former post and done states to this one)
- refused --> expense has been refused by responsible (mapped from former cancel state)

And on hr_expense_sheet:
- submit
- approve
- post
- done
- cancel
We therefore have the same states as before on hr_expense except for draft one, that I mapped to submit. Maybe I should modify the command creating the sheet to create only for those lines not in draft...

This is however my undestanding of the changes performed, explaining the mapping I tried to make. Please let me know your view.
Thanks,
Rémi

StefanRijnhart:
Thanks! merging this, so the work can continue on https://github.com/OCA/OpenUpgrade/pull/1130

PR #764, opened 2017-02-16
[UPD] upgrade this script(compare_noupdate_xml_records.py) to be compatible with 10.0 and small bugs of module: openupgrade_records

duanyp1991:
Changes to be committed:
	modified:   scripts/compare_noupdate_xml_records.py

Description of the issue/feature this PR addresses:

Current behavior before PR:

Desired behavior after PR is merged:

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #1008, opened 2017-08-09
10.0 compare noupdate xml records

pedrobaeza:
Superseed of #764


pedrobaeza:
Everything is correct now with also my review on the previous commit, so I merge optimistically

PR #1009, opened 2017-08-09
[IMP] Migrate fields.Binary to attachments automatically

### LINKED ISSUE #1007: [9.0/10.0] openupgrade_records: Detect changes in binary field attachment argument
Impacted versions: 9.0/10.0

openupgrade_records doesn't log changes in the attribute `attachment` of binary fields, which is important, as this change requires to make a process on the migration script.

Example:
9.0: https://github.com/odoo/odoo/blob/a31820d0cb780108ff1f4a0d322ca8e3bd180da3/addons/website_slides/models/slides.py#L282
10.0: https://github.com/odoo/odoo/blob/64190abec53847027f4db66950bf8c164dc74d74/addons/website_slides/models/slides.py#L311

I looked at this manually on 9.0 and match almost all the occurrences (see https://github.com/OCA/OpenUpgrade/pull/745), and I expect not to be more on 10.0, but it's not the case, so we should include this check.

I'm adding it, but do you have a viable full OpenUpgrade installation done for performing the analysis quickly, @hbrunn @StefanRijnhart @jbeficent @mreficent ?

StefanRijnhart:
Closes #1007

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/post-migration.py
@@ -5,7 +5,6 @@
 
 import logging
 from openupgradelib import openupgrade
-from openupgradelib import openupgrade_90
 
 
 logger = logging.getLogger('OpenUpgrade')
@@ -15,20 +14,6 @@ logger = logging.getLogger('OpenUpgrade')
 column_copies = {
     'ir_actions': [
         ('help', None, None),
-    ],
-}
-
-attachment_fields = {
-    'res.partner': [
-        ('image', None),
-        ('image_medium', None),
-        ('image_small', None),
-    ],
-    'res.country': [
-        ('image', None),
-    ],
-    'ir.ui.menu': [
-        ('web_icon_data', None),
     ],
 }
 
@@ -98,4 +83,3 @@ def migrate(env, version):
         env.cr, 'base', 'migrations/9.0.1.3/noupdate_changes.xml',
     )
     assign_view_keys(env)
-    openupgrade_90.convert_binary_field_to_attachment(env, attachment_fields)
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -23,17 +23,6 @@ column_renames = {
     'res_partner_bank': [
         ('bank', 'bank_id'),
     ],
-    'res_partner': [
-        ('image', None),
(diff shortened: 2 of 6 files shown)


pedrobaeza:
The problem here is that the content of the field is erased AFAIK.

StefanRijnhart:
Yes it looks like it, doesn't it? Locally the attachments are there but the test still fails. Hold on...

pedrobaeza:
OK, thanks for the investigation. Indeed this method is better. We should also apply this in 10.0, 11.0...

StefanRijnhart:
Oh no, I expect tests to be green now, but I get the dreaded 'fatal: Could not parse object error' from 'git reset'

pedrobaeza:
The solution is to close the PR and open it again. It's a problem with GitHub + Travis. It happens when you rebase, or force push... I don't wait directly in one of that cases, and cancel the current build, close and open again the PR.

StefanRijnhart:
Thanks for showing me the workaround! About the code: I tested for the presence of the attachments in the base module, but of course the deferred step was not yet executed at that point. I added a deferred test script, and (after moving it until after the commit of the deferred step ;-) after that I could successfully test the code from the deferred step.

pedrobaeza:
There are a lot more places where I used the `convert_binary_field_to_attachment`. I understand that you should remove them as well for not having duplicate attachments. You can see them on the corresponding PR: https://github.com/OCA/OpenUpgrade/pull/745/files

StefanRijnhart:
@pedrobaeza I don't care about them. When a field was migrated already, this new attempt should just be harmless.

pedrobaeza:
OK, I will check that and forward port it to 10.0 (or are you already with it?) On v10, we don't need deferred method, but put and `end-migration.py` on base.

StefanRijnhart:
Yes, I was on it. Thanks for the reminder of end-migration.py

StefanRijnhart:
I was thinking for 11, we should put a check in the analysis so that we know if we need to port this further or not. @pedrobaeza you were working on that, can you share the code that keeps track of the attachment attribute on binary fields for 10.0? I'd like to think about a mechanism that allows us to ignore this flag in the 9 to 10 analysis (to prevent useless backporting) but not in the 10 to 11 analysis.

pedrobaeza:
Yes, here you have it: https://github.com/OCA/OpenUpgrade/pull/1014

The most difficult part was to understand how this works as it was the first time I dealt with this module.

pedrobaeza (approved):

PR #1010, opened 2017-08-09
[FIX] analytic: Remove analytic account of type='view'

### LINKED ISSUE #578: [9.0][analytic] Analytic accounts of type='view' should be deleted
Impacted versions: 9.0

Steps to reproduce:
Migrate a database in 8.0 with module 'analytic' installed.

Current behavior:
The analytic accounts of type 'view' are migrated to analytic accounts with tag 'view'.

Expected behavior:
These analytic accounts will not be used anymore in v9. 

Instead, the child analytic accounts may incorporate a tag with the name of the old analytic account.
In case of parent analytic accounts that are not of type 'view' a tag should also be created and assigned downwards to all the children down the hierarchy.

V8:
image

V9:
image

pedrobaeza:
Closes #578

### DIFF
--- a/addons/analytic/migrations/9.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/analytic/migrations/9.0.1.1/openupgrade_analysis_work.txt
@@ -39,7 +39,8 @@ analytic     / account.analytic.account / template_id (many2one)        : DEL re
 #Nothing to do : No longer needed.
 
 analytic     / account.analytic.account / type (selection)              : DEL required: required, selection_keys: ['contract', 'normal', 'template', 'view'], req_default: normal
-# Converted to tags
+# Done: Hide view accounts
+# Done: Converted to tags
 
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
@pedrobaeza I mentioned this before, I believe. As what Hintjens has said about optimistic merging, I would prefer that people do not merge their own PRs. Can you agree with that?

pedrobaeza:
Sorry, but there has been around no other people with merge possibilities all this time. I needed to contact specifically Jordi last week and disturb him for pressing the Merge button.

There's no sense on that. If there's a problem, then make a PR solving it or put an issue and don't worry that I will solve it very quickly, but we can't have 30 open PRs as last week.

pedrobaeza:
And as this is a continuous chain, there are inter-dependencies between the PRs. For example, for this one, we have https://github.com/OCA/account-analytic/pull/126, and the same was for a lot of other modules even inside OpenUpgrade.

PR #1013, opened 2017-08-09
[10.0] binary fields to attachment deferred

StefanRijnhart:
Port of https://github.com/OCA/OpenUpgrade/pull/1009


pedrobaeza (approved):
I confirm this works, and it seems that only `datas` field of `slide.slide` model have changed to `attachment=True`.

StefanRijnhart:
There is also field cache from model pos.cache. Anyway, green now.

StefanRijnhart:
Let me add a protection to allow tests_deferred not to break tests in which website_slides is not installed.

pedrobaeza:
Right, in my test DB I hadn't installed that module.  Merging when Travis finishes.

PR #1014, opened 2017-08-09
[IMP] openupgrade_records: Log attachment property on binary fields

pedrobaeza:
cc @Tecnativa @StefanRijnhart

### DIFF
--- a/odoo/addons/openupgrade_records/models/openupgrade_record.py
+++ b/odoo/addons/openupgrade_records/models/openupgrade_record.py
@@ -37,6 +37,7 @@ class Record(models.Model):
     @api.model
     def field_dump(self):
         keys = [
+            'attachment',
             'module',
             'mode',
             'model',
--- a/odoo/openupgrade/openupgrade_loading.py
+++ b/odoo/openupgrade/openupgrade_loading.py
@@ -157,6 +157,8 @@ def log_model(model, local_registry):
                     sorted([x[0] for x in v.selection]))
             else:
                 properties['selection_keys'] = 'function'
+        elif v.type == 'binary':
+            properties['attachment'] = str(getattr(v, "attachment", False))
         default = model._fields[k].default
         if v.required and default:
             if isinstance(default, types.FunctionType):


odoo/openupgrade/openupgrade_loading.py
```diff
@@ -157,6 +157,8 @@ def isrelated(model, k):
                     sorted([x[0] for x in v.selection]))
             else:
                 properties['selection_keys'] = 'function'
+        elif v.type == 'binary':
+            properties['attachment'] = str(getattr(v, "attachment", False))
```
pedrobaeza, line 161: If I understand well the mechanism, current code will require to include the same fragment in `openupgrade_records` v9. One variation for not requiring it would be:
```
        elif v.type == 'binary':
            if getattr(v, "attachment", False):
                properties['attachment'] = "True"
```
and this will generate the text `NEW attachment=True` or similar. Isn't it?

StefanRijnhart: Ah, I see that compare.py is not shared in openupgradelib, but instead kept locally in odoo/addons/openupgrade_records/lib/. So we can include this here and then add the 'attachment' key in Odoo 11 to make it show up in the analysis files.

compare.py does however assume that keys are present for every field, not just binary fields. So you should add a default in the properties dictionary right above.

pedrobaeza: Seeing the code of `compare.py`, we only need to keep the same set of properties by field on both v10/v11, but it's not needed to have `attachment=False` on the rest of the field types. That's why I put this code and not a `False` value on the default dict.

StefanRijnhart: compare.py is not field type aware, but you are right in the sense that you may also add the key here, so that the default value is set to False (probably a better choice, too): https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/openupgrade_records/models/openupgrade_record.py#L39

pedrobaeza: Then I don't get what do you want me to do? I let it as is? I think it's better to not need to propagate this value to the rest of the type of fields.

StefanRijnhart: Please add the attribute here: https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/openupgrade_records/models/openupgrade_record.py#L39. It is safe to do so. The value of the field will not be compared and reported until we add the field to these lists here (which we should do in OpenUpgrade 11): https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/openupgrade_records/lib/compare.py#L190

pedrobaeza:
Changed according your comments. You can squash on merge.

StefanRijnhart:
@pedrobaeza Sorry about the confusion, but I changed my mind about the best place for this attribute during the discussion. If you add it here https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/openupgrade_records/models/openupgrade_record.py#L39 instead, it saves us one database record for every field that is not a binary attachment with attachment=True

pedrobaeza:
@StefanRijnhart But we need to fill also the values that have `attachment=True` for v10. I think with the last change you propose + my initial commit filling only attachment=True records is everything we need. Should I proceed this way?

StefanRijnhart:
Yes, I agree. Only register the property when it is True as you suggested in https://github.com/OCA/OpenUpgrade/pull/1014/commits/a753b17345c231cc1109daab156b4dd50032b4df#r132307670, and then set a default value when comparing records (https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/openupgrade_records/models/openupgrade_record.py#L39), not when storing records in the database.

pedrobaeza:
@StefanRijnhart I have changed the PR according your comments. Please review.

odoo/addons/openupgrade_records/models/openupgrade_record.py
```diff
@@ -40,6 +40,7 @@ def field_dump(self):
             'module',
             'mode',
             'model',
+            'no_update',
```
StefanRijnhart: `attachment` ;-)

pedrobaeza: hahaha, this is what happens when we mix late hours, topics, and so on, changing...

pedrobaeza:
Done, but I suppose that this is going to finish with error due to Travis commit glitch.

pedrobaeza:
Green! We can merge then.

StefanRijnhart:
Thanks! I will now implement the 11.0 part

PR #1015, opened 2017-08-11
[FIX][8.0] point_of_sale migration (picking type)

legalsylvain:

stock picking type from pos config are bad created and unused.

- stock_picking_type has two bad fields. (default_location_src_id and default_location_dest_id)
- stock_picking created before the migration are not affected to the new stock_picking_type created.

--

CC : @kirca, @StefanRijnhart, @pedrobaeza

### DIFF
--- a/addons/point_of_sale/migrations/8.0.1.0.1/post-migration.py
+++ b/addons/point_of_sale/migrations/8.0.1.0.1/post-migration.py
@@ -101,7 +101,7 @@ def get_pricelist_id(cr, pool, shop_id, pc):
     return pricelist_id[0]
 
 
-def get_picking_type_id(cr, pool, shop_id, pc):
+def get_picking_type_id(cr, pool, shop_id, pc, stock_loc_id):
     """
         Create a new picking_type_id using shop info
     """
@@ -114,9 +114,8 @@ def get_picking_type_id(cr, pool, shop_id, pc):
             cr, SUPERUSER_ID, 'point_of_sale.seq_picking_type_posout'),
         'code': 'outgoing',
         'warehouse_id': wh.id,
-        'default_location_src_id': wh.wh_output_stock_loc_id.id,
-        'default_location_dest_id': pool['ir.model.data'].xmlid_to_res_id(
-            cr, SUPERUSER_ID, 'stock.stock_location_customers'),
+        'default_location_src_id': stock_loc_id,
+        'default_location_dest_id': wh.wh_output_stock_loc_id.id,
     })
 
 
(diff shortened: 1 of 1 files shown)


addons/point_of_sale/migrations/8.0.1.0.1/post-migration.py
```diff
@@ -114,9 +114,8 @@ def get_picking_type_id(cr, pool, shop_id, pc):
             cr, SUPERUSER_ID, 'point_of_sale.seq_picking_type_posout'),
         'code': 'outgoing',
         'warehouse_id': wh.id,
-        'default_location_src_id': wh.wh_output_stock_loc_id.id,
-        'default_location_dest_id': pool['ir.model.data'].xmlid_to_res_id(
-            cr, SUPERUSER_ID, 'stock.stock_location_customers'),
+        'default_location_src_id': stock_loc_id,
+        'default_location_dest_id': wh.wh_output_stock_loc_id.id,
```
StefanRijnhart, line 118: I agree with the modification of the source location, but are you sure you don't want to put a customer location in the destination here?

legalsylvain: Hi @StefanRijnhart, thanks for taking a look on that patch.

About default_location_dest_id, for the time being, it uses allways the same (got by xml_id), that doesn't work in a multicompany context.
I so selected wh_output_stock_loc_id of the current warehouse, as it is the default 'Output Location' of the warehouse. So it is a customer location... Or did I missed something ? 
If yes, what do you propose to use instead ? 

regards.

StefanRijnhart: The output location could be a pack location, which is wrong for the POS destination. Fetch the warehouse specific customer location using something like this: https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L415 (with usage='customer')

StefanRijnhart: @legalsylvain I'm guessing you don't agree? Is there anyone else with an opinion on this?

legalsylvain: No, I just fully forgot this PR ! Holidays are killing peer reviews. ;-)
I'm fully agree with your PoV. (even if I personally don't use pack location.)

I will not have time to implement your proposal immediately. but I'll review and accept any PR against grap:8.0_FIX_migration_pos_config. 

regards.

pedrobaeza:
Are anyone going to continue with this PR?

legalsylvain:
I personnaly migrated on 8.0 so I'll not work anymore on that branch we can keep open if any people are interested to finish / change the work. Otherwise, not a problem to close this PR.
regards.

kirca:
Not working on it also, although I have migrated already with it.

pedrobaeza:
Closing as no further work. Anybody interested can rescue this work completing it with the comments.

PR #1018, opened 2017-08-14
[9.0][ADD] portal_stock: Migration to 9.0

cubells:
- Checked noupdate records
 

Nothing to do


addons/portal_stock/migrations/9.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,9 @@
+---Fields in module 'portal_stock'---
+
+---XML records in module 'portal_stock'---
+NEW ir.model.access: portal_stock.access_stock_pack_operation_portal
+NEW ir.model.access: portal_stock.access_stock_picking_type_portal
+NEW ir.rule: portal_stock.portal_stock_pack_operation_portal_user_rule
+NEW ir.rule: portal_stock.portal_stock_picking_portal_user_rule
+DEL ir.rule: portal_stock.portal_stock_picking_user_rule
```
pedrobaeza, line 8: The rule is noupdate. You should delete it.

pedrobaeza (changes requested):

MiquelRForgeFlow (approved):

addons/portal_stock/migrations/9.0.0.1/openupgrade_analysis_work.txt
cubells: It's done.

pedrobaeza (approved):

PR #1021, opened 2017-08-14
[9.0][MIG] portal_claim: Renamed to website_crm_claim

cubells:
Features are merged in crm_claim module


openerp/addons/base/migrations/9.0.1.3/pre-migration.py
```diff
@@ -49,6 +49,7 @@ def cleanup_modules(cr):
             ('contacts', 'mail'),
             ('marketing_crm', 'crm'),
             ('email_template', 'mail'),  # mail_template class
+            ('portal_claim', 'crm_claim'),
```
pedrobaeza: Looking at the code at 8.0 and 9.0, it's better to just rename (in apriori.py) `portal_claim` to `website_crm_claim`

pedrobaeza (changes requested):

addons/crm_claim/migrations/9.0.1.0/pre-migration.py
```diff
@@ -17,7 +17,15 @@
...
+    openupgrade.rename_columns(env.cr, column_renames)
+    openupgrade.rename_tables(env.cr, table_renames)
+    # Portal_claim is merged here and It is needed rename or delete ir.rule
+    if openupgrade.is_module_installed(env.cr, 'website_crm_claim'):
+        openupgrade.rename_xmlids(env.cr, [
+            ('portal_claim.portal_claim_rule',
+             'website_crm_claim.portal_claim_rule'),
+        ])
+    else:
+        env.ref('portal_claim.portal_claim_rule').unlink()
```
MiquelRForgeFlow: Delete this `else` because, in this case, there would be nothing to unlink

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow:
And please, update documentation

addons/crm_claim/migrations/9.0.1.0/pre-migration.py
```diff
@@ -17,7 +17,13 @@
...
 
-@openupgrade.migrate()
-def migrate(cr, version):
-    openupgrade.rename_columns(cr, column_renames)
-    openupgrade.rename_tables(cr, table_renames)
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    openupgrade.rename_columns(env.cr, column_renames)
+    openupgrade.rename_tables(env.cr, table_renames)
+    # Portal_claim is merged here and It is needed rename xml-id
```
pedrobaeza: This is not needed anymore making the renaming in the other place. Please restore it as it was.

addons/crm_claim/migrations/9.0.1.0/pre-migration.py
```diff
@@ -17,7 +17,12 @@
...
 
-@openupgrade.migrate()
-def migrate(cr, version):
-    openupgrade.rename_columns(cr, column_renames)
-    openupgrade.rename_tables(cr, table_renames)
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    openupgrade.rename_columns(env.cr, column_renames)
+    openupgrade.rename_tables(env.cr, table_renames)
+    if openupgrade.is_module_installed(env.cr, 'website_crm_claim'):
```
pedrobaeza: I'm talking about leaving this as it was before this PR. You don't need to do this with the renaming of the module.

MiquelRForgeFlow: True :+1:

addons/crm_claim/migrations/9.0.1.0/pre-migration.py
```diff
@@ -17,7 +17,7 @@
 ]
 
 
-@openupgrade.migrate()
-def migrate(cr, version):
-    openupgrade.rename_columns(cr, column_renames)
-    openupgrade.rename_tables(cr, table_renames)
+@openupgrade.migrate(use_env=True)
```
pedrobaeza: You're still modifying this script. Please restore it as it was for making the diff smaller.

cubells:
You are right @pedrobaeza 

Sorry.

pedrobaeza:
Agh, I miss to say to you about the documentation... Please put `Done - Renamed to website_crm_claim`

pedrobaeza:
And you have to fill also `website_crm_claim` line with `Done - Renamed from portal_claim`

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #1024, opened 2017-08-15
[FIX] don't break after writing the first cover image

hbrunn:
...and don't make the overlay div opaque

### DIFF
--- a/addons/website_blog/migrations/9.0.1.0/post-migration.py
+++ b/addons/website_blog/migrations/9.0.1.0/post-migration.py
@@ -31,7 +31,6 @@ def convert_blog_post_cover(env):
             ' "background-color": "oe_none",'
             ' "resize_class": "cover cover_full"}}'
         ).format(row[1])
-        break
 
 
 @openupgrade.migrate(use_env=True)
--- a/addons/website_blog/migrations/9.0.1.0/post-migration.py
+++ b/addons/website_blog/migrations/9.0.1.0/post-migration.py
@@ -27,7 +27,7 @@ def convert_blog_post_cover(env):
         post = post_obj.browse(row[0])
         post.cover_properties = (
             '{{"background-image": "url({})",'
-            ' "opacity": "1",'
+            ' "opacity": "0",'
             ' "background-color": "oe_none",'
             ' "resize_class": "cover cover_full"}}'
         ).format(row[1])


addons/website_blog/migrations/9.0.1.0/post-migration.py
```diff
@@ -27,11 +27,10 @@ def convert_blog_post_cover(env):
         post = post_obj.browse(row[0])
         post.cover_properties = (
             '{{"background-image": "url({})",'
-            ' "opacity": "1",'
+            ' "opacity": "0",'
```
pedrobaeza, line 30: If this doesn't have `opacity`, then the image won't be shown. This was copied from the style given by Odoo.

addons/website_blog/migrations/9.0.1.0/post-migration.py
```diff
@@ -27,11 +27,10 @@ def convert_blog_post_cover(env):
...
             ' "background-color": "oe_none",'
             ' "resize_class": "cover cover_full"}}'
         ).format(row[1])
-        break
```
pedrobaeza, line 34: Why I put this `break`??!! Maybe a debug trick...? I'm crazy :stuck_out_tongue_closed_eyes:

addons/website_blog/migrations/9.0.1.0/post-migration.py
hbrunn: in 10, it's the other way around. 10 creates a div that overlays the image, and opacity 1 combined with color none (=white) will give you just blank: https://github.com/OCA/OpenUpgrade/blob/10.0/addons/website_blog/views/website_blog_templates.xml#L185
Hmmm, change it back for 9 just to have to change it back again for 10. Would be cleaner...

pedrobaeza: Yeah, we should add another migration script in website_blog for v10, as not everybody makes v8 > v10 directly.

pedrobaeza:
Any news on this, Holger?

pedrobaeza:
@hbrunn can you leave this PR without the break and then we merge, and handle the other thing later in another PR (to v10)?

pedrobaeza:
I have made a direct commit for the removal of the break for unlocking this in https://github.com/OCA/OpenUpgrade/commit/a9f1eaa0ba92ab6f160aeb7d29f7a774e932637c. And I have also proposed the opacity meaning change in 10.0 in the PR #1091, so this one can be closed.

PR #1003, opened 2017-08-05
[ADD] hr_attendance: Migration scripts

pedrobaeza:
* Checked noupdate records
* Analysis work file
* Updated documentation
* Migration scripts

PR #1027, opened 2017-08-15
[FIX] don't assign purchase price lists to websites

hbrunn:
Odoo will grab the first price list available, and we don't want to have purchase price lists slipped in here


pedrobaeza:
On v9/v10, there are no purchase pricelists. You should treat this with another technique on v9, disabling purchase pricelists there if possible, and if not, doing as I did with attendances of type 'action' (which don't have equivalent on new version (copy them to another table). See https://github.com/OCA/OpenUpgrade/pull/1003/files#diff-1d6e98c42c455785122a139ed675bc5aR20

pedrobaeza:
Any news on this, Holger?

hbrunn:
yes, I think we're better off with https://github.com/OCA/OpenUpgrade/pull/1072

PR #1029, opened 2017-08-16
[9.0][MIG] portal: migration scripts

cubells:
- Rename access rules xml-ids.
- Checked noupdate records


addons/portal/migrations/9.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,22 @@
...
+DEL ir.model.access: portal.access_mail_notification_portal
+DEL ir.ui.menu: portal.portal_inbox
+DEL ir.ui.menu: portal.portal_mail_archivesfeeds
+DEL ir.ui.menu: portal.portal_mail_starfeeds
+DEL ir.ui.menu: portal.portal_messages
+DEL ir.ui.view: portal.assets_backend
+DEL ir.ui.view: portal.share_step1_form_portal
+DEL ir.ui.view: portal.share_step2_form_portal
+DEL ir.ui.view: portal.wizard_user_tree_view
+NEW mail.template: portal.mail_template_data_portal_welcome
```
pedrobaeza: You haven't put any comment here.

addons/portal/migrations/9.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.rename_xmlids(cr, [
```
pedrobaeza: As they are noupdate="0" records (all CSV records are this mode), and these records are not assigned manually to anything, you don't need to rename them, so put it as "Nothing to do" instead and remove this script.

pedrobaeza (changes requested):

openerp/openupgrade/doc/source/modules80-90.rst
```diff
@@ -406,7 +406,7 @@ missing in the new release are marked with |del|.
 +----------------------------------------+------------------------------------------+
 |point_of_sale                           |                                          |
 +----------------------------------------+------------------------------------------+
-|portal                                  |                                          |
+|portal                                  | Done                                     |
```
MiquelRForgeFlow: Should be `Nothing to do` if nothing is done

MiquelRForgeFlow (changes requested):

cubells:
@mreficent done!

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #1032, opened 2017-08-17
[10.0][IMP] compare_noupdate_xml_records: analyze repositories

MiquelRForgeFlow:
This PR introduce a third optional argument, the mode, that is `module` or `repository`:
* The 'module' option does the usual analysis between two modules. In this case, the two first arguments are the module paths. Default option.
* The 'repository' option compares all the modules that have the same name in both repositories. In this case, the two first arguments are the repository paths.

### DIFF
--- a/scripts/compare_noupdate_xml_records.py
+++ b/scripts/compare_noupdate_xml_records.py
@@ -3,6 +3,7 @@
 # Copyright 2016 Opener B.V. <https://opener.am>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from os.path import join as opj
 import argparse
 import ast
 import os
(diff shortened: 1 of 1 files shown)


scripts/compare_noupdate_xml_records.py
```diff
@@ -3,6 +3,8 @@
 # Copyright 2016 Opener B.V. <https://opener.am>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from odoo.modules.module import MANIFEST_NAMES
```
pedrobaeza: This import is not valid.

scripts/compare_noupdate_xml_records.py
```diff
@@ -211,5 +182,76 @@ def main(argv=None):
...
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'olddir', metavar='older_directory')
+    parser.add_argument(
+        'newdir', metavar='newer_directory')
+    parser.add_argument(
+        'mode', metavar='module/repository')
+    arguments = parser.parse_args(argv)
+
```
pedrobaeza: Don't let these blank lines. They are IMO very anti-aesthetic and makes harder to read method lines together

pedrobaeza (changes requested):

scripts/compare_noupdate_xml_records.py
MiquelRForgeFlow: Deleted all blank lines 😅

MiquelRForgeFlow:
@pedrobaeza, is it ok now? 🤓

pedrobaeza:
I have to re-check. I'll tell you

scripts/compare_noupdate_xml_records.py
```diff
@@ -211,5 +181,68 @@ def main(argv=None):
...
+    - Does not handle the shorthand records <menu>, <act_window> etc.,
+      although that could be done using the same expansion logic as
+      is used in their parsers in openerp/tools/convert.py
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'olddir', metavar='older_directory')
+    parser.add_argument(
+        'newdir', metavar='newer_directory')
+    parser.add_argument(
```
pedrobaeza, line 216: Can you make this argument optional and default to `module` for having the same command line as previously?

MiquelRForgeFlow: Done :+1:

pedrobaeza (approved):

PR #1038, opened 2017-08-21
Merge connector into queue_job

cubells:
These changes are needed in connector migration from 9.0 to 10.0

cc @Tecnativa

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -3,6 +3,11 @@ to help the matching process
 """
 
 renamed_modules = {
+    # OCA/connector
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
Why a rename and not a module merge in the base pre-migration? 🤔
And what is the need of it? The module https://github.com/OCA/connector/tree/10.0/connector is still `installable=True`!

pedrobaeza:
See explanation here: https://github.com/OCA/connector/pull/256#issuecomment-323327247

pedrobaeza:
@cubells please include a bit of the explanation as a comment in the code

pedrobaeza (approved):

PR #1042, opened 2017-08-22
[FIX] calendar: Load noupdate=1 records

pedrobaeza:


yajo (approved):

rafaelbn (approved):

MiquelRForgeFlow (commented):
Could you update the <code>analysis_work.txt</code>?

pedrobaeza:
> Could you update the analysis_work.txt?

No, as they are not reflected there. But the problem is another: `compare_noupdate_xml_records.py` script is not properly working...  I'll submit a PR soon. Please stay tuned.

MiquelRForgeFlow (approved):

PR #1044, opened 2017-08-24
[FIX][8.0] product_price_history initialization

legalsylvain:

in a multicompany context, after a migration from 7.0 to 8.0, all items in product_price_history has the same company_id as the current company of the super user. That is due to this line.
https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/post-migration.py#L71

if a product doesn't belong to this company, the history will be bad.

the PR execute a extra SQL request to alter history and affect the company, according to the company of the template.

Note : It is a partial fix. I don't know how to manage products that doesn't have a company. (global product) because : 
* product_price_history.company_id > required = True
* product_template.company_id > required = False

regards.

--

### DIFF
--- a/addons/product/migrations/8.0.1.1/post-migration.py
+++ b/addons/product/migrations/8.0.1.1/post-migration.py
@@ -78,6 +78,24 @@ def create_properties(cr, pool):
         AND name = 'standard_price'""")
     openupgrade.logged_query(cr, sql)
 
+    # Affect history to the template's company
+    # Note, for the time being, in a multi company context the history will
+    # be incomplete for products  that don't belong to a given company
+    # (Global products), because the history will be associated to the
+    # company of the SUPERUSER, during the migration
+    cr.execute("""
+        UPDATE product_price_history pph
+        SET company_id = template_company.company_id
+        FROM (
+            SELECT
+                pt.id template_id,
+                pt.company_id
+            FROM product_template pt
+            WHERE company_id is not null
+        ) template_company
+        WHERE pph.product_template_id = template_company.template_id
+    """)
+
     # product.price.history entries have been generated with a value for
     # today, we want a value for the past as well, write a bogus date to
     # be sure that we have an historic value whenever we want
--- a/addons/product/migrations/8.0.1.1/post-migration.py
+++ b/addons/product/migrations/8.0.1.1/post-migration.py
@@ -85,22 +85,16 @@ def create_properties(cr, pool):
     # company of the SUPERUSER, during the migration
     cr.execute("""
         UPDATE product_price_history pph
-        SET company_id = template_company.company_id
-        FROM (
-            SELECT
-                pt.id template_id,
-                pt.company_id
-            FROM product_template pt
-            WHERE company_id is not null
-        ) template_company
-        WHERE pph.product_template_id = template_company.template_id
+        SET company_id = product_template.company_id
+        FROM product_template
+        WHERE pph.product_template_id = product_template.id
+        AND product_template.company_id IS NOT NULL
     """)
 
     # product.price.history entries have been generated with a value for
     # today, we want a value for the past as well, write a bogus date to
     # be sure that we have an historic value whenever we want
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):
LGTM :+1:

addons/product/migrations/8.0.1.1/post-migration.py
```diff
@@ -78,6 +78,24 @@ def create_properties(cr, pool):
         AND name = 'standard_price'""")
     openupgrade.logged_query(cr, sql)
 
+    # Affect history to the template's company
+    # Note, for the time being, in a multi company context the history will
+    # be incomplete for products  that don't belong to a given company
+    # (Global products), because the history will be associated to the
+    # company of the SUPERUSER, during the migration
+    cr.execute("""
+        UPDATE product_price_history pph
```
pedrobaeza, line 87: This can be done without subquery:
```
UPDATE product_price_history pph
SET company_id = product_template.company_id
FROM product_template
WHERE pph.product_template_id = product_template.template_id
AND product_template.company_id IS NOT NULL
```

pedrobaeza (changes requested):
Can you point in the analysis work which field change motivates this extra query?

I also think that the method `create_properties` is not the best place for adding this.

addons/product/migrations/8.0.1.1/post-migration.py
legalsylvain: Hi @pedrobaeza, 

thanks for your review.

> Can you point in the analysis work which field change motivates this extra query?

this is a new table, populated automaticaly by ORM, when creating properties (here : https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/post-migration.py#L71)


For that reason, in the xxx_work file, there is "nothing to do" (https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/openupgrade_analysis_WORK.txt#L36), but there are some things to fix ;-)

> I also think that the method create_properties is not the best place for adding this.

This is the right place to write this extra SQL query. There is another query to fix datetime, just after the code I added. ref : https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/post-migration.py#L84

> This can be done without subquery

You're right, I'll change that. Thanks.

pedrobaeza:
> this is a new table, populated automaticaly by ORM, when creating properties (here : https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/post-migration.py#L71)

Why don't you assign company there or make that Odoo itself writes the company there?

> For that reason, in the xxx_work file, there is "nothing to do" (https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/openupgrade_analysis_WORK.txt#L36), but there are some things to fix ;-)

And that's why I'm asking, because you should change also this file changing the text to something similar to `Done: price history table created and first record populated from product prices`

> I also think that the method create_properties is not the best place for adding this.
> > This is the right place to write this extra SQL query. There is another query to fix datetime, just after the code I added. ref : https://github.com/OCA/OpenUpgrade/blob/8.0/addons/product/migrations/8.0.1.1/post-migration.py#L84

This is not really a property (company_dependent=True field), so I think it's better to relocate all the product price history code in a separate method, but I'm not going to block the PR for this reason.

legalsylvain:
> Why don't you assign company there or make that Odoo itself writes the company there?

Could you explain what do you propose, because I don't get it.

> This is not really a property (company_dependent=True field) 

Company-dependent fields : Formerly known as 'property' fields. (ref : https://www.odoo.com/documentation/10.0/reference/orm.html)

> but I'm not going to block the PR for this reason.

nice !  ;-)

pedrobaeza:
Can you please make the changes for merging this, @legalsylvain ?

pedrobaeza (approved):

PR #1045, opened 2017-08-25
[10.0] compare_noupdate_xml_records: 3 fixes

pedrobaeza:
* Add elements to <odoo>, not <data>.
* Really pretty print.
* Don't strip CDATA tags.

cc @Tecnativa

### DIFF
--- a/scripts/compare_noupdate_xml_records.py
+++ b/scripts/compare_noupdate_xml_records.py
@@ -108,7 +108,10 @@ def get_records(addon_dir):
         for xml_file in manifest[key]:
             xml_path = xml_file.split('/')
             try:
-                tree = etree.parse(os.path.join(addon_dir, *xml_path))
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@mreficent here you have the fixes I told about. It's needed now to recheck full OpenUpgrade addons folder, but now that you have developed the folder analysis, it's very easy to you :wink: 

An example of proper output is in #1042

MiquelRForgeFlow:
Nice :+1: 

But who knows, maybe I was already aware that the compare_noupdate_xml_records was broken and maybe I developed in advance the folder analysis part because I supposed someday you will fix it and then we would need to recheck all the addons folder. 😎😉

pedrobaeza:
Yeah, this is Minority "OpenUpgrade" reports :stuck_out_tongue_winking_eye:

PR #1051, opened 2017-08-29
[FIX] sales_team: noupdate records changes + XML-ID rename

pedrobaeza:
* Website Sales team has been moved from module XML_ID and this way is kept
* We continue paying #1045 bug, so here there are other noupdate records load

cc @Tecnativa


rafaelbn (approved):
👍

JordiBForgeFlow (approved):

PR #1054, opened 2017-08-30
[9.0] Openupgrade: Rename module account_payment_term_multi_day to account_payment_term_extension

sergio-teruel:
cc @Tecnativa


pedrobaeza:
In this case is not correct. It should be merged instead in base module, as both modules exist in v8.

pedrobaeza:
This is already done:

https://github.com/OCA/OpenUpgrade/blob/8d182d2377a6eed04abd159f62b9ecff8d97651c/openerp/addons/base/migrations/9.0.1.3/pre-migration.py#L67-L68

PR #1055, opened 2017-08-30
[FIX] hr_timesheet: Project filling fix

pedrobaeza:
There can be timesheet analytic lines that are not linked to a task nor an issue.
We need to link them to the associated project of the analytic account.

### DIFF
--- a/addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
@@ -43,7 +43,8 @@ def migrate_missing_projects(env):
 
 
 def fill_analytic_line_project(env):
-    """Fill project with the linked one in the related task or issue."""
+    """Fill project with the linked one in the related task, issue, or project
+    related to the analytic account."""
     openupgrade.logged_query(
         env.cr,
         """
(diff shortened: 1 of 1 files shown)


cubells (approved):
:+1: 

LGTM

JordiBForgeFlow:
I am concerned about timesheet lines entered against pure analytic accounts, not linked to projects. This was allowed in v9.

Perhaps in project module we need to create projects for those, or perhaps here, since it is the only place where there could be a break in consistency.

rafaelbn (approved):
👍 
In other case after migration from v9 to v10 you will not see in Timesheets -> Activities any activity which was not linked to a task or issue in v9

pedrobaeza:
@jbeficent that's what this PR does: fill lines with the project for still being considered as "timesheet lines". Previously, we have created projects for those analytic lines that don't have it (https://github.com/OCA/OpenUpgrade/pull/1055/files#diff-87d2c9f7c3a0d64a5388d7dd4daf3dacL21)

MiquelRForgeFlow (approved):

JordiBForgeFlow:
Oh, you're right. :+1:

PR #1057, opened 2017-09-01
[9.0][MIG] all pending noupdate=1 records

MiquelRForgeFlow:
This PR contains the pending migration scripts of the noupdate=1 records for the modules that in the [modules80-90.rst](https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/doc/source/modules80-90.rst) list are as `Done ` or `Nothing to do`.


pedrobaeza:
Uhm, is this happening also on v9?

MiquelRForgeFlow:
Yeah, and perhaps previous versions, but I think that in previous versions is not absolute needed. I'm a bit tired haha

pedrobaeza:
Yeah, I know, you are making a great work. I think this are omissions from people that made the migration scripts (maybe me included).

addons/website/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -9,6 +9,10 @@
     <record id="menu_homepage" model="website.menu">
         <field name="website_id" ref="default_website"/>
     </record>
+    <record id="default_website" model="website">
```
pedrobaeza: This is removed conscientiously, as we don't want this to be changed in your production DB

MiquelRForgeFlow: Ok, I will delete now.

addons/account/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+  <data>
+    <record id="email_template_edi_invoice" model="mail.template"><field name="body_html">
```
pedrobaeza: Ouch, as you haven't backported #1045 and you are using v9 script, `CDATA` tags have been stripped... You need to take care of this.

pedrobaeza (changes requested):

addons/account/migrations/9.0.1.1/noupdate_changes.xml
MiquelRForgeFlow: SHIT

MiquelRForgeFlow: I included now that improvement in the other PR.

pedrobaeza: OK, waiting for Travis for merging.

MiquelRForgeFlow: And doing now the update of this PR

pedrobaeza:
Migrating to v10, I have just found another possible problem with noupdate loading: `forcecreate="0"`. I have an error on project module because it's trying to create (not to update) a record that has been removed and Odoo doesn't recreate it because of that switch.

Thus, it's better that you stay until I analyze this and find a solution. Suggestions are welcome.

addons/account/migrations/9.0.1.1/noupdate_changes.xml
MiquelRForgeFlow: Now `CDATA` tags are included 😌

pedrobaeza:
Have you seen my comment?

MiquelRForgeFlow:
Now I saw it 😄

pedrobaeza:
Please recheck if there's a forcecreate="0" record in these.

MiquelRForgeFlow:
there aren't any `forcecreate=false` here

addons/account/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,90 @@
...
+    </record>
+    <record id="account_payment_term_immediate" model="account.payment.term">
+      <field name="note">Payment term: Immediate Payment</field>
+    </record>
+    <record id="sequence_reconcile_seq" model="ir.sequence">
+      <field eval="False" name="company_id"/>
+    </record>
+    <record id="account_payment_term_15days" model="account.payment.term">
+      <field name="note">Payment term: 15 Days</field>
+      <field name="line_ids" eval="[(0, 0, {'value': 'balance', 'value_amount': 0.0, 'sequence': 500, 'days': 15, 'option': 'day_after_invoice_date'})]"/>
```
pedrobaeza: Remove this line, as there's no actual change (why is recognized by the script) and due to the addition operation, it's failing.

addons/account/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,90 @@
...
+      <field name="type">receivable</field>
+    </record>
+    <record id="data_account_type_payable" model="account.account.type">
+      <field name="close_method"/>
+      <field name="report_type"/>
+      <field name="code"/>
+      <field name="include_initial_balance" eval="True"/>
+      <field name="type">payable</field>
+    </record>
+    <record id="account_payment_term_net" model="account.payment.term">
```
pedrobaeza: Remove also this one.

pedrobaeza (changes requested):

addons/account/migrations/9.0.1.1/noupdate_changes.xml
MiquelRForgeFlow: but v8 doesn't have this line

pedrobaeza: Right, I have seen v9 and v10, but this for v8 > v9. Any way, we have the reason why it appears, but it's still incorrect, because it's adding a second line, not replacing current one. Please replace it then with 2 commands: `[(5, ), (0, 0, ...`

MiquelRForgeFlow: Done, let's see now.

MiquelRForgeFlow:
More stuff in `account` is failing due to some constrains.

pedrobaeza:
Uhm, seems to be not easy to solve, as I don't know where that account is coming from. Any way, the problem is the same for other DBs with accounts not properly set. The best bet is to just remove that record.

hbrunn:
@pedrobaeza the problem is that load_data defaults to init mode, and there forcecreate is ignored (it's init mode): https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L230
So the simplest fix is to load the data in question with `mode='update'`, and my only reason not to change the default in openupgradelib is because that's a breaking api change

pedrobaeza:
Yeah, you're totally right, Holger. I was about to raise the same thing on project v10 migration.

pedrobaeza:
@mreficent can you please change it?

MiquelRForgeFlow:
Finally it's :green_apple:!!

@pedrobaeza What do you want me to change exactly? :thinking:

addons/account_accountant/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+  <data>
+    <record id="module_install_notification" model="mail.message">
```
pedrobaeza: This one is not needed as this a trick for Odoo for notifying all employees a module installation, and I'm afraid this would be triggered also on the migration.

addons/account_voucher/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+  <data>
+    <record id="module_install_notification" model="mail.message">
```
pedrobaeza: The same as previous comment

addons/calendar/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,360 @@
...
+                                        </tr>
+                                    </table>
+                                </td>
+                            </tr>
+                        </table>
+                    </div>
+                </div>
+                ]]></field>
+    </record>
+    <record id="alarm_notif_5" model="calendar.alarm">
```
pedrobaeza: I'm not sure we should change these values, as users could customize this previously (and they are not data that can fail).

addons/calendar/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,360 @@
...
+                            <a style="padding: 8px 30px 8px 30px;border-radius: 6px;border: 1px solid #CCCCCC;background:#808080;margin : 0 15px 0 0;text-decoration: none;color:#FFFFFF;" href="/calendar/meeting/decline?db=${'dbname' in ctx and ctx['dbname'] or ''}&token=${object.access_token}&action=${'action_id' in ctx and ctx['action_id'] or ''}&id=${object.event_id.id}">Decline</a>
+                            <a style="padding: 8px 30px 8px 30px;border-radius: 6px;border: 1px solid #CCCCCC;background:#D8D8D8;text-decoration: none;color:#FFFFFF;" href="/calendar/meeting/view?db=${'dbname' in ctx and ctx['dbname'] or ''}&token=${object.access_token}&action=${'action_id' in ctx and ctx['action_id'] or ''}&id=${object.event_id.id}">View</a>
+                    </div>
+                </div>
+                ]]></field>
+    </record>
+    <record id="alarm_notif_1" model="calendar.alarm">
```
pedrobaeza: The same as previous

addons/crm/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,46 @@
...
+      <field ref="model_crm_opportunity_report" name="model_id"/>
+    </record>
+    <record id="mt_salesteam_lead_won" model="mail.message.subtype">
+      <field name="relation_field">team_id</field>
+      <field name="res_model">crm.team</field>
+    </record>
+    <record id="mail_alias_lead_info" model="mail.alias">
+      <field name="alias_parent_model_id" ref="model_crm_team"/>
+    </record>
+    <record id="module_install_notification" model="mail.message">
```
pedrobaeza: Don't include this record

addons/hr_timesheet_sheet/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version='1.0' encoding='utf-8'?>
```
pedrobaeza: This file is not needed

addons/mrp/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='utf-8'?>
```
pedrobaeza: This file is not needed

addons/project_issue/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -1,28 +1,38 @@
...
 <openerp>
   <data>
-    <!--Changes in noupdate data-->
     <record id="mt_issue_new" model="mail.message.subtype">
-        <field name="description">Issue opened</field>
-        <field name="name">Issue Opened</field>
+      <field name="description">Issue opened</field>
+      <field name="name">Issue Opened</field>
+    </record>
+    <record id="module_install_notification" model="mail.message">
```
pedrobaeza: Remove this record

addons/project_issue/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -1,28 +1,38 @@
...
                                             '&amp;',
                                                 ('project_id.privacy_visibility', '=', 'followers'),
                                                 ('message_partner_ids', 'in', [user.partner_id.id]),
                                         ]</field>
-        <field name="name">Project/Issue: employees: portal, employee or (followers and following)</field>
+      <field name="name">Project/Issue: employees: portal, employee or (followers and following)</field>
     </record>
     <record id="mt_issue_ready" model="mail.message.subtype">
-        <field name="name">Issue Ready</field>
+      <field name="description">Issue ready for next sStage</field>
```
pedrobaeza: Are you sure this field is correct (`sStage`)?

addons/purchase/migrations/9.0.1.2/noupdate_changes.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version='1.0' encoding='utf-8'?>
```
pedrobaeza: File not needed

addons/sale/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -1,15 +1,19 @@
 <?xml version='1.0' encoding='utf-8'?>
 <openerp>
   <data>
-    <!--Changes in noupdate data-->
+    <record id="module_install_notification" model="mail.message">
```
pedrobaeza: Delete this record

addons/stock/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version='1.0' encoding='utf-8'?>
```
pedrobaeza: File not needed

addons/stock_account/migrations/9.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+  <data>
+    <record id="default_cost_method" model="ir.property">
```
pedrobaeza, line 4: This is resetting the default values. Are you sure this is correct?

pedrobaeza (changes requested):
I'm afraid some noupdate changes shouldn't be blindly included, as they can have a side effect or simply not needed because they are thought for an interactive installation, not a migration.

addons/calendar/migrations/9.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: This means removing that?

pedrobaeza: That's it. Maybe you can include them commented and include a text clarifying it for having future references and not saying "Why I didn't include these records?"

addons/project_issue/migrations/9.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: That's because it is exactly what the message of the record mt_issue_ready has. But I can fix that if you like.

addons/stock_account/migrations/9.0.1.1/noupdate_changes.xml
MiquelRForgeFlow: Yes...

addons/project_issue/migrations/9.0.1.0/noupdate_changes.xml
pedrobaeza: Curious that Odoo contains that typo. I think we can fix it, yes.

MiquelRForgeFlow:
Comments attended

pedrobaeza (approved):

PR #1060, opened 2017-09-02
[FIX] compare_noupdate_xml_records: Don't ignore forcecreate

pedrobaeza:
Records with forcecreate="0"/"False" are not recreated when they
have been previously removed. If we try to load some changes in
records of this type and the records have been removed, we will get
for sure an error of missing required fields.

Preserving the attribute on changed records the check for existence
is done and you won't get errors

cc @ Tecnativa

### DIFF
--- a/scripts/compare_noupdate_xml_records.py
+++ b/scripts/compare_noupdate_xml_records.py
@@ -152,6 +152,9 @@ def main_analysis(old_update, old_noupdate, new_update, new_noupdate):
 
         element = etree.Element(
             "record", id=xml_id, model=record_new.attrib['model'])
+        # Add forcecreate attribute if exists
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@mreficent here is the solution I mentioned in the other PR.

MiquelRForgeFlow:
Can you do the same to v9?

MiquelRForgeFlow (approved):

pedrobaeza:
Cherry-picked directly in 9.0

MiquelRForgeFlow:
Should I also update the records that have now `forcecreate="True"`?

pedrobaeza:
No if you don't want. It's the default value.

PR #1062, opened 2017-09-05
[FIX] mass_mailing: Create one utm.source per mass mailing

pedrobaeza:
Odoo now inherits by delegation from utm.source and create a different
record per mass mailing, using the name of the source as subject, so we
need to pre-create a source for each record.

WARNING: This replaces sources already set, but it's the only way to
preserve the subject on the mass mailing and this binding is a feature
hardly used on previous versions.


pedrobaeza:
@mreficent this one is ready for being reviewed

yajo (approved):

MiquelRForgeFlow (approved):

PR #1063, opened 2017-09-06
[FIX] project: Deal with forcecreate=False records

pedrobaeza:
Let's see if we have all the cases covered. I have included a test for that.

cc @Tecnativa

### DIFF
--- a/addons/project/migrations/10.0.1.1/post-migration.py
+++ b/addons/project/migrations/10.0.1.1/post-migration.py
@@ -15,4 +15,5 @@ def migrate(env, version):
     )
     openupgrade.load_data(
         env.cr, 'project', 'migrations/10.0.1.1/noupdate_changes.xml',
+        mode='update',
     )
--- a/addons/project/migrations/10.0.1.1/post-migration.py
+++ b/addons/project/migrations/10.0.1.1/post-migration.py
@@ -15,5 +15,5 @@ def migrate(env, version):
     )
     openupgrade.load_data(
         env.cr, 'project', 'migrations/10.0.1.1/noupdate_changes.xml',
-        mode='update',
+        mode='init_no_create',
     )
--- a/addons/project/migrations/10.0.1.1/tests/test_openupgrade_project.yml
+++ b/addons/project/migrations/10.0.1.1/tests/test_openupgrade_project.yml
@@ -1,0 +1,4 @@
+-
+    remove an existing record with forcecreate=False to see if we have an error
+-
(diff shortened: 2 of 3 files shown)


pedrobaeza:
@mreficent this is what I was talking about on the other PR, but I decided to do it myself as I don't have it clear that everything is going to be correct.

@hbrunn my fear with `mode='update'` is that we don't get updated the rest of the existing records.

hbrunn:
I tried this yesterday but `mode=update` didn't work, because of https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_model.py#L1212 - we have an xmlid, so this will act just the same. Haven't found a good solution yet.

pedrobaeza:
I think we need to modify the core..

hbrunn:
I'd rather not do anything at all with those records as far as xml files are concerned, and do conditional writes in python code where necessary. We're talking about a couple of records per release, right?

pedrobaeza:
Well the pity is to not have an standard way to do it. Maybe what we can do is to filter out the non existing records in openupgradelib method.

pedrobaeza:
You know: if you don't have it automated, for sure we will have problems in the future when we forget to do the specific thing.

hbrunn:
sure, standard ways are better. If you want to commit time to this, perfect. Otherwise, we'll better revert the commit loading this for the project migration, as I believe real world databases tend to have the record deleted

hbrunn:
@pedrobaeza  this PR works with https://github.com/OCA/openupgradelib/pull/80

hbrunn:
@pedrobaeza as of https://github.com/OCA/openupgradelib/pull/80, you'll have to pass `mode='init_no_create'`

StefanRijnhart:
Did you consider leaving out the noupdate entries from the project module in 10 that only disable the earlier forcecreate=False items from 9.0? If the migrated database is from Odoo 8.0 the entries won't exist (which causes the problem you are trying to solve here) and if the database was installed from version 9.0 the entries will have been cleaned up or reused(!) during the 9.0 lifecycle by the customer.

StefanRijnhart:
As per my previous comment I am proposing https://github.com/OCA/OpenUpgrade/pull/1124

StefanRijnhart:
After the merge of #1124, this is no longer needed I believe.

pedrobaeza:
Well, yes, but I still want to try the test.

pedrobaeza:
Uhm, strange error, but anyway tests are correct. I close this PR then.

PR #1064, opened 2017-09-11
[IMP] migrate pricelist items with computed fixed price to fixed

hbrunn:
this tries to detect fixed prices that were hacked into the formula by 100% discount and the real price as surcharge. As I don't want to repeat the appropriate calculations in SQL, I only do this if there's no min/max margin.


pedrobaeza (commented):
Shouldn't you empty also `price_surcharge` field?

hbrunn:
that's only used in the formula (https://github.com/OCA/OCB/blob/9.0/addons/product/pricelist.py#L228) so I figure it doesn't hurt and makes it simpler to revert if ever you really want a formula for whatever reason

pedrobaeza (approved):
OK, let's keep it then

PR #1067, opened 2017-09-14
[IMP] account: Improve merge journals method

pedrobaeza:
* Lower the journal sequence, for avoiding that refund journal is selected automatically when making an invoice.
* Record the target journal where a refund journal has been merged, so that other modules can use this information if needed. For example, l10n_es_account_invoice_sequence needs this.

cc @Tecnativa @sergio-teruel


pedrobaeza:
Example of use in https://github.com/OCA/l10n-spain/pull/638

sergio-teruel (approved):

pedrobaeza:
Thanks for the review, Sergio. Merging for continuing with the dependencies.

PR #1068, opened 2017-09-15
[ADD] l10n_fr: migration scripts

cubells:
Nothing to do.

cc @Tecnativa


pedrobaeza (approved):
Next time, it should be better that you indicate about noupdate check and also to create the analysis work file, but there's indeed nothing to do.

PR #1069, opened 2017-09-15
[ADD] l10n_fr: migration scripts to 10.0

cubells:
Nothing to do.

cc @Tecnativa


pedrobaeza (approved):
The same comment about analysis work file + checked noupdate confirmation, but :+1:

PR #1070, opened 2017-09-15
[ADD] account_chart: migration scripts to 9.0

cubells:
Account_chart is now merged now in account module

cc @Tecnativa


pedrobaeza (changes requested):
You have to put in on pre-migration.py script in `base` module in the "merging operation", not in apriori that is a pure renamed, and thus it gives error as it exists another `account` module.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #1071, opened 2017-09-15
[ADD] bank_statement_period_from_line_date: migration scripts to 9.0

cubells:
bank_statement_period_from_line_date module is merged now in account module

cc @Tecnativa


pedrobaeza (changes requested):
The same error about putting it on apriori.py instead of base module.

MiquelRForgeFlow:
Why not put it here https://github.com/OCA/OpenUpgrade/pull/1071/files#diff-7c38f07ba9fa0c25b74e99a4641d3999R70?

MiquelRForgeFlow (approved):

PR #1073, opened 2017-09-19
[FIX] base: Renamed modules in apriori.py

pedrobaeza:
It was not done previously.

cc @Tecnativa


pedrobaeza:
cc @cubells @mreficent

MiquelRForgeFlow (approved):
:star:

PR #1074, opened 2017-09-19
[10.0][FIX][Base] noupdate_changes

MiquelRForgeFlow:
Just in case, I compared base modules.

### DIFF
--- a/odoo/addons/base/migrations/10.0.1.3/noupdate_changes.xml
+++ b/odoo/addons/base/migrations/10.0.1.3/noupdate_changes.xml
@@ -9,6 +9,17 @@
     <record id="tj" model="res.country">
         <field name="currency_id" ref="TJS"/>
     </record>
+    <record id="HUF" model="res.currency">
+        <field name="rounding">1.00</field>
+    </record>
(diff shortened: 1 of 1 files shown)


odoo/addons/base/migrations/10.0.1.3/noupdate_changes.xml
```diff
@@ -9,6 +9,17 @@
     <record id="tj" model="res.country">
         <field name="currency_id" ref="TJS"/>
     </record>
+    <record id="HUF" model="res.currency">
+        <field name="rounding">1.00</field>
+    </record>
+    <record id="user_root" model="res.users">
```
pedrobaeza: Better to comment out this change as people might change previously the signature at their taste.

MiquelRForgeFlow: True!

MiquelRForgeFlow: Done!

pedrobaeza (approved):

PR #1075, opened 2017-09-19
[9.0][FIX][Base] noupdate changes

MiquelRForgeFlow:
Just in case, I compared base modules.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
+++ b/openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
@@ -1,8 +1,1334 @@
 <?xml version='1.0' encoding='utf-8'?>
 <openerp>
-    <data>
-        <record model="res.groups" id="group_user">
-            <field name="implied_ids" eval="[(4, ref('group_no_one'))]"/>
-        </record>
-    </data>
+  <data>
+    <record model="res.groups" id="group_user">
+      <field name="implied_ids" eval="[(4, ref('group_no_one'))]"/>
+    </record>
+    <record id="gw" model="res.country">
+      <field eval="245" name="phone_code"/>
+    </record>
+    <record id="gu" model="res.country">
+      <field eval="0" name="phone_code"/>
+    </record>
+    <record id="gt" model="res.country">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
We are getting these warnings:
```
2017-09-19 13:01:22,182 23732 WARNING openupgrade openerp.models: res.currency.write() with unknown fields: accuracy
```
and I'm not sure why we have Travis red.

openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
```diff
@@ -1,8 +1,1511 @@
 <?xml version='1.0' encoding='utf-8'?>
 <openerp>
-    <data>
-        <record model="res.groups" id="group_user">
-            <field name="implied_ids" eval="[(4, ref('group_no_one'))]"/>
-        </record>
-    </data>
```
MiquelRForgeFlow, line 7: Travis is red because the `test_technical_features` test fails. It seems it didn't liked the removal of this lines.

MiquelRForgeFlow:
Just added previous deleted lines and now is green :green_apple:

openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
pedrobaeza: So if you restore this it throws error? But before it was green...

MiquelRForgeFlow: Here is the explanation: https://github.com/OCA/OpenUpgrade/pull/607

pedrobaeza:
You still have the accuracy warning. Please remove that lines.

openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
pedrobaeza: OK, my question was why do you remove it, but I see that you are only changing indentation and they appear later.

MiquelRForgeFlow:
Done! :wink:

pedrobaeza (approved):

PR #1076, opened 2017-09-19
[10.0][MIG] mrp

MiquelRForgeFlow:
- Completed analysis
- Completed migrations scripts
- Updated documentation
- mrp_operations migration added


PCatinean (approved):

pedrobaeza:
You will need to take into account rename fields when OCA/openupgradelib#84 will be merged.

MiquelRForgeFlow:
@pedrobaeza the rename_fields method is now taken into account

addons/mrp/migrations/10.0.2.0/openupgrade_analysis.txt
```diff
@@ -1,20 +1,20 @@
 ---Fields in module 'mrp'---
 mrp          / ir.attachment            / priority (selection)          : NEW selection_keys: ['0', '1', '2', '3']
-mrp          / mrp.bom                  / date_start (date)             : DEL
```
pedrobaeza: Please don't modify this file

addons/mrp/__manifest__.py
```diff
@@ -65,7 +65,11 @@
     ],
     'demo': [
         'data/mrp_demo.xml',
-        'data/mrp_lot_demo.yml'],
+        # 'data/mrp_lot_demo.yml'
```
pedrobaeza, line 68: Isn't there any other option? I don't like this, as you are disabling demo data that also servers for performing tests. Please put the error you get for debugging the problem. FYI, functions in YAML are also executed on update.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+# NOTHING TO DO: because we do not have pre-existing data of the picking type relating to mrp.bom
+
+mrp          / mrp.bom                  / position (char)               : DEL
+# NOTHING TO DO
+
+mrp          / mrp.bom                  / product_efficiency (float)    : DEL required: required, req_default: function
+# NOTHING TO DO
+
+mrp          / mrp.bom                  / product_rounding (float)      : DEL
+# NOTHING TO DO: Rounding is taken from the product.uom now
```
pedrobaeza, line 22: Is it possible that the rounding is not the same from product one? What to do in that cases?

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+# Copyright 2017 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_defaults(env):
+    # Set default values from field definition as this is a new field and
+    # the default val resembles the previous behavior best
+    default_specs = {
+        'mrp.bom': [('ready_to_produce', None)],
```
pedrobaeza: No need of this one, as there's already a default value applied:

https://github.com/OCA/OpenUpgrade/blob/4c1d81199e42924ae356fc390db2df8ba08a30c1/addons/mrp/models/mrp_bom.py#L53

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+# change of boms using validity dates is no longer support as it is considered risky. Manual action needs
+# to be taken from ECO in order to make changes.
+
+mrp          / mrp.bom.line             / operation_id (many2one)       : NEW relation: mrp.routing.workcenter
+# NOTHING TO DO
+
+mrp          / mrp.bom.line             / product_efficiency (float)    : DEL required: required, req_default: function
+mrp          / mrp.bom.line             / product_rounding (float)      : DEL
+mrp          / mrp.bom.line             / property_ids (many2many)      : DEL relation: mrp.property
+mrp          / mrp.bom.line             / product_uom (many2one)        : was renamed to product_uom_id [nothing to do]
```
pedrobaeza: It's not really nothing to do. You have called `rename_fields` for this field.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.bom.line             / date_stop (date)              : DEL
+# NOTHING TO DO: According to https://www.scribd.com/document/343980661/oDoo-10-Manufacturing automatic
+# change of boms using validity dates is no longer support as it is considered risky. Manual action needs
+# to be taken from ECO in order to make changes.
+
+mrp          / mrp.bom.line             / operation_id (many2one)       : NEW relation: mrp.routing.workcenter
+# NOTHING TO DO
+
+mrp          / mrp.bom.line             / product_efficiency (float)    : DEL required: required, req_default: function
+mrp          / mrp.bom.line             / product_rounding (float)      : DEL
```
pedrobaeza: The same question about rounding with mrp.bom

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.message              / bom_id (many2one)             : NEW relation: mrp.bom
+mrp          / mrp.message              / message (html)                : NEW required: required
+mrp          / mrp.message              / product_id (many2one)         : NEW relation: product.product
+mrp          / mrp.message              / product_tmpl_id (many2one)    : NEW relation: product.template
+mrp          / mrp.message              / routing_id (many2one)         : NEW relation: mrp.routing
+mrp          / mrp.message              / valid_until (date)            : NEW required: required, req_default: function
+mrp          / mrp.message              / workcenter_id (many2one)      : NEW relation: mrp.workcenter
+# NOTHING TO DO
+
+mrp          / mrp.production           / date_planned (datetime)       : was renamed to date_planned_start [nothing to do]
```
pedrobaeza, line 57: The same about `rename_fields`

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.production           / date_planned_finished (datetime): NEW
+mrp          / mrp.production           / move_created_ids (one2many)   : DEL relation: stock.move
```
pedrobaeza: Please include also one2many fields in `rename_fields` method. Although it's not an SQL column per se, you have an ir.model.field entry, it can be used in a filter, etc.

addons/mrp/migrations/10.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,63 @@
...
+    ('base.menu_mrp_config', 'mrp.menu_mrp_config'),
+    ('base.menu_mrp_root', 'mrp.menu_mrp_root'),
+]
+
+
+def prepopulate_fields(cr):
+    cr.execute(
+        """
+        ALTER TABLE mrp_production
+        ADD COLUMN picking_type_id INTEGER
```
pedrobaeza, line 154: You need to fill it with the first picking type available for having the constraint correctly applied. On post-migration you will overwrite that value with the correct one.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+    cr.execute(
+        """
+        UPDATE stock_warehouse
+        SET manufacture_to_resupply = TRUE
+        """
+    )
+
+
+def populate_production_picking_type_id(cr):
+    cr.execute(
```
pedrobaeza: Please include here as docstring the same text put on analysis work file:
```
Updated all mrp.production records trying to find
first the most suitable picking_type_id based on the destination location,
and fall back to the default picking type search if none found previously.
```

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        UPDATE mrp_production mp
```
pedrobaeza: Avoid subqueries if not needed, although the behavior changes a bit:
```
UPDATE mrp_production mp
SET picking_type_id = spt.id
FROM stock_picking_type spt
WHERE code = 'mrp_operation'
AND mp.location_dest_id = spt.default_location_dest_id
```
What happens if there's more than 1 picking type with that location? In this case, it's overwritten on each pass, so the last value is the one that remains (the most modern picking type).

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+            SELECT min(id) as id, default_location_dest_id
+            FROM stock_picking_type
+            WHERE code = 'mrp_operation'
+            GROUP BY default_location_dest_id
+        ) spt
+        WHERE mp.location_dest_id = spt.default_location_dest_id
+            AND mp.picking_type_id IS NULL
+        """
+    )
+    cr.execute(
```
pedrobaeza: Use `openupgrade.logged_query` in this kind of SQL queries (this one and the previous one) for having a trace in the log.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+# NOTHING TO DO
+
+mrp          / mrp.production           / picking_type_id (many2one)    : NEW relation: stock.picking.type, required: required, req_default: function
+# DONE: Pre-migration: pre-created the field
+# DONE: Post-migration: Updated all mrp.production records trying to find
+# first the most suitable picking_type_id based on the destination location,
+# and fall back to the default picking type search if none found previously.
+
+mrp          / mrp.production           / procurement_group_id (many2one): NEW relation: procurement.group
+# NOTHING TO DO. Reminder: this change affects to OCA module 'mrp_production_raw_material_procurement_group' in v9, where probably
```
pedrobaeza: Don't put this here, but on migration issue on OCA modules section.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+# the field 'raw_material_procurement_group_id' will need to be renamed.
+
+mrp          / mrp.production           / state (selection)             : selection_keys is now '['cancel', 'confirmed', 'done', 'planned', 'progress']' ('['cancel', 'confirmed', 'done', 'draft', 'in_production', 'ready']')
+# DONE: Post-migration. Mapped some values
```
pedrobaeza: You can mention that you needed to copy column on pre-migration.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+        'mrp.workcenter': [('sequence', None)],
+    }
+    openupgrade.set_defaults(env.cr, env, default_specs)
+
+
+def map_mrp_production_state(cr):
+    openupgrade.map_values(
+        cr,
+        openupgrade.get_legacy_name('state'), 'state',
+        [('draft', 'confirmed'),
```
pedrobaeza: Isn't there anything more to do for draft orders? Just put state=confirmed? Or you have to run anything more? I suspect that you need to make anything more.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+
+mrp          / mrp.production           / procurement_ids (one2many)    : NEW relation: procurement.order
```
pedrobaeza: Same comments about missing one2many renamings and already done renamings

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.production           / product_uom (many2one)        : was renamed to product_uom_id [nothing to do]
+mrp          / mrp.production           / propagate (boolean)           : NEW
+mrp          / mrp.production           / scrap_ids (one2many)          : NEW relation: stock.scrap
+mrp          / mrp.production           / website_message_ids (one2many): DEL relation: mail.message
+mrp          / mrp.production           / workcenter_lines (one2many)   : relation is now 'mrp.workorder' ('mrp.production.workcenter.line')
+mrp          / mrp.production           / workcenter_lines (one2many)   : was renamed to workorder_ids [nothing to do]
+# NOTHING TO DO
+
+
+mrp          / mrp.routing              / workcenter_lines (one2many)   : was renamed to operation_ids [nothing to do]
```
pedrobaeza: `rename_fields`

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+# NOTHING TO DO
+
+mrp          / mrp.routing.workcenter   / batch (selection)             : NEW required: required, selection_keys: ['no', 'yes'], req_default: function
+# DONE: Used the default value
```
pedrobaeza: No need of anything. It's already assigned by default.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_defaults(env):
+    # Set default values from field definition as this is a new field and
+    # the default val resembles the previous behavior best
+    default_specs = {
+        'mrp.bom': [('ready_to_produce', None)],
+        'mrp.routing.workcenter': [('batch', None)],
```
pedrobaeza: No need also

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+        'mrp.workcenter': [('sequence', None)],
```
pedrobaeza: And neither this one

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+            WHERE code = 'mrp_operation'
+            LIMIT 1
+        ) spt
+        WHERE mp.picking_type_id IS NULL
+        """
+    )
+
+
+def populate_routing_workcenter_routing_id(cr):
+    cr.execute(
```
pedrobaeza: Use here ORM (this is post-migration). It's simpler:
```
routing = env['mrp.routing'].search([], limit=1)
```

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        SELECT COUNT(id)
+        FROM mrp_routing
+        """
+    )
+    if cr.fetchall()[0][0] == 0:
+        cr.execute(
+            """
+            INSERT INTO mrp_routing (id, name, active)
```
pedrobaeza: ORM also.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+            VALUES (1, 'Dummy routing', TRUE)
+            """
+        )
+    cr.execute(
```
pedrobaeza: Use `openupgrade.logged_query` and use the ID you already have and remember: subqueries are very ugly and performance drainer (you are executing one query per record).

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.unbuild              / mo_id (many2one)              : NEW relation: mrp.production
+mrp          / mrp.unbuild              / name (char)                   : NEW
+mrp          / mrp.unbuild              / produce_line_ids (one2many)   : NEW relation: stock.move
+mrp          / mrp.unbuild              / product_id (many2one)         : NEW relation: product.product, required: required
+mrp          / mrp.unbuild              / product_qty (float)           : NEW required: required
+mrp          / mrp.unbuild              / product_uom_id (many2one)     : NEW relation: product.uom, required: required
+mrp          / mrp.unbuild              / state (selection)             : NEW selection_keys: ['done', 'draft']
+
+
+mrp          / mrp.workcenter           / capacity_per_cycle (float)    : was renamed to capacity [nothing to do]
```
pedrobaeza, line 135: rename_fields comment

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.workcenter           / color (integer)               : NEW
+mrp          / mrp.workcenter           / costs_cycle (float)           : DEL
```
pedrobaeza: Isn't this time now on the table `time_ids`? If so, you have to create corresponding entries

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.workcenter           / oee_target (float)            : NEW
+mrp          / mrp.workcenter           / order_ids (one2many)          : NEW relation: mrp.workorder
+mrp          / mrp.workcenter           / product_id (many2one)         : DEL relation: product.product
+mrp          / mrp.workcenter           / routing_line_ids (one2many)   : NEW relation: mrp.routing.workcenter
+mrp          / mrp.workcenter           / time_cycle (float)            : DEL
+mrp          / mrp.workcenter           / time_ids (one2many)           : NEW relation: mrp.workcenter.productivity
+# NOTHING TO DO
+
+mrp          / mrp.workcenter           / sequence (integer)            : NEW required: required, req_default: function
+# DONE: Used the default value
```
pedrobaeza: Not needed. Done automatically

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.workcenter.productivity / loss_id (many2one)            : NEW relation: mrp.workcenter.productivity.loss, required: required
+mrp          / mrp.workcenter.productivity / user_id (many2one)            : NEW relation: res.users
+mrp          / mrp.workcenter.productivity / workcenter_id (many2one)      : NEW relation: mrp.workcenter, required: required
+mrp          / mrp.workcenter.productivity / workorder_id (many2one)       : NEW relation: mrp.workorder
+mrp          / mrp.workcenter.productivity.loss / loss_type (selection)         : NEW required: required, selection_keys: ['availability', 'performance', 'productive', 'quality'], req_default: function
+mrp          / mrp.workcenter.productivity.loss / manual (boolean)              : NEW
+mrp          / mrp.workcenter.productivity.loss / name (char)                   : NEW required: required
+mrp          / mrp.workcenter.productivity.loss / sequence (integer)            : NEW
+
+####### DONE: Renamed table 'mrp_production_workcenter_line' to 'mrp_workorder' ################
```
pedrobaeza: Do all fields match one by one?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / mrp.workorder            / time_ids (one2many)           : NEW relation: mrp.workcenter.productivity
+mrp          / mrp.workorder            / workcenter_id (many2one)      : NEW relation: mrp.workcenter, required: required
+# NOTHING TO DO
+
+mrp          / procurement.order        / property_ids (many2many)      : DEL relation: mrp.property
+# NOTHING TO DO
+
+
+mrp          / stock.move               / consumed_for (many2one)       : DEL relation: stock.move
+# DONE: Pre-migration: Renamed to None and copied this column
```
pedrobaeza: Indicate in the comment for what is used (consumed quants)?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+
+mrp          / stock.move               / active_move_lot_ids (one2many): NEW relation: stock.move.lots
+mrp          / stock.move               / bom_line_id (many2one)        : NEW relation: mrp.bom.line
+mrp          / stock.move               / consume_unbuild_id (many2one) : NEW relation: mrp.unbuild
+mrp          / stock.move               / move_lot_ids (one2many)       : NEW relation: stock.move.lots
+mrp          / stock.move               / operation_id (many2one)       : NEW relation: mrp.routing.workcenter
+mrp          / stock.move               / quantity_done_store (float)   : NEW
+mrp          / stock.move               / unbuild_id (many2one)         : NEW relation: mrp.unbuild
+mrp          / stock.move               / unit_factor (float)           : NEW
+mrp          / stock.move               / workorder_id (many2one)       : NEW relation: mrp.workorder
```
pedrobaeza: This one should be set for old workorders not done yet, or better, for all of them for having correctly the history. The same can apply to other fields of this model as well.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,419 @@
...
+mrp          / stock.move.lots          / done_wo (boolean)             : NEW
+mrp          / stock.move.lots          / lot_id (many2one)             : NEW relation: stock.production.lot
+mrp          / stock.move.lots          / lot_produced_id (many2one)    : NEW relation: stock.production.lot
+mrp          / stock.move.lots          / lot_produced_qty (float)      : NEW
+mrp          / stock.move.lots          / move_id (many2one)            : NEW relation: stock.move
+mrp          / stock.move.lots          / production_id (many2one)      : NEW relation: mrp.production
+mrp          / stock.move.lots          / quantity (float)              : NEW
+mrp          / stock.move.lots          / quantity_done (float)         : NEW
+mrp          / stock.move.lots          / workorder_id (many2one)       : NEW relation: mrp.workorder
+# NOTHING TO DO
```
pedrobaeza, line 230: I think you need to fill this table with historic data.

addons/mrp/migrations/10.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,63 @@
...
+    ('mrp.production', 'mrp_production', 'date_planned', 'date_planned_start'),
+    ('mrp.production', 'mrp_production', 'move_lines', 'move_raw_ids'),
+    ('mrp.production', 'mrp_production', 'product_uom', 'product_uom_id'),
+    ('mrp.production', 'mrp_production', 'workcenter_lines', 'workorder_ids'),
+    ('mrp.routing', 'mrp_routing', 'workcenter_lines', 'operation_ids'),
+    ('mrp.workcenter', 'mrp_workcenter', 'capacity_per_cycle', 'capacity'),
+    ('mrp.workorder', 'mrp_workorder', 'date_planned', 'date_planned_start'),
+]
+
+_table_renames = [
```
pedrobaeza: Use also `model_renames` method

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+            ON sqmr_consumed.move_id = sm_consumed.id
+        INNER JOIN stock_quant_move_rel sqmr_produced
+            ON sqmr_produced.move_id = sm_produced.id
+        """ % openupgrade.get_legacy_name('consumed_for')
+    )
+    quants = cr.fetchall()
+    for quant in quants:
+        cr.execute(
+            """
+            INSERT INTO stock_quant_consume_rel
```
pedrobaeza: You can make only one query (with `openupgrade.logged_query`, please) with the insert and the select as source. Check https://www.w3schools.com/sql/sql_insert_into_select.asp

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,140 @@
...
+        cr,
+        openupgrade.get_legacy_name('state'), 'state',
+        [('draft', 'pending'),
+         ('starworking', 'progress'),
+         ('pause', 'ready'),
+         ], table='mrp_workorder', write='sql')
+
+
+def update_stock_warehouse(cr):
+    cr.execute(
```
pedrobaeza: You are doing this by SQL, so no recomputation is triggered as you mention on the analysis file (`...so as to force the creation of the default picking types and pull rules for manufacturing on that warehouse`)

pedrobaeza (changes requested):

pedrobaeza:
In addition, you have to check also mrp_operations possible changes (as now is integrated on mrp).

addons/mrp/migrations/10.0.2.0/openupgrade_analysis.txt
MiquelRForgeFlow: It was done automatically by PCatinean :S

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: that was previous I used the rename_fields...

addons/mrp/migrations/10.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,63 @@
...
+
+from openupgradelib import openupgrade
+from odoo import tools
+
+_column_copies = {
+    'mrp_production': [
+        ('state', None, None)
+    ],
+    'mrp_workorder': [
+        ('state', None, None)
```
StefanRijnhart: This field is from module mrp_operations, not mrp.

StefanRijnhart: https://github.com/Eficent/OpenUpgrade/pull/11/files

pedrobaeza: But this PR merges both modules before.

StefanRijnhart: @pedrobaeza this migration script breaks the migration when mrp_operations was not installed in the first place. I fixed it in my PR on this branch.

pedrobaeza: Ah, OK, got it!

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: What do you propose?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: In this case, should it be this?

        ('mrp.production', 'mrp_production', 'move_created_ids', None),

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: At least emit a warning. Other possibility is to create a specific uom with that rounding called `openupgrade_...`

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: No, it should be a target one2many. How is this field now named?

MiquelRForgeFlow: `move_finished_ids`?

addons/mrp/migrations/10.0.2.0/pre-migration.py
MiquelRForgeFlow: Which constraint? What I am doing in post-migration is just caring for the cases where <code>mp.picking_type_id IS NULL</code>, because may exist new test data that may fill that table and column, and I don't want to rewrite test data.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: Which historic data? This table is new

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: And through which relation can I fill them?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: Some fields come from the <code>mrp.operations</code> module and some other fields are new. But there are many matches.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: You tell me... I haven't dug on v9/v10 data structures enough

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: Please double check if any of the fields has been renamed

MiquelRForgeFlow: Yes, I checked them :wink:, that's why in the <code>_field_renames</code> I put

            ('mrp.production', 'mrp_production', 'date_planned', 'date_planned_start'),

addons/mrp/migrations/10.0.2.0/post-migration.py
MiquelRForgeFlow: @jbeficent, what do you think of this?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: done

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: done

addons/mrp/migrations/10.0.2.0/pre-migration.py
MiquelRForgeFlow: how? I proved putting it before and after the table_renames and in both cases fails

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: Again I don't know data structure, but I think it should be there in previous version somewhere

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: Then reflect it here on the work file

addons/mrp/migrations/10.0.2.0/post-migration.py
StefanRijnhart: In 10.0, in its initial state any manufacturing order already has the end product stock moves and the raw material stock moves generated. See https://github.com/odoo/odoo/blob/10.0/addons/mrp/models/mrp_production.py#L267

addons/mrp/__manifest__.py
MiquelRForgeFlow: `mrp_demo.xml` has `<function>` tags that later `mrp_lot_demo.yml` uses.
This `<function>` tags only are executed on install, not in updates.
When migrating (migration is an update), this tags are not executed and `mrp_lot_demo.yml` breaks. Thus, `mrp_lot_demo.yml` is disabled to avoid the breaking.

OpenUpgrade is a tool just for migrating, not for testing nor working.

pedrobaeza: No, if you are disabling this, something is not correct. We need to test the migration, and it's done through demo data, so enable it and solve any problem. If not, comment the exact problems you find.

MiquelRForgeFlow: You cannot test the demo data that is not loaded. The `<function>` tags in the `mrp_demo.xml` are not loaded!

pedrobaeza: OK, I see this doesn't include demo data. Please add a comment about that.

MiquelRForgeFlow: The comment is already added in the manifest...

pedrobaeza: OK, OK, I didn't see them here in the simplified widget

addons/mrp/migrations/10.0.2.0/post-migration.py
MiquelRForgeFlow: @StefanRijnhart So, for those that were `draft` and became `confirmed`, we should do a `production._generate_moves()`, right?

leio:
What's the status here? How can others help? Mostly also, has there since been any work on this outside what we can see on this PR, to not duplicate any work?
If work is needed to migrate successfully (as opposed to just some style and perfection issues still present in PR), then I might be able to help, as I need this as well, just don't want to waste and duplicate efforts.

addons/mrp/migrations/10.0.2.0/post-migration.py
leio: This should be `UPDATE mrp_workcenter SET time_start = time_start*60, time_stop=time_stop*60` (at least to actually launch on postgres). There is no **AND** in the syntax.

StefanRijnhart: @mreficent sorry, missed your question. But yes, indeed!

addons/mrp/migrations/10.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,72 @@
...
+        ADD COLUMN picking_type_id INTEGER
+        """
+    )
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    cr = env.cr
+    tools.drop_view_if_exists(cr, 'mrp_workorder')
+    openupgrade.rename_models(cr, _model_renames)
```
leio: This is renaming models based on data that's actually about tables, not models.
`_model_renames` should be renamed to `_table_renames` and `rename_tables` called with it, not `rename_models`. Additionally might have to do something about the model information as well, though it's already filled by the previous VIEW backed model, maybe that can be deleted and then the model rename done too.

Without these changes, the migration actually fails for me, because there will be no mrp_workorder table at all, so first the sequence deletion from _column_copies fails, and afterwards post-migration on mrp_workorder as well.

leio: "[FIX] without model_renames" commit seems to already fix this up, but then "[ADD] More migration scripts" pushed later messes it up again

leio (changes requested):
The picking_type_id seems to be failing as-is with this patchset for me (plus the other fixes I've mentioned). There will be no `stock.picking.type` with code=='mrp_operation', because it seems only created on first need inside _get_manufacture_pull_rules_values, called inside mrp addons stock.warehouse models create_routes additions (amongst other places). But create_routes is only called upon creating a new warehouse - but we are migrating a v9 existing warehouse.
EDIT: For new warehouses, it's actually first created via create_sequences_and_picking_types additions in mrp module, it seems; but the point is the same.

So as far as I can see, the necessary additional warehouse routes will need to be created in mrp pre-migration for each stock.warehouse present.

Otherwise things fall over as follows:

* mrp module upgrade (after pre-migration) creates ir_model_data entry against default warehouses manu_type_id via `addons/mrp/data/mrp_data.yml`, which doesn't exist as there's no code=='mrp_operation' stock.picking.type's for it (there would be for fresh databases as stock.warehouse create would have called create_routes)
* post-migration script goes looking for suitable mrp_operation code picking types to assign to the new picking_type_id field, but none will be found -- picking_type_id will remain empty for all existing mrp.production records. All picking_type_id will remain unset and NOT NULL contraint addition will still fail after end-migration --update=all calls as well.
* `odoo.addons.base.ir.ir_model: Deleting 0@stock.picking.type (mrp.picking_type_manufacturing)` is done, as it doesn't point to any existing record (expected, just confusing if thinking in terms of stock.picking.type instead of ir.ir_model)
* If the database would be used afterwards (e.g. if it was a migration from v9 where MRP was used to setup BOMs and other things, but no actual manufacturing orders yet existed), then there will be picking_type_id unset errors when trying, and curiously in some cases it could end up using an internal transfer stock.picking.type instead of mrp_operation (I have one such instance somehow in my ongoing migration test results).

tl;dr - stock.warehouse/stock.picking.type/etc needs adapting for mrp use cases in mrp pre-migration script

EDIT: Adding `env['stock.warehouse'].search([])._create_manufacturing_picking_type()` at start of post-migration seems to "solve" it for me, but I haven't reviewed potential other new record needs for mrp stock, and this beats the purpose of adding the column in pre-migration with the intention of trying to have things ready by pre-migration time; but registry doesn't seem to exist in pre-migration, so for now went the easy way via post-migration myself locally to get further.

addons/mrp/migrations/10.0.2.0/pre-migration.py
```diff
@@ -36,16 +38,32 @@
...
+    tools.drop_view_if_exists(cr, 'mrp_workorder')
+    openupgrade.rename_tables(
+        cr, [('mrp_production_workcenter_line', 'mrp_workorder')],
+    )
+    openupgrade.logged_query(
+        cr,
+        "DELETE FROM ir_model WHERE model = %s",
+        ('mrp.workorder', )
+    )
+    openupgrade.rename_models(
```
leio: This now errors with the following after pre-migration (module own upgrade, before post-migration, it seems):

```
2018-01-16 10:23:20,719 28239 INFO migrate9_migrated odoo.modules.loading: loading mrp/security/ir.model.access.csv
2018-01-16 10:23:22,495 28239 INFO migrate9_migrated odoo.sql_db: bad query: INSERT INTO "ir_model_access" ("id", "perm_read", "name", "active", "perm_unlink", "perm_write", "p
erm_create", "group_id", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval('ir_model_access_id_seq'), true, 'mrp.workorder.user', true, true, true, true, 3
5, 1, 1, (now() at time zone 'UTC'), (now() at time zone 'UTC')) RETURNING id
2018-01-16 10:23:22,537 28239 INFO migrate9_migrated odoo.sql_db: bad query: INSERT INTO "ir_model_access" ("id", "perm_read", "name", "active", "perm_unlink", "perm_write", "p
erm_create", "group_id", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval('ir_model_access_id_seq'), true, 'mrp.workorder', true, true, true, true, 36, 1,
 1, (now() at time zone 'UTC'), (now() at time zone 'UTC')) RETURNING id
2018-01-16 10:23:23,943 28239 ERROR migrate9_migrated odoo.modules.registry: Failed to load registry
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/registry.py", line 82, in new
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 409, in load_modules
    force, status, report, loaded_modules, update_module, upg_registry)
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 299, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks, upg_registry=upg_registry)
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 181, in load_module_graph
    _load_data(cr, module_name, idref, mode, kind='data')
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 98, in _load_data
    tools.convert_file(cr, module_name, filename, idref, mode, noupdate, kind, report)
  File "/var/tmp/openupgrade/10.0/server/odoo/tools/convert.py", line 843, in convert_file
    convert_csv_import(cr, module, pathname, fp.read(), idref, mode, noupdate)
  File "/var/tmp/openupgrade/10.0/server/odoo/tools/convert.py", line 899, in convert_csv_import
    raise Exception(_('Module loading %s failed: file %s could not be processed:\n %s') % (module, fname, warning_msg))
Exception: Module loading mrp failed: file mrp/security/ir.model.access.csv could not be processed:
 No matching record found for external id 'model_mrp_workorder' in field 'Object'
Missing required value for the field 'Object' (model_id)
No matching record found for external id 'model_mrp_workorder' in field 'Object'
Missing required value for the field 'Object' (model_id)
2018-01-16 10:23:23,951 28239 CRITICAL migrate9_migrated odoo.service.server: Failed to initialize database `migrate9_migrated`.
  File "/var/tmp/openupgrade/10.0/server/odoo/service/server.py", line 914, in preload_registries
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/registry.py", line 82, in new
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 409, in load_modules
    force, status, report, loaded_modules, update_module, upg_registry)
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 299, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks, upg_registry=upg_registry)
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 181, in load_module_graph
    _load_data(cr, module_name, idref, mode, kind='data')
  File "/var/tmp/openupgrade/10.0/server/odoo/modules/loading.py", line 98, in _load_data
    tools.convert_file(cr, module_name, filename, idref, mode, noupdate, kind, report)
  File "/var/tmp/openupgrade/10.0/server/odoo/tools/convert.py", line 843, in convert_file
    convert_csv_import(cr, module, pathname, fp.read(), idref, mode, noupdate)
  File "/var/tmp/openupgrade/10.0/server/odoo/tools/convert.py", line 899, in convert_csv_import
    raise Exception(_('Module loading %s failed: file %s could not be processed:\n %s') % (module, fname, warning_msg))
Exception: Module loading mrp failed: file mrp/security/ir.model.access.csv could not be processed:
 No matching record found for external id 'model_mrp_workorder' in field 'Object'
Missing required value for the field 'Object' (model_id)
No matching record found for external id 'model_mrp_workorder' in field 'Object'
Missing required value for the field 'Object' (model_id)
2018-01-16 10:23:23,954 28239 INFO migrate9_migrated odoo.service.server: Initiating shutdown
2018-01-16 10:23:23,954 28239 INFO migrate9_migrated odoo.service.server: Hit CTRL-C again or send a second signal to force the shutdown.
```

LoisRForgeFlow:
I updated the PR and tested it in an environment without work centers but several manufacturing WH's.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,228 @@
...
+        wh_id = loc.get_warehouse()
+        wh_ids.append(wh_id.id)
+        loc_wh_map[loc.id] = wh_id.id
+    # update manufacture_to_ressupply
+    cr.execute(
+        """
+        UPDATE stock_warehouse
+        SET manufacture_to_resupply = TRUE
+        WHERE id in %s
+        """ % (tuple(set(wh_ids)),)
```
StefanRijnhart: Replace `%` with `,`. It prevents 

```
  File "/home/odoo/buildout/ou100/parts/odoo/addons/mrp/migrations/10.0.2.0/post-migration.py", line 86, in update_stock_warehouse
    """ % (tuple(set(wh_ids)),)
  File "/home/odoo/buildout/ou100/parts/odoo/odoo/sql_db.py", line 154, in wrapper
  File "/home/odoo/buildout/ou100/parts/odoo/odoo/sql_db.py", line 231, in execute
ProgrammingError: syntax error at or near ")"
LINE 4:         WHERE id in (1,)
```
Probably best not to execute this code when there are no wh_ids (even if that is not very likely at this point)

LoisRForgeFlow:
Hi @leio 

I'm reading now your comments (sorry for the delay) and I find them very detailed and helpful, thanks! I did a commit some days ago that I think addresses some of your comments. Could you check?

Right now, I found the same error you point here: https://github.com/OCA/OpenUpgrade/pull/1076#discussion_r161761157. I will continue the discussion there.

You can always collaborate opening a PR against [Eficent/Openupgrade/10.0-mig-mrp](https://github.com/eficent/openupgrade/tree/10.0-mig-mrp).

addons/mrp/migrations/10.0.2.0/pre-migration.py
LoisRForgeFlow: have you found the way to overcome this?

LoisRForgeFlow: @leio Just fixed it here: https://github.com/OCA/OpenUpgrade/pull/1076/commits/7fe9dc234b78bbbc314421bf716516a44f26b9ee

MiquelRForgeFlow:
I don't know what is happening now. Should I rebase?

leio:
As another small issue I've noticed (and I'm not sure this is an issue with mrp migration) is that after the migration, new MO's created by e.g. mrp scheduler get the name taken from a different sequence than correct (so I'm getting WH/IN/ prefix instead of MO/ for new manufacturing orders).

leio:
Is the mrp.workorder sequence column copy really necessary? Is there a chance it gets deleted before post-migration runs? For me it's still there even after migration and it's still there in my latest full migration run too (the one with a "-u all" done after migrate.py), with all the same values as `openupgrade_legacy_10_0_sequence`, as sequence column doesn't exist anymore.

If the copy is made, the original should be deleted, or it should rather be a rename to begin with, as otherwise there'll be an unused column leftover that isn't visibly detectable from the openupgrade prefix?

leio:
It seems lots of mrp.production.workcenter.line migration is not considered due to the analysis showing its removal and mrp.workorder addition, without considering the fields?

For example mrp.workcenter.productivity data could be generated from old delay and date_start/date_finished data, I believe.

The differences I see between mrp.product.workcenter.line migrated to v10 mrp.workorder from v8/v9 and mrp.workorder in a fresh v10 database and a v8 database are as follows (this is from psql output sorting and comparing, not a perfect way and might contain mistakes and meaning/content changes not considered this way):

1. sequence field is still around after migration, despite migration to next_work_order_id existing
2. delay, hour and cycle field are still around and unused and NOT migrated yet (mrp.workcenter.productivity maybe?)
3. qty field is still around and unused. Missing migration to qty_produced and qty_producing?
4. New mrp.workorder fields duration, duration_expected, capacity, production_date, duration_unit, duration_percent, production_availability, date_planned_finished, operation_id and final_lot_id - might be missing filling up with correct data during migration. Though some of these are computed fields that will probably be computed in, but only correctly if the non-computed ones are properly filled - they aren't, e.g. duration stuff can't be computed if time_ids (mrp.workcenter.productivity) isn't properly filled with old delay/hour/cycle data from point 2.

Some of these might be relevant only if mrp_operations was used in v8/v9. So might need some conditionals?

The main difference between v9 and v10 that isn't considered yet in this PR seems to be the changes for mrp.workorder to support recording of different workers doing part of the work and recording that data per-worked in the new mrp.workcenter.productity model, while before just the duration and time without employee was somewhat saved. Without migrating as much as possible, OEE data and reports seem rather useless.

This is a bit more important for my use case, as I had made v8 mrp.production.workcenter.delay writable (not just copy from workcenter data) and added an employee_id field; but that just gives me possibility to end up with more exact work duration times and correct user_id association in mrp.workcenter.productivity, while I believe without those the data is still important enough to migrate over as good as possible.

I'm not sure how much I can work on this in a useful way, as I'm not sure I want to sign a CLA just for this. Let me know if OCA CLA is not required for OpenUpgrade.

StefanRijnhart:
@leio we'd appreciate your contribution. CLA is required as on any OCA project I'm afraid.

leio:
Here's a little SQL query with which I started with to get some data into mrp.workcenter.productivity from old data:

```
INSERT INTO mrp_workcenter_productivity
(user_id, description,
create_uid, create_date, write_uid, write_date,
loss_id, date_start, date_end,
workorder_id, workcenter_id, temp_employee_id)
SELECT
1 user_id, 'Time Tracking: ' || emp.name_related AS description,
wo.create_uid, wo.create_date, wo.write_uid, wo.write_date,
7 loss_id, wo.date_start, wo.date_start + (wo.delay || ' hour')::INTERVAL AS date_end,
wo.id workorder_id, wo.workcenter_id, wo.employee_id
FROM mrp_workorder wo LEFT JOIN hr_employee emp ON wo.employee_id=emp.id
WHERE wo.state NOT IN ('cancel', 'pending') AND wo.employee_id IS NOT NULL;
```

It's far from something suitable for the migration script, but maybe it helps someone. As further notes, I had an employee_id added into v8 myself that lets record who did the workorder, which is kept here via temp_employee_id, to be post-processed into user_id later on instead of right now. So ignore that part (and the relevant table join) I guess. `loss_id` is hard-coded to my database. Cancelled and pending stuff ignored on purpose right now.
I'll need to come back to this later for my own needs to get this more right, but maybe it gives some ideas for OpenUpgrade purposes meanwhile.

I'm not going to sign a CLA for just this (and I am very wary of any CLAs in general), so it's probably better to not contribute real code for OpenUpgrade by me, or you'll end up with copyright and licensing issues when looking at any actual code I'd write...

pedrobaeza:
@leio thanks for the contribution, but there's no problem with the CLA. It's only a way to help us to make any action for protecting the code quickly, but it's your call. Any way, don't expect certain things to get done if you don't involve enough in the process.

leio:
Thanks, I would involve myself more, but I can't - the CLA is hindering. So I hope my comments are individually useful to someone and not requiring the CLA.
And there are many problems with the CLA, but this is not the forum to discuss that.

pedrobaeza:
OK, please expose your problems with the CLA in the contributors list and we can debate about it, but every open source serious software has its own CLA.

leio:
`Efficiency Factor` is wrong after migration (1% mostly, as 1 in DB was the default until v9.0), but this is really a bug in the migration of the `resource` module, which changed `time_efficiency` field to store actual percent value, not float representation; thus resource module needs migration that multiplies all existing `resource.resource:time_efficiency` by 100.0

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,424 @@
...
+
+# Good reference for tracking changes between versions:
+# https://www.scribd.com/document/343980661/oDoo-10-Manufacturing
+# https://github.com/odoo/odoo/commit/2ddc35a5300e89981c0572ae6749aee946f34e9d
+
+mrp          / ir.attachment            / priority (selection)          : NEW selection_keys: ['0', '1', '2', '3']
+# NOTHING TO DO
+
+mrp          / mrp.bom                  / date_start (date)             : DEL
+mrp          / mrp.bom                  / date_stop (date)              : DEL
```
leio, line 11: At the very least BoMs that are not valid anymore should be archived for v10, otherwise they'll suddenly be possibly used again during production as a valid BoM. Not sure what could be done with date_start in future BoMs. Also date_stop values in the future will have information lost, but alas, I guess nothing that can be done about that, other than warn in some documentation or whatnot.

leio (changes requested):

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -0,0 +1,230 @@
...
+    wh_pt_map = dict()
+    for wh in env["stock.warehouse"].browse(list(set(wh_ids))):
+        wh._create_manufacturing_picking_type()
+        wh_pt_map[wh.id] = wh.manu_type_id.id
+    for loc in loc_wh_map.keys():
+        loc_pt_map[loc] = wh_pt_map[loc_wh_map[loc]]
+    return loc_pt_map
+
+
+def populate_production_picking_type_id(cr, loc_pt_map):
```
leio, line 151: Old procurement.rule records with `action == 'manufacture'` need updated as well to the newly created manufacturing picking type. Otherwise new manufacturing orders created by procurements (e.g. scheduler from orderpoint rules on manufactured products) are created from WH/IN/ sequence, not MO/ sequence, resulting in completely wrong MO names and all the other issues that might come from wrong picking type. In other words, if not doing that, this method only updates old MO's to have correct picking type, but new ones will still be wrong.

leio (changes requested):

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,424 @@
...
+
+
+mrp          / mrp.routing              / workcenter_lines (one2many)   : was renamed to operation_ids [nothing to do]
+# DONE: pre-migration: Renamed field.
+
+
+mrp          / mrp.routing.workcenter   / batch (selection)             : NEW required: required, selection_keys: ['no', 'yes'], req_default: function
+mrp          / mrp.routing.workcenter   / batch_size (float)            : NEW
+mrp          / mrp.routing.workcenter   / cycle_nbr (float)             : DEL required: required, req_default: function
+mrp          / mrp.routing.workcenter   / hour_nbr (float)              : DEL required: required, req_default: function
```
leio: Shouldn't this be migrated to time_cycle_manual?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,424 @@
...
+# DONE: pre-migration: Renamed fields.
+
+
+mrp          / mrp.routing              / workcenter_lines (one2many)   : was renamed to operation_ids [nothing to do]
+# DONE: pre-migration: Renamed field.
+
+
+mrp          / mrp.routing.workcenter   / batch (selection)             : NEW required: required, selection_keys: ['no', 'yes'], req_default: function
+mrp          / mrp.routing.workcenter   / batch_size (float)            : NEW
+mrp          / mrp.routing.workcenter   / cycle_nbr (float)             : DEL required: required, req_default: function
```
leio: Shouldn't this be migrated to somewhere, or hour_nbr multipled by this while migrating hour_nbr to time_cycle_manual?

leio (changes requested):

addons/mrp/migrations/10.0.2.0/pre-migration.py
```diff
@@ -50,17 +50,29 @@ def rename_mrp_workorder(cr):
...
         cr,
         "DELETE FROM ir_model WHERE model = %s",
-        ('mrp.workorder', )
+        ('mrp.workorder',)
+    )
+    openupgrade.rename_tables(
+        cr, [('mrp_production_workcenter_line', 'mrp_workorder')]
     )
     openupgrade.rename_models(
         cr, [('mrp.production.workcenter.line', 'mrp.workorder')]
```
leio: For some reason I'm now failing migration with:
```
OpenUpgrade: mrp: pre-migration script called with version 9.0.1.1
OpenUpgrade: Running DELETE FROM ir_model WHERE model = mrp.workorder
OpenUpgrade: 1 rows affected
OpenUpgrade: table mrp_production_workcenter_line: renaming to mrp_workorder
OpenUpgrade: table mrp_production_workcenter_line_id_seq: renaming to mrp_workorder_id_seq
OpenUpgrade: model mrp.production.workcenter.line: renaming to mrp.workorder
odoo.sql_db: bad query: UPDATE ir_model_data SET name='model_mrp_workorder' WHERE name='model_mrp_production_workcenter_line' AND model = 'ir.model'
mrp: error in migration script mrp/migrations/10.0.2.0/pre-migration.py: IntegrityError('duplicate key value violates unique constraint "ir_model_data_module_name_uniq"\nDETAIL:  Key (name, module)=(model_mrp_workorder, mrp) already exists.\n',)
OpenUpgrade: duplicate key value violates unique constraint "ir_model_data_module_name_uniq"
DETAIL:  Key (name, module)=(model_mrp_workorder, mrp) already exists.
  File "/var/lib/odoo/openupgrade/openupgradelib/openupgradelib/openupgrade.py", line 1427, in wrapped_function
    if use_env2 else cr, version)
  File "/var/lib/odoo/openupgrade/10.0/addons/mrp/migrations/10.0.2.0/pre-migration.py", line 90, in migrate
    rename_mrp_workorder(cr)
  File "/var/lib/odoo/openupgrade/10.0/addons/mrp/migrations/10.0.2.0/pre-migration.py", line 62, in rename_mrp_workorder
    cr, [('mrp.production.workcenter.line', 'mrp.workorder')]
  File "/var/lib/odoo/openupgrade/openupgradelib/openupgradelib/openupgrade.py", line 548, in rename_models
    ('model_' + _new, 'model_' + _old),
  File "/var/lib/odoo/openupgrade/10.0/server/odoo/sql_db.py", line 154, in wrapper
  File "/var/lib/odoo/openupgrade/10.0/server/odoo/sql_db.py", line 231, in execute
IntegrityError: duplicate key value violates unique constraint "ir_model_data_module_name_uniq"
DETAIL:  Key (name, module)=(model_mrp_workorder, mrp) already exists.
```

I don't recall it failing back when I tested these fixed at end of January.. Odd.

Anyhow, it seems I still have a stale entry about mrp.workorder view in ir_model_data table as well (I have no cascade delete rules on this from ir_model in my DB at least), so need to delete that as well, in addition to ir_model above. Something like
```
openupgrade.logged_query(
    cr,
    "DELETE FROM ir_model_data WHERE name = %s",
    ('model_mrp_workorder',)
)
```

I'm not sure what changed meanwhile, maybe some openupgrade change (I see some small improvements in that area, but didn't look deep if they could affect this).
It does feel like there should be some openupgrade helper method to call for this use case (dropping a view that had an Odoo model backing it, or perhaps just together with a full model delete that checks if it's a view or table), instead of handrolling it here. Maybe there is something suitable for the ir_model + ir_model_data munging already too?

leio (changes requested):

addons/mrp/migrations/10.0.2.0/pre-migration.py
leio: After applying that quick ir_model_data "fix", the next failure related to this appears (this was definitely changed meanwhile in openupgradelib, but maybe just the way the query is done to be without regexp maybe?):

```
odoo.modules.migration: module mrp: Running migration [>10.0.2.0] pre-migration
OpenUpgrade: mrp: pre-migration script called with version 9.0.1.1
OpenUpgrade: Running DELETE FROM ir_model WHERE model = mrp.workorder
OpenUpgrade: 1 rows affected
OpenUpgrade: Running DELETE FROM ir_model_data WHERE name = model_mrp_workorder
OpenUpgrade: 1 rows affected
OpenUpgrade: table mrp_production_workcenter_line: renaming to mrp_workorder
OpenUpgrade: table mrp_production_workcenter_line_id_seq: renaming to mrp_workorder_id_seq
OpenUpgrade: model mrp.production.workcenter.line: renaming to mrp.workorder
odoo.sql_db: bad query: UPDATE ir_model_data imd
            SET name = 'field_' || 'mrp_workorder' || '_' || imf.name
            FROM ir_model_fields imf
            WHERE imd.model = 'ir.model.fields'
                AND imd.name = 'field_' || 'mrp_production_workcenter_line' || '_' || imf.name
                AND imf.model = 'mrp.production.workcenter.line'
OpenUpgrade: mrp: error in migration script mrp/migrations/10.0.2.0/pre-migration.py: IntegrityError('duplicate key value violates unique constraint "ir_model_data_module_name_uniq"\nDETAIL:  Key (name, module)=(field_mrp_workorder_id, mrp) already exists.\n',)
OpenUpgrade: duplicate key value violates unique constraint "ir_model_data_module_name_uniq"
DETAIL:  Key (name, module)=(field_mrp_workorder_id, mrp) already exists.
  File "/var/lib/odoo/openupgrade/openupgradelib/openupgradelib/openupgrade.py", line 1427, in wrapped_function
    if use_env2 else cr, version)
  File "/var/lib/odoo/openupgrade/10.0/addons/mrp/migrations/10.0.2.0/pre-migration.py", line 95, in migrate
    rename_mrp_workorder(cr)
  File "/var/lib/odoo/openupgrade/10.0/addons/mrp/migrations/10.0.2.0/pre-migration.py", line 67, in rename_mrp_workorder
    cr, [('mrp.production.workcenter.line', 'mrp.workorder')]
  File "/var/lib/odoo/openupgrade/openupgradelib/openupgradelib/openupgrade.py", line 557, in rename_models
    (AsIs(_new), AsIs(_old), old),
  File "/var/lib/odoo/openupgrade/10.0/server/odoo/sql_db.py", line 154, in wrapper
  File "/var/lib/odoo/openupgrade/10.0/server/odoo/sql_db.py", line 231, in execute
IntegrityError: duplicate key value violates unique constraint "ir_model_data_module_name_uniq"
DETAIL:  Key (name, module)=(field_mrp_workorder_id, mrp) already exists.
```

leio: Solved that one with
```
openupgrade.logged_query(
    cr,
    "DELETE FROM ir_model_data WHERE model = %s AND module = %s and name LIKE %s",
    ('ir.model.fields', 'mrp', 'field_mrp_workorder_%',)
)
```

but clearly it could use a helper. Additionally there's old properties that might remain stale and potentially clash with new models with the same name, which could use proper removing for model deletion. As this wasn't a case with old mrp.workorder view, I didn't look into doing that here; got past this now and completed a migration test finally again with these two additional DELETE FROM commands.

leio:
I think production final product moves (stock.move's that have production_id set) should get their quantity_done_store fixed up. At least all such moves that are in state='done' should get their quantity_done_store set to product_uom_qty. That's because otherwise mrp.production qty_produced computed field is rather wrong otherwise. There might be more work in this area for full correctness in migration results...
Relatedly stock.move's that have raw_material_production_id should get a similar treatment, so that the "Consumed" column in MO view shows the correct thing

leio:
Similarly mrp.workorder qty_produced and qty_producing could use filling up correctly. All done and 0 producing in case of state done. But I suspect this (workorder produced/producing qties) is less important for common use cases compared to mine (I have the equivalent of multiple mrp.workcenter.productivity entries for one mrp.workorder in play in v8 already with custom mods)

leio:
I think old mrp.workorder `hour` should get migrated to `duration_expected` (multiplied by 60 due to hour vs minutes storage)

leio:
In v10 mrp.production state='confirmed' is followed by state='planned'. During transition workorders are created.
In v8 mrp.production state='confirmed' already has workorders.

Result with current lack of handling this is that a Manufacturing Order that was in confirmed state in v8 (or presumably v9) already has workorder_ids → user clicks "Create Workorders" to proceed → MO ends up with double the work orders, all of which need production recorded to get MO done.
In my use case I need to unlink all MO workorders that are in confirmed state, something like `env['mrp.production'].search([('state','=','confirmed'),('workorder_ids','!=',False)]).mapped('workorder_ids').unlink()`, which takes a LOT of time apparently. For generic case, it might be OK to instead assume workorders are created correctly from routing already and thus set MO state to planned if workorder_ids != False.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,424 @@
...
+
+mrp          / stock.move               / active_move_lot_ids (one2many): NEW relation: stock.move.lots
+mrp          / stock.move               / bom_line_id (many2one)        : NEW relation: mrp.bom.line
+mrp          / stock.move               / consume_unbuild_id (many2one) : NEW relation: mrp.unbuild
+mrp          / stock.move               / move_lot_ids (one2many)       : NEW relation: stock.move.lots
+mrp          / stock.move               / operation_id (many2one)       : NEW relation: mrp.routing.workcenter
+mrp          / stock.move               / quantity_done_store (float)   : NEW
+mrp          / stock.move               / unbuild_id (many2one)         : NEW relation: mrp.unbuild
+mrp          / stock.move               / unit_factor (float)           : NEW
+# NOTHING TO DO
```
leio: There is **much** to do here. If this isn't filled up, no manufacturing orders that were brought over will consume any raw materials, because the code in mrp.workorder record_production method won't have a unit_factor, and thus it skips raising consumed qty (quantity_done) when the workorder is done. At least the workorder_id filling is properly done in populate_stock_move_workorder_id, but without unit_factor filled, that isn't useful alone.

If unit_factor isn't calculated for it and filled in, the result is the following for MO's that were created in earlier Odoo version (and weren't in done state):

 - We record production, to keep it simple lets say it had nothing produced before and we produce it all now
 - record_production() on the last workorder (as stock.move workorder_id IS calculated to point at the last workorder of the production it's a raw material to consume for) will check all stock.moves for that production (via raw_move_ids one2many, the reverse of stock.move workorder_id) unit_factor, and raise quantity_done accordingly. quantity_done stays 0 because unit_factor is not set.
- The production is finished and user clicks "Mark as Done"
- mrp module action_done override is called and the production moves are funneled through move_validate() instead of main action_done()
- move_validate() checks what quantity_done is, and because it has remained zero due to missing unit_factor, it cancels the move instead (button_mark_done does that after post_inventory is done, which ended up in that move_validate)
- The raw material isn't consumed, the reservation is released by stock.move action_cancel, raw material moves are all in cancel state now
- All raw material stock is completely wrong (the quantity that was reserved for production is released back to quantity on hand), because they were obviously consumed in real life production, but Odoo has cancelled the moves instead.

leio (changes requested):

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,424 @@
...
+
+mrp          / procurement.order        / property_ids (many2many)      : DEL relation: mrp.property
+# NOTHING TO DO
+
+
+mrp          / stock.move               / consumed_for (many2one)       : DEL relation: stock.move
+# DONE: Pre-migration: Renamed to None and copied this column. That will be used in post-migration to known which moves are consumed for.
+
+mrp          / stock.move               / active_move_lot_ids (one2many): NEW relation: stock.move.lots
+mrp          / stock.move               / bom_line_id (many2one)        : NEW relation: mrp.bom.line
```
leio: This has to be filled as well, like also unit_factor https://github.com/OCA/OpenUpgrade/pull/1076#pullrequestreview-122561036

If this (or unit_factor) is not set, mrp.workorder record_production will NOT consume the move and it will cancel the move afterwards in button_mark_done, resulting in completely wrong quantities in stock for all the raw materials used in manufactoring orders that were created in the older version than v10..

leio (changes requested):

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: how? what is the relation?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: fill <code>unit_factor</code> with what?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: fill <code>bom_line_id</code> with what?

addons/mrp/migrations/10.0.2.0/post-migration.py
MiquelRForgeFlow: What do you mean by updating rules and picking types? What kind of change?

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
leio: I did `UPDATE mrp_routing_workcenter SET time_mode='manual', time_cycle_manual=hour_nbr * 60;`, because old data was entered in terms of manual prediction, without any "10 last work orders average" kind of logic, but that might be OK to leave time_mode to automatic (I just knew my customer wants manual left in place). So I think it should be hour_nbr * 60, but it might need to be multiplied or divided by cycle_nbr or something to get the same conceptual result. For me cycle_nbr data was always 1, so I didn't look into it deeply, but the migration from hour_nbr was critical.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
leio: With the unit factor as it is filled upon new workorder generation for fresh Manufacturing Orders..

I fixed it up for my needs after migration with the following odoo shell script:
```
moves_to_fix = env['stock.move'].search([('raw_material_production_id','!=',False),('state','not in',['cancel','done']),('unit_factor','=',False)])
for move in moves_to_fix:
    move.unit_factor = move.product_uom_qty / ((move.raw_material_production_id.product_qty - move.raw_material_production_id.qty_produced) or 1.0)
```

Something like that needs doing during migration. Maybe should bother with cancel or done moves as well, in case they get reset at some point, but the workflow used in my instance didn't have a need for it. If it is done for cancel/done moves as well, it'll take a significant amount of time if done via odoo methods; if with pure SQL, should be very fast (but potentially different rounding that needs to be cared for maybe?).

If this is not done, all raw materials used by already generated (prior to migration) MO-s would go to cancel after MO is marked as done, completely breaking stock quantities.

addons/mrp/migrations/10.0.2.0/openupgrade_analysis_work.txt
leio: Pretty much anything, otherwise as with unit_factor, the stock will go completely wrong when MO is marked as done. The (questionable) check only checks if it's set or not; if not set, it'll not increase quantity_done and end up cancelled in button_mark_done.

In my case I post-processed by checking through all BoM lines of the MO to find the same product - if only one is found (there could be multiple lines for same product), I assigned that to bom_line_id. But that will not solve all of them, as BoM might have changed in time, multiple same products found, etc. To finish it up, I just assigned what is left to basically any bom line of the MO's BoM - in practice it only matters if MO produced quantity is updated via wizard, and my customer doesn't use that. Multiple raw material stock moves of the same MO-s pointing at the same bom line is fine as well, with the same caveat of update wizard not working right. So it's under the understanding that MO Produced Qty Update wizard is not going to be used for old MOs, and then all is good.
That said, v10 code is still rather questionable - nothing stops you from removing a BoM line after it's already assigned to a stock.move, thus breaking the on hand quantities later on as well; for that an issue should be filed against Odoo and at the very least stock.move:bom_line_id made ondelete='restrict' instead of the current default SET NULL. I made that change in my own module for now with issue filing as TODO :(

addons/mrp/migrations/10.0.2.0/post-migration.py
leio: I had to do the following, otherwise automatically created MO-s take WH/IN/ sequence, not MO sequence:
`UPDATE procurement_rule SET picking_type_id=7 WHERE action='manufacture';` where 7 is the ID of stock_picking_type record with code='mrp_operation', which gets created during migration at some point (per earlier follow-up fixes that are now part of this PR already) -- manufacture procurement rule has to use the correct picking type, as-is it is wrong (unset) and the code ends up with an incoming picking type, taking sequence from WH/IN/

MiquelRForgeFlow:
Hey, everybody! This PR is GREEN again. So, I think it would be nice if I squash a bit and we do merge of this PR. Then, if something can be improved, it could be done in another PR instead.

@pedrobaeza as you can see in premigration, I think that maybe a generalization of `delete_old_workorder_model` could be implemented in `openupgradelib`.

@leio I hope I implemented correctly what you said.

AaronHForgeFlow (approved):
:+1:  Working fine from my side. Functional test excluding work.orders cause I do not use them:
- mrp.production is good: including consumed and production moves. New picking type and the picking type in the procurement group updated fine.

- bom migrations seems also good: and the bom_line relation in the stock move is correct.

- The relation between the consumed quants and the stock move is correct IMO.

Thanks

pedrobaeza (approved):

MiquelRForgeFlow:
@pedrobaeza recently I thought that maybe we could rename the old `mrp.workorder` (model and table) to `mrp.workorder.report` for example instead of using the `delete_old_workorder_model`. What do you think?

pedrobaeza:
That can be a good option. I was waiting to Travis to merge, but if you make this change, I can wait more.

MiquelRForgeFlow:
Ok, I will make then.

pedrobaeza:
Any way, as `mrp.workorder.report` is an SQL view, I'm not sure if there's going to be a problem.

pedrobaeza:
Merging this as a valid one, and waiting for other PR with the change to see if there's a problem.

MiquelRForgeFlow:
Ok :)

PR #1077, opened 2017-09-20
[FIX] account: Change journal entry date when current date not in period

pedrobaeza:
As journal entries in v8 takes the invoice date, if you forced another period for that entry, when moving to v9 you will have an incorrect date on the journal entry, and that is now the only reference, so we need to change that date to for example the start date of the forced period.

cc @Tecnativa

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -624,6 +624,31 @@ def update_account_invoice_date(cr):
     )
 
 
+def update_move_date(cr):
+    """Update journal entries date when the date is not inside the indicated
+    period for respecting the accounting period on v9.
+    """
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE account_move am
+        SET date = ap.date_start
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Ready to be reviewed.

sergio-teruel (approved):

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -717,6 +742,7 @@ def migrate(env, version):
     fill_move_line_invoice(cr)
     merge_invoice_journals(env)
     update_account_invoice_date(cr)
+    update_move_date(cr)
```
MiquelRForgeFlow, line 745: Shouldn't be this line before than the previous one? :thinking:

pedrobaeza: It doesn't matter

MiquelRForgeFlow: Ok

MiquelRForgeFlow (approved):

PR #1078, opened 2017-09-21
[FIX] mass_mailing_sending_queue: Delete cron noupdate job

pedrobaeza:
This module includes a noupdate record for a cronjob. If we don't remove it, then on migrated DB there will be a warning about an inexisting model, because these kind of records are not auto-removed on update.

cc @Tecnativa

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/post-migration.py
@@ -77,6 +77,14 @@ def publish_attachments(env):
     )
 
 
+def cleanup_modules_post(env):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@cubells take note for other possible module mergings in the future.

cubells:
Sure! And thank you for your guidance...

pedrobaeza:
@cubells It needs to be done on post migration as ir.model.data is still not available in pre, so it's less tied to the original method, but well, it works also.

pedrobaeza:
Ready to be reviewed

cubells (approved):
:+1: 

LGTM

PR #1084, opened 2017-09-26
[IMP] account: Merge supplier invoice references

pedrobaeza:
In v8, there are 2 fields for writing references: supplier_invoice_number and reference. Now in v9 there's only the last one. We merge the first field content in the second one for avoiding data loss. Note that previously the `reference` field has been copied for preserving the original field contents.

@Tecnativa

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -2,7 +2,9 @@
 # © 2016 Sylvain LE GAL <https://twitter.com/legalsylvain>
 # © 2016 Serpent Consulting Services Pvt. Ltd.
 # © 2016 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
 from openupgradelib import openupgrade
 
 column_renames = {
@@ -41,6 +43,9 @@ column_copies = {
         ('type_tax_use', None, None),
         ('type', None, None),
     ],
+    'account_invoice': [
+        ('reference', None, None),
+    ]
 }
 
 table_renames = [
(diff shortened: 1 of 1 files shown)


addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -212,6 +217,31 @@ def blacklist_field_recomputation(env):
     ]
 
 
+def merge_supplier_invoice_refs(env):
```
sergio-teruel, line 220: This method in pre or post??

addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -212,6 +217,31 @@ def blacklist_field_recomputation(env):
...
+    """In v8, there are 2 fields for writing references:
+    supplier_invoice_number and reference. Now in v9 there's only the last one.
+    We merge the first field content in the second one for avoiding data loss.
+    Note that previously the `reference` field has been copied for preserving
+    the original field contents.
+    """
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_invoice
+        SET reference = supplier_invoice_number || ' - ' || reference
```
sergio-teruel, line 230: SET reference = supplier_invoice_number || ' - ' || %s ......
 % openupgrade.get_legacy_name('reference') ??

sergio-teruel (changes requested):

addons/account/migrations/9.0.1.1/pre-migration.py
pedrobaeza: It can fit in both, but better sooner than later.

addons/account/migrations/9.0.1.1/pre-migration.py
pedrobaeza: No, the column exists without legacy name.

sergio-teruel: but here https://github.com/OCA/OpenUpgrade/pull/1084/files#diff-050c291be549d31c2482669f8281e619R46 you are renaming the column...

pedrobaeza: No, I'm copying it.

sergio-teruel: Ok

sergio-teruel (approved):

Issue #1046, opened 2017-08-25
[RFC] Detect removed noupdate records


pedrobaeza:
noupdate="1" records are not automatically removed on any update/migration, and they can be very harmful if they are for example record rules (example: https://github.com/OCA/OpenUpgrade/pull/848).

We should detect them and maybe automatically delete them. Ideal would be on OpenUpgrade itself, in a similar way we have made with binary stored as attachments: https://github.com/OCA/OpenUpgrade/pull/1013.

If not possible, then at least we can detect them in the `compare_noupdate_xml_records.py` script. There's a `<delete>` tag for XML that maybe we can use for automatic removal loading the XML file.

Opinions @StefanRijnhart @hbrunn @mreficent ?

MiquelRForgeFlow:
I love automatic things :+1:

pedrobaeza:
Any opinion/advice for implementing this?

StefanRijnhart:
I'm all for reporting on the noupdate property in the analysis file (working on that now), but removing them by default would be unfortunate (for example, the default service product is in use by some of my customers, but it has been removed in Odoo 9.0).

pedrobaeza:
OK, then let's go for the manual one. About the issue you mention, we can study it for solving all cases. Maybe an unlink encapsulated in a try... Can you open an issue for it?

StefanRijnhart:
@pedrobaeza I am not particularly interested in automatic removal, so I won't open the issue.

pedrobaeza:
Well I was talking about the concrete issue of the default service product, but if you have solved it...

StefanRijnhart:
Oh, sorry I misunderstood. The service product will be kept exactly because it is a noupdate record. I gave it as an example why it might be a bad idea to delete such records by default when they are removed in a new Odoo release.

StefanRijnhart:
Here is the analysis with checking for the noupdate property. Is that what you had in mind? https://github.com/StefanRijnhart/OpenUpgrade/commit/a524b8362ee9d1d548009b75d1e13bd6452403ef

pedrobaeza:
Yeah, that's it! With that, we can see from the analysis file which records have to be taken into account. We still need to keep the script for getting the XML, but this way we can compare 2 ways. Thanks for that.

StefanRijnhart:
Well, thanks for suggesting this feature. I will create a PR with this change as soon as the current analysis and framework PR's are merged.

[...]

PR #1086, opened 2017-09-28
[ADD] Analysis: store noupdate property of xml records

StefanRijnhart:
Store the noupdate property now, so that we can report on it in the 11.0 analysis. 
https://github.com/OCA/OpenUpgrade/issues/1046

### DIFF
--- a/odoo/addons/openupgrade_records/models/analysis_wizard.py
+++ b/odoo/addons/openupgrade_records/models/analysis_wizard.py
@@ -64,6 +64,7 @@ class AnalysisWizard(models.TransientModel):
         res = compare.compare_sets(remote_records, local_records)
 
         # Retrieve xml id representations and compare
+        # In version 11, add 'noupdate' as a field to retrieve and compare
         fields = ['module', 'model', 'name']
         local_xml_records = [
             dict([(field, record[field]) for field in fields])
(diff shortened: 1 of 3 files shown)


odoo/addons/openupgrade_records/models/analysis_wizard.py
```diff
@@ -64,6 +64,7 @@ def write_file(module, version, contents,
         res = compare.compare_sets(remote_records, local_records)
 
         # Retrieve xml id representations and compare
+        # In version 11, add 'noupdate' as a field to retrieve and compare
```
pedrobaeza, line 67: Why this comment here? Is it a TODO?

pedrobaeza:
@StefanRijnhart what about #1014?

odoo/addons/openupgrade_records/models/analysis_wizard.py
StefanRijnhart: Yes, it is a TODO for when this is ported to OpenUpgrade 11. If we add the field now, it will break the analysis between 9 and 10 because 9 does not have this code.

pedrobaeza: Then maybe you can add the `TODO` or `FIXME` word before.

StefanRijnhart: Oh come on @pedrobaeza.

pedrobaeza: OK, let's keep it as is, but if we won't remember to change this on v11, it's not my fault :stuck_out_tongue_winking_eye:

pedrobaeza (approved):

PR #1088, opened 2017-09-30
[9.0][FIX] account: Set default inbound and outbound payment methods

sergio-teruel:
cc @Tecnativa
The account_payment_mode module from bank-payment repo updates the default payment mode for this fields so i need update these module too to set all payment_methods.

default methods in account module:
https://github.com/OCA/OCB/blob/9.0/addons/account/models/account.py#L203

default methods in account_payment_mode:
https://github.com/OCA/bank-payment/blob/10.0/account_payment_mode/models/account_journal.py#L11

Please @pedrobaeza @carlosdauden Can you review?

--

### DIFF
--- a/addons/account/migrations/9.0.1.1/end-migration.py
+++ b/addons/account/migrations/9.0.1.1/end-migration.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def update_journal_payment_methods(env):
+    """Update default payment methods in journals when type is 'bank' or
+    'cash'.
+    """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As this is totally new, there's no reference of what are the expected allowed payment methods, so you have to allow of them IMO. The steps are:

* Put an end-migration.py script, for covering the case where other modules create payment methods (which happens on OCA/bank-payment suite).
* Assign all the available outbound and inbound payment methods to each journal.
* You have to make the same for cash journals.

sergio-teruel:
@pedrobaeza changes done in end-migration script. I don't know about this option.

pedrobaeza:
Sergio, sorry for the confusion, but end-migration scripts are only allowed on >=10.0. For 9.0, you have to put the code in https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/openupgrade/deferred_90.py

pedrobaeza:
It finally seems that there is end migration scripts on v9, so I can merge. Thanks for the contribution.

PR #1090, opened 2017-09-30
11.0 analysis

StefanRijnhart:


StefanRijnhart:
@pedrobaeza I'll cover the new analysis features you have been promoting in a follow up PR

pedrobaeza (approved):
This seems OK as it's the result of the output of the module, so I merge more or less blindly. When we start to analyze each file, if we find any issue then we will come back to this.

@StefanRijnhart , you can continue with next PR about noupdate records, thanks.

PR #1091, opened 2017-09-30
[FIX] website_blog: Change opacity for cover images

pedrobaeza:
Switch opacity value for posts, as now in v10 the meaning of that
property is for an upper solid frame, not for the image itself:

* opacity=1.0 in v9 means solid image
* opacity=1.0 in v10 means solid color frame.

So what we do is to get the inverse percentage of the opacity: abs(1 -opacity).

@Tecnativa


OSevangelist:
I haven't tested it with an actual migration but that change seems reasonable (and after all relatively small) to me. So +1 from me for merging the PR

cubells (approved):
:+1: 

LGTM

addons/website_blog/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+from odoo.tools import safe_eval
+
+
+def convert_blog_post_cover(env):
+    """Switch opacity value for posts, as now in v10 the meaning of that
+    property is for an upper solid frame, not for the image itself.
+    """
+    post_obj = env['blog.post'].with_context(active_test=False)
+    for post in post_obj.search([('cover_properties', '!=', False)]):
+        props = safe_eval(post.cover_properties)
```
hbrunn, line 15: upstream uses JSON here, so I think you should do the same: https://github.com/OCA/OCB/blob/10.0/addons/website_blog/controllers/main.py#L166

addons/website_blog/migrations/10.0.1.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+        if 'opacity' in props:
+            props['opacity'] = str(abs(1 - float(props['opacity'])))
+        post.cover_properties = str(props).replace("'", '"')
```
hbrunn, line 18: and then here dumps

pedrobaeza:
OK, you're right. I'll do the PR changing it.

PR #1093, opened 2017-10-02
Analysis/11.0 attachment property

StefanRijnhart:
Log if binary fields are stored as an attachment.


pedrobaeza (approved):
We have 2 changes! Are we going to make it on the base end-migration?

StefanRijnhart:
@pedrobaeza not sure what you mean

pedrobaeza:
On v10, we use a deferred solution: https://github.com/OCA/OpenUpgrade/pull/1013, so I'm asking if we are going to use the same solution. Merging this any way.

PR #1095, opened 2017-10-02
[11.0][ADD] renames

MiquelRForgeFlow:
Updated apriori and documentation with the module renames.


odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -3,6 +3,12 @@
 """
 
 renamed_modules = {
+    # Odoo
+    'base_action_rule': 'base_automation',
+    'crm_project_issue': 'crm_project',
+    'stock_picking_wave': 'stock_picking_batch',
+    'website_issue': 'website_form_project',
```
pedrobaeza, line 10: Are you sure this is renaming? I see it more as merging, specially the last one

StefanRijnhart: Same thing as far as the analysis is concerned. I'd like to have all the merged modules in here.

StefanRijnhart: So this list is not complete but it's a start.

MiquelRForgeFlow:
@pedrobaeza , as @StefanRijnhart suggested, I created two different lists.

odoo/addons/openupgrade_records/lib/apriori.py
MiquelRForgeFlow: Now, separated in two lists, I think they are complete.

StefanRijnhart:
Can you change https://github.com/OCA/OpenUpgrade/blob/11.0/odoo/addons/openupgrade_records/lib/compare.py#L18 so that it also takes merged modules into account?

MiquelRForgeFlow:
sure

StefanRijnhart:
@rruebner can you call https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.update_module_names on the modules in these lists in your base module pre-migrate script? like we do here: https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/base/migrations/10.0.1.3/pre-migration.py#L16, but then also for the merge modules.

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -15,7 +15,8 @@
 
 
 def module_map(module):
-    return apriori.renamed_modules.get(module, module)
+    return apriori.renamed_modules.get(module, module) or \
+           apriori.merged_modules.get(module, module)
```
StefanRijnhart: the first clause will always return `module`. Try
```
return apriori.renamed_modules.get(
    module,
    apriori.merged_modules.get(module, module))
```

MiquelRForgeFlow:
Done!

MiquelRForgeFlow:
ping: @pedrobaeza

StefanRijnhart:
Thanks! Could you also hunt for renamed models by comparing new and obsolete models in odoo/addons/base/migrations/openupgrade_general_log.txt?

MiquelRForgeFlow:
Ok, but better do it in an other PR, right?

StefanRijnhart:
No, I'd like it in this PR so that I can regenerate the analysis before this PR is merged and give you feedback based on that.

MiquelRForgeFlow:
ah, ok

StefanRijnhart:
Thanks! I'll start an analysis with this now.

StefanRijnhart:
See here for the followup: https://github.com/OCA/OpenUpgrade/pull/1102

PR #1098, opened 2017-10-02
[11.0][MIG] base_action_rule -> base_automation

yvaucher:
Migration of base_action_rule into base_automation

Changes models, table and one XMLID

--

### DIFF
--- a/addons/base_automation/migrations/11.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/base_automation/migrations/11.0.1.0/openupgrade_analysis_work.txt
@@ -1,0 +1,138 @@
+---Fields in module 'base_automation'---
+base_automation / base.automation          / _inherits (False)             : NEW  -> warning delegate field on action_server_id this makes it an _intherits
+base_automation / base.automation          / action_server_id (many2one)   : NEW relation: ir.actions.server, required: required
+base_automation / base.automation          / active (boolean)              : NEW 
+base_automation / base.automation          / filter_domain (char)          : NEW 
+base_automation / base.automation          / filter_pre_domain (char)      : NEW 
+base_automation / base.automation          / last_run (datetime)           : NEW 
+base_automation / base.automation          / on_change_fields (char)       : NEW 
+base_automation / base.automation          / trg_date_calendar_id (many2one): NEW relation: resource.calendar
+base_automation / base.automation          / trg_date_id (many2one)        : NEW relation: ir.model.fields
+base_automation / base.automation          / trg_date_range (integer)      : NEW 
+base_automation / base.automation          / trg_date_range_type (selection): NEW selection_keys: ['day', 'hour', 'minutes', 'month']
+base_automation / base.automation          / trigger (selection)           : NEW required: required, selection_keys: ['on_change', 'on_create', 'on_create_or_write', 'on_time', 'on_unlink', 'on_write']
+base_automation / base.automation.lead.test / active (boolean)              : NEW 
+base_automation / base.automation.lead.test / date_action_last (datetime)   : NEW 
+base_automation / base.automation.lead.test / is_assigned_to_admin (boolean): NEW 
+base_automation / base.automation.lead.test / line_ids (one2many)           : NEW relation: base.automation.line.test
+base_automation / base.automation.lead.test / name (char)                   : NEW required: required
+base_automation / base.automation.lead.test / partner_id (many2one)         : NEW relation: res.partner
+base_automation / base.automation.lead.test / priority (boolean)            : NEW 
+base_automation / base.automation.lead.test / state (selection)             : NEW selection_keys: ['cancel', 'done', 'draft', 'open', 'pending']
+base_automation / base.automation.lead.test / user_id (many2one)            : NEW relation: res.users
+base_automation / base.automation.line.test / lead_id (many2one)            : NEW relation: base.automation.lead.test
+base_automation / base.automation.line.test / name (char)                   : NEW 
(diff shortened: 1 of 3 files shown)


odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -3,6 +3,7 @@
 """
 
 renamed_modules = {
+    'base_action_rule': 'base_automation',
```
yvaucher: This will depends on ~~#1095~~ #1102

lepistone (approved):

addons/base_automation/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,40 @@
...
+base_automation / base.automation.lead.test / line_ids (one2many)           : NEW relation: base.automation.line.test
+base_automation / base.automation.lead.test / name (char)                   : NEW required: required
+base_automation / base.automation.lead.test / partner_id (many2one)         : NEW relation: res.partner
+base_automation / base.automation.lead.test / priority (boolean)            : NEW 
+base_automation / base.automation.lead.test / state (selection)             : NEW selection_keys: ['cancel', 'done', 'draft', 'open', 'pending']
+base_automation / base.automation.lead.test / user_id (many2one)            : NEW relation: res.users
+base_automation / base.automation.line.test / lead_id (many2one)            : NEW relation: base.automation.lead.test
+base_automation / base.automation.line.test / name (char)                   : NEW 
+base_automation / base.automation.line.test / user_id (many2one)            : NEW relation: res.users
+base_automation / ir.actions.server        / usage (False)                 : NEW selection_keys: ['base_automation', 'ir_actions_server', 'ir_cron'], mode: modify
```
StefanRijnhart, line 26: Need to set usage to base_automation for actions that are linked to an action rule

addons/base_automation/migrations/11.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,33 @@
...
+from openupgradelib import openupgrade
+
+
+model_renames_base_action_rule = [
+    ('base.action.rule', 'base.automation')
+]
+
+table_renames_base_action_rule = [
+    ('base_action_rule', 'base_automation')
+]
```
yvaucher, line 22: @StefanRijnhart 
are model and table rename deprecated by apriori ?

https://github.com/OCA/OpenUpgrade/blob/11.0/odoo/addons/openupgrade_records/lib/apriori.py#L28

StefanRijnhart: No, we leave the renames to the originating modules. Apriori is to improve the analysis.

yvaucher: Ok

pedrobaeza:
@yvaucher can you please finish this PR?

addons/base_automation/migrations/11.0.1.0/openupgrade_analysis_work.txt
yvaucher: Not sure what this usage field is, I'll check that tomorrow

yvaucher:
@StefanRijnhart @pedrobaeza I made some advances on this.

I didn't expect that _inherits implied by the attribute `delegate` so I had to create `ir.actions.servers records`

Not sure if we want to drop the unused columns from table `base_automation` that I copy to `ir_act_server`.

And any idea of what to do with the 3 dropped fields? seems to me that we can't do much with them the methods of `base.automation` have been simplified.

addons/base_automation/migrations/11.0.1.0/openupgrade_analysis_work.txt
yvaucher: In fact we needs to create a ir.actions.server for each base.automation
The field server_action_id being set as delegate=True

addons/base_automation/migrations/11.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,104 @@
...
+        'code': DEFAULT_PYTHON_CODE,
+        'state': 'object_write',
+        'type': 'ir.actions.server',
+        'usage': 'base_automation',
+    }
+    cr.execute("""
+        SELECT id, create_uid, create_date, write_uid, write_date
+          name, sequence, model_id, kind
+        FROM base_automation
+        WHERE action_server_id IS NULL
```
njeudy: Missing ";" ?

pedrobaeza: Not needed on pyscopg2 queries

pedrobaeza:
I'm going to check pending things in this for finishing it.

pedrobaeza:
> Not sure if we want to drop the unused columns from table base_automation that I copy to ir_act_server.

We should keep them for other migration scripts (even for custom modules) that make use of them for other purposes. Module database_cleanup can be used later for dropping the garbage (columns, fields, tables, models...)

> And any idea of what to do with the 3 dropped fields? seems to me that we can't do much with them the methods of base.automation have been simplified.

act_user_id is very specific and no special problem in not being transferred. If someone is using it for anything special, can take the SQL column. The other 2 serves for filtering records, and we should dump the domain in the filter to the domains fields (I have already done it).

pedrobaeza (approved):

yvaucher:
@pedrobaeza thanks

pedrobaeza:
@yvaucher there's more work in #1303

PR #1100, opened 2017-10-02
[11.0][MIG][WIP] base

rruebner:
Added first part / try of base module migration, it is not done yet. Missing parts are marked with "TODO" in openupgrade_analysis_work.txt.

--

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
+++ b/odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
@@ -63,7 +63,7 @@ base         / ir.actions.server        / action_id (many2one)          : DEL re
 
 base         / ir.actions.server        / binding_model_id (many2one)   : NEW relation: ir.model
 base         / ir.actions.server        / binding_type (selection)      : NEW required: required, selection_keys: ['action', 'report'], req_default: function
-# NOTHING TO DO
+# DONE: post-migration: map value 'menu_ir_values_id' with the help of 'model_id' to binding_model_id
 
 base         / ir.actions.server        / condition (char)              : DEL 
 base         / ir.actions.server        / copyvalue (char)              : DEL 
@@ -76,8 +76,7 @@ base         / ir.actions.server        / ref_object (reference)        : DEL
 # DONE: pre-migration: rename column
 
 base         / ir.actions.server        / state (selection)             : selection_keys is now '['code', 'multi', 'object_create', 'object_write']' ('function')
-# DONE: pre-migration: copy column
-# TODO: post-migration: map values 'trigger' and 'client_action' to default value
+# NOTHING TO DO, invalid options will be removed by migration of ir.actions.server 
 
 base         / ir.actions.server        / sub_model_object_field (many2one): DEL relation: ir.model.fields
 base         / ir.actions.server        / sub_object (many2one)         : DEL relation: ir.model
@@ -118,7 +117,7 @@ base         / ir.cron                  / function (char)               : DEL
 
 base         / ir.cron                  / interval_type (selection)     : selection_keys is now '['days', 'hours', 'minutes', 'months', 'weeks']' ('['days', 'hours', 'minutes', 'months', 'weeks', 'work_days']')
 # DONE: pre-migration: copy column
-# TODO: post-migration: map value 'work_days' to default value
+# TODO: post-migration: map value 'work_days' to 'days', because changing from work_days to month (default) is maybe not the best idea
 
 base         / ir.cron                  / ir_actions_server_id (many2one): NEW relation: ir.actions.server, required: required
 # TODO: Check which ir.actions.server should be used, because the field is required
--- a/odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
+++ b/odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
@@ -69,7 +69,7 @@ base         / ir.actions.server        / state (selection)             : select
 base         / ir.actions.server        / usage (char)                  : now required, default = function
 base         / ir.actions.server        / usage (char)                  : selection_keys is now '['ir_actions_server', 'ir_cron']' ('False')
 base         / ir.actions.server        / usage (char)                  : type is now 'selection' ('char')
-# TODO: post-migration: 'ir_cron' value is related to new functionality, so set existing records to 'ir_actions_server'
+# NOTHING TO DO, manually check needed after migration, 'ir_cron' value is related to new functionality, so set existing records to 'ir_actions_server'
 
 base         / ir.actions.todo          / groups_id (many2many)         : DEL relation: res.groups
 # NOTHING TO DO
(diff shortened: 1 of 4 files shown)


odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,429 @@
...
+base         / ir.actions.server        / binding_model_id (many2one)   : NEW relation: ir.model
+base         / ir.actions.server        / binding_type (selection)      : NEW required: required, selection_keys: ['action', 'report'], req_default: function
+# NOTHING TO DO
+
+base         / ir.actions.server        / condition (char)              : DEL 
+base         / ir.actions.server        / copyvalue (char)              : DEL 
+base         / ir.actions.server        / id_object (reference)         : DEL 
+base         / ir.actions.server        / id_value (char)               : DEL 
+base         / ir.actions.server        / link_new_record (boolean)     : DEL 
+base         / ir.actions.server        / menu_ir_values_id (many2one)  : DEL relation: ir.values
```
StefanRijnhart, line 52: For actions with a menu_ir_values_id, set `binding_model_id` to action.model_id.id and `binding_type` to `action`. Delete the ir.values entry.

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,429 @@
...
+base         / ir.actions.server        / model_object_field (many2one) : DEL relation: ir.model.fields
+base         / ir.actions.server        / ref_object (reference)        : DEL 
+# DONE: pre-migration: rename column
+
+base         / ir.actions.server        / state (selection)             : selection_keys is now '['code', 'multi', 'object_create', 'object_write']' ('function')
+# DONE: pre-migration: copy column
+# TODO: post-migration: map values 'trigger' and 'client_action' to default value
```
StefanRijnhart: Leave `trigger` value, so that it can be identified as invalid (there are no workflows anymore in Odoo 11). Leave client_action entries as well. They should be removed by the migration of ir.actions.server, and other entries should be recognizable as obsolete.

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,429 @@
...
+base         / ir.cron                  / _inherits (False)             : NEW 
+# NOTHING TO DO
+
+base         / ir.cron                  / args (text)                   : DEL 
+base         / ir.cron                  / function (char)               : DEL 
+# DONE: pre-migration: rename column
+
+base         / ir.cron                  / interval_type (selection)     : selection_keys is now '['days', 'hours', 'minutes', 'months', 'weeks']' ('['days', 'hours', 'minutes', 'months', 'weeks', 'work_days']')
+# DONE: pre-migration: copy column
+# TODO: post-migration: map value 'work_days' to default value
```
StefanRijnhart: How about mapping 'work_days' to 'days', as the default value is 'months'?

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
rruebner: Done.

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
rruebner: Done.

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
rruebner: Updated todo text.

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,428 @@
...
+base         / res.country              / address_view_id (many2one)    : NEW relation: ir.ui.view
+base         / res.country              / name_position (selection)     : NEW selection_keys: ['after', 'before']
+base         / res.country              / vat_label (char)              : NEW 
+base         / res.currency             / currency_subunit_label (char) : NEW 
+base         / res.currency             / currency_unit_label (char)    : NEW 
+# NOTHING TO DO
+
+base         / res.currency.rate        / name (datetime)               : type is now 'date' ('datetime')
+# DONE: pre-migration: copy column
+# DONE: pre-migration: change field type
```
StefanRijnhart: I can't find where this happens. Am I missing something?

StefanRijnhart:
Making some suggestions here: https://github.com/rruebner/OpenUpgrade/pull/1

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
rruebner: :) your are right. Its not done yet.

rruebner:
@StefanRijnhart PR reviewed and merged, added 'default' ir.values merging into ir.default

StefanRijnhart:
Thank you!

StefanRijnhart (approved):

odoo/addons/base/migrations/11.0.1.3/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,399 @@
...
+# NOTHING TO DO
+
+base         / ir.actions.act_url       / usage (char)                  : DEL 
+base         / ir.actions.act_window_close / usage (char)                  : DEL 
+base         / ir.actions.actions       / usage (char)                  : DEL 
+base         / ir.actions.client        / usage (char)                  : DEL 
+# NOTHING TO DO, obsolete column
+
+base         / ir.actions.act_window    / binding_model_id (many2one)   : NEW relation: ir.model
+base         / ir.actions.act_window    / binding_type (selection)      : NEW required: required, selection_keys: ['action', 'report'], req_default: function
```
hbrunn, line 13: there's a lot to do for this: This fields (actually, this happens in the action's base model) replace the horrible ir.values model to link actions to models for the sidebar. You'll have to fill those values from the appropriate ir.values record

odoo/addons/base/migrations/11.0.1.3/post-migration.py
```diff
@@ -0,0 +1,92 @@
...
+import json
+
+from odoo.tools import pickle
+
+from openupgradelib import openupgrade
+
+from psycopg2.extensions import AsIs
+
+
+def map_ir_actions_server_fields(cr):
```
hbrunn, line 13: as written above, you need to do this for all ir.actions.* models. Don't rely on the link from the action to the ir.value record, do it the other way around via https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_values.py#L95, possibly filtering for key and key2

hbrunn (changes requested):

odoo/addons/base/migrations/11.0.1.3/post-migration.py
StefanRijnhart: @hbrunn I addressed your comments in #1121 (except that ir.values' action_id is almost never filled and when it is, the key already provides the action reference as it does in all other instances)

StefanRijnhart:
Closing as review can continue on #1121

PR #1102, opened 2017-10-02
11.0 add renames mreficent

StefanRijnhart:
Slightly modified version of https://github.com/OCA/OpenUpgrade/pull/1095 plus the effects on the analysis


StefanRijnhart:
BTW I took out the merge of website_portal_sale with sale even if they have a shared entry in the analysis. The reason is that the former module shares entries with other modules too, such as account. This resulted in a misleading analysis that implied that fields moved from sale to account.

MiquelRForgeFlow (approved):

StefanRijnhart:
Merging after @mreficent's approval

PR #1111, opened 2017-10-03
[11.0][MIG] product

Benniphx:

Current behavior before PR:

--


addons/product/migrations/11.0.1.2/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+            """, (product_rows[0], r[0]))
+            # 'copy' existing product_packaging for additional
+            # product_product records
+            for p in product_rows[1:]:
+                fields = values = column_names
+                cr.execute("""
+                    INSERT INTO product_packaging %s
+                    SELECT %s
+                    FROM product_packaging WHERE id = %s;
+                """, (tuple(fields.append('product_id')),
```
StefanRijnhart: Now 'product_id' is also added to fields and column_names. Try `fields + ['product_id']`

addons/product/migrations/11.0.1.2/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+                      ','.join(values.append(p[0])), r[0]))
```
StefanRijnhart: The lists of fields and values are instantiated as a single element in the SQL.

>>> session.env.cr.mogrify("SELECT %s FROM res_partner", (','.join(['id', 'name']),))
"SELECT 'id,name' FROM res_partner"

You'll want something like

```
cr.execute(
     """
    INSERT INTO product_packaging %s
    SELECT %s
    FROM product_packaging WHERE id = %%s""" % (
        ','.join(fields + ['product_id']), ','.join(values + [p[0]])),
    (r[0],))
```

StefanRijnhart:
@benniphx will you be able to have look at the requests (and rebase on the main branch, so that tests are run on this PR)? Otherwise, let me know.

pedrobaeza:
@Benniphx are you going to finish this PR or should we take it over?

Benniphx:
I will continue today

StefanRijnhart:
Needs a rebase because of the new column length in the module status table, and a flake8 fix (see travis output), otherwise good!

StefanRijnhart (changes requested):
See previous comment

Benniphx:
I am working on it and I have to change a part in sql function as well
I am on it

odoo/openupgrade/doc/source/modules100-110.rst
```diff
@@ -1,5 +1,5 @@
-Module coverage 10.0 -> 11.0
-============================
+Module coverage 9.0 -> 10.0
```
MiquelRForgeFlow: Why this change?

Benniphx: ah this was from the rebase i thought i changed it back

MiquelRForgeFlow:
Please rebase and squash

addons/product/migrations/11.0.1.2/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+            # 'copy' existing product_packaging for additional
+            # product_product records
+            for p in product_rows[1:]:
+                fields = values = column_names
+                cr.execute("""
+                   INSERT INTO product_packaging %s
+                   SELECT %s
+                   FROM product_packaging WHERE id = %%s""" % (
+                    ','.join(fields + ['product_id']),
+                    ','.join(values + ['product_id'])), (r[0],))
```
Benniphx: @StefanRijnhart  I think i should use something like [p[0]] instead of ['product_id'] in the  ','.join(values + ['product_id'])), (r[0],)) line

but @rruebner thinks this will not work. Maybe a hint? or idear

Benniphx: ok Travis said it is ok so i changed it to [p[0]]

Benniphx:
It was rebased and squashed and is now working correctly

StefanRijnhart (approved):

PR #1103, opened 2017-10-03
[11.0][MIG] delivery

yvaucher:
Description of the issue/feature this PR addresses:

Migration of delivery module.

_inherits an product.product was removed from delivery.carrier. name and active fields

to merge before:

- [x] stock

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #1115, opened 2017-10-03
[ADD] Update coverage status of delivery module

StefanRijnhart:
Follow up of https://github.com/OCA/OpenUpgrade/pull/1103

--


pedrobaeza:
Why not doing directly in the PR as it's opened?

StefanRijnhart:
Good point, just confused.

PR #1119, opened 2017-10-03
Py3 compat, openupgrade records and travis data loading

StefanRijnhart:
* Implemented Python version switching in Travis to load yml data
* Finish conversion to Python3 of OpenUpgrade code
* Move from openerplib to odoorpc for openupgrade_records communication because of Python3 compatibility
* Switch to pip installed sphinx because sphinx3-python did not seem to use the active virtualenv
* Move deprecated sphinx-build command line option to new option in config file


pedrobaeza:
Travis is failing.  Can you check?

StefanRijnhart:
@pedrobaeza that is to be expected. It tries to migrate modules that are marked Done but there are dependencies that are not yet done (such as `base`). Please merge so that we can rebase other PRs.

pedrobaeza:
OK, merging then

PR #1121, opened 2017-10-04
11.0 base

StefanRijnhart:
Continuation of https://github.com/OCA/OpenUpgrade/pull/1100. Take care of migration of ir.values to action fields and add a test for this. Rebased onto https://github.com/OCA/OpenUpgrade/pull/1119. Merge that one first. Tests for this one then should be green before merge (or maybe not, given other migrated modules with missing dependencies).

--


MiquelRForgeFlow (approved):
LGTM

odoo/addons/base/migrations/11.0.1.3/post-migration.py
```diff
@@ -0,0 +1,104 @@
...
+from odoo.tools import pickle
+
+from openupgradelib import openupgrade
+
+from psycopg2.extensions import AsIs
+
+
+def map_ir_actions_server_fields(cr):
+    """Map model_id from ir_values to ir_actions' binding_model_id. As per
+    table inheritance, this will set the values in the related tables such
```
rruebner, line 14: @StefanRijnhart Is this true? I didn't know that by simple "UPDATE" related tables are also updated. The rest is fine from my side.

rruebner (approved):
Really cool. I didn't know about the table inheritance. So here is my 👍

odoo/addons/base/migrations/11.0.1.3/post-migration.py
hbrunn: this is not generally true for Odoo models inheriting other models/tables, but for those specific ones it is, because the init sql does the appropriate trickery: https://github.com/OCA/OCB/blob/11.0/odoo/addons/base/base.sql#L5

hbrunn (approved):
should we merge or are you planning to add more commits?

StefanRijnhart:
Merge please ;-)

PR #1123, opened 2017-10-06
[FIX] Migration of resource

StefanRijnhart:
Fixes tests and prevents an exception when the ORM retrieves a default value for the new column resource_calendar_id on res_company.

--


MiquelRForgeFlow (approved):
LGTM :+1:

mreficent:
I will fix this migration. I was wrong :S

PR #1124, opened 2017-10-06
[IMP] Don't deactivate sample project items from 9.0

StefanRijnhart:
It seems a bad idea to try and set this data to inactive even if it exists in the database. If the sample project and tasks were not removed or set inactive by the users, they might have been reused as actual projects and setting them inactive would be harmful.


pedrobaeza (approved):
It's difficult to balance what each user makes with the data, but yours is a good assumption.

PR #1142, opened 2017-10-22
[11.0][IMP+FIX] Update documentation about merged modules + portal_stock merging

pedrobaeza:
`portal_stock` is incorrectly merged to `portal` instead of `stock`

@Tecnativa


pedrobaeza:
@StefanRijnhart you changed `portal_stock` merging operation to `portal` instead of `stock`. Can you explain me why? Is it only a question of already migrated modules?

I also have doubts about `website_portal` being merged in `website` instead of `portal`.

StefanRijnhart:
@pedrobaeza the renames are partially based on mismatches that occur in the analysis when other choices are made. See for a concrete example https://github.com/OCA/OpenUpgrade/pull/1102#issuecomment-333629253.

I was thinking that merging portal_stock with portal is harmless because there were no shared entries with either stock or portal. Was I incorrect?

MiquelRForgeFlow:
I think this can be closed, right?

pedrobaeza:
No, as there are modules still not documented, but I have to update this with latest checks.

pedrobaeza:
This is finally not needed as documentation has been updated meanwhile and the other renaming is not critical.

PR #1144, opened 2017-10-23
[MIG] base_vat: Nothing to do

cubells:
Nothing to do

cc @Tecnativa


addons/base_vat/migrations/11.0.1.0/openupgrade_analysis.txt
```diff
@@ -3,3 +3,4 @@
 NEW ir.ui.view: base_vat.res_config_settings_view_form
 NEW ir.ui.view: base_vat.view_partner_short_form
 DEL ir.ui.view: base_vat.company_form_vat
+# Nothing to do
```
pedrobaeza: You should create a new `openupgrade_analysis_work.txt` file, not modifying this one.

pedrobaeza (changes requested):

MiquelRForgeFlow (approved):

StefanRijnhart (approved):

StefanRijnhart:
@cubells I see you addressed @pedrobaeza's comment. That's great ;-) You should also put down a comment about it so that this PR shows up in Pedro's notifications. That way, Pedro has a way of knowing that this PR can be reassessed.

StefanRijnhart:
I'll merge as it is easy for me to validate on Pedro's behalve in this trivial case.

pedrobaeza:
Thanks

PR #1146, opened 2017-10-23
[MIG] bus: Nothing to do

cubells:
Nothing to do

cc @Tecnativa


pedrobaeza (approved):
Well, there's no analysis work file, although the original analysis file doesn't contain anything, so no matter at all.

PR #1149, opened 2017-10-23
[FIX][9.0] hr_expense: merge expense lines in expense

cubells:
- Rename he_expense_line to hr_expense, this change allows other modules that depends on hr_expense not to need migration scripts
- Update post-migration script, as hr_expense_line is renamed, update data is not necessary

cc @Tecnativa

### DIFF
--- a/addons/hr_expense/migrations/9.0.2.0/post-migration.py
+++ b/addons/hr_expense/migrations/9.0.2.0/post-migration.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 # © 2016 Eficent Business and IT Consulting Services S.L.
 # © 2016 Serpent Consulting Services Pvt. Ltd.
+# Copyright 2017 Tecnativa - Vicent Cubells
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-import logging
+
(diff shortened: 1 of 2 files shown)


addons/hr_expense/migrations/9.0.2.0/post-migration.py
```diff
@@ -13,57 +13,26 @@ def map_expense_state(env):
...
+            employee_id = ot.employee_id,
+            state = ot.state,
+            name = ot.name,
+            date = ot.date,
+            account_move_id = ot.account_move_id, 
+            untaxed_amount = 0.00,
+            payment_mode = 'own_account'
+        FROM %(old_table)s ot
+        WHERE ot.id = he.expense_id
+    """ % {'old_table': openupgrade.get_legacy_name('hr_expense_expense')})
```
StefanRijnhart: If you import AsIs from psycopg2.extensions you should be able to use Psycopg2 substitution

```
""", {'old_table': AsIs(openupgrade.get_legacy_name('hr_expense_expense'))})
```

MiquelRForgeFlow:
@cubells, please, check conflicts

pedrobaeza (approved):

PR #1156, opened 2017-10-23
[MIG] survey: Migration scripts

cubells:
- Preserve datetime columns data

cc @Tecnativa


addons/survey/migrations/11.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,26 @@
+---Fields in module 'survey'---
+survey       / survey.question          / type (selection)              : selection_keys is now '['date', 'free_text', 'matrix', 'multiple_choice', 'numerical_box', 'simple_choice', 'textbox']' ('['datetime', 'free_text', 'matrix', 'multiple_choice', 'numerical_box', 'simple_choice', 'textbox']')
+# Nothing to do: selections keys are the same
```
pedrobaeza: They are not the same: before it was `datetime`. Now it's `date`. You have to map them. Use the openupgradelib method `map_values`

addons/survey/migrations/11.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,26 @@
...
+
+survey       / survey.question          / validation_max_date (datetime): type is now 'date' ('datetime')
+DONE: pre-migration: Renamed to preserve data
```
pedrobaeza: Include a comment saying that the field is then converted automatically to date.

pedrobaeza (changes requested):

pedrobaeza:
@cubells please attend my comments.

pedrobaeza (approved):

PR #1157, opened 2017-10-27
[ADD] paydays: renamed to account_payment_term_multi_day

cubells:
Paydays module features are now in account_payment_term_multi_day module

cc @Tecnativa


StefanRijnhart (approved):
As per https://github.com/OCA/account-payment/pull/10

PR #1159, opened 2017-11-05
[9.0][IMP] *: Use new rename_fields method

pedrobaeza:
Depends on 

- [x] OCA/openupgradelib#84

This is a huge change, but it's only the result of applying 2 techniques:

* Use `rename_fields` with the existing renamed columns that are actually fields.
* Use `rename_fields` with the fields that have been handled by Odoo with `oldname` attribute (thanks @StefanRijnhart for including that lines in the analysis!)

With this, we improve even more our migration, handling:

* Fields that have been renamed and are translatable. The translations were missing before, so it can be considered this even a bug.
* If the fields are included in an export profile (first level for now), they are automatically handled.
* If the fields are included in a filter as domain (first level for now) or a group by, they are automatically handled.

@Tecnativa

### DIFF
--- a/addons/sale/migrations/9.0.1.0/pre-migration.py
+++ b/addons/sale/migrations/9.0.1.0/pre-migration.py
@@ -1,29 +1,32 @@
 # -*- coding: utf-8 -*-
-# © 2015 Microcom
-# © 2016 Eficent Business and IT Consulting Services S.L. -
+# Copyright 2015 Microcom
+# Copyright 2016 Eficent Business and IT Consulting Services S.L. -
 # Jordi Ballester Alomar
-# © 2016 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
-# © 2016 Opener B.V. - Stefan Rijnhart
+# Copyright 2016 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
+# Copyright 2016 Opener B.V. - Stefan Rijnhart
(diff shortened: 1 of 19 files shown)


StefanRijnhart (approved):

pedrobaeza:
We need to remove the commit for changing openupgradelib alternative source before merging.

addons/sale/migrations/9.0.1.0/pre-migration.py
```diff
@@ -1,30 +1,33 @@
...
     ],
-    'sale_order_tax': [
-        ('order_line_id', 'sale_order_line_id'),
-        ('tax_id', 'account_tax_id'),
-    ],
 }
 
+field_renames = [
+    ('sale.order', 'sale_order', 'section_id', 'team_id'),
+    ('account.invoice', 'accouwnt_invoice', 'section_id', 'team_id'),
```
MiquelRForgeFlow: account_invoice

MiquelRForgeFlow (changes requested):

addons/sale/migrations/9.0.1.0/pre-migration.py
pedrobaeza: Good catch! Changed

pedrobaeza:
@mreficent please merge when CIs finish

MiquelRForgeFlow (approved):

PR #1160, opened 2017-11-05
[10.0][IMP] *: Use new rename_fields method

pedrobaeza:
Depends on 

- [x] OCA/openupgradelib#84

This is a huge change, but it's only the result of applying 2 techniques:

* Use `rename_fields` with the existing renamed columns that are actually fields.
* Use `rename_fields` with the fields that have been handled by Odoo with `oldname` attribute.

With this, we improve even more our migration, handling:

* Fields that have been renamed and are translatable. The translations were missing before, so it can be considered this even a bug.
* If the fields are included in an export profile (first level for now), they are automatically handled.
* If the fields are included in a filter as domain (first level for now) or a group by, they are automatically handled.

@Tecnativa

### DIFF
--- a/addons/account_asset/migrations/10.0.1.0/pre-migration.py
+++ b/addons/account_asset/migrations/10.0.1.0/pre-migration.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 Tecnativa - Vicent Cubells
+# Copyright 2017 Tecnativa - Pedro M. Baeza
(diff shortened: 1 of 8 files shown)


StefanRijnhart (approved):

MiquelRForgeFlow (approved):

pedrobaeza:
@mreficent @StefanRijnhart you can merge this one as well when CIs finish now that the dependency has been merged.

PR #1161, opened 2017-11-07
[FIX] hr_attendance: Assure order on check-out assign

pedrobaeza:
This way, the retrieved check-out is the nearest one.


rafaelbn (approved):
Tested in production enviroment

MiquelRForgeFlow (approved):

PR #1162, opened 2017-11-07
[FIX] mass_mailing: Link `mail.mass_mailing.campaign` to `utm.campaign` records

yajo:
Method docstring explains the change.

--
@Tecnativa

### DIFF
--- a/addons/mass_mailing/migrations/10.0.2.0/pre-migration.py
+++ b/addons/mass_mailing/migrations/10.0.2.0/pre-migration.py
@@ -32,32 +32,7 @@ def fill_mass_mailing_sources(env):
         )
 
 
-def convert_campaigns_to_utm(env):
-    """Link ``mail.mass_mailing.campaign`` to ``utm.campaign`` records.
-
-    This method creates one ``utm.campaign`` record for each
-    ``mail.mass_mailing.campaign`` that has none, and fills its name if it had
-    it empty.
-    """
-    mass = env["mail.mass_mailing.campaign"].search([
-        "|", ("campaign_id", "=", False), ("campaign_id.name", "=", False),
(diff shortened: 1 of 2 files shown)


rafaelbn (approved):
Tested functionally in production

pedrobaeza (changes requested):
Do it on post-migration, as models are still not registered and you are using ORM.

addons/mass_mailing/migrations/10.0.2.0/pre-migration.py
```diff
@@ -32,7 +32,32 @@ def fill_mass_mailing_sources(env):
...
+    """
+    mass = env["mail.mass_mailing.campaign"].search([
+        "|", ("campaign_id", "=", False), ("campaign_id.name", "=", False),
+    ])
+    for one in mass:
+        env.cr.execute(
+            "SELECT name FROM mail_mass_mailing_campaign WHERE id = %s",
+            (one.id,)
+        )
+        name = env.cr.fetchone()[0]
```
pedrobaeza: Please protect this from non results:
```
row = env.cr.fetchone()
if not row:
    continue
...
```

StefanRijnhart: @pedrobaeza given the previous lines it seems that each id is guaranteed to exist in the database.

yajo: Indeed, the failure is not possible here.

yajo:
Thanks for the tips, let's see now

StefanRijnhart (approved):

pedrobaeza (approved):
Please squash before merging

PR #1168, opened 2017-11-14
[IMP] No need to install "contract" without recurring invoicing

StefanRijnhart:
The apriori mapping migrates account_analytic_analysis to contract, but if the module was only used for the analytic report that it included and not for recurring invoicing, the installation of the contract module is not needed.


StefanRijnhart:
Also fixes an unrelated pep8 error.

pedrobaeza (approved):

PR #1170, opened 2017-11-15
[UPD] documentation: renames, merges & bigger table

MiquelRForgeFlow:
- Updated documentation with renames and merges.
- Provided bigger table.


pedrobaeza:
Can't we avoid to make bigger and bigger tables with other syntax that doesn't need this fixed size? I think RST have a syntax for that

MiquelRForgeFlow:
I don't know, but I don't think we will need to make the table bigger anymore (in this version). Let's begin to think that for v12 :stuck_out_tongue_winking_eye:

mreficent:
Let's begin to think that for v12 :stuck_out_tongue_winking_eye:

MiquelRForgeFlow:
I don't know what happened with Travis, but can we fast-track it? :S

StefanRijnhart:
Can you not tag obsolete modules with 'Done' as it triggers travis to try and upgrade these modules. Not sure if that broke travis now but it's not pretty.

pedrobaeza:
@StefanRijnhart we tag them as "Done" where there's code for handling this disappearance. For example, calling `update_module_names` with `merge_modules=True` argument. So for that part, I think it's correct, but you are right that it's going to try to install that module on Travis, but the only output we will get is a warning about not found module on this version. On the other way, I think it's needed for having the module installed on the previous version.

Any way, the error should come from other place.

StefanRijnhart:
So don't tag those modules with Done, or modify travis config even if it does not break travis __now__. Just tagging them with `Merged with...` is just as clear.

pedrobaeza:
But are you sure they don't need to be installed on version N-1?

StefanRijnhart:
Travis runs the tests on a prefab database with all modules installed, so the migration status of a module does not influence the set of installed modules, if that is what you mean.

StefanRijnhart:
But you are onto something. When a module is renamed, its ir_module_module entry is renamed as well. Travis will set its status to uninstalled and will not run the new module's migration scripts. So we need to split up the selection of modules to uninstall and the selection of modules to upgrade.

For the scope of this PR then we can leave it like this for now if it does not break anything. Travis is already broken so we can merge this and clean up afterwards.

StefanRijnhart (approved):

MiquelRForgeFlow:
Then I merge this as fast-tracking.

MiquelRForgeFlow:
Hmmm... The problem is still there :S

MiquelRForgeFlow:
The problem it's the table's size or what? I don't understand why changing a `.rst` file is giving problems.

StefanRijnhart:
What is the problem?

MiquelRForgeFlow:
I don't know, but Travis now is red in the new PRs

PR #1171, opened 2017-11-15
[FIX] event: Take into account nb_register

pedrobaeza:
On v8, we have one event.registration record = one or several attendees according nb_register field.

But on v9 we have one event.registration record = one attended, so here we duplicate the existing record for representing the same number of records than attendees.

Done in end-migration for having all the possible fields in event.registration already loaded.

@Tecnativa

### DIFF
--- a/addons/event/migrations/9.0.0.1/end-migration.py
+++ b/addons/event/migrations/9.0.0.1/end-migration.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def create_multiple_attendees_registration_records(env):
+    """On v8, we have one event.registration record = one or several attendees
+    according nb_register field.
+
+    But on v9 we have one event.registration record = one attended, so here
+    we duplicate the existing record for representing the same number of
+    records than attendees.
+
+    Done in end-migration for having all the possible fields in
+    event.registration already loaded.
+    """
+    # Pre-create a column for storing original reference
+    column_name = openupgrade.get_legacy_name('original_registration_id')
+    table = 'event_registration'
+    if not openupgrade.column_exists(env.cr, table, column_name):
+        openupgrade.logged_query(
+            env.cr, """
+            ALTER TABLE %s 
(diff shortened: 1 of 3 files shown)


addons/event/migrations/9.0.0.1/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from . import test_event
```
StefanRijnhart: As travis notes, you do not need to import migration tests.

addons/event/migrations/9.0.0.1/end-migration.py
```diff
@@ -0,0 +1,47 @@
...
+
+    But on v9 we have one event.registration record = one attended, so here
+    we duplicate the existing record for representing the same number of
+    records than attendees.
+
+    Done in end-migration for having all the possible fields in
+    event.registration already loaded.
+    """
+    # Pre-create a column for storing original reference
+    column_name = openupgrade.get_legacy_name('original_registration_id')
```
StefanRijnhart, line 20: Can you prefix this column with 'openupgrade_' so that it is clear where it came from?

pedrobaeza: It's already done calling `get_legacy_name`...

addons/event/migrations/9.0.0.1/tests/__init__.py
pedrobaeza: No, that's not the problem I think. I'll tell you later.

addons/event/migrations/9.0.0.1/end-migration.py
StefanRijnhart: True!

addons/event/migrations/9.0.0.1/tests/__init__.py
StefanRijnhart: It is __a__ problem though.
```
$ flake8 addons/*/migrations/*/tests/ --max-line-length=120
addons/event/migrations/9.0.0.1/tests/__init__.py:4:1: F401 '.test_event' imported but unused
The command "flake8 addons/*/migrations/*/tests/ --max-line-length=120" exited with 1.
```

pedrobaeza:
@StefanRijnhart I have changed tests for being deferred

StefanRijnhart (approved):
Thanks!

PR #1172, opened 2017-11-16
[RFC] Fix and simplify tax computation.

StefanRijnhart:
Make sure that every line with a tax code gets a tax assigned. Between taxes
with the same tax code it is less important which one is assigned, because
presumably taxes with the same tax code will be mapped to the same tax tag in
9.0 so that the tax report remains unchanged.


pedrobaeza:
Sorry, but all the code I did is needed with Spanish localization taxes. Tags are not the solution.

StefanRijnhart:
@pedrobaeza @jbeficent @NL66278 can you give your insights on this?

The current migration from tax codes on move lines to taxes goes through a lot of effort to find a unique mapping from tax codes to taxes. In my experience, a lot of taxes share the same tax codes. For instance, the Dutch tax chart will have the same tax codes on the 21% excluded tax, the 21% included tax and the obsolete 19% taxes from before the VAT raise. The current code fails to migrate move lines with any of these taxes because there is no perfect match.

And IMHO the goal of the migration should be that the VAT report looks the same before and after the migration. The VAT report in Odoo 8.0 is based on tax codes, while the VAT report in Odoo 9.0 is based on tax tags (not on taxes). Tags are generated from the obsolete tax codes so move lines with a specific tax code will be associated with the correct tax tag no matter what actual tax is assigned to the move line.

So I'd like to know if this change is acceptable for you.

What is obviously still missing is the mapping of lines without a debit and credit amount. This situation does not occur in my localization so I need some guidance here. This part of the migration seems to suffer from the same problem: when no perfect match is found, no action is taken. Instead of trying to find a line with the exact amount, I would like to suggest to modify the given move line and assign the tax amount as debit/credit, and add a counterpart line on the same account so that the move balance does not change. This is ugly, but it will retain the integrity of the tax report.

A particular question about this last aspect of the current migration is that the tax amount is taken to be positive and is then mapped to either a debit or a credit line: https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/post-migration.py#L359. Intuitively, I would think that the sign of the amount matters so that the tax needs to be assigned to a move line with the correct sign. My suggestion above would fix this.

StefanRijnhart:
@pedrobaeza please read the comment above. I need to fix the migration for other localizations.

pedrobaeza:
For Spain is not acceptable of matching the first found. A possible agreement can be to search for an unique match and if after all, it's not found, then take the first one.

About debit/credit, let me check with more time.

StefanRijnhart:
If we add my code just as a fallback I have to sit and wait until every move line with a non-unique tax code mapping is checked against the translation table. Can we add both variants as a 'strategy' and then apply the strategy depending on the localizations? The code would have to be made company aware but that should not be hard. We could add a list of localizations that need perfect tax matching.

pedrobaeza:
OK, if you prefer... but this code is very fast any way. Have you tried? I have migrated DBs with millions of records without too much lag.

StefanRijnhart:
@pedrobaeza maybe so! Or maybe that is because in your case, the unique tax code is usually found and the translation fallback is not triggered. On my databases, the translation fallback is triggered for every move line with a tax code.

pedrobaeza:
OK, let's go with your strategy, but please activate for Spain the "detailed" one.

StefanRijnhart:
For the record, it's 58 seconds to run the current migration on a database with only 33.000 tax lines. My method takes 3.3 seconds so I think I go for implementing different strategies.

pedrobaeza:
@StefanRijnhart are you going to continue working on this?

sbidoul:
@StefanRijnhart thanks a lot for pointing me here. 

[I missed the big warning sign somewhere that says the tax migration is not reliable in all cases :wink:]

The approach of this PR is interesting. It looks like it's indeed more robust for the purpose of preserving the correctness of the VAT declaration reports. I'm not very comfortable with the idea of taking the first tax that matches the tax code, though: the result could be very confusing for accountants looking at move lines.

I'm playing (in my head so far) with an alternative approach.

* per localization, declare a map {tax code: tax tag} (such as the csv in #1432)
* for each tax code used, create a new inactive tax named `Code {code}` with the corresponding tag 
* for tax lines, find the corresponding Code tax and use it as tax_line_id
* for base lines, find the corresponding Code tax and add it to tax_ids
* ~~for tax lines with debit=credit=0, find/create the corresponding Code tax and add it to the tax_ids of the base line where debit or credit = tax_amount~~ (*is that correct? => no it's not*)

We could even do this only when several taxes share the same code.

This approach should be quite robust with little surprises and work in all situations I can think of.
Of course it does not preclude keeping specific strategies for situations where something smarter can be done.

What do you think?

sbidoul:
@StefanRijnhart @pedrobaeza I implemented the above strategy in #1432. The only limitation I found so far is that it is sometimes impossible to disambiguate between base codes and tax codes (in rare situations where the same code is used for base and tax).

sbidoul:
@StefanRijnhart at the end of the day I implemented your suggestion to handle lines without a debit and credit amount

https://github.com/acsone/OpenUpgrade/blob/236cebbd84721c0cb1c5ab49e72ba1fbaebc7787/addons/l10n_be/migrations/9.0.1.1/post-migration.py#L200-L212

pedrobaeza:
> The only limitation I found so far is that it is sometimes impossible to disambiguate between base codes and tax codes (in rare situations where the same code is used for base and tax).

That's the problem I found in Spanish localization, where each tax code is shared by a lot of taxes. We have to pay the sins of the past...

> at the end of the day I implemented your suggestion to handle lines without a debit and credit amount

I also implemented something similar in https://github.com/OCA/OpenUpgrade/blob/35e5aee1ac8ca2c6c47a4fef492a387d611bab1c/addons/account/migrations/9.0.1.1/post-migration.py#L421-L452

StefanRijnhart:
Indeed, that was part of @pedrobaeza's implementation, not mine.

pedrobaeza:
Should we close this?

pedrobaeza:
No answer, so closing.

PR #1175, opened 2017-11-16
[11.0][MIG] calendar

yuriqp:
migration of calendar module

Current behavior before PR:

--


addons/calendar/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,22 @@
+
+---Fields in module 'calendar'---
+calendar     / calendar.contacts        / user_id (many2one)            : now required, default = function
+# Nothing to do, now required but filled by default function
```
StefanRijnhart, line 4: The default is user specific. Can you set user_id to create_uid where user_id is null in the pre-migration?

addons/calendar/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,267 @@
...
+    <field name="model"/>
+    <field name="name">Calendar: Event Reminder</field>
+    <field name="function"/>
+    <field name="code">model.get_next_mail()</field>
+    <field name="model_id" ref="model_calendar_alarm_manager"/>
+    <field name="state">code</field>
+  </record>
+
+  <record id="calendar_template_meeting_reminder" model="mail.template">
+    <field name="body_html"><![CDATA[
```
StefanRijnhart, line 14: When we reset noupdate data during a migration, we need to consider if it would be better if the customer's customizations are retained. This is especially true for report and email templates. If you do reset the data, it would be a good idea to unlink the related translations, otherwise the translated template will still be the old version.

Personally, I think in the case of the calendar reminder templates it would be good to take the new version.

MiquelRForgeFlow:
@yuriqp, please, if you do more changes, do a `git pull` before doing them, because I resolved the conflicts.

StefanRijnhart (approved):
Thanks for the updates!

addons/calendar/migrations/11.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Now user_id is required, set it to the one who created the entry
+    """
+    cr.execute("""
+        UPDATE calendar_contacts SET user_id = create_uid
+        WHERE user_id IS NULL;
+    """)
+
+
+def cleanup_translations(cr):
```
pedrobaeza, line 18: I insist this should be an openupgradelib because it's normal to use this same code across several migration scripts.

StefanRijnhart: Not sure what you mean by 'insist' but I think it's a good suggestion!

pedrobaeza: hehe, it was for @yuriqp because he has done it in another PR and I have put the same comment.

yuriqp: As I mentioned in other PR (https://github.com/OCA/OpenUpgrade/pull/1188) @rruebner already created a PR with this function in openupgradelib (OCA/openupgradelib#87). What should I do, wait until it is merged and then create a new PR to use it in this migration?

pedrobaeza: When merged, we make another PR calling that function instead.

pedrobaeza:
@yuriqp please make a PR using the recently merged `delete_record_translations` method in openupgradelib.

yuriqp:
I created a new PR using the new `delete_record_translations` method (https://github.com/OCA/OpenUpgrade/pull/1247)

PR #1178, opened 2017-11-16
[11.0][MIG] hr

nadiaafa:

Current behavior before PR:

--


nadiaafa:
Hi,
I've signed the CLA, but still unknown to clabot, could you check please?

addons/hr/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,17 @@
...
+# Nothing to do
+
+---XML records in module 'hr'---
+NEW ir.actions.act_window: hr.hr_config_settings_action
+NEW ir.ui.menu: hr.hr_menu_configuration
+NEW ir.ui.view: hr.hr_job_view_kanban
+NEW ir.ui.view: hr.res_config_settings_view_form
+DEL ir.values: hr.hr_employee_action_subordinate_hierachy_mutli
+DEL mail.message: hr.module_install_notification (noupdate)
+DEL res.groups: base.group_user (noupdate)
```
MiquelRForgeFlow: noupdate records. You should create the `noupdate_changes.xml` file the same way as here: https://github.com/OCA/OpenUpgrade/pull/1175

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow:
Don't worry, the CLA has been processed just now. The clabot will be updated in any moment.

addons/hr/migrations/11.0.1.1/openupgrade_analysis_work.txt
nadiaafa: when i run 
python compare_noupdate_xml_records.py ../../OpenUpgrade/addons/hr ../addons/hr

i got this:

```
<odoo/>
```

StefanRijnhart: True. What you should do is remove the mail message in the post script. Because it is noupdate, it won't be removed automatically. Some records are better left in the database, but this message was never good for anything.

The res.groups is a false positive, of course. In this case it is the override that was removed from the module, not the group itself.

odoo/openupgrade/doc/source/modules100-110.rst
```diff
@@ -137,7 +137,7 @@ missing in the new release are marked with |del|.
 +-----------------------------------+-------------------------------------------------+
 |google_spreadsheet                 |                                                 |
 +-----------------------------------+-------------------------------------------------+
-|hr                                 |                                                 |
+|hr                                 | Done                                            |
```
pedrobaeza: This should be `Nothing to do`

pedrobaeza (changes requested):

odoo/openupgrade/doc/source/modules100-110.rst
StefanRijnhart: Not if my change request is followed up ;-)

pedrobaeza: I though it was a false positive. About `hr.module_install_notification`, it's harmless to let it, because it's not going to be sent anymore.

StefanRijnhart: True!

addons/hr/migrations/11.0.1.1/openupgrade_analysis_work.txt
StefanRijnhart: Then again, as Pedro mentions below, the notification was already fired if this module is being upgraded.... So I agree with him, just make this a 'Nothing to do'

nadiaafa: Thank you @StefanRijnhart and @pedrobaeza , i did

StefanRijnhart:
Ugh. Already merged with https://github.com/OCA/OpenUpgrade/pull/1193

StefanRijnhart:
I used the web editor to resolve the conflict.

StefanRijnhart:
Squashed and merged. Thanks @nadiaafa!

nadiaafa:
Thank you

PR #1180, opened 2017-11-16
[11.0][MIG] hr_contract

fayezqandeel:

Current behavior before PR:

--


addons/hr_contract/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,28 @@
...
+hr_contract  / hr.contract              / state (selection)             : selection_keys is now '['cancel', 'close', 'draft', 'open', 'pending']' ('['close', 'draft', 'open', 'pending']')
+# Nothing to do, nothing related
+
+hr_contract  / hr.contract              / trial_date_start (date)       : DEL
+# Nothing to do, nothing related
+
+hr_contract  / hr.contract              / website_message_ids (one2many): NEW relation: mail.message
+# Nothing to do, new relations
+
+hr_contract  / hr.contract              / working_hours (many2one)      : DEL relation: resource.calendar
```
pedrobaeza: This is clearly a rename of the field from `working_hours` to `resource_calendar_id`, and more if you check the label from both fields in v10 and v11: `Working Schedule`. Please pay attention to these details.

pedrobaeza (changes requested):

fayezqandeel:
@pedrobaeza rename field done on pre_migration.py

pedrobaeza:
Completed the migration with:

* Fill company in contract from job.
* Remove deleted noupdate records.

pedrobaeza (approved):

pedrobaeza:
The unlink of the filter has revealed a possible problem with renamed_models method in OpenUpgrade. I'll investigate this further.

pedrobaeza:
This depends indirectly of `base_automation`. We will probably need to add the dependency in the manifest for avoiding problems in the loading order (as we are unlinking base.automation records that are not anymore in this version, nor the dependency).

pedrobaeza:
I just think about other better solution. Do the unlinking on end-migration where all the models are already loaded.

pedrobaeza:
Correct after merging base_automation, so I merge.

PR #1181, opened 2017-11-16
[11.0][MIG] Migration Script for l10n_de, l10n_de_skr03 and l10n_de_skr04

christophlsa:
In general here was nothing todo except the noupdate changes but I have one question/suggestion:

The l10n_de module removed some "account.account.tag" records. These entries have no `noupdate=1` which means that the entries will be removed. Do you think this is good? This means if these tags are used in account.account that these will be removed there during the update. Do you think I should write a pre migration script to set these entries to `noupdate = true`?

--


addons/l10n_de/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,146 @@
...
+DEL account.account.tag: l10n_de.tag_de_50
+DEL account.account.tag: l10n_de.tag_de_51
+DEL account.account.tag: l10n_de.tag_de_54
+DEL account.account.tag: l10n_de.tag_de_55
+DEL account.account.tag: l10n_de.tag_de_56
+DEL account.account.tag: l10n_de.tag_de_57
+DEL account.account.tag: l10n_de.tag_de_58
+DEL account.account.tag: l10n_de.tag_de_70
+DEL account.account.tag: l10n_de.tag_de_71
+DEL account.account.type: l10n_de.account_type_other (noupdate)
```
pedrobaeza: It seems that this old type has been replaced by `account.data_account_type_other_income` (I have checked several accounts, but not all). I think you should search all occurrences of your account.account with this `user_type_id` and replace them with the other type, and finally remove the account type.

pedrobaeza (changes requested):

addons/l10n_de/migrations/11.0.1.0/openupgrade_analysis_work.txt
christophlsa: I'm not 100% sure here. I also went through the account types and this one is the only one I can imagine to be right here. I found the PR https://github.com/odoo/odoo/pull/17447 and the discussion in https://github.com/odoo/odoo/issues/18571 but couldn't find out why it was removed.

Nevertheless I wrote a script to replace the type with the other one.

christophlsa:
I did the Review Changes but the travis build timed out. I'm not sure if it is possible to rerun the build.

pedrobaeza (approved):

pedrobaeza:
I have rebased for having account module migrated and prevent a problem if the account type is used elsewhere in other model. Merging this.

PR #1184, opened 2017-11-16
[11.0][FIX] Documentation

MiquelRForgeFlow:
@StefanRijnhart you said:
> Can you not tag obsolete modules with 'Done' as it triggers travis to try and upgrade these modules. Not sure if that broke travis now but it's not pretty.

> So don't tag those modules with Done, or modify travis config even if it does not break travis now. Just tagging them with Merged with... is just as clear.

So, I will delete "done"s until the Travis gets green.


StefanRijnhart:
Thanks for the fix!

PR #1188, opened 2017-11-17
[11.0][MIG] Migration project

yuriqp:
Migration of project module

Current behavior before PR:

--


addons/project/migrations/11.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,37 @@
...
+        SELECT res_id
+        FROM ir_model_data
+        WHERE
+            module = 'project' AND model = 'mail.template' AND name in %s
+    """, (updated_templates,))
+    record_ids = tuple([r[0] for r in cr.fetchall()])
+
+    query = ("""
+        DELETE FROM ir_translation
+        WHERE module = 'calendar' AND res_id IN %s
```
pedrobaeza: calendar?

pedrobaeza: Isn't better to create an openupgradelib function for handling this?

pedrobaeza (dismissed):

addons/project/migrations/11.0.1.1/pre-migration.py
yuriqp: I saw that @rruebner already created a PR with this function in openupgradelib (https://github.com/OCA/openupgradelib/pull/87). What should I do, wait until it is merged?

pedrobaeza: Yes, let's merge it and then call it from here.

pedrobaeza:
Still waiting to fix and merge PR in openupgradelib for translation cleanup.

pedrobaeza:
@yuriqp please use the recently merged `delete_record_translations` method in openupgradelib.

pedrobaeza:
Travis is still red

yuriqp:
@pedrobaeza Travis is now green :), can this go further?

pedrobaeza:
I'm not seeing the conversion of all project issues into project tasks. My strategy would be:

* Look for projects with issues.
* Duplicate that projects, with a suffix like " - Issues".
* Create one task per issue associated to the duplicated project.

hbrunn:
...and only create tasks for issues that don't already have a task I'd say

yuriqp:
but wouldn't we be losing some information if we get rid of issues related with tasks? For example, let say I have an issue related with a task, and added some information in the description field of the issue.

yuriqp:
Travis is failing, but looks like it is not related with the changes in the branch.

yuriqp:
Any news on why is travis failing?

StefanRijnhart:
@pedrobaeza thanks for picking this one up. Will you migrate issue messages and attachments to the new tasks as well?

pedrobaeza:
Good catch! I haven't thought about that, but yes, I can do it. For now, I was checking that everything was more or less in line. Adding both things in next push.

pedrobaeza:
@StefanRijnhart This is ready. Maybe we can inject some values for testing followers, but I'm a bit exhaust with this module.

StefanRijnhart (approved):
Thanks!

PR #1189, opened 2017-11-17
[11.0][MIG] mass mailing

Benniphx:
Migration done


Benniphx:
I squashed with my main account cla problem should not be a problem

addons/mass_mailing/migrations/11.0.2.0/post-migration.py
```diff
@@ -0,0 +1,40 @@
...
+            INSERT INTO mail_mass_mailing_contact_list_rel (contact_id,
+            list_id) values(%s,%s)
+            """, (r[0], r[1],))
+
+
+def map_mailing_model_to_mailing_model_id(cr):
+    """Set the mailing_model_id for the mailing_model in mail.mass_mailing"""
+    cr.execute(
+        """
+        SELECT mm.id, m.id FROM mail_mass_mailing AS mm
```
StefanRijnhart: Try this in a single query using something like `UPDATE mail_mass_mailing mm SET mailing_model_id = im.id FROM ir_model WHERE ir.model = mm.mailing_model` (untested)

StefanRijnhart: In this case, performance can really take a hit if you don't refactor this into a single query. Or you can use the API https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.m2o_to_x2m

StefanRijnhart:
Please take the noupdate changes in the ir.cron record.

StefanRijnhart (changes requested):
See my individual comments

addons/mass_mailing/migrations/11.0.2.0/post-migration.py
pedrobaeza: First query is correct. The openupgradelib method can't be used as this was previously a Char field.

addons/mass_mailing/migrations/11.0.2.0/post-migration.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# © 2017 bloopark systems (<http://bloopark.de>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def map_list_id_to_list_ids(cr):
+    """Set the list_ids for the list_id in mail.mass_mailing.contact"""
+    cr.execute(
```
pedrobaeza: This is the case where you should use openupgradelib `m2o_to_x2m` (https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.m2o_to_x2m)

addons/mass_mailing/migrations/11.0.2.0/post-migration.py
StefanRijnhart: @pedrobaeza thanks for the correction. Hit the wrong line with that second comment.

pedrobaeza:
I have changed the script for using m2o_to_x2m, and use only one query for the update, so @StefanRijnhart's comment has been honored, so this can be merged.

pedrobaeza (approved):

PR #1195, opened 2017-11-17
[11.0][MIG] account

rruebner:

account migration

--


rruebner:
Can someone help. Travis shows an error with the copy_columns() function in the pre-migration.py but I don't know what is wrong. I use the same format like I did it in other migration scripts.

addons/account/migrations/11.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,46 @@
...
+        ('state', None, None),
+    ],
+}
+
+
+def add_required_colum_chart_template_id(cr):
+    """The model account.reconcile.model.template has now a new required
+    many2one field chart_template_id with no default value. To avoid errors
+    on module update we add already the column here and fill it with the
+    right chart template ID."""
```
StefanRijnhart: It should not be a fatal error if a uniqueness constraint cannot be set on a column during a module upgrade. I believe the chart template id is set by each of the localization's data files, so you can leave out this whole bit if I am not mistaken.

StefanRijnhart:
copy_columns breaks on the space in `'account_invoice '`

addons/account/migrations/11.0.1.1/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+        openupgrade.get_legacy_name('state'), 'state',
+        [('proforma', 'draft'), ('proforma2', 'draft')],
+        table='account_invoice', write='sql')
+
+    # copy statement_line_id values from account.move to account.move.line
+    env.cr.execute("""
+        UPDATE account_move_line aml
+        SET statement_line_id = am.statement_line_id
+        FROM account_move am
+        WHERE aml.move_id = am.id;
```
StefanRijnhart: It's quicker if you only perform this operations on moves where statement_line_id is set.

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,195 @@
...
+account      / account.group            / code_prefix (char)            : NEW 
+account      / account.group            / name (char)                   : NEW required: required
+account      / account.group            / parent_id (many2one)          : NEW relation: account.group
+account      / account.group            / parent_left (integer)         : NEW 
+account      / account.group            / parent_right (integer)        : NEW 
+account      / account.invoice          / access_token (char)           : NEW 
+account      / account.invoice          / activity_ids (one2many)       : NEW relation: mail.activity
+account      / account.invoice          / cash_rounding_id (many2one)   : NEW relation: account.cash.rounding
+account      / account.invoice          / refund_invoice_ids (one2many) : NEW relation: account.invoice
+# NOTHING TO DO
```
StefanRijnhart: Can you note here that refund_invoice_ids is the One2many counterpart of an existing Many2one field?

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,195 @@
...
+account      / account.fiscal.position.template / vat_required (boolean)        : NEW 
+account      / account.fiscal.position.template / zip_from (integer)            : NEW 
+account      / account.fiscal.position.template / zip_to (integer)              : NEW 
+account      / account.full.reconcile   / exchange_partial_rec_id (many2one): DEL relation: account.partial.reconcile
+account      / account.group            / code_prefix (char)            : NEW 
+account      / account.group            / name (char)                   : NEW required: required
+account      / account.group            / parent_id (many2one)          : NEW relation: account.group
+account      / account.group            / parent_left (integer)         : NEW 
+account      / account.group            / parent_right (integer)        : NEW 
+account      / account.invoice          / access_token (char)           : NEW
```
StefanRijnhart: Can you note here that Odoo already pre-populates this column with unique tokens?

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,197 @@
...
+account      / account.payment          / message_ids (one2many)        : NEW relation: mail.message
+account      / account.payment          / message_last_post (datetime)  : NEW 
+account      / account.payment          / state (selection)             : selection_keys is now '['cancelled', 'draft', 'posted', 'reconciled', 'sent']' ('['draft', 'posted', 'reconciled', 'sent']')
+account      / account.payment          / website_message_ids (one2many): NEW relation: mail.message
+account      / account.payment          / writeoff_label (char)         : NEW 
+account      / account.payment.term     / sequence (integer)            : NEW required: required, req_default: function
+# NOTHING TO DO
+
+account      / account.reconcile.model.template / chart_template_id (many2one)  : NEW relation: account.chart.template, required: required
+# DONE: pre-migration: add column to table and add correct chart template IDs
```
MiquelRForgeFlow: then change this message

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,197 @@
+---Fields in module 'account'---
+account      / account.account          / group_id (many2one)           : NEW relation: account.group
+account      / account.account.tag      / active (boolean)              : NEW
```
pedrobaeza, line 3: @StefanRijnhart Doesn't the analysis now tell about fields with default function?

addons/account/migrations/11.0.1.1/pre-migration.py
rruebner: Interesting. I didn't know that it would work but I tested it and yes it works. You will get first a warning in your server log but later the chart_template_id field will be set by the XML record (l10n_fr has a chart_template_id for example).

rruebner:
@StefanRijnhart Fixes / updates done.

pedrobaeza (changes requested):
You should reset also translations of the templates. @yuriqp has created a good function in other PRs that can be put in openupgradelib for reusability.

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
StefanRijnhart: Only for required fields now. I've now created https://github.com/OCA/OpenUpgrade/issues/1204 to track this issue.

pedrobaeza: Thanks!

StefanRijnhart:
@pedrobaeza all in good faith, but I prefer you change my comments to ~~strikethrough~~ instead of deleting them. Now I don't remember what I wrote on this particular PR!

pedrobaeza:
You're right. It was for removing things that have been already done . They were comments about adding an explanation of one2many.

pedrobaeza:
@rruebner please use the recently merged `delete_record_translations` method in openupgradelib.

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
rruebner: Done.

rruebner:
Done.

rruebner:
@pedrobaeza Should I do something because of the travis error? Right now I don't see an issue based on my changes.

StefanRijnhart:
@rruebner please rebase to fix that pip error

njeudy:
Hello @rruebner you have to migrate Pro-forma Invoices groupe from account to sale module too (or should it be in sale module migration ? @StefanRijnhart )

_xmlid_renames = [
    ('account.group_proforma_invoices', 'sale.group_proforma_sales'),
]

njeudy:
@rruebner a way to add me in your personnal branch to help you ?

rruebner:
@njeudy I could do the renaming in account migration with a check if sale is installed but as you also said I would put it in sale migration. @pedrobaeza, @StefanRijnhart What do you think?

Pedro, is it enough if I add some account tests to make the coverage green? I ask because also other modules miss some coverage.

pedrobaeza:
No need to check if `sale` module is installed, as you can perform the same the renaming, and if the module is not installed on migration, when installed on new version, you will have already the proper group assigned.

njeudy:
@pedrobaeza ok for you ?

addons/account/migrations/11.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,93 @@
...
+    <field name="name">Invoicing: Invoice email</field>
+  </record>
+  <record model="account.payment.term" id="account_payment_term_immediate">
+    <field name="note">Payment terms: Immediate Payment</field>
+  </record>
+  <record model="ir.sequence" id="sequence_payment_supplier_refund">
+    <field name="name">Payments supplier credit notes sequence</field>
+  </record>
+  <record model="account.payment.term" id="account_payment_term_15days">
+    <field name="note">Payment terms: 15 Days</field>
```
StefanRijnhart, line 83: Prefer not to reset payment terms, which may be modified. And nothing substantially was changed anyway compared to https://github.com/odoo/odoo/blob/10.0/addons/account/data/account_data.xml#L28

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,193 @@
...
+account      / res.company              / account_setup_bank_data_done (boolean): NEW 
+account      / res.company              / account_setup_bar_closed (boolean): NEW 
+account      / res.company              / account_setup_coa_done (boolean): NEW 
+account      / res.company              / account_setup_company_data_done (boolean): NEW 
+account      / res.company              / account_setup_fy_data_done (boolean): NEW 
+account      / res.company              / tax_exigibility (boolean)     : NEW 
+account_tax_cash_basis / account.tax              / use_cash_basis (boolean)      : selection_keys is now '['on_invoice', 'on_payment']' ('False')
+account_tax_cash_basis / account.tax              / use_cash_basis (boolean)      : type is now 'selection' ('boolean')
+account_tax_cash_basis / account.tax              / use_cash_basis (boolean)      : was renamed to tax_exigibility [nothing to do]
+# NOTHING TO DO
```
StefanRijnhart, line 75: looks like we need to make sure that tax_exigibility is set to 'on_payment' if use_cash_basis was set to true, and 'on_invoice' otherwise.

StefanRijnhart: Set the company's tax_exigibility to True if any of its taxes was set to use_cash_basis as well.

mvaled:
Hi all, 

Do you need more hands with this?

pedrobaeza:
Please attend latest comments.

AMhadhbi:
Hi all, I would like to know if the module account already done migrated? because I tested with my database and the server is still running more than one hour ...
here 
screen shot 2018-03-08 at 14 44 39

the server stuck in line  **Storing computed values of account.partial.reconcile.max_date**

legalsylvain:
@AMhadhbi. Such duration sometimes occure for new computed field. Nothing's anormal. could you mention how many record do you have in your table account_partial_reconcile ?

To reduce migration duration, you can work to write an SQL script that populate correctly and more quickly the new field.

regards.

AMhadhbi:
@legalsylvain thanks for your fast reply, in my table there is **12412 records.** I have a big data in my account data. 
Yes I would like to reduce migration duration, how can I work please with SQL script, have you an example? the script SQL should write in script migration account? 
regrads.

legalsylvain:
Hi @AMhadhbi 
basically, to reduce migration time of a computed field you should : 

1. check how the compute function works. In that case, that function https://github.com/OCA/OpenUpgrade/blob/11.0/addons/account/models/account_move.py#L1546

2. Add in the pre-migration script, a function that 
* create the field early
* populate the values, by an SQL request.

See exemple here. https://github.com/OCA/OpenUpgrade/blob/8.0/addons/account/migrations/8.0.1.1/pre-migration.py#L25

3. make a Pull request to share your work !

kind regards.

StefanRijnhart:
Thank you so far @rruebner. We have not heard from you for sometime so I've taken the liberty to pick up where you left off here: https://github.com/OCA/OpenUpgrade/pull/1295

reinka:
@legalsylvain I tried to [implement your suggestion ](https://github.com/OCA/OpenUpgrade/pull/1195#issuecomment-371500223)

    def pre_compute_max_date_field(cr):
        """Create and compute values for the new computed field max_date.
    
        Bypass the ORM auto creation and computation because it can take
        a long time and consume a lot of memory for large
        account_partial_reconcile.
        """
        cr.execute("""
                ALTER TABLE account_partial_reconcile
                ADD COLUMN max_date DATE;
                """)
    
        cr.execute("""
                UPDATE account_partial_reconcile
                    SET max_date = (
                      SELECT
                        CASE WHEN ml.date > ml2.date
                          THEN ml.date
                          ELSE ml2.date
                        END
                        From account_partial_reconcile pr
                        LEFT JOIN account_move_line ml
                          ON pr.debit_move_id = ml.id
                        LEFT JOIN account_move_line ml2
                          on pr.credit_move_id = ml2.id
                      WHERE account_partial_reconcile.id = pr.id
                    );
                """)

However, now the computation of `account.move.line.tax_base_amount` (which normally happens right before the `max_date` computation) is **extremly** slow (it's been runing for 25 min on my local machine right now and still has not ended). Without the `max_date` pre-migration step that computation goes almost instantly. Any thoughts on this?

PR #1196, opened 2017-11-17
[11.0][MIG] sale_stock migration

yuriqp:
Depends on:

- [x] stock

sale_stock migration

Current behavior before PR:

--


addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,45 @@
+---Fields in module 'sale_stock'---
+sale_stock   / procurement.group        / sale_id (many2one)            : NEW relation: sale.order
+# Nothing to do
```
pedrobaeza: You need to fill this field, or everything previous to the migration will not be connected, including pending sales orders. Get it from the inverse `procurement_group_id` field in sale.order or old procurement order.

addons/sale/migrations/11.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def update_field_module(cr):
```
pedrobaeza: I think this should be a function in openupgradelib.

addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,45 @@
...
+
+sale_stock   / sale.order               / picking_ids (many2many)       : not a function anymore
+sale_stock   / sale.order               / picking_ids (many2many)       : type is now 'one2many' ('many2many')
+# Nothing to do
+
+sale_stock   / sale.order               / procurement_group_id (many2one): previously in module sale
+# Done, rename references on sale pre-migration script
+
+sale_stock   / sale.order.line          / move_ids (one2many)           : NEW relation: stock.move
+sale_stock   / stock.move               / sale_line_id (many2one)       : NEW relation: sale.order.line
```
pedrobaeza: This field should be filled also or the work flows will break. You have to get it from old procurement orders.

addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,45 @@
...
+NEW ir.ui.view: sale_stock.product_template_view_form_inherit_stock
+# Nothing to do
+
+NEW ir.ui.view: sale_stock.res_config_settings_view_form_sale
+NEW ir.ui.view: sale_stock.res_config_settings_view_form_stock
+# Nothing to do
+
+NEW ir.ui.view: sale_stock.stock_location_route_view_form_inherit_sale_stock
+DEL ir.ui.view: sale_stock.stock_location_route_form_view_inherit_sale_stock
+# Done, xml_ids renamed
```
pedrobaeza: Not needed. As they are noupdate=0 records, they are removed/added automatically.

pedrobaeza (changes requested):

addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
yuriqp: @pedrobaeza I wonder if something similar would be needed for the field move_ids

addons/sale/migrations/11.0.1.1/pre-migration.py
yuriqp: @pedrobaeza how should I proceed, create a PR in openupgradelib with this function?

pedrobaeza: Yes!

yuriqp: @pedrobaeza : I created a new method to handle this case in openupgradelib (https://github.com/OCA/openupgradelib/pull/92). After this is merged, I will update this.

addons/sale_stock/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+
+def update_stock_move_field_from_procurement_order(cr):
+    """
+    Filling the values of new sale_order_id field of stock_move
+    with respective values of old procurement_order
+    """
+    query = ("UPDATE stock_move AS sm "
+             "SET sale_order_id = po.sale_order_id "
```
njeudy: I do not have `sale_order_id` on `procurement.order` model in V10 only `sale_line_id` if i'm right ?

njeudy (changes requested):

addons/sale_stock/migrations/11.0.1.0/post-migration.py
yuriqp: you are right

njeudy:
@yuriqp @pedrobaeza I push a merde request to correct the tavis fail.

It's because sale module is not correctly migrated.

pedrobaeza:
Well, this is going also to be incorrect until stock module is migrated.

njeudy:
@pedrobaeza there is a conflict with #1251 .. we should merge #1251 and rebase to merge this one I think. Ok add sale part to #1251 .

what do you think?

pedrobaeza:
This PR can't be merged until `stock` is migrated.

njeudy:
@pedrobaeza so have to merge #1251  and rebase this merge request tu update pre-migration.py ?

pedrobaeza:
Yeah, that one can be merged as there's no other dependency and then rebase this one for solving conflicts, and then wait for `stock` migration.

nadiaafa:
Hi @pedrobaeza , can i do stock migration? i know you are busy but we need a feedback from you. Thank you.

jholze:
@pedrobaeza could you give us a feedback please.

pedrobaeza:
I'm stuck with the quant merge procedure. I'm making it at openupgradelib level, but I have to deal with a lot of possible inconsistencies during the migration process.

nadiaafa:
@pedrobaeza can i help please? can we have a hangout meeting maybe today or tomorrow according to your availability? thank you for your efforts.

jholze:
@pedrobaeza Morning, just a question, possible that you push the code you still have ?

AMhadhbi:
Hi @pedrobaeza how was the updated for migration module stock and the quant merge procedure. any updated  at openupgradelib level?

MiquelRForgeFlow (approved):
:green_apple: 

I think this can be merged

addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,49 @@
...
+# Nothing to do
+
+sale_stock   / stock.move               / sale_line_id (many2one)       : NEW relation: sale.order.line
+# Filled in post-migration script
+
+sale_stock   / stock.move               / to_refund_so (boolean)        : DEL
+# Nothing to do
+
+sale_stock   / stock.picking            / sale_id (many2one)            : not a function anymore
+# Nothing to do, converted to a related
```
StefanRijnhart: It is converted to a related field on the procurement group's sale_id, but this is set in the post script. The field is recomputed during the migration for every picking (which can take some time on a large database) but at that time, the procurement group's sale id is not set in your scripts.
For best performance, pre-create both fields in the pre migration script and fill them there using SQL.

addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,49 @@
...
+# Done, rename references on sale pre-migration script
+
+sale_stock   / sale.order.line          / move_ids (one2many)           : NEW relation: stock.move
+# Nothing to do
+
+sale_stock   / stock.move               / sale_line_id (many2one)       : NEW relation: sale.order.line
+# Filled in post-migration script
+
+sale_stock   / stock.move               / to_refund_so (boolean)        : DEL
+# Nothing to do
```
StefanRijnhart: Copy values into `to_refund` if stock_account is installed. You may have to create the column as the module loading order seems undefined to me.

addons/sale_stock/migrations/11.0.1.0/openupgrade_analysis_work.txt
AMhadhbi: @StefanRijnhart, you are right. During my migration database to odoo11 I found that there is no previous data to link sale order and stock picking because the procurement group's sale id is not set so I added in the post migration new function to fix this issue : 

```python
def update_stock_picking_field_from_procurement_group(cr):
    """
    Filling the values of new sale_id field of stock_picking
    with respective values of old procurement_group
    """
    qery = ("UPDATE stock_picking AS sp "
             "SET sale_id = pg.sale_id "
             "FROM procurement_group AS pg "
             "WHERE sp.group_id = pg.id")
    openupgrade.logged_query(
        cr, query, ()
    )
```
@yuriqp  could you please check Stefan review's to approve this module and will be merged. thanks

pedrobaeza (approved):

pedrobaeza:
I have fixed remaining comments, so we can merge.

PR #1197, opened 2017-11-17
[11.0][MIG]11.0 payment

nadiaafa:

Current behavior before PR:

--


addons/hr/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,17 @@
+---Fields in module 'hr'---
```
StefanRijnhart: Can you submit the PR without these HR changes?

addons/payment/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,168 @@
...
+                <li><i class="fa fa-check"/>Manual Capture
+                </li>
+            </ul>
+        </field>
+        <field name="payment_icon_ids"
+               eval="[(6, 0, [ref(&quot;payment.payment_icon_cc_maestro&quot;),                                                         ref(&quot;payment.payment_icon_cc_mastercard&quot;),                                                         ref(&quot;payment.payment_icon_cc_discover&quot;),                                                         ref(&quot;payment.payment_icon_cc_diners_club_intl&quot;),                                                         ref(&quot;payment.payment_icon_cc_jcb&quot;),                                                         ref(&quot;payment.payment_icon_cc_visa&quot;)])]"/>
+        <field name="sequence">3</field>
+    </record>
+
+</odoo>
```
MiquelRForgeFlow: ~~add new line below~~ (done)

MiquelRForgeFlow (approved):

addons/payment/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,168 @@
...
+        <field name="description" type="html">
+            <p>
+                A generic payment acquirer to send payment instructions to any payment acquirer.
+            </p>
+            <ul class="list-inline">
+                <li><i class="fa fa-check"/>Online Payment
+                </li>
+            </ul>
+        </field>
+        <field name="auto_confirm"/>
```
StefanRijnhart: This is a false positive from the compare-noupdate-xml script. It notices that there is no value present anymore for this field, and as a result it adds an empty element. But the reason for the missing value is that the field was removed. In that case, it is better to remove these elements to prevent the following messages in the migration log

`WARNING openupgrade odoo.models: payment.acquirer.write() with unknown fields: auto_confirm`

addons/payment/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,48 @@
+---Fields in module 'payment'---
+payment      / payment.acquirer         / auto_confirm (selection)      : DEL required: required, selection_keys: ['authorize', 'confirm_so', 'generate_and_pay_invoice', 'none'], req_default: function
+payment      / payment.acquirer         / capture_manually (boolean)    : NEW 
+payment      / payment.acquirer         / country_ids (many2many)       : NEW relation: res.country
+payment      / payment.acquirer         / payment_flow (selection)      : NEW required: required, selection_keys: ['form', 's2s'], req_default: function
+payment      / payment.acquirer         / payment_icon_ids (many2many)  : NEW relation: payment.icon
```
pedrobaeza, line 6: You should move old version image in payment.acquirer to a payment.icon record, taking care of not duplicating standard ones. This way, you would preserve any custom image you have put in your current payment methods.

addons/payment/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,48 @@
...
+payment      / payment.acquirer         / specific_countries (boolean)  : NEW 
+payment      / payment.icon             / acquirer_ids (many2many)      : NEW relation: payment.acquirer
+payment      / payment.icon             / image (binary)                : NEW attachment: True
+payment      / payment.icon             / image_payment_form (binary)   : NEW attachment: True
+payment      / payment.icon             / name (char)                   : NEW 
+payment      / payment.token            / verified (boolean)            : NEW 
+payment      / payment.transaction      / callback_eval (char)          : DEL
```
pedrobaeza, line 13: I have been searching in current 11.0 code (not OU one, which is outdated), possible uses of new fields that can be referred to this old field in previous version, but I can't find any. Did you figure out any possible translation needed for this field?

pedrobaeza (changes requested):

addons/payment/migrations/11.0.1.0/openupgrade_analysis_work.txt
StefanRijnhart: Not sure if that is useful. It will likely lead to duplicates. If people customized this, they can add the same image again.

pedrobaeza: It is, Stefan! Don't make the customer to put again their specific icons for the payments. I have several with custom ones.

nadiaafa: @pedrobaeza i didn't get it, payment.icon seems to store icon's cards, and payment.acquirer icon didn't change(stored as ir.attachment) so it will be stored after migration even if there is custom ones.

pedrobaeza: You are right. I have been confused with the name. Nothing to do indeed.

pedrobaeza:
`auto_confirm` field should still be removed from XML file

nadiaafa:
@pedrobaeza i remode auto_confirm, thank you.

pedrobaeza (approved):

PR #1199, opened 2017-11-17
[FIX] board: Include noupdate data not detected by script

pedrobaeza:
The script doesn't inspect `<menuitem>` and other shortcut elements

@Tecnativa


rafaelbn (approved):
Tested in customer environment

addons/board/migrations/10.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record model="ir.ui.menu" id="menu_board_my_dash">
```
MiquelRForgeFlow, line 3: I don't see what is the change. The exact same record is in v10. Nothing has changed.

pedrobaeza: Please check again: the parent is not the same.

MiquelRForgeFlow: v10: https://github.com/OCA/OCB/blob/10.0/addons/board/views/board_views.xml#L45

        parent="base.menu_board_root"
v11: https://github.com/OCA/OCB/blob/11.0/addons/board/views/board_views.xml#L45

        parent="base.menu_board_root"

pedrobaeza: You are looking the wrong branch :wink:

https://github.com/OCA/OCB/blob/9.0/addons/board/board_view.xml#L31

MiquelRForgeFlow: Ah, stupid of me. I saw your "added this to the **11.0** milestone 3 days ago" and that tricked me. :laughing:

pedrobaeza: Yeah, I changed it immediately because I clicked on the wrong line...

MiquelRForgeFlow: Hey, but the <code>id</code> is different! And also the <code>sequence</code>, <code>action</code> and <code>groups</code>.

pedrobaeza: Wrong link: https://github.com/OCA/OCB/blob/9.0/addons/board/board_mydashboard_view.xml#L45

MiquelRForgeFlow (approved):

PR #1200, opened 2017-11-20
10.0

Aaqiu:
     Hi, my English is very poor, so I hope you can understand what I will say next. I used your module which is named base_import_match, and here is the problem, I can't import the one2many fields, the error message after Vaildated is like this:
No matching record found for name 'Casting' in field 'Hierarchy' at row 2

Can you help me out? Looking forward to your reply, THX!

--


pedrobaeza:
Sorry, OpenUpgrade has nothing to do with base_import_match.

PR #1201, opened 2017-11-22
[FIX+IMP] hr_expense: Fixes and improvements to current script

pedrobaeza:
* RFR: Extract methods with some migration operations.
* IMP: Use openupgrade.logged_query for seeing results in the log.
* IMP: Don't use subqueries in UPDATE where no needed
* FIX: set expense responsible was setting employee IDs, where it should be user IDs.
* IMP: Use openupgradelib `map_values` instead of several plain SQLs.
* FIX: Assign correctly product accounts using ORM and taking into account multi-company

There are still 2 operations without being checked, that are still in the main method.

@Tecnativa

### DIFF
--- a/addons/hr_expense/migrations/10.0.2.0/post-migration.py
+++ b/addons/hr_expense/migrations/10.0.2.0/post-migration.py
@@ -23,11 +23,13 @@ def set_expense_responsible(env):
     openupgrade.logged_query(
         env.cr, """
         UPDATE hr_expense_sheet s
-        SET responsible_id = e.user_id
+        SET responsible_id = r.user_id
         FROM hr_department d,
-             hr_employee e
+             hr_employee e,
+             resource_resource r
         WHERE s.department_id = d.id
-            AND d.manager_id = e.id""",
+            AND d.manager_id = e.id
+            AND r.id = e.resource_id""",
     )
 
 
--- a/addons/hr_expense/migrations/10.0.2.0/post-migration.py
+++ b/addons/hr_expense/migrations/10.0.2.0/post-migration.py
@@ -1,7 +1,65 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 Le Filament (<https://le-filament.com>)
+# Copyright 2017 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 from openupgradelib import openupgrade
+
+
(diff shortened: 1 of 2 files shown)


addons/hr_expense/migrations/10.0.2.0/post-migration.py
```diff
@@ -1,9 +1,69 @@
...
+        ], table='hr_expense',
+    )
+
+
+def fill_expense_product_account(env):
+    """Set correct account corresponding to product subject to this expense."""
+    for expense in env['hr.expense'].search([]):
+        expense.account_id = expense.product_id.with_context(
+            force_company=expense.company_id.id,
+        ).product_tmpl_id._get_product_accounts()['expense']
```
StefanRijnhart, line 53: You could build a map of account values for all unique combinations of (product, company) extracted with an sql query

```SELECT product_id, company_id, ARRAY_AGG(id) FROM hr_expense group by product_id, company_id```

and then write the expenses in groups. If you have a large amount of expenses it could improve performance. But it's just a suggestion, I'll leave it up to you.

**Review by StefanRijnhart — APPROVED** [review]

File: addons/hr_expense/migrations/10.0.2.0/post-migration.py (hunk 1 above)
Comment (pedrobaeza) [reply addons/hr_expense/migrations/10.0.2.0/post-migration.py:53]: It's not so easy, Stefan, due to:

- The account can be at product or category level.
- Properties themselves have 3 possible level of values: general default value, company default value, record value.

So this performance trade-off avoids possible problems.

Comment (StefanRijnhart) [reply addons/hr_expense/migrations/10.0.2.0/post-migration.py:53]: Pedro, if you retrieve the account for every combination of (product, company) you have all of that covered.

Comment (pedrobaeza) [reply addons/hr_expense/migrations/10.0.2.0/post-migration.py:53]: It still lacks the default value without company for those without specific value

Comment (StefanRijnhart) [reply addons/hr_expense/migrations/10.0.2.0/post-migration.py:53]: Let me clarify: 
```
self.env.cr.execute("SELECT product_id, company_id, ARRAY_AGG(id) FROM hr_expense group by product_id company_id")
for product_id, company_id, expense_ids in self.env.cr.fetchall():
    self.env['hr.expense'].browse(expense_ids).write({
        'account_id': self.env['product.product'].with_context(
            force_company=company_id).browse(
                product_id).product_tmpl_id._get_product_accounts()['expense'].id})
```
I don't think you will run into problems of missing accounts with that, do you?

PR #1202, opened 2017-11-23
[ADD] Migration of purchase order line invoice lines

StefanRijnhart:
Odoo 8.0 introduced account.invoice.line's Many2one 'purchase_line_id but was not used until 9.0. Here, the 'invoice_lines' counterpart on the purchase order line is migrated to a One2many field. The field already exists, so it is easy to do in the pre-script so that the computation of the invoiced quantities will be correct.

### DIFF
--- a/addons/purchase/migrations/9.0.1.2/openupgrade_analysis_work.txt
+++ b/addons/purchase/migrations/9.0.1.2/openupgrade_analysis_work.txt
@@ -51,7 +51,7 @@ purchase     / purchase.order.line      / date_planned (date)           : type i
 #Nothing To Do : will get datetime automatically.
 
 purchase     / purchase.order.line      / invoice_lines (many2many)     : type is now 'one2many' ('many2many')
-#Nothing To Do : As there is already m2o (purchase_line_id) for purchase.order.line in account.invoice.line.
+#Done: migrated Many2many table to Many2one field on the invoice line (purchase_line_id)
 
 purchase     / purchase.order.line      / invoiced (boolean)            : DEL 
 #Nothing To Do : Functionality removed now no longer needed.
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
purchase / account.invoice / purchase_id (many2one) : NEW relation: purchase.order
        #Nothing To Do : New Functionality added.

I think that this field also may need a script (executed after your script) to get filled. Maybe:

        UPDATE account_invoice ai
        SET purchase_id = pol.order_id
        FROM account_invoice_line ail
        LEFT JOIN purchase_order_line pol ON ail.purchase_line_id = pol.id
        WHERE ail.invoice_id = ai.id

StefanRijnhart:
@mreficent `purchase_id` on the invoice is a dummy field to add the purchase order's lines to the invoice. If you check the onchange method, you can see that the field is set to False immediately after creating invoice lines for the purchase order https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/account_invoice.py#L78

MiquelRForgeFlow:
@StefanRijnhart In theory, yes. But may you can have another module that changes this method in a way that it's not put to False. I think we should not be guided by the methods, only by the fields and how they are declared. And the `purchase_id` of account_invoice is a stored not related not computed field.

StefanRijnhart:
@mreficent I disagree with that. You should always look at how a field is used. In this case, it is used as an oblique method to add new invoice lines. I think leaving the field empty is our best option.

StefanRijnhart:
By the way, if in your words, there is another module that changes the semantics of the field, the migration of that module should take care of filling in the values.

MiquelRForgeFlow (approved):

PR #1207, opened 2017-11-26
[9.0]  [FIX] purchase: PO line taxes were not kept because m2m rel table renamed

### LINKED ISSUE #1205: v8 -> v9 Taxes lost in purchase.order.line
When you migrate from v8 to v9 with openupgrade, taxes are lost on purchase.order.line.

My analysis of the problem:
field in v8 (old API):
```
'taxes_id': fields.many2many('account.tax', 'purchase_order_taxe', 'ord_id', 'tax_id', 'Taxes', domain=['|', ('active', '=', False), ('active', '=', True)]),
```
So the many2many link table is "purchase_order_taxe" with 2 columns "ord_id" and "tax_id"

field in v9 (new API):
```
taxes_id = fields.Many2many('account.tax', string='Taxes', domain=['|', ('active', '=', False), ('active', '=', True)])
```
So the M2M field name is the same as in v8 ("taxes_id"), but as the table and field name are not given in field definition, they are auto-generated, and therefore the table name is "account_tax_purchase_order_line_rel" and the field names are "purchase_order_line_id" and "account_tax_id".

I had a quick look in the openupgrade scripts of the purchase module between v8 and v9 and it seems that it doesn't handle the rename of the table and its 2 columns. This certainly explains the bug.

pedrobaeza:
Closes #1205 

@Tecnativa

### DIFF
--- a/addons/purchase/migrations/9.0.1.2/pre-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/pre-migration.py
@@ -19,6 +19,17 @@ field_renames = [
      'fiscal_position_id'),
 ]
 
+column_renames = {
+    'purchase_order_taxe': [
+        ('ord_id', 'purchase_order_line_id'),
+        ('tax_id', 'account_tax_id'),
+    ]
+}
+
+table_renames = [
+    ('purchase_order_taxe', 'account_tax_purchase_order_line_rel'),
+]
+
 
 def map_order_state(cr):
     """ Map values for state field in purchase.order and purchase.order.line.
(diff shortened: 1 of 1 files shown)


alexis-via:
Don't forget that you also have to rename the 2 fields in that table, cf my bug report.

pedrobaeza:
@alexis-via thanks for the reminder. I didn't check that part. Now they are renamed also.

MiquelRForgeFlow (approved):

alexis-via (approved):
Code review, no test

StefanRijnhart (approved):

PR #1210, opened 2017-11-30
[9.0][ADD] account_invoice_reorder_lines

pedrobaeza:
The feature has been included now in core. The field sequence already existed in 8, but the _order was not set. Now that order has been set, and the field name is the same, so nothing more to do, except merging the module for avoiding a missing module message.
 
@Tecnativa

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -87,6 +87,7 @@ def cleanup_modules(cr):
             # from OCA/crm - included in core
             ('crm_lead_lost_reason', 'crm'),
             # from OCA/sale-workflow - included in core
+            ('account_invoice_reorder_lines', 'account'),
             ('sale_order_back2draft', 'sale'),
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
Thus, you should delete the module from `apriori` file.

pedrobaeza:
Uhm, I see, but why have you continued with that module if core makes already the order? Can you ask in your company?

MiquelRForgeFlow:
Some people like to migrate things like a hobby. I will ask to @aheficent

AaronHForgeFlow:
@pedrobaeza You're are right. But we also need to see the sequence number in the view because it's easier in invoices with many invoice lines. Perhaps the module should be refactored and only show the  sequence number in the view.

pedrobaeza:
But then it should be optional to install that additional feature. People coming from the old v8 module only need core. I will remove the apriori reference then.

pedrobaeza:
I have removed the other reference. Please squash on merge.

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:

PR #1211, opened 2017-11-30
[IMP] account: Remove view accounts

pedrobaeza:
On v8, they were not selectable on documents and they are only for support in totals, but without more use. On v9, as is, they are selectable and can be confused in any selection, so better remove them, previously making a copy for preserving them in case you need.

@Tecnativa

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -7,6 +7,7 @@
 import operator
 from openupgradelib import openupgrade
 from openerp.modules.registry import RegistryManager
+from psycopg2.extensions import AsIs
 
 
 def map_bank_state(cr):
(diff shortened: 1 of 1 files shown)


rafaelbn (approved):

StefanRijnhart:
This is perfectly acceptable IMHO, but just curious why you did not set the accounts to 'deprecated' instead?

pedrobaeza:
I prefer to clean them out than to be kept with the rest of possible deprecated accounts you make on your normal v9 accounting.

StefanRijnhart (approved):
Thanks for your feeback!

rb3606:
Because of deletion of these accounts its giving me error in migration. 

IntegrityError: null value in column "account_id" violates not-null constraint
CONTEXT:  SQL statement "UPDATE ONLY "public"."account_invoice_line" SET "account_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "account_id""

How to solve it ?

pedrobaeza:
You mustn't select view accounts in your invoice lines. That's an inconsistency in your DB.

rb3606:
Thanks @pedrobaeza

legalsylvain:
Hi @rb3606.

I faced the same problem. I replace the ``DELETE FROM account_account`` by 
``
       UPDATE account_account
        SET name = name || ' (DISABLED BY OpenUpgrade 9.0)',
        active = false
``

**Pro :** 
- can migrate without integrity error
- faster. (unlink massively a lot of ``account.account`` make the upgrade slow).

**Cons :** 
- something is wrong in your database. If it's about old account.move / account.invoice maybe there is nothing to do, but you should take a look.

rb3606:
@legalsylvain good idea..  but we also need to migrate  invoice lines which have view account set.

Issue #1212, opened 2017-12-01
[9.0] delivery_carrier table wrong ids 


AaronHForgeFlow:
When migrated a database from v8 to v9 the ids of the delivery_carrier table changed unexpectedly:

In the following image I am showing the ids and the name of the carrier. See in first place the values in v8, then the values in v9 after migrate and then the openupgrade_legacy table for carrier_id. The wrong rows are highlighted in red.

![image](https://user-images.githubusercontent.com/19620251/33494451-914ad64c-d6c3-11e7-8b35-0f463a55d3b1.png)

@hbrunn @pedrobaeza  Can you shed some light here? :confused:

pedrobaeza:
That's because the delivery carrier is now populated with the lines of v8. That table is renamed to delivery_carrier. This is for propagating correctly all the fields that are in the lines, which are more suitable. If other modules use a reference to all table, you need a migration script for them.

pedrobaeza:
Fixed in #1213

PR #1213, opened 2017-12-04
[9.0][IMP]reassign carried_id

AaronHForgeFlow:
This solves this issue: https://github.com/OCA/OpenUpgrade/issues/1212

When migrating, the ids for table delivery_carrier may change. There are references in sale_order and in stock_picking to delivery_carrier that need to be updated.

### DIFF
--- a/addons/delivery/migrations/9.0.1.0/post-migration.py
+++ b/addons/delivery/migrations/9.0.1.0/post-migration.py
@@ -28,9 +28,10 @@ def rename_property(cr, model, old_name, new_name):
 
 def reassign_carrier_id(cr):
     cr.execute("""
-        select id, openupgrade_legacy_9_0_carrier_id from delivery_carrier
-        where id != openupgrade_legacy_9_0_carrier_id;
-    """)
+        select id, %s from delivery_carrier
+        where id != %s;
+    """ % (openupgrade.get_legacy_name('carrier_id'),
+           openupgrade.get_legacy_name('carrier_id')),)
     for new_id, old_id in cr.fetchall():
         if openupgrade.table_exists(cr, 'sale_order') and \
                 openupgrade.column_exists(cr, 'sale_order', 'carrier_id'):
--- a/addons/delivery/migrations/9.0.1.0/post-migration.py
+++ b/addons/delivery/migrations/9.0.1.0/post-migration.py
@@ -28,20 +28,20 @@ def rename_property(cr, model, old_name, new_name):
 
 def reassign_carrier_id(cr):
     cr.execute("""
-        select id, openupgrade_legacy_9_0_carrier_id from delivery_carrier 
-        where id != openupgrade_legacy_9_0_carrier_id; 
+        select id, openupgrade_legacy_9_0_carrier_id from delivery_carrier
+        where id != openupgrade_legacy_9_0_carrier_id;
     """)
     for new_id, old_id in cr.fetchall():
         if openupgrade.table_exists(cr, 'sale_order') and \
                 openupgrade.column_exists(cr, 'sale_order', 'carrier_id'):
             cr.execute("""
-                UPDATE sale_order set carrier_id = %s 
+                UPDATE sale_order set carrier_id = %s
                 where carrier_id = %s
             """ % (new_id, old_id))
         if openupgrade.table_exists(cr, 'stock_picking') and \
                 openupgrade.column_exists(cr, 'stock_picking', 'carrier_id'):
             cr.execute("""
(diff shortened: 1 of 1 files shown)


addons/delivery/migrations/9.0.1.0/post-migration.py
```diff
@@ -26,6 +26,44 @@ def rename_property(cr, model, old_name, new_name):
         (new_name, field_ids))
 
 
+def reassign_carrier_id(cr):
+    print("""Delivery carrier ids have changed""")
```
pedrobaeza: Don't print

addons/delivery/migrations/9.0.1.0/post-migration.py
```diff
@@ -26,6 +26,44 @@ def rename_property(cr, model, old_name, new_name):
...
+    update_stock_picking = False
+    if cr.fetchone():
+        update_stock_picking = True
+
+    cr.execute("""
+        select id, openupgrade_legacy_9_0_carrier_id from delivery_carrier 
+        where id != openupgrade_legacy_9_0_carrier_id; 
+    """)
+    for new_id, old_id in cr.fetchall():
+        if update_sale:
```
pedrobaeza: Use `openupgrade.column_exists`

pedrobaeza (changes requested):

addons/delivery/migrations/9.0.1.0/post-migration.py
```diff
@@ -26,6 +26,44 @@ def rename_property(cr, model, old_name, new_name):
...
+    if cr.fetchone():
+        update_sale = True
+
+    update_stock_picking = False
+    cr.execute("""
+        SELECT table_name
+        FROM information_schema.tables
+        WHERE table_name='stock_picking'
+    """)
+    update_stock_picking = False
```
MiquelRForgeFlow: delete this line because it's repeated

MiquelRForgeFlow (changes requested):

addons/delivery/migrations/9.0.1.0/post-migration.py
pedrobaeza: It's not needed directly using `openupgrade.column_exists`

addons/delivery/migrations/9.0.1.0/post-migration.py
```diff
@@ -26,6 +26,26 @@ def rename_property(cr, model, old_name, new_name):
...
+    """)
+    for new_id, old_id in cr.fetchall():
+        if openupgrade.table_exists(cr, 'sale_order') and \
+                openupgrade.column_exists(cr, 'sale_order', 'carrier_id'):
+            cr.execute("""
+                UPDATE sale_order set carrier_id = %s 
+                where carrier_id = %s
+            """ % (new_id, old_id))
+        if openupgrade.table_exists(cr, 'stock_picking') and \
+                openupgrade.column_exists(cr, 'stock_picking', 'carrier_id'):
```
MiquelRForgeFlow: line too long

AaronHForgeFlow: the length is 78 characters, is that bad?

MiquelRForgeFlow: ah sorry, I thought that was the travis error 😅

pedrobaeza:
The errors are:
```
./addons/delivery/migrations/9.0.1.0/post-migration.py:31:75: W291 trailing whitespace
./addons/delivery/migrations/9.0.1.0/post-migration.py:32:55: W291 trailing whitespace
./addons/delivery/migrations/9.0.1.0/post-migration.py:38:54: W291 trailing whitespace
./addons/delivery/migrations/9.0.1.0/post-migration.py:44:57: W291 trailing whitespace
```
Put in your IDE configuration to strip them.

addons/delivery/migrations/9.0.1.0/post-migration.py
```diff
@@ -26,6 +26,26 @@ def rename_property(cr, model, old_name, new_name):
         (new_name, field_ids))
 
 
+def reassign_carrier_id(cr):
+    cr.execute("""
+        select id, openupgrade_legacy_9_0_carrier_id from delivery_carrier
```
pedrobaeza: It's better to use `openupgrade.get_legacy_name('carrier_id')`

MiquelRForgeFlow: and also in line 32

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #1214, opened 2017-12-05
Fix/9.0/account/fiscalyear sequence

StefanRijnhart:


addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -336,6 +336,31 @@ def map_account_tax_template_type(cr):
         table='account_tax_template', write='sql')
 
 
+def migrate_account_sequence_fiscalyear(cr):
+    """ Migrate subsequences from fiscalyears to sequence date ranges
+    so that the next number is aligned with the last assigned number """
+    openupgrade.logged_query(
+        cr,
+        """ \
```
pedrobaeza, line 344: Why the `\`?

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -336,6 +336,31 @@ def map_account_tax_template_type(cr):
...
+        WHERE asf.fiscalyear_id = af.id
+             and asf.sequence_id = irs.id; """)
+    openupgrade.logged_query(
+        cr,
+        """ \
+        UPDATE ir_sequence SET
+            prefix = REPLACE(prefix, '%%(year)s', '%%(range_year)s'),
+            suffix = REPLACE(suffix, '%%(year)s', '%%(range_year)s'),
+            use_date_range = TRUE
+        WHERE id IN (SELECT sequence_id FROM ir_sequence_date_range)
```
pedrobaeza, line 360: Use `...IN (SELECT DISTINCT(sequence_id) FROM...` for avoiding repetitive updates.

addons/account/migrations/9.0.1.1/post-migration.py
StefanRijnhart: I *think* it looks more aligned in the logfile.

addons/account/migrations/9.0.1.1/post-migration.py
StefanRijnhart: Possible SQL antipattern: https://www.quora.com/Should-I-use-DISTINCT-in-a-subquery-when-using-IN and this set of records is always very small so I'd rather leave this as is.

pedrobaeza: OK, got it. You're right

pedrobaeza (approved):

PR #1217, opened 2017-12-07
[FIX] don't choke on deleted stages

hbrunn:
stages are noupdate: https://github.com/OCA/OpenUpgrade/blob/9.0/addons/crm/data/crm_stage_data.xml


pedrobaeza (approved):

pedrobaeza:
There's an strange error in Travis...

pedrobaeza (changes requested):
This seems either the XML file is not correct or the code in openupgradelib for handling `init_no_create` mode is not correct.

hbrunn:
@pedrobaeza fixed, I passed the parameter wrongly, shame!

pedrobaeza (approved):

PR #1220, opened 2017-12-09
11.0 crm migration

nadiaafa:

Current behavior before PR:

--


addons/crm/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,22 @@
...
+def migrate(env, version):
+    env.cr.execute("""
+         INSERT INTO mail_activity(res_id, res_model_id, res_model,
+         res_name, activity_type_id, summary,
+         date_deadline, user_id, create_uid, create_date, write_uid, write_date)
+         SELECT cl.id,  im.id, 'crm.lead',cl.title_action, cl.next_activity_id,
+         cl.title_action, cl.date_action,cl.user_id, cl.create_uid, cl.create_date,
+         cl.write_uid, cl.write_date
+         FROM crm_lead as cl, ir_model as im
+         WHERE type='opportunity' AND  date_action IS NOT NULL  AND im.model='crm.lead'
```
pedrobaeza: I miss here the condition `AND next_activity_id IS NOT NULL`, because that's not a required field AFAIK

pedrobaeza: Why filtering only for `type='opportunity'`? Maybe you have a lead with an action.

pedrobaeza (changes requested):

addons/crm/migrations/11.0.1.0/post-migration.py
nadiaafa: @pedrobaeza yes you are right, i updated it with just next_activity_id IS NOT NULL condition.

nadiaafa: @pedrobaeza i have a question about the field used_id of mail.activity in v11, it's a required field, here in my query, i used user_id of crm.lead to fill it, but this field is not required, so should i add a test to check if user_id of crm.lead if not null? or i use just create_uid of crm.lead?

nadiaafa:
@pedrobaeza i made the changes, could you check them please . Thank you

pedrobaeza:
I have taken this PR for fixing several things and improving others:

* Create or match mail_activity_types
* Fix creation of mail_activity records
* Complete migration work file
* Added tests

pedrobaeza (approved):

PR #1225, opened 2017-12-12
[9.0][IMP] deferred: Disable invalid filters

pedrobaeza:
Use new openupgradelib method for disabling invalid filters.

@Tecnativa

### DIFF
--- a/openerp/openupgrade/deferred_90.py
+++ b/openerp/openupgrade/deferred_90.py
@@ -5,7 +5,8 @@
 #    (<https://odoo-community.org>)
 #
 #    Contributors:
-#    -
+#    - Stefan Rijnhart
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
Why in this case it's not put at the `end-migration.py` like you did in v10 and v11?

pedrobaeza:
Because this already exists for other things.

StefanRijnhart:
@mreficent the 'end' stage was only introduced after the release of Odoo 10 and then backported to 9.0, long after work started on OpenUpgrade 9.0. That is why OpenUpgrade 9.0 still uses the deferred script.

StefanRijnhart (approved):

PR #1226, opened 2017-12-12
[10.0][IMP] base: disable invalid filters

pedrobaeza:
Use new openupgradelib method for disabling invalid filters

@Tecnativa


rvalyi (approved):

StefanRijnhart (approved):
Thanks!

PR #1229, opened 2017-12-14
Upstream 11.0 96fcd4b56c78684c1983eb1b3a896955970b1a4a

mvaled:
Update upstream.

There were DB changes after the first release of 11.0. 

Conflicts:

- `requirements.txt`.  They added latest psycogpg2.  

- `odoo/modules/loading.py`.  They added a check for `package.name` possibly None.  I kept the condition, kept the `cr.commit_org()`, but also protected the `OpenUpgrade edit start` with the same condition.

- `README.md`.  I kept OpenUpgrade' s versions (minus white spaces).

- `addons/l10n_br/models/account.py`.  They removed a couple of fields in commit 5d0d80afa5bce8780305173c93e291466a133f7f.  @StefanRijnhart can explain further ;)


pedrobaeza (approved):
Everything seems correct, but I prefer to have a second opinion by @StefanRijnhart on the loading.py part. Next time, please make separate commits for the conflicts found, so we can easily analyze them.

StefanRijnhart (approved):
Thanks @mvaled, great work again! I checked the conflicts that the merge of Odoo@96fcd4b causes and you resolved them all perfectly.

mvaled:
Hi @pedrobaeza,

I would like discuss a little further the procedure to split the conflict-resolution commits from the merge.  I see two variants:

- Commit by resolving by one side (theirs or ours).  Then I would have to remember which files I have to actually merge.

- Commit the merge with conflict "<<<<" markers.  And the do the commits that fix them.  This has the advantage of being (normally) a SyntaxError and would make it easier to spot.

Both alternatives interact badly with `git rerere` (which I have enable globally, but I can disable for OpenUpgrade).  If any, I would opt for the second; but I think it should be like an OCA-way to do upstream merges with conflicts.  So, I put this issue to consideration.

Best regards.

pedrobaeza:
I don't have a strong opinion on that, as I haven't deal yet with conflicts of this type. @hbrunn and @StefanRijnhart has dealt with OCB conflicts this way. What do you think?

hbrunn:
well, I'm also unhappy with the way we do it, but I also don't really see a better option. We could also do it the other way around, force push the odoo branch with just one commit on top where we keep all of the OpenUpgrade patches. As this is bad for nearly everything else I don't propose it.

StefanRijnhart:
I think the way we deal with this now is fine. The conflicts are either trivial or simple to solve, at least for someone that is familiar with the framework. If someone proposes a merge, someone else resolves the same conflicts and arrives at a similar (or equivalent) solution, and tests are still green, then the PR is cleared for merging.

pedrobaeza:
OK, let's continue this way then

PR #1230, opened 2017-12-15
[DBG] Why does pip install fail on travis?

StefanRijnhart:
https://travis-ci.org/OCA/OpenUpgrade/builds/315807044#L642


StefanRijnhart:
Follow up here: https://github.com/OCA/OpenUpgrade/pull/1231

PR #1234, opened 2017-12-15
[FIX] delivery: We don't need to change name, but create product

pedrobaeza:
As now delivery.carrier inherits by delegation from product.product, we need to create a specific product for each carrier, independently from the previous product that was assigned. If not, things like the carrier name will be taken from old product name, carriers will share products...

@Tecnativa

### DIFF
--- a/addons/delivery/migrations/9.0.1.0/pre-migration.py
+++ b/addons/delivery/migrations/9.0.1.0/pre-migration.py
@@ -99,28 +99,40 @@ def fill_missing_delivery_grid_records(cr):
     )
 
 
-def correct_names(env):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@mreficent @aheficent sorry for bothering you again, but this module is my black beast. Continuing the checking, I have discovered that the only solution is to make this for having fixed everything. In fact, I'm removing the TODO.

lasley (approved):
Code and concept 👍  - we ran into similar issue with our Easypost connector

AaronHForgeFlow (approved):
:+1: From my side it's ok like this, I use just one grid for each carrier, that's why the previous approach was valid to me. This is more correct in any case.

Issue #690, opened 2016-12-03
Translations change v8 > v9


pedrobaeza:
I have just detected that the comments to find the translation of fields texts have changed.

In version 8, you have following translation comment for the field label (the `string` attribute)
```
#: field:model.name.with.dots,field_name:0
```

However, in version 9, this is the equivalent comment for translating the field label:
```
#: model:ir.model.fields,field_description:module_name.field_xml_id
```
Being `field_xml_id` generally composed of:  `field_` + `model_name_with_underscores` + `_` + `field_name` (I'm not sure if `model_name_with_underscores` can vary if the `_table_name` attribute is modified in the model).

Are you aware of this, @jbeficent, @hbrunn, @StefanRijnhart ? Have you handled this in any way?

StefanRijnhart:
Currently not handled, I'm afraid. All custom field label translations will be reset by the looks of it.

pedrobaeza:
As there's no easy solution, I close this although it's a pity to lose custom translations from v8 to v9.

StefanRijnhart:
@pedrobaeza I don't agree with closing issues just because we don't have a solution yet. Do you mind if I reopen it?

StefanRijnhart:
What is the biggest problem in this? Is it the performance of a couple of thousands of sql queries?

pedrobaeza:
No, the problem is to determine how to transform old "syntax" to the new one. Check first comment for a bigger explanation.

There's an additional problem that I can expand with fields with `translate=html_translate`.

StefanRijnhart:
Doesn't the following query give you the source and target value for the transformation?

```
SELECT 'field:'||imf.model||','||imf.name||':0',
    'model:ir.model.fields,field_description:'||imd.module||'.'||imd.name 
    FROM ir_model_data imd
    JOIN ir_model_fields imf ON imf.id = imd.res_id
    WHERE imd.model = 'ir.model.fields';
```

pedrobaeza:
But you have to change existing ones to new ones, so that doesn't serve as is. I'm also doubting about the part of model name with underscores.

StefanRijnhart:
I misunderstood. Your comment features the description from the translation file, not from the translation table. But I still don't understand the problem :-(

StefanRijnhart:
Do you want to change the translations in the database or in the translation files of custom modules?

pedrobaeza:
Existing translations in the database. I will write further about this later.

[...]

PR #1241, opened 2017-12-24
[ADD] Migration of field name translations

StefanRijnhart:
Translations of field names are encoded differently in Odoo 9.0:
```
     version |           name                    | res_id |  type
    ---------+-----------------------------------+--------+-------
     8.0     | ir.module.module,summary          |      0 | field
     9.0     | ir.model.fields,field_description |    759 | model
```
This change migrates existing entries.

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -131,6 +131,32 @@ def has_recurring_contracts(cr):
     return False
 
 
+def migrate_translations(cr):
+    """ Translations of field names are encoded differently in Odoo 9.0:
+     version |           name                    | res_id |  type
+    ---------+-----------------------------------+--------+-------
+     8.0     | ir.module.module,summary          |      0 | field
+     9.0     | ir.model.fields,field_description |    759 | model
+    """
+    openupgrade.logged_query(
+        cr, """
+        WITH mapping AS (
+            SELECT imd.module,
+                imf.model||','||imf.name AS name80,
+                'ir.model.fields,field_description' AS name90,
+                imd.res_id
+            FROM ir_model_data imd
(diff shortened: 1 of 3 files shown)


pedrobaeza:
This is what I reported in #690 . Have you find an exact pattern? I didn't have time for investigating this.

Thanks for doing it!

StefanRijnhart:
Yes, this is working fine for field names. I am only tweaking the test now. Are there any other classes of translations that have a different notation that you know of?

pedrobaeza:
Code translations remain the same.

The only type of translation that remains incorrect is the user translations for fields with new `translate=html_translate` attribute. This change is awful, and it makes some parts unusable. Imagine this HTML text:
```
<h3>A title</h3>
<p>First paragraph</p>
<p>Second paragraph</p>
```
On v8, you have a total translation of this field as is, so virtually you can put anything totally different in the translation.

But on >=v9, that text is passed through a filter that split and cleanup translations this way (this is the content of the source field):
`A title`
`First paragraph`
`Second paragraph`

So this makes:

* Functionally: you need to adapt your translation to the exact scheme in the source, which is not something reasonable for a decent CMS.
* Technically: it's barely impossible to adapt old translations to fit the source scheme. Possible tricks is to try to match HTML tags to know which part belongs to the original source, letting empty the not matched ones, and injecting missing ones at the end, but the heuristic is very complicated and for sure there will be some that doesn't work.

StefanRijnhart:
Thanks for the clarification! This PR is now ready for review. Additional benefit of migrating the translations in the database is that you can then update the translations for each custom module using something like
```
odoo-server --stop-after-init -d migrated90 -l nl_NL --i18n-export=../../parts/repo/$module/i18n/nl.po --modules=$module
```

StefanRijnhart:
And thank you for the initial analysis on #690!

pedrobaeza (approved):

PR #1246, opened 2018-01-02
[IMP] account_voucher: Use the voucher's number before comment.

mvaled:
In a DB migration, I've found that most of the payments end-up being named 'Write-off' which is the default comment in Odoo 8.  This is because our clients use preferably the number instead of the name.  The comment is normally not changed because it's set when there are differences between credits and debits which is rare in our case.

Proposal: Use the number instead of the comment first.

--


mvaled:
Hi @hbrunn, @StefanRijnhart 

I would appreciate your opinion on this PR.  Maybe it's a too-local issue.

Happy new year!

StefanRijnhart (approved):
Agreed! We don't use vouchers that much, but I see that every voucher has the 'Write-off' default comment which is uninformative as the payment's name.

PR #1248, opened 2018-01-06
[FIX] Product level income/expense accounts

StefanRijnhart:
--


legalsylvain (commented):
Hi @StefanRijnhart. Thanks for this patch. LGTM for me.
But I don't understand why this patch is needed.
the rename should be handled by native migration (oldname=...) don't you think ? 

https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/models/product.py#L29

That's why it is written "nothing to do" in the analysis file. 
https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt#L156

Thanks for the clarification.

StefanRijnhart:
All that the oldname attribute does in Odoo is rename the database column if there is one. But the name of property fields is encoded in the ir_property model table, so properties still need manual migration. The 'nothing to do' is a false positive in the analysis.

legalsylvain (approved):
Thanks ! I didn't know that point.
:+1:

pedrobaeza (approved):
Shouldn't we propose to Odoo to handle this case with `oldname`?

StefanRijnhart:
@pedrobaeza Ha, we could propose all our migration scripts as well, see what they say ;-) I can imagine that we patch OpenUpgrade to handle oldname for properties as well if it occurs more often, but I think it is pretty rare especially because oldname is not applied consistently in upstream Odoo.

pedrobaeza:
Well, indeed, but this case is more straight-forward. Well, let's annotate it as something to improve in openupgrade analysis, isn't it?

StefanRijnhart:
Look, it's already fixed for 10.0 and up with https://github.com/oca/openupgrade/pull/681!

pedrobaeza:
Oh, I didn't remember that one! Thanks.

PR #1252, opened 2018-01-09
[11.0][FIX] remove unwanted migrations directory (2.0, 9.0.1.1)

njeudy:
Some directory prevent from running openupgrade 11.0 script in french 10.0 databases :)

--

### DIFF
--- a/addons/l10n_fr/migrations/2.0/post-migrate_tags_on_taxes.py
+++ b/addons/l10n_fr/migrations/2.0/post-migrate_tags_on_taxes.py
@@ -1,7 +1,0 @@
-from openerp.modules.registry import RegistryManager
-
-
-def migrate(cr, version):
-    registry = RegistryManager.get(cr.dbname)
-    from openerp.addons.account.models.chart_template import migrate_tags_on_taxes
-    migrate_tags_on_taxes(cr, registry)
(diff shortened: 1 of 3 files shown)


hbrunn:
I can imagine the 2.0 folder interferes, but do the 9... too?

njeudy:
@hbrunn we are migrating from 10.0 to 11.0 so no need to have all migration directory I think. 9.0.1.1 is in V9 openupgrade branch .. 

less code, clean fonctionnality ...

hbrunn:
the thing is that we regularly merge from upstream, and if ever they change something on this scripts, we'll have to do manual stuff to fix the conflict. So the risk is worth it imho if the existing scripts break something, but if it's only for cleaning up, I'd prefer to not to

njeudy:
I will try with 9.0.1.1 directory.

StefanRijnhart:
The 2.0 migration is actually part of Odoo 11.0, see https://github.com/odoo/odoo/commit/9a2b9b1ae1dc77ad9a59de585e22388dd2a5f828! As @hbrunn is right and the 9.0.x.x scripts should not intervene I'm closing this.

njeudy:
@StefanRijnhart but openupgrade script failed with 2.0 folder. How can we handle this ?

V10.0 -> V11.0 with l10n_fr fail with .. we need to handle this I think.

StefanRijnhart:
@njeudy it depends on how it fails. Is the function already covered by OpenUpgrade 11, or does it not apply to existing installations (only fresh installations of Odoo 11 before this 2.0 script was added?)?

PR #1253, opened 2018-01-09
[10.0] store fields nulls

mvaled:
After migrating a fairly small DB, I noticed that many ir_model_fields tuples (more than 800) have their 'store' column NULL.

This should update them all.

--


mvaled:
I now realized those are fields that are now missing in the model.  The only problem is the *manual* fields, but those can be easily identified.

PR #1254, opened 2018-01-10
[IMP] [FIX] hr_timesheet: Precompute department_id in analytic lines.

mvaled:
The analytic lines have a new (related) reference to the user's department.  Which for many analytic lines (I have a DB with more than 270k) may take a very long time doing in it via the ORM.

--

### DIFF
--- a/addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
+++ b/addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
@@ -11,6 +11,25 @@ COLUMN_RENAMES = {
 }
 
 
+def create_and_populate_department(cr):
+    cr.execute('''
+       ALTER TABLE account_analytic_line ADD COLUMN department_id INT;
+       ALTER TABLE account_analytic_line
+          ADD CONSTRAINT account_analytic_line_department_id_fkey
+          FOREIGN KEY (department_id) REFERENCES hr_department (id);
+
+       WITH departments AS (
+          SELECT r.user_id AS user_id, MAX(e.department_id) AS dpt_id
+             FROM hr_employee e JOIN resource_resource r
+                                ON e.resource_id = r.id
+          GROUP BY user_id
+       )
+       UPDATE account_analytic_line aal SET department_id=departments.dpt_id
+       FROM departments WHERE aal.user_id = departments.user_id;
+    ''')
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.rename_columns(env.cr, COLUMN_RENAMES)
+    create_and_populate_department(env.cr)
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I would prefer this kind of operations to use `openupgrade.logged_query` for let trace in the log.

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -11,6 +11,25 @@
 }
 
 
+def create_and_populate_department(cr):
+    cr.execute('''
+       ALTER TABLE account_analytic_line ADD COLUMN department_id INT;
+       ALTER TABLE account_analytic_line
```
pedrobaeza, line 17: Isn't constraint added automatically in the update process? It's risky to replicate ORM in this.

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -11,6 +11,25 @@
...
+          ADD CONSTRAINT account_analytic_line_department_id_fkey
+          FOREIGN KEY (department_id) REFERENCES hr_department (id);
+
+       WITH departments AS (
```
pedrobaeza, line 21: Is this subquery needed?

pedrobaeza:
@hbrunn I think this time you have been very fast merging this...

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
hbrunn: right, this was a bit premature. No manual constraints, they'll be recreated also when the column exists, cf https://github.com/OCA/OCB/blob/11.0/odoo/models.py#L2175 - @mvaled could you PR a followup for this?

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
hbrunn: I think this CTE is a very elegant solution. A subquery would be bad indeed

mvaled: Since a single user_id can be related with more than one resource, it can also have many employees, and thus belong to many departments.  This is why I even used a MAX, because I don't care for the department.  

I would have to look how does Odoo decides which department to assign using the related when it finds more that one employee.

This is probably an error, but the model allows for it.  In fact, a test in Odoo does set two employees to the same user.

mvaled:
@hbrunn, @pedrobaeza I could made another PR to address the constraint.

BTW, I'm doing another one to add the 'children_hours' which is also new in Odoo 10.

pedrobaeza:
Yes, please

PR #1255, opened 2018-01-11
[9.0][FIX] account: Fill tax_id in account.invoice.tax

pedrobaeza:
Method followed
===============

* If both codes are present in only one tax, then we can safely fill tax_id with that tax.
* If there are more than one tax with that tax codes, try to match tax with the name of the tax line.
* If still no single match, try to match only with active taxes.
* Finally, if there is no match, see if there's any tax code manual mapping for starting again the matching.

@Tecnativa


rafaelbn (approved):

StefanRijnhart (approved):
This clearly mimics the way that move taxes are retrieved, and as such does not work for some localizations such as l10n_nl. I'll merge this for reasons of consistency. Adaptation for l10n_nl and other localizations can be considered part of https://github.com/OCA/OpenUpgrade/pull/1172

pedrobaeza:
Thanks for merging this, though this is slightly different from the other one. 

* It performs more with SQL for performance.
* It includes a way to manually map some tax codes if no direct match.
* I think cases you told about should work right now.

Any way, tell me whatever you need.

PR #1257, opened 2018-01-11
[IMP][FIX] hr_timesheet: Pre-compute news fields for sub-tasks.

mvaled:
Odoo 10 adds a tree of sub-tasks (`parent_id` and `child_ids`) and two new computed fields `children_hours` and `total_hours_spent` over that structure.

The first one is defined recursively over the child_ids.  At the leaves of the tree children_hours is 0.  Since all tasks in the migration process would end-up being leaves of this structure, children_hours would be 0.

The field `total_hours_spent` is simply defined by `effective_hours + children_hours`.

The other fields in this computed method are kept invariant.

--

### DIFF
--- a/addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
+++ b/addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
@@ -26,7 +26,40 @@ def create_and_populate_department(cr):
     ''')
 
 
+def create_and_populate_children(env):
+    cr = env.cr
+    openupgrade.add_fields(env, [
+        _task_field('parent_id', 'integer', 'INT'),
+        _task_field('children_hours', 'float', 'DOUBLE PRECISION'),
+        _task_field('total_hours_spent', 'float', 'DOUBLE PRECISION'),
+    ])
+    #  Since the parent_id is new there's no point in computing the
+    #  children_hours, they will be 0; however there's a single method
+    #  computing children_hours, effective_hours, remaining_hours,
+    #  total_hours, total_hours_spent, and delay_hours; but only
+    #  children_hours and total_hours_spent need recomputation.
+    #
+    #  In _get_hours, ``task.total_hours_spent = task.effective_hours + task.children_hours``
+    #  No need to bother to add 0.
+    openupgrade.logged_query(cr, '''
+    UPDATE project_task SET children_hours=0, total_hours_spent=effective_hours;
+    ''')
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.rename_columns(env.cr, COLUMN_RENAMES)
     create_and_populate_department(env.cr)
+    create_and_populate_children(env)
+
+
+def _task_field(field_name, field_type, sql_type, module='project'):
+    # See `create_and_populate_children` above.
+    return (
+        field_name,
+        'project.task',
+        'project_task',
+        field_type,
+        sql_type,
+        module
+    )
--- a/addons/hr_timesheet/migrations/10.0.1.0/openupgrade_analysis.txt
+++ b/addons/hr_timesheet/migrations/10.0.1.0/openupgrade_analysis.txt
@@ -8,6 +8,8 @@ hr_timesheet / project.project          / allow_timesheets (boolean)    : NEW
 hr_timesheet / project.project          / subtask_project_id (many2one) : NEW relation: project.project
 hr_timesheet / project.task             / parent_id (many2one)          : NEW relation: project.task
 hr_timesheet / project.task             / timesheet_ids (one2many)      : NEW relation: account.analytic.line
+hr_timesheet / project.task             / children_hours (float)        : NEW computed field
+hr_timesheet / project.task             / total_hours_spent (float)     : NEW computed field
 ---XML records in module 'hr_timesheet'---
 NEW ir.actions.act_window: hr_timesheet.act_hr_timesheet_line
 NEW ir.actions.act_window: hr_timesheet.act_hr_timesheet_line_by_project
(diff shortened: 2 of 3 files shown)


mvaled:
Please consider this still WIP.  I'm testing it with a DB with 50k tasks.

mvaled:
@hbrunn, @pedrobaeza 

I think this ready.

mvaled:
@pedrobaeza I've just noticed your comment about using `openupgrade.logged_query`.  Doing change now.

mvaled:
Done here and in #1258

mvaled:
@pedrobaeza, @hbrunn 

I think this PR is ready.

StefanRijnhart (approved):

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -29,7 +29,32 @@ def create_and_populate_department(cr):
     ''')
 
 
+def create_and_populate_children(cr):
+    openupgrade.logged_query(cr, '''
+    ALTER TABLE project_task ADD COLUMN parent_id INT;
+
+    -- Since the parent_id is new there's no point in computing the
```
pedrobaeza: Why this explanation is part of the logged query?

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -29,7 +29,32 @@ def create_and_populate_department(cr):
...
+    -- children_hours, they will be 0; however there's a single method
+    -- computing children_hours, effective_hours, remaining_hours,
+    -- total_hours, total_hours_spent, and delay_hours.  However, only
+    -- children_hours and total_hours_spent need recomputation.
+
+    -- These are new columns, the others existed before but created in another
+    --  addon.
+
+    ALTER TABLE project_task ADD COLUMN children_hours DOUBLE PRECISION;
```
pedrobaeza: As stated by @hbrunn in https://github.com/OCA/openupgradelib/pull/93#issue-162200382, these fields won't have XML-ID, so we should change approach of pre-creating columns having an openupgradelib method that makes the rest of the work (create the column, create the field and assign an XML-ID).

pedrobaeza (changes requested):

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
mvaled: I could make it outside the query, but I think the comment belongs near to the "funny" code.   I like to leave my thoughts as clear as I deem possible.  In this case, the reason behind why I can put `children_hours=0` is not quite obvious.

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
mvaled: I didn't realize that when I made the PR.   I guess I should have used the `add_ir_model_fields` function.  Am I right?

mvaled: Now that I read better is not `add_ir_model_fields`.

pedrobaeza:
@mvaled can you please attend pending comments?

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
pedrobaeza: @mvaled I have created the method in this PR: https://github.com/OCA/openupgradelib/pull/100

mvaled: Hi @pedrobaeza   

Thanks for the new method. I made a rebase on 10.0 and squash all the commits into one.

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -26,7 +26,39 @@ def create_and_populate_department(cr):
...
+    openupgrade.logged_query(cr, '''
+    UPDATE project_task SET children_hours=0, total_hours_spent=effective_hours;
+    ''')
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.rename_columns(env.cr, COLUMN_RENAMES)
     create_and_populate_department(env.cr)
+    create_and_populate_children(env.cr)
```
MiquelRForgeFlow: pass the env instead

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -26,7 +26,39 @@ def create_and_populate_department(cr):
     ''')
 
 
+def create_and_populate_children(cr):
+    openupgrade.add_fields([
```
MiquelRForgeFlow: `openupgrade.add_fields(env, [`

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
```diff
@@ -26,7 +26,39 @@ def create_and_populate_department(cr):
...
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.rename_columns(env.cr, COLUMN_RENAMES)
     create_and_populate_department(env.cr)
+    create_and_populate_children(env.cr)
+
+
+def _task_field(field_name, field_type, sql_type, module='project'):
```
MiquelRForgeFlow: should this method be above the `migrate` method by convention? 🙂

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
mvaled: Opps!  Fixed in the last push.

addons/hr_timesheet/migrations/10.0.1.0/pre-migration.py
mvaled: I usually organize my modules in order of "interest-ness": definitions that are interesting and/or important come first.  This function just put repeated things like the model, table and module.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #1258, opened 2018-01-11
[10.0][IMP] purchase: Speed up copying of 'state' field in purchase lines.

mvaled:
This is just a speed improvement.  For me it also avoids an error, because we added another possible choice in the 'state' field in an addon.  Without this I get a `ValueError: Wrong value for purchase.order.line.state: 'paid'`.

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #1260, opened 2018-01-16
Revert "[10.0][IMP] purchase: Speed up copying of 'state' field in purchase lines."

pedrobaeza:
Reverts OCA/OpenUpgrade#1258

I'm reverting this commit because it's incorrect and provokes an error (I don't know why Travis was not failing when this was merged).

On v9, there's exactly the same field that in v10:

v9: 

https://github.com/odoo/odoo/blob/11c73cd0aa2ff9cbfde8e5eebf3f22dfa1c805f1/addons/purchase/purchase.py#L558

v10:

https://github.com/odoo/odoo/blob/ef8cd11a6be25cbb5288e8ab88b4aafa2fa0119c/addons/purchase/models/purchase.py#L632

The error comes because the column already exists:

```
ProgrammingError: column "state" of relation "purchase_order_line" already exists
```

@mvaled you have to seriously check customizations in your DB or last update because your changes make standard to fail.


hbrunn:
I'd like to know why this didn't break in travis. Possibly we'll have to regenerate the 9.0 test database, this was merged quite late: https://github.com/OCA/OCB/commit/dbe3d8035f6f48b630f043707ffc6bac0b1cb204

pedrobaeza:
Yeah, that can be the problem indeed.

mvaled:
Hi all,

Indeed you're right.  It seems that it has to do with a customized 'state' in one of our addons.

I'm doing (for the first time) a double migration: from Odoo 8 to 9 and then to 10.

The DB I get after migrating from 8 to 9 does not have the 'state' column:

```
$ psql -d o9-mercurio -c "select state from purchase_order_line"
ERROR:  column "state" does not exist
LINE 1: select state from purchase_order_line
```

I was prompted to do the original patch because our customization introduced in a new state for the PO, which while migrating from 9 to 10 (after migrating away from 8) provoked an error.

That's why I assumed the column was new in 10.  In any case, I assume then that original patch can be applied to the 9.0 branch.  I will check both issues:

First, why do I end up with a missing 'state' column?  Have any of you witnessed anything similar?

Strange thing: after the OpenUpgrade migration from 8 to 9, I do a `-u all` with Odoo code base, and that field exists:

https://github.com/merchise-autrement/odoo/blob/c699d3d51af27a9eaff205f8b29f1d3b44dac549/addons/purchase/purchase.py#L566

... and still the column does not get added to the DB. :O


Second, should we apply the original patch to branch 9.0?


Thanks for your patience.

pedrobaeza:
I don't think it's needed in v9 as well.

PR #1263, opened 2018-01-18
[WIP] [FIX] account: Reconcile items that appear in reconciliation objects.

### LINKED ISSUE #1262: [9.0] Reconciled items are being left un-reconciled after the migration.
### Summary of the issue

After migrating from Odoo 8 to 9, automatic gain/loss journal items in the debtor/creditor accounts are not reconciled, even they were in the Odoo 8 DB.

Impacted versions:  Migration from 8 to 9.

### Steps to reproduce

This is rather lengthy process and there may be some subtleties involved in the issue.  In steps where there's another (obvious) choice to do things I explain further.

1. Create a new DB (no demonstration data).

2. Install the Accounting and Finance module.  Make your company currency USD, and taxes 0% for both sales and purchases (this is not unrealistic: Cuba).  Use a general chart of accounts.

3. Activate multi-currency.

4. Modify your Bank account and diary to have currency EUR, also change your Debtor account to have secondary currency EUR.  Basically this means that you report in USD but keep your assets in EUR.

5. Create accounts for Foreign Exchange Gains and Loss and put them in the configuration.

7. Go to the Currencies configuration, chose USD and add new (for the current date) rate of 1.000.  Important: keep the old rate of 1.2834.  Notice that even the company reports in USD, we don't set its rate to 1, but keep the EUR with rate 1 and update the USD/EUR daily.

8. Create a customer invoice with a single line.  Price 100 €.  Use a date several days back so that the rate be 1.2834.  Validate the invoice.  This would make the journal entry to look like:

screenshot from 2018-01-18 10-58-47

9. Pay the invoice entirely (`account.voucher`).  Use the Bank diary, 100 €, date as of today (USD/EUR should be 1).  Validate the payment.   The journal entry should be like:

screenshot from 2018-01-18 11-01-14

Notice that the 28.34 USD in the Debtors account is reconciled.  

10. Post all entries.

11.  Stop there and migrate the DB to Odoo 9 with OpenUpgrade (I used branch 9.0 at 2f1674913392be418d5ddd3e632cf4ec3e7a62a5).

12.  Upgrade the migrated DB with Odoo 9 (I used branch 9.0 at odoo/odoo@40998a4dedcfa1aef85c1e94fd0648a51b8f56f2).

##### Current behavior

The migrated DB shows item crediting Debtors by 28.34 unreconciled:

screenshot from 2018-01-18 11-07-28

##### Expected behavior:

The journal item should be reconciled.

##### Notes

Even though the item is not 'reconciled' it does appear in the A1 reconciliation object:

screenshot from 2018-01-18 11-08-16

mvaled:
Fixes #1262.

--

### DIFF
--- a/addons/account/migrations/9.0.1.1/end-migration.py
+++ b/addons/account/migrations/9.0.1.1/end-migration.py
@@ -22,6 +22,17 @@ def update_journal_payment_methods(env):
     })
 
 
+def mark_reconciled(env):
+    '''Update move lines fully reconciled.'''
+    openupgrade.logged_query(
+        env.cr,
+        '''UPDATE account_move_line SET reconciled=true
+           WHERE full_reconcile_id IS NOT NULL AND reconciled=false
+        '''
+    )
+
+
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     update_journal_payment_methods(env)
+    mark_reconciled(env)


mvaled:
Hi @hbrunn, @pedrobaeza, @NL66278, 

I would like you to review this PR and the issue it attempts to solve (#1262).  Please consider this a WIP.  I've just did the first thing that came to mind.  Maybe the best place would be in the 'account_full_reconcile' module...

mvaled:
This solves the reported issue:

screenshot from 2018-01-18 12-31-25

mvaled:
In https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account_full_reconcile/migrations/9.0.1.1/post-migration.py#L160 the `reconciled` column is not always True despite the value of `full_reconcile_id`.

So probably the issue is not with the `reconciled` itself but about the residuals.

mvaled:
In the DB I see this:

```
to9-acc-rec=# select debit, credit, amount_residual, amount_residual_currency, reconciled, full_reconcile_id, currency_id, amount_currency from account_move_line;
 debit  | credit | amount_residual | amount_residual_currency | reconciled | full_reconcile_id | currency_id | amount_currency 
--------+--------+-----------------+--------------------------+------------+-------------------+-------------+-----------------
   0.00 | 128.34 |             0.0 |                      0.0 | f          |                   |           1 |         -100.00
  28.34 |   0.00 |             0.0 |                      0.0 | f          |                   |           1 |            0.00
 100.00 |   0.00 |             0.0 |                      0.0 | f          |                   |           1 |          100.00
   0.00 |  28.34 |             0.0 |       22.080000000000002 | t          |                 1 |           1 |            0.00
   0.00 | 100.00 |             0.0 |      -22.080000000000002 | t          |                 1 |           1 |         -100.00
 128.34 |   0.00 |             0.0 |                      0.0 | t          |                 1 |           1 |          100.00
```

Note: the items are reconciled because of this PR, but notice the `amount_residual_currency`.  That's why the items are [not being marked](https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account_full_reconcile/migrations/9.0.1.1/post-migration.py#L173) as reconciled.

mvaled:
Doing the same steps in Odoo 9 (no migration) yields:

```
to9-acc1=# select debit, credit, amount_residual, amount_residual_currency, reconciled, full_reconcile_id, currency_id, amount_currency from account_move_line;
 debit  | credit | amount_residual | amount_residual_currency | reconciled | full_reconcile_id | currency_id | amount_currency 
--------+--------+-----------------+--------------------------+------------+-------------------+-------------+-----------------
  100.0 |    0.0 |             0.0 |                      0.0 | f          |                   |           1 |           100.0
    0.0 | 128.34 |             0.0 |                      0.0 | f          |                   |           1 |          -100.0
  28.34 |    0.0 |             0.0 |                      0.0 | f          |                   |           1 |             0.0
 128.34 |    0.0 |             0.0 |                      0.0 | t          |                 2 |           1 |           100.0
    0.0 |  100.0 |             0.0 |                      0.0 | t          |                 2 |           1 |          -100.0
    0.0 |  28.34 |             0.0 |                      0.0 | t          |                 2 |           1 |             0.0
```

So, indeed the problem is with the `amount_residual_currency` and, thus, the solution should deal with that problem.

Please don't merge this yet.

However, I would like some feedback and/or ideas from @pedrobaeza, @NL66278, @hbrunn  and @jbeficent. Since they were involved in the discussion of #663.

mvaled:
I conducted a *guided* review of a migrated DB to watch the hot spots.  The following SQL (or modifications of it) was instrumental:

```sql
--- Shows move lines within the same full reconcilation group for which the sum of amount_currency is not the same as amount_residual_currency.

WITH recs AS (
    SELECT id, debit, credit, amount_residual,
           amount_residual_currency,
           reconciled, full_reconcile_id,
           currency_id, amount_currency,
           (ROUND(SUM(debit) OVER (PARTITION BY full_reconcile_id), 4) - ROUND(SUM(credit) OVER (PARTITION BY full_reconcile_id), 4)) AS balance,
           ROUND(SUM(amount_residual_currency) OVER (PARTITION BY full_reconcile_id), 2) AS diff,
           ROUND(SUM(amount_currency) OVER (PARTITION BY full_reconcile_id), 2) AS cdiff
    FROM account_move_line
    WHERE full_reconcile_id IS NOT NULL
) SELECT *, ABS(diff-cdiff) as deviation FROM recs WHERE diff != cdiff ORDER BY full_reconcile_id;

```


In this DB, only 1011 move line showed a deviation between 0.01 and 0.03.  The remaining 174366 move lines showed no difference.

Other results:

- `balance` is always 0, meaning that debits match credits.  
- `cdiff` is always 0, meaning that the amount in currency units is always being completely "consumed".

I think all the deviations are due to rounding errors. Even the deviations larger than 0.01 can be explained by that: reconciliations with 10 or more lines.

In this DB, I think I could safely set `reconciled` to true whenever the line has a full reconcile id.  So the solution proposed "works".  Alternatively we could try to improve the how we compute the `amount_residual_currency`.

What do you think?

mvaled:
Any comments?

StefanRijnhart:
Thanks for your work! Are the amounts in fact unrounded in your database, or are the rounding errors introduced in the migration of the account module?

mvaled:
I believe they were introduced in the migration.  In the original DB the following queries return 0 rows.

```
select debit, trunc(debit), debit-trunc(debit) from account_move_line where 0 <  debit-trunc(debit)  and  debit-trunc(debit)  < 0.01;

select debit, trunc(debit), debit-trunc(debit) from account_move_line where 0 <  debit-trunc(debit)  and  debit-trunc(debit)  < 0.01;

select abs(amount_currency), trunc(abs(amount_currency)), abs(amount_currency)-trunc(abs(amount_currency)) from account_move_line where 0 <  abs(amount_currency)-trunc(abs(amount_currency))  and  abs(amount_currency)-trunc(abs(amount_currency))  < 0.01;
```

I've migrated the original DB from 8 to 10 already with this PR merged.  I discovered some inconsistencies with amount_residual for client refunds when exchanges rates were involved.  But only a very small fraction.

mvaled:
In the migrated DB I see unrounded amounts in credits and amount_currency, in newly created lines (non migrated).

mvaled:
Ah! @StefanRijnhart 

Notice the error is reproducible in a test DB as explained #1262.

StefanRijnhart:
@mvaled Did you create a fresh database with the instructions of #1262 so that it can be used to reproduce the bug? Can you shared it?

mvaled:
@StefanRijnhart I did when I reported the issue.  I will recreate it and upload the SQL dumps.

mvaled:
I have just reproduced the bug.  

[bug_1262.zip](https://github.com/OCA/OpenUpgrade/files/1759441/bug_1262.zip)


The file 'odoo8_bug_1262.sql' contains the Odoo 8 DB before migration.  I prepared the DB in a virtualenv with Odoo 8 at commit odoo/odoo@515a5ed900194050930d27d06f237a7c2e627354.

I cloned that DB and did the following in OpenUpgrade (commit 2f1674913392be418d5ddd3e632cf4ec3e7a62a5):

```
./openerp-server -d odoo9_bug_1262 -u all --stop-after-init 2>&1 | tee odoo9_bug_1262.txt
```

The file odoo9_bug_1262.sql is the dump just after the migration.

In order to actually see the DB with Odoo 9 (commit odoo/odoo@2014fabfbaf2e24a047881deb404844a81c7bbc1), I have to manually remove the views for im_chat and edi:

```
psql -d odoo9_bug_1262 -c "
         DELETE FROM ir_ui_view WHERE arch like '%/im_chat/static/%';
         DELETE FROM ir_ui_view WHERE arch like '%/edi/static/%';"
"
```

And then upgrade the DB using Odoo 9.

StefanRijnhart:
Thanks, it's perfect! I've restored the database locally and I'm having a look at the migration of the move lines and the reconciliations.

StefanRijnhart:
Fixed with https://github.com/OCA/OpenUpgrade/pull/1277

Issue #1240, opened 2017-12-20
Problem with table account_statement_operation_template in migration 9.0


hmartinet:
Impact migration 8.0 to 9.0, module account.

This migration try to rename table account_statement_operation_template into account_operation_template. I have an INFO in log and no other logs about this table. There is no ERROR in the migration process.

In the result database the table account_statement_operation_template already exists but not the table account_operation_template. Then the next migration to 10.0 will failed because account_operation_template does not exists. I don't understand why this table can be not renamed without error in logs. I try this migration 3 times with the same result. Someone has an idea in which direction I have to investigate ?

```2017-12-20 13:48:34,611 53675 INFO odoo_8_migrated OpenUpgrade: table account_statement_operation_template: renaming to account_operation_template```

**Comment by pedrobaeza** [comment]
Sorry, no idea. I haven't got any problem with that.

**Comment by hmartinet** [comment]
I will try to rename it manually and then running migration 9.0 to 10.0. If an update process specialist has an idea how a query can not be applied without an ERROR in log.

**Comment by StefanRijnhart** [comment]
What is the error from the 10.0 migration?

**Comment by hmartinet** [comment]
The error (table account_operation_template does not exists) appears in 10.0 migration. But the cause is that the table rename in 9.0 migration was not effective (but correctly logged). The the table name is always account_statement_operation_template after the 9.0 migration (like in 8.0).

**Comment by StefanRijnhart** [comment]
The reason I asked is that I don't understand the problem, so I was hoping that a log snipped of a failed migration (or the whole set of logs maybe) could provide more insight. I am currently finishing a migration from 7.0 to 10.0 and I do not have this problem but then again I have no entries in either table which might make a difference.

BTW if the problem is that table account_statement_operation_template remains unchanged after the rename_tables operation it should actually go unnoticed because an upgrade of account 9.0 should create the new table anyway.

**Comment by hmartinet** [comment]
Ok. I have this problem only with one customer database, another customer was in same version of Odoo 7.0 with same version of custom addons and the database was successfully migrated. Then I think the problem is in the database but I have no idea where to investigate.

The error logged in 10.0 migration is:
```
2017-12-20 10:07:28,400 30257 ERROR odoo_9_migrated OpenUpgrade: relation "account_operation_template" does not exist
Traceback (most recent call last):
  File "[...]/workdir/openupgradelib/openupgradelib/openupgrade.py", line 1391, in wrapped_function
    if use_env2 else cr, version)
[...]

PR #1269, opened 2018-02-06
[8.0] Improve reconcile_ref creation.

hmartinet:
Create and set values for reconcile_ref field of account_move_line
to bypass ORM auto update (memory usage and speed improvement for
large account_move_line).

#1240

--

### DIFF
--- a/addons/account/migrations/8.0.1.1/pre-migration.py
+++ b/addons/account/migrations/8.0.1.1/pre-migration.py
@@ -60,7 +60,6 @@ def migrate(cr, version):
     openupgrade.rename_columns(cr, column_renames)
     openupgrade.rename_tables(cr, tables_renames)
     pre_compute_reconcile_ref_field(cr)
-
     # drop views that inhibit changing field types. They will be recreated
     # anyways
     for view in [
--- a/addons/account/migrations/8.0.1.1/pre-migration.py
+++ b/addons/account/migrations/8.0.1.1/pre-migration.py
@@ -59,7 +59,6 @@ def migrate(cr, version):
             cr, 'account', 'exp', 'account.analytic.journal', res[0], True)
     openupgrade.rename_columns(cr, column_renames)
     openupgrade.rename_tables(cr, tables_renames)
-
     pre_compute_reconcile_ref_field(cr)
 
     # drop views that inhibit changing field types. They will be recreated
--- a/addons/account/migrations/8.0.1.1/pre-migration.py
+++ b/addons/account/migrations/8.0.1.1/pre-migration.py
@@ -22,23 +22,12 @@ tables_renames = [
 ]
 
 
-@openupgrade.migrate()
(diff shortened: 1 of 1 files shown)


hmartinet:
@StefanRijnhart This will solve my problem with 1'000'000 records in account_move_line. The migration from 7.0 to 8.0 was successful and complete on a 8Go RAM instance. I have to write similar statements for 9.0 and 10.0 migration (I have same problem in 9.0, Now I use a monitoring tool to check memory usage graph during the process).

legalsylvain (approved):
hi Thanks a lot.
It seems replace correctly the computed function.
https://github.com/OCA/OCB/blob/8.0/addons/account/account_move_line.py#L445

maybe one minor approvment : switch between the two updates

LGTM otherwise. 
```
    cr.execute("""
        update account_move_line ml
            set reconcile_ref = r.name
        from account_move_reconcile r
        where ml.reconcile_id = r.id;
        """)


    cr.execute("""
        update account_move_line ml
            set reconcile_ref = r.name
        from account_move_reconcile r
        where ml.reconcile_partial_id = r.id
        and ml.reconcile_ref is null
        """)
```

pedrobaeza (changes requested):
A little refactoring for better reading: please enclose all the code in a method that is called from the main method (`migrate`), and the initial comment can be method doc.

hmartinet:
@legalsylvain I think this order is the right one. A line should never be partially AND fully reconciled (it would be a data inconsistency but it's not the point here to fix that). But if that is still the case we want to keep the full reconcile name, then we update the full after the partial to guarantee that.

@pedrobaeza Ok I will fix it

legalsylvain:
Hi @hmartinet. I wrote **and ml.reconcile_ref is null** in the second statement, so IMO the two proposals make the same things and I guess that the second update will be executed faster. But not a big deal.
regards.

addons/account/migrations/8.0.1.1/pre-migration.py
```diff
@@ -35,6 +59,9 @@ def migrate(cr, version):
             cr, 'account', 'exp', 'account.analytic.journal', res[0], True)
     openupgrade.rename_columns(cr, column_renames)
     openupgrade.rename_tables(cr, tables_renames)
+
```
pedrobaeza: Please don't add these extra lines. PEP8 advises not to abuse of these lines and in this case, it doesn't add any extra value.

hmartinet:
@legalsylvain Ok I understand your point of view. But I still think that my proposal is more logical (functionally). And yes it's a philosophical choice that does not really matter.

pedrobaeza (approved):

pedrobaeza:
Thank you for your contribution.

StefanRijnhart:
I've come late to the party, but thanks for following this up!

hmartinet:
@StefanRijnhart :wink:

PR #1277, opened 2018-02-27
[FIX] Computation of residual currency amount when reconciling moves …

### LINKED ISSUE #1262: [9.0] Reconciled items are being left un-reconciled after the migration.

StefanRijnhart:
…in the same currency

Don't apply the rate computation between outstanding amount and currency amount when
the currencies of both reconciled lines are the same, but derive the residual currency
amount directly from the two currency amounts of both lines. Analogous to
https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_move.py#L857-L861

Fixes #1262

### DIFF
--- a/addons/account_full_reconcile/migrations/9.0.1.1/post-migration.py
+++ b/addons/account_full_reconcile/migrations/9.0.1.1/post-migration.py
@@ -92,10 +92,15 @@ def migrate_reconcile(cr):
         currency_id = False
         rate = 1.0
         amount_currency = 0.0
-        if debit_record.line_currency_id and debit_record.amount_currency:
+        if debit_record.line_currency_id:
             currency_id = debit_record.line_currency_id
-            rate = abs(debit_record.balance / debit_record.amount_currency)
-            amount_currency = amount / rate
+            if debit_record.line_currency_id == credit_record.line_currency_id:
+                amount_currency = min(
+                    abs(debit_record.amount_residual_currency),
+                    abs(credit_record.amount_residual_currency))
+            elif debit_record.amount_currency:
+                rate = abs(debit_record.balance / debit_record.amount_currency)
+                amount_currency = amount / rate
         elif credit_record.line_currency_id and credit_record.amount_currency:
             currency_id = credit_record.line_currency_id
             rate = abs(credit_record.balance / credit_record.amount_currency)


StefanRijnhart:
@mvaled with this, I got

```
odoo1263=> select date,debit,credit,account_id,amount_currency, amount_residual, amount_residual_currency, reconciled from account_move_line order by account_id
;
    date    | debit  | credit | account_id | amount_currency | amount_residual | amount_residual_currency | reconciled 
------------+--------+--------+------------+-----------------+-----------------+--------------------------+------------
 2018-02-08 | 128.34 |   0.00 |          8 |          100.00 |             0.0 |                      0.0 | t
 2018-02-26 |   0.00 | 100.00 |          8 |         -100.00 |             0.0 |                      0.0 | t
 2018-02-26 |   0.00 |  28.34 |          8 |            0.00 |             0.0 |                      0.0 | t
 2018-02-08 |   0.00 | 128.34 |         19 |         -100.00 |             0.0 |                      0.0 | f
 2018-02-26 | 100.00 |   0.00 |         26 |          100.00 |             0.0 |                      0.0 | f
 2018-02-26 |  28.34 |   0.00 |         28 |            0.00 |             0.0 |                      0.0 | f
```

StefanRijnhart:
-- edit -- Forget the earlier version of this comment, I did not understand how the 9.0 database in your zip was created. For the record, here is the table of partial reconciliations in the migrated database. Does it look alright to you?
```
select credit_move_id,debit_move_id,amount,amount_currency,full_reconcile_id from account_partial_reconcile;
 credit_move_id | debit_move_id | amount | amount_currency | full_reconcile_id 
----------------+---------------+--------+-----------------+-------------------
              4 |             2 |  100.0 |           100.0 |                 1
              5 |             2 |  28.34 |             0.0 |                 1
(2 rows)
```

mvaled:
In Odoo 9, those reconciliations seem OK.  If I understand the model, matched_credit_ids (and matched_debit_ids) are actually a kind of relation between the move lines themselves, but with attributes attached.

The journal entry of the invoice was 100 € valuated as $ 128.34.  But the payment entered 100 € valuated at $ 100, so the 0€ / $ 28.34 line is just the adjustment needed to report the gain/loss due to change in the rates.  (~~I believe in Odoo 10, they went further I put the gain/loss in a separate move~~ They do that in Odoo 9 as well).

In fact, I've just created this DB in Odoo 9 where I played the same setup as in the Odoo 8.  You can see that the partials produced in Odoo 9 are just the same (except for ids) as the ones produced from the migration:

[o9.sql.zip](https://github.com/OCA/OpenUpgrade/files/1765146/o9.sql.zip)

mvaled:
It seems that you need to also do the same for credits (refunds).  I have detected another possibly related bug.  In all cases the invoices/refunds are in EUR, but the company reports in USD. Summary of the bug: Some refunds that are not-paid after migration but they were before. Some invoices for which I have partial payments got the wrong residual after migration.

Disclaimer: Today I'm not with a full Internet connection, so I won't be able to be online.  I've checkout your branch and testing with the case in #1262 and also the other two I have sketched before.

StefanRijnhart:
This code is run for every full or partial reconciliation of entries which are in the same non-company currency, no matter if it is a refund or something else so I am hoping this fixes all your problems. Please keep us updated as to what your results are after doing a full migration.

mvaled:
I've already done the migration from Odoo 8 to 9 and then to 10.   It's been a long process.  These issues, I've managed to fix them in the migrated DB.  But in any case it's good to have them fixed in the code. ;)

I will try to report back as soon as I'm able.

StefanRijnhart:
Sure, take your time!

pedrobaeza:
@mvaled I'm facing the issue with currencies on an already migrated database (there are invoices in other currency that appears as pending debits when they were paid). Can you share with me the queries you made for fixing the problem in your migrated DB?

mvaled:
Hi @pedrobaeza,

I used a rather manual approach and I did it after the Odoo 10 migration.  I worked in the Odoo shell: `odoo shell -d DB` instead of SQL directly:

```python
>>> Move = self.env['account.move']
>>> Invoice = self.env['account.invoice']
>>> type(Move)._check_lock_date  = lambda s: True  # ensure to be able to write locked moves
```

Find all invoices in a foreign currency created before the migration which are still open and find those that have partials:

```python
>>> invoices = Invoice.search([('currency_id', '!=', 3), ('state', '=', 'open'), ('create_date', '<=', '2018-01-27')])
>>> partials = invoices.filtered(lambda i: i.residual and i.move_id.line_ids.filtered(lambda l: l.account_id.reconcile).filtered(lambda l: l.matched_debit_ids.full_reconcile_id or l.matched_credit_ids.full_reconcile_id))
```

Since the `partials` were less than 50, I decided to inspect each of it and I didn't create an automated script.  These are (some of) the things I checked:

```python
>>> invoice = partials[0]
>>> # Select the move that contains the Receivable.
>>> line = invoice.move_id.line_ids.filtered(lambda l: l.account_id.reconcile)
```

NOTE: if the invoice was a refund or not, I had to check for matched_debits or credits.

I checked the invoice was fully consumed in its currency:

```python
>>> sum(line.matched_debit_ids.full_reconcile_id.partial_reconcile_ids.mapped('amount_currency'))  == invoice.amount_total
```


The exchange gain/loss adjuments could be positive or negative, I check for them by going to the payment move in the UI.

Once I was statisfied this invoice was actually paid, I did:

```python
>>> line.amount_residual_currency = 0
>>> line.amount_residual = 0
>>> invoice.state = 'paid'  # This triggers recomputations or residual.

>>> env.cr.commit()
```

I'm sorry I can't be more helpful.

pedrobaeza:
Hi @mvaled thank you very much for your instructions. Digging more in detail in my DB, the problem was due to a user misuse, so it's not directly a migration problem, so sorry for making you explain this, but it can be useful for anyone more.

But right now I'm not sure which cases this PR solves, as I haven't faced it.

mvaled (approved):
I tested this today.  The issue explained in #1262 is solved.

StefanRijnhart:
Thanks for testing @mvaled!

PR #1280, opened 2018-03-02
[9.0][MIG] portal_sale migration scripts

cubells:
- Add noupdate xml records.

cc @Tecnativa


pedrobaeza (changes requested):
Please use the recent method for resetting the mail template translations, or they can fail too. See as example https://github.com/OCA/OpenUpgrade/pull/1247

pedrobaeza:
@cubells please attend comments

cubells:
done!

pedrobaeza (approved):

Issue #1283, opened 2018-03-07
[9.0] account_voucher. Link from invoice to payment not created.


AaronHForgeFlow:
The field "Originator payment" in account move lines is intended to be used as the link that relates the move lines to the payment that originated them, not the the invoice that was being paid the payment_id. https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_payment.py#L511.

However,  the migration script is trying to match the account move lines generated from the invoices and the ones generated from the payment. 
https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account_voucher/migrations/9.0.1.0/post-migration.py#L86

As they are not related, the table `account_invoice_payment_rel` remains empty after the migration.

pedrobaeza:
Do you have a patch for this?

AaronHForgeFlow:
@pedrobaeza I created https://github.com/OCA/OpenUpgrade/pull/1284 but still WIP

pedrobaeza:
Closing as there's an opened PR

PR #1284, opened 2018-03-07
[9.0][FIX] account_voucher. Link between payment and invoice

AaronHForgeFlow:
Solves this issue: https://github.com/OCA/OpenUpgrade/issues/1283

### DIFF
--- a/addons/account_voucher/migrations/9.0.1.0/post-migration.py
+++ b/addons/account_voucher/migrations/9.0.1.0/post-migration.py
@@ -20,7 +20,7 @@ def create_payments_from_vouchers(env):
     # As far as I can see there is no connection between invoices in 8.0
     # that we can or have to migrate to 9.0.
     env.cr.execute(
-        """\
+        """
         INSERT INTO account_payment (
             id, create_date, communication, company_id, currency_id,
             partner_id, partner_type,
@@ -68,7 +68,7 @@ def create_payments_from_vouchers(env):
     )
     # Statement below works because new payments have same id as old vouchers
     env.cr.execute(
-        """\
+        """
         UPDATE account_move_line aml2
         SET payment_id = av.id
         FROM account_voucher av
@@ -85,13 +85,19 @@ def create_payments_from_vouchers(env):
     )
     # Also recreate link from invoice to payment
     env.cr.execute(
-        """\
+        """
         INSERT INTO account_invoice_payment_rel (payment_id, invoice_id)
-        SELECT aml.payment_id, ai.id
-        FROM account_move_line aml
-        JOIN account_invoice ai ON aml.move_id = ai.move_id
-        WHERE NOT aml.payment_id IS NULL
-        GROUP BY aml.payment_id, ai.id
+        SELECT av.id as payment_id, ai.id as invoice_id
+        FROM account_invoice ai
+        JOIN account_move_line amli ON ai.move_id = amli.move_id
+        JOIN account_partial_reconcile apr ON
+             (apr.credit_move_id = amli.id OR apr.debit_move_id = amli.id)
+        JOIN account_move_line amlp ON
+             (apr.credit_move_id = amlp.id OR apr.debit_move_id = amlp.id)
+        JOIN account_voucher av ON av.move_id = amlp.move_id
+        WHERE av.voucher_type IN ('receipt', 'payment')
+        AND av.state IN ('draft', 'posted')
+        GROUP BY av.id, ai.id
         """
     )
 


pedrobaeza:
Next time, please publish directly the PR with the issue text inside, instead of making 2 things that are worst to be followed. About this PR, please check Travis.

AaronHForgeFlow:
This is working fine for me, but creating an extra column maybe is not the best way to do it. @NL66278 @jbeficent

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -83,15 +83,56 @@ def create_payments_from_vouchers(env):
         AND aml.id = aml2.id
         """
     )
+    env.cr.execute("""SELECT column_name
```
pedrobaeza: Use `openupgradelib.column_exists`

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -83,15 +83,56 @@ def create_payments_from_vouchers(env):
...
+        column_name='payment_from_invoice_id'""")
+    if not env.cr.fetchone():
+        env.cr.execute("""
+            ALTER TABLE account_move_line ADD COLUMN
+            payment_from_invoice_id INTEGER
+            """)
+
+    env.cr.execute(
+        """\
+        WITH Q1 AS (
```
pedrobaeza: Do you really need the subquery?

AaronHForgeFlow: I couldn't find a way

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -83,15 +83,54 @@ def create_payments_from_vouchers(env):
         AND aml.id = aml2.id
         """
     )
+
+    if not openupgrade.column_exists(
+            env.cr, 'account_move_line', 'payment_from_invoice_id'):
```
StefanRijnhart: Can you use `openupgrade.get_legacy_name('payment_from_invoice_id')` so that the legacy character of this table is recognizable at database cleanup time?

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -83,11 +83,50 @@ def create_payments_from_vouchers(env):
...
+    env.cr.execute("""SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name='account_move_line' AND
+        column_name='payment_from_invoice_id'""")
+    if not env.cr.fetchone():
+        env.cr.execute("ALTER TABLE account_move_line ADD COLUMN"
+                       "payment_from_invoice_id INTEGER")
+
+    env.cr.execute(
+        """\
```
mreficent: delete the `\`

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -83,17 +83,57 @@ def create_payments_from_vouchers(env):
...
+            env.cr, 'account_move_line',
+            openupgrade.get_legacy_name('payment_from_invoice_id')):
+        new = openupgrade.get_legacy_name('payment_from_invoice_id')
+        env.cr.execute("""
+            ALTER TABLE account_move_line ADD COLUMN
+            %s INTEGER
+            """ % new)
+
     env.cr.execute(
         """\
```
MiquelRForgeFlow: delete the `\`

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -83,17 +83,57 @@ def create_payments_from_vouchers(env):
...
+                    SELECT aml_id FROM Q1 WHERE av_id = av.id
+                )
+            )
+        )""" % openupgrade.get_legacy_name('payment_from_invoice_id'))
+
     # Also recreate link from invoice to payment
     env.cr.execute(
-        """\
+        """
         INSERT INTO account_invoice_payment_rel (payment_id, invoice_id)
```
StefanRijnhart, line 89: It seems to me that you can fill this table directly from the following query:
```
SELECT av.id as payment_id, ai.id as invoice_id
FROM account_invoice ai
JOIN account_move_line amli ON ai.move_id = amli.move_id
JOIN account_partial_reconcile apr ON 
     (apr.credit_move_id = amli.id OR apr.debit_move_id = amli.id)
JOIN account_move_line amlp ON
     (apr.credit_move_id = amlp.id OR apr.debit_move_id = amlp.id)
JOIN account_voucher av ON av.move_id = amlp.move_id
WHERE av.voucher_type IN ('receipt', 'payment')
AND av.state IN ('draft', 'posted');
```
I might be missing something though, so please have a careful look. If it works, it would be good because it is a bit quicker than the query above.

AaronHForgeFlow: Thanks. I'll test this one and I'll come back with the result

AaronHForgeFlow: Tested with slight modification. Same result as previous approach. Thank you!

addons/account_voucher/migrations/9.0.1.0/post-migration.py
```diff
@@ -85,13 +85,19 @@ def create_payments_from_vouchers(env):
...
-        GROUP BY aml.payment_id, ai.id
+        SELECT av.id as payment_id, ai.id as invoice_id
+        FROM account_invoice ai
+        JOIN account_move_line amli ON ai.move_id = amli.move_id
+        JOIN account_partial_reconcile apr ON 
+             (apr.credit_move_id = amli.id OR apr.debit_move_id = amli.id)
+        JOIN account_move_line amlp ON
+             (apr.credit_move_id = amlp.id OR apr.debit_move_id = amlp.id)
+        JOIN account_voucher av ON av.move_id = amlp.move_id
+        WHERE av.voucher_type IN ('sale', 'purchase')
```
StefanRijnhart: Why did you change this to sale or purchase type vouchers? Only vouchers with 'receipt' or 'payment' are migrated to account.payment in the first place. type https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account_voucher/migrations/9.0.1.0/post-migration.py#L55

AaronHForgeFlow: I'm still reviewing. It came out at the time the post-migration script was executed that the voucher_type was already updated to the selection values for v9. In that case the condition is useless because it's fetching all vouchers of all types. What I don't know is why the voucher_type column was updated when there is no script for that.

StefanRijnhart: Any news on this?

pedrobaeza:
@aheficent do you finish this?

AaronHForgeFlow:
@pedrobaeza @StefanRijnhart Yes I will continue with this, sorry for the delay. I've using this script for a custom migration and it seems to be working but I'll double check to make sure it works in all cases.

PS: If anybody has used this Pull Request for testing a migration it'll be great to get some feedback here. :grimacing:

AaronHForgeFlow:
@StefanRijnhart you are right. The previous solution was not working. Besides, it makes sense to use the same voucher_type as the one used to create the payments.

To sum up, this is working fine. Tested migration in a database with more than 10000 payments. Selected   random invoices and the payment was successfully linked.

Sorry for the delay.

StefanRijnhart (approved):

StefanRijnhart:
There has been a visit from the trailing whitespace monster.

AaronHForgeFlow:
@pedrobaeza @mreficent

PR #1285, opened 2018-03-09
[11.0][MIG] migrate account_asset

AMhadhbi:
Depends on:

- [x] account - ~~#1195~~ #1295 

--


StefanRijnhart:
Dear Bloopark people, thank you for working so hard on OpenUpgrade 11. However, it is less useful to keep on proposing migration scripts that depend on account while the account script itself has not yet been completed. As a team, can you please focus on getting https://github.com/OCA/OpenUpgrade/pull/1195 in good shape, cause it is a real blocker. And after that one, keep on focussing on the migration scripts in the correct dependency order. Thanks!

AMhadhbi:
@StefanRijnhart thank for your message.
You are right the same also about migration stock module has not yet already completed. it is real big blocker (account and stock) for migration to odoo11.
we should focus on the migration in the correct dependency order: account and stock.
First we need a feedback and migration progress by @rruebner with account module and also a feedback and migration progress by @pedrobaeza  with module stock, to do not duplicated work and may be need more hands with these two blocker.
Thanks.

pedrobaeza (approved):

PR #1287, opened 2018-03-13
[IMP] Set invalid custom views to inactive

### LINKED ISSUE #1271: Error migrating from 7.0 to 8.0
I am using the migrate.py script to upgrade from v7.0 to v8.0. I run into the following error: 
"""
2018-01-24 21:22:32,629 12436 ERROR TO_migrated openerp.addons.base.ir.ir_ui_view: Element '\<group name="status" version="7.0"\>' cannot be located in parent view

Error context:
View `product.product.form`
[view_id: 221, xml_id: product.product_normal_form_view, model: product.product, parent_id: 239]
2018-01-24 21:22:32,649 12436 CRITICAL TO_migrated openerp.service.server: Failed to initialize database `TO_migrated`.
  File "/var/tmp/openupgrade/8.0/server/openerp/service/server.py", line 929, in preload_registries
  File "/var/tmp/openupgrade/8.0/server/openerp/modules/registry.py", line 370, in new
  File "/var/tmp/openupgrade/8.0/server/openerp/modules/loading.py", line 509, in load_modules
    if not Views._validate_custom_views(cr, SUPERUSER_ID, model):
  File "/var/tmp/openupgrade/8.0/server/openerp/api.py", line 268, in wrapper
  File "/var/tmp/openupgrade/8.0/server/openerp/addons/base/ir/ir_ui_view.py", line 1133, in _validate_custom_views
    return self._check_xml(cr, uid, ids, context=context)
  File "/var/tmp/openupgrade/8.0/server/openerp/api.py", line 268, in wrapper
  File "/var/tmp/openupgrade/8.0/server/openerp/addons/base/ir/ir_ui_view.py", line 197, in _check_xml
    view_def = self.read_combined(cr, uid, view.id, ['arch'], context=context)
  File "/var/tmp/openupgrade/8.0/server/openerp/api.py", line 268, in wrapper
  File "/var/tmp/openupgrade/8.0/server/openerp/addons/base/ir/ir_ui_view.py", line 561, in read_combined
    cr, uid, arch_tree, view_arch, parent_view['id'], context=context)
  File "/var/tmp/openupgrade/8.0/server/openerp/api.py", line 268, in wrapper
  File "/var/tmp/openupgrade/8.0/server/openerp/addons/base/ir/ir_ui_view.py", line 498, in apply_inheritance_specs
    self.raise_view_error(cr, uid, _("Element '%s' cannot be located in parent view") % tag, inherit_id, context=context)
  File "/var/tmp/openupgrade/8.0/server/openerp/api.py", line 268, in wrapper
  File "/var/tmp/openupgrade/8.0/server/openerp/addons/base/ir/ir_ui_view.py", line 383, in raise_view_error
    raise AttributeError(message)
AttributeError: Element '\<group name="status" version="7.0"\>' cannot be located in parent view

Error context:
View `product.product.form`
[view_id: 221, xml_id: product.product_normal_form_view, model: product.product, parent_id: 239]

"""

My customizations to v7.0 have not touched the <group name="status"> element or anything inside the <page name="Information"> in which it occurs, neither in product.product_normal_form_view nor its parent, product.product_template_form_view.

I have tried inserting the version="7.0" attribute in the relevant ir_ui_view record of my original v7.0 database, but without success. I cannot find the reference to <group name="status" version="7.0"> in the v8.0 code. I am at a loss as to how to proceed, and, more generally, I do not understand why upgrades to ui objects are critical. I mean, once other objects are upgraded, can't 8.0 menus and views be installed from scratch in the new environment?

Thanks in advance

StefanRijnhart:
Fixes #1271

--

### DIFF
--- a/openerp/addons/base/ir/ir_ui_view.py
+++ b/openerp/addons/base/ir/ir_ui_view.py
@@ -206,7 +206,7 @@ class view(osv.osv):
                     # caused by views introduced by models not year loaded
                     _logger.warn(
                         "Can't render view %s for model: %s. If you are "
-                        "migrating between major versions of OpenERP, "
+                        "migrating between major versions of Odoo, "
                         "this is to be expected (otherwise, do not run "
                         "OpenUpgrade server).", view.xml_id, view.model)
                 # RNG-based validation is not possible anymore with 7.0 forms
@@ -224,7 +224,7 @@ class view(osv.osv):
                         # caused by views introduced by models not year loaded
                         _logger.warn(
                             "Can't render view %s for model: %s. If you are "
-                            "migrating between major versions of OpenERP, "
+                            "migrating between major versions of Odoo, "
                             "this is to be expected (otherwise, do not run "
                             "OpenUpgrade server).", view.xml_id, view.model)
                         return True
@@ -233,7 +233,7 @@ class view(osv.osv):
                         # caused by views introduced by models not year loaded
                         _logger.warn(
                             "Can't render view %s for model: %s. If you are "
-                            "migrating between major versions of OpenERP, "
+                            "migrating between major versions of Odoo, "
                             "this is to be expected (otherwise, do not run "
                             "OpenUpgrade server).", view.xml_id, view.model)
                         return True
@@ -1130,7 +1130,21 @@ class view(osv.osv):
 
         ids = map(itemgetter(0), cr.fetchall())
         context = dict(load_all_views=True)
-        return self._check_xml(cr, uid, ids, context=context)
+        # OpenUpgrade: set invalid custom views to inactive
+        for view_id in ids:
+            try:
+                self._check_xml(cr, uid, [view_id], context=context)
+            except AttributeError:
+                view = self.browse(cr, uid, view_id, context=context)
+                _logger.warn(
+                    "Can't render custom view %s for model %s. "
+                    "Assuming you are migrating between major versions of "
+                    "Odoo, this view is now set to inactive. Please "
+                    "review the view contents manually after the migration.",
+                    view.xml_id, view.model)
+                self.write(
+                    cr, uid, [view_id], {'active': False}, context=context)
+        return True
 
     def _validate_module_views(self, cr, uid, module):
         """Validate architecture of all the views of a given module"""


legalsylvain (approved):
:+1:  Thanks a lot ! 

LGTM. (Code review / no test).

StefanRijnhart:
The 9.0 port here https://github.com/OCA/OpenUpgrade/pull/1288 has a test (OpenUpgrade 8.0 does not yet support testing migration scripts).

pedrobaeza:
This should be done at the end, or you can have inactive views that are later updated and converted to valid ones.

StefanRijnhart:
@pedrobaeza absolutely, and that is indeed the case. This code resides in _validate_custom_views, and it is (only) called at the very end of the register initialization: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/modules/loading.py#L509

openerp/addons/base/ir/ir_ui_view.py
```diff
@@ -1130,7 +1130,21 @@ def _validate_custom_views(self, cr, uid, model):
...
+        # OpenUpgrade: set invalid custom views to inactive
+        for view_id in ids:
+            try:
+                self._check_xml(cr, uid, [view_id], context=context)
+            except AttributeError:
+                view = self.browse(cr, uid, view_id, context=context)
+                _logger.warn(
+                    "Can't render custom view %s for model %s. "
+                    "Assuming you are migrating between major versions of "
+                    "OpenERP, this view is now set to inactive. Please "
```
pedrobaeza: s/OpenERP/Odoo

openerp/addons/base/ir/ir_ui_view.py
```diff
@@ -1130,7 +1130,21 @@ def _validate_custom_views(self, cr, uid, model):
...
+                    "review the view contents manually after the migration.",
+                    view.xml_id, view.model)
+                self.write(
```
pedrobaeza: Isn't bettet to do it via SQL for avoiding any ORM constraint?

pedrobaeza (changes requested):
Thanks for the info, Stefan, so it seems it's the way to go! Little comments.

openerp/addons/base/ir/ir_ui_view.py
StefanRijnhart: Sorry, I don't see why. And I still prefer ORM operations in Openupgrade unless for performance reasons or when such breakage actually occurs.

openerp/addons/base/ir/ir_ui_view.py
StefanRijnhart: Done

openerp/addons/base/ir/ir_ui_view.py
pedrobaeza: For not running code like the standard one: 

https://github.com/OCA/OpenUpgrade/blob/3ea086efb36dc42dd5c15a5ae428d1a34407fa6b/openerp/addons/base/ir/ir_ui_view.py#L280-L297

or any other custom code attached to the model that can be expensive or directly blocking.

StefanRijnhart: I don't understand the aversion against using the ORM all of a sudden. Using the ORM ensures the best consistency of the data. And the code you refer to is completely harmless especially in this case. Get a grip Pedro.

pedrobaeza: It's not aversion, but prevent possible future problems as this is affecting a general process and you can't imagine what people have made in their DBs/custom modules. That's why in this case, I prefer the SQL (although normally I would go for ORM). We haven't got also a rollback mechanism in this part if it fails, isn't it?

pedrobaeza (approved):

PR #1293, opened 2018-03-21
[FIX] im_odoo_support view breaks web client (internal server error)

StefanRijnhart:
im_odoo_support is a deprecated module, merged into im_livechat in OpenUpgrade 10.0
If the latter module is not installed on the 9.0 database, a broken version of im_odoo_support.assets_backend is left in the database which breaks the web client after the migration.

Note that im_odoo_support is installed by default on Odoo but not on OCB so none of the usual OpenUpgrade contributors noticed...

--

### DIFF
--- a/odoo/addons/base/migrations/10.0.1.3/post-migration.py
+++ b/odoo/addons/base/migrations/10.0.1.3/post-migration.py
@@ -1,11 +1,25 @@
 # coding: utf-8
-# © 2017 Opener BV <http://therp.nl>
+# © 2017-2018 Opener BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
 
+def handle_im_odoo_support_views(env):
+    """ im_odoo_support is a deprecated module, merged into im_livechat.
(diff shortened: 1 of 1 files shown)


hbrunn (approved):
looks fine, but still I ask myself why this record is not deleted by the standard xmlid mechanism. do we set the noupdate flag somewhere erroneously? i think i remember we had some case before where a record wasn't cleaned up that should have, but we never really got to the bottom of this

StefanRijnhart:
@hrunn it is not deleted because the module is merged into a module that is not installed. Oh, I see that I left out a *not* in the comment which must have been confusing. Fixed that now.

hbrunn:
ah, that makes sense. I'll merge optimistically

StefanRijnhart:
Oh, you were too quick to include my improved comment

hbrunn:
sorry for this. if you don't want to force push on the main branch, ping me for the fixup commit

StefanRijnhart:
I pushed https://github.com/OCA/OpenUpgrade/commit/cc215b95d3b8655d249c3139d2abc5a0f89d4c03 to the main branch. Problem solved!

PR #1295, opened 2018-03-27
11.0/account

StefanRijnhart:
Continued from https://github.com/OCA/OpenUpgrade/pull/1195 by @rruebner

--


StefanRijnhart:
Tests are currently broken on a mysterious glitch in our reference database
```
10.0-ref=> select name,module from ir_model_data where model = 'account.tax';
          name           | module  
-------------------------+---------
 1_sale_tax_template     | l10n_ae
 1_purchase_tax_template | l10n_ae
(2 rows)
```

StefanRijnhart:
I have recreated the test database (now documented at https://github.com/OCA/OpenUpgrade/wiki/How-to-create-a-reference-database). I split off #1296 for easier review of non-migration changes in this PR.

pedrobaeza:
Please rebase

pedrobaeza:
And thanks for documenting the DB preparation process!

StefanRijnhart:
Thanks for merging the other PR. I rebased this one, now we just have to wait until ansible-odoo's PR has installed Odoo 42 times https://travis-ci.org/OCA/ansible-odoo/builds/359377647

pedrobaeza:
Wow, that's not too much scalable...

StefanRijnhart:
Something is wrong with the coverage. It does not check any file. Did I mess things up in #1296?

pedrobaeza:
I'm seeing several differences with previous one: https://github.com/OCA/OpenUpgrade/blob/10.0/.coveragerc

StefanRijnhart:
Thanks. I didn't know we had one before. I'll port that one then.
https://github.com/OCA/OpenUpgrade/pull/1297

pedrobaeza:
Let's see now rebasing.

StefanRijnhart:
Done, finally!

addons/account/migrations/11.0.1.1/post-migration.py
```diff
@@ -0,0 +1,71 @@
...
+        """UPDATE account_payment
+        SET state = 'cancelled'
+        WHERE state = 'draft' AND move_name IS NOT NULL
+        AND id NOT IN (
+            SELECT payment_id FROM account_move_line
+            WHERE payment_id IS NOT NULL)""")
+    # Populate new 'sequence' field according to previous order field 'name'
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE account_payment_term apt
```
pedrobaeza, line 61: Isn't simpler to just assign `id` to `sequence`?

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,201 @@
...
+account      / account.tax.template     / cash_basis_account (many2one) : NEW relation: account.account.template
+account      / account.tax.template     / tax_exigibility (selection)   : NEW selection_keys: ['on_invoice', 'on_payment']
+# NOTHING TO DO: fields of chart template elements which do not affect installed charts
+account      / res.company              / account_opening_move_id (many2one): NEW relation: account.move
+account      / res.company              / account_setup_bank_data_done (boolean): NEW 
+account      / res.company              / account_setup_bar_closed (boolean): NEW 
+account      / res.company              / account_setup_coa_done (boolean): NEW 
+account      / res.company              / account_setup_company_data_done (boolean): NEW 
+account      / res.company              / account_setup_fy_data_done (boolean): NEW
+# NOTHING TO DO: setup wizards that are harmless or even useful to revisit
```
pedrobaeza: I would set all these fields to True, as we don't want to have these steps as pending on a migrated DB. Someone migrating a DB doesn't need to revisit these configurations steps IMO.

addons/account/migrations/11.0.1.1/post-migration.py
StefanRijnhart: Indeed, my query is more advanced ;-) and it preserves the original order.

pedrobaeza: OK

pedrobaeza:
@StefanRijnhart now that Odoo includes a field in refunds for referring to the origin invoice, and we have the community module account_invoice_refund_link that makes the same (https://github.com/OCA/account-invoicing/tree/10.0/account_invoice_refund_link), what do you think about including here a detection of the table and fill the field accordingly? I can do it if you want.

addons/account/migrations/11.0.1.1/openupgrade_analysis_work.txt
StefanRijnhart: Done

StefanRijnhart:
What field are you talking about? I can only see refund_invoice_id but that field was already implemented in Odoo 10.0: https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L1057. ~~I don't see the field that could replace origin_invoice_id from account_invoice_refund_link.~~

StefanRijnhart:
Ah, I see that the new One2many `refund_invoice_ids` is added, which does not actually replace the OCA module's origin_invoice_ids because the latter is a Many2many and allows for more than one original invoice to be combined in the same refund. So I guess this new Odoo field is not a replacement for the field from the OCA module.

pedrobaeza:
Ouch, right, the other field was already there. Then the addition should be in 10.0. I'll do it later. Right now, I'm fixing account_invoice_refund_link for 11.0 for avoiding the conflict with the field (actually removing both fields in invoice for using standard ones. The reason for being many2many instead of many2one/one2many is historical, as in the migration path, it was kept that way for not losing compatibility, but on v11 there's no reason for having a many2many). I'll ping you on the refactoring of the module, but should be fast-track because we have a broken module that if installed, can harm the DB.

pedrobaeza (approved):
Thanks for the extra tests!

PR #1296, opened 2018-03-28
Fix travis yml loading and coverage configuration

StefanRijnhart:
--


pedrobaeza (approved):
Thanks for the work refining this, which is not easy normally.

PR #1301, opened 2018-04-06
[ADD] sale_stock_picking_return_invoicing: Feature included in core sale_stock

pedrobaeza:
The field "to_refund_so" has been included in module sale_stock and the
rest of the features are the same, so only merging operation is needed.


MiquelRForgeFlow:
The name of that field in the sale_stock is "to_refund", no "to_refund_so".
So maybe a rename of that field should be performed?

pedrobaeza:
No, that's on v11. Here in v10 is still the same:

https://github.com/OCA/OpenUpgrade/blob/0de5eafc87f51eac2a0bdcffb67da1ee0eb31bfe/addons/sale_stock/models/stock.py#L16

I will rename the field on OpenUpgrade 11.0.

MiquelRForgeFlow:
Oh, true! 😂

MiquelRForgeFlow (approved):

PR #1302, opened 2018-04-06
[REN] rename mail_log_messages_to_process to mail_log_message_to_process

LoisRForgeFlow:
Module renamed on migration from v9 to v10

* [ ] Depends on https://github.com/OCA/server-tools/pull/1214


pedrobaeza (approved):
Although the PR is still pending, we can include this already, as it doesn't hurt.

PR #1303, opened 2018-04-08
[11.0][FIX] base_automation: Fix migration

pedrobaeza:
Depends on:

- [X] `add_fields` method in https://github.com/OCA/openupgradelib/pull/100

It turns out that migration was not correct all:

* Checking if module "base_action_rule" is installed will return always False, as the module has been renamed. Any way, we don't need to check it, as if the script is being executed is because the module was installed.
* Query for inserting ir_action_server record was not correct.
* The column for referencing the delegate field was not created.
* By default, queries don't return dictionaries with column names.
* Associated server actions to run weren't linked.

In addition, I have added:

* More tests
* If the fields for assigning a responsible or for adding followers are set, this is taken into account and translated to new system, so we can consider this now as a "pixel perfect" conversion.

@Tecnativa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ script:
     # on unreleased features in openupgradelib
     - pip install --ignore-installed git+https://github.com/OCA/openupgradelib.git@master
     # select modules and perform the upgrade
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Now I know why Travis didn't complain about the errors in `base_automation` migration scripts: the module was not being migrated due to addition of `|new|` and `|del|` words on the documentation table. I have adjusted the regexp expression in 9c29fb50a8cd7bba027cc74d94abf31430ea20f8

pedrobaeza:
Finally green! But this is pending of the openupgradelib PR.

pedrobaeza:
Library dependency merged, complete coverage and this is needed for fixing this like the hr_contract migration, so I merge it.

PR #1304, opened 2018-04-08
[10.0][FIX] .travis.yml: Perform migration in proper modules

pedrobaeza:
With the addition of the words |new| and |del| before the name of the module in the table, the regexp was not working anymore for that rows.

This has been detected trying base_automation module in #1303 

If everything goes green because it doesn't affect the final result, I'll merge for having a correct comparison.

@Tecnativa


pedrobaeza:
Green!

PR #1306, opened 2018-04-09
[FIX+IMP] base: Add server action on cron + warnings + tests

pedrobaeza:
I continue refining existing scripts. For base, ir.cron records shouldn't be let in an inconsistent state without proper server action, as they won't update properly. There's a warning about a missing field in noupdate XML data removed, and I have added tests for the ir.cron thing.

@Tecnativa


pedrobaeza:
This is green and with tests. Merging the refinement!

PR #1307, opened 2018-04-09
[FIX] .travis.yml: Module list for previous version

pedrobaeza:
The list of modules for marking as installed in the DB should be the one from the previous version, so it should include those with the `|del|` mark.

Let's see if this one is the definitive.

@Tecnativa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,10 +62,11 @@ script:
     # on unreleased features in openupgradelib
     - pip install --ignore-installed git+https://github.com/OCA/openupgradelib.git@master
     # select modules and perform the upgrade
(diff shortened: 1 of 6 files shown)


cubells (approved):
:+1: 

LGTM.

pedrobaeza:
Well, this is finally getting the modules, but now other errors has arisen due to the modules are correctly tested. I'll fix them here.

pedrobaeza:
And green!

MiquelRForgeFlow:
Shouldn't this travis fix be backported to previous versions?

pedrobaeza:
Yes, totally. At least for v10.

MiquelRForgeFlow:
Would you? or me?

pedrobaeza:
If you can do it, it would be great. I'm in a rush today.

PR #1318, opened 2018-04-25
[FIX+IMP] sale_stock: Recompute delivered qty + mgiration from v8

pedrobaeza:
* After setting `to_refund_so` field, we need to recompute the delivered qty, as it's not a computed field.
* If you come from a v8 migration, take into account the `invoice_status` field that excludes some returns from being invoiceable.

@Tecnativa


MiquelRForgeFlow (approved):
LGTM

rafaelbn (approved):

PR #1322, opened 2018-04-30
[MIG][11.0] website_crm: migration scripts

cubells:
Depends on:

- [X] website_form

cc @Tecnativa


addons/website_crm/migrations/11.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,11 @@
+---Fields in module 'website_crm'---
+website_crm  / website                  / crm_default_team_id (many2one): NEW relation: crm.team
+website_crm  / website                  / crm_default_user_id (many2one): NEW relation: res.users
+---XML records in module 'website_crm'---
+DEL crm.team: sales_team.salesteam_website_sales (noupdate)
+# Done: noupdate record deleted on post-migration
```
pedrobaeza: This has not been finally done.

addons/website_crm/migrations/11.0.2.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2018 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
```
pedrobaeza: The parameter is env by default since v10, so this is should be renamed and put below `env.cr`

pedrobaeza (changes requested):
Please migrate `website_form` as it's a dependency.

cubells:
@pedrobaeza website_from dependency is merged. Can you recheck this?

addons/website_crm/migrations/11.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,11 @@
+---Fields in module 'website_crm'---
+website_crm  / website                  / crm_default_team_id (many2one): NEW relation: crm.team
+website_crm  / website                  / crm_default_user_id (many2one): NEW relation: res.users
+
+---XML records in module 'website_crm'---
+DEL crm.team: sales_team.salesteam_website_sales (noupdate)
```
pedrobaeza, line 6: This is still not done. Protect the unlink the same way as seen before.

pedrobaeza: Wait, I have seen that this has been now incorporated to `sales_team` module, so you should rename in that module the XML-ID from `website_crm.salesteam_website_sales` to `sales_team.salesteam_website_sales`

cubells: @pedrobaeza 

done!

pedrobaeza (approved):

pedrobaeza (changes requested):
I'm thinking that you should make the renaming on `sales_team` module, not here, as if not, you will get duplicated XML-ID and record.

cubells:
@pedrobaeza 

done!

pedrobaeza (approved):

PR #1323, opened 2018-04-30
[MIG][11.0] payment_paypal: migration scripts

cubells:
cc @Tecnativa


addons/payment_paypal/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="payment.payment_acquirer_paypal" model="payment.acquirer">
+    <field name="view_template_id" ref="paypal_acquirer_button"/>
```
pedrobaeza: This is not correct, as this one is the old v10 value. I think you have run the script with the reversed arguments. This should be `paypal_form`.

addons/payment_paypal/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,7 @@
...
+    <field name="paypal_pdt_token"/>
```
pedrobaeza: This one is incorrect as well. It should be `dummy`.

pedrobaeza (changes requested):

cubells:
@pedrobaeza done

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #1326, opened 2018-05-02
[MIG] pad_project: Migration to 11.0

cubells:
Depends on:

- [x] project

cc @Tecnativa


addons/pad_project/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,8 @@
+---Fields in module 'pad_project'---
+pad_project  / project.project          / use_pads (boolean)            : NEW
```
pedrobaeza, line 2: You have to put this to True on all existing projects.

pedrobaeza (changes requested):

addons/pad_project/migrations/11.0.1.0/openupgrade_analysis_work.txt
cubells: @pedrobaeza  done.

pedrobaeza (approved):

PR #1329, opened 2018-05-08
[MIG][11.0] event: migration scripts

cubells:
- No update changes
- Remove rule

cc @Tecnativa


addons/event/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,87 @@
+---Fields in module 'event'---
+event        / event.event              / auto_confirm (boolean)        : not a function anymore
```
pedrobaeza, line 2: You have to write the same value as it was in v10 (the defaut `True` might not be valid): 

https://github.com/odoo/odoo/blob/14f59f4b780fa1e757cfce6a46c62c7a6b00d13e/addons/event/models/event.py#L179

addons/event/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,340 @@
...
+        </tbody>
+    </table>
+</div>
+            ]]></field>
+    <field name="subject">${object.event_id.name}: ${object.get_date_range_str()}</field>
+    <field name="reply_to" eval="False"/>
+  </record>
+  <record id="event_mail_scheduler" model="ir.cron" forcecreate="True">
+    <field name="active"/>
+    <field name="model"/>
```
pedrobaeza: Always check Travis output, as at least this field is not valid anymore in v11, and thus you get a warning. Remove it and other possible non existing fields.

addons/event/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,87 @@
...
+event        / event.event              / date_begin_located (datetime) : type is now 'char' ('datetime')
+# Nothing to do
+
+event        / event.event              / date_end_located (datetime)   : type is now 'char' ('datetime')
+# Nothing to do
+
+event        / event.event              / is_online (boolean)           : NEW
+# Nothing to do
+
+event        / event.event              / reply_to (char)               : DEL
```
pedrobaeza, line 14: You have to put as follower this email address, searching for a partner with that email or creating a new one if not exist.

addons/event/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,87 @@
...
+# Nothing to do
+
+event        / event.event              / twitter_hashtag (char)        : previously in module website_event
+# Nothing to do
```
pedrobaeza: You have to use openupgradelib method: https://github.com/OCA/openupgradelib/blob/aab221abfcf7100402942a935ef9c9cc34e01a94/openupgradelib/openupgrade.py#L1885

addons/event/migrations/11.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2018 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env.ref('event.report_event_registration_company_rule').unlink()
```
pedrobaeza: You have to prevent if the rule has been manually deleted or there's a constraint:
```python
try:
    env.ref('event.report_event_registration_company_rule').unlink()
except Exception:
    pass
```

pedrobaeza (changes requested):

pedrobaeza:
Delete also template translations

cubells:
@pedrobaeza can you recheck?

addons/event/migrations/11.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,337 @@
...
+            </tr>
+        </tbody>
+    </table>
+</div>
+            ]]></field>
+    <field name="subject">${object.event_id.name}: ${object.get_date_range_str()}</field>
+    <field name="reply_to" eval="False"/>
+  </record>
+  <record id="event_mail_scheduler" model="ir.cron" forcecreate="True">
+    <field name="active"/>
```
StefanRijnhart: Can you remove this field? It is empty because it was removed. Just as well not to override it with a truthy value because if people disabled this task they probably won't want to have it re-enabled upon migration.

pedrobaeza (approved):
With the comment by @StefanRijnhart already changed and with another little improvement that will be both squashed on merge.

PR #1330, opened 2018-05-08
[MIG][11.0] event_sale: migration scripts

cubells:
Depends on:

- [x] event

cc @Tecnativa


addons/event_sale/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,23 @@
+---Fields in module 'event_sale'---
+event_sale   / event.event.ticket       / event_type_id (many2one)      : NEW relation: event.type
+# Nothing to do
```
pedrobaeza, line 3: You have to fill this field with the proper value of the existing event tickets. I think you can have it through event_id.event_type_id, but you have to be sure that this is the expected value.

addons/event_sale/migrations/11.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2018 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env.ref('event_sale.event_type').unlink()
```
pedrobaeza: You should enclose it in a try and with savepoint, because it can be in use:
```python
try:
    with env.cr.savepoint():
        env.ref('event_sale.event_type').unlink()
except Exception:
    pass
```

pedrobaeza (changes requested):

addons/event_sale/migrations/11.0.1.1/openupgrade_analysis_work.txt
cubells: @pedrobaeza 
there is a constraint here that prevent to fill the ticket type or get the type from event.
https://github.com/OCA/OpenUpgrade/blob/11.0/addons/event_sale/models/event.py#L154
Am I right?

pedrobaeza: You're right, as I didn't get the meaning of this field.

pedrobaeza (approved):

PR #1331, opened 2018-05-08
[MIG][11.0] wesite_sale: migration scripts

cubells:
- Add noupdate changes file.
- Convert hidden selection jey to radio

cc @Tecnativa


pedrobaeza (changes requested):
You have forgotten to reset template translations. You can squash my direct commits also at the same time.

cubells:
@pedrobaeza can you recheck?

pedrobaeza (approved):

PR #1336, opened 2018-05-09
[MIG][11.0] rating: migration scripts

cubells:
cc @Tecnativa


MiquelRForgeFlow (approved):

addons/rating/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def update_res_model_id(env):
+    env.cr.execute(
+        """
+            UPDATE rating_rating
```
pedrobaeza: Incorrect SQL:
```python
from psycopg2.extensions import AsIs

openupgrade.logged_query(
    env.cr, """"
    UPDATE rating_rating rr
    SET res_model_id = im.id
    FROM ir_model im
    WHERE im.name=rr.%s""", 
    (AsIs(openupgrade.get_legacy_name('res_model')), ),
)
```

addons/rating/migrations/11.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+COLUMN_COPIES = {
+    'rating_rating': [
+        ('res_model', None, None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, COLUMN_COPIES)
```
pedrobaeza, line 16: Why copying the column?

pedrobaeza (changes requested):

addons/rating/migrations/11.0.1.0/pre-migration.py
cubells: I'm copying the column because field res_model in version 10.0 is a string with the name of  the model. In version 11.0 the field is a related to res_model_id. I copy the field to not lose the value.

pedrobaeza: Related non stored? Then better to rename it instead of copying it

cubells: It's stored so I have to copy field.

pedrobaeza: OK, thanks for the info.

pedrobaeza:
Flake8 error:
```
./addons/rating/migrations/11.0.1.0/post-migration.py:10:33: W291 trailing whitespace
```

addons/rating/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+from psycopg2.extensions import AsIs
+
+
+def update_res_model_id(env):
+    openupgrade.logged_query(
+        env.cr, """"
```
StefanRijnhart: One quote too much ;-)

cubells:
@StefanRijnhart done!

StefanRijnhart (approved):

pedrobaeza (approved):

PR #1339, opened 2018-05-16
[MIG][11.0] hr_timesheet: migration scripts

cubells:
Depends on:
- [x] sale_timesheet

cc @Tecnativa


StefanRijnhart:
Looks like you need to cover sale_timesheet as well here to make travis green, or at least that field 'is_service' which you need to precreate and fill with an SQL query.

cubells:
hi @StefanRijnhart I'm working on it.

addons/hr_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,74 @@
+---Fields in module 'hr_timesheet'---
+hr_timesheet / account.analytic.line    / employee_id (many2one)        : NEW relation: hr.employee
```
pedrobaeza, line 2: Fill this with the SQL equivalent to:
`self.env['hr.employee'].search([('user_id', '=', result['user_id'])], limit=1).id`

addons/hr_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,74 @@
...
+# Nothing to do
+
+hr_timesheet / project.project          / subtask_project_id (many2one) : module is now 'project' ('hr_timesheet')
+# Nothing to do: it must be done in project migration
```
pedrobaeza: You can group all these lines together with the same comment for better reading

pedrobaeza (changes requested):

pedrobaeza:
`sale_timesheet` gets auto-installed when installing this one, so let's see if we can merge sale_timesheet to get this one green.

addons/hr_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: Yes, this must be filled with this.

addons/hr_timesheet/migrations/11.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+
+_xmlid_renames = [
+    ('website_project_timesheet.portal_task_timesheet_rule',
+     'hr_timesheet.timesheet_rule_portal'),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_xmlids(env.cr, _xmlid_renames)
```
MiquelRForgeFlow, line 14:
```python
if openupgrade.table_exists(cr, 'website'):
```

pedrobaeza: No need of that, as the rename won't have any effect if non existant

MiquelRForgeFlow: :+1:

pedrobaeza (approved):

PR #1342, opened 2018-05-17
[MIG][11.0] sale_timesheet: migration scripts

cubells:
cc @Tecnativa


addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,66 @@
...
+# Nothing to do
+
+sale_timesheet / project.project          / sale_line_id (many2one)       : NEW relation: sale.order.line
+# Nothing to do
+
+sale_timesheet / project.task             / procurement_id (many2one)     : DEL relation: procurement.order
+# Nothing to do
+
+sale_timesheet / sale.order.line          / task_id (many2one)            : NEW relation: project.task
+# Nothing to do
```
pedrobaeza: You have to fill this field through old procurement

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,66 @@
...
+
+sale_timesheet / product.template         / service_type (False)          : NEW selection_keys: ['manual', 'timesheet'], mode: modify
+# Nothing to do
+
+sale_timesheet / product.template         / track_service (False)         : DEL selection_keys: ['manual', 'task', 'timesheet'], mode: modify
+# Nothing to do
+
+sale_timesheet / project.project          / sale_line_id (many2one)       : NEW relation: sale.order.line
+# Nothing to do
```
pedrobaeza: You have to fill this field when the SO line also created a project. This is not easy to determine, but you can take the product in the line to see if there's a project put there or not.

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,66 @@
...
+
+sale_timesheet / account.invoice          / timesheet_ids (one2many)      : NEW relation: account.analytic.line
+# Nothing to do
+
+sale_timesheet / hr.employee              / account_id (many2one)         : DEL relation: account.account
+# Nothing to do
+
+sale_timesheet / product.template         / service_tracking (selection)  : NEW selection_keys: ['no', 'project_only', 'task_global_project', 'task_new_project']
+# Nothing to do
```
pedrobaeza, line 16: You have to map old values from `track_service` to this new field.

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,66 @@
+---Fields in module 'sale_timesheet'---
+sale_timesheet / account.analytic.line    / timesheet_invoice_id (many2one): NEW relation: account.invoice
+# Nothing to do
```
pedrobaeza: Both this field and `timesheet_revenue` (and getting also o2m field `timesheet_ids`) should be computed for old entries. This can be done calling:
```python
env['account.invoice'].search([
    ('invoice_line_ids.product_id.type', '=', 'service'),
])._compute_timesheet_revenue()
```

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,66 @@
...
+
+sale_timesheet / hr.employee              / account_id (many2one)         : DEL relation: account.account
+# Nothing to do
+
+sale_timesheet / product.template         / service_tracking (selection)  : NEW selection_keys: ['no', 'project_only', 'task_global_project', 'task_new_project']
+# Nothing to do
+
+sale_timesheet / product.template         / service_type (False)          : NEW selection_keys: ['manual', 'timesheet'], mode: modify
+# Nothing to do
```
pedrobaeza: You have to set 'timesheet' value here when `invoice_policy` = `delivery` and `track_service` in ('timesheet', 'task')

pedrobaeza (changes requested):

nachoserra-domatix:
@cubells  we're gonna use this, is there anything we can do to help?

cubells:
@NachoSerra sure! you can improve my commits.

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
cubells: It's not possible because project_id is a field that not exist. Update from orm?

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
cubells: It's not possible because track_service field is not available even in pre.

pedrobaeza:
@cubells what's missing here?

cubells:
@pedrobaeza 
I guess all is done but travis is red. !!??

MiquelRForgeFlow:
Could the `sale_timesheet` and `hr_timesheet` migrations be on a same PR? I think this may solve the problem, because both "should" be merged at the same time because `sale_timesheet` has auto_install.

@cubells

pedrobaeza:
The error is the same as #1434. I have cherry-picked it on main branch and rebase this PR to see.

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,64 @@
...
+NEW ir.actions.act_window: sale_timesheet.timesheet_action_plan_pivot
+NEW ir.actions.act_window: sale_timesheet.timesheet_action_report_cost_revenue
+NEW ir.actions.act_window.view: sale_timesheet.timesheet_action_report_cost_revenue_form
+NEW ir.actions.act_window.view: sale_timesheet.timesheet_action_report_cost_revenue_pivot
+NEW ir.actions.act_window.view: sale_timesheet.timesheet_action_report_cost_revenue_tree
+NEW ir.filters: sale_timesheet.timesheet_filter_billing
+# Nothing to do
+
+DEL ir.model.data: sale_timesheet.duplicate_field_xmlid (noupdate)
+# Done: rename xmlids
```
pedrobaeza: Deleted record, not XML-ID renaming

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+
+def migrate_project_sale_line(env):
+    openupgrade.logged_query(
+        env.cr, """
+            UPDATE project_project pj
+            SET sale_line_id = sol.id
+            FROM sale_order_line sol, product_product pd, sale_order so
+            WHERE sol.product_id = pd.id
+            AND sol.order_id = so.id
+            AND so.analytic_account_id = pj.id
```
pedrobaeza: There are several things more to have here into account: the SO state, as a draft SO can't generate that project, and the product old tracking values, being sure that the product was configured to generate a new project.

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.map_values(
+        env.cr, openupgrade.get_legacy_name('service_type'), 'service_type',
+        [('task', 'timesheet')], table='product_template',
+    )
+    # Update invoice_policy
+    env['product.template'].search([
+        ('type', '=', 'service'),
+    ])._compute_service_policy()
```
pedrobaeza, line 44: I miss before doing this computation, to check if `invoice_policy` should change according old fields.

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,64 @@
+---Fields in module 'sale_timesheet'---
+sale_timesheet / account.analytic.line    / timesheet_invoice_id (many2one): NEW relation: account.invoice
+# Done: recomputed by calling _compute_timesheet_revenue method
+
+sale_timesheet / account.analytic.line    / timesheet_invoice_type (selection): NEW selection_keys: ['billable_fixed', 'billable_time', 'non_billable', 'non_billable_project']
+# Nothing to do
+
+sale_timesheet / account.analytic.line    / timesheet_revenue (float)     : NEW
+sale_timesheet / account.invoice          / timesheet_ids (one2many)      : NEW relation: account.analytic.line
+# Done: recomputed by calling _compute_timesheet_revenue method in post
```
pedrobaeza: If you do this recomputation after, then blacklist the initial computation. Check https://github.com/OCA/OpenUpgrade/pull/1421/files#diff-8022026a0884faddafd8988f759b4bb4R14 for seeing how to do it.

pedrobaeza: I'm also worried on how this is going to perform. Maybe an SQL replacement?

pedrobaeza (changes requested):

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
cubells: I don't think so. `invoice_policy` is a field that not change.

pedrobaeza: OK

addons/sale_timesheet/migrations/11.0.1.0/openupgrade_analysis_work.txt
cubells: I think It's very difficult to replace this method:
https://github.com/OCA/OCB/blob/11.0/addons/sale_timesheet/models/account_invoice.py#L54

pedrobaeza: @cubells this is definitively not needed. `timesheet_invoice_type` field is new, so no line will have this value set to `billable_time` or `billable_fixed`, and thus having all to 0 is enough.

cubells:
I'll got to do to test, a migration of sale_timesheet + hr_timesheet.

pedrobaeza:
No need of that, @cubells. The problem right now is:
```
./addons/sale_timesheet/migrations/11.0.1.0/post-migration.py:5:1: F401 'psycopg2.extensions.AsIs' imported but unused
```

MiquelRForgeFlow (approved):
🎉

cubells:
@pedrobaeza all is now :green_heart:

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,49 @@
...
+    migrate_sale_line_tasks(env)
+    migrate_project_sale_line(env)
+
+    # Update timesheet_invoice_id and timesheet_revenue
+    env['account.invoice'].search([
+        ('invoice_line_ids.product_id.type', '=', 'service'),
+    ])._compute_timesheet_revenue()
```
pedrobaeza: I have said that you have to remove this.

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,49 @@
...
+            UPDATE project_project pj
+            SET sale_line_id = sol.id
+            FROM sale_order_line sol, product_product pd, sale_order so,
+              product_template pt
+            WHERE sol.product_id = pd.id
+            AND sol.order_id = so.id
+            AND so.analytic_account_id = pj.id
+            AND pt.id = pd.product_tmpl_id
+            AND so.state in ('sale', 'done')
+            AND pt.service_tracking = 'task_new_project'
```
pedrobaeza: You can't put this condition. This is a new value in v11 and no product will have this. Remember how a new project was generated in v10. It was leaving the project field in blank and selecting the option to create new task

pedrobaeza (changes requested):

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
cubells: Ok, I understand.

cubells:
@pedrobaeza :green_heart: again!!

PR #1352, opened 2018-05-22
[MIG][10.0] rating: nothing to do

cubells:
cc @Tecnativa


addons/rating/migrations/10.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,7 @@
+---Fields in module 'rating'---
+rating       / rating.rating            / consumed (boolean)            : NEW
```
pedrobaeza, line 2: This analysis file is not correct... this field exists also in v10, but other new fields like `rating_image` or `rating_text` are not reported. @StefanRijnhart do you have idea why?

StefanRijnhart: Ah @pedrobaeza you are looking at the analysis of 10.0 so this field should be here. As for rating_image and rating_text in 11.0, these are computed fields and have been traditionally left out of the analysis as they presumably do not need any work.

pedrobaeza: Agh, I have mixed v10 and v11 as we have both. Sorry for the noise.

pedrobaeza, line 2: @cubells you have to set this to True if `rating != -1`, which is the previous default value when unset.

pedrobaeza (changes requested):

addons/rating/migrations/10.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: EDIT: Sorry, the correct comparison value is `!= -1`

cubells: @pedrobaeza 
done!

StefanRijnhart (approved):

pedrobaeza:
I'm thinking that for having the same data as in v11, you have to put the field `rating` to 0 when the value is `-1`. Maybe that illegal value makes something to crash.

cubells:
@pedrobaeza 
Can you recheck?

pedrobaeza:
@cubells you have made the contrary. Read again what I wrote. You need 2 queries.

cubells:
Oops. You are right.
thanks @pedrobaeza

pedrobaeza (approved):

PR #1365, opened 2018-06-05
[11.0][FIX] resource & project

### LINKED ISSUE #1363: [11.0] migration error in project when project_issue was not previously installed
Impacted versions:
11
Steps to reproduce:
Migrate a v10 database without project_issue installed.

Current behavior:
Traceback regarding missing relation project_issue, pretty much identical to #1000 but I lost it.

Traced it to  https://github.com/OCA/OpenUpgrade/blob/11.0/addons/project/migrations/11.0.1.1/post-migration.py#L38 and just added a return to get past and rest worked fine but obviously not the fix for all.

Expected behavior:

MiquelRForgeFlow:
Some bugs I found.

Fixes #1363

### DIFF
--- a/addons/project/migrations/11.0.1.1/post-migration.py
+++ b/addons/project/migrations/11.0.1.1/post-migration.py
@@ -43,7 +43,7 @@ def convert_issues(env):
        as the original one + ' (issues)'. This is the default option. Maximum
        information will be preserved, like the label used for issues, or the
        subtype followers.
-    2. Convert issues into tasks in the same project. It there's an issue
+    2. Convert issues into tasks in the same project. If there is an issue
        linked to a task, then the same task will be preserved for both records.
 
     This is controlled via a system parameter inserted in the DB prior to the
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Can you explain the rationale behind the dummy calendar?

MiquelRForgeFlow:
I don't understand why really, but when I was doing a migration, the resource crashed with an integrity error. There was some query executed after pre-migration and before post-migration that checks if the required column had null values and if it had then crashes.

pedrobaeza:
Please dig a bit more and further if you can assign other calendar that is not a dummy (maybe the default one).

MiquelRForgeFlow:
Ok, I will dig a bit more 🙄

MiquelRForgeFlow:
@pedrobaeza this is the traceback that I receive between the premigration and postmigration of the resource:
selection_022

pedrobaeza:
That's only a warning, not an error.

MiquelRForgeFlow:
Ok, but is there a way to fix this warning?

pedrobaeza:
Well, I haven't analyze it. Don't you have any clue?

MiquelRForgeFlow:
When a field is declared as `required`, it seems this `set_not_null` method is called. It adds a not_null_constrain to this column in the table. So that's why I thought that filling the null values with an unused value easy to find in post-migration would do the trick.

pedrobaeza:
That's not the question here on migrations. You should wonder why a field is now required and before don't (or it even doesn't exist), and which is the normal expected value (maybe there's a sane default defined in the model and you only need to call set_default method of openupgradelib?)

MiquelRForgeFlow:
Well, I give up. I will merge at least the project fix.

pedrobaeza:
OK, merging this as is. I will investigate about the resource thing.

PR #1367, opened 2018-06-07
[FIX] load noupdate data with mode init_no_create

hbrunn:
without this, migrations can break if some of the records are removed

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/post-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/post-migration.py
@@ -98,6 +98,7 @@ def migrate(env, version):
     set_filter_active(env.cr)
     openupgrade.load_data(
         env.cr, 'base', 'migrations/9.0.1.3/noupdate_changes.xml',
+        mode='init_no_create',
     )
     assign_view_keys(env)
     publish_attachments(env)


StefanRijnhart:
Have you got more details? At the first sight, it looks like all of that data is forcecreate'd, so I don't see what's the difference.

hbrunn:
won't any deleted country or currency fail because name is mandatory and none of the records is `forcecreate="True"`? I'll provide the good error message next week if you're not convinced

pedrobaeza:
There's a strange Travis error with this

StefanRijnhart:
If I am not mistaken, the default value for forcecreate is True, so all records in the update xml are forcecreated before the post script runs. (https://github.com/odoo/odoo/blob/9.0/openerp/tools/convert.py#L660)

StefanRijnhart:
If this change prevents loading admin as a member of base.group_no_one (https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml#L4-L5), then maybe there is an error in the implementation of init_no_create as well?

hbrunn:
I was indeed fixing a symptom. In the database in question, the record was gone but the xmlid record still existed, this confuses the loading mechanism. Cleaned that up, no need for the patch now

PR #1368, opened 2018-06-08
[11.0][ADD] stock: Migration scripts for v11

pedrobaeza:
Depends on:

- [x] merge_records method - https://github.com/OCA/openupgradelib/pull/105

And it's finally here! :tada: 

cc @Tecnativa


yajo (approved):

pedrobaeza:
I'm merging this one as well for continuing with the rest of the migration scripts. Any possible later problem, on subsequent PRs.

ke-real:
@pedrobaeza Hi! is stock_move_line state left out intentionally from the population script? Currently it leaves nulls there. Or rather recompute on state field is needed.

pedrobaeza:
@ke-real you are right about the state field. I will add it in a next PR.

ke-real:
@pedrobaeza One more small thing that is not clear to me. What is the reason behind merge_records in the end_migration? Why can't you call env['stock.quant']._merge_quants() there as it is the standard logic for quant merging and it is different from the logic you have there (standard logic groups by in_date as well). Basically what will happen is that migration logic merges all different in_dates together, even though the standard logic would not do it.

Not that big a deal but still a difference from standard.

pedrobaeza:
That method doesn't serve as it's used only in some cases where data is in specific state. AFAIK, `in_date` is a residual field that doesn't serve to control the real entry date (it's done through stock.move.line), so minimum value serves as a good value.

Rathnakreeswaran:
For quantity field in stock_quant.py?

ke-real:
@pedrobaeza , found another bug with move line creation.

spo.qty_done should be replaced with coalesce(spol.qty, spo.qty_done)
and actually the same applies to all the product_qty and probably ordered_qty usages.
Otherwise the lot quantities will be multiplied by the number of lots used.

pedrobaeza:
You are using outdated code. Please check new code, explicitly https://github.com/OCA/OpenUpgrade/pull/1494, that is already merged.

ke-real:
Ok, sorry my bad. This looks better.

However how do you cover Inventory moves? In v9 and v10 no operations were generated for these. And stock.move.line is always generated in v11. So as of now you would miss out the inventory move lines. I think the whole script should be based on  not stock_pack_operation but rather on stock_quant and stock_quant_move_rel. Or a combination of both.

pedrobaeza:
You can't base this on quants. I don't think there's any problem with inventory lines, as they generate stock moves, and picking is not required in the SQL.

ke-real:
Well I'd rather keep the database consistent. Especially having in mind that quants are something temporary in v11 - otherwise you won't be able to track anything.
Btw, why do you think it is not possible to base on quants? (I mean before merging them)

pedrobaeza:
Because they don't track traceability for example and lot of reasons.

PR #1371, opened 2018-06-13
[IMP] improve the documentation for manual upgrade

ap-wtioit:
* Move automatic upgrade down (it is somehow deprecated)
* Move general tips up to the top for users to discover them before
starting the upgrade
* Add (code) samples how to perform the steps
* Add a section with general errors that can happen during upgrade and
  their solutions
* Minor wording stuff

We tried to improve the documentation for the manual upgrade to be easier to follow

Current behavior before PR:
The documentation for the manual upgrade is hard to follow through and it is not obvious that the manual migration is the prefered variant

It should be easier for developers new to OpenUpgrade to perform an Odoo upgrade with OpenUpgrade

Info @wt-io-it

Thanks for the help during OCA CodeSprint Sevilla 2018
@js-landoo
@ig-landoo
@pabloqb 

--


odoo/openupgrade/doc/source/migration_details.rst
```diff
@@ -6,27 +6,65 @@ and upgrade your database by calling odoo-bin (or openerp-server) directly. You
...
 2. Compare your set of installed modules with the modules that are covered by
    the OpenUpgrade Addons branch you are using. Upgrading a database that has
    uncovered modules installed is likely to fail. Authoritative in this respect
    is the existence and contents of a *user_notes.txt* file in the
    migrations/[version] subdirectory of each module. We also try to indicate
    module coverage in the documentation but it sometimes lags behind.
 
+   .. code-block:: bash
+
+       $ diff odoo/addons OpenUpgrade/addons/ | grep -v "Common subdirectories" | grep -v "OpenUpgrade"
```
StefanRijnhart, line 44: What does this do?

odoo/openupgrade/doc/source/migration_details.rst
```diff
@@ -38,20 +76,11 @@ uncovered modules.
...
-* When installing the openupgradelib make sure you check out the latest version from github to get the latest updates and fixes. Either by following the method from *migrate.py* or installing via pip like so::
-
-    $ pip install git+git://github.com/OCA/openupgradelib.git
-
 Configuration options
 +++++++++++++++++++++
 
 OpenUpgrade allows for the following configuration options. Add these options
-to a separate stanza in the server configuration file under a header
+to a separate line in the server configuration file under a header
```
StefanRijnhart, line 83: I think `a separate stanza` (a group of lines) as a description of one or more lines is better than `a separate line` as a description of one or more lines.

StefanRijnhart:
In all honesty, part of this change concerns installing Odoo from source. While you need to be able to do that when installing Odoo, writing a manual to do so is out of scope for the OpenUpgrade documentation and needs to be delegated elsewhere. What is useful of course is mentioning specific pitfalls like having an outdated release of openupgradelib installed instead of the master branch.

odoo/openupgrade/doc/source/migration_details.rst
ap-wtioit: this compares the addons directory of odoo with OpenUpgrade and ignores lines with "Common subdirectory" (existing in OpenUpgrade) or "OpenUpgrade" (existing only in OpenUpgrade, hence not installed on the odoo server)

StefanRijnhart: Still not clear what you are trying to do here. I assume `OpenUpgrade` is the distribution of openupgrade that you cloned under `1` , but what is `odoo` in this case?

ap-wtioit: odoo would be your local Odoo instance

js-landoo:
Hi @StefanRijnhart 

We have no problem updating content in the way you suggest.
Just confirm us the changes to be done.

Many thanks.

odoo/openupgrade/doc/source/migration_details.rst
pedrobaeza: I don't see this relevant for a migration

odoo/openupgrade/doc/source/migration_details.rst
```diff
@@ -65,3 +94,107 @@ to a separate stanza in the server configuration file under a header
...
+        migrate(self.cr, installed_version)
+      File "/home/user/.virtualenvs/python3/lib/python3.6/site-packages/openupgradelib/openupgrade.py", line 1204, in wrapped_function
+        (module, filename, str(e).decode('utf8')))
+    AttributeError: 'str' object has no attribute 'decode'
+
+  You need to make sure to use the openupgradelib from github:
+
+  .. code-block:: bash
+
+    $ pip install --upgrade git+git://github.com/OCA/openupgradelib.git
```
pedrobaeza, line 139: You have already said this requirement, so this is redundant

odoo/openupgrade/doc/source/migration_details.rst
```diff
@@ -65,3 +94,107 @@ to a separate stanza in the server configuration file under a header
...
+        loaded_modules, processed_modules = load_module_graph(cr, graph, status, perform_checks=update_module, report=report, upg_registry=upg_registry)
+      File "/home/user/OpenUpgrade/odoo/modules/loading.py", line 139, in load_module_graph
+        migrations.migrate_module(package, 'pre')
+      File "/home/user/OpenUpgrade/odoo/modules/migration.py", line 165, in migrate_module
+        mod = load_script(pyfile, name)
+      File "/home/user/OpenUpgrade/odoo/modules/migration.py", line 26, in load_script
+        return imp.load_module(module_name, fp, fname, ('.py', 'r', imp.PY_SOURCE))
+    ValueError: load_module arg#2 should be a file or None
+
+  You need to make sure you are calling OpenUpgrade for upgrading to version 11 with python3:
```
pedrobaeza, line 165: This is a requirement of v11, not OpenUpgrade itself. One should know this, but well we can remind it, but not with all this text.

odoo/openupgrade/doc/source/migration_details.rst
```diff
@@ -65,3 +94,107 @@ to a separate stanza in the server configuration file under a header
...
+        from . import db, graph, loading, migration, module, registry
+      File "/home/user/OpenUpgrade/odoo/modules/graph.py", line 10, in <module>
+        import odoo.tools as tools
+      File "/home/user/OpenUpgrade/odoo/tools/__init__.py", line 7, in <module>
+        from . import pdf
+      File "/home/user/OpenUpgrade/odoo/tools/pdf.py", line 4, in <module>
+        from PyPDF2 import PdfFileWriter, PdfFileReader
+    ModuleNotFoundError: No module named 'PyPDF2'
+
+  You may have created the virtualenv for a python version that differs from the one used in odoo-bin.
```
pedrobaeza, line 195: Again something about your skills with Python, not with OpenUpgrade itself.

odoo/openupgrade/doc/source/migration_details.rst
ap-wtioit: All this text should help users that strugle with using OpenUpgrade (and the goal was to improve the documentation so OpenUpgrade is easier to use) to find the solution to their problem with google. I could shorten the stacktrace to the first and last line with 3 dots in between. Would that help?

pedrobaeza: The text can vary on time, so the stack trace is not useful at all. Just say that OpenUpgrade v11 needed Python 3, and not here in this section.

ap-wtioit: If we remove the stack trace completely we would loose the ability to search for errors on google and find the correct solution.

pedrobaeza: I repeat: this stack trace would change in the future, so it's useless (maybe it has changed already now that we merged latest changes in #1398)

js-landoo:
I suggest to leave "traceback" line and last one, this way.
Just 3 lines, but useful, doesn't matter if in future they are exactly the
same, they will help:

> +    Traceback (most recent call last):
+    ...
+    ValueError: load_module arg#2 should be a file or None


-----------------------------------------------------------
Josean Soroa
LANDOO SL


2018-07-03 9:48 GMT+02:00 Pedro M. Baeza <notifications@github.com>:

pedrobaeza:
Closing this as the reasons stated due to not being useful for OpenUpgrade itself, but mixed knowledge about deployment, Python, etc.

PR #1372, opened 2018-06-13
[IMP] add a guide how to setup pycharm to perform a migration

ap-wtioit:
We are adding a new step by step guide how to perform a migration/upgrade with pycharm

Current behavior before PR:
-

It should be easier for developers to prepare an OdooUpgrade on their system

This was done at the OCA Code Sprint Sevilla 2018 by:
@js-landoo
@ig-landoo
@pabloqb
@ap-wtioit

Info @wt-io-it

--


odoo/openupgrade/doc/source/scripts/post-migration.py
```diff
@@ -0,0 +1,121 @@
...
+            ON imd.module = imm.name
+            WHERE imm.state <> 'installed'
+            AND imd.model = 'ir.ui.view')
+    """)
+    conn.commit()
+
+
+def set_not_ported_modules_to_installed(conn, cr):
+    print("""set not yet ported modules to installed
+    (otherwise, updating a module to work on becomes tricky)""")
```
StefanRijnhart, line 89: I don't understand. What is the purpose of this? I think setting the module state to 'installed' is misleading. Keeping the 'to update' state is useful to validate the result of the migration.

odoo/openupgrade/doc/source/scripts/post-migration.py
```diff
@@ -0,0 +1,121 @@
...
+for a in options[:]:
+    if a[0] == '--db_port' and a[1] != '':
+        db_port = a[1]
+        options.remove(a)
+        break
+
+
+def disable_inherit_unported_modules(conn, cr):
+    print("""defuse inheriting views originating from
+    not yet ported modules""")
```
StefanRijnhart, line 70: Same as below. It makes your database work but it does not seem proper to me to do it in this way.

odoo/openupgrade/doc/source/intro.rst
```diff
@@ -54,4 +54,5 @@ Migrating your database
 
    migrate.py
    migration_details
+   migration_pycharm
```
StefanRijnhart, line 57: Do you need to add a migration_pycharm.rst?

ap-wtioit: yep sorry, two files were missing. thanks for checking so fast.

odoo/openupgrade/doc/source/scripts/post-migration.py
ap-wtioit: This helps login in to your not fully migrated Odoo instance and check in the UI what worked and what is missing. If modules are marked as "to update" Odoo will try to update them again on login (keeping the user from analysing it with the UI). This is an additional script that's not needed always but useful sometimes.

js-landoo:
Hi @StefanRijnhart what's next step?

Explanations from @ap-wtioit are good for you?
I not, we can delete both methods, they are not the main goal of this issue. 
When we started the analysis on monday, @mreficent just suggested them and we thought it was a nice way to add some additional (and quite personnel) logic.

I have reviewed generated html and I find it correct.

We wait for your news. Thanks.

MiquelRForgeFlow (commented):
Why this <code>.rst</code> file is focused on PyCharm? I think this PR could be generalized, because several IDEs are compatible with what here is written.

js-landoo:
Yes @mreficent you are right but we feel that when we try to make more general documents, at the end we leave it too open and finally becames more difficult for someone with less skills.
Our main goal has been to make a very easy guide, without different meanings.
Once this guide is published, it will be nice to publish a more generic one as you say. It will be easy just copying commands and parameters given to pycharm and running them in shell. If you want, we can colaborate to do it.

pedrobaeza:
I feel that this is very bloated for an operation that should depend on each implementator. For example, we use our Docker system and don't need any of these things. Maintaining and supporting (errors and questions generated from it) this documentation is out of scope IMO. If you want this, you should keep it in your side.

pedrobaeza:
Again very specific for one environment, and no useful for others. This kind of "documentation" is something to be put on a personal/company blog, but not on OpenUpgrade documentation.

PR #1373, opened 2018-06-15
[9.0][MIG] im_chat: merged in mail

MiquelRForgeFlow:
- `im_chat.session` was replaced by `mail.channel`
- `im_chat.shortcode` was renamed into `mail.shortcode`
- `im_chat.presence` was moved to `bus` module
- js and controller code was moved from `im_chat` to `mail` module

--


MiquelRForgeFlow:
Are you waiting for a proof? Ok. As you can see in https://github.com/odoo/odoo/commit/4af96310721237d3640bbd162ff9ea84e627028d, https://github.com/odoo/odoo/commit/a9a78ee9b6a499c29b284e1612951cc20249502f and https://github.com/odoo/odoo/commit/449f25463843767ba41a9a8c2adfdd4ceb2f2e5c, `im_chat` code was moved to `mail` and `bus`.

MiquelRForgeFlow:
@pedrobaeza can you look at this PR?

StefanRijnhart:
I'm not familiar with the chat module and merging is fine by me but does the migration need anything else?

pedrobaeza:
@mreficent shouldn't you need to rename/merge models like im.chat.session, im.shortcode, and so on?

MiquelRForgeFlow:
maybe yes or maybe not, because after those commits Odoo refactored mail. I need to check in detail in a dummy migration if those renames are needed or not.

pedrobaeza:
OK, please check to be sure no information is lost.

MiquelRForgeFlow:
I have the scripts nearly finished. I will push them when finished.

pedrobaeza:
OK

MiquelRForgeFlow:
@pedrobaeza @StefanRijnhart Scripts uploaded.

But I have found a strange bug that I don't know how to fix. After the migration, if for example as an admin you create a new user and try to chat with that user, you will have an AccessError message because the new channel is not really created and I don't know why. Any clue??

StefanRijnhart:
@mreficent does that bug only occur on migrated databases?

MiquelRForgeFlow:
@StefanRijnhart Yes, but I don't have tested if without this PR also occurs.

StefanRijnhart:
I cannot reproduce on a migrated database with 8.0 demo data, but that one does not have any preexisting mail channels. But do you have duplicate entries in mail_channel_partner? This query determines whether a new channel is going to be created, and I think it would be sensitive to such entries: https://github.com/odoo/odoo/blob/9.0/addons/mail/models/mail_channel.py#L357-L364

MiquelRForgeFlow:
No, this code doesn't create duplicate entries in mail_channel_partner

pedrobaeza:
Let's merge this one and experience the possible side effects. I will have feedback soon, but don't block the general method as everything is green.

MiquelRForgeFlow:
@pedrobaeza OMG, you merged this, haha. Any sweet feedback?

pedrobaeza:
Not time to check it yet. Will tell in a few days.

PR #1375, opened 2018-06-18
[MIG][11.0] website_portal_sale: merged into sale_payment

cubells:
cc @Tecnativa

### DIFF
--- a/addons/account/migrations/11.0.1.1/pre-migration.py
+++ b/addons/account/migrations/11.0.1.1/pre-migration.py
@@ -12,6 +12,10 @@ column_copies = {
     ],
 }
 
+_portal_xmlid_renames = [
+    ('website_portal_sale.portal_my_invoices', 'account.portal_my_invoices'),
+]
+
 
 @openupgrade.migrate()
 def migrate(env, version):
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
I don't see the change in website_sale for renaming the remaining XML-ID. And please rebase (fetching first the origin) for avoiding the current weird error.

sergio-teruel:
@cubells I'm not sure of this... I think that website_portal_sale is merged into sale module, you can see the views definitions and controllers here
v10 (https://github.com/OCA/OCB/blob/10.0/addons/website_portal_sale/views/website_portal_sale_templates.xml#L23) 
v11 (https://github.com/OCA/OCB/blob/11.0/addons/sale/views/sale_portal_templates.xml#L20)

pedrobaeza:
See the internal task for an explanation, but indeed this lacks the the XML-ID renaming I have already mentioned

sergio-teruel:
OK. We wait to remame XML-ID

cubells:
@sergio-teruel @pedrobaeza can you recheck?

pedrobaeza (approved):

PR #1378, opened 2018-06-19
[MIG][11.0] website_sale_b2c: merged into sale

cubells:
cc @Tecnativa


pedrobaeza (changes requested):
The approach is not the correct: making the merge of the module directly with sale, you have automatically renamed the XML-IDs and the module is not marked as missing

pedrobaeza:
I'm seeing that the renaming is already performed (and in the proper place - please take care with these details -): https://github.com/OCA/OpenUpgrade/pull/1378/files#diff-2512f1b5911da20e24162e3cbe2fc068R80

Thus I close this PR.

cubells:
@pedrobaeza ok, thanks!

PR #1381, opened 2018-06-19
[9.0][MIG] portal_project & portal_project_issue

MiquelRForgeFlow:
- Commit https://github.com/odoo/odoo/commit/50fa3b6ba4227946d60eeaf57af70fdaab73dce2 renamed `portal_project_issue` to `website_project_issue`.

- Commit https://github.com/odoo/odoo/commit/8bd6074f03232926d679662a4d8de07a8cef8ddc deleted `portal_project` (putting content to `project`).

--


openerp/addons/openupgrade_records/lib/apriori.py
```diff
@@ -4,6 +4,7 @@
 
 renamed_modules = {
     'portal_claim': 'website_crm_claim',
+    'portal_project_issue': 'website_project_issue',
```
pedrobaeza: This renaming is dangerous, as if you don't have website module installed, this will install all the website stack. You should merge it with `project_issue` instead.

pedrobaeza (changes requested):
Check also if you need to change any data in projects (like visibility), for allowing the same kind of access before the migration.

openerp/addons/openupgrade_records/lib/apriori.py
MiquelRForgeFlow: Done

pedrobaeza (approved):

PR #1389, opened 2018-06-21
[MIG][11.0] fleet: migration scripts

cubells:
cc @Tecnativa


addons/fleet/migrations/11.0.0.1/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.map_values(
+        env.cr, openupgrade.get_legacy_name('category'), 'category',
+        [('both', 'contract')], table='fleet_service_type',
+    )
+    openupgrade.map_values(
```
pedrobaeza: You have made it the contrary: the values you map are new, not the contrary. The best chance is to execute manually `scheduler_manage_contract_expiration` method of `fleet.vehicle.log.contract` for having the possible contracts on the correct state.

pedrobaeza (changes requested):

addons/fleet/migrations/11.0.0.1/post-migration.py
pedrobaeza: Obviously, there's no need of copying the column.

cubells:
@pedrobaeza thanks!

pedrobaeza (approved):

PR #1392, opened 2018-06-21
[MIG][11.0] stock_dropshipping: no update changes

cubells:
cc @Tecnativa


addons/stock_dropshipping/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,5 @@
+---Fields in module 'stock_dropshipping'---
+stock_dropshipping / purchase.order.line      / sale_line_id (many2one)       : NEW relation: sale.order.line
```
pedrobaeza, line 2: You have to fill this field with the proper value. Take it from the associated procurement.order

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #1397, opened 2018-06-22
[MIG][10.0] website_sale_delivery: migration scripts

cubells:
cc @Tecnativa


addons/website_sale_delivery/migrations/10.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+COLUMN_COPIES = {
+    'delivery_carrier': [
+        ('website_description', None, None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, COLUMN_COPIES)
```
pedrobaeza: This line gives error on Travis. Are you sure the content is not already on HTML?

pedrobaeza (changes requested):

addons/website_sale_delivery/migrations/10.0.1.0/pre-migration.py
cubells: No, It's a related field to product.description_sale.

The conversion to html it must be done in product migration.

pedrobaeza (approved):

PR #1402, opened 2018-06-24
[ADD] stock_account: Migration scripts

pedrobaeza:
This one is not easy when you have stock valuation through accounting. As OCA instance nor my customers have this kind of accounting, I'm afraid I'm leaving half this migration, covering only the rest of the cases.

cc @Tecnativa


davidtranhp:
I believe the costing method fifo is to replace the real in the previous version. Fifo in combination with Lot/Serial will be real costing method, in term of accounting.

StefanRijnhart:
I think @davidtranhp has a point. Also, leaving the `remaining` fields empty looks harmless as it just applies the standard price in that case. I'll try to look into the impact some more later on. In the meantime, can you attend to the test error 
```
2018-06-24 17:04:28,003 4943 ERROR openupgrade OpenUpgrade: stock_account: error in migration script stock_account/migrations/11.0.1.1/pre-migration.py: not enough arguments for format string
```

pedrobaeza:
I checked some code and seems that `average` fits more with previous `real` one, but I have change to `fifo` according your comments. Travis rebuild with the fix for the error in progress.

pedrobaeza:
Travis is OK now. Please check.

StefanRijnhart:
I need to look into the impact of the valuation types some more later on. If you are in a hurry, can you add this to the module status as `Done ('real' product valuation not yet covered')` and you can merge this.

pedrobaeza:
@StefanRijnhart I have changed then the description on the documentation for reflecting that missing part. You can merge when you want.

addons/stock_account/migrations/11.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,45 @@
+---Fields in module 'stock_account'---
+stock_account / account.move             / stock_move_id (many2one)      : NEW relation: stock.move
+stock_account / stock.move               / account_move_ids (one2many)   : NEW relation: account.move
+# TODO: Not easy way to figure out previous valuation moves from which stock.move comes from. Only impacts if stock valuation accounting is enabled.
+
+stock_account / product.category         / property_cost_method (selection): selection_keys is now '['average', 'fifo', 'standard']' ('['average', 'real', 'standard']')
+stock_account / product.template         / property_cost_method (selection): selection_keys is now '['average', 'fifo', 'standard']' ('['average', 'real', 'standard']')
+# Done: pre-migration: The real costing method doesn't exist anymore, so the most similar one is 'average'. Values mapped according this.
```
MiquelRForgeFlow: -> fifo 🤓

odoo/openupgrade/doc/source/modules100-110.rst
```diff
@@ -473,7 +473,7 @@ missing in the new release are marked with |del|.
 +-----------------------------------+-------------------------------------------------+
 |stock                              | Done                                            |
 +-----------------------------------+-------------------------------------------------+
-|stock_account                      |                                                 |
+|stock_account                      | Done (real product valuation not yet covered')  |
```
MiquelRForgeFlow, line 476: -> `'real'`

MiquelRForgeFlow (approved):
😏

addons/stock_account/migrations/11.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: hehe

odoo/openupgrade/doc/source/modules100-110.rst
pedrobaeza: It's very long, so I remove quotes

addons/stock_account/migrations/11.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: Changed by UI, so please squash on merge

StefanRijnhart (approved):

ke-real:
Hi

Will FIFO be covered some time soon? I can see this migration does not touch value and remaining_value at all, which means that all valuation will fallback to std price.

pedrobaeza:
Not plans by my part. If you are interested, you can fund any of this project's contributors for doing it (I'm not the most suitable one, as I don't use it).

ke-real:
I could try to do it myself, but I'm not sure if there is a point since FIFO implementation in Odoo 11 is useless in its current state..

pedrobaeza:
Well, I can't give an opinion on that matter I'm afraid.

StefanRijnhart:
@ke-real what is wrong with FIFO in Odoo 11? Have you got any pointers?

ke-real:
@StefanRijnhart well, first it is company level FIFO, not warehouse/location level FIFO- this can be ok for single warehouse setup, probably not acceptable for multi-warehouse with separate cost centers (of course business needs can be different as well).
Another issue is that it does not take lots/serials into account - this can be arguable but in the end it is up to business to decide which way it has to go.

I'm not quite through all the changes yet, so there might be something else.

PR #1403, opened 2018-06-25
[MIG][10.0] purchase_requisition: migration scripts

cubells:
cc @Tecnativa


addons/purchase_requisition/migrations/10.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
...
+purchase_requisition / purchase.order.line      / quantity_tendered (float)     : DEL
+purchase_requisition / purchase.requisition     / exclusive (selection)         : DEL required: required, selection_keys: ['exclusive', 'multiple'], req_default: multiple
+purchase_requisition / purchase.requisition     / multiple_rfq_per_supplier (boolean): DEL
+# Nothing to do
+
+purchase_requisition / purchase.requisition     / type_id (many2one)            : NEW relation: purchase.requisition.type, required: required, req_default: function
+# Nothing to do
+
+purchase_requisition / purchase.requisition     / vendor_id (many2one)          : NEW relation: res.partner
+# Done: upgrade vendor with partner_id of purchase order
```
pedrobaeza: You should leave it blank, as it's not required and you can have quotations from several vendors. This is a new feature for autocompleting new RfQs only.

addons/purchase_requisition/migrations/10.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
...
+
+purchase_requisition / purchase.requisition     / website_message_ids (one2many): DEL relation: mail.message
+# Nothing to do
+
+purchase_requisition / purchase.requisition.line / price_unit (float)            : NEW
+# Nothing to do
+
+purchase_requisition / purchase.requisition.line / product_id (many2one)         : now required
+# Nothing to do
```
pedrobaeza: You should assign to an existing product if null. Check if there's a product in standard that you can assign, and if not, you should create a new one

addons/purchase_requisition/migrations/10.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
+---Fields in module 'purchase_requisition'---
+purchase_requisition / purchase.order.line      / quantity_tendered (float)     : DEL
+purchase_requisition / purchase.requisition     / exclusive (selection)         : DEL required: required, selection_keys: ['exclusive', 'multiple'], req_default: multiple
+purchase_requisition / purchase.requisition     / multiple_rfq_per_supplier (boolean): DEL
+# Nothing to do
+
+purchase_requisition / purchase.requisition     / type_id (many2one)            : NEW relation: purchase.requisition.type, required: required, req_default: function
```
pedrobaeza, line 7: As this is created before having data loaded, the default is not applied. You should reapply it after, and try to avoid the ERROR on NOT NULL constraint meanwhile.

pedrobaeza (changes requested):
flake8 error:
```
./addons/purchase_requisition/migrations/10.0.0.1/post-migration.py:5:1: F401 'psycopg2.extensions.AsIs' imported but unuse
```

pedrobaeza (approved):

Issue #1380, opened 2018-06-19
Bad SQL request at migration tool - execution possible in postgres database


piapatricktresp:
Impacted versions: v11

Steps to reproduce: use a odoo 10 database and try to migrate to v11

Current behavior: stops at a bad query:
ALTER TABLE "hr_holidays" ADD FOREIGN KEY ("manager_id") REFERENCES "hr_employee"("id") ON DELETE set null'
ERROR: column "manager_id" referenced in foreign key constraint does not exist

Expected behavior: execute query

Video/Screenshot link (optional):

I am working on a database upgrade from v10 to v11 the last couple of days. i got stuck many times, because there were some table datas missing or some constrains violated, which i could get passed.

But now i get an error with this statement:

ALTER TABLE "hr_holidays" ADD FOREIGN KEY ("manager_id") REFERENCES "hr_employee"("id") ON DELETE set null'
with this ERROR: column "manager_id" referenced in foreign key constraint does not exist

If i execute this SQL-Query direct in the database i do not get any errors. Since i have no clue where in the migration tool this query is created, i do not see any chance to move forward. Do you have any hints, what to do?

( i have the community edition so i can not use the service from odoo )

Thanks!

Patrick

pedrobaeza:
hr_holidays module is still not migrated, so the error should come from that. Please check current coverage in #1005.

piapatricktresp:
ok thank you very much for this hint!

pedrobaeza:
`hr_holidays` is already migrated

BT-fschubert:
Got the same with hr_holidays commit - any idea why this still happens?

pedrobaeza:
Check if other third party module adds the same column

PR #1405, opened 2018-06-25
[MIG][11.0] hr_holidays: migration scripts

cubells:
cc @Tecnativa


novawish:
I'm just curious: Is this issues (https://github.com/OCA/OpenUpgrade/issues/1380) fixed in here?

pedrobaeza:
That's the idea, but right now, it seems that there are no changes, so it can be 2 things:

* We have to do something, as the red Travis suggests.
* There's something bad on the DB of the issue.

cubells:
@novawish Yes, it is.

Now, this pr fix https://github.com/OCA/OpenUpgrade/issues/1380

addons/hr_holidays/migrations/11.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,28 @@
+---Fields in module 'hr_holidays'---
+hr_holidays  / hr.employee              / leaves_count (integer)        : type is now 'float' ('integer')
+# Nothing to do
+
+hr_holidays  / hr.holidays              / first_approver_id (many2one)  : NEW relation: hr.employee
+# Done: renamed field
```
pedrobaeza: You shouldn't rename the manager field, but:

* Add the field in pre.
* Copy the values from manager_id in pre.

addons/hr_holidays/migrations/11.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,28 @@
...
+
+hr_holidays  / hr.holidays              / manager_id2 (many2one)        : was renamed to second_approver_id [nothing to do]
```
pedrobaeza: Always remember that until v12, you have to rename the field also although it has `oldname` attribute

pedrobaeza (changes requested):

cubells:
done @pedrobaeza

addons/hr_holidays/migrations/11.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,19 @@
...
+
+field_renames = [
+    ('hr.holidays', 'hr_holidays', 'manager_id2', 'second_approver_id')
+]
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    openupgrade.rename_fields(env, field_renames)
+    openupgrade.logged_query(
```
pedrobaeza: You should use for better completion the new openupgradelib method `add_fields`: https://github.com/OCA/openupgradelib/blob/7d9650376e1b64838b6caf2e90581176ac683808/openupgradelib/openupgrade.py#L1815

I'm adding it already

cubells: Perfect @pedrobaeza 

I didn't know this method.

pedrobaeza: It was added recently by me as I observed that there was a gap (the created XML-ID) and to try to simplify which PG type to use for adding the column.

On v12 won't be needed for the XML-ID part.

StefanRijnhart (approved):
What a sneaky bug in Odoo, to rename a field that is still defined...

Issue #1404, opened 2018-06-25
[9.0] migrate database to odoo 9 : ValueError: External ID not found in the system: portal_stock.portal_stock_picking_user_rule


AMhadhbi:
Impacted versions:
9.0
Steps to reproduce:

./openerp-server -d yarab -u all --stop-after-init

Current behavior:
ValueError: External ID not found in the system: portal_stock.portal_stock_picking_user_rule
2018-06-25 13:29:26,936 2324 **CRITICAL** yarab openerp.service.server: Failed to initialize database `yarab`.
Traceback (most recent call last):
  File "/var/www/openerp/service/server.py", line 897, in preload_registries
    registry = RegistryManager.new(dbname, update_module=update_module)
  File "/var/www/openerp/modules/registry.py", line 390, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/var/www/openerp/modules/loading.py", line 412, in load_modules
    loaded_modules, update_module, upg_registry)
  File "/var/www/openerp/modules/loading.py", line 299, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks, upg_registry=upg_registry)
  File "/var/www/openerp/modules/loading.py", line 137, in load_module_graph
    migrations.migrate_module(package, 'pre')
  File "/var/www/openerp/modules/migration.py", line 169, in migrate_module
    mod.migrate(self.cr, pkg.installed_version)
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.3.2-py2.7.egg/openupgradelib/openupgrade.py", line 1430, in wrapped_function
    if use_env2 else cr, version)
  File "/var/www/addons/portal_stock/migrations/9.0.0.1/pre-migration.py", line 10, in migrate
    env.ref('portal_stock.portal_stock_picking_user_rule').unlink()
  File "/var/www/openerp/api.py", line 792, in ref
    return self['ir.model.data'].xmlid_to_object(xml_id, raise_if_not_found=raise_if_not_found)
  File "/var/www/openerp/api.py", line 248, in wrapper
    return new_api(self, *args, **kwargs)
  File "/var/www/openerp/api.py", line 490, in new_api
    result = method(self._model, cr, uid, *args, **old_kwargs)
  File "/var/www/openerp/addons/base/ir/ir_model.py", line 1028, in xmlid_to_object
[...]

PR #1406, opened 2018-06-25
[ADD] [9.0] portal_stock: avoid unlink error

cubells:
cc @Tecnativa

Fix https://github.com/OCA/OpenUpgrade/issues/1404


MiquelRForgeFlow:
I don't know if it's related but, @pedrobaeza wanted to change the merge of `portal_stock` from `portal` to `stock`. I don't know if he still wants to change that or no.

pedrobaeza:
Not related. This is v9

PR #1408, opened 2018-06-26
[FIX] Move dates on last day of period are rewritten to first day

### LINKED ISSUE #1407: [9.0] Moves dated on the last day of a period are re-written to the first day
When migrating from v8 to v10, one of my users found that some moves that were dated on the last day of a period are re-written to the first day of a period !!!
I think the bug is located here: https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/post-migration.py#L1102

This SQL request takes into account the case where the move is dated at the first day and at the last day of the period, whereas it should only take into account the moves that are OUT of the period. So it seems to me that the SQL request should be ("<=" changed to "<" and "=>" changed to ">":

```
UPDATE account_move am
        SET date = ap.date_start
        FROM account_period ap
        WHERE am.period_id = ap.id
            AND (am.date < ap.date_start OR am.date > ap.date_stop)"""
```

I haven't tested the fix above, but the bug seems obvious. If I understand this SQL request correctly, it seems that ALL account moves dated at the last day of a period are re-written to the first day of the same period. It was introduced by @pedrobaeza in September 2017 in this commit: https://github.com/OCA/OpenUpgrade/commit/62d391560e80037f8bfe8925d756e804bdc3a89d

It is a very important bug because it modifies the date of a valid account move, which should never happen.

StefanRijnhart:
[FIX] Move dates on last day of period are rewritten to first day
Original SQL also selected moves on the last day of the period.
Fixes #1407, courtesy of Alexis de Lattre

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -1088,7 +1088,7 @@ def update_account_invoice_date(cr):
         WHERE am.period_id = ap.id
             AND am.id = ai.move_id
             AND ai.date IS NULL
-            AND (am.date <= ap.date_start OR am.date >= ap.date_stop)"""
+            AND (am.date < ap.date_start OR am.date > ap.date_stop)"""
     )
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

alexis-via:
@StefanRijnhart  Thank you very much for the fix

PR #1410, opened 2018-06-27
[9.0][FIX] sale_stock: Compute delivered quantities in some conditions

rousseldenis:
If the product delivered is different from the ordered one, don't call compute_qty_obj
as the unit categories could be different.
This is typically the use case with mrp.bom.

That avoid crash
--

### DIFF
--- a/addons/sale_mrp/migrations/9.0.1.0/post-migrate.py
+++ b/addons/sale_mrp/migrations/9.0.1.0/post-migrate.py
@@ -1,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+# TODO: Set delivered_qty in sale order lines with bom
+# See https://github.com/OCA/OpenUpgrade/pull/1410
--- a/addons/sale_stock/migrations/9.0.1.0/post-migrate.py
+++ b/addons/sale_stock/migrations/9.0.1.0/post-migrate.py
@@ -83,6 +83,9 @@ def set_sale_order_qty_delivered(env):
                 AND sm.location_dest_id in %s
             GROUP by sol.id) a
          WHERE sol.id = line_id""", (loc_ids,))
+    # Take only records that have the same products on both move and
+    # sale order lines as unit categories could be different if
+    # bom (phantom) are used.
     env.cr.execute("""\
         SELECT sol.id, sm.id
         FROM sale_order_line sol
@@ -90,6 +93,7 @@ def set_sale_order_qty_delivered(env):
             JOIN stock_move sm ON sm.procurement_id = po.id
         WHERE sm.state = 'done'
             AND sm.product_uom != sol.product_uom
+            AND sm.product_id = sol.product_id
             AND sm.location_dest_id in %s""", (loc_ids,))
     so_qty_map = {}
     uom_obj = env['product.uom']


pedrobaeza:
But AFAIK, this is a forbidden condition. How do you get to that?

rousseldenis:
@StefanRijnhart @pedrobaeza 

This is the minimal code to avoid crash on migrations.

It should trigger a PR on OpenUpgrade/sale_mrp to get the right delivered_qty on sale order lines that depend on boms.
BUT IMHO the _get_delivered_qty on sale order lines should not be called as in Odoo code (sale_mrp), to find the corresponding bom, the computed domain is based on current date compared to date_start/date_end bom's which could lead to incorrect bom mapping.

rousseldenis:
@pedrobaeza You can make phantom boms with the base product unit which is 'Unit()' composed by products with units as 'meter' or ....

pedrobaeza:
I think this is not the problem. The problem comes when you have changed the UoM of the product and you have already stock.move with previous UoM of incompatible type, which was not forbidden previously, but I don't see this use case.

pedrobaeza:
> You can make phantom boms with the base product unit which is 'Unit()' composed by products with units as 'meter' or ....

OK, then we have 2 separate cases, but I didn't know that the `qty_received` was correctly handled when having kit BoMs.

rousseldenis:
> OK, then we have 2 separate cases, but I didn't know that the qty_received was correctly handled when having kit BoMs.

@pedrobaeza Didn't catch your  comment

rousseldenis:
> I think this is not the problem. The problem comes when you have changed the UoM of the product and you have already stock.move with previous UoM of incompatible type, which was not forbidden previously, but I don't see this use case.

You cannot change the unit category of an existing product. You have to deactivate it and create new one

pedrobaeza:
Sorry, I meant `qty_delivered`. I have seen now the proper code in

https://github.com/odoo/odoo/blob/ef444da57a0e5b3a8426a6fac8cef3f6d856b148/addons/sale_mrp/sale_mrp.py#L33-L61

pedrobaeza (approved):
Agree with the change, although you should add:

* the TODO in sale_mrp migration script
* a comment before this query for indicating the reason of including the product_id comparison

pedrobaeza:
> You cannot change the unit category of an existing product. You have to deactivate it and create new one

That happened on the middle of 8.0 life, but there's a lot of DBs with already this problem. Any way, what I do is to fix the DB before migration homogenizing uom across all records (not ideal solution indeed).

rousseldenis:
@pedrobaeza Ok. I just wanted to say that this PR is not enough to avoid delivered_qty wrong computation on migrations (for boms). Just avoid crash with the use case described.

rousseldenis:
> Agree with the change, although you should add:

> the TODO in sale_mrp migration script
> a comment before this query for indicating the reason of including the product_id comparison

Ok, I'll do it

rousseldenis:
@pedrobaeza Done

MiquelRForgeFlow (approved):

AaronHForgeFlow:
I don't understand. Will the delivered_qty will not be computed for the BOM components only?

rousseldenis:
@aheficent For the cases (I don't know all the implementations) if products are not the same in sale order lines and outgoing stock moves.

I think the actual computation is wrong if bom are used (it does not crash with same unit categories) as the query is a simplification of existing use cases.

AaronHForgeFlow (approved):
@rousseldenis ok. It will still compute the delivered qty for the products in the so line so :+1:

PR #1419, opened 2018-06-30
[FIX+IMP] stock: Several refinements

pedrobaeza:
* Error in comparison in create_specific_procurement_rules_from_globals
* Compute stock.move.reference & stock.picking.scheduled_date through SQL
* Add logging trace for each post-migration method

@Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -1,10 +1,51 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2017-2018 Tecnativa - Pedro M. Baeza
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
I think you also should update the `openupgrade_analysis_work.txt` file

pedrobaeza:
That fields are not present there, as there were before. The change has been in their `store` attribute. Maybe we should log that change also in `openupgrader_records` module.

MiquelRForgeFlow:
Yes, I think that should be logged. In fact, I thought that it was already logging that.

pedrobaeza:
But that's something out of the scope of this PR.

MiquelRForgeFlow:
Of course.

MiquelRForgeFlow (approved):
LGTM

cubells (approved):
:+1: 

LGTM too

PR #1421, opened 2018-07-02
[IMP] purchase: Compute fields through SQL

pedrobaeza:
This saves a lot of time.

@Tecnativa

### DIFF
--- a/addons/purchase/migrations/11.0.1.2/post-migration.py
+++ b/addons/purchase/migrations/11.0.1.2/post-migration.py
@@ -27,9 +27,72 @@ def update_procurement_fields(env):
     )
 
 
+@openupgrade.logging()
+def fill_purchase_order_picking_fields(env):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
This one is ready to review. I have introduced in this version the same hack that @NL66278 put for v9 for blacklisting fields to recompute in order to get columns/m2m tables created by ORM, but compute fields by SQL.

MiquelRForgeFlow (approved):
But I don't like this kind of blacklist hack

pedrobaeza:
> But I don't like this kind of blacklist hack

Why not? I have got to it because on many2many fields, there are a lot of SQL operations for getting the exact result as with ORM: create tables, indexes, foreign keys, comments... And at the end, it's better to simply not recomputing records.

StefanRijnhart (approved):
Thank you. Openupgrade needs this kind of code to become performant.

Issue #1308, opened 2018-04-12
[9.0][delivery] post migration issue


florian-dacosta:
Hello,
I have an issue when migrating the delivery module from v8 to 9 with Openupgrade.
I think it is a bug, but the code is 2 years old, which make me wonder why nobody would have meet the same issue... (since it is a very used module)
Here : https://github.com/OCA/OpenUpgrade/blame/9.0/addons/delivery/migrations/9.0.1.0/post-migration.py#L45 , The field `property_delivery_carrier` is renamed to `property_delivery_carrier_id` in the post-migration field. I have an error, since the field `property_delivery_carrier_id` is already created during the update of the module. 
I guess it should be in the pre-upgrade script to avoid this problem. 
This comment seems to go this way too : https://github.com/OCA/OpenUpgrade/blame/9.0/addons/delivery/migrations/9.0.1.0/post-migration.py#L11

I'd like to know the opinion of the authors of this script if possible!
What do you think? @pedrobaeza @hbrunn

pedrobaeza:
Being a property, it's the same to do it on pre or post right now, as the values are in other table. This can only affect to default values.

florian-dacosta:
Thanks for your answer Pedro.
Are you sure it is the same?
Old property is property_delivery_carrier.
New one is property_delivery_carrier_id.

So, on update of the module in version 9, Odoo will create a `ir.model.fieds`, with name `property_delivery_carrier_id`.
Then, on post upgrade https://github.com/OCA/OpenUpgrade/blame/9.0/addons/delivery/migrations/9.0.1.0/post-migration.py#L13 Odoo will rename old `ir.model.fields property_delivery_carrier` to `property_delivery_carrier_id`.
Doing this, Odoo fails because there is an SQL constraint so we can't have 2 ir.model.fields for same name, same model.

The problem here is not a loss of data, it is just a failure that will block openupgrade.

StefanRijnhart:
Makes sense, @florian-dacosta. Could you do propose a fix? You can use https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1672

pedrobaeza:
OK, it's weird, as I have migrated several DBs without hitting the problem, but OK to do it that part on pre-migration. We have now a new method in openupgradelib called `rename_fields` that you can use for this purpose.

pedrobaeza:
Any news about this?

pedrobaeza:
Closing as no answer

[...]

PR #1427, opened 2018-07-03
[9.0][FIX] Switch property rename from post migration to pre migration script

florian-dacosta:
Pull Request to fix the problem described here : https://github.com/OCA/OpenUpgrade/issues/1308

I dug into the problem to understand why I had this bug when nobody else seemed to have it...
And here is the reason. It is a migration from 7 to 9 and the module base_custom_attributes add a sql constraint on the name column of ir_model_fields : https://github.com/OCA/product-attribute/blob/7.0/base_custom_attributes/ir_model.py#L37

That is why Openupgrade failed for my database.
For other, the constraint does not exist, so OpenUpgrade allow to have a duplicate ir_model_fields.

I juste made a verification on a recenttly migrated database of a colleague, and the database contains 2 records of ir.model.fields with name `property_delivery_carrier_id` (and `model` `res.partner`).
Which is only possible because Openupgrade by pass the orm.

@pedrobaeza @StefanRijnhart 

--

### DIFF
--- a/addons/delivery/migrations/9.0.1.0/post-migration.py
+++ b/addons/delivery/migrations/9.0.1.0/post-migration.py
@@ -6,26 +6,6 @@
 from openupgradelib import openupgrade
 
 
-def rename_property(cr, model, old_name, new_name):
-    """Rename property old_name to new_name. This should happen in a
-    pre-migration script."""
-    # TODO: propose this to openupgradelib
-    cr.execute(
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

bealdav (approved):

StefanRijnhart:
Thanks!

PR #1428, opened 2018-07-04
[FIX] Purchase requisition setting on products

StefanRijnhart:
[FIX] Purchase requisition setting on products
The previously used method openupgrade.map_values currently does not support
mapping away from boolean values as these are mapped to 'is (not) set'.

--


pedrobaeza:
Can you please check Travis error?
```
TypeError: argument of type 'SQL' is not iterable
```

MiquelRForgeFlow (approved):

StefanRijnhart:
I gave up on psycopg2 sql composibles for now. @pedrobaeza now green.

pedrobaeza (approved):

PR #1432, opened 2018-07-05
[MIG] l10n_be: tax tags from tax codes

sbidoul:
The current migration of taxes on move lines does not work for l10n_be (and other localizations - see for instance #1172)

This one works for l10n_be and is quite generic, so should work for other localizations.

In a nutshell, it does the following:
* it uses a CSV file that maps tax codes to tax tags, tells if a code is for base or tax, and tells if a positive tax_amount is debit or credit
* from that it creates an inactive tax named "Mig Code {code}"
* it uses these taxes in the migrated move lines

--

### DIFF
--- a/addons/l10n_be/migrations/9.0.1.1/post-migration.py
+++ b/addons/l10n_be/migrations/9.0.1.1/post-migration.py
@@ -13,11 +13,6 @@ _logger = logging.getLogger(__name__)
 
 
 MIG_TAX_LINE_PREFIX = "[8 to 9 tax migration] "
-
-# Set this to True if you want to attempt to reuse existing taxes
-# on move lines. If False, a new inactive tax is created for each
-# tax code used on move lines.
-MIG_TAX_REUSE = False
 MIG_TAX_PREFIX = "Mig Code "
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
For activating tests, you have to put on the documentation the module as `Done`. Is it worth to put the general methods elsewhere (on account for example)?

sbidoul:
If you think this method (ie mapping old account tax code to new account tags) is useful for other localizations, I can move the method to account, yes.

pedrobaeza:
Well, I don't know if it can be useful for others, but it's probably, don't you think?

rousseldenis (approved):
Tested.

sbidoul:
@pedrobaeza where is the best place for such a reusable method? `account/migrations` is not importable.

StefanRijnhart:
@sbidoul we don't have it at this point, but conventionally you can create `openerp/openupgrade/openupgrade_90.py` for that.

StefanRijnhart:
BTW tests are fixed here: https://github.com/OCA/OpenUpgrade/pull/1434

pedrobaeza:
Please rebase and do the documentation stuff

sbidoul:
@pedrobaeza @StefanRijnhart I'm now quite satisfied with this algorithm. I encourage you to give it a try.

In short
* you need this table mapping tax codes to tax tags, which should be quite easy to create for each localization
* by default it creates one new inactive tax for each tax code, to ensure the result is both understandable and preserves the vat reports (there is a flag to force it to try and reuse existing taxes, and fall back to creating new taxes if there is an ambiguity)
* it goes to great length to find out if a move line with a tax_code_id is a base or a tax (by looking at the linked invoice)
* it handles debit=credit=0 lines by adding the tax to the corresponding base move line if it can find it reliably (same amount) and falls back to creating two dummy move lines otherwise
* it does not rely on move line labels
* it logs errors for everything that needs to be examined manually 
* it has a method to reset taxes on move lines so you can run it several times in case of problems

pedrobaeza:
@sbidoul is this going to be fixed?

pedrobaeza:
Closing this for now until further answer.

PR #1439, opened 2018-07-09
[FIX] calendar: Handle recurring events

pedrobaeza:
Solves #1412 

@Tecnativa

### DIFF
--- a/addons/calendar/migrations/11.0.1.0/pre-migration.py
+++ b/addons/calendar/migrations/11.0.1.0/pre-migration.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-# © 2017 bloopark systems (<http://bloopark.de>)
+# Copyright 2017 bloopark systems (<http://bloopark.de>)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This tries to showcase for now the problem. If it arises, then I will add the migration code for avoiding it.

pedrobaeza:
And here we have it!

https://travis-ci.org/OCA/OpenUpgrade/builds/401748434#L2494

Let's go for the solution.

pedrobaeza:
And pre-creating the column everything's back to normal.

@davidtranhp @jonerikceberio you can review this.

StefanRijnhart (approved):

StefanRijnhart:
Waiting for feedback from @davidtranhp @jonerikceberio

jonerikceberio:
Good on my side, although I still do not understand why it wasn't able to 'translate' the fake ids to real ids to write on DB.

davidtranhp:
I confirm that this PR solves the issue

StefanRijnhart:
Thanks for the feedback!

PR #1440, opened 2018-07-10
[10.0][FIX]purchase_requisition fill product

AaronHForgeFlow:
Missing mandatory field in purchase requisition post_migration script

Current behavior before PR:
An error raises asking for categ_id field when creating a product

No error is raised and the product is created for the purchase_requisition line

Solution: adding a default category. Actual migration run successfully after the fix.

--

### DIFF
--- a/addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
+++ b/addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
@@ -10,8 +10,14 @@ def fill_product(env):
     """Fill with an pre-created product as it's required."""
     product = env['product.product'].create({
         'active': False,
+        'categ_id': env.ref('product.product_category_all').id,
         'name': 'OU purchase requisition line wildcard',
     })
+    if product.categ_id.type == 'view':
+        normal = env['product.category'].search(
+            [('type', '=', 'normal')], limit=1)
+        product.categ_id = normal
+
     openupgrade.logged_query(
         env.cr, """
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
ping @mreficent

MiquelRForgeFlow (approved):
LGTM

But it's strange that the default `_get_default_category_id` doesn't execute.

pedrobaeza (changes requested):
Check the code for the default value:

https://github.com/OCA/OpenUpgrade/blob/e1623b0ca8f8663da5b42529da796af952dbaa83/addons/product/models/product_template.py#L23

What I think is that the category "All" has been changed in your DB from normal to view, and that's the reason why it doesn't work, so the product you are creating is going to be inside a view category. You should find the first category that has `type = 'normal'` when default categ is not of that type.

pedrobaeza:
Are you going to change this, @aheficent ?

AaronHForgeFlow:
I am not sure if I understood. Are you proposing this?

1.- check the type of the category.
2.- if the type is view find a type normal and then assign that type.

pedrobaeza:
Exactly

AaronHForgeFlow:
Ok, it makes sense. I'll do that

AaronHForgeFlow:
Closed as solved here: https://github.com/OCA/OpenUpgrade/pull/1637

AaronHForgeFlow:
BTW please look for existing PRs before opening new ones.

PR #1443, opened 2018-07-11
[IMP] sale: Speed up computed fields through SQL

pedrobaeza:
These fields are very heavy to be computed through regular way.

@Tecnativa

### DIFF
--- a/addons/sale/migrations/11.0.1.1/post-migration.py
+++ b/addons/sale/migrations/11.0.1.1/post-migration.py
@@ -102,19 +102,19 @@ def fill_sale_order_line_computed(env):
     # Finally, launch an ORM computation of lines with different invoice
     # currency than sales order one
     env.cr.execute(
-        """SELECT ail.id
-        FROM account_invoice_line ail
+        """SELECT sol.id
+        FROM sale_order_line sol
         INNER JOIN sale_order_line_invoice_rel rel
-            ON rel.invoice_line_id = ail.id
-        INNER JOIN sale_order_line sol
-            on sol.id = rel.order_line_id
+            ON rel.order_line_id = sol.id
+        INNER JOIN account_invoice_line ail
+            on ail.id = rel.invoice_line_id
         WHERE ail.currency_id != sol.currency_id
         """
     )
-    inv_lines = env['account.invoice.line'].browse([
+    so_lines = env['sale.order.line'].browse([
         x[0] for x in env.cr.fetchall()
     ])
-    inv_lines._compute_invoice_amount()
+    so_lines._compute_invoice_amount()
 
 
 @openupgrade.migrate()
--- a/addons/sale/migrations/11.0.1.1/post-migration.py
+++ b/addons/sale/migrations/11.0.1.1/post-migration.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
-# © 2017 bloopark systems (<http://bloopark.de>)
+# Copyright 2017 bloopark systems (<http://bloopark.de>)
 # Copyright 2018 Eficent Business and IT Consulting Services S.L.
+# Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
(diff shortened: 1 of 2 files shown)


addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -8,9 +9,64 @@
...
+    # We have to do this second query that reduces the invoiced amount due to
+    # the refunded amount as the computation mixes several tables that can't
+    # be combined together
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET amt_invoiced = sol.amt_invoiced - sub.amount
+        FROM (
+            SELECT rel.order_line_id,
+                SUM(ai2.amount_total) AS amount
```
StefanRijnhart: I see you correctly emulate the bug https://github.com/odoo/odoo/pull/25417! Maybe you can add a comment about that.

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -8,9 +9,64 @@
...
+    openupgrade.logged_query(
+        env.cr, "UPDATE sale_order_line sol SET amt_invoiced = 0.00",
+    )
+    # Then we set the quantity for the order lines with related invoice lines
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET amt_invoiced = sub.amount
+        FROM (
+            SELECT rel.order_line_id, SUM(ail.price_total) AS amount
```
StefanRijnhart: Can you do a CASE switch to use the alternative amount in case of delivery products? https://github.com/odoo/odoo/blob/11.0/addons/sale/models/sale.py#L743

pedrobaeza:
Waiting until upstream PR odoo/odoo#25417 or odoo/odoo#26054 get resolved for changing the algorithm.

pedrobaeza:
@StefanRijnhart I have adjusted query to match current more polite upstream code, and also cover the "delivery" invoicing policy. Please check.

pedrobaeza:
@StefanRijnhart this one is green and ready to be merged.

JordiBForgeFlow:
@mreficent can you please try to replicate the migration using this PR?

StefanRijnhart (approved):

StefanRijnhart:
Thanks @pedrobaeza. Because there are no tests in this PR validating the results, I'd love to hear back from @mreficent before merging.

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -8,9 +9,77 @@
...
+    # discounting refunded invoiced. We apply a fixed rounding for avoiding
+    # resolution problems. It also uses a trick using the same table 2 times
+    # for avoiding to perform a subquery.
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET amt_to_invoice = ROUND(
+            (CASE WHEN COALESCE(pt.invoice_policy, 'x') = 'delivery'
+             THEN sol.price_reduce_taxinc * sol.qty_delivered
+             ELSE sol.price_total - sol.amt_invoiced END), 5
```
MiquelRForgeFlow:
```suggestion
             ELSE sol.price_total END) - sol.amt_invoiced, 5
```

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -8,9 +9,77 @@
...
+        )
+        FROM sale_order_line sol2
+        LEFT JOIN product_product pp ON pp.id = sol2.product_id
+        LEFT JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        WHERE sol2.id = sol.id""",
```
MiquelRForgeFlow:
```suggestion
        WHERE sol2.id = sol.id AND sol.state IN ('sale', 'done')""",
```

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -8,9 +9,77 @@
...
+                INNER JOIN account_invoice_line ail
+                    ON rel.invoice_line_id = ail.id
+                INNER JOIN account_invoice ai ON ai.id = ail.invoice_id
+                INNER JOIN account_invoice ai2 ON ai2.refund_invoice_id = ai.id
+            WHERE ai.state IN ('open', 'paid')
+                AND ai2.state IN ('open', 'paid')
+                AND ail.product_id = sol.product_id
+            GROUP BY sol.id
+        ) AS sub
+        WHERE sol.id = sub.id""",
```
MiquelRForgeFlow, line 100:
```sql
UPDATE sale_order_line sol
SET amt_invoiced = sol.amt_invoiced - sub.amount
FROM (
	SELECT rel.order_line_id, SUM(ail.price_total) AS amount
	FROM sale_order_line_invoice_rel rel
	LEFT JOIN sale_order_line sol2 ON sol.id = rel.order_line_id
	INNER JOIN account_invoice_line ail ON
	    rel.invoice_line_id = ail.id
	INNER JOIN account_invoice ai ON ai.id = ail.invoice_id
        INNER JOIN account_invoice ai2 ON ai2.refund_invoice_id = ai.id
    WHERE ai.state IN ('open', 'paid')
	AND ai2.state IN ('open', 'paid')
	AND ail.product_id = sol2.product_id
    GROUP BY rel.order_line_id
) AS sub
WHERE sol.id = sub.order_line_id
```

addons/sale/migrations/11.0.1.1/post-migration.py
MiquelRForgeFlow: now is line https://github.com/odoo/odoo/blob/11.0/addons/sale/models/sale.py#L763, and yes, it's done in https://github.com/OCA/OpenUpgrade/pull/1443/files#diff-2e401a61b8a9480b8995b85d4c21e9b0R43

marked as resolved

AaronHForgeFlow:
Conflicts here, @pedrobaeza can you rebase?

MiquelRForgeFlow:
(and attend comments)

rafaelbn:
Please @pedrobaeza revolve conflicts thanks

pedrobaeza:
I have fixed conflicts and push a new version with hopefully everything attended. @mreficent I have seen your comments too late, and do the things by myself. Please check if everything's correct.

addons/sale/migrations/11.0.1.1/post-migration.py
MiquelRForgeFlow: I don't remember why I put you this query. It seems both do the same 😅

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -19,9 +20,107 @@ def activate_proforma(env):
...
+    # before discounting refunded invoiced. We apply a fixed rounding for
+    # avoiding resolution problems. It also uses a trick using the same table
+    # 2 times for avoiding to perform a subquery.
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET amt_to_invoice = ROUND(
+            (CASE WHEN COALESCE(pt.invoice_policy, 'x') = 'delivery'
+             THEN sol.price_reduce_taxinc * sol.qty_delivered
+             ELSE sol.price_total END) - sol.amt_invoiced END, 5
```
MiquelRForgeFlow:
```suggestion
             ELSE sol.price_total END) - sol.amt_invoiced, 5
```

addons/sale/migrations/11.0.1.1/post-migration.py
pedrobaeza: Hehe, let's forget it then!

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -19,9 +20,107 @@ def activate_proforma(env):
...
+            AND sol.state IN ('sale', 'done')
+        """,
+    )
+    # We have to do this second query that reduces the invoiced amount due to
+    # the refunded amount as the computation mixes several tables that can't
+    # be combined together. The computation is not exact, as we reduce amount
+    # for all lines with same product, but it's not even correct on upstream,
+    # as there's no real link.
+    # See https://github.com/odoo/odoo/pull/26054#pullrequestreview-143472733
+    openupgrade.logged_query(
```
MiquelRForgeFlow, line 81: Some time ago, Odoo added this commit https://github.com/odoo/odoo/commit/c05abb2db837138ef784034aa8bab2fe2e014206. Should we do a change in the query?

addons/sale/migrations/11.0.1.1/post-migration.py
pedrobaeza: Ouch, fixed!

addons/sale/migrations/11.0.1.1/post-migration.py
pedrobaeza: Well, there's a slight difference in the computation algorithm through SQL and through ORM: if there are multiple lines with the same product, with SQL, the amount is discounted on each line while in ORM only once, but it's not possible to fine-tune better and it's a very edge case. What do you think?

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -19,9 +20,107 @@ def activate_proforma(env):
...
+            ON rel.invoice_line_id = ail.id
+        INNER JOIN sale_order_line sol
+            on sol.id = rel.order_line_id
+        WHERE ail.currency_id != sol.currency_id
+        """
+    )
+    inv_lines = env['account.invoice.line'].browse([
+        x[0] for x in env.cr.fetchall()
+    ])
+    inv_lines._compute_invoice_amount()
```
MiquelRForgeFlow: You should extract the sale order lines instead, not the invoice lines. This method is from the so lines.

addons/sale/migrations/11.0.1.1/post-migration.py
MiquelRForgeFlow: Ok! :+1:

addons/sale/migrations/11.0.1.1/post-migration.py
pedrobaeza: Thanks for fixing this. I'm thinking that we don't need to link `sale_order_line` table, as we already have the ID in sale_order_line_invoice_rel.order_line_id.

Would you?

MiquelRForgeFlow: hahaha

MiquelRForgeFlow: but you need the sale_order_line table for the currency!!!

pedrobaeza: Ah, right, my first intention was correct then, as I haven't taken that into account this second sight

pedrobaeza:
Let's merge it then if everything correct

PR #1446, opened 2018-07-11
[IMP] stock: drop slow constraint

pedrobaeza:
Removing this constraint that doesn't affect new data structure, we get tons of more performance on quant removal.

Credits go to @carlosdauden, who discovered the bottleneck.

@Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/pre-migration.py
+++ b/addons/stock/migrations/11.0.1.1/pre-migration.py
@@ -38,8 +38,9 @@ def delete_quants_for_consumable(env):
 
 
 def drop_slow_constraint(env):
-    """Removing this constraint, that doesn't affect new data structure, we
-    get tons of more performance on quant removal.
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/11.0.1.1/pre-migration.py
```diff
@@ -37,10 +37,21 @@ def delete_quants_for_consumable(env):
     )
 
 
+def drop_slow_constraint(env):
+    """Removing this constraint, that doesn't affect new data structure, we
+    get tons of more performance on quant removal.
```
StefanRijnhart: Please add to the comment that the reason this does not affect the data model is that it belongs to an obsolete model.

pedrobaeza: Changed. You can squashed on merge for removing the extra commit as I edited by UI.

StefanRijnhart (approved):

PR #1447, opened 2018-07-11
[FIX] Prefer 'move' rules over buy and manufacture

StefanRijnhart:
and [IMP] Call native Odoo method if no 'move' rule is found

Currently in Openupgrade, make_to_stock procurements for moves from stock to
customer location could get assigned a manufacture action if mrp was installed, which does not make sense regardless of the supply method because a make_to_stock procurement should always result in a move.

Also, group procurements and look up the rule per group instead of iterating over all pending procurements.

Backport the deferred tests mechanism and add a deferred test for this.

--

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false
 
 python:
   - "2.7"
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):
Thanks for the tests that makes this more reliable, and the great work adding the testing mechanism

PR #1449, opened 2018-07-16
[ADD] write account.account.type#type for noupdate records

hbrunn:
those seem to have not yet been different when the script was run

### DIFF
--- a/addons/account/migrations/9.0.1.1/noupdate_changes.xml
+++ b/addons/account/migrations/9.0.1.1/noupdate_changes.xml
@@ -72,5 +72,17 @@
       <field name="note">Payment term: 30 Net Days</field>
       <field name="line_ids" eval="[(5, ), (0, 0, {'value': 'balance', 'value_amount': 0.0, 'sequence': 500, 'days': 30, 'option': 'day_after_invoice_date'})]"/>
     </record>
+    <record model="account.account.type" id="data_account_type_receivable">
+      <field name="type">receivable</field>
+      <field name="include_initial_balance" eval="True"/>
+    </record>
+    <record model="account.account.type" id="data_account_type_payable">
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
I'm looking at a migrated database right now and the only value that I don't have is data_account_type_liquidity's type. Let me have a closer look.

hbrunn:
watch out for the effect of https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/post-migration.py#L293 - took a while for me to understand that in my case, the types were inconsistent so in the first go, the migration would create new report types with the expected internal types

StefanRijnhart:
I guess I was lucky with the russian roulette that is https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/post-migration.py#L301-L324. You gotta wonder though which account in both our setups triggered setting an incorrect type on the account type. Apart from that though, this does not hurt.

StefanRijnhart:
Travis
```
You cannot have a receivable/payable account that is not reconciliable. (account code: 451055)
None" while parsing /home/travis/build/OCA/OpenUpgrade/addons/account/migrations/9.0.1.1/noupdate_changes.xml:75, near
```

hbrunn:
hmmm. For consistency, setting affected accounts to `reconcile=True` seems fine to me. Any complaints?

StefanRijnhart:
No, that's fine.

pedrobaeza:
Travis seems to be red.

pedrobaeza:
What happens with this PR?

StefanRijnhart:
@hbrunn care to fix this PR according to your last comment?

pedrobaeza:
Closing this until further answer

PR #1454, opened 2018-07-19
 [MIG] account: set default date_maturity to date 

### LINKED ISSUE #1437: 8 to 9: date_maturity required on account move line
Impacted versions:

8 to 9

Steps to reproduce:

After migration, the not null constraint on date_maturity cannot be set.

Should the migration script set date_maturity = date if not set?

sbidoul:
closes #1437

--

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -310,12 +310,13 @@ def merge_supplier_invoice_refs(env):
     )
 
 
-def set_date_maturity(cr):
-    cr.execute("""
+def set_date_maturity(env):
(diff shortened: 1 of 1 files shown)


addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -310,6 +310,14 @@ def merge_supplier_invoice_refs(env):
     )
 
 
+def set_date_maturity(cr):
+    cr.execute("""
```
pedrobaeza: Please use `openupgrade.logged_query` for logging the operation and know how much rows this affects

sbidoul:
done

pedrobaeza (approved):

PR #1456, opened 2018-07-20
[WIP] 9.0 Migration preupgrade checks CLI command

dreispt:
This adds an Odoo CLI command to check what modules are missing migration scripts.
To run: `./odoo.py preupgrade -d <dbname>`

Addons directories can include `*.migrate` files documenting the modules to ignore or renamed.
Such a file is provided for the core addons, based on the modules marked as "nothing to do" in #400.

May have some rough edges to improve, but it works and is good to review.


StefanRijnhart:
Thanks. This PR implies that we need to keep track of module migration status in two places, in the *.migrate files and in the openupgrade coverage documentation. Isn't it just as easy for you, and easier to mainain, if you display the comment on the matching row of the rst doc (just fetch it live from github). Maybe you can decouple this from Odoo, and add it as a standalone script to openupgradelib. You can fetch the list of installed modules in SQL, and determine the Odoo release number from the version of the base module. The script can then be used across Odoo releases instead of having to be ported from release to release.

dreispt:
I was not aware of the Openupgrade coverage documentation. You mean this?:
https://doc.therp.nl/openupgrade/_sources/modules100-110.txt

I can't find these docs sources. 
Shouldn't they be on a branch of the OCA/Openupgrade repo?

pedrobaeza:
https://github.com/OCA/OpenUpgrade/blob/11.0/odoo/openupgrade/doc/source/modules100-110.rst

StefanRijnhart:
Oh sorry for not adding a link. The coverage of each release is kept in its own source tree. See how the documentation build script fetches the correct file for each release: https://github.com/OCA/OpenUpgrade/blob/11.0/scripts/build_openupgrade_docs

pedrobaeza:
Closing this as the approach is not the correct as stated. You can open a new PR when switching.

PR #1458, opened 2018-07-25
[9.0][procurement] remove the value of 'src_model' from view 'action_procurement_compute',

JordiBForgeFlow:
Having value 'procurement.order' value in 'src_model' field of view  'action_procurement_compute' will cause errors during migration to 11.0. We have to force this removal because the view definition from 9.0
omits the field 'src_model'.

--

### DIFF
--- a/addons/procurement/migrations/9.0.1.0/noupdate_changes.xml
+++ b/addons/procurement/migrations/9.0.1.0/noupdate_changes.xml
@@ -1,0 +1,15 @@
+<?xml version='1.0' encoding='utf-8'?>
+<openerp>
+  <data>
+        <!-- Force removal of values in src_model, as it is not needed in 9.0.
+        Having this field completed causes issues during migration in 11.0.-->
+        <record id="action_procurement_compute" model="ir.actions.act_window">
+            <field name="name">Run Reordering Rules</field>
+            <field name="res_model">procurement.orderpoint.compute</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form</field>
+            <field name="target">new</field>
+            <field name="src_model"></field>
+        </record>
+  </data>
+</openerp>
--- a/addons/procurement/migrations/9.0.1.0/pre-migration.py
+++ b/addons/procurement/migrations/9.0.1.0/pre-migration.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# © 2018 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+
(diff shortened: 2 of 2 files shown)


addons/procurement/migrations/9.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,14 @@
...
+  <data>
+        <!-- Force removal of values in src_model, as it is not needed in 9.0.
+        Having this field completed causes issues during migration in 11.0.-->
+        <record id="action_procurement_compute" model="ir.actions.act_window">
+            <field name="name">Run Reordering Rules</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form</field>
+            <field name="target">new</field>
+            <field name="src_model"></field>
+        </record>
```
mreficent: `null value in column "res_model" violates not-null constraint`

MiquelRForgeFlow:
@jbeficent 

I don't understand this PR. The record with id `action_procurement_compute` was in v7, but it was deleted in v8.

The error in travis says that `procurement.orderpoint.compute` model doesn't exist. And it's true, it was deleted in v8.

pedrobaeza:
FYI, I have fixed the problem on the final stage in ad1c2b9413f007d4b4b3409afe48d95d6164c28c

MiquelRForgeFlow:
@pedrobaeza shouldn't your fix be done in previous versions?

pedrobaeza:
The problem is you may have already migrated DBs with incorrect field filled, so we should take care when it fails.

MiquelRForgeFlow:
Yeah, but there are people that still don't have migrated to v11 and maybe are still in v7. The fix should be backported I think.

pedrobaeza:
Yes, I agree

MiquelRForgeFlow:
Should I backport it? 🤔

pedrobaeza:
OK, but please check the proper version

MiquelRForgeFlow:
Ah, I see what happened.

In v7, the action was in  `procurement` module with `src_model="procurement.order"`.
But in v8 the same action exists, but in `stock` module, and maintaining the `src_model="procurement.order"`.
And then in v9, v10 and v11, the same action (still in `stock`) appears without `src_model`.

So the proper version where to put the fix is v9.

MiquelRForgeFlow:
Shouldn't we make a PR to Odoo to add `<field name="res_model"/>`?

pedrobaeza:
For sure they won't accept it.

MiquelRForgeFlow:
They should at least accept it in `master` 😅

pedrobaeza:
Well, you can try it, but the problem will arise only on migrations, so I think it's better to just limit this to be fixed on OpenUpgrade

MiquelRForgeFlow:
I made both things. 😎

MiquelRForgeFlow:
Closed by https://github.com/OCA/OpenUpgrade/pull/1498.

PR #1466, opened 2018-08-02
Add/11.0/merge upstream ec7109c9

StefanRijnhart:
Merging upstream Odoo up until commit ec7109c, mainly to include @hbrunn's odoo/odoo@65ad606.

I had to update the noupdate changes in l10n_de_skr03/4 because the original data had become invalid after https://github.com/odoo/odoo/commit/a3bc08d9d8aa40e8d41f99e5b102997c6990396a

Only conflict:

```
--- a/odoo/modules/migration.py
+++ b/odoo/modules/migration.py
@@@ -18,12 -18,19 +18,25 @@@ from odoo.tools import pycompa
  if pycompat.PY2:
      import imp
      def load_script(path, module_name):
 +        # OpenUpgrade edit start:
 +        # Removed a copy of migration script to temp directory
 +        # Replaced call to load_source with load_module so frame isn't lost and breakpoints can be set
          fp, fname = tools.file_open(path, pathinfo=True)
++<<<<<<< HEAD
++=======
+         fp2 = None
+
+         # pylint: disable=file-builtin,undefined-variable
+         if not isinstance(fp, file):
+             # imp.load_source need a real file object, so we create
+             # one from the file-like object we get from file_open
+             fp2 = os.tmpfile()
+             fp2.write(fp.read())
+             fp2.seek(0)
+
++>>>>>>> 06de947
          try:
 -            return imp.load_source(module_name, fname, fp2 or fp)
 +            return imp.load_module(module_name, fp, fname, ('.py', 'r', imp.PY_SOURCE))
          finally:
              if fp:
                  fp.close()
```

Conflict caused by https://github.com/odoo/odoo/commit/9f8eae34d1d89d7cf01eef8a1a89dc431ef5f4e2#diff-6d3fdea7af4358946ee202d25095b776

I took the opportunity to readd the lines that were removed in OpenUpgrade, but in a disabled way. Hope that this leads to less conflicts in this section in the future.

Diff after resolving conflict:

```
--- a/odoo/modules/migration.py
+++ b/odoo/modules/migration.py
@@ -18,15 +18,28 @@ from odoo.tools import pycompat
 if pycompat.PY2:
     import imp
     def load_script(path, module_name):
-        # OpenUpgrade edit start:
-        # Removed a copy of migration script to temp directory
-        # Replaced call to load_source with load_module so frame isn't lost and breakpoints can be set
         fp, fname = tools.file_open(path, pathinfo=True)
+        fp2 = None
+
+        # OpenUpgrade edit start:
+        # Don't copy the migration script into a temp directory. Instead,
+        # replace the call to load_source with a call to load_module so  that
+        # the frame isn't lost and breakpoints can be set
+        # pylint: disable=file-builtin,undefined-variable
+        if False and not isinstance(fp, file):
+            # imp.load_source need a real file object, so we create
+            # one from the file-like object we get from file_open
+            fp2 = os.tmpfile()
+            fp2.write(fp.read())
+            fp2.seek(0)
+
         try:
             return imp.load_module(module_name, fp, fname, ('.py', 'r', imp.PY_SOURCE))
         finally:
             if fp:
                 fp.close()
+            if fp2:
+                fp2.close()
```

The full diff of this section with core Odoo 11.0 is now:

```
>         # OpenUpgrade edit start:
>         # Don't copy the migration script into a temp directory. Instead,
>         # replace the call to load_source with a call to load_module so  that
>         # the frame isn't lost and breakpoints can be set
29c25
>         if False and not isinstance(fp, file):
---
<         if not isinstance(fp, file):
37c33
>             return imp.load_module(module_name, fp, fname, ('.py', 'r', imp.PY_SOURCE))
---
<             return imp.load_source(module_name, fname, fp2 or fp)
```


StefanRijnhart:
@christophlsa can you review the changes to the migration of the German localization? See https://github.com/odoo/odoo/commit/a3bc08d9d8aa40e8d41f99e5b102997c6990396a. We found out because the original data broke after this merge:
```
2018-08-01 08:27:59,986 4954 INFO openupgrade OpenUpgrade: l10n_de_skr04: loading migrations/11.0.2.0/noupdate_changes.xml
2018-08-01 08:27:59,990 4954 ERROR openupgrade OpenUpgrade: l10n_de_skr04: error in migration script l10n_de_skr04/migrations/11.0.2.0/post-migration.py: "External ID not found in the system: l10n_de_skr04.account_3300" while parsing /home/travis/build/OCA/OpenUpgrade/addons/l10n_de_skr04/migrations/11.0.2.0/noupdate_changes.xml:3, near
<record id="l10n_chart_de_skr04" model="account.chart.template">
    <field name="property_account_payable_id" ref="account_3300"/>
    <field name="property_account_receivable_id" ref="account_1200"/>
    <field name="property_account_expense_categ_id" ref="account_5400"/>
    <field name="property_account_income_categ_id" ref="account_4400"/>
    <field name="transfer_account_id" ref="account_1520"/>
    <field name="expense_currency_exchange_account_id" ref="account_6880"/>
    <field name="income_currency_exchange_account_id" ref="account_4840"/>
  </record>
```
I simply regenerated the noupdate data but I can imagine the whole migration needs to be revised.

MiquelRForgeFlow (approved):

pedrobaeza:
We can merge this one although @christophlsa hasn't answered yet, as the build is green and only installations with modules l10n_de_skr* can be affected.

PR #1468, opened 2018-08-06
[ADD] product_variant_supplierinfo: Merged in product

pedrobaeza:
Odoo core now manages vendors for variants in core.

@Tecnativa


StefanRijnhart:
Does the data model of the OCA module correspond one on one with the Odoo functionality?

pedrobaeza:
Yes, there's only a new field in product.supplierinfo: `product_id`, that it's now on core (in fact, it was in v10, but without use). The rest were tricks for making it work. You can check it in https://github.com/OCA/product-variant/tree/10.0/product_variant_supplierinfo/models

pedrobaeza:
OK to go?

StefanRijnhart (approved):
Thanks!

PR #1471, opened 2018-08-07
[9.0][ADD] product_pricelist_fixed_price: Merged in product

pedrobaeza:
There's no need of nothing more as these lines cover the fixed price and at the end, this module adapts the UI for setting that values on background:

https://github.com/OCA/OpenUpgrade/blob/9df238f191237bcacd84b3c01209094b1a1b1f62/addons/product/migrations/9.0.1.2/post-migration.py#L108-L116

@Tecnativa


StefanRijnhart (approved):

pedrobaeza:
Thanks for the quick prompt!

PR #1477, opened 2018-08-10
[FIX] Delete translations of refreshed mail template contents

StefanRijnhart:


addons/purchase/migrations/10.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate
+def migrate(env, version):
+    """ Delete the translations of the email templates that are being
+    overwritten in the noupdate_changes.xml """
+    openupgrade.delete_record_translations(
+        env.cr, 'purchase', [
+            'module_install_notification',
```
pedrobaeza: This one is a "fake" template and it doesn't need to be reset.

StefanRijnhart: Yes I know, but it is included in the noupdate_data.xml so I'm just including it here as well for consistency's sake.

pedrobaeza: OK, as you prefer.

StefanRijnhart: On second thought, I removed both.

addons/purchase/migrations/10.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# coding: utf-8
+# Copyright 2018 Opener B.V. <stefan@opener.amsterdam>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate
```
pedrobaeza: You have forgotten here `()`

pedrobaeza (approved):

PR #1482, opened 2018-08-12
[IMP] Store originating id on converted bom lines

StefanRijnhart:


StefanRijnhart:
I need this because I have a customization to add an attachment to a BoM line and I need the original mrp_bom id to preserve these attachments. Additionally I can propose the query to preserve any attachments of bom lines here as well, although there is no way to add such an attachment in the core Odoo interface.

addons/mrp/migrations/8.0.1.1/post-migrate.py
```diff
@@ -20,6 +20,7 @@
 #
 ##############################################################################
 import logging
+from postgresql.extensions import AsIs
```
pedrobaeza: s/postgresql/psycopg2

pedrobaeza (changes requested):

pedrobaeza:
> Additionally I can propose the query to preserve any attachments of bom lines here as well, although there is no way to add such an attachment in the core Odoo interface.

I feel that we should have an openupgradelib method for preserving "linked" stuff, independently from the UI possibilities:

* attachments
* mail.message 
* Maybe filters?
* Maybe defaults?
* Or maybe a similar technique like the one used when merging quants:

  https://github.com/OCA/openupgradelib/blob/60669f07fc3bea130148e6690b979d612b4eb5da/openupgradelib/openupgrade_merge_records.py#L355-L357

It can have this signature: `move_linked_resources(env, old_id, new_id)`. Then we can apply it for this case or others. Any way, storing the originating ID is a good move for any module on top of this one. Example applied on v11:

* Here I stored the old issue record ID:

  https://github.com/OCA/OpenUpgrade/blob/850a88ee4f8adb892fcf47ffd676095a51f7a056/addons/project/migrations/11.0.1.1/post-migration.py#L109-L133

* And I make use of it for migrating data from `project_issue_sheet`:

  https://github.com/OCA/OpenUpgrade/pull/1475/files#diff-422089833c862d2c2f545ffef6b2f79fR35

StefanRijnhart:
I'm thinking it might be overengineered to have such method, or at least out of scope for this one. Branch is fixed BTW, thanks for noting.

pedrobaeza (approved):

PR #1486, opened 2018-08-14
Revert "[MIG] Patch demo data to enable travis testing."

pedrobaeza:
This reverts commit 539105978f90ae947813623765625e66bdf9993b.

This patch is provoking some integration tests to fail due to the no installation of a chart of accounts (as there are accounts manually created). I'm suspecting this is going to work as is due to latest Odoo patches, so let's see.

@Tecnativa


pedrobaeza:
@StefanRijnhart @hbrunn this is important for me for passing integration tests in middle versions. Is this acceptable for you?

StefanRijnhart:
@NL66278 alternative fix for l10n_be insanity.

StefanRijnhart (approved):

NL66278 (approved):
:+1:  This seems a better solution then the original one from me.

PR #1489, opened 2018-08-15
[11.0][FIX] website: unlisted website page record not linked with new model

karan-dreambits:
Modified query to link ir_ui_view records of website page to new module of website_page. This fix makes sure that the pages which are not listed in menu are also link properly
--

### DIFF
--- a/addons/website/migrations/11.0.1.0/post-migration.py
+++ b/addons/website/migrations/11.0.1.0/post-migration.py
@@ -15,21 +15,21 @@ def fill_website_pages(env):
         INSERT INTO website_page
             (url, view_id, website_indexed, website_published, create_uid,
              create_date, write_uid, write_date)
-        SELECT
-            wm.url, iuv.id, TRUE, TRUE, iuv.create_uid,
+        SELECT COALESCE(wm.url, '/page/' || iuv.name), iuv.id, TRUE, TRUE,
+            iuv.create_uid,
             iuv.create_date, iuv.write_uid, iuv.write_date
         FROM
-            ir_ui_view iuv
-        INNER JOIN
-            website_menu wm ON wm.url = '/page/' || iuv.name AND
-            (wm.website_id = iuv.website_id OR
-             wm.website_id IS NULL OR
-             iuv.website_id IS NULL)
+            website_page wp
+        RIGHT JOIN
+            ir_ui_view iuv ON wp.view_id = iuv.id
         LEFT JOIN
-            website_page wp ON wp.view_id = iuv.id
+            website_menu wm ON wm.url = '/page/' || iuv.name
+            AND (wm.website_id = iuv.website_id
+            OR wm.website_id IS NULL
+            OR iuv.website_id IS NULL)
         WHERE
-            wp.id IS NULL AND
-            iuv.page = true""",
+            wp.id IS NULL
+            AND iuv.page = TRUE""",
     )
 
 


pedrobaeza:
Thanks for the PR. Instead of doing a UNION, just change the order of the JOINs, putting last one with LEFT JOIN the one that links to website_menu table, and when selecting page url, put `COALESCE(wm.url, '/page/')` instead, and this way, we make only one query.

Is `/page/` the correct url for pages without menu then?

karan-dreambits:
@pedrobaeza  /page/ is correct url for pages without menu, atleast for 10.0
Not sure about 11.0 

And l like your idea. Trying the query again with your suggestions

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -15,21 +15,20 @@ def fill_website_pages(env):
...
-            website_menu wm ON wm.url = '/page/' || iuv.name AND
-            (wm.website_id = iuv.website_id OR
-             wm.website_id IS NULL OR
-             iuv.website_id IS NULL)
-        LEFT JOIN
-            website_page wp ON wp.view_id = iuv.id
-        WHERE
-            wp.id IS NULL AND
-            iuv.page = true""",
+	SELECT COALESCE(wm.url, '/page/' || iuv.name), iuv.id, TRUE, TRUE, iuv.create_uid,
```
pedrobaeza: This exceeds the 80 cols limit

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -15,21 +15,20 @@ def fill_website_pages(env):
...
+		iuv.create_date, iuv.write_uid, iuv.write_date
+	FROM
+	    website_page wp
+	RIGHT JOIN
```
pedrobaeza: Why `RIGHT JOIN`? It should be `LEFT JOIN` as well

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -15,21 +15,20 @@ def fill_website_pages(env):
...
+	    ir_ui_view iuv ON wp.view_id = iuv.id
+	LEFT JOIN
+	    website_menu wm ON wm.url = '/page/' || iuv.name
+	    AND (wm.website_id = iuv.website_id
+	    OR wm.website_id IS NULL
+	    OR iuv.website_id IS NULL)
+	WHERE
+	    wp.id IS NULL
+	   AND iuv.page = TRUE"",
```
pedrobaeza: One space more

pedrobaeza (changes requested):

addons/website/migrations/11.0.1.0/post-migration.py
karan-dreambits: I doubt that @pedrobaeza .. LEFT JOIN would just fetch all records of website_page table whereas we need to fetch all records of ir_ui_view table inrespective of them satisfying the join condition of not so RIGHT JOIN seems more appropriate.

Also LEFT JOIN doesn't work ;)

pedrobaeza: OK, I get you. I thought all views had `website.page` record.

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -15,21 +15,21 @@ def fill_website_pages(env):
...
-            website_menu wm ON wm.url = '/page/' || iuv.name AND
-            (wm.website_id = iuv.website_id OR
-             wm.website_id IS NULL OR
-             iuv.website_id IS NULL)
-        LEFT JOIN
-            website_page wp ON wp.view_id = iuv.id
-        WHERE
-            wp.id IS NULL AND
-            iuv.page = true""",
+	SELECT COALESCE(wm.url, '/page/' || iuv.name), iuv.id, TRUE, TRUE,
```
pedrobaeza: Why now this deindentation? Please clean the PR and make only 1 commit with all.

karan-dreambits: I committed via a server where migration is being done so this issues are creeping. I will clean up the PR and commit

pedrobaeza (approved):

karan-dreambits:
Thank you @pedrobaeza for the quick merge :)

PR #1490, opened 2018-08-15
[ADD] sale_journal: Converted to sale_order_type

pedrobaeza:
We convert deprecated Odoo module into OCA one. The rest of the migration is in the module in OCA/sale-workflow#680

@Tecnativa

### DIFF
--- a/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/9.0.1.3/pre-migration.py
@@ -50,7 +50,7 @@ def cleanup_modules(cr):
             ('account_followup', 'account_credit_control'),
             # We convert deprecated Odoo module into OCA one - The rest of
             # the migration is in the module in OCA/sale-workflow
-            ('account_journal', 'sale_order_type'),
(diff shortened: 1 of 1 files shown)


openerp/addons/base/migrations/9.0.1.3/pre-migration.py
```diff
@@ -48,6 +48,9 @@ def cleanup_modules(cr):
         cr, [
             ('account_chart', 'account'),
             ('account_followup', 'account_credit_control'),
+            # We convert deprecated Odoo module into OCA one - The rest of
+            # the migration is in the module in OCA/sale-workflow
+            ('account_journal', 'sale_order_type'),
```
StefanRijnhart: DId you mean `sale_journal`?

pedrobaeza: Of course... what a typo!

pedrobaeza:
Changed by GitHub UI. You can squash on merge.

StefanRijnhart (approved):
Thanks!

PR #1493, opened 2018-08-16
Update 9.0 analysis

StefanRijnhart:


StefanRijnhart:
I recreated the analysis to check the impact of the GDPR private partner functionality that was introduced in March in 9, 10 and 11.0. No work seems to be needed there (I checked the noupdate changes of the base module as well but there were none).

I also fixed a regression in the openupgrade_records module that prevented my standard xmlrpc script from running.

MiquelRForgeFlow:
There seems to be errors in the `openupgrade_general_log.txt` (when puts `nothing has changed in this moduleERROR: ...`)

MiquelRForgeFlow:
Also, an hours ago the v8 branch has been updated from upstream

StefanRijnhart:
@mreficent that is a side effect of the analysis when obsolete modules are being replaced by or merged into external modules which are not in the analysis scope. Apart from the missing newline there is not a lot that can be done about this I'm afraid.

I will recreate the analysis on the updated 8.0 branch when I have the time.

PR #1494, opened 2018-08-17
[FIX][11.0] stock: Several fixes

pedrobaeza:
* Copy state on stock.move.line
* Don't error loading stock.action_procurement_compute 
* WIP: Use pack operation > move link qty

@Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -115,7 +115,17 @@ def set_partially_available_state(env):
 @openupgrade.logging()
 def create_stock_move_line(env):
     """This method creates stock.move.line recreated from old
-    stock.pack.operation records.
+    stock.pack.operation records. These records are created only for done
+    moves, as for those not done, there's no need of creating stock.move.line,
+    as they can be recreated in v11 easily:
+
+    * For outgoing or internal transfers, clicking on "Check availability",
+      which takes current reserved quants information for recreating the
+      information.
+    * For incoming transfers, clicking on validate for accepting full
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is now finished. Text extracted from the main commit:

When having several stock.move with the same product, Odoo grouped them on the same stock.pack.operation record, so use it for determining the quantities on stock.move.line records (at least one per stock.move), will make this quantity to be incorrectly higher.

We also discard stock.pack.operation.lot information, as this doubles the records on the query, and we can recreate that information using the lot of the quant linked in the stock.move.operation.link record.

Tried this query in:

- Incoming transfer without lots
- Incoming transfer with lots
- Outgoing/Internal transfer without lots
- Outgoing/Internal transfer with lots

Finally, we only insert move lines for done moves, as for those not done, there's no need of creating stock.move.line, as they can be recreated in v11 easily:

* For outgoing or internal transfers, clicking on "Check availability", which takes current reserved quants information for recreating the information
* For incoming transfers, clicking on validate for accepting full quantities, or writing the quantity in the lines you want in "Operations" tab. The only drawback is that if you want to use the detailed operation mode, you will have to enter quantity manually.

MiquelRForgeFlow (approved):

pedrobaeza:
@StefanRijnhart what do you think about this?

StefanRijnhart:
Can you add a comment about not creating stock move lines for pending transfers? Then it's fine by me.
- edit -: I mean copy the comments from this PR into the code.

pedrobaeza:
Added comments per @StefanRijnhart's instructions, so I merge.

StefanRijnhart:
Thanks!

PR #1497, opened 2018-08-22
[FIX] Selection value of account.payment.term.line value

StefanRijnhart:
[FIX] Selection value of account.payment.term.line value

This fixes

```
File "/home/odoo/odoo/addons/account/models/account_invoice.py", line 1472, in compute
    if amt:
UnboundLocalError: local variable 'amt' referenced before assignment
```

in migrated databases (when validating invoices).

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -249,6 +249,14 @@ def map_account_tax_template_type(cr):
         table='account_tax_template', write='sql')
 
 
+def map_payment_term_line_value(cr):
+    """ Someone fixed a Flemishism """
+    openupgrade.logged_query(
+        cr,
+        """UPDATE account_payment_term_line
+        SET value = 'percent' WHERE value = 'procent';""")
+
+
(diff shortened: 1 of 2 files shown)


addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -249,6 +249,13 @@ def map_account_tax_template_type(cr):
         table='account_tax_template', write='sql')
 
 
+def map_payment_term_line_value(cr):
+    """ Someone fixed a Flemishism """
+    openupgrade.logged_query(
+        """UPDATE account_payment_term_line
```
pedrobaeza: You have missed the `cr` argument

StefanRijnhart: Fixed now. I tend not to commit these small fixes untested because I rely on Travis, but I don't want it to be more work for you going into the travis build and look around. So if you like, you can just ignore these red PRs of mine.

pedrobaeza: There's not too much problem, but as you prefer.

PR #1498, opened 2018-08-23
[FIX] stock: Don't error loading stock.action_procurement_compute

MiquelRForgeFlow:
Cherry-pick of https://github.com/OCA/OpenUpgrade/pull/1494/commits/ad1c2b9413f007d4b4b3409afe48d95d6164c28c as discussed in https://github.com/OCA/OpenUpgrade/pull/1458.

--

### DIFF
--- a/addons/stock/migrations/9.0.1.1/pre-migration.py
+++ b/addons/stock/migrations/9.0.1.1/pre-migration.py
@@ -28,9 +28,27 @@ field_renames = [
 ]
 
 
+def fix_act_window(env):
+    """Action window with XML-ID 'stock.action_procurement_compute' has
+    set src_model='procurement.order' in v8, but in v9 it doesn't have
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/9.0.1.1/pre-migration.py
```diff
@@ -28,9 +28,28 @@
 ]
 
 
+def fix_act_window(env):
+    """Action window with XML-ID 'stock.action_procurement_compute' has
+    set src_model='procurement.order', and this will provoke an error as
+    new definition doesn't overwrite this field. We empty that value before
+    that happens. The source of this assignation is not clear, but it doesn't
```
pedrobaeza: You can change this text for adapting according new investigations

MiquelRForgeFlow: Done

pedrobaeza (approved):

Issue #1337, opened 2018-05-11
[V9] OpenUpgrade migration V8 to V9 error - NoneType and float


FranciscoBigWings:
Impacted versions: OpenUpgrade from V8 to V9
Steps to reproduce: execute --update all in V9 OpenUpgrade directory pointing to a database in V8
Current behavior: I attach screenshots because I couldnt copy/paste the log file

In this image I want to show what was the script doing before giving the error
![fullsizeoutput_b5b](https://user-images.githubusercontent.com/8320244/39944748-137bdcec-5568-11e8-87df-ed5a45a079a8.jpeg)

In this image I show the first traceback
![fullsizeoutput_b6a](https://user-images.githubusercontent.com/8320244/39944830-68c6a790-5568-11e8-890e-32f7bc08f81c.jpeg)

In this image I show the second traceback, after this I have to shutdown because it doesnt do anything else
![fullsizeoutput_b69](https://user-images.githubusercontent.com/8320244/39944869-878d93f0-5568-11e8-9b12-7baf9818a7d9.jpeg)

I tried to understand the error as I always do, but this time I am too blind or I just dont see the source of the error

Thanks,

pedrobaeza:
It seems incorrect tax configuration, but nothing to do with OpenUpgrade itself.

FranciscoBigWings:
Thank you for your answer Pedro,

What would you recommend to check the problem? I saw that there are products that dont have taxes set, I mean the field is empty, could that be the problem?

Besides that all taxes are managed by l10n_es module which is in its last version 8.0.5.4

I just realized that there is another issue opened with a very similar problem, is it not the same?
https://github.com/OCA/OpenUpgrade/issues/716


Thank you

pedrobaeza:
#716 is about dependency graph, so I don't think it's that. Sorry, but I'm not able to give you more clues as you have to dig on to your data and trigger when it gets the value None.

FranciscoBigWings:
I understand Pedro, thank you

pedrobaeza:
Due to the investigation in OCA/l10n-spain#911, the solution seems to be #1499

FranciscoBigWings:
Thanks Pedro, I will wait for it to be merged

pedrobaeza:
You can test it also

FranciscoBigWings:
Hello, just confirm that I test it and the issue has been solved. Now it works

pedrobaeza:
Great

PR #1499, opened 2018-08-27
[9.0][FIX] account: Switch type='none' to type='percent' + amount=0 in account.tax

pedrobaeza:
Solves #1337 

cc @Tecnativa


MiquelRForgeFlow (approved):
LGTM

PR #1505, opened 2018-08-31
[FIX] mass_mailing: Fix mass mailing list handling

pedrobaeza:
* Doing a mass mailing on lists now points to the 'mail.mass_mailing.list' model instead of the contact one.
* Mass mailings always use a domain for getting the recipients. If you select some mailing lists in UI, on background the mailing domain is changed accordingly. As now mass mailing contacts can belong to multiple lists, we force the computation of the domain for having the correct one.

@Tecnativa


cubells (approved):
:+1: 

LGTM.

Issue #1338, opened 2018-05-12
[9.0][product] Product weight data wrongly migrated


leio:
In 8.0 there was product.template `weight` (Gross Weight) and `weight_net` (Net Weight).
In 9.0 there is product.product `weight` — Gross Weight; help="The weight of the contents in Kg, not including any packaging, etc.".
In 10.0 there is product.product `weight` — Weight; help="The weight of the contents in Kg, not including any packaging, etc."

9.0 had a bug in field string value stating gross weight, but it always meant to be only net weight now, as described in commit cae29df7d984d20c2cb736498b37bfb42545d67a when weight_net got removed:
```
    product
    - [REM] weight_net field. Weight net is removed from the product model.
    Indeed a product should be modeled without its possible packaging data.
    Moreover in the codebase the weight_net field is not really used.
```

The forgotten string change has been fixed by 10.0, but the tooltip (help) was already correctly describing it in 9.0 as well. This is in commit cd11497bd81a8501d5c63589f6ac6c438db5760b but doesn't seem to have reached 9.0 branch.

As-is, migrations lose important data, if data has been properly entered - nett weight into `weight_net` and not `weight` in v7 and v8.
v8 product.template weight_net values should be put into product.product weight, not a pure name by name move with net weight data incorrectly lost instead of gross weight data.

StefanRijnhart:
Agreed! As a solution, do you agree to keep the template's weight as the product weight if weight_net was not entered? Would you like to create the PR?

leio:
I guess that'd make sense, yes. I will not create the PR, as I have not and will not sign any CLAs, sorry.

pedrobaeza:
Why don't you want to sign the CLA?

StefanRijnhart:
That's a shame @leio, I see some good input from your side here on this project so it would be great if you could submit some code every now and then. I'll see what I can do about this one. In the mean time, it would be interesting to hear what your problem is with the OCA CLA in particular.

leio:
Technically it'd be easier for me to submit my code to not worry about the changes being used by my customers under the AGPL-3 license they happen to be licensed at this point. But you refuse to take and integrate such code, so it's a wasted effort more or less. (To be clear, in this case I simply copied all weight_net into weight with SQL prior to migration, so no AGPL-3 derivative code was written). I'd be happy to contribute under AGPL-3+, GPL-2+, LGPL-2.1+ or all of those together, but I do not see a CLA as something acceptable to me, sorry.
[...]

PR #1510, opened 2018-09-04
[9.0] #1338, set weight from weight_net

sergiocorato:
Description of the issue/feature this PR addresses: #1338 

Current behavior before PR: weight_net is not migrated

Desired behavior after PR is merged: weight_net if != 0.0 is migrated in weight

--

### DIFF
--- a/addons/product/migrations/9.0.1.2/pre-migration.py
+++ b/addons/product/migrations/9.0.1.2/pre-migration.py
@@ -125,3 +125,11 @@ def migrate(env, version):
     cr.execute(
         "update product_pricelist set active=False where type='purchase'"
     )
+
+    # if weight_net != 0.0 in product_template, override weight
+    # if not, do nothing
+    openupgrade.logged_query(cr, """
+            UPDATE product_template
+            SET weight = weight_net
+            WHERE weight_net != '0.0'
+            """)


addons/product/migrations/9.0.1.2/pre-migration.py
```diff
@@ -125,3 +125,11 @@ def migrate(env, version):
     cr.execute(
         "update product_pricelist set active=False where type='purchase'"
     )
+
+    # if weight_net != 0.0 in product_template, override weight
+    # if not, do nothing
+    openupgrade.logged_query(cr, """
+            UPDATE product_template
+            SET weight = weight_net
+            WHERE weight_net != '0.0'
```
StefanRijnhart, line 134: I find examples in one of my customer's database where weight_net is null. Can you exclude these as well?

sergiocorato: mmm, adding
<code> and weight_net is not Null </code>
give me the same amount of record, and I have many records null

StefanRijnhart: Sorry, I did not realize that NULL != 0.0 evaluates to `NULL` instead of `true`. That means that null values are ignored anyway and you can ignore my comment.

StefanRijnhart (approved):

StefanRijnhart (commented):
Thanks!

PR #1515, opened 2018-09-06
[FIX] mass_mailing: Rename list_id column in contact

pedrobaeza:
Odoo uses an SQL for getting the number of contacts using the m2m table, which has a column named `list_id` for referencing the mailing list. Keeping old list_id column in mass mailing contact, provokes an ambiguous column error.

This renames the column the usual OpenUpgrade way.

cc @Tecnativa


MiquelRForgeFlow (approved):
LGTM

AaronHForgeFlow (approved):

PR #1519, opened 2018-09-11
[FIX+IMP] sale_timesheet: Map track_service to service_tracking + use openupgradelib method

pedrobaeza:
* If you had "Create a task and track hours" option in `track_service` v10 field, the most similar one respecting task creation (new v11 field `service_tracking`) is "Create a task in a new project". Thus, we map that value.
* Use openupgradelib `delete_records_safely_by_xml_id` method for removing the `duplicate_field_xmlid` record.

cc @Tecnativa

### DIFF
--- a/addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
+++ b/addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
@@ -1,4 +1,5 @@
 # Copyright 2018 Tecnativa - Vicent Cubells
+# Copyright 2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
@@ -16,7 +17,7 @@ def migrate_sale_line_tasks(env):
 
 
 def migrate_project_sale_line(env):
-    service_column = openupgrade.get_legacy_name('service_type')
+    service_column = openupgrade.get_legacy_name('track_service')
     openupgrade.logged_query(
         env.cr, """
             UPDATE project_project pj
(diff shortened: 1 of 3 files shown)


cubells (approved):
:+1: 

LGTM

JordiBForgeFlow:
@mreficent can you please try to replicate the migration using this PR?

pedrobaeza:
I switch to WIP because I want to check some things (and there's a little error right now).

MiquelRForgeFlow (changes requested):
The `column_copies` in the pre-migration file is wrong because the field `service_type` is new, it didn't exist before. Also, try to not use uppercase.

rafaelbn:
sugerencia mapep

pedrobaeza:
@mreficent you can check this one now, as it's completed.

addons/sale_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -47,12 +48,72 @@ def migrate_project_sale_line(env):
...
+
+    Project field depends on company, so this is applicable as soon as one
+    company has any project set.
+    """
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name('track_service'),
+        'service_tracking', [
+            ('manual', 'no'),
+            ('timesheet', 'project_only'),
```
MiquelRForgeFlow, line 70: you put above that `timesheet` is mapped to `task_new_project` or `task_global_project` but not `project_only`

pedrobaeza: sorry, incorrect docstring, but proper scripts. Fixing docs.

pedrobaeza: Done

MiquelRForgeFlow (approved):
LGTM if CI green

PR #1520, opened 2018-09-11
[FIX] project: Transfer also magic columns

pedrobaeza:
Magic columns (aka log fields) were not transferred on tasks created out of issues, and thus that fields were empty. With this, we transfer that information.

cc @Tecnativa


rafaelbn (approved):
Tested in our own instance :+1:

PR #1524, opened 2018-09-19
[ADD] Rename of 'res.partner.sector' model

MiquelRForgeFlow:
Followup of https://github.com/OCA/OpenUpgrade/pull/1523.

Functionally tested with https://github.com/OCA/partner-contact/pull/619/commits/5983f2cd617e57d0fd6b0f0706062b96a99b094f.

--

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
@@ -24,6 +24,15 @@ model_renames_ir_actions_report = [
     ('ir.actions.report.xml', 'ir.actions.report')
 ]
 
+# OCA/partner-contact
+# migration of partner_sector to partner_industry_secondary
+# Done here for avoiding the IntegrityError of duplicating a model:
+# The installing of base module already creates the res.partner.industry,
+# so if the rename is done later in the module, it gives that error.
+model_renames_partner_sector = [
+    ('res.partner.sector', 'res.partner.industry'),
+]
+
 
 def fill_cron_action_server_pre(env):
     """Prefill the column with a fixed value for avoiding the not null error,
@@ -55,6 +64,7 @@ def migrate(env, version):
     openupgrade.copy_columns(env.cr, column_copies)
     openupgrade.rename_columns(env.cr, column_renames)
     openupgrade.rename_models(env.cr, model_renames_ir_actions_report)
+    openupgrade.rename_models(env.cr, model_renames_partner_sector)
     env.cr.execute(
         """UPDATE ir_actions SET type = 'ir.actions.report'
         WHERE type = 'ir.actions.report.xml'""")


pedrobaeza:
This should be done on pre-migration of the module itself, not here.

MiquelRForgeFlow:
But if I do the rename in the module, then it breaks.
```
psycopg2.IntegrityError: duplicate key value violates unique constraint "ir_model_obj_name_uniq"
DETAIL:  Key (model)=(res.partner.industry) already exists.
```

pedrobaeza:
Uhm, OK, I see, please restore the branch and I reopen this. And what happens if the module is not installed?

MiquelRForgeFlow:
Nothing. If would happen something, the travis should be red. But I am not sure if I understood correctly your question.

MiquelRForgeFlow:
Why travis is not executing correctly?

pedrobaeza:
Travis is that way.... sometimes it simply doesn't work. I have retried and we need to wait.

MiquelRForgeFlow:
:green_apple: :green_apple:

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -75,4 +75,6 @@
     'ir.actions.report.xml': 'ir.actions.report',
     'stock.pack.operation': 'stock.move.line',
     'stock.picking.wave': 'stock.picking.batch',
+    # OCA/partner-contact
```
pedrobaeza: I think you should put a longer comment indicating the reason for this.

MiquelRForgeFlow: ok 😑

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -75,4 +75,6 @@
...
+    'res.partner.sector': 'res.partner.industry',
```
StefanRijnhart: How about the pre_init_hook of the OCA module?

odoo/addons/openupgrade_records/lib/apriori.py
MiquelRForgeFlow: done

odoo/addons/openupgrade_records/lib/apriori.py
pedrobaeza: The problem is that Odoo already creates this model on their base module, so it will conflict. Maybe we should only leave that old model to die? I'm thinking it's the best solution, as this one is not complete:

* Model and fields XML will still point to other module (not `base`).
* There's no real advantage of making this, as records are moved other way.

StefanRijnhart: Ah, I see. Don't you want to add a full migration in the base modules' migration scripts that is only run if the old sector module was installed?

MiquelRForgeFlow: Interesting question @StefanRijnhart I will do that.

MiquelRForgeFlow:
@StefanRijnhart @pedrobaeza done. Now shouldn't be any problems :)

pedrobaeza:
I still think it's a bad idea as explained in https://github.com/OCA/OpenUpgrade/pull/1524#discussion_r218904848. Better to just move data from one table to the one created by base.

MiquelRForgeFlow:
But your issues of your linked comment have been solved by doing the migration in the base.

StefanRijnhart:
@mreficent indeed, this takes care of the data migration, plus we should have a patch in OpenUpgrade that rewrites the owning module of an XMLID to the first module it encounters. Maybe if you could check your test migration that the XMLIDs of model and the name field are indeed modified to belong to the base module?

pedrobaeza:
Yeah, it's not complete here, but I insist it doesn't worth. Simply move data from one table to the other.

StefanRijnhart:
@pedrobaeza I find it difficult to see it your way. Moving a table instead of copying over the data lets the records keep their ids, which is very much preferred.

pedrobaeza:
@StefanRijnhart That data has been entered manually, so no possible XML-ID. Remember that what he is trying to do here is to have sectors information that was created in previous version with the module partner_sector as industries in core model. I think it's very convoluted to rename table, fields, etc instead of simply copying data.

StefanRijnhart:
I mean the `id` column of the table.

Why do you mention the need to rename fields? Isn't it just this model rename?

pedrobaeza:
No, and in fact, I'm seeing @mreficent has already taken my approach in the migration script of the module itself:

https://github.com/OCA/partner-contact/pull/619/commits/67934245d5ee8ed0daea0f454483ef7eb8534949

StefanRijnhart:
Indeed, @mreficent that migration in partner-contact seems to preclude this change. Is this PR obsolete then?

MiquelRForgeFlow:
@StefanRijnhart This PR is not strictly necessary for the migration as you can check. But it's just that I wanted to avoid having a deprecated model (res.partner.sector) in the migrated database and thus I wanted some kind of rename.

MiquelRForgeFlow:
Closing this because not strictly needed.

PR #1525, opened 2018-09-25
[FIX] Table renamed crm_lead_lost_reason using rename_tables

diggy128:
This fixes renaming crm_lead_lost_reason to crm_lost_reason using
openupgrade's rename_tables so that sequences are renamed as well.

Current behavior before PR:
Migration fails at sequence with new name

Migration runs.

--

### DIFF
--- a/addons/crm/migrations/9.0.1.0/pre-migration.py
+++ b/addons/crm/migrations/9.0.1.0/pre-migration.py
@@ -12,6 +12,7 @@ table_renames = [
     ('crm_case_categ', 'crm_lead_tag'),
     ('crm_case_stage', 'crm_stage'),
     ('section_stage_rel', 'crm_team_stage_rel'),
+    ('crm_lead_lost_reason', 'crm_lost_reason'),
 ]
 
 column_renames = {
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Looks like this was conditional for a reason, but maybe you can call rename tables on this one table if the table exists?

pedrobaeza:
This was made on purpose because the table comes from on OCA module (crm_lead_lost_reason), as stated in the comment, so this change has no sense.

StefanRijnhart:
@pedrobaeza the issue at hand is that the table's id sequence is not renamed. I would appreciate a proper fix.

diggy128:
@pedrobaeza I guess that performing the migration in core has some meaning too, right? 
So, should it be reverted and rename the id sequence with a query?
Either case I need to fix this to proceed with my migration so I might as well contribute a valid fix.

StefanRijnhart:
I think you should replace the sql with something like 

if openupgrade.table_exists(cr, old):
    openupgrade.rename_tables(cr, [(old, new)])

pedrobaeza:
OK about the sequence renaming, but you can't do it outside the condition to check if there's the table.

pedrobaeza (approved):
Now it's OK

pedrobaeza:
Thanks for the PR!

PR #1531, opened 2018-10-02
[ADD] Base

StefanRijnhart:
--


pedrobaeza:
You can rebase here.

StefanRijnhart:
Will be green when ~https://github.com/OCA/openupgradelib/pull/118~ and ~https://github.com/OCA/openupgradelib/pull/119~ are merged.

odoo/modules/loading.py
```diff
@@ -583,6 +583,7 @@ def load_modules(db, force_demo=False, status=None, update_module=False):
         registry.updated_modules += processed_modules
 
         # OpenUpgrade: run deferred tests
+        cr.commit()
```
StefanRijnhart, line 586: Without this commit, the transaction with the end stage migration script is rolled back. Probably due to the test run below.

odoo/addons/base/migrations/12.0.1.3/end-migration.py
```diff
@@ -0,0 +1,57 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """ Fork user admin off from user system. User admin keeps the original
+    partner, and user system gets a new partner. """
+    user_root = env.ref('base.user_root')
+    partner_admin = env.ref('base.partner_admin')
+    partner_root = env.ref('base.partner_admin').copy({'name': 'System'})
+    login = user_root.login
+    user_root.login = '__system__'
```
mvaled, line 50: I think you should also remove the password for the __system__ superuser.

StefanRijnhart:
@mreficent did you push to my branch on purpose?

MiquelRForgeFlow:
@StefanRijnhart yes.

StefanRijnhart:
@mreficent slightly unconventional but alright then.

odoo/addons/base/migrations/12.0.1.3/end-migration.py
StefanRijnhart: Thanks, fixed now! (down below)

odoo/addons/base/migrations/12.0.1.3/openupgrade_analysis.txt
```diff
@@ -74,7 +74,9 @@ DEL ir.ui.view: base.res_request_link_search_view
 DEL ir.ui.view: base.res_request_link_tree-view
 NEW res.country: base.xk (noupdate)
 NEW res.lang: base.lang_en_CA
-NEW res.partner: base.partner_admin (noupdate)
 DEL res.request.link: base.req_link_partner (noupdate)
 NEW res.users: base.template_portal_user_id
+# Done: moved XMLID of this user from auth_signup
 NEW res.users: base.user_admin (noupdate)
+NEW res.partner: base.partner_admin (noupdate)
+# Done: split up this user and partner from what is now the 'system' user
```
MiquelRForgeFlow: I think this kind of comments should go into the `openupgrade_analysis_work.txt` file instead.

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,69 @@
+# © 2018 Opener B.V. (stefan@opener.amsterdam)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.addons.openupgrade_records.lib import apriori
+from openupgradelib import openupgrade
+
+xmlid_renames = [
+    ('auth_signup.default_template_user', 'base.template_portal_user_id'),
+]
```
MiquelRForgeFlow, line 10: why not also add this?
```python
('auth_signup.default_template_user_config', 'base.default_template_user_config')
```

StefanRijnhart: I think the module merge should take care of this. This is not the case for the user as the name part of the XML id is different.

StefanRijnhart: Sorry, you are right. This module is not actually merged into base.

odoo/addons/base/migrations/12.0.1.3/openupgrade_analysis.txt
StefanRijnhart: Indeed, done!

MiquelRForgeFlow: But also I think you should revert the changes done in this `openupgrade_analysis.txt`

StefanRijnhart: Ha, that's what I tried but squashed against the wrong commit. Let me try again.

StefanRijnhart:
Even if this PR is green there will be errors in the logs because the way testing is currently setup means that the filestore of the test database is not available during the migration. I'd need to regenerate this database with storage backend `db` to fix that. Not sure if I have time for this today.

MiquelRForgeFlow:
So you are saying that this PR still will not be merged, right?

StefanRijnhart:
Well, technical it can be merged because it will likely be green. In that case I'll upload the new test database later. I'll leave it up to you guys.

mreficent:
:green_apple:

StefanRijnhart:
I really don't like this active branch pushing if it's not just a bugfix. Just merge the green branch and push another commit. @mreficent, have you got merge permissions on this project?

MiquelRForgeFlow:
Sorry, I was just testing the CI. Commit deleted.

mreficent:
Commit deleted.

StefanRijnhart:
Sorry for my absence. I see you waited for me to push a version of the database with the attachments in the filesystem. I've updated https://github.com/OCA/OpenUpgrade/wiki/How-to-create-a-reference-database on how to do so.

The tests are still green, and now there are no errors anymore concerning missing filestorage attachments. Ready for merge?

MiquelRForgeFlow (approved):
Nice, Then I will merge

pedrobaeza:
Have you put the TODOs for the remaining things? Like convert translations to new format.

MiquelRForgeFlow:
I can't merge it with "rebase and merge" option. @StefanRijnhart rebase to resolve the conflicts! :S

StefanRijnhart:
@pedrobaeza I am not aware of any todo's that I did not put in. How is the translation format changed for example?

MiquelRForgeFlow:
Translations comments are changed in https://github.com/odoo/odoo/pull/26925/files#diff-a41938cd3ff4c7e353103fcd581f839a (to be handled on `base`). Important comment about this on https://github.com/odoo/odoo/pull/26925#issuecomment-422445897.

StefanRijnhart:
Thanks for the pointers. I've added the migration of the translations now. Did I miss anything else?

MiquelRForgeFlow:
By the way, a new full analysis is need.

StefanRijnhart:
Agreed. Let's merge this first and I'll generate it.

StefanRijnhart:
Ah, it's merged!

PR #1532, opened 2018-10-02
[MIG][11.0] l10n_fr: Nothing to do

cubells:
cc @Tecnativa


addons/l10n_fr/migrations/11.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,177 @@
+---Fields in module 'l10n_fr'---
+l10n_fr      / res.partner              / siret (char)                  : NEW
```
legalsylvain, line 2: Hi. If OCA module l10n_fr_siret was installed on V10 -, the field is not new. So maybe we should change the full xml_id of the field, if it exists on the database. l10n_fr_siret.xxxx --> l10n_fr.xxxx.
I'm not sure about what I propose. (Not sure if it required in new versions). Do you have an insight @StefanRijnhart ?

StefanRijnhart: This should be done automatically by OpenUpgrade! But it would be good to check.

cubells: @StefanRijnhart  @legalsylvain 

As old field is store=True

https://github.com/OCA/l10n-france/blob/10.0/l10n_fr_siret/models/partner.py#L104

then the value exist on migration. I think I do not have to do anything else.

cubells: @legalsylvain field remains the same in version 11.0.

cubells: https://github.com/OCA/l10n-france/blob/11.0/l10n_fr_siret/models/partner.py#L80

legalsylvain: Hi @cubells. The question was not about the data in the table, but about the name of the field. 
if l10n_fr_siret is installed on V10, you have an xml_id named : "l10n_fr_siret.field_res_partner_siret"
In v11, once migrated, the database should contain the xml_id renamed : "l10n_fr.field_res_partner_siret". 

As @StefanRijnhart said previously, it should be done automatically, but maybe it will not works because the field exist in both modules in V11. (because of redefinition). But the renamed should be done, because if we uninstall l10n_fr_siret module in V11, the siret field should stay and not be dropped.

Except that question, :+1: Thanks for migrating this module !

regards.

cubells: @legalsylvain 

To review.

PR #1533, opened 2018-10-02
12.0 product uom migration

PCatinean:

Migration to v12 of uom and product module

Current behavior before PR:

It doesn't load?

It loads if everything goes as planned :)

--


hmrodrigues:
I've a question regarding the base model migration.

How will we handle the admin user id?
In Odoo 12, the admin user now has the ID 2 and the root user (aka System) has the ID 1

I believe that the root user (ID 1) will not be a problem (the role is the same, but now has a different name and can't login (unable to set active=True, at least via ORM), but the admin user (now with id 2) can be, because the new admin will be a old normal user (or even a public one)

An example for this is the flow to set the [web.base.url](https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_users.py#L581) parameter. If a module implements something like this but without a try except block, the user will have issues with permissions.

StefanRijnhart:
hmoridriques, can you create an issue to discuss this? It is not relevant for this particular PR

pedrobaeza:
Can you rebase and solve existing conflicts with the new base? @mreficent do you have a replacement for this?

MiquelRForgeFlow:
Yes and yes, but this PR also puts migration scripts for `product`, and `product` depends of `mail`. Until `mail` gets merged, I don't see the need to update this PR.

pedrobaeza:
OK, I continue the escalation over dependencies and merging things. Any way, should we close this?

MiquelRForgeFlow:
No need to close this. I will just update this branch in the following days :)

MiquelRForgeFlow:
I doesn't understand the origin of the error in the CI

pedrobaeza:
The problem here is that the module `uom` doesn't exist in v11, so no migration scripts are going to be processed for this module, as it's going being installed for the first time during migration. We have to do all the stuff in other module migration scripts (base?) or trick this. Investigating the best way right now...

pedrobaeza:
And here we go! I have moved the `uom` renamings to `base` module, fix some things inside `product` migration scripts, and also fixed main attachment id of `mail` module method.

addons/mail/migrations/12.0.1.0/end-migration.py
```diff
@@ -37,21 +37,25 @@ def fill_mail_thread_message_main_attachment_id(env):
...
+            prefetch_fields=False,
+        )
+        grouped_attachments = attachment_obj.read_group(
             [('res_model', '=', model)],
-            fields=['res_id', 'id:array_agg'],
+            fields=['res_id'],
             groupby=['res_id'])
-        attachs = [(a['res_id'], a['id']) for a in grouped_attachments]
+        attachs = {
+            a['res_id']: attachment_obj.search(a['__domain'])
```
MiquelRForgeFlow, line 48: shouldn't we add `.with_context(active_test=False)` in this `search`? 🤔

pedrobaeza: Has attachment an `active` field?

MiquelRForgeFlow: Yes, it has.

MiquelRForgeFlow: https://github.com/OCA/OpenUpgrade/pull/1678

davidtranhp: attachment in Odoo 11 does not have `active` field.

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
```diff
@@ -1,12 +1,47 @@
...
+    ('product_uom', 'uom_uom'),
+    ('product_uom_categ', 'uom_category'),
+]
+
 xmlid_renames = [
     ('auth_signup.default_template_user', 'base.template_portal_user_id'),
     ('auth_signup.default_template_user_config',
      'base.default_template_user_config'),
+    ('product.group_uom', 'uom.group_uom'),
+    ('product.product_uom_categ_unit', 'uom.product_uom_categ_unit'),
```
cuongnmtm, line 24: Missing  `('product.product_uom_categ_vol', 'uom.product_uom_categ_vol'),`

cuongnmtm (commented):
Missing an item in `xmlid_renames`

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
pedrobaeza: Can you point where in 11.0 code this record is defined?

davidtranhp: Is it? https://github.com/odoo/odoo/blob/45bd7a58d5202d6b9387636317eb33efa569f458/addons/product/data/product_data.xml#L15

pedrobaeza: I mean `product_uom_categ_vol` pointed by @cuongnmtm. That one doesn't exist...

davidtranhp: This one: https://github.com/odoo/odoo/blob/45bd7a58d5202d6b9387636317eb33efa569f458/addons/product/data/product_data.xml#L27

pedrobaeza: Is that v11?

davidtranhp: Yes, that is. https://github.com/odoo/odoo/blob/11.0/addons/product/data/product_data.xml

pedrobaeza: Thanks. Fixed in ab5e14825252bfd41d87c481d1318c148855b057

PR #1538, opened 2018-10-09
[10.0][IMP] mrp: bottleneck created to compute unit_factor

LoisRForgeFlow:
Without this filtering calculating the unit_factor for all the moves associated to all MOs could take more than 24 hours in big databases. Plus, the field is not useful for done and cancel MOs, so not worth to compute it.

In my local testing each loop lasted around 0.8 seconds but sometimes it went up to 3!. This means that in a database with +100k MOs you will be in this loop, in the best case, for 80k seconds (22 hours). With this fix it only took 7 minutes in my use case.


AaronHForgeFlow (approved):
:+1:

MiquelRForgeFlow (approved):

JordiBForgeFlow (approved):

PR #1539, opened 2018-10-09
[11.0][MIG] product_expiry & stock_picking_wave

MiquelRForgeFlow:
The `stock` module is already migrated.

--


MiquelRForgeFlow:
How I fix this? @pedrobaeza @StefanRijnhart 

selection_027

pedrobaeza:
Next time please only one module per PR.

pedrobaeza:
Why do you need to merge instead of renaming?

MiquelRForgeFlow:
Because with renaming I got the error I showed you above. Also, in the migration issue https://github.com/OCA/OpenUpgrade/issues/1005 it is put as a merging.

pedrobaeza:
Uhm, that's very weird. I think that maybe `update_module_names` method doesn't perform operations in the proper order. @StefanRijnhart do you have any clue about this?

StefanRijnhart:
The error implies that model/table stock.picking.wave is not properly renamed to stock.picking.batch. Can you point out to me where that is supposed to happen?

MiquelRForgeFlow:
Well, finally, what should we do with this PR?

StefanRijnhart:
Don't merge, but rename and properly migrate stock.picking.wave to stock.picking.batch in the module's migration script.

MiquelRForgeFlow:
@StefanRijnhart Yes, done! (I don't know why but before I didn't see analysis file, thus why I asked that)

Ready to review and merge! :green_apple:

StefanRijnhart (approved):
Great, thanks!

pedrobaeza:
I don't see the name renaming in this PR. Have you removed it?

MiquelRForgeFlow:
The rename was already in the aprori file.

pedrobaeza:
OK then

PR #1542, opened 2018-10-11
[10.0][FIX] account_asset: Fill expense account from asset one

pedrobaeza:
Following the behavior in previous version, if you don't have expense account, then the asset one is used. As now the field is required, we copy the asset account in those empty expense accounts

cc @Tecnativa


cubells (approved):
:+1: 
LGTM

PR #1543, opened 2018-10-11
[12.0][MIG] barcodes

MiquelRForgeFlow:
Migration scripts of `barcodes` module.

--


pedrobaeza (approved):
I think that nomenclature is not used by anything in community right now, but good to have a sane default when possible.

PR #1565, opened 2018-10-17
[FIX] uom_id is required when creating a product

diggy128:
Creating a product requires the uom to be not null. This fix adds the
Unit(s) uom when creating the dummy product

Creating a product requires the uom to be not null.

Current behavior before PR:
Migration fails when trying to create the product with null uom_id

The dummy product is created and migration continues

--

### DIFF
--- a/addons/sale/migrations/9.0.1.0/post-migration.py
+++ b/addons/sale/migrations/9.0.1.0/post-migration.py
@@ -40,6 +40,7 @@ def set_dummy_product(env):
         'name': 'Any product',
         'type': 'service',
         'order_policy': 'manual',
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
But isn't the default uom set in your database as per https://github.com/odoo/odoo/blob/9.0/addons/product/product.py#L724?

diggy128:
@StefanRijnhart You are right. 
This happened because I add uom_id to product.product (in a separate module) to allow setting a different uom per variant and defaults don't get propagated when redefining a field.

Please forgive the mistake and close this as invalid.

PR #1566, opened 2018-10-17
[12.0][MIG] Auto-installed modules with base and web

MiquelRForgeFlow:
`#Nothingtodo` in these auto installed modules.

This PR will be the first to merged after base is green and merged :)

--


pedrobaeza:
I prefer to have one module per PR, although all of them are auto-installed. Please take also into account default values of new fields that doesn't modify the behavior of previous databases.

MiquelRForgeFlow:
I any module needs some script, I will create a separate commit for that module

pedrobaeza:
But it's not only that: maybe you don't consider they need a migration script, but reviewers think it is, and then multiple discussions arise from that, so the thread is a mess. But not only that, you force reviewers to check all the modules you include in this PR. Maybe they want to check one, but not all.

Another tip: FYI, I remove `iap` module as soon as I create a DB. If I want a module that use that services, then it's installed by dependencies.

MiquelRForgeFlow:
Yeah, good points and good tip, really :+1:  Although in this case I think multiple discussions is unlikely and I would give bet a pea that the reviewers of this PR would be the same reviewers of the several PRs if I made them separated.

pedrobaeza:
Hehe, that's right, basically Stefan and me, but it's a general approach, and sometimes discussions arise in the most unexpected things!

odoo/openupgrade/doc/source/modules110-120.rst
```diff
@@ -529,19 +529,19 @@ missing in the new release are marked with |del|.
...
+|web_diagram                                 | Nothing to do                                   |
 +--------------------------------------------+-------------------------------------------------+
-|web_editor                                  |                                                 |
+|web_editor                                  | Nothing to do                                   |
 +--------------------------------------------+-------------------------------------------------+
-|web_kanban_gauge                            |                                                 |
+|web_kanban_gauge                            | Nothing to do                                   |
 +--------------------------------------------+-------------------------------------------------+
-| |del| web_planner                          |                                                 |
+| |del| web_planner                          | Nothing to do                                   |
```
pedrobaeza: If this module disappears, you have to see where you have to merge it (web?) for not having this one reported as missing in new version.

pedrobaeza (changes requested):

odoo/openupgrade/doc/source/modules110-120.rst
MiquelRForgeFlow: There are a lot of modules that go to missing in each version because they don't have continuity, it's something difficult to avoid. I think the merging should be done only when it makes sense. And in the web_planner module I think it doesn't makes sense. Odoo just removed it, don't merged it with any other. What I am saying is that, if hypothetically we have a custom module that continues the work of the web_planner, in this case we may not want the merge. So we should do only the essential merges ones.

pedrobaeza: I don't agree. Nobody is going to replace this shitty feature, so we should just ease the migration to this version merging this module with am existing one. You don't decide to install this one in previous version, but was a hard dependency, so you shouldn't need to do any extra stuff of migration.

MiquelRForgeFlow: ok

MiquelRForgeFlow:
Ready :green_apple:

pedrobaeza (approved):

PR #1569, opened 2018-10-18
[12.0][MIG] mail

MiquelRForgeFlow:
Migration scripts of `mail`.

--


addons/mail/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+mail         / mail.message             / moderator_id (many2one)       : NEW relation: res.users
+mail         / mail.moderation          / channel_id (many2one)         : NEW relation: mail.channel, required: required
+mail         / mail.moderation          / email (char)                  : NEW required: required
+mail         / mail.moderation          / status (selection)            : NEW required: required, selection_keys: ['allow', 'ban']
+# NOTHING TO DO
+
+mail         / mail.notification        / email_status (selection)      : selection_keys is now '['bounce', 'canceled', 'exception', 'ready', 'sent']' ('['bounce', 'exception', 'ready', 'sent']')
+mail         / mail.notification        / failure_reason (text)         : NEW 
+mail         / mail.notification        / failure_type (selection)      : NEW selection_keys: ['BOUNCE', 'RECIPIENT', 'SMTP', 'UNKNOWN']
+mail         / mail.notification        / mail_id (many2one)            : NEW relation: mail.mail
```
pedrobaeza: You should fill this field matching the `message_id` of `mail.mail` model with field `message_id` of this model.

addons/mail/migrations/12.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="mail_activity_data_todo" model="mail.activity.type">
+    <field name="days"/>
```
pedrobaeza: Remove this deprecated field, and I think `delay_count` as well, as it's the renamed field.

addons/mail/migrations/12.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,23 @@
...
+    <field name="name">To Do</field>
+    <field name="sequence">12</field>
+    <field name="delay_count">5</field>
+  </record>
+  <record id="mt_activities" model="mail.message.subtype">
+    <field name="default" eval="False"/>
+  </record>
+  <record id="mail_activity_data_call" model="mail.activity.type">
+    <field name="days"/>
```
pedrobaeza: _notify_prepare_template_context

addons/mail/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+# NOTHING TO DO: don't bother about test models
+
+mail         / mail.tracking.value      / track_sequence (integer)      : NEW
+# DONE: end-migration: filled using environment the tracked fields
+
+mail         / res.partner              / activity_date_deadline (date) : now a function
+# NOTHING TO DO
+
+mail         / res.partner              / message_last_post (datetime)  : DEL
+mail         / res.partner              / message_main_attachment_id (many2one): NEW relation: ir.attachment
```
pedrobaeza, line 118: We should fill the field following this logic:

https://github.com/OCA/OpenUpgrade/blob/0dfe8b21959f6c38f7d2becac286bb68d66cfe88/addons/mail/models/mail_thread.py#L2065-L2070

addons/mail/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def fill_mail_blacklist_res_partner(cr):
+    cr.execute(
+        """
+        INSERT INTO mail_blacklist (email, active,
+            create_uid, create_date, write_uid, write_date)
+        SELECT email, active, create_uid, create_date, write_uid, write_date
+        FROM res_partner rp
+        WHERE rp.%s = TRUE
```
pedrobaeza: Nitpicking: no need for `= TRUE` part.

addons/mail/migrations/12.0.1.0/end-migration.py
```diff
@@ -0,0 +1,36 @@
...
+        FROM mail_tracking_value mtv
+        INNER JOIN mail_message mm ON mtv.mail_message_id = mm.id
+        INNER JOIN ir_model_fields imf ON (
+            imf.name = mtv.field AND mm.model = imf.model) 
+        WHERE mm.model IS NOT NULL AND imf.track_visibility IS NOT NULL AND
+            imf.track_visibility != 'false'
+        """
+    )
+    for field, model in env.cr.fetchall():
+        if field in list(env[model]._fields.keys()):
```
pedrobaeza: For avoiding errors do:
```suggestion
        if env.get(model) and field in list(env[model]._fields.keys()):
```

pedrobaeza (changes requested):

addons/mail/migrations/12.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: ?

pedrobaeza: hehe, wrong copy/paste. The same about deprecated field.

MiquelRForgeFlow:
Ready :green_apple:

addons/mail/migrations/12.0.1.0/end-migration.py
```diff
@@ -0,0 +1,62 @@
...
+        type(env[k]), type(env['mail.thread'])) and env[k]._auto]
+    for model in thread_models:
+        records = env[model].with_context(active_test=False).search(
+            [('message_main_attachment_id', '=', False)])
+        grouped_attachments = env['ir.attachment'].read_group(
+            [('res_model', '=', model)],
+            fields=['res_id', 'id:array_agg'],
+            groupby=['res_id'])
+        attachs = [(a['res_id'], a['id']) for a in grouped_attachments]
+        for record in records:
```
pedrobaeza, line 45: Is this more performant than including directly the condition in the read_group `('res_id', 'in', records.ids)`?

addons/mail/migrations/12.0.1.0/end-migration.py
```diff
@@ -0,0 +1,62 @@
...
+            attachment_ids = [a[1] for a in attachs if a[0] == record.id]
+            if attachment_ids:
+                all_attachments = env['ir.attachment'].browse(
+                    attachment_ids[0])
```
pedrobaeza, line 49: I suppose that this contains all the IDs, not only first, isn't it?

pedrobaeza (commented):
I have been searching a way to do this by SQL, but no luck in getting the hard condition.

rvalyi:
Hello @mreficent I just tested your mail migration script for the 2nd time and it looks there is an issue with the fill_mail_blacklist_res_partner method when some mail is empty. Indeed, with my production database I get the following a stack trace:
```
2018-12-18 20:40:39,147 276 ERROR oe-partial-to-12 odoo.modules.loading: Error executing post migration script for module mail                                                                               
`-> analytic                                                                                                                                                                                                 
`-> auth_signup                                                                                                                                                                                              
`-> base_automation                                                                                                                                                                                          
`-> calendar                                                                                                                                                                                                 
`-> contacts                                                                                                                                                                                                 
   `-> mass_mailing                                                                                                                                                                                          
`-> fetchmail                                                                                                                                                                                                
`-> hr                                                                                                                                                                                                       
`-> mail_bot                                                                                                                                                                                                 
`-> portal                                                                                                                                                                                                   
   `-> digest                                                                                                                                                                                                
      `-> account                                                                                                                                                                                            
         `-> account_bank_statement_import                                                                                                                                                                   
         `-> account_facturx                                                                                                                                                                                 
         `-> base_iban                                                                                                                                                                                       
         `-> base_vat                                                                                                                                                                                        
            `-> l10n_fr                                                                                                                                                                                      
               `-> l10n_fr_fec                                                                                                                                                                               
               `-> l10n_fr_siret                                                                                                                                                                             
         `-> payment                                                                                                                                                                                         
            `-> payment_transfer                                                                                                                                                                             
            `-> sale                                                                                                                                                                                         
               `-> sale_stock                                                                                                                                                                                
         `-> stock_account                                                                                                                                                                                   
      `-> crm                                                                                                                                                                                                
         `-> mass_mailing_crm                                                                                                                                                                                
      `-> project                                                                                                                                                                                            
`-> product                                                                                                                                                                                                  
   `-> stock                                                                                                                                                                                                 
`-> rating                                                                                                                                                                                                   
`-> sales_team                                                                                                                                                                                               
`-> sms                                                                                                                                                                                                      
`-> snailmail                                                                                                                                                                                                
: null value in column "email" violates not-null constraint                                                                                                                                                  
DETAIL:  Failing row contains (2, null, t, null, 53, 2017-11-29 13:11:28.19984, 1, 2018-12-11 11:45:05.920844).                                                                                              
                                                                                                                                                                                                             
2018-12-18 20:40:39,182 276 WARNING oe-partial-to-12 odoo.modules.loading: Transient module states were reset                                                                                                
2018-12-18 20:40:39,187 276 ERROR oe-partial-to-12 odoo.modules.registry: Failed to load registry                                                               

2018-12-18 20:40:39,188 276 CRITICAL oe-partial-to-12 odoo.service.server: Failed to initialize database `oe-partial-to-12`.
  File "/odoo/src/odoo/service/server.py", line 1042, in preload_registries
  File "/odoo/src/odoo/modules/registry.py", line 86, in new
  File "/odoo/src/odoo/modules/loading.py", line 480, in load_modules
  File "/odoo/src/odoo/modules/loading.py", line 364, in load_marked_modules
    upg_registry=upg_registry,
  File "/odoo/src/odoo/modules/loading.py", line 259, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/odoo/src/odoo/modules/migration.py", line 187, in migrate_module
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 1445, in wrapped_function
    if use_env2 else cr, version)
  File "/odoo/src/addons/mail/migrations/12.0.1.0/post-migration.py", line 33, in migrate
    fill_mail_blacklist_res_partner(cr)
  File "/odoo/src/addons/mail/migrations/12.0.1.0/post-migration.py", line 16, in fill_mail_blacklist_res_partner
    """, (AsIs(openupgrade.get_legacy_name('opt_out')), ),
psycopg2.IntegrityError: null value in column "email" violates not-null constraint
DETAIL:  Failing row contains (2, null, t, null, 53, 2017-11-29 13:11:28.19984, 1, 2018-12-11 11:45:05.920844).
```
(Then I commented the function call as a work around to keep the migration going)

Hope this helps.

addons/mail/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def fill_mail_blacklist_res_partner(cr):
+    cr.execute(
+        """
+        INSERT INTO mail_blacklist (email, active,
+            create_uid, create_date, write_uid, write_date)
+        SELECT email, active, create_uid, create_date, write_uid, write_date
+        FROM res_partner rp
+        WHERE rp.%s
```
pedrobaeza: From @rvalyi's error comment:
```suggestion
        WHERE rp.%s AND rp.email IS NOT NULL
```

addons/mail/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+            AND rp.email NOT IN (SELECT email FROM mail_blacklist)
```
pedrobaeza: Instead of this, you can use `GROUP BY email` and put `MAX` group operator in the rest of fields.

pedrobaeza (changes requested):

addons/mail/migrations/12.0.1.0/end-migration.py
MiquelRForgeFlow: Right, because here `attachment_ids` is a list of lists.

addons/mail/migrations/12.0.1.0/post-migration.py
MiquelRForgeFlow: max(active)? active is boolean :S

pedrobaeza: Right, for that field, you can use `BOOL_OR(active)`.

addons/mail/migrations/12.0.1.0/end-migration.py
MiquelRForgeFlow: but... including this domain in the read_group is just the same as not including it because `records` are all the records of the model, the domain doesn't restrict anything :S

pedrobaeza: Yeah, if that field is not pre-filled by other means, it's indeed the same. OK, forget this.

addons/mail/migrations/12.0.1.0/post-migration.py
MiquelRForgeFlow: but what happens if active = true is an old record and active = false is a newer record, both having same mail? if grouped, they will be put in the blacklist as active = true.

I think the good approach is not grouping, just ordering the records by id DESC before the insert

pedrobaeza: The problem here is performance doing this, and I'm not sure about the correctness of this any way. The case you are talking about can be the same conflictive if you have disabled and blacklisted the last created contact and leave the other one. Take into account that having one archived contact blacklisted and the non blacklisted one active, is not going to disable the sending, as the blacklist record will be archived as well.

Thus, my recommendation is still to group with OR.

MiquelRForgeFlow: Also MAX(create_uid) doesn't makes sense. It should have the last create_uid, not the max :S

pedrobaeza: Not really. Use `MIN` or `MAX` as you prefer, but selecting the last or the first one doesn't follow any business logic.

MiquelRForgeFlow:
Ready :green_apple:

addons/mail/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,38 @@
...
+def fill_mail_blacklist_res_partner(cr):
+    cr.execute(
+        """
+        INSERT INTO mail_blacklist (email, active,
+            create_uid, create_date, write_uid, write_date)
+        SELECT email, BOOL_OR(active), MAX(create_uid), MAX(create_date),
+            MAX(write_uid), MAX(write_date)
+        FROM res_partner
+        WHERE %s AND email IS NOT NULL
+            AND email NOT IN (SELECT email FROM mail_blacklist)
```
pedrobaeza: You don't need this part as you are already grouping by email. That's the main improvement why why group.

pedrobaeza (approved):
I have pushed a better version of the fill blacklist using OpenUpgrade 9.5+ feature `ON CONFLICT`.

PR #1570, opened 2018-10-19
[ADD] crm_phonecall: Auto-install it if present and don't convert phonecalls to logged activities

pedrobaeza:
cc @Tecnativa

### DIFF
--- a/addons/crm/migrations/9.0.1.0/post-migration.py
+++ b/addons/crm/migrations/9.0.1.0/post-migration.py
@@ -7,6 +7,16 @@ from openupgradelib import openupgrade
 
 
 def migrate_phonecalls(env):
+    """Convert each phonecall to a logged activity. If you have crm_phonecall
+    OCA module in your environment, this is not done, and the module is marked
+    instead to be installed, as the module replaces exactly this feature and
+    you save this time.
+    """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@cubells please review

pedrobaeza:
@mreficent can you please check?

MiquelRForgeFlow:
But the `crm_phonecall` doesn't exist in v8.

pedrobaeza:
This is not for v8, but for v9. It detects if you have the new v9 module in the module in the list (this checks happens after base update, which refresh module list), and it avoids the heavy task of converting calls to the log.

MiquelRForgeFlow:
I don't understand. Why not put `auto_install = TRUE` in that module?

pedrobaeza:
You don't want that module autoinstalled on a fresh v9. You do if you are migrating from v8. And not only that, you don't want that hacky and ugly way of having the calls somehow logged on the chatter if you are going to have the same features than in v8.

MiquelRForgeFlow (approved):
Ah, now I understand :+1:

MiquelRForgeFlow:
Shouldn't you check if `calendar` module is already installed? I ask this because `crm_phonecall` depends on `calendar`.

pedrobaeza:
It's a dependency of `crm`, so no problem on that.

PR #1571, opened 2018-10-19
[FIX] sale_timesheet: Restrict the SO line assignment to projects

pedrobaeza:
Previous query were very greedy, assigning a lot of projects that have several sales order lines related or that have been created after the project creation.

This query prevents that restricting the update.

cc @Tecnativa


pedrobaeza:
@cubells please review

MiquelRForgeFlow:
There is two "project_project pj". I am afraid this doesn't give problems. I think it's clearer to use "project_project pj1" and "project_project pj2".

pedrobaeza:
It's not a problem. They are in different scope.

cubells (approved):
:+1: 

LGTM

Tested.

PR #1572, opened 2018-10-19
[11.0] sale_proforma_report Incorporated into standard in sale

LoisRForgeFlow:
Pro-forma invoices can be printed from SO after activating a setting.

### DIFF
--- a/addons/sale/migrations/11.0.1.1/post-migration.py
+++ b/addons/sale/migrations/11.0.1.1/post-migration.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2017 bloopark systems (<http://bloopark.de>)
+# Copyright 2018 Eficent Business and IT Consulting Services S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow:
@pedrobaeza I'm don't see clear how to automate the activation (in premigration I don't have the new group in post-migration I don't have the old module) of the group

pedrobaeza:
You can detect if the column `proforma` is present on the table `sale_order`, and then activate the group if so.

LoisRForgeFlow:
@pedrobaeza Thanks for the :bulb:! Done.

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

addons/sale/migrations/11.0.1.1/post-migration.py
```diff
@@ -8,9 +9,20 @@
...
 
 
+def activate_proforma(env):
+    """If module `sale_proforma_report` (OCA/sale-reporting) was installed we
+    need to activate the "Pro-forma Invoices" setting as the module was
+    deprecated in favour of it."""
+    if openupgrade.column_exists(env.cr, 'sale_order', 'proforma'):
+        employee_group = env.ref('base.module_category_human_resources')
+        proforma_group = env.ref('sale.group_proforma_sales')
+        employee_group.write({'implied_ids': [(4, proforma_group.id)]})
```
pedrobaeza, line 19: Is this the way that Odoo uses when you mark the field in settings?

LoisRForgeFlow: yes, exactly, I took it from res.settings.

LoisRForgeFlow: btw, I already tested it and worked :green_heart:

PR #1573, opened 2018-10-19
[12.0][MIG] fetchmail

MiquelRForgeFlow:
`#Nothingtodo` in `fetchmail` module.

--


pedrobaeza (approved):
`action_id` feature is being lost in fact, and can be replaced as explained on the commit with `base_automation` module + action on creation, but I don't think it deserves the effort, as I don't know any DB that has this filled.

PR #1574, opened 2018-10-19
[12.0][MIG] lunch

MiquelRForgeFlow:
`#Nothingtodo` in `lunch` module.

--


pedrobaeza:
Is this for encouraging Odoo to use OpenUpgrade to migrate their instance? :stuck_out_tongue_closed_eyes: 

Please check Travis status.

MiquelRForgeFlow:
> Is this for encouraging Odoo to use OpenUpgrade to migrate their instance? 😝

hahaha good one. It's just that I think that we should always try to do (regardless of our interest) the modules that have a `#NothingToDo` or only a noupdate_changes file.

MiquelRForgeFlow:
green ci

pedrobaeza (approved):

PR #1578, opened 2018-10-22
[12.0][MIG] resource

MiquelRForgeFlow:
Migration scripts for `resource` module.

--


addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Fields in module 'resource'---
+resource     / resource.calendar        / hours_per_day (float)         : NEW
```
pedrobaeza: You should compute this field getting the average of the hours per day. You can call `_onchange_hours_per_day` on post-migration for getting this.

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Fields in module 'resource'---
+resource     / resource.calendar        / hours_per_day (float)         : NEW 
+resource     / resource.calendar.attendance / day_period (selection)        : NEW required: required, selection_keys: ['afternoon', 'morning'], req_default: function
```
pedrobaeza, line 5: I would filled this depending on the ~~`date_to`~~ `hour_from` field: if before 12:00 a.m, `morning`. `afternoon` elsewhere.

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
...
+resource     / resource.calendar.leaves / time_type (selection)         : NEW selection_keys: ['leave', 'other']
```
pedrobaeza, line 8: I don't have this clear if the default value `leave` is the proper one for getting the same result as in v11. Have you investigated this?

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
...
+# NOTHING TO DO
+
+resource     / resource.calendar.leaves / tz (selection)                : DEL selection_keys: function
+resource     / resource.calendar        / tz (selection)                : NEW required: required, selection_keys: function, req_default: function
+# Done: post-migration: filled tz from leave_ids.tz
+
+resource     / resource.resource        / tz (selection)                : NEW required: required, selection_keys: function, req_default: function
+# Done: post-migration: filled tz from user_id.tz
```
pedrobaeza, line 16: There are other possible resources to take into account when doing the migration scripts. For example, workcenters for `mrp` module. But this is to be done in that module :wink:

pedrobaeza (changes requested):

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: and how I put 12:00 a.m in postgres?

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: It should be handled in hr_holidays module? The resource.calendar.leaves if linked to hr.holiday with holiday_id, then the resource.calendar.leaves should have the same time_type as its holiday.
See https://github.com/odoo/odoo/blob/12.0/addons/hr_holidays/models/hr_leave.py#L544

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: And thus I should do something?

pedrobaeza: Not yet. Only for taking into account for the future.

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: OK, then we reassign in that module.

addons/resource/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: No need for that in the proper field. It's a float field.

pedrobaeza:
I have finished the scripts with the remaining points.

@mreficent several things for remembering when doing other migration scripts:

* no need since v10 to put `use_env=True`. That's the default.
* Use `openupgrade.logged_query` for most of the SQL for having the proper trace.
* You can use the first line for the SQL expression as it will be output correctly on log without the extra line break.

pedrobaeza (approved):

PR #1579, opened 2018-10-22
[12.0][MIG] crm

MiquelRForgeFlow:
Migration scripts for `crm` module.

--


pedrobaeza:
Finish migration scripts with:

* Try to remove deleted noupdate records
* Create XML-IDs for linking properly act_window.view records
* Optimize blacklist filling

@mreficent please switch your habit of putting `use_env=False`. This is limiting without sense the possibilities to use and makes larger the diff when we need to use it.

PR #1580, opened 2018-10-23
[9.0] update upstream 75388edd9c

MiquelRForgeFlow:
Merging upstream Odoo up until commit https://github.com/odoo/odoo/commit/75388edd9c6cb54ee04c2bcac6886dbedd717700.

There were no conflicts.

--


openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
```diff
@@ -145,6 +145,7 @@
       <field name="active" eval="False"/>
     </record>
     <record id="res_partner_title_doctor" model="res.partner.title">
+      <field name="domain"/>
```
pedrobaeza: This field no longer exists. That's why it appears in the noupdate. Please remove it from all for avoiding the warnings.

MiquelRForgeFlow: Done

openerp/addons/base/migrations/9.0.1.3/noupdate_changes.xml
```diff
@@ -607,6 +608,7 @@
       <field eval="27" name="phone_code"/>
     </record>
     <record id="GTQ" model="res.currency">
+      <field name="company_id"/>
```
pedrobaeza: This one also doesn't exist. You should check WARNING in logs for knowing it.

MiquelRForgeFlow:
I thought to also include here https://github.com/OCA/OpenUpgrade/pull/1449 :)

MiquelRForgeFlow:
I don't understand, really. I created locally a v8 database with demo data. I installed all `l10n_*` modules (except `l10n_be_hr_payroll_account` that for an unknown reason I couldn't install it). Then I migrated it and this error didn't appear.

pedrobaeza:
@mreficent v8 database is static and was generated by @StefanRijnhart some time ago. For sure this is a new constraint that the old DB don't comply in demo data, so you should remove that and do it via SQL code.

MiquelRForgeFlow:
Ok, ready to merge. I just removed it. I won't waste more time in something that I really don't bother  😥

pedrobaeza:
Well, can you remind me which one was, because maybe it's important (I think it was the receivable account type, which should be reconciled or reconcile system won't work).

MiquelRForgeFlow:
Both the receivable and the payable. Both have to be related to reconcilable accounts.

pedrobaeza:
Then the SQL code is needed. What I don't get is why this was working before and not now...

MiquelRForgeFlow:
Could you merge this PR and do that SQL code in another PR? I ask it because this PR is unrelated with that thing of the accounts.

pedrobaeza:
OK, but I'm not going to work on that part, as I don't need it for now in my projects.

PR #1582, opened 2018-10-24
[10.0] update upstream bfbaa09268

MiquelRForgeFlow:
Merging upstream Odoo up until commit https://github.com/odoo/odoo/commit/bfbaa092688c6282ae8468ab0cb6bfc2fcdfc552.

There were no conflicts.

--


pedrobaeza:
There's a strange warning with country states:
```
2018-10-24 11:16:42,805 22275 INFO openupgrade odoo.sql_db: bad query: ALTER TABLE "res_country_state" ADD CONSTRAINT "res_country_state_name_code_uniq" unique(country_id, code)
2018-10-24 11:16:42,806 22275 WARNING openupgrade odoo.models.schema: Table 'res_country_state': unable to add constraint 'unique(country_id, code)'!
If you want to have it, you should update the records and execute manually:
ALTER TABLE "res_country_state" ADD CONSTRAINT "res_country_state_name_code_uniq" unique(country_id, code)
```
Why they are duplicated? Can you check if there's a missing XML-ID renaming in base?

MiquelRForgeFlow:
I think the problem is that in v9 this sql_constraint doesn't exist. So, when in v10 tries to add this contrain, it happens that it can't because the records in database are not compatible with that constraint. Some records in v9 maybe have same `country_id` and `code` 🤔

MiquelRForgeFlow:
I don't understand, really. I created locally a v9 database with demo data. I installed all `l10n_*` modules. Then I migrated it and this warning didn't appear.

pedrobaeza (approved):
Well, I suppose it's due to bad initial v9 DB with duplicated data, so let's move forward.

PR #1583, opened 2018-10-24
[REF] move merged_modules list into apriori files to make the analysis easier

legalsylvain:
This PR is around a work I make, to analysis Openupgrade coverage of Odoo instances to evaluate the workload.

One of the difficulties is to know if a module is unported or if the features has been merged in another module. (OCA / Odoo SA)

For that purpose, I'm moving this list into the apriori file to make the analysis easier.

the PR doesn't change any behaviour of openupgrade, and has been locally tested.

CC : @StefanRijnhart 

--


MiquelRForgeFlow:
Could you also do the same to v10? :)

MiquelRForgeFlow (approved):

legalsylvain:
> Could you also do the same to v10? :)

#1584

PR #1586, opened 2018-10-25
9.0 ref apriori file

legalsylvain:
This PR is to continue the work of analysis I began about migration coverage.
(Réf : https://github.com/OCA/OpenUpgrade/pull/1583)

For that purpose, I have to distinguish merged / renamed modules from OCA + odoo and modules from others repositories.
This PR is just refactoring.

I'll do it the same for version 10.0+ after.

--

### DIFF
--- a/openerp/addons/openupgrade_records/lib/apriori.py
+++ b/openerp/addons/openupgrade_records/lib/apriori.py
@@ -61,28 +61,29 @@ _oca_odoo_renamed_modules = {
 }
 
 _other_renamed_modules = {
-    # ??? -> OCA/manufacture-reporting
+    # odoomrp-wip -> OCA/manufacture-reporting
     'mrp_bom_structure_xls_level_1': 'mrp_bom_structure_xlsx_level_1',
 
-    # ??? -> OCA/product-attribute
+    # odoomrp-wip -> OCA/product-attribute
     'product_m2m_categories': 'product_multi_category',
 
-    # ??? -> OCA/product-variant
+    # odoomrp-wip -> OCA/product-attribute 10.0
+    'sale_product_variants': 'sale_variant_configurator',
+    'sale_stock_product_variants': 'sale_stock_variant_configurator',
+
+    # odoomrp-wip -> OCA/product-variant
     'product_variants_no_automatic_creation': 'product_variant_configurator',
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't think that comments are the best way to reflect things. We should have other structured data.

MiquelRForgeFlow (changes requested):
Sorry, but I don't like this approach ❌

legalsylvain:
> I don't think that comments are the best way to reflect things. We should have other structured data.

I mainly splited each list into 2 lists. (the OCA / odoo modules), and others modules that are not under OCA umbrella. And comments are already present in this file. So I don't understand which comments you're talking about.

> Sorry, but I don't like this approach

Could you give a reason @mreficent  ? ;-)

MiquelRForgeFlow:
> Could you give a reason @mreficent  ? ;-)

Aesthetic changes don't need reasons. You just like them or not. 😏

pedrobaeza:
> I mainly splited each list into 2 lists. (the OCA / odoo modules)

@legalsylvain that's why I say that this is not enough for identifying everything with full information, and more in an automated way IMO.

MiquelRForgeFlow:
> and others modules that are not under OCA umbrella

It's not important if the origin name of the module is not in OCA/Odoo. What's important is that the target rename/merge is in OCA/Odoo.

legalsylvain:
> It's not important if the origin name of the module is not in OCA/Odoo.

for the work i began during odoo experience this year, yes it's important. 

> What's important is that the target rename/merge is in OCA/Odoo.

And some modules are not under oca umbrella at all. 

```
    'sale_product_variants': 'sale_variant_configurator',
    'sale_stock_product_variants': 'sale_stock_variant_configurator',
```

pedrobaeza:
That modules were a transfer from odoomrp-wip to OCA as agreed when done.

MiquelRForgeFlow:
> for the work i began during odoo experience this year, yes it's important.

Then upload that work so this way we can value correctly this PR :)

StefanRijnhart:
I don't understand why you need the distinction. As is pointed out above, the odoo-mrp project is an exception. And the distinction is only important if there is a name clash. Do you have an example of such a clash?

pedrobaeza:
Closing as no further news.

Issue #1587, opened 2018-10-25
Outfactoring db analyzing


blaggacao:
@StefanRijnhart Don't [bark at me](https://github.com/OCA/OpenUpgrade/issues/502#issuecomment-225495301). :) It's a suggestion to make the UX of OpenUpgrade simpler.
I already started [`click-odoo-migrator`](https://github.com/xoe-labs/click-odoo-migrator) to drive the migration process in a scriptable way.

I'm evaluating the possibility to outfactor the relevant parts of the analysis plugin into a subcommand like `click-odoo-migrator analyze` (eg. via a `openupgrade-analyzer` lib)

Obviously that does not make any sense at all if there is no shared single repository for the apriori and blacklist information. And it would only make limited sense if it would not be a true and coordinated attempt to become the preferred way of driving database analysis (eventually through CI automation) in the community.

Please let me know your thoughts and feelings about this idea.

Best, David

/cc @rvalyi

blaggacao:
Ps: My vision of a workflow would look like this:

- `click-odoo-migrator analyze 11.0..12.0` on any given git repository that has both branches/tags `11.0` and `12.0`
- Would report a complete module tree with migration folders only (for all modules recognized in this repository), and the analysis therein.
- This tree could then be passed to `click-odoo-migrator --mig-directory` for applying the migration(s)
- And for the shared apriori, a multioption flag needs to be provided that takes one or several file(s) or URL(s) to pre-load know mutations (preferably from `json` format).

Furthermore: Maintaining a priori data within such library should be avoided in order to disconnect release cycles.

blaggacao:
This thought leads me to another question: All methods in the [compare module](https://github.com/OCA/OpenUpgrade/blob/11.0/odoo/addons/openupgrade_records/lib/compare.py) are supposedly public. I guess that does not express the correct usage intention. Is my assumption correct that there are only 2 really public methods in this module?

blaggacao:
Once that's achieved, what about code-generators (e.g. jinja tamplates) for some of the more commonly identifiable ("patternable") cases, like: 

- https://doc.therp.nl/openupgrade/xmlids.html
- https://doc.therp.nl/openupgrade/format.html

in a `openupgrade-generators` lib package?

blaggacao:
Here we go, with complete history: https://github.com/xoe-labs/openupgradelib-contrib

lina2358:
Any opinion on this one? It would be nice to further this thought.

[...]

PR #1588, opened 2018-10-25
Make private methods private

blaggacao:
Assuming that there are only two true entry points into this module,
mark all other methods as private to the module.
Preparing: https://github.com/OCA/OpenUpgrade/issues/1587#issuecomment-433121344

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -14,12 +14,12 @@ import copy
 from odoo.addons.openupgrade_records.lib import apriori
 
 
-def module_map(module):
+def _module_map(module):
     return apriori.renamed_modules.get(
         module, apriori.merged_modules.get(module, module))
 
 
-def model_map(model):
+def _model_map(model):
     return apriori.renamed_models.get(model, model)
 
 
@@ -32,7 +32,7 @@ IGNORE_FIELDS = [
     ]
 
 
-def compare_records(dict_old, dict_new, fields):
+def _compare_records(dict_old, dict_new, fields):
     """
     Check equivalence of two OpenUpgrade field representations
     with respect to the keys in the 'fields' arguments.
@@ -42,24 +42,24 @@ def compare_records(dict_old, dict_new, fields):
     """
     for field in fields:
         if field == 'module':
-            if (module_map(dict_old[field]) != dict_new[field]):
+            if (_module_map(dict_old[field]) != dict_new[field]):
                 return False
         elif field == 'model':
-            if (model_map(dict_old[field]) != dict_new[field]):
+            if (_model_map(dict_old[field]) != dict_new[field]):
                 return False
         elif dict_old[field] != dict_new[field]:
             return False
     return True
 
 
-def search(item, item_list, fields):
+def _search(item, item_list, fields):
     """
     Find a match of a dictionary in a list of similar dictionaries
     with respect to the keys in the 'fields' arguments.
     Return the item if found or None.
     """
     for other in item_list:
-        if not compare_records(item, other, fields):
+        if not _compare_records(item, other, fields):
             continue
         return other
     # search for renamed fields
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
What does it matter if they are private or public methods?

blaggacao:
I wanted to outfactor this into a pip installable library. So it should become syntactically correct and transparent to consume, I guess.

Another way to look at it:
'' Vs '_' conveys meaning.
It's like a faulty documentation.

pedrobaeza:
I'm not sure if you would be able to decouple this from ORM and module architecture itself.

blaggacao:
@pedrobaeza, well the interface of this `compare` module is damn pretty minimal, so how you provide ORM capabilities to the wrapper code is not a concern of this module, actually.

~But you are right, it will not work without ORM, but there are other ways to provide it.~

Actually looking carefully this module only interfaces with a handful table in an odoo DB, so ORM could be easily obviated.

pedrobaeza:
The problem I see decoupling the analysis to this architecture is that there are some information that you can't get only browsing the code (or it's very hard to do). Are you sure you want to go to that ride that can lead to frustration?

blaggacao:
> The problem I see decoupling the analysis to this architecture is that there are some information that you can't get only browsing the code (or it's very hard to do). Are you sure you want to go to that ride that can lead to frustration?

Do you mean analyzing (in other words report generation) is not entirely contained in this module (excluding file writing and such stuff, just data structure wise)? If so, then we (I) should first change the interface of this module, of course.

Are there any specific side effects for analyzing not contained here or for which no clear interface description could be defined (eventually requiring an ORM)?

pedrobaeza:
No, it's all here, but with the help of ORM of both analyzed versions and communicating through XML-RPC.

blaggacao:
> No, it's all here, but with the help of ORM of both analyzed versions and communicating through XML-RPC.

Understand. Thanks, that's not a problem, all _that_ boilerplate can live in `click-odoo-migrator` without any problem.

blaggacao:
My motivation for all this is: I don't need a whole OpenUpgrade > 1GB repo just as boilerplate to get a report of the changes that have occurred between two databases.

pedrobaeza:
Yeah, I understand and share the motivation, but sometimes it's not so easy.

blaggacao:
Here we go, with complete history: https://github.com/xoe-labs/openupgradelib-contrib

blaggacao:
@pedrobaeza 

> Are you sure you want to go to that ride that can lead to frustration?

It took me 2 days, so far, and I'm already pretty happy... :smile: 

> No, it's all here

[Is it really?](https://github.com/xoe-labs/dodoo-migrator/tree/master/src/analyzer/schema_analyzer) :wink:

pedrobaeza:
Can you put an example of what this analyzes?

blaggacao:
Sure, it's still very `WIP`, but I'd explain the strategy:
```python
# 0. define receiving data structures for field spec of both branches
# 1. execute subroutine on two branches
# # 1. checkout branch (recursively)
# # 2. start odoo environment
# # 3. run get_dataframe_of_all_fields
# # 4. stop odoo environment
# 2. return a diff data frame with changed rows
# 3. identify known transitions
# 3bis. report known transition in a txt format similar to the OU analyzer format
# 4. generate PRE migration code stubs based on the known transitions - divided into: 
# - fully endogenous stubs (nothing to do)
# - exogenous stubs which provide code scaffolding and must be reviewed / completed manually
# 5. analyze noupdate=True body changes
# 6. Not sure yet: define body transformation functions for noupdate=True body changes
```
The key to understanding this strategy is:
- The [exogenous file](https://github.com/xoe-labs/dodoo-migrator/blob/master/src/analyzer/exogenous_schema_information.yaml)
- The [field slot analysis](https://github.com/xoe-labs/dodoo-migrator/blob/933c6e75a70c6c443e6b208f465d474aee9d0fe4/src/analyzer/schema_analyzer/analyzer.py#L39-L64) and it's [comparison](https://github.com/xoe-labs/dodoo-migrator/blob/933c6e75a70c6c443e6b208f465d474aee9d0fe4/src/analyzer/schema_analyzer/analyzer.py#L178)

Based on the field slot transitions we can catch transitions we couldn't before with the OU approach (because they are not visible from db introspection, only), furthermore those transitions are defined in [such an atomic way](https://github.com/xoe-labs/dodoo-migrator/blob/master/src/analyzer/schema_analyzer/_transitions.py) so that they can become the building blocks of composable code generators. The generated code might not be the shortest and most elegant code, but it will be standardized and repeatable.

After dealing with schema changes in PRE migration scripts and after executing a migration we can analyze [data body (`text`, `char`, `html`, `binary`) changes](https://github.com/xoe-labs/dodoo-migrator/blob/933c6e75a70c6c443e6b208f465d474aee9d0fe4/src/analyzer/data_analyzer/_transitions.py#L16) on `noupdate=true` records and see if we can find generic transformation functions for specific cases, and [high-lite suspect cases](https://github.com/xoe-labs/dodoo-migrator/blob/933c6e75a70c6c443e6b208f465d474aee9d0fe4/src/analyzer/data_analyzer/_transitions.py#L34-L37) otherwise.

pedrobaeza:
OK, got it, but for now it's very similar in `openupgrade_records` module way of working from what I'm seeing. When you have a real full analysis file, please show me.

blaggacao:
@pedrobaeza Yes, it's very similar. After all, that's the module I wanted to out-factor and give it's independence. Yet, it sticks with more atomic definitions than `openupgrade_records` this should be the game changer for enabling migration script stub generators. Generated code will be generally more transparent than borrowed library functions.

PR #1591, opened 2018-10-26
[10.0][FIX] procurement: Delete company view

rousseldenis:
Related to https://github.com/OCA/maintainer-tools/issues/390

--

### DIFF
--- a/addons/procurement/migrations/10.0.1.0/pre-migration.py
+++ b/addons/procurement/migrations/10.0.1.0/pre-migration.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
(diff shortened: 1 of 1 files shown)


rousseldenis:
@pedrobaeza

MiquelRForgeFlow:
Why?

rousseldenis:
See above.

MiquelRForgeFlow:
OMG, I didn't see before  😅

MiquelRForgeFlow:
Have you tested it in a test migration?

rousseldenis:
Not yet.

StefanRijnhart:
It is not entirely clear to me what the problem is with breaking views during migration. The errors are supposed to be disarmed by changes that we make in OpenUpgrade so that they don't interrupt the migration. In the migration post processing, Odoo deletes the views that are no longer present in the module such as this one (unless it is defined as noupdate). What problems causes this view exactly?

rousseldenis:
Yes. False positive. Happens without Open upgrade. Closing

StefanRijnhart:
OK, thanks for the followup!

PR #1594, opened 2018-10-30
[11.0] update upstream a219bdb85e

MiquelRForgeFlow:
Merging upstream Odoo up until commit https://github.com/odoo/odoo/commit/a219bdb85e9078f88bb5a7b43b3598762a7bd819.

There were no conflicts.

--


StefanRijnhart:
Why is this commit part of this merge? https://github.com/OCA/OpenUpgrade/pull/1594/commits/ef5f69093a500ea262275cfa809e3f9737fda678

MiquelRForgeFlow:
Because this is what my `compare_no_update_records` outputs. As it's likely the same but reordered, I am a bit lazy to check record by record if a new thing has appeared. So I overwrote it.

StefanRijnhart:
Can you propose that one separately from the upstream update?

MiquelRForgeFlow:
@StefanRijnhart I think last commit is better and easier to review than before :)

StefanRijnhart:
Sorry if I was unclear but I think such updates should be proposed in a separate PR

MiquelRForgeFlow:
Ok, no problem. Commit deleted. What to do now?

StefanRijnhart (approved):
We can now merge once green

MiquelRForgeFlow:
But doesn't get green :S

StefanRijnhart:
It's the dreaded `fatal: Could not parse object` error from git. In my experience, the only thing that helps in this case is a rebase.

MiquelRForgeFlow:
It's already rebased.

MiquelRForgeFlow:
:green_apple:

PR #1595, opened 2018-10-31
[11.0][MIG] product_by_supplier

MiquelRForgeFlow:
As @lreficent commented here https://github.com/OCA/purchase-workflow/pull/596#issuecomment-434623285, I made the migration of `product_by_supplier`.

--

### DIFF
--- a/addons/purchase/migrations/11.0.1.2/post-migration.py
+++ b/addons/purchase/migrations/11.0.1.2/post-migration.py
@@ -89,6 +89,24 @@ def fill_purchase_order_picking_fields(env):
     )
 
 
+def activate_vendor_pricelists(env):
+    """If module `product_by_supplier` (OCA/purchase-workflow) was installed we
+    need to activate the "Vendor Pricelists" setting as the module was
+    deprecated in favour of it."""
+    env.cr.execute(
+        """
+        SELECT id
+        FROM ir_model_data
+        WHERE module = 'purchase' AND
+            name = 'view_product_supplierinfo_search'
+        """
+    )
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
LGTM

addons/purchase/migrations/11.0.1.2/pre-migration.py
```diff
@@ -6,6 +6,24 @@
 from openupgradelib import openupgrade
 
 
+def detect_product_by_supplier(cr):
+    cr.execute(
+        """
+        SELECT id
+        FROM ir_model_data
+        WHERE module = 'product_by_supplier' AND
```
pedrobaeza: This is going to be always Falsy, as you have merged module in `purchase`. Any way, do this or other check in post for avoiding this ugly hack of creating a fake column. Do you know that in OpenUpgrade views are not removed? Use that.

pedrobaeza (changes requested):

addons/purchase/migrations/11.0.1.2/pre-migration.py
StefanRijnhart: For the record, while OpenUpgrade does not drop seemingly obsolete columns and tables like Odoo does, it does delete module data (such as views) that is not defined as noupdate but only once all the modules have been processed.

pedrobaeza: Thanks for the complementary information, Stefan!

pedrobaeza (approved):

PR #1601, opened 2018-11-03
[FIX] stock: Handle partial reserved moves + fixes in sml creation

pedrobaeza:
In v11, the only way to handle reserved quantities (partial or full) are through stock.move.line records, but there are no hard link between quant reserved qty and on which stock.move is reserved.

Thus, we need to create stock.move.line (sml) for recreating this reservations, but this time information is got directly from old stock.quant through reservation_id field.

There are also 2 small fixes on previous sml generation code:

* `partially_available` state should be set on state != 'done', not only when waiting another move.
* We should group sml also by package, owner and destination package for preserving that options in new version (they are not usual, but can happen).

cc @Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -108,7 +108,7 @@ def set_partially_available_state(env):
         env.cr, """
         UPDATE stock_move
         SET state = 'partially_available'
-        WHERE partially_available AND state='waiting'"""
+        WHERE partially_available AND state != 'done'"""
     )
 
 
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -185,9 +187,74 @@ def create_stock_move_line(env):
             LEFT JOIN stock_quant sq ON sq.id = smol.reserved_quant_id
             LEFT JOIN stock_production_lot spl ON spl.id = sq.lot_id
         WHERE sm.state = 'done'
-        GROUP BY sq.lot_id, spo.product_id""",
+        GROUP BY sq.lot_id, spo.product_id, spo.owner_id, spo.package_id,
+            spo.result_package_id""",
```
MiquelRForgeFlow, line 191: I think you should also group by `sm.id`

MiquelRForgeFlow: Or something similar, I dunno. I did a test, and I see that there is a need to create in some way more stock move lines with done moves.

MiquelRForgeFlow (approved):
Functionally tested. Please merge ASAP! :)

pedrobaeza:
I have detected one last case, that is on incoming draft pickings. If we don't add stock.move.line, we have to fill by hand each of the picking lines for validating the picking (they are not automatically created), but I'll handle this in other PR.

MiquelRForgeFlow:
Please, let's wait until https://github.com/OCA/OpenUpgrade/pull/1597 is updated and merged.

pedrobaeza:
Too late

MiquelRForgeFlow:
No, I was referring to your new PR.

pedrobaeza:
Ah, OK, but they are independent, unless you are doing things that don't belong to `mrp` module in the other PR. Why do you want me to wait?

MiquelRForgeFlow:
Some things of `mrp` moved to `stock`, so `stock` migration scripts are affected.

pedrobaeza:
Uhm, I'm not sure that this is correct. `mrp` depends on stock, so you can perform everything inside `mrp` module. No need to move things. Please point me in the other PR the reasons and I'll check.

MiquelRForgeFlow:
@pedrobaeza Please, do the other PR when you are ready (ASAP is possible) :)

PR #1603, opened 2018-11-05
[10.0][FIX] hr_timesheet: pass active field from analytic accounts to new projects

MiquelRForgeFlow:
It doesn't make sense to create active projects from inactive analytic accounts.

--


LoisRForgeFlow (approved):
mekes sense

pedrobaeza (approved):

PR #1614, opened 2018-11-12
[10.0][FIX] fix fill_stock_move_unit_factor perf and use openupgrade.logged…

mourad-ehm:
When DB contains thousands (or millions) of mrp_production row migration of mrp can exceed 90h (more than 4 days). 

Current behavior before PR:
First: productions = env['mrp.production'].search([]) is not good because of number of records in data base
second: the query is executed in for loop one by one (because of qty fields wich is not stored)

To save time, we compute fields by sql queries and use mass update (by 2000 records at time)

### DIFF
--- a/addons/mrp/migrations/10.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/10.0.2.0/post-migration.py
@@ -8,7 +8,8 @@ from odoo.tools import DEFAULT_SERVER_DATE_FORMAT
 
 
 def populate_stock_move_quantity_done_store(cr):
-    cr.execute(
+    openupgrade.logged_query(
+        cr,
         """
         UPDATE stock_move
         SET quantity_done_store = product_uom_qty
@@ -20,7 +21,8 @@ def populate_stock_move_quantity_done_store(cr):
 
 
 def archive_mrp_bom_date_stop(cr):
-    cr.execute(
+    openupgrade.logged_query(
+        cr,
         """
         UPDATE mrp_bom
         SET active = FALSE
@@ -34,7 +36,8 @@ def archive_mrp_bom_date_stop(cr):
 
 def update_mrp_workorder_hour(cr):
     """It's a sum because the user used hours or used cycles."""
-    cr.execute(
+    openupgrade.logged_query(
+        cr,
         """
         UPDATE mrp_workorder
         SET duration_expected = (COALESCE(%s, 0) + COALESCE(%s, 0)) * 60
@@ -46,7 +49,8 @@ def update_mrp_workorder_hour(cr):
 
 def update_mrp_workcenter_times(cr):
     """The time now in minutes instead of hours."""
-    cr.execute(
+    openupgrade.logged_query(
+        cr,
         """
         UPDATE mrp_workcenter
         SET time_start = COALESCE(time_start, 0) * 60,
@@ -56,7 +60,8 @@ def update_mrp_workcenter_times(cr):
 
 
 def fill_mrp_routing_workcenter_time_cycle_manual(cr):
-    cr.execute(
+    openupgrade.logged_query(
+        cr,
         """
         UPDATE mrp_routing_workcenter
         SET time_cycle_manual = COALESCE(%s, 0) * COALESCE(%s, 1) * 60
@@ -175,7 +180,8 @@ def populate_production_picking_type_id(cr, loc_pt_map):
 
 
 def update_manufacture_procurement_rules(cr):
-    cr.execute(
+    openupgrade.logged_query(
+        cr,
         """
         UPDATE procurement_rule pr
         SET picking_type_id = spt.id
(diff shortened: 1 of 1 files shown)


addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -278,37 +284,110 @@ def populate_stock_move_workorder_id(cr):
...
+            'field_type': 'float',
+        })
+
+    # compute qty_produced for mrp_production
+
+    openupgrade.logged_query(cr, """
+        UPDATE mrp_production mp
+        SET %(field)s = qty_done_tb.quantity_done
+        FROM (
+        SELECT sum(sm.quantity_done) as quantity_done, sm.production_id FROM
```
MiquelRForgeFlow: runbot says `sm.quantity_done` column doesn't exist.

bealdav (approved):
Tested in several migrations +1

StefanRijnhart:
@mourad-ehm is this still WIP? Otherwise, can you change the title?

mourad-ehm:
Thanks @StefanRijnhart for your comment.  I change the title.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -278,37 +284,112 @@ def populate_stock_move_workorder_id(cr):
 
 def fill_stock_move_unit_factor(env):
```
StefanRijnhart, line 294: Took me some time to figure out what's happening here. Can you document here that the unit factor represents the rate of consumption of the raw material move lines given the produced quantities of the production order? And mention that your approach is to add temporary columns for computed fields that represent these produced quantities.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -278,37 +284,112 @@ def populate_stock_move_workorder_id(cr):
...
-            WHERE qty_produced IS NULL
-            """
-        )
-        ids = [r[0] for r in cr.fetchall()]
-        productions = env['mrp.production'].browse(ids)
-    productions._get_produced_qty()
+    # to fix performence issue, we use sql to compute quantity_done of
+    # stock move and qty_produced of mrp_production
+    quantity_done_field = 'quantity_done'
+    if not openupgrade.column_exists(cr, 'stock_move', 'quantity_done'):
```
StefanRijnhart: Even if this field exists, I'm thinking it would be wrong to take custody of it. For your purposes, I would *always* create the new column using get_legacy_name.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -278,37 +284,112 @@ def populate_stock_move_workorder_id(cr):
...
+        INNER JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        WHERE pp.id = sm.product_id
+        AND (pt.tracking = 'none' or pt.tracking IS NULL)
+        AND %(field)s IS NULL;
+        """ % {
+        'field': quantity_done_field,
+    })
+
+    qty_produced_field = 'qty_produced'
+    if not openupgrade.column_exists(cr, 'mrp_production', 'qty_produced'):
```
StefanRijnhart: Same, I would always create the new column using get_legacy_name.

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -278,37 +284,112 @@ def populate_stock_move_workorder_id(cr):
...
+        AND %(qty_produced_field)s IS NULL;
+        """ % {
+        'quantity_done_field': quantity_done_field,
+        'qty_produced_field': qty_produced_field,
+    })
+
+    mp_obj = env['mrp.production']
+    domain = [('state', 'not in', ['done', 'cancel'])]
+    offset = 0
+    while len(mp_obj.search(
```
StefanRijnhart: Chunking here is slower than just running the query once across all open production orders.

StefanRijnhart:
Looks good! I have some minor comments. Let me know what you think.

pedrobaeza:
@mourad-ehm are you going to answer the comments?

sebastienbeau:
I plan to work on this PR this month

legalsylvain:
@sebastienbeau, @mourad-ehm : any news regarding this work ?

thanks !

sebastienbeau:
I am back ! sorry for the delay. I start to work on it today !

sebastienbeau:
change are done, I am going to rerun a migration and give some feedback

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -369,36 +378,28 @@ def fill_stock_move_unit_factor(env):
...
-            END
-            FROM mrp_production mp
-            WHERE sm.raw_material_production_id = mp.id
-            AND sm.raw_material_production_id in %(mp_ids)s
-            """ % {
-            'field': qty_produced_field,
-            'mp_ids': tuple(productions.ids,),
-        })
-        offset += 2000
+    openupgrade.logged_query(cr, """
```
mourad-ehm, line 389: Depending on number of data to update, the mass update can become too long than update by block of 2000 rows.

sebastienbeau: I am re-running a migration so we will have data to compare really soon

pedrobaeza:
@StefanRijnhart do you agree with these changes then?

pedrobaeza:
@StefanRijnhart friendly reminder

StefanRijnhart (approved):

Issue #1613, opened 2018-11-12
Duplicated countries/states


dcorio:
Hello, not sure this can considered a bug but... using the latest migration script, while trying to migrate from 8.0 to 9.0,10.0,11.0, the migration fails because there's a constraint that prohibits to create duplicated countries or states.
Countries and states may exist because customers create them by hand.
Is there a common approach to this?

StefanRijnhart:
Thanks for noting! A patch to catch these duplicates would be welcome, I
think! Countries can easily be identified by their ISO code, so based on
that you could add an xml id in the pre-script.

pedrobaeza:
You can use openupgradelib merge records feature for merging countries with same code.

pedrobaeza:
Please identify in which version the constraint is added for adding the code in the proper branch.

dcorio:
Hi @pedrobaeza i wasn't aware of this "merge records feature". 
For instance I'm facing this problem migrating an 8.0 to 11.0 and the constraint is present in 10.0 and 11.0. Not on 9.0 AFAIK.
I can try to help here if you can point some examples

Many thanks in advance.

pedrobaeza:
Ok, then it seems you have to add the merge operation on OpenUpgrade 10.0. I would put in on the migration scripts of base module, but doing on pre-migration phase would lead that you can't use it with ORM mode, so this gets more difficult and not completed by SQL, so we can have the initial warning for not appearing anymore after the post-migration.

This is the method to call:

https://github.com/OCA/openupgradelib/blob/0ac601497c9849c3f257ca505bc00d28af368155/openupgradelib/openupgrade_merge_records.py#L407

and here you have an example with stock module in v11 for merging quants with same "significant key" (as v11 does):

https://github.com/OCA/OpenUpgrade/blob/168b1ab166b29955ead25b36428b464622df7710/addons/stock/migrations/11.0.1.1/end-migration.py#L14

MiquelRForgeFlow:
Solved in https://github.com/OCA/OpenUpgrade/pull/1619.

PR #1619, opened 2018-11-15
[10.0][FIX] Ensure no duplications in res.country.state model

MiquelRForgeFlow:
Solves https://github.com/OCA/OpenUpgrade/issues/1613.

In fact, solves several things:
- If manual data is introduced previously, then assures there is no duplicated data to avoid error of new constraint.
- If manual data is introduced previously, then solves that cvs v10 doesn't gives conflicts when migrating.
- Also, deletes marked "duplicated" data at the post-migration.

--

### DIFF
--- a/odoo/addons/base/migrations/10.0.1.3/openupgrade_analysis_work.txt
+++ b/odoo/addons/base/migrations/10.0.1.3/openupgrade_analysis_work.txt
@@ -68,6 +68,8 @@ DEL ir.ui.menu: base.menu_sale_config
 DEL ir.ui.menu: base.menu_sales
 NEW ir.ui.view: base.action_view_company_form_link_2_currencies
 NEW ir.ui.view: base.view_translation_dialog_tree
+# NOTHING TO DO
+
 NEW res.country.state: base.state_ae_aj
 NEW res.country.state: base.state_ae_az
 NEW res.country.state: base.state_ae_du
@@ -654,6 +656,8 @@ NEW res.country.state: base.state_za_nc
 NEW res.country.state: base.state_za_nl
 NEW res.country.state: base.state_za_nw
 NEW res.country.state: base.state_za_wc
+# DONE: post-migration: assured no duplications, due to new 'name_code_uniq' constraint
+
 NEW res.currency: base.TJS
 DEL res.currency: base.GWP
 DEL res.currency: base.SIT
--- a/odoo/addons/base/migrations/10.0.1.3/post-migration.py
+++ b/odoo/addons/base/migrations/10.0.1.3/post-migration.py
@@ -1,7 +1,9 @@
 # coding: utf-8
 # © 2017-2018 Opener BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from openupgradelib import openupgrade
+import csv
+from openupgradelib import openupgrade, openupgrade_merge_records
+from odoo.modules.module import get_module_resource
 
 
 def handle_im_odoo_support_views(env):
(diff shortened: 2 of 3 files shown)


JordiBForgeFlow:
The record to keep after the merge should always be the last created.

StefanRijnhart:
I understood #1613 differently. If a user creates a country that does not exist in an earlier version, the constraint on unique country codes is added without a problem, because the codes are still unique. But loading the base data, which now does contain that country will try to add another country with the same code because the XMLIDs don't match. This breaks the migration.

Your code does not fix that. I think a fix for that should assign XMLIDs to manually created countries and states by code before loading the base data.

JordiBForgeFlow:
@StefanRijnhart ahh that would solve it. I gyess that we should then read the csv data file, look for matches and then assign.

odoo/addons/base/migrations/10.0.1.3/end-migration.py
```diff
@@ -25,7 +26,26 @@ def convert_binary_attachment_style(env):
         openupgrade_90.convert_binary_field_to_attachment(env, field_spec)
 
 
+def merge_country_states(env):
+    group_list = [
+        'country_id', 'code',
```
pedrobaeza: I would deduplicate by code only.

pedrobaeza:
I think both are needed, and the most frequent case we are going to face is not the absent of a country that is now on v10, but the involuntary creation of duplicated countries, due to the infamous quick create action for example, you may end up with 3 "Spain" countries with code "ES", as the code is required. In that case, you need to deduplicate them.

For having the most fine tuned deduplication, we should add a new option to merge values dictionaries to the the first not falsy, so for example the first address format filled is preserved. @mreficent you should pass also this dictionary for changing default action over some fields, or you will end up with some undesired concatenated strings.

odoo/addons/base/migrations/10.0.1.3/end-migration.py
MiquelRForgeFlow: Why?

JordiBForgeFlow: There are codes that are used by multiple countries. We cannot do that

MiquelRForgeFlow:
> you should pass also this dictionary for changing default action over some fields, or you will end up with some undesired concatenated strings.

The only other field of the model is a `char`, and this type by default doesn't concatenate.

MiquelRForgeFlow:
@dcorio, could you test if this fix solves your problem?

odoo/addons/base/migrations/10.0.1.3/end-migration.py
pedrobaeza: OK, I have mixed countries with states, so I was talking all the time about countries. Clear now.

pedrobaeza:
@mreficent default operation for char fields is merge:

https://github.com/OCA/openupgradelib/blob/0ac601497c9849c3f257ca505bc00d28af368155/openupgradelib/openupgrade_merge_records.py#L246

so please change that.

MiquelRForgeFlow:
But this code says the contrary:

https://github.com/OCA/openupgradelib/blob/0ac601497c9849c3f257ca505bc00d28af368155/openupgradelib/openupgrade_merge_records.py#L290

pedrobaeza:
I see... please update docstring of the method.

MiquelRForgeFlow:
NOTE: All of you that want to test this PR, please update first your openupgradelib to latest changes (should include https://github.com/OCA/openupgradelib/pull/127).

MiquelRForgeFlow:
@pedrobaeza I have documented all methods and queries.

LoisRForgeFlow (approved):
I already tested it several times and it works good :+1:

JordiBForgeFlow (approved):

pedrobaeza:
This contains a lot of duplicated code that shouldn't be kept as is. You can call a method in one script from the other if you pretend to repeat the same check.

MiquelRForgeFlow:
And how I import a pre-migration method in the the post-migration file?

workspace 1_042

pedrobaeza:
The problem is the dash. You need to add another file, for example `tools.py`, and do on both scripts:
`from .tools import method_to_run`.

MiquelRForgeFlow:
Still some problems:

- case 1:

selection_044
(with empty init file)
```python
2018-11-30 17:27:30,762 11095 ERROR test_database odoo.modules.registry: Failed to load registry
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/registry.py", line 83, in new
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/loading.py", line 375, in load_modules
    report=report, models_to_check=models_to_check, upg_registry=upg_registry)
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/loading.py", line 146, in load_module_graph
    migrations.migrate_module(package, 'pre')
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/migration.py", line 154, in migrate_module
    mod = imp.load_module(name, fp, pathname, ('.py', 'r', imp.PY_SOURCE))
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/addons/base/migrations/10.0.1.3/pre-migration.py", line 6, in <module>
    from . import tools
ValueError: Attempted relative import in non-package
```
- case 2:

and also I cannot do `from odoo.addons.base.migrations.10.0.1.3` because the points and also because migration folder is not declared in base init file.

pedrobaeza:
But this is not what I have told you... You don't have to put `__init__.py` file, and the import sentence is `from .tools import <method>`. I haven't tried, but this is the theory.

dcorio:
> @dcorio, could you test if this fix solves your problem?

Hi @mreficent, I'm really sorry for the late reply. Even if I'm late, I want to thank you and the others for this fix. It definitely solved the issue I was facing!

MiquelRForgeFlow:
@dcorio thanks for your comment.

@pedrobaeza doing what you said, I have this error:
```python
2018-12-03 09:39:52,146 5791 ERROR test_database odoo.modules.registry: Failed to load registry
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/registry.py", line 83, in new
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/loading.py", line 375, in load_modules
    report=report, models_to_check=models_to_check, upg_registry=upg_registry)
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/loading.py", line 146, in load_module_graph
    migrations.migrate_module(package, 'pre')
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/modules/migration.py", line 154, in migrate_module
    mod = imp.load_module(name, fp, pathname, ('.py', 'r', imp.PY_SOURCE))
  File "/home/eficent/openupgrade/10.0/OpenUpgrade/odoo/addons/base/migrations/10.0.1.3/pre-migration.py", line 6, in <module>
    from .tools import ensure_country_state_id_on_existing_records
ValueError: Attempted relative import in non-package
```

pedrobaeza:
OK, it might not be possible without real hackings, but still I don't get the "just in case". Why do you want to repeat the exact same steps in post than in pre? Please put me a use case where that happens.

PR #1620, opened 2018-11-15
[10.0] update upstream 8a9569b31e

MiquelRForgeFlow:
Merging upstream Odoo up until commit https://github.com/odoo/odoo/commit/8a9569b31e503562c046722901dbc9c51e7b5118.

There were no conflicts.

--


StefanRijnhart:
Why do we need all these upstream merges?

StefanRijnhart:
I mean, given that Odoo has its stability policy, the things that are important to Openupgrade are not very likely to change after the release. And we could do with a little less of these to review.

MiquelRForgeFlow:
Yeah, it's true what you say but I like to do this to be more assured. There are a few migration scripts that instead of doing sql queries are using the environment to do some searchs and unlinks.

pedrobaeza (approved):

PR #1622, opened 2018-11-15
fix product_packaging INSERT query invalid syntax

rvalyi:
the SQL query to clone the product packaging lines from the template to the variants has a bad syntax. Of course if you don't have such packaging you won't see the issue. This patch fixes the syntax error.

cc @Benniphx

### DIFF
--- a/addons/product/migrations/11.0.1.2/post-migration.py
+++ b/addons/product/migrations/11.0.1.2/post-migration.py
@@ -34,11 +34,11 @@ def map_product_tmpl_id_to_product_id(cr):
             for p in product_rows[1:]:
                 fields = values = column_names
                 cr.execute("""
-                   INSERT INTO product_packaging %s
+                   INSERT INTO product_packaging(%s)
                    SELECT %s
-                   FROM product_packaging WHERE id = %%s""" % (
+                   FROM product_packaging WHERE id = %s""" % (
                     ','.join(fields + ['product_id']),
-                    ','.join(values + [p[0]])), (r[0],))
+                    ','.join(values + [str(p[0])]), r[0]))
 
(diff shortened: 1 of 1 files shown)


addons/product/migrations/11.0.1.2/post-migration.py
```diff
@@ -34,11 +34,11 @@ def map_product_tmpl_id_to_product_id(cr):
             for p in product_rows[1:]:
                 fields = values = column_names
```
pedrobaeza, line 35: If `len(fields) == len(values)` according to this, why are you passing in the SQL an extra `str(p[0])` column in the SELECT part?

pedrobaeza (changes requested):

addons/product/migrations/11.0.1.2/post-migration.py
rvalyi: well, we insert the fields ```','.join(fields + ['product_id'])```, that is ```len(fields) + 1``` (product_id has been filtered out earlier in the script) and the values we select are ```','.join(values + [str(p[0])])```, that is again ```len(fields) + 1``` (we basically copy all the values except product_id which is taken from an other earlier request. And finally the last argument r[0] is the product_packaging.id. Finally the str has been added to ensure p[0] which is an integer can be concatenated with another string. Sorry I don't see mistake, while I'm very sure the previous script was not correct. Are you sure about your remark?

pedrobaeza: Yes, sorry. I saw bad the first string replacement. You're right.

pedrobaeza (approved):

PR #1625, opened 2018-11-19
[11.0][FIX] website: Move social media data

pedrobaeza:
On v10, we had these social media links in website:

https://github.com/odoo/odoo/blob/258eaa7a3ce7c25e93f1b9940351335bfcda2aeb/addons/website/models/website.py#L170-L176

But on v11, they are non-stored, related fields to company:

https://github.com/odoo/odoo/blob/cac91731221b5ac122ad745ac408b406d335e5c7/addons/website/models/website.py#L55-L60

So we need to move the information from website to company.

@Tecnativa

### DIFF
--- a/addons/website/migrations/11.0.1.0/post-migration.py
+++ b/addons/website/migrations/11.0.1.0/post-migration.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
+from psycopg2.extensions import AsIs
 
 
 def fill_website_pages(env):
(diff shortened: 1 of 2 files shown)


rafaelbn (approved):

MiquelRForgeFlow:
But first you need to check if `mass_mailing` module is installed because you may have installed `website` but not `mass_mailing`. This company fields are defined in `mass_mailing`. See https://github.com/odoo/odoo/commit/e38e2515882dacc73ab7d87fa76da3c5a395c6a6.

pedrobaeza:
If you look to the link on the main comment, you'll see that these field are also declared in website module.

MiquelRForgeFlow:
Declared in `website` model in `website` module in v10. Declared in `res.company` model in `mass_mailing` module in v11. Thus, in the migration, before execute that query you should check if `mass_mailing` was installed.

pedrobaeza:
No, declared in `res.company` model in `website` module. Please check correctly before asserting things:

https://github.com/OCA/OpenUpgrade/blob/59abd3b221caf31bf08ffdb1fe37192032288616/addons/website/models/res_company.py#L11

MiquelRForgeFlow:
Oh, true. It's duplicated! hahaha

MiquelRForgeFlow (approved):

PR #1629, opened 2018-11-21
[11.0][MIG] mrp_repair

MiquelRForgeFlow:
Migration scripts for `mrp_repair`.

--


addons/mrp_repair/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,15 @@
+---Fields in module 'mrp_repair'---
+mrp_repair   / mrp.repair               / activity_ids (one2many)       : NEW relation: mail.activity
+mrp_repair   / mrp.repair               / website_message_ids (one2many): NEW relation: mail.message
+mrp_repair   / mrp.repair.fee           / to_invoice (boolean)          : DEL
```
pedrobaeza: How is handled now non invoiceable repairs?

MiquelRForgeFlow: Good question :S

LoisRForgeFlow: @pedrobaeza What do you mean? you still have the invoice method in the repair header, and you can set it to "no invoice".

pedrobaeza: It's not the same. In v10, you had the option to decide on each fee line if it's going to be invoiceable or not through this field. Now on v11 all fee lines are considered invoiceable, so someone migrating from v10 that has non invoiceable fees will find that in v11, that repairs will increase the total amount without doing anything, so the option can be:

- Remove from the table the non invoiceable lines (you can move them to another "backup" table - SQL allows to do directly from the creation of the table without having to define the schema).
- Put the total amount of that lines to 0 and move the real amount to another column.

LoisRForgeFlow: Ok, I follow you. I would go for the second option as it is what makes more sense in new version. If you want to have a line but do not want to charge for it, you have to set its amount to 0. 

Another new different behavior to note is that a part to be removed (type "Remove" in the tab parts [technically the field is called `operations`, how confusing as the next tab is called operations too but they are `fees_lines`...:sweat:]) are never added to the invoice, while in v10 they were.

pedrobaeza: Well, then it's interesting also that you fix this. I'm personally not a user of this module in any of my customers, but it's good to let this the most "pixel perfect" done.

MiquelRForgeFlow: Done. Please, update review!

LoisRForgeFlow (approved):
LGTM :+1:

addons/mrp_repair/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,40 @@
...
+        UPDATE mrp_repair_line
+        SET price_unit = 0
+        WHERE %s != TRUE
+        """, (AsIs(openupgrade.get_legacy_name('to_invoice')), ),
+    )
+    cr.execute(
+        """
+        UPDATE mrp_repair_fee
+        SET price_unit = 0
+        WHERE %s != TRUE
```
pedrobaeza: nitpicking:

```suggestion
        WHERE NOT %s
```

addons/mrp_repair/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+    )
+
+
+def reset_mrp_repair_price_unit(cr):
+    cr.execute(
+        """
+        UPDATE mrp_repair_line
+        SET price_unit = 0
+        WHERE %s != TRUE
```
pedrobaeza: nitpicking:

```suggestion
        WHERE NOT %s
```

pedrobaeza (changes requested):
You are not preserving the previous value in any place, so that's something destructive, which we avoid generally in OpenUpgrade. Before updating, you can copy column for avoiding this.

MiquelRForgeFlow:
Done! :)

pedrobaeza (changes requested):
* Please use `openupgrade.logged_query` on SQL statements.
* You can't rename `price_unit` field, as it's used when the line is invoiceable. You have to copy it.

MiquelRForgeFlow:
Done

pedrobaeza (approved):

PR #1630, opened 2018-11-21
rename_xmlids now takes a list

rvalyi:

l10n_fr migration is broken  because the openupgrade.rename_xmlids method was previously accepting a single tuple while now it requires a list of tuples.

Current behavior before PR:

breaks when running l10n_fr migration

it works

--


pedrobaeza (approved):
Signature hasn't changed, so I think this was incorrectly done from the beginning. Any way, merging when CIs finish

PR #1631, opened 2018-11-21
[11.0][FIX] Ensure no conflicts with constrain of res.country.state model

MiquelRForgeFlow:
v11 of https://github.com/OCA/OpenUpgrade/pull/1619.

This PR ensures no conflicts when migrating the `res.country.state` data with the new v11 csv data. As the existing constraint already assures that data is not duplicated, then it's not need any related post-migration script to delete duplications.

--

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
@@ -3,9 +3,11 @@
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+import csv
 from psycopg2.extensions import AsIs
 from odoo.addons.openupgrade_records.lib import apriori
 from openupgradelib import openupgrade
+from odoo.modules.module import get_module_resource
 
 
 # backup of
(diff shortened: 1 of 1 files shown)


odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -45,6 +47,121 @@ def fill_cron_action_server_pre(env):
     )
 
 
+def assure_country_state_id_on_existing_records(cr):
+    with open(get_module_resource('base', 'res', 'res.country.state.csv'),
+              newline='') as country_states_file:
+        states = csv.reader(country_states_file, delimiter=',', quotechar='"')
+        for row, state in enumerate(states):
```
pedrobaeza, line 57: Why are you doing this? It's done when loading the CSV by regular Odoo update process

pedrobaeza (changes requested):
I have started to review the PR, but seeing such a big amount of code, the only thing I can think on is that this is not needed. There's so much select, then updates... If everything is needed, add docstring and comments for illustrating the process. Don't be allergic to comments.

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
MiquelRForgeFlow: I need each csv record to check with and ensure to solve possible conflicts that later may appear when odoo loads the file

MiquelRForgeFlow:
@pedrobaeza I have documented all methods and queries.

LoisRForgeFlow (approved):
Works good :+1:

JordiBForgeFlow (approved):

pedrobaeza:
Please wait before merging this and this seems redundant to me.

MiquelRForgeFlow:
@pedrobaeza It have been tested with a client database that needed this because if not their migration fails. @lreficent can warranty that.

LoisRForgeFlow:
@pedrobaeza From an overall perspective, this fixed an issue we had on a migration. So from my end, it is more than fine. What is you concern here? is technically any issue with the approach? I think it is important to resolve this as this fix will likely be needed for v12 and future ones.

pedrobaeza:
The question is that on v10, you already have the constraint:

https://github.com/OCA/OpenUpgrade/blob/1b05ffea3546f31b022668d56a6dd5f8d2063bc8/odoo/addons/base/res/res_country.py#L102

so any native v10 DB won't have any duplicated state, so there's no need for deduplicating again. For v9 > v10 migration, there can be the need as the constraint was introduced and the DB might contain these duplicates, but not here. If you have the case is because you have a migrated DB without the deduplication, so the constraint was not applied and the violations of it might increase.

And finally, I don't see new country states in the file that may overlap with manual created countries, except Colombian ones, that are not covered right now by Geonames, so no special problem.

v10: https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/base/res/res.country.state.csv
v11: https://github.com/OCA/OpenUpgrade/blob/11.0/odoo/addons/base/res/res.country.state.csv

MiquelRForgeFlow:
In fact, if I remember correctly, our problem was because we had introduced manually the Colombian ones haha

pedrobaeza:
hahaha, well, that's a total serendipity... Well, it's a lot of code for only that remote case, but I won't oppose if you want to merge, except a couple of points that I'm reviewing right now.

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -45,6 +47,107 @@ def fill_cron_action_server_pre(env):
...
+                SET name = '%(data_name)s', module = 'base'
+                WHERE id = %(data_id)s AND model = 'res.country.state'
+                """ % {
+                    'data_name': data_name,
+                    'data_id': found_id[0],
+                }
+            )
+            cr.execute(
+                """
+                UPDATE res_country_state rcs
```
pedrobaeza, line 138: Don't do that, as when loading the CSV it will be automatically reloaded again

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -45,6 +47,107 @@ def fill_cron_action_server_pre(env):
...
+        for row, state in enumerate(states):
+            if row == 0:
+                continue
+            data_name = state[0]
+            country_code = state[1]
+            name = state[2]
+            state_code = state[3]
+            # first: query to ensure the existing odoo countries have
+            # the code of the csv file, because maybe some code has changed
+            cr.execute(
```
pedrobaeza, line 66: This is not needed, as the pre-requisite is always to have a DB of the previous version totally updated, and this data is noupdate="0" (being a CSV), so no possible problem. If any problem, go to first step ;)

pedrobaeza:
I think the general approach is incorrect, as it has a lot of useless code. What you have to do is:

- Search in one query all the `res_id` for `ir.model.data` that points to `res.country.state`.
- Get existing `res.country.state` ids.
- Substracting 2nd from 1st, you get records without XML-IDs.
- That records are the one that you have to compare with new CSV to see if any matches.
- And that should be done preloading in a dictionary the CSV with key (country_id, code). The rest is useless.
- If matched, then add the ir.model.data record, but don't do anything more. Odoo update process will do it.

pedrobaeza:
Let's close this as it's a very specific case and the code is a bit convoluted.

PR #1637, opened 2018-11-26
[FIX] purchase_requisition post-migration create product only if necessry & fix bug categ_id required

mourad-ehm:
product creation fails because categ_id is required. 
update is only for 
Current behavior before PR:
we create a new product even update is not applicable (no purchase_requisition_line with product_id IS NULL)
Create product only if there is some purchase_requisition_line with product_id IS NUL.
Fix bug categ_id required for product creation.

### DIFF
--- a/addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
+++ b/addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
@@ -8,16 +8,24 @@ from openupgradelib import openupgrade
 
 def fill_product(env):
     """Fill with an pre-created product as it's required."""
-    product = env['product.product'].create({
-        'active': False,
-        'name': 'OU purchase requisition line wildcard',
-    })
-    openupgrade.logged_query(
-        env.cr, """
-        UPDATE purchase_requisition_line
-        SET product_id = %s
-        WHERE product_id IS NULL;""", (product.id, ),
-    )
+    if env['purchase.requisition.line'].search_count(
+            [('product_id', '=', False)]) > 0:
+        categ_id = env["product.template"]._get_default_category_id()
+        if not categ_id:
+            categ_id = env["product.category"].search(
+                [], limit=1, order='id asc').id
+        product = env['product.product'].create({
+            'active': False,
+            'categ_id': categ_id,  # categ_id is required
+            'company_id': False,  # tacke into account multi-company context
+            'name': 'OU purchase requisition line wildcard',
+        })
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE purchase_requisition_line
+            SET product_id = %s
+            WHERE product_id IS NULL;""", (product.id, ),
+        )
 
 
 @openupgrade.migrate()


mourad-ehm:
ping @pedrobaeza , @bealdav

bealdav (approved):

StefanRijnhart:
Can you check travis
```
flake8 --max-line-length=120 $(find . \( -name 'pre-*.py' -or -name 'post-*.py' \) -exec grep -q openupgrade {} \; -print)
./addons/purchase_requisition/migrations/10.0.0.1/post-migration.py:14:27: E261 at least two spaces before inline comment
The command "flake8 --max-line-length=120 $(find . \( -name 'pre-*.py' -or -name 'post-*.py' \) -exec grep -q openupgrade {} \; -print)" exited with 1.
```
 and, as an additional fix, can you set 'company_id' to False in the created product? It should work better in multicompany context.

mourad-ehm:
Hi @StefanRijnhart  I fixed flake8. Thanks

addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
```diff
@@ -8,16 +8,19 @@
...
-        UPDATE purchase_requisition_line
-        SET product_id = %s
-        WHERE product_id IS NULL;""", (product.id, ),
-    )
+    if env['purchase.requisition.line'].search_count(
+            [('product_id', '=', False)]) > 0:
+        product = env['product.product'].create({
+            'active': False,
+            'categ_id': 1,  # categ_id is required
+            'name': 'OU purchase requisition line wildcard',
```
StefanRijnhart, line 21: Thanks for fixing flake8! Can you insert an additional fix for me? When creating utility products in other cases, we found out that we need to set 'company_id' to False to cover multicompany contexts. Can you do that here?

mourad-ehm: Ok

mourad-ehm: Could you approve this PR ? Thanks

addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
```diff
@@ -8,16 +8,20 @@
...
-        env.cr, """
-        UPDATE purchase_requisition_line
-        SET product_id = %s
-        WHERE product_id IS NULL;""", (product.id, ),
-    )
+    if env['purchase.requisition.line'].search_count(
+            [('product_id', '=', False)]) > 0:
+        product = env['product.product'].create({
+            'active': False,
+            'categ_id': 1,  # categ_id is required
```
StefanRijnhart: Sorry, I only now spotted the hardcoded id. Can you replace that with `env.ref('product.product_category_all').id`?

StefanRijnhart: In fact, that category is selected by default (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/product/models/product_template.py#L19-L23). Are you sure you need to insert it explicitly here?

MiquelRForgeFlow:
@mourad-ehm could you squash commits, please?

mourad-ehm:
squash OK. Could you merge it, please?

addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
MiquelRForgeFlow: shouldn't this be
```python
'categ_id': env["product.product"]._get_default_category_id(),
```
?

StefanRijnhart: But it's called by default (so I'd think that no value is necessary)

mourad-ehm: I make change for default_category.

StefanRijnhart (approved):

addons/purchase_requisition/migrations/10.0.0.1/post-migration.py
mourad-ehm: Openupgrade is used for migrate custom database. We are not sure to have ref('product.product_category_all') or default values (bug in my case).

pedrobaeza: @mourad-ehm that category is required in all Odoo databases and it's recreated each time you upgrade product module, so this not a problem and the first requirement to run OpenUpgrade is that you have consistency in your DB (which is assured in some cases updating your DB modules).

pedrobaeza: @StefanRijnhart you have merged this with the ID hardcoded.

pedrobaeza: Ah, sorry, this comment is outdated. I would prefer the reference, but OK.

PR #1639, opened 2018-11-29
[FIX+IMP] stock: Handle receptions not done + fix grouping on done

pedrobaeza:
* stock.move.line creation from done moves should group by stock move.
* Receptions not yet validated now have stock.move.line for not affecting normal flow clicking directly on validate button.

@Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -240,7 +240,7 @@ def create_stock_move_line_incoming(env):
             INNER JOIN stock_picking_type spt ON spt.id = sp.picking_type_id
         WHERE sm.state = 'assigned'
             AND sm.origin_returned_move_id IS NULL
-            AND spt.code == 'incoming'
+            AND spt.code = 'incoming'
         """,
     )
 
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -188,14 +184,73 @@ def create_stock_move_line(env):
...
+            COALESCE(sp.name, sm.name),
+            'assigned',
+            sm.write_date,
+            sm.write_uid
+        FROM stock_move sm
+            INNER JOIN stock_picking sp ON sp.id = sm.picking_id
+            INNER JOIN stock_picking_type spt ON spt.id = sp.picking_type_id
+        WHERE sm.state = 'assigned'
+            AND sm.origin_returned_move_id IS NULL
+            AND spt.code == 'incoming'
```
MiquelRForgeFlow: s/==/=

pedrobaeza: hehe, right, Python deformation

pedrobaeza:
Done directly in GitHub. To be squashed on merge.

MiquelRForgeFlow (approved):

PR #1640, opened 2018-12-01
[FIX] stock: Don't fail on incompatible conversion

pedrobaeza:
After stock.move.line creation, all records are checked to see if the have different UoM from product one, and re-compute `product_qty`. If you have changed improperly the unit of measure of the product to an incompatible UoM (this was possible to be made on <= v8, or now deactivating the product, changing the UoM, and activating it again), the conversion will fail.

We enclose it in a try for ignoring these conversions.

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -2,6 +2,7 @@
 # Copyright 2017-2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from odoo.exceptions import UserError
 from openupgradelib import openupgrade
 from psycopg2.extensions import AsIs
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@mreficent maybe we don't need this method at all?

addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -319,15 +320,18 @@ def recompute_stock_move_line_qty_different_uom(env):
...
-        env.cr.execute(
-            "UPDATE stock_move_line SET product_qty = %s WHERE id = %s",
-            (product_qty, line.id),
-        )
+        try:  # Don't fail on incompatible conversions
+            product_qty = line.product_uom_id._compute_quantity(
+                line.product_uom_qty, line.product_id.uom_id,
+                rounding_method='HALF-UP',
+            )
+            # Can't assign by ORM, so by SQL
```
MiquelRForgeFlow, line 328: why can't assign by ORM?

MiquelRForgeFlow (approved):
But LGTM

addons/stock/migrations/11.0.1.1/post-migration.py
pedrobaeza: Because there's associated code to the `write` on stock.move.line that makes a lot of reserve/unreserve operations on quants that will mess up quant state.

pedrobaeza:
What about the other question?

MiquelRForgeFlow:
I don't know with 100% security if it's needed or not. I see it as a "just in case" and better it than not to have it. I always see good any method that checks if something is wrong and automatically corrects it.

pedrobaeza:
OK, I prefer to not overload processes and have several point of failures, but let's move on this case.

PR #1646, opened 2018-12-04
[10.0] sale_margin : nothing to do

legalsylvain:
mark as nothing to do sale_margin.

Don't know why the migration folder is not present, but the two simple modules are indentical. (just migrated into new API).

V9 : https://github.com/odoo/odoo/blob/9.0/addons/sale_margin/sale_margin.py
V10 : https://github.com/odoo/odoo/blob/10.0/addons/sale_margin/models/sale_order.py


pedrobaeza:
There's no `migrations` folder because there isn't any change. It's created when there's any change and `openupgrade_analysis.txt` file has to be created.

pedrobaeza:
Please avoid to make the patch directly on OCA branch as this triggers 2 Travis builds that are very heavy.

legalsylvain:
> Please avoid to make the patch directly on OCA branch as this triggers 2 Travis builds that are very heavy.

yes sorry. It occures when I use the github interface. Do you know how to force github to create the branch in /legalsylvain instead of OpenUpgrade ?

regards.

pedrobaeza:
I'm afraid GitHub doesn't offer this for now, only to change email address for committing.

PR #1648, opened 2018-12-05
9.0 imp account migration by sql

legalsylvain:
* add a new function fast_create, to create and populate a new field, by SQL.

* fast create of ``account_invoice_tax.currency_id``, ``account_bank_statement.difference``, ``account_invoice.amount_total_signed``

* fast create of ``account_invoice.amount_total_company_signed``, ``account_invoice.amount_untaxed_signed`` if there are not invoices with currency different from that of company.

migration of account is slow

make the migration faster

note : the function fast_create could be integrated in openupgradelib.

--

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -384,7 +384,6 @@ def fast_create(env, settings):
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     cr = env.cr
-
     # 9.0 introduces a constraint enforcing this
     cr.execute(
         "update account_account set reconcile=True "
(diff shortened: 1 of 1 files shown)


rvalyi:
Hello, I can confirm the migration completes using this PR. Reviewing account things quickly it seems it works well. However I cannot claim I did some deep verification yet. Also it seems to be faster. My migration went form around 4h (approximative measure) down to 3 hours. I also ran that on a different hardward but that was going from my laptop i7-5600U CPU @ 2.60GHz to a lower end Intel Core Processor Haswell 2Gz. So there is no doubt left it is faster. So thanks. Would be cool if others can confirm this PR works for them too.
cc @mourad-ehm @bealdav that PR may interest you

pedrobaeza (approved):
Checked this after a long delay. Sorry for that.

Issue #1521, opened 2018-09-14
[11.0][base] Company logo is reset after migration


JordiBForgeFlow:
@mreficent Can you review?

cc @pedrobaeza

Ark74:
Can confirm.
Using c83a78bdeba39c8de1d9ce5937aa068bbe39fb55
From 10 to 11

pedrobaeza:
Fixed in #1649

PR #1649, opened 2018-12-05
[11.0][FIX] base: don't reset company logo

MiquelRForgeFlow:
I don't understand why we didn't detected this before.

Fixes https://github.com/OCA/OpenUpgrade/issues/1521.

--

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/noupdate_changes.xml
+++ b/odoo/addons/base/migrations/11.0.1.3/noupdate_changes.xml
@@ -417,9 +417,11 @@
     <field name="currency_subunit_label">Centime</field>
     <field name="currency_unit_label">Franc</field>
   </record>
+<!-- Disabled because may affect somebody into resetting their company logo
   <record model="res.partner" id="main_partner">
     <field name="image" type="base64" file="base/static/img/main_partner-image.png"/>
   </record>
+-->
   <record model="res.currency" id="SLL">
     <field name="currency_subunit_label">Cents</field>
     <field name="currency_unit_label">Leone</field>


rvalyi (approved):

cubells (approved):
:+1: 

LGTM

odoo/addons/base/migrations/11.0.1.3/noupdate_changes.xml
```diff
@@ -417,9 +417,9 @@
     <field name="currency_subunit_label">Centime</field>
     <field name="currency_unit_label">Franc</field>
   </record>
-  <record model="res.partner" id="main_partner">
-    <field name="image" type="base64" file="base/static/img/main_partner-image.png"/>
-  </record>
+  <!--<record model="res.partner" id="main_partner">-->
```
pedrobaeza: Simply remove it instead of commenting it.

MiquelRForgeFlow: I did it before, but then I thought that if commented, then if in the future we update this file, we will know this should not be added again. Thus, I prefer commenting it.

pedrobaeza: I also thought about that, but I'm not sure if it's going to be effective, as you generate the whole file from the external tool and overwrite it. You should check manually the diff to see this difference. Any way, if we keep it:

1. Enclose all in one comment, not one comment per line.
2. Put an explanatory text in the beginning of the comment.

pedrobaeza (approved):

PR #1650, opened 2018-12-05
[10.0][FIX] mrp: mrp_production state is 'planned' only if has workorders

MiquelRForgeFlow:
Fixes mo states.

--

### DIFF
--- a/addons/mrp/migrations/10.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/10.0.2.0/post-migration.py
@@ -78,11 +78,20 @@ def map_mrp_production_state(cr):
         """
     )
     draft_ids = [r[0] for r in cr.fetchall()]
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow (approved):

pedrobaeza (changes requested):
I don't get why you need this. Is it for not changing manufacturing orders without work orders? Then just do it in one query:
```python
    cr.execute(
        """
        UPDATE mrp_production mp
        SET state = 'planned'
        FROM mrp_workorder mw
        WHERE mp.state = 'confirmed'
            AND mp.id = mw.production_id
        """
    )
```
And please, ALWAYS add comment on code for explaining what you are doing.

MiquelRForgeFlow:
Updated :)

pedrobaeza (approved):

PR #1652, opened 2018-12-06
[10.0][ADD] OCA implementation of hr_timesheet_invoice

hbrunn:
this is currently more RFC than a serious PR: In v8, hr_timesheet_invoice was deprecated, and I created a [functional equivalent](https://github.com/OCA/account-analytic/pull/213) for v10 (let's not discuss if I'm better off with the standard sale order way, I'm not).

Given there won't be a version 9 of this (at least from my side), this is the only place I can put this in order to work, but conceptually it seems wrong to me because the module was dropped already in 9. Any opinion on this?

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -31,6 +31,8 @@ renamed_modules = {
     # OCA/stock-logistics-workflow:
     'product_customer_code_picking':
         'product_supplierinfo_for_customer_picking',
+    # OCA/account-analytic:
+    'hr_timesheet_invoice': 'account_invoice_analytic_line',
 }
 
 merged_modules = [


pedrobaeza:
Well, you can do the same on v9 with the same effect, as I deduce you are migrating from v8 to v10, but any way, there's no special problem to leave it here. If some improbable day someone migrate the module to v9, we can change this to v9.

hbrunn:
thanks, that makes things convenient. Then I'll leave this in work in progress until the above is merged

pedrobaeza:
Any news on this?

hbrunn:
not enough reviews for merging https://github.com/OCA/account-invoicing/pull/462 yet

pedrobaeza:
As this is not finished and I'm not completely agree with the raw renaming, as you can have installed the Odoo module, but not because you are invoicing from analytic lines, but as dependency (due to weird dependency scheme in this version), so I'm closing this for clarity and can be reopened later.

PR #1654, opened 2018-12-10
[FIX] account.analytic.line#partner_id is not the user's partner, but the analytic account's

hbrunn:


pedrobaeza:
@hbrunn are you sure that timesheet entries from users (time spent) have that partner in the lines instead of analytic account partner?

hbrunn:
yes

pedrobaeza:
Do you have any code line link for asserting that? I'm having a hard moment for finding where this field is declared and populated. I think this method shouldn't be in `analytic` module.

hbrunn:
https://github.com/OCA/OCB/blob/9.0/addons/account/models/account_analytic_line.py#L25

It does occur to me that the field might already be defined by some community / private model, so we might be better off adding `where account_id is not null and parent_id is null`

pedrobaeza:
But that's on 9.0. What about 8.0?

hbrunn:
it doesn't exist. From the the point of view of `analytic`, it's a new field

pedrobaeza:
That's why I say that this will fail if the field doesn't exist in v8, and less done in this module (the field is defined in v9 account). So my doubt is: if the field doesn't exist in v8, when you upgrade `account` module in v9, the field will be created and populated with that related value, so no need of this. If it's created in other place, then it should be in v8 somewhere, and the code should be executed somewhere else.

hbrunn:
no it won't: https://github.com/OCA/OCB/blob/9.0/addons/analytic/models/analytic.py#L60

pedrobaeza:
OK, that's the missing piece! Thanks for the patience with me. The extra stuff `WHERE partner_id IS NULL` can be very interesting for not overriding possible values having `partner_id` field declared on community values.

hbrunn:
no worries, I think it's a good thing to be critical on all input, everyone fucks up eventually and that we don't want in our code tree.

Added the `partner_id IS NULL` part and we politely forget about the nonsense I wrote about `account_id`

pedrobaeza (approved):

PR #1655, opened 2018-12-10
[12.0][FIX] migration of base

MiquelRForgeFlow:
This PR does:

- Updates the noupdate_changes file.
- Fills correctly base_onboarding_company_state field of res.company

--

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
+++ b/odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
@@ -1,17 +1,174 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
-  <record id="gt" model="res.country">
-    <field name="currency_id" ref="GTQ"/>
+  <record id="cc" model="res.country">
+    <field eval="61" name="phone_code"/>
   </record>
-  <record id="sg" model="res.country">
-    <field name="vat_label">GST No.</field>
+  <record id="aq" model="res.country">
+    <field eval="672" name="phone_code"/>
+  </record>
+  <record id="kn" model="res.country">
+    <field eval="1869" name="phone_code"/>
+  </record>
+  <record id="mp" model="res.country">
+    <field eval="1670" name="phone_code"/>
+  </record>
+  <record id="eh" model="res.country">
+    <field eval="212" name="phone_code"/>
+  </record>
+  <record id="do" model="res.country">
+    <field eval="1849" name="phone_code"/>
+  </record>
+  <record id="gu" model="res.country">
+    <field eval="1671" name="phone_code"/>
+  </record>
+  <record id="nf" model="res.country">
+    <field eval="672" name="phone_code"/>
+  </record>
+  <record id="pn" model="res.country">
+    <field eval="64" name="phone_code"/>
+  </record>
+  <record id="sx" model="res.country">
+    <field eval="1721" name="phone_code"/>
+  </record>
+  <record id="ax" model="res.country">
+    <field eval="358" name="phone_code"/>
+  </record>
+  <record id="im" model="res.country">
+    <field eval="44" name="phone_code"/>
+  </record>
+  <record id="bm" model="res.country">
+    <field eval="1441" name="phone_code"/>
+  </record>
+  <record id="lc" model="res.country">
+    <field eval="1758" name="phone_code"/>
+  </record>
+  <record id="bq" model="res.country">
+    <field eval="599" name="phone_code"/>
+  </record>
+  <!-- not needed (because end-migration)
+  <record id="user_root" model="res.users">
+    <field name="signature"><![CDATA[<span>&#45;&#45; <br/>
+System</span>]]></field>
+  </record>
+  -->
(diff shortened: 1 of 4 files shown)


odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
```diff
@@ -1,17 +1,174 @@
...
+    <field eval="1268" name="phone_code"/>
+  </record>
+  <record id="bl" model="res.country">
+    <field eval="590" name="phone_code"/>
+  </record>
+  <record id="bv" model="res.country">
+    <field eval="55" name="phone_code"/>
+  </record>
+  <!--
+  <record id="partner_root" model="res.partner">
```
pedrobaeza, line 165: Why do you comment this? I see this important for following Odoo's paradigm

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
```diff
@@ -1,17 +1,174 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
-  <record id="gt" model="res.country">
-    <field name="currency_id" ref="GTQ"/>
```
pedrobaeza, line 4: Isn't this country finally assigned that currency?

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
MiquelRForgeFlow: I just commented it because I wasn't sure. I will put again now :)

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
MiquelRForgeFlow: Yes, but was also assigned in v11, so obviously disappears now from here.

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
```diff
@@ -1,17 +1,172 @@
...
+    <field eval="1441" name="phone_code"/>
+  </record>
+  <record id="lc" model="res.country">
+    <field eval="1758" name="phone_code"/>
+  </record>
+  <record id="bq" model="res.country">
+    <field eval="599" name="phone_code"/>
+  </record>
+  <!--
+  <record id="user_root" model="res.users">
```
MiquelRForgeFlow, line 49: @pedrobaeza Should I decomment also this one?

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
MiquelRForgeFlow: Travis failed because of decommenting this.
Error: "You cannot archive a contact linked to an internal user."

pedrobaeza: Then you should do it by SQL.

pedrobaeza: Or see if the problem is that the user is not inactive.

MiquelRForgeFlow: The user (base.user_root) is active

pedrobaeza: And the question is: should it be? How is in v12 default?

MiquelRForgeFlow: in v11, both partner and user were active. Is in v12 that the partner becomes inactive. I think the way to handle it it's to let the partner be active in this file but later put inactive with sql.

pedrobaeza: OK, good for me that way.

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
pedrobaeza: What the signature contains?

odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
MiquelRForgeFlow: ah, no, I see now. the user also becomes inactive in v12. So the correct way to handle it it's add the active = false and let the partner be as it is.

pedrobaeza: OK

MiquelRForgeFlow: After some checking, I concluded that this should be commented instead, because the user_root/user_admin/partner_root/partner_admin thing is handled in the pre-migration and in the end-migration. It was done by Stefan.

pedrobaeza: OK, then put these conclusions here as well for your future "you" and the rest of us.

PR #1656, opened 2018-12-11
[10.0] Merge website_blog_mgmt into website_blog

chienandalu:
`website_blog_mgmt` is deprecated from 10.0 since there's Odoo core functionality for it.

cc @Tecnativa

### DIFF
--- a/addons/website_blog/migrations/10.0.1.0/pre-migration.py
+++ b/addons/website_blog/migrations/10.0.1.0/pre-migration.py
@@ -3,7 +3,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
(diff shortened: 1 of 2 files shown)


addons/website_blog/migrations/10.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+from odoo import tools
```
MiquelRForgeFlow: delete this line

pedrobaeza (approved):
I have removed the not used variable, so I merge.

PR #1657, opened 2018-12-11
[10.0][FIX] mrp: bom_line_id was wrongly mapped.

LoisRForgeFlow:
This field is needed when updating qty to produce, if not correctly set a error will raise, the query I'm proposing solved the issue.

@mreficent @jbeficent

### DIFF
--- a/addons/mrp/migrations/10.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/10.0.2.0/post-migration.py
@@ -324,17 +324,21 @@ def fill_stock_move_bom_line_id(cr):
     openupgrade.logged_query(
         cr,
         """
-        UPDATE stock_move sm
+        UPDATE stock_move sm_update
         SET bom_line_id = mbl.bom_line_id
-        FROM mrp_production mp
-        LEFT JOIN mrp_bom mb ON mp.bom_id = mb.id
-        LEFT JOIN (SELECT min(id) AS bom_line_id, bom_id, product_id
-                   FROM mrp_bom_line
-                   GROUP BY bom_id, product_id
-                   ) mbl ON (mbl.bom_id = mb.id
-                                AND mbl.product_id = mp.product_id)
-        WHERE sm.production_id = mp.id AND sm.product_id = mp.product_id
-            AND sm.bom_line_id IS NULL
+        FROM stock_move sm
+            INNER JOIN mrp_production mp
+            ON sm.raw_material_production_id = mp.id
+            LEFT JOIN (
+                SELECT min(id) AS bom_line_id, bom_id, product_id
+                FROM mrp_bom_line
+                GROUP BY bom_id, product_id
+                ) mbl ON (
+                    mbl.bom_id = mp.bom_id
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow (approved):

addons/mrp/migrations/10.0.2.0/post-migration.py
```diff
@@ -324,17 +324,21 @@ def fill_stock_move_bom_line_id(cr):
...
+            ON sm.raw_material_production_id = mp.id
+            LEFT JOIN (
+                SELECT min(id) AS bom_line_id, bom_id, product_id
+                FROM mrp_bom_line
+                GROUP BY bom_id, product_id
+                ) mbl ON (
+                    mbl.bom_id = mp.bom_id
+                    AND mbl.product_id = sm.product_id)
+        WHERE sm_update.raw_material_production_id = mp.id
+            AND sm_update.product_id = sm.product_id
```
pedrobaeza: This join is very weak and will cross a lot of records.

JordiBForgeFlow: @pedrobaeza Why weak?

pedrobaeza: Because you are crossing all the moves that has the same product only, without a strong condition.

JordiBForgeFlow: Not all moves. The inner join with mrp_production reduces the query very significantly

pedrobaeza: See that `stock_move` is used 2 times, one with `sm_update` and another with just `sm`, and the join for both is the `product_id` field, so you are crossing all the moves with a certain product with themselves.

LoisRForgeFlow: I see what you mean, would you like `sm_update.id = sm.id` better?

pedrobaeza: Yes. I suppose you need the second `stock_move` table for using it on the JOIN. Then that's the correct link.

But I still doubt this is the correct approach, as you are still opening this to all the possible BoM lines, instead of those linked through the BoM selected on the manufacturing order. Not only that, I wonder if you have added manually components that are not on the BoM...

LoisRForgeFlow: I'm using the BoM in the MO: https://github.com/OCA/OpenUpgrade/pull/1657/files#diff-a4ece93235eea01f8f86ac1f94e62a7bR337. Why would I have added components?? this is an update of stock_move, isn't it?

pedrobaeza: OK, right, then just change the link between stock_move tables.

LoisRForgeFlow: Done!

pedrobaeza (approved):

PR #1661, opened 2018-12-13
[12.0][ADD] apriori: mpr_repair renamed models

MiquelRForgeFlow:
Also reordered them alphabetically.

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -26,15 +26,19 @@ merged_modules = {
 
 renamed_models = {
     # Odoo
+    'hr.holidays': 'hr.leave',
+    'hr.holidays.status': 'hr.leave.type',
+    'mrp.repair': 'repair.order',
+    'mrp.repair.fee': 'repair.fee',
+    'mrp.repair.line': 'repair.line',
     'procurement.rule': 'stock.rule',
     'product.attribute.line': 'product.template.attribute.line',
     'product.attribute.price': 'product.template.attribute.value',
     'product.uom': 'uom.uom',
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
@pedrobaeza I think that now I did not leave any model rename 😅

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -26,15 +26,19 @@
...
     'product.uom.category': 'uom.category',
-    'hr.holidays': 'hr.leave',
-    'hr.holidays.status': 'hr.leave.type',
-    'sale.quote.template': 'sale.order.template',
-    'sale.quote.option': 'sale.order.template.option',
     'sale.quote.line': 'sale.order.template.line',
+    'sale.quote.option': 'sale.order.template.option',
+    'sale.quote.template': 'sale.order.template',
     'stock.incoterms': 'account.incoterms',
+    'stock.location.path': 'stock.rule',
```
MiquelRForgeFlow: I don't know if this will give problems. Having two models renamed into a new one.

pedrobaeza: Yes. You should rename one of them, and merge the other one in this. Any way, is this model renaming being called in base module right now?

MiquelRForgeFlow: No, it's just only used in the openupgrade analysis.

pedrobaeza: The error would raise there, but any way, I think it's not a good idea to put both for the analysis.

MiquelRForgeFlow: Ok, I commented it instead.

PR #1662, opened 2018-12-14
[12.0][MIG] analytic

MiquelRForgeFlow:
Migration scripts for `analytic` module.

--


pedrobaeza (approved):
Fixed 2 things:

* The column that we have to copy for preserving previous data is the one for `account.analytic.line`, not for `account.analytic.account`.
* We have to set the company unconditionally from analytic account, as it's a required field in that table, and otherwise we can't have inconsistent data that are checked in a Python constraint. I don't understand why Odoo didn't put clearly here a related field, and instead constrains to only have as valid value the same as the related one.

pedrobaeza:
Still blocked by `uom` dependency that is a new one and thus I didn't have in my system.

PR #1663, opened 2018-12-14
[12.0][MIG] event

MiquelRForgeFlow:
Migration scripts for `event` module.

--


MiquelRForgeFlow:
hahaha it's green despite a lot of its dependencies don't have the migration scripts

pedrobaeza:
Yes, that can happen if the depending modules doesn't break normal update process, although data are not correct (or maybe yes...). Any way, it should be checked in the hierarchy order.

MiquelRForgeFlow:
I see it as a minus minus = plus case :dancer:

pedrobaeza:
I have mixed PRs and merge this one accidentally. There's missing the mail template translation reset. I'll do it directly in a commit.

pedrobaeza:
Done in 235b2acfeb259d8008db272c8a66608814ac8f0d

I have to check if translations way of being stored have been changed in this version so `delete_record_translations` might not work correctly.

PR #1664, opened 2018-12-14
[12.0][MIG] mass_mailing

MiquelRForgeFlow:
Migration scripts for `mass_mailing` module.

--


pedrobaeza:
Fixed migration scripts, as the opt_out is not filled by blacklist, but a field in the `rel` table.

Please remove the habit of putting `use_env=False`, as this reduces the possibilities to do with migration scripts. If you want to not put `env.cr`, which is any way very short, put the initial `cr = env.cr` line.

Please also use ALWAYS `openupgrade.logged_query` for better output and control, unless this hammers a lot the log because it's a sequential operation.

PR #1666, opened 2018-12-15
[FIX] stock: Used stock source location instead of move location for reserved

pedrobaeza:
On reserved moves, the right location is the quant one, as you are able to indicate
a parent location in the move, and then the reserve takes place in a children one.

cc @Tecnativa


JordiBForgeFlow (approved):
Tested and worka as expected!

PR #1673, opened 2019-01-08
[12.0][MIG] social_media

MiquelRForgeFlow:
`#NothingToDo`

--


pedrobaeza (approved):
`mass_mailing` was depending on that one, but as there's nothing to do, the result is the same. It would be needed to use https://github.com/OCA/openupgradelib/blob/67ebfa8fa77ea404a1c0ce711b7bc37da2ecaf23/openupgradelib/openupgrade.py#L1921 if there wasn't that dependency, but it's not the case.

PR #1674, opened 2019-01-08
[12.0][MIG] project & pad_project

MiquelRForgeFlow:
Migration scripts for `project` module.

--


rvalyi:
Hello, I'm trying on a live database upgrading from v8, v9, v11 and then v12 and I'm getting this stack trace with this project script:

```
  File "/odoo/src/addons/project/migrations/12.0.1.1/post-migration.py", line 9, in migrate
    cr, 'project', 'migrations/12.0.1.1/noupdate_changes.xml')
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 296, in load_data
    tools.convert_xml_import(cr, module_name, fp, idref, mode=mode)
  File "/odoo/src/odoo/tools/convert.py", line 869, in convert_xml_import
    obj.parse(doc.getroot(), mode=mode)
  File "/odoo/src/odoo/tools/convert.py", line 768, in parse
    exc_info[2]     
  File "/odoo/src/odoo/tools/pycompat.py", line 86, in reraise
    raise value.with_traceback(tb)
  File "/odoo/src/odoo/tools/convert.py", line 762, in parse
    self._tags[rec.tag](rec, de, mode=mode)
  File "/odoo/src/odoo/tools/convert.py", line 667, in _tag_record
    record = model.with_context(rec_context)._load_records([data], self.mode == 'update')
  File "/odoo/src/odoo/models.py", line 3810, in _load_records
    records = self.create([data['values'] for data in to_create])
  File "<decorator-gen-258>", line 2, in create
  File "/odoo/src/odoo/api.py", line 433, in _model_create_single
    return self.browse().concat(*(create(self, vals) for vals in arg))
  File "/odoo/src/odoo/api.py", line 433, in <genexpr>
    return self.browse().concat(*(create(self, vals) for vals in arg))
  File "/odoo/src/addons/project/models/project.py", line 693, in create
    task = super(Task, self.with_context(context)).create(vals)
  File "<decorator-gen-125>", line 2, in create
  File "/odoo/src/odoo/api.py", line 451, in _model_create_multi
    return create(self, [arg])
  File "/odoo/src/addons/mail/models/mail_thread.py", line 271, in create
    threads = super(MailThread, self).create(vals_list)
  File "/odoo/src/odoo/api.py", line 452, in _model_create_multi
    return create(self, arg)
  File "/odoo/src/odoo/models.py", line 3511, in create
  File "/odoo/src/odoo/models.py", line 3611, in _create
    cr.execute(query, params)
odoo.tools.convert.ParseError: "null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (1629, 2019-01-09 21:12:05.988338, 10, null, null, 1, null, null, null, 1, 1, 2019-01-09 21:12:17, null, 1, 0, null, null, 2019-01-09 21:12:17, null, normal, 2019-01-09 21:12:
05.988338, t, null, null, null, null, null, null, null, null, null, null, 2019-01-09 21:12:17, null, null, null, null, null, null, null, null, null, null, null, null).
" while parsing /odoo/src/addons/project/migrations/12.0.1.1/noupdate_changes.xml:3, near
<record id="project_task_data_14" model="project.task" forcecreate="False">
    <field name="user_id" ref="base.user_admin"/>
    <field name="displayed_image_id"/>
  </record>
```

I'm not sure if the project script is faulty or if there is something corrupted with my database. I can tell that I have no such ir_model_data record with name='project_task_data_14' and I'm not sure if it is right or not.

However I'm surprised you init only just a few of the fields of these data records here:
https://github.com/OCA/OpenUpgrade/pull/1674/files#diff-926f0f5eabc7fb5e956f3f92840a301cR3
while the v12 code sets up many more fields: https://github.com/odoo/odoo/blob/12.0/addons/project/data/project_data.xml#L271

Is there any explanation I'm missing? Also where are you taking this displayed_image_id field from as it is not set in https://github.com/odoo/odoo/blob/12.0/addons/project/data/project_data.xml#L271 ?

Sorry if I'm wrong about something.

MiquelRForgeFlow:
@rvalyi Thanks for your feedback, you are very helpful! :)

In the `noupdate_changes.xml` we only put the fields that have changed between versions. The `displayed_image_id` was from https://github.com/odoo/odoo/blob/11.0/addons/project/data/project_data.xml#L326.

MiquelRForgeFlow:
@rvalyi Can you test it from a fresh v11 to v12 without v10 and older versions?

MiquelRForgeFlow:
I just detected that the `project_task_data_14` in v8 was demo data but in v9 it was moved to real data. I don't know if this is the cause of the problem or not.

MiquelRForgeFlow:
> I just detected that the project_task_data_14 in v8 was demo data but in v9 it was moved to real data. I don't know if this is the cause of the problem or not.

Ah, that's the real cause. Yeah. @rvalyi In your v8 database you don't have the `project_task_data_14` record. Thus, you migrate to v9, to v10, to v11, without adding this record in the database. Then, when migrating to v12, the `noupdate_changes.xml` is trying to update this record. But as you don't have it, it breaks because the field "name" that is required is not defined.

MiquelRForgeFlow:
@pedrobaeza, what do you recommend to @rvalyi in this case? Maybe we should add this kind of records in v9 `noupdate_changes.xml` file to ensure they are added when migrating to v9.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
+---Fields in module 'project'---
+project      / account.analytic.account / company_uom_id (many2one)     : module is now 'hr_timesheet' ('project')
+project      / account.analytic.account / company_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
+project      / account.analytic.account / project_count (integer)       : module is now 'hr_timesheet' ('project')
+project      / account.analytic.account / project_ids (one2many)        : module is now 'hr_timesheet' ('project')
+# NOTHING TO DO
```
pedrobaeza: For this we might need to use `https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L939`, as I'm thinking that if not, we can have data loss. This is how I think Odoo will handle this:

1. `project` module is updated first, as `hr_timesheet` depends on it.
2. As the fields are removed from `project` module, they are going to be removed, as there are no more references to that fields.
3. Then, when updating `hr_timesheet`, the fields will be created again, but now with empty data.

Can you please confirm that this is what is happening? We are using any way a buggy base right now for this, as there was a patch for properly removing SQL columns when no references left: https://github.com/odoo/odoo/pull/30018, so we will need to update from upstream code and then try.

If you confirm my fears, then we need to update docstring of the method to reflect this new reality, and call the method for these fields.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+project      / digest.digest            / kpi_project_task_opened (boolean): NEW
+# NOTHING TO DO
+
+project      / project.project          / access_token (char)           : NEW
```
pedrobaeza: If you have researched for each field why it's `Nothing to do`, the perfect thing to do is to comment the results of your investigation here for not having to perform the same investigation when reviewing. For example, for this field I would put:
`Nothing to do: The token is empty by default and generated when requested sharing the link`

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.project          / analytic_account_id (many2one): module is now 'hr_timesheet' ('project')
```
pedrobaeza: The same about updating field module.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.project          / message_last_post (datetime)  : DEL 
+project      / project.project          / message_main_attachment_id (many2one): NEW relation: ir.attachment
+project      / project.project          / percentage_satisfaction_project (integer): previously in module rating_project
```
pedrobaeza: All these moved fields are indeed nothing to do, but an explanation about why is also good (the same as I have made previously in my comment on the contrary case).

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.project          / percentage_satisfaction_task (integer): previously in module rating_project
+project      / project.project          / portal_show_rating (boolean)  : previously in module website_rating_project
+project      / project.project          / rating_request_deadline (datetime): previously in module rating_project
+project      / project.project          / rating_status (selection)     : previously in module rating_project
+project      / project.project          / rating_status_period (selection): previously in module rating_project
+# NOTHING TO DO
+
+project      / project.task             / access_token (char)           : NEW
```
pedrobaeza, line 21: You can group both fields with the same name and put the same explanation about `Nothing to do`.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.task             / activity_date_deadline (date) : now a function
+project      / project.task             / message_last_post (datetime)  : DEL 
+project      / project.task             / message_main_attachment_id (many2one): NEW relation: ir.attachment
+# NOTHING TO DO
+
+project      / project.task             / rating_count (integer)        : previously in module rating_project
```
pedrobaeza, line 45: You can group this with the other fields from `project.project` and put the same explanation.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.task             / rating_ids (one2many)         : previously in module rating_project
+project      / project.task             / rating_last_feedback (text)   : previously in module rating_project
+project      / project.task             / rating_last_image (binary)    : previously in module rating_project
+project      / project.task             / rating_last_value (float)     : previously in module rating_project
+# NOTHING TO DO
+
+project      / project.task             / remaining_hours (float)       : module is now 'hr_timesheet' ('project')
```
pedrobaeza: The same about updating field module.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.task             / remaining_hours (float)       : now a function
+# NOTHING TO DO
+
+project      / project.task.type        / auto_validation_kanban_state (boolean): previously in module rating_project
```
pedrobaeza: Group this with the other similar fields

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+project      / project.task.type        / rating_template_id (many2one) : previously in module rating_project
+project      / res.company              / project_time_mode_id (many2one): module is now 'hr_timesheet' ('project')
```
pedrobaeza: The same about updating field module

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+---XML records in module 'project'---
+NEW digest.digest: digest.digest_digest_default (noupdate)
+NEW digest.tip: project.digest_tip_project_0
+NEW ir.actions.act_window: project.rating_rating_action_project_report
+NEW ir.actions.act_window: project.rating_rating_action_task
+NEW ir.actions.act_window: project.rating_rating_action_view_project_rating
+DEL ir.actions.act_window: project.action_project_task_user_tree_filtered
+DEL ir.actions.act_window: project.project_task_action_activity
+DEL ir.actions.act_window: project.project_task_actions_act_window_merge_tasks
+DEL ir.attachment: project.msg_task_data_14_attach (noupdate)
```
pedrobaeza, line 75: You have to try to remove these records (with the openupgradelib method for that)

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+NEW ir.ui.view: project.portal_project_rating_progressbar
+NEW ir.ui.view: project.quick_create_task_form
+NEW ir.ui.view: project.rating_index
+NEW ir.ui.view: project.rating_project_rating_page
+NEW ir.ui.view: project.rating_rating_view_search_project
+DEL ir.ui.view: project.analytic_account_inherited_form
+DEL ir.ui.view: project.mail_template_task_merge
+DEL ir.ui.view: project.project_planner
+DEL ir.ui.view: project.project_task_merge_wizard_form
+DEL mail.message: project.msg_task_data_14 (noupdate)
```
pedrobaeza, line 122: To remove

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+DEL mail.message: project.msg_task_data_8 (noupdate)
+NEW mail.message.subtype: project.mt_project_task_rating
+NEW mail.message.subtype: project.mt_task_rating
+NEW mail.template: project.mail_template_data_project_task (noupdate)
+NEW mail.template: project.rating_project_request_email_template (noupdate)
+DEL mail.template: project.mail_template_data_module_install_project (noupdate)
+DEL project.task: project.project_task_data_8 (noupdate)
```
pedrobaeza, line 138: To remove

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+DEL res.company: base.main_company (noupdate)
+NEW res.groups: project.group_project_rating
+DEL res.groups: project.group_time_work_estimation_tasks
```
pedrobaeza, line 143: To remove? This will probably be renamed to something else in `hr_timesheet` module.

pedrobaeza (changes requested):
Apart from the inline comment, it seems clear seeing analysis file that `rating_project` is now merged in `project`, so you can proceed with this merging in the same PR.

I think also I deserve co-authorship of this.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Can you do a last check of the state of this PR? I am sure I have all done except this check you want me to confirm.

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+DEL ir.actions.act_window: rating_project.rating_rating_action_view_project_rating_task
+# NOTHING TO DO
+
+DEL ir.attachment: project.msg_task_data_14_attach (noupdate)
+DEL ir.attachment: project.msg_task_data_8_attach (noupdate)
+# DONE: post-migration: Deleted by xml_id
+
+NEW ir.cron: project.ir_cron_rating_project (noupdate)
+DEL ir.cron: rating_project.ir_cron_rating_project
+# DONE: pre-migration: renamed xml ids
```
pedrobaeza: No need. It's already renamed on the merge modules operation. Maybe you have to switch noupdate status, as previously it seems that it was noupdate="0".

addons/project/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+    openupgrade.load_data(
+        env.cr, 'project', 'migrations/12.0.1.1/noupdate_changes.xml')
+    openupgrade.delete_records_safely_by_xml_id(
+        env, [
+            'project.msg_task_data_14_attach',
+            'project.msg_task_data_8_attach',
+            'project.msg_task_data_14',
+            'project.msg_task_data_8',
+            'project.project_task_data_8',
+            'project.mail_template_data_module_install_project',
```
pedrobaeza: This is one is a fake one that you can remove. I think there will be a warning if not removed.

addons/project/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+xmlid_renames = [
```
pedrobaeza: No need of these renamings

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+
+DEL project.task: project.project_task_data_8 (noupdate)
+# DONE: post-migration: Deleted by xml_id
+
+DEL res.company: base.main_company (noupdate)
+NEW res.groups: project.group_project_rating
+# NOTHING TO DO
+
+DEL res.groups: project.group_time_work_estimation_tasks
+# ?? maybe use update_module_names
```
pedrobaeza: Please check it here in this PR before being merged

addons/project/migrations/12.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="rating_project_request_email_template" model="mail.template">
```
pedrobaeza: No need of this. It was previously noupdate="0". What you have to do is to set this XML-ID as noupdate.

pedrobaeza: However, have you checked all the `rating_task_*` and the changes to tasks that are noupdate="1" in `rating_project`

addons/project/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+        ],
+    )
+    openupgrade.delete_record_translations(
```
pedrobaeza: No need of this

pedrobaeza (changes requested):
Several comments and the `hr_timesheet` thing

addons/project/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Where/how to switch?

pedrobaeza: `openupgrade.update_module_names` already performs this. Check the code.

pedrobaeza: What I'm missing is the merge operation in apriori.py...

MiquelRForgeFlow: https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/addons/openupgrade_records/lib/apriori.py#L17

pedrobaeza: OK, it hasn't been done in this PR then, but as said, you have to remove this as it's already done.

addons/project/migrations/12.0.1.1/noupdate_changes.xml
MiquelRForgeFlow: The `rating_task_*` are demo data in both modules.

pedrobaeza: OK, great, then less noise...

MiquelRForgeFlow: > No need of this. It was previously noupdate="0". What you have to do is to set this XML-ID as noupdate.

I put the this record in v11 project module (in noupdate data) and did a comparison with v12. The result is what I put in this noupdate_changes file.

pedrobaeza: But as this ir.model.data is already created with noupdate="0", the record is going to be update with all data. That's why I told you that you have to switch this flag. Take into account that the noupdate part is only evaluated when creating the ir.model.data. The rest of the time the value at ir.model.data is the one used.

MiquelRForgeFlow: I did the switch flag in the pre-migration. So then in post-migration needs the noupdate file change.

pedrobaeza: Do it the contrary (on post) and don't get complicated...

MiquelRForgeFlow: 😂

mreficent:
> hahaha, `fixup!`

Because if I had done a force push then you couldn't see the previous force push changes... 😅

pedrobaeza:
> Because if I had done a force push then you couldn't see the previous force push changes...

For the record, the problem happens when you perform several operations on local (like rebasing + amend commit), and then forced push, so the line of the diff that GitHub presents all the changes on the added commits of the rebase + the real changes. That's why I asked you to 1. rebase, 2. forced push, 3. amend commit, 4. forced push. This way, you have the diff properly on the second line that GitHub presents.

MiquelRForgeFlow:
But you are forgetting something: If I force push commit A, then amend commit and force push, then amend commit and force push, and again and again, then Github doesn't create force push lines for each of this force pushes. Github will merge all this force push lines into one or two.

pedrobaeza:
> But you are forgetting something: If I force push commit A, then amend commit and force push, then amend commit and force push, and again and again, Github doesn't create force push lines for each of this force pushes. Github will merge all this force push lines into one or two.

That's not true, GitHub keeps each push with its diff:

seleccion_011

Try it next time.

MiquelRForgeFlow:
It doesn't happen in the first two consecutive force pushes. I think it fuses the second and the third. And maybe there is involved also a time condition. I don't know the exact details, but I saw some force push lines merged in other PRs.

pedrobaeza:
OK, let's try next time to see.

MiquelRForgeFlow:
Ok, it's ready except checking the `hr_timesheet` part :)

pedrobaeza:
Please check it and then I can press the merge button if everything is OK.

pedrobaeza:
The rest of changes seems good for me, and the double forced push has worked the way it's intended :wink:

MiquelRForgeFlow:
@pedrobaeza I checked now a migration test with the updated v11 and v12. The hr_timesheet-from-project columns remained filled, they don't get empty! 😋 So, don't need to do something about and you can merge happily now.

By the way, I saw that `remaining_hours` of project.task in v11 has `digits=(16,2)` (in db is numeric). But in v12 don't have this digits, so in db is double precision without decimal places. We need to do something about this?

pedrobaeza (approved):

pedrobaeza:
> By the way, I saw that `remaining_hours` of project.task in v11 has `digits=(16,2)` (in db is numeric). But in v12 don't have this digits, so in db is double precision without decimal places. We need to do something about this?

Very good question... The fact is that indeed they are different data types (with their advantages and disadvantages that we are not going to debate here), but might affect to the results of the field. For example, this can be one of the cases: previously, there were only a length of 2 decimals (the second part of the digits), which in a hour field can lead to inaccuracies in the result if you round to that 2 decimals, so it will be a problem in some cases if you migrate without changing that field. And I think we should take care of all of these changes across all versions. One approach can be to include in core code the conversion when detected for not needing to include this in migration scripts. What do you think?

MiquelRForgeFlow:
> What do you think?

Ok for me, but I don't know how to do  the conversion :S

rvalyi (changes requested):
I found a problem: until v11 the project.project object has an ```_inherits = {'account.analytic.account': "analytic_account_id"} ``` see https://github.com/odoo/odoo/blob/11.0/addons/project/models/project.py#L50 and this is where project names are stored for instance. However, in v12, there is no more such _inherits and the name is a column on the project_project table itself https://github.com/odoo/odoo/blob/12.0/addons/project/models/project.py#L175.
However the current migration in this PR doesn't take care of propagating the name (and perhaps other fields?) to the project,project object. As a result: migrated projects have a NULL name!

MiquelRForgeFlow:
Thanks @rvalyi !! This change wasn't detected by the openupgrade analysis. Fixed.

@pedrobaeza @StefanRijnhart What do you think? Can openupgrade analysis method be fixed to detect this kind of change?

addons/project/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+column_copies = {
+    'project_project': [
+        ('analytic_account_id', None, None),
```
pedrobaeza, line 7: We shouldn't rename this column, as it requires afterwards to be renamed/copied again in `hr_timesheet`, which is useless, as the meaning of the field doesn't change, and can lead to some errors if other data is linked to this SQL column.

pedrobaeza (changes requested):
Don't rename the column and it's OK to go.

addons/project/migrations/12.0.1.1/pre-migration.py
MiquelRForgeFlow: it's a column_copies, not a rename

pedrobaeza: Yes, right. As talked by chat, let's keep this in case we need it.

pedrobaeza (approved):

pedrobaeza:
As @rvalyi's concern is solved, I merge.

PR #1675, opened 2019-01-08
[12.0][MIG] hr

MiquelRForgeFlow:
Migration scripts for `hr` module.

--


addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+# NOTHING TO DO
+
+hr           / hr.job                   / message_last_post (datetime)  : DEL
+hr           / hr.job                   / message_main_attachment_id (many2one): NEW relation: ir.attachment
+hr           / mail.channel             / subscription_department_ids (many2many): NEW relation: hr.department
+# NOTHING TO DO
+
+---XML records in module 'hr'---
+NEW hr.employee: hr.employee_admin (noupdate)
+DEL hr.employee: hr.employee_root (noupdate)
```
pedrobaeza, line 32: We should do something here:

1. We can remove the old `hr.employee_root`, as no longer exists, but probably there will be other records linked to this one.
2. We can rename XML-ID for the new `admin` employee, and update all the proper fields.

Option 2 seems more perfect for the purpose intention.

pedrobaeza (changes requested):
Also reset template translations. I think all these suggestions deserve my co-authorship on the scripts although I'm not directly typing them (but I can do if that's a problem).

addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
+---Fields in module 'hr'---
+hr           / hr.department            / message_last_post (datetime)  : DEL 
+hr           / hr.department            / message_main_attachment_id (many2one): NEW relation: ir.attachment
+# NOTHING TO DO
+
+hr           / hr.employee              / activity_ids (one2many)       : NEW relation: mail.activity
+hr           / hr.employee              / additional_note (text)        : NEW 
+hr           / hr.employee              / certificate (selection)       : NEW selection_keys: ['bachelor', 'master', 'other']
+hr           / hr.employee              / children (integer)            : previously in module hr_contract
```
sergio-teruel, line 9: @mreficent I think that you have move children and place_of_brith from hr_contract to hr.
https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1921

sergio-teruel (changes requested):

addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: @sergio-teruel This is not needed in this case, as `hr` is loaded before than `hr_contract`, so the XML-ID for `hr` will be created first, and thus no removal will be done.

addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Ok, applied the option 2

MiquelRForgeFlow:
Ready

addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: But you are not loading in `noupdate_records` the changed fields...

MiquelRForgeFlow: Yes, in the beginning of the file. Here https://github.com/OCA/OpenUpgrade/pull/1675/files#diff-e962c6851ef25abb2b16c9bb04bf8c9aR3.

pedrobaeza: OK, sorry. I didn't see it.

pedrobaeza (approved):

PR #1676, opened 2019-01-08
[12.0][MIG] maintenance

MiquelRForgeFlow:
Migration scripts for `maintenance` module.

--


addons/maintenance/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
+---Fields in module 'maintenance'---
+maintenance  / maintenance.equipment    / activity_date_deadline (date) : now a function
+maintenance  / maintenance.equipment    / company_id (many2one)         : NEW relation: res.company
```
pedrobaeza, line 2: You have to empty this field for avoiding losing access to the equipments as previously this field and the security rule didn't exist.

addons/maintenance/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
...
+maintenance  / maintenance.equipment    / effective_date (date)         : NEW required: required, req_default: function
```
pedrobaeza: Fill this with `date(create_date)` for having the same behavior than previously.

addons/maintenance/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
...
+maintenance  / maintenance.equipment    / message_last_post (datetime)  : DEL 
+maintenance  / maintenance.equipment    / message_main_attachment_id (many2one): NEW relation: ir.attachment
+# NOTHING TO DO
+
+maintenance  / maintenance.equipment    / warranty (date)               : was renamed to warranty_date [nothing to do]
+# DONE: pre-migration: renamed.
+
+maintenance  / maintenance.equipment.category / company_id (many2one)         : NEW relation: res.company
```
pedrobaeza: Empty the field

addons/maintenance/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
...
+maintenance  / maintenance.equipment.category / message_last_post (datetime)  : DEL 
+maintenance  / maintenance.equipment.category / message_main_attachment_id (many2one): NEW relation: ir.attachment
+# NOTHING TO DO
+
+maintenance  / maintenance.request      / activity_date_deadline (date) : now a function
+maintenance  / maintenance.request      / company_id (many2one)         : NEW relation: res.company
```
pedrobaeza: Empty the field

addons/maintenance/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,41 @@
...
+maintenance  / maintenance.request      / message_last_post (datetime)  : DEL 
+maintenance  / maintenance.request      / message_main_attachment_id (many2one): NEW relation: ir.attachment
+# NOTHING TO DO
+
+maintenance  / maintenance.request      / technician_user_id (many2one) : was renamed to user_id [nothing to do]
+# DONE: pre-migration: renamed.
+
+maintenance  / maintenance.team         / active (boolean)              : NEW
+maintenance  / maintenance.team         / company_id (many2one)         : NEW relation: res.company
```
pedrobaeza: Empty the field

pedrobaeza (changes requested):

MiquelRForgeFlow:
Ready, all changes done.

pedrobaeza:
Next time you make a rebase, please force pushed right after the rebase and then force pushed again with the changes, so that the GitHub diff viewer can take only the important changes

pedrobaeza (approved):
And please avoid to use `use_env=False`. Why do you hate so much the default one?

MiquelRForgeFlow:
> please force pushed right after the rebase and then force pushed again with the changes

oh, right. And what did you when this option to see the force pushed changes not yet existed? 😁

> And please avoid to use use_env=False. Why do you hate so much the default one?

You are right. I hate so much the default one, but I don't know why haha. Is there any difference in efficiency, performance or something?

pedrobaeza:
> oh, right. And what did you when this option to see the force pushed changes not yet existed? :grin:

Ask the contributor to commit a fixup instead :wink: Don't you remember these old times with `fixup! ...` commits?

> You are right. I hate so much the default one, but I don't know why haha. Is there any difference in efficiency, performance or something?

It's because if tomorrow we need to expand the scripts and need full environment, then we have to dirty the diff changing that and all the lines involved. It's not critical, but it's not clean. The same applies in one specific method. That's why I always pass `env`, and use `env.cr` when needed (it's very short, so no special problem with PEP8 and so on).

PR #1678, opened 2019-01-08
[12.0][FIX] base, mail: minor fixes

MiquelRForgeFlow:
This PR fixes:
- base: `street` is a computed field in `res.company`, so to search for this field we need to compute it.
- mail: a list cannot be filtered.

--

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/post-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/post-migration.py
@@ -25,7 +25,9 @@ def generate_thumbnails(env):
 def update_res_company_onboarding_company_state(env):
     # based on old base_onboarding_company_done
     good_companies = env["res.company"].with_context(
-        active_test=False).search([('street', '!=', False)])
+        active_test=False).search([]).filtered(
+        lambda c: c.partner_id.browse(c.partner_id.sudo().address_get(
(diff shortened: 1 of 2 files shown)


odoo/addons/base/migrations/12.0.1.3/post-migration.py
```diff
@@ -25,7 +25,9 @@ def generate_thumbnails(env):
 def update_res_company_onboarding_company_state(env):
     # based on old base_onboarding_company_done
     good_companies = env["res.company"].with_context(
-        active_test=False).search([('street', '!=', False)])
+        active_test=False).search([]).filtered(
```
pedrobaeza: For making it more readable and as there's no `active` field in `res.company`:
```suggestion
    good_companies = env["res.company"].search([]).filtered(lambda c: (
        c.partner_id.browse(
            c.partner_id.sudo().address_get(adr_pref=['contact'])['contact']
        ).sudo().street
    ))
```

pedrobaeza (changes requested):
On v11, `ir.attachment` didn't have `active` field, so putting `active_test=False` is a nonsense, as all attachments are going to be active.

MiquelRForgeFlow:
Done

PR #1679, opened 2019-01-08
[12.0][MIG] payment

MiquelRForgeFlow:
Migration scripts for `payment` module.

--


addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
+---Fields in module 'payment'---
+payment      / account.invoice          / transaction_ids (many2many)   : NEW relation: payment.transaction
```
pedrobaeza, line 2: I'm not able to find where is this field populated. Do you? If there's a flow that feeds it, then we should try to feed past invoices/transactions the same way.

addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
...
+payment      / payment.acquirer         / qr_code (boolean)             : NEW 
+# NOTHING TO DO
+
+payment      / payment.transaction      / date (datetime)               : NEW
+payment      / payment.transaction      / date_validate (datetime)      : DEL
+# DONE: pre-migration: renamed field
+
+payment      / payment.transaction      / invoice_ids (many2many)       : NEW relation: account.invoice
```
pedrobaeza: This is the twin field of `transaction_ids` in `account.invoice`, so you should group both together.

addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
...
+payment      / payment.transaction      / is_processed (boolean)        : NEW
```
pedrobaeza: We should assume all the current transactions are processed, so please set this value by SQL

addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
...
+payment      / payment.acquirer         / qr_code (boolean)             : NEW 
+# NOTHING TO DO
+
+payment      / payment.transaction      / date (datetime)               : NEW
+payment      / payment.transaction      / date_validate (datetime)      : DEL
+# DONE: pre-migration: renamed field
+
+payment      / payment.transaction      / invoice_ids (many2many)       : NEW relation: account.invoice
+payment      / payment.transaction      / is_processed (boolean)        : NEW 
+payment      / payment.transaction      / payment_id (many2one)         : NEW relation: account.payment
```
pedrobaeza, line 17: You have to set this field through the field `payment_transaction_id`, which is the inverse of it (as `one2one` doesn't exist in the framework).

addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
...
+payment      / payment.transaction      / return_url (char)             : NEW
```
pedrobaeza: This can be nothing to do as this is used as a kind of transient data while the transaction is in process.

addons/payment/migrations/12.0.1.0/post- 😎.py
```diff
@@ -0,0 +1,21 @@
...
+from openupgradelib import openupgrade
+
+
+def map_payment_transaction_state(cr):
+    openupgrade.map_values(
+        cr,
+        openupgrade.get_legacy_name('state'),
+        'state',
+        [('refunding', 'pending'),
+         ('refunded', 'done'),
```
pedrobaeza: It should be `'cancel'`, not `'done'`:

v11:
https://github.com/OCA/OpenUpgrade/blob/65767162fcd60037a79532b31df0c5f253b6f74d/addons/payment_authorize/models/payment.py#L336
v12:
https://github.com/OCA/OpenUpgrade/blob/e2fdf7e64894efc786e4b56df2899087a0925398/addons/payment_authorize/models/payment.py#L320

https://github.com/OCA/OpenUpgrade/blob/e2fdf7e64894efc786e4b56df2899087a0925398/addons/payment/models/payment_acquirer.py#L762

The same may apply to `refunding`. At the end, it means cancel.

pedrobaeza (changes requested):
Please change the names of the files to `pre-migration.py` and `post-migration.py`.

Co-authorship requested!

addons/payment/migrations/12.0.1.0/post- 😎.py
MiquelRForgeFlow: I am not sure, but if you say so...

pedrobaeza: Yes, I have spent 10 minutes following the trace of this.

addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: I am not sure, but if you says so...

addons/payment/migrations/12.0.1.0/post- 😎.py
MiquelRForgeFlow: Good 👍😂

addons/payment/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: The `many2many` is filled in https://github.com/OCA/OpenUpgrade/blob/12.0/addons/payment/models/account_invoice.py#L81 and in https://github.com/OCA/OpenUpgrade/blob/12.0/addons/payment/models/payment_acquirer.py#L633. I don't know if I should do something about. Could you check?

pedrobaeza: I have finally filled this from linked account.payment

pedrobaeza (approved):

PR #1681, opened 2019-01-08
[12.0][MIG] stock & stock_account

MiquelRForgeFlow:
Migration scripts for `stock` & `stock_account` modules.

--


odoo/openupgrade/doc/source/modules110-120.rst
```diff
@@ -629,3 +629,14 @@ missing in the new release are marked with |del|.
...
 |website_twitter                             |                                                 |
 +--------------------------------------------+-------------------------------------------------+
+
+OCA modules
+===========
+
+Here you will find the coverage of OpenUpgrade for other OCA modules that has
+suffered any kind of transformation and it has been taken into account here:
+
+OCA/account-invoicing
```
pedrobaeza: OCA/stock-logistics-warehouse

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+            env, 'stock.rule',
+            [row[1]],
+            row[0],
+        )
+
+
+def merge_stock_putaway_strategy_product(env):
+    openupgrade.logged_query(
+        env.cr, """
+        INSERT INTO stock_fixed_putaway_strat (product_id, putaway_id,
```
pedrobaeza: In this case, as they are very similar, and there are no extra records already, go for a renaming of the table and the column in pre instead. Or at least copy the exact same `ID`. No need to have a "mapping" over other column. That's performed when the destination table can already contain other records, and thus, `id` column can be assured to be the same.

addons/stock/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,42 @@
...
+    openupgrade.rename_models(cr, _model_renames)
+    openupgrade.rename_tables(cr, _table_renames)
+    cr.execute(
+        """
+        ALTER TABLE stock_rule
+        ADD COLUMN {} integer;
+        """.format(openupgrade.get_legacy_name('loc_path_id'))
+    )
+    if openupgrade.table_exists(cr, 'stock_product_putaway_strategy'):
+        cr.execute(
```
pedrobaeza: No need of adding this column according my other comment in post.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+stock        / stock.location           / parent_path (char)            : NEW 
+stock        / stock.location           / parent_right (integer)        : DEL 
+# DONE: post-migration: called _parent_store_compute
+
+stock        / stock.location.route     / pull_ids (one2many)           : DEL relation: procurement.rule
+stock        / stock.location.route     / push_ids (one2many)           : DEL relation: stock.location.path
+stock        / stock.location.route     / rule_ids (one2many)           : NEW relation: stock.rule
+# NOTHING TO DO
+
+stock        / stock.move               / ordered_qty (float)           : DEL
```
pedrobaeza, line 33: It's worth to say that this field is a copy of the initial value of `product_uom_qty`, but not used more anywhere (only as reference), so removing it it's not a problem.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+stock        / stock.move               / package_level_id (many2one)   : NEW relation: stock.package_level
+stock        / stock.move               / product_uom (many2one)        : relation is now 'uom.uom' ('product.uom')
+# NOTHING TO DO
+
+stock        / stock.move               / push_rule_id (many2one)       : DEL relation: stock.location.path
+stock        / stock.move               / rule_id (many2one)            : relation is now 'stock.rule' ('procurement.rule')
+# DONE: post-migration: merged push_rule_id into rule_id
+
+stock        / stock.move.line          / ordered_qty (float)           : DEL
```
pedrobaeza: The same comment as the one in `stock.move`, and you can even group both together.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+stock        / stock.location           / parent_right (integer)        : DEL 
+# DONE: post-migration: called _parent_store_compute
+
+stock        / stock.location.route     / pull_ids (one2many)           : DEL relation: procurement.rule
+stock        / stock.location.route     / push_ids (one2many)           : DEL relation: stock.location.path
+stock        / stock.location.route     / rule_ids (one2many)           : NEW relation: stock.rule
+# NOTHING TO DO
+
+stock        / stock.move               / ordered_qty (float)           : DEL
+stock        / stock.move               / package_level_id (many2one)   : NEW relation: stock.package_level
```
pedrobaeza: This should go grouped with the new fields of `stock.package_level` model.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+stock        / stock.move.line          / package_level_id (many2one)   : NEW relation: stock.package_level
+stock        / stock.move.line          / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
+# NOTHING TO DO
+
+stock        / stock.package_level      / location_dest_id (many2one)   : NEW relation: stock.location
+stock        / stock.package_level      / move_ids (one2many)           : NEW relation: stock.move
+stock        / stock.package_level      / move_line_ids (one2many)      : NEW relation: stock.move.line
+stock        / stock.package_level      / package_id (many2one)         : NEW relation: stock.quant.package, required: required
+stock        / stock.package_level      / picking_id (many2one)         : NEW relation: stock.picking
+DONE: post-migration: filled model (tentative).
```
pedrobaeza: Have you found the utility of this model? I'm not able to see it in even in interface... If not useful, I would say to not fill it, as it's optional. Any way, I'm putting some comments on your code.

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def fill_stock_package_level(env):
+    all_moves = env['stock.move'].search([])
+    all_pickings = env['stock.picking'].search([])
+
+    # case 1:
+    moves = all_moves.filtered(lambda m: m.state == 'assigned')
+    for move in moves:
+        move.mapped('picking_id')._check_entire_pack()
```
pedrobaeza: This is the same as doing only one line `moves.mapped('picking_id')._check_entire_pack()`, and why not using directly `all_pickings.filtered(lambda p: p.state == 'assigned')_check_entire_pack()`?

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+    # case 2:
+    pickings = all_pickings.filtered(
+        lambda p: p.state not in ('done', 'cancel'))
+    for picking in pickings:
+        move_line_ids = picking.move_line_ids.filtered(
+            lambda ml: ml.qty_done > 0 and not ml.result_package_id)
+        if move_line_ids and len(
+                move_line_ids.mapped('location_dest_id')) == 1:
+            move_lines_packed = move_line_ids.filtered(
+                lambda ml: ml.result_package_id)
```
pedrobaeza: You are filtering initially by the condition `not ml.result_package_id`, but now you try to do it the contrary, resulting in an empty recordset.

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+    all_pickings = env['stock.picking'].search([])
+
+    # case 1:
+    moves = all_moves.filtered(lambda m: m.state == 'assigned')
+    for move in moves:
+        move.mapped('picking_id')._check_entire_pack()
+
+    # case 2:
+    pickings = all_pickings.filtered(
+        lambda p: p.state not in ('done', 'cancel'))
```
pedrobaeza: I think this should be done over the rest of the pickings except `draft` and `assigned` ones. And even more, I would say, that you can include here the `assigned` as well, and forget the step 1 that can hurt if you have specific code in `_check_entire_pack` or unneeded steps.

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+            move_lines_packed = move_line_ids.filtered(
+                lambda ml: ml.result_package_id)
+            package = move_lines_packed.result_package_id
+            env['stock.package_level'].create({
+                'package_id': package.id,
+                'picking_id': picking.id,
+                'location_id': False,
+                'location_dest_id': move_line_ids.mapped(
+                    'location_dest_id').id,
+                'move_line_ids': [(6, 0, move_lines_packed.ids)]
```
pedrobaeza: And what about stock.move?

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+        lambda p: p.state not in ('done', 'cancel'))
+    for picking in pickings:
+        move_line_ids = picking.move_line_ids.filtered(
+            lambda ml: ml.qty_done > 0 and not ml.result_package_id)
+        if move_line_ids and len(
+                move_line_ids.mapped('location_dest_id')) == 1:
+            move_lines_packed = move_line_ids.filtered(
+                lambda ml: ml.result_package_id)
+            package = move_lines_packed.result_package_id
+            env['stock.package_level'].create({
```
pedrobaeza: You should group by move.

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+                'package_id': package.id,
+                'picking_id': picking.id,
+                'location_id': False,
+                'location_dest_id': move_line_ids.mapped(
+                    'location_dest_id').id,
+                'move_line_ids': [(6, 0, move_lines_packed.ids)]
+            })
+
+    # case 3:
+    moves = all_moves.filtered(
```
pedrobaeza: This shouldn't be needed with the other changes

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+        lambda m: m.picking_id and m.state not in ('done', 'cancel'))
+    for move in moves:
+        move.mapped('package_level_id').write({
+            'picking_id': move.picking_id.id,
+        })
+
+    # case 4:
+    moves = all_moves.filtered(lambda m: m.state == 'assigned')
+    moves.mapped('package_level_id').write({'is_done': True})
```
pedrobaeza: Is this condition correct?

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+
+stock        / stock.picking            / activity_date_deadline (date) : now a function
+stock        / stock.picking            / backorder_ids (one2many)      : NEW relation: stock.picking
+stock        / stock.picking            / immediate_transfer (boolean)  : NEW 
+stock        / stock.picking            / message_last_post (datetime)  : DEL 
+stock        / stock.picking            / message_main_attachment_id (many2one): NEW relation: ir.attachment
+stock        / stock.picking            / move_ids_without_package (one2many): NEW relation: stock.move
+stock        / stock.picking            / move_line_ids_without_package (one2many): NEW relation: stock.move.line
+stock        / stock.picking            / package_level_ids (one2many)  : NEW relation: stock.package_level
```
pedrobaeza: Group these fields with the package level part

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+from openupgradelib import openupgrade, openupgrade_merge_records
+from psycopg2.extensions import AsIs
+
+
+def fill_stock_picking_type_barcode(env):
+    picking_types = env['stock.picking.type'].with_context(
+        active_test=False).search([('code', '!=', False),
+                                   (['warehouse_id', '!=', False])])
+    default_names_dict = {
+        _('Receipts'): '-RECEIPTS',
```
pedrobaeza: Translations are not working here, so make sure when you query later this that the name is got in English.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+stock        / stock.picking.type       / barcode (char)                : NEW
+# DONE: post-migration: Filled the same way Odoo does
+
+stock        / stock.picking.type       / barcode_nomenclature_id (many2one): DEL relation: barcode.nomenclature
+# NOTHING TO DO: handled in barcodes module
+
+stock        / stock.production.lot     / activity_ids (one2many)       : NEW relation: mail.activity
+stock        / stock.production.lot     / message_last_post (datetime)  : DEL 
+stock        / stock.production.lot     / message_main_attachment_id (many2one): NEW relation: ir.attachment
+stock        / stock.production.lot     / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
```
pedrobaeza: You can group all the uom related changes and say `Nothing to do: model renamed in base`

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+# DONE: post-migration (merge stock.location.path into stock.rule)
+
+stock        / stock.scrap              / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
+# NOTHING TO DO
+
+stock        / stock.warehouse          / default_resupply_wh_id (many2one): DEL relation: stock.warehouse
+# NOTHING TO DO
+
+stock        / stock.warehouse          / mto_pull_id (many2one)        : relation is now 'stock.rule' ('procurement.rule')
+# DONE: pre-migration (rename of procurement.rule into stock.rule)
```
pedrobaeza, line 89: Group with the rest of the fields that involves this. I'm seeing that you don't group and keep this "random" order, which is less readable, and the idea of the work file is that you reorder at your convenience, grouping and commenting by logical sets.

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_model_data imd
+        SET model = 'stock.rule', res_id = sr.id
+        FROM stock_rule sr
+        WHERE imd.res_id = sr.%s AND model = 'stock.location.path'
+        """, (AsIs(openupgrade.get_legacy_name('loc_path_id')), ),
+    )
+    env.cr.execute(
```
pedrobaeza, line 72: Why are you performing this until the end of the method? I don't think it's needed.

addons/stock/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2018 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_column_renames = {
+    'stock_location_path': [
```
pedrobaeza: Why renaming this if at the end, you are copying this table?

addons/stock/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,42 @@
...
+@openupgrade.migrate(use_env=False)
+def migrate(cr, version):
+    openupgrade.rename_columns(cr, _column_renames)
+    openupgrade.rename_models(cr, _model_renames)
+    openupgrade.rename_tables(cr, _table_renames)
+    cr.execute(
+        """
+        ALTER TABLE stock_rule
+        ADD COLUMN {} integer;
+        """.format(openupgrade.get_legacy_name('loc_path_id'))
```
pedrobaeza, line 37: I usually used `orig_` prefix for indicating that this comes from an insert operation and this is the reference of the original record

pedrobaeza (commented):
Wow, it has been huge to study all the module and review the migration scripts...

Please give me the credits :wink:

pedrobaeza (changes requested):
Comments above

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+DEL ir.model.access: stock.access_product_attribute_line_manager
+DEL ir.model.access: stock.access_product_attribute_price_manager
+DEL ir.model.access: stock.access_product_uom_categ_stock_manager
+DEL ir.model.access: stock.access_product_uom_stock_manager
+DEL ir.model.access: stock.access_stock_incoterms_all
+DEL ir.model.access: stock.access_stock_incoterms_manager
+DEL ir.model.access: stock.access_stock_location_path_internal_user
+DEL ir.model.access: stock.access_stock_location_path_stock_user
+DEL ir.model.access: stock.access_stock_location_path_user
+DEL ir.rule: stock.stock_location_path_comp_rule (noupdate)
```
pedrobaeza, line 136: Delete this rule

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,196 @@
...
+DEL ir.ui.view: stock.stock_location_path_form
+DEL ir.ui.view: stock.stock_location_path_tree
+DEL ir.ui.view: stock.view_incoterms_tree
+DEL ir.ui.view: stock.view_procurement_rule_filter
+DEL ir.ui.view: stock.view_procurement_rule_form
+DEL ir.ui.view: stock.view_procurement_rule_tree
+DEL ir.ui.view: stock.view_quant_package_picking_tree
+NEW res.groups: stock.group_lot_on_delivery_slip
+DEL res.request.link: stock.req_link_tracking (noupdate)
+DEL stock.incoterms: stock.incoterm_CFR
```
pedrobaeza, line 179: Is this renamed somewhere?

pedrobaeza (commented):
A couple of comments more

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: But how can you rename A to B when you have A and B at the same time?

pedrobaeza: Not on pre-migration

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: This model is like stock quant, only technical without interface. I think they created it as a result of making compatible the stock move lines with the stock packages.

pedrobaeza: stock.quant are shown on stock on hand smart-button, reports, etc, but not this one. That's why I say of not populating it at all. If one day this becomes visible, then we can do it. But I let it your choice. Check my comments if you go ahead.

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: What do you mean later? I need this translations here because `_get_picking_type_create_values` method used translated names.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Yeah, but lazy to group them. But ok, I will do.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: My "random" order is not random. I tried to maintain the default order, that is "model name" order, to do the minimum changes possible.

addons/stock/migrations/12.0.1.1/pre-migration.py
MiquelRForgeFlow: Yeah, a bit stupid. It's the habit. Are you just asking me to point the silliness or requiring me to revert this?

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Don't know. I just simply did "stock.incoterms -> account.incoterms" rename in `account` migration. I didn't check all records of stock.incoterms. Maybe some of them went to localization modules.

addons/stock/migrations/12.0.1.1/post-migration.py
pedrobaeza: Translations are not going to work here, as you don't have a translation pointing to this file. I mean by "later" a couple of lines below.

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: Then remove the `_work` part in the file name :stuck_out_tongue_winking_eye:

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: That's why I put "random" between quotes, because it's the order provided by `openupgrade_records`, which is not meaningful.

addons/stock/migrations/12.0.1.1/pre-migration.py
pedrobaeza: Better to revert it :wink:

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
pedrobaeza: Please check. They are standard, so they should be somewhere else apart from localization modules. The case is that I remember some `ADD` of these records in one review, but I didn't take that into account...

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: What do you mean? If I am migrating from v11, and in v11 database I have both tables filled, then what do you expect to do in this migration? With two tables filled I cannot rename one as the other.

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: I just used the same as done here:

https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_move.py#L1011-L1012

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: I just used what is done here: https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_picking.py#L1024

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: I just copy pasted this:

https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_picking.py#L1045-L1050

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: same as above

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: check here:

https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_picking.py#L572-L574

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: This is the back order case:

https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_picking.py#L820-L837

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: Are the translations working in end-migration? if yes, maybe I could do that in end-migration. The problem here is that the names of the picking types are translated.

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: I think it's needed because the stock_move table has rule_id and push_rule_id columns. So as I add the push rules with a different id in the stock rule table, then I need to make sure that the rule_id column of stock_move is properly filled in two cases: (1) if rule_id and push_rule_id both were filled, and (2) rule_id not filled but push_rule_id filled.

addons/stock/migrations/12.0.1.1/pre-migration.py
MiquelRForgeFlow: well, but if we are using openupgrade to perform this changes then I prefer to use the get_legacy_name here

addons/stock/migrations/12.0.1.1/openupgrade_analysis_work.txt
MiquelRForgeFlow: Checked, all of them are in `account`. Happens that in the last openupgrade analysis update the rename of this model was not yet in apriori.

rvalyi:
Hello,
I should say that after migrating a production database from v8 to v11 without any problem, when trying to migrate to v12 with this script, at the end of the migration, Odoo curiously tries to delete records from stock_picking_type for which there is an ir_model_data record with noupdate=False. From what I understand such records where what was used previously and should somehow be re-mapped to the new default records may be? Cause of course the brutal deletion here violates not-null constraint on many foreign keys then...

```
2019-02-12 05:14:48,026 807 INFO db odoo.addons.base.models.ir_model: Deleting 1@stock.picking.type (stock.picking_type_in) 
2019-02-12 05:14:48,029 807 ERROR db odoo.sql_db: bad query: b'DELETE FROM stock_picking_type WHERE id IN (1)'
ERROR: null value in column "picking_type_id" violates not-null constraint
DETAIL:  Failing row contains (760, null, 2014-12-17 13:27:18.088813, 1, 1, 2008-11-18 00:00:00, 18, null, 1, 4, 1455.0, null, 16, null, 2, -291.0, done, null, 4, null, 1, null, null, 2019-02-09 05:30:22.465769, , null, 1164.0, CF00012, null, picking, f, null, 2008-12-03, 12, null, done, null, no, 2008-12-03 00:00:00, null, 0, 0, null, null, 60).
CONTEXT:  SQL statement "UPDATE ONLY "public"."purchase_order" SET "picking_type_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "picking_type_id""
  File "/odoo/src/odoo/addons/base/models/ir_model.py", line 1646, in _process_end
    record.unlink()
  File "/odoo/external-src/connector/component_event/models/base.py", line 112, in unlink
    result = super(Base, self).unlink()
  File "/odoo/src/odoo/models.py", line 3108, in unlink
    cr.execute(query, (sub_ids,))
psycopg2.IntegrityError: null value in column "picking_type_id" violates not-null constraint
DETAIL:  Failing row contains (760, null, 2014-12-17 13:27:18.088813, 1, 1, 2008-11-18 00:00:00, 18, null, 1, 4, 1455.0, null, 16, null, 2, -291.0, done, null, 4, null, 1, null, null, 2019-02-09 05:30:22.465769, , null, 1164.0, CF00012, null, picking, f, null, 2008-12-03, 12, null, done, null, no, 2008-12-03 00:00:00, null, 0, 0, null, null, 60).
CONTEXT:  SQL statement "UPDATE ONLY "public"."purchase_order" SET "picking_type_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "picking_type_id""
```

and the same problem happens with stock.location with errors such as:
```
2019-02-12 05:34:37,919 850 INFO db odoo.models.unlink: User #1 deleted ir.ui.view records with IDs: [1901] 
2019-02-12 05:34:37,921 850 INFO db odoo.addons.base.models.ir_model: Deleting 15@stock.location (stock.stock_location_output) 
2019-02-12 05:34:38,054 850 ERROR db odoo.sql_db: bad query: b'DELETE FROM stock_location WHERE id IN (15)'
ERROR: null value in column "location_id" violates not-null constraint
DETAIL:  Failing row contains (3, 2014-11-29 09:38:26.603823, 20, 6, 2015-12-27 17:51:54.169874, f, 1, WH: Input -> Output, null, move, null, propagate, null, 1, t, make_to_order, null, null, 20, 3, 0, 4, null, , 20.00, null, manual).
CONTEXT:  SQL statement "UPDATE ONLY "public"."stock_rule" SET "location_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "location_id""
    record.unlink()
  File "/odoo/external-src/connector/component_event/models/base.py", line 112, in unlink
    result = super(Base, self).unlink()
  File "/odoo/src/odoo/models.py", line 3108, in unlink
    cr.execute(query, (sub_ids,))
psycopg2.IntegrityError: null value in column "location_id" violates not-null constraint
DETAIL:  Failing row contains (3, 2014-11-29 09:38:26.603823, 20, 6, 2015-12-27 17:51:54.169874, f, 1, WH: Input -> Output, null, move, null, propagate, null, 1, t, make_to_order, null, null, 20, 3, 0, 4, null, , 20.00, null, manual).
CONTEXT:  SQL statement "UPDATE ONLY "public"."stock_rule" SET "location_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "location_id""

2019-02-12 05:34:38,064 850 CRITICAL db odoo.service.server: Failed to initialize database `db`. 
  File "/odoo/src/odoo/service/server.py", line 1055, in preload_registries
  File "/odoo/src/odoo/modules/registry.py", line 86, in new
  File "/odoo/src/odoo/modules/loading.py", line 455, in load_modules
    env['ir.model.data']._process_end(processed_modules)
  File "/odoo/src/odoo/addons/base/models/ir_model.py", line 1646, in _process_end
    record.unlink()
  File "/odoo/external-src/connector/component_event/models/base.py", line 112, in unlink
    result = super(Base, self).unlink()
  File "/odoo/src/odoo/models.py", line 3108, in unlink
    cr.execute(query, (sub_ids,))
psycopg2.IntegrityError: null value in column "location_id" violates not-null constraint
DETAIL:  Failing row contains (3, 2014-11-29 09:38:26.603823, 20, 6, 2015-12-27 17:51:54.169874, f, 1, WH: Input -> Output, null, move, null, propagate, null, 1, t, make_to_order, null, null, 20, 3, 0, 4, null, , 20.00, null, manual).
```

As a work around, I set noupdate=True for the stock.picking.type and stock.location records involved. I'm pretty sure this is not the correct thing, but at least I could keep testing...

cuongnmtm:
Hello guys, i have an issue when doing the migration from 11.0 to 12.0: The options of field selection are changed here (https://github.com/odoo/odoo/commit/a35df8d37170790257876a54aa54f2c3ab596c44#diff-23b5396fd047e334041d578a52ace527L38).
So that when migrated to 12.0, there are still records that have `move` option, can you help me how should we deal with that?

Thanks

MiquelRForgeFlow:
@cuongnmtm thanks, I didn't see this before. Solved.

MiquelRForgeFlow:
> As a work around, I set noupdate=True for the stock.picking.type and stock.location records involved. I'm pretty sure this is not the correct thing, but at least I could keep testing...

@rvalyi what do you mean by you set noupdate=True? If I am not seeing wrong, they (`stock.stock_location_output` and `stock.picking_type_in`) are already as noupdate=True because https://github.com/OCA/OpenUpgrade/blob/12.0/addons/stock/data/stock_data.xml#L81 🤔

nkhuyit:
I got an error while I was trying to migrate model stock. Can anyone help me to handle it, please?

  File "/opt/odoo/code/projects/OpenUpgrade/odoo/service/server.py", line 1055, in preload_registries
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/modules/registry.py", line 86, in new
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/modules/loading.py", line 476, in load_modules
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/modules/loading.py", line 360, in load_marked_modules
    upg_registry=upg_registry,
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/modules/loading.py", line 257, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/modules/migration.py", line 187, in migrate_module
  File "/opt/odoo/code/projects/OpenUpgrade/openupgradelib/openupgradelib/openupgrade.py", line 1489, in wrapped_function
    if use_env2 else cr, version)
  File "/opt/odoo/code/projects/OpenUpgrade/addons/stock/migrations/12.0.1.1/post-migration.py", line 170, in migrate
    fill_stock_package_level(env)
  File "/opt/odoo/code/projects/OpenUpgrade/addons/stock/migrations/12.0.1.1/post-migration.py", line 145, in fill_stock_package_level
    'move_line_ids': [(6, 0, move_lines_packed.ids)]
  File "<decorator-gen-178>", line 2, in create
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/api.py", line 430, in _model_create_single
    return create(self, arg)
  File "/opt/odoo/code/projects/OpenUpgrade/addons/stock/models/stock_package_level.py", line 132, in create
    result = super(StockPackageLevel, self).create(vals)
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/api.py", line 451, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/models.py", line 3540, in create
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/models.py", line 3640, in _create
    cr.execute(query, params)
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/sql_db.py", line 148, in wrapper
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/sql_db.py", line 225, in execute
psycopg2.IntegrityError: null value in column "package_id" violates not-null constraint
DETAIL:  Failing row contains (1, null, 11680, 10, 1, 2019-05-10 10:16:40.525537, 1, 2019-05-10 10:16:40.525537).

addons/stock/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: solved

MiquelRForgeFlow:
@nkhuyit  I think I solved your error 😉

nkhuyit:
I'm facing with problem about stock module again .Please anyone help me fix it : 
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/tools/convert.py", line 762, in parse
    self._tags[rec.tag](rec, de, mode=mode)
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/tools/convert.py", line 667, in _tag_record
    record = model.with_context(rec_context)._load_records([data], self.mode == 'update')
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/models.py", line 3818, in _load_records
    data['record']._load_records_write(data['values'])
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/models.py", line 3756, in _load_records_write
    self.write(values)
  File "/opt/odoo/code/projects/OpenUpgrade/addons/mrp/models/stock_warehouse.py", line 237, in write
    return super(StockWarehouse, self).write(vals)
  File "/opt/odoo/code/projects/OpenUpgrade/addons/stock/models/stock_warehouse.py", line 167, in write
    warehouse._create_or_update_global_routes_rules()
  File "/opt/odoo/code/projects/OpenUpgrade/addons/stock/models/stock_warehouse.py", line 247, in _create_or_update_global_routes_rules
    self[rule_field] = self.env['stock.rule'].create(values)
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/api.py", line 451, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/models.py", line 3540, in create
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/models.py", line 3640, in _create
    cr.execute(query, params)
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/sql_db.py", line 148, in wrapper
  File "/opt/odoo/code/projects/OpenUpgrade/odoo/sql_db.py", line 225, in execute
psycopg2.IntegrityError: null value in column "picking_type_id" violates not-null constraint
DETAIL:  Failing row contains (301, WKH: Stock → Pre-Production (MTO), f, propagate, null, pull, 20, 1, 365, 16, 1, make_to_order, null, null, 0, null, t, 1, null, 1, 2019-06-12 07:35:30.818248, 1, 2019-06-12 07:35:30.818248, null, null, manual).

mreficent:
@nkhuyit this sounds to me it's a mrp problem, not stock one.

MiquelRForgeFlow:
@nkhuyit could you says us when this error is produced? which module is being loaded when this error happens?

nkhuyit:
@mreficent When i migrate purchase_stock module and got a error like that.

MiquelRForgeFlow:
@nkhuyit I solved the issue in the migration of mrp. Updated that PR. 😉

pedrobaeza:
I have reworked 2 things:

* Rename table and columns for putaway product instead of inserting
* Simplify stock.quant_level creation based on previous version output

Please check.

I still need to check stock.rule merging.

addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -103,58 +103,35 @@ def merge_stock_location_path_stock_rule(env):
...
+    for group in groups:
+        if group['package_id'] == group['result_package_id']:
+            move_lines = StockMoveLine.search(group['__domain'])
+            StockPackageLevel.create({
+                'package_id': group['result_package_id'][0],
+                'picking_id': group['picking_id'][0],
+                'location_id': group['location_id'][0],
+                'location_dest_id': group['location_dest_id'][0],
+                'move_line_ids': [(6, 0, move_lines.ids)]
             })
```
MiquelRForgeFlow, line 133: I think it's missing to fill the `state` and `is_done` fields.

pedrobaeza: Both are computed non stored.

MiquelRForgeFlow: Oh, good

pedrobaeza (approved):
All OK now. Merging.

PR #1683, opened 2019-01-10
[12.0][MIG] mrp

MiquelRForgeFlow:
Migration scripts for `mrp` module.

--


alexis-via:
Here is my experience trying to migrate a DB with mrp module installed using the code of this PR (and several other PR available on OCA/openupgrade for v12) :
When Odoo loads purchase_stock/data/purchase_stock_data.xml, it will arrive on:
```
<record id="stock.warehouse0" model="stock.warehouse">
            <field name="buy_to_resupply" eval="True"/>
</record>
```
it will run the method _create_or_update_global_routes_rules() on stock.warehouse, as defined in the "stock" module. This method creates a stock.rule. This object stock.rule has a M2O field picking_type_id which is required. The values that are used to create this stock.rule is prepared by the method _get_global_route_rules_values() which is inherited in the MRP module. In the MRP module, that method returns the result dict with a key "pbm_mto_pull_id" which uses the field 'pbm_type_id' of the related warehouse as "picking_type_id" for the stock.rule. But 'pbm_type_id' is a new field in v12, so it is empty, so it blocks the migration:

```
odoo.tools.convert.ParseError: "null value in column "picking_type_id" violates not-null constraint
DETAIL:  Failing row contains (115, 1, 2019-04-19 09:18:12.165898, WH3: Stock → Pré-fabrication (MTO), 20, 1, 1, pull, 2019-04-19 09:18:12.165898, f, null, propagate, null, 138, 18, null, 0, 3, t, make_to_order, null, 1, null, null, null, null, null, manual).
```

MiquelRForgeFlow:
In the post-migration of `mrp`, I called a method to update the picking type values of the warehouses.

@alexis-via Could you try again if you have the same error?

OthmaneGhandi:
> In the post-migration of `mrp`, I called a method to update the picking type values of the warehouses.
> 
> @alexis-via Could you try again if you have the same error?

@mreficent Thanks for your efforts :  I tried it but still got the same issue, I think that the call of method **_get_global_route_rules_values()** happens before the run of the post-migration scripts

@alexis-via did you manage to avoid this issue

MiquelRForgeFlow:
As I understand the problem, the load of the `mrp` is done before the load of the `purchase_stock`, right? If so, the post-migration of `mrp` also is done before the load of the `purchase_stock`. Thus, in the post-migration of mrp, the `pbm_type_id` should have been filled and in the load of `purchase_stock` shouldn't have that error.

@OthmaneGhandi could you check this? Could you see if the `mrp` is loaded before the `purchase_stock`?

pedrobaeza:
@mreficent you can't determine the order of modules loading if there's no direct dependency, so you have to take this into account.

MiquelRForgeFlow:
@pedrobaeza Interesting. Is there a way to force the order?

MiquelRForgeFlow:
Well, in fact, if the `purchase_stock` is loaded before the `mrp`, it shouldn't be a problem. That issue only happened if the `mrp` was loaded before the `purchase_stock`. That's why, it's a bit strange that @OthmaneGhandi still has the same issue. @OthmaneGhandi could you post here your error traceback?

pedrobaeza:
Migration scripts must not depend on loading order. If we are doing something here that affects `purchase_stock`, then we should move that part to a common dependency module for both (`base`?), and check them the existence of tables/models/etc, or move this to end-migration.

MiquelRForgeFlow:
@OthmaneGhandi @alexis-via Hey guys, I solved the issue 🎉🎉

OthmaneGhandi:
> @OthmaneGhandi @alexis-via Hey guys, I solved the issue tadatada

Indeed because the method _create_or_update_sequences_and_picking_types create the picking types but we need to fill them in warehouse --> the write method (you added) was the missing puzzle 
Thanks @mreficent

OthmaneGhandi:
Hi @mreficent 

I got a weird error related to company_id constraint while trying to work with mrp production (That was migrated from V10 -> V11 then V11 -> V12)

I can't create a new mrp production because the stock.move couldn't be created
 
https://user-images.githubusercontent.com/6293989/61187869-0b23ff80-a66f-11e9-8393-2c39f72b63a6.png

I checked that in **_generate_finished_moves** (method related to create the stock move) all values (company_id within them) are well set --> but in the creation of the stock move, **some of** these values are not taken into consideration, which is very weird. Is there an explanation to this ?

@pedrobaeza do you have any suggestion for this issue ?
Is there any one who got this same issue ?

Thanks

OthmaneGhandi:
@mreficent @pedrobaeza 
Also when I try to update the mrp module I got this error 
https://user-images.githubusercontent.com/6293989/61188707-1bda7280-a67b-11e9-8fa3-f822becd5c92.png

MiquelRForgeFlow:
@OthmaneGhandi I was unable to reproduce exactly the same errors you showed, but I solved I little issue with updating the module.

If you can provide more info in how you get those errors, then I will try to help.

OthmaneGhandi:
> @OthmaneGhandi I was unable to reproduce exactly the same errors you showed, but I solved I little issue with updating the module.
> 
> If you can provide more info in how you get those errors, then I will try to help.

Thanks @mreficent 
Are you able to create **mrp production** with **stock move** lines ?

MiquelRForgeFlow:
@OthmaneGhandi how I can create mrp production with stock moves?

OthmaneGhandi:
@mreficent  when I go to manufacturing --> manufacturing order -> I choose a product => when I click save I got the error because it couldn't create stock moves == Consumed materials 

https://user-images.githubusercontent.com/6293989/61226326-20a23380-a71a-11e9-9439-be02bda9fcb3.png

MiquelRForgeFlow:
As I said, I don't have that problem. :(

pedroslopez:
> I got a weird error related to company_id constraint while trying to work with mrp production (That was migrated from V10 -> V11 then V11 -> V12)
> 
> I can't create a new mrp production because the stock.move couldn't be created
> 
> https://user-images.githubusercontent.com/6293989/61187869-0b23ff80-a66f-11e9-8393-2c39f72b63a6.png
> 
> _(...)_
>
> Is there any one who got this same issue ?
> 
> Thanks

@OthmaneGhandi  As far as the `company_id` validation error goes, we had this issue after migrating our database and it was due to `stock.rule`'s `company_id` field not being set. It's not required on that model, but it _is_ required when creating a manufacturing order. We were able to solve it by making sure the `company_id` field was properly set on all `stock.rule`s.

OthmaneGhandi:
> As far as company_id missing goes, we had this issue after migrating our database and it was due to stock.rule's company_id field not being set. It's not required on that model, but it _is_ required when creating a manufacturing order.

hats off @pedroslopez when a filled the company_id field in stock rule it worked. Thanks
Still I don't know why I got the error, do you have an explanation to that.

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
+---Fields in module 'mrp'---
+mrp          / mrp.bom                  / message_last_post (datetime)  : DEL 
+mrp          / mrp.bom                  / message_main_attachment_id (many2one): NEW relation: ir.attachment
+mrp          / mrp.bom                  / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
+mrp          / mrp.bom.line             / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
+# NOTHING TO DO
+
+mrp          / mrp.production           / activity_date_deadline (date) : not related anymore
+mrp          / mrp.production           / activity_date_deadline (date) : not stored anymore
```
pedrobaeza, line 9: I'm seeing now that this is a problem in all models that implements `mail.activity.mixin`: if we keep the column in DB, there is going to be incorrect readings of it, so we must rename it. There can be 2 approaches:

* Rename the column on each module.
* Try to get all the existing fields on `mail` migration and rename in "batch".

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+mrp          / mrp.production           / activity_date_deadline (date) : now a function
+mrp          / mrp.production           / message_last_post (datetime)  : DEL 
+mrp          / mrp.production           / message_main_attachment_id (many2one): NEW relation: ir.attachment
```
pedrobaeza, line 12: Put why nothing to do: computed on mail end-migration script

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+mrp          / mrp.production           / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
```
pedrobaeza: Put why nothing to do: renamed model on `base` module.

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+mrp          / mrp.production           / product_uom_qty (float)       : NEW
```
pedrobaeza: Put why nothing to do: computed field. By the way: why this doesn't reflect that is a computed stored field?

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+# NOTHING TO DO
+
+mrp          / mrp.unbuild              / activity_date_deadline (date) : not related anymore
+mrp          / mrp.unbuild              / activity_date_deadline (date) : not stored anymore
+mrp          / mrp.unbuild              / activity_date_deadline (date) : now a function
```
pedrobaeza, line 21: The same comment about this field coming from `mail.activity.mixin`

addons/mrp/migrations/12.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_column_renames = {
+    'mrp_workcenter_productivity_loss': [
+        ('loss_type', None),
```
pedrobaeza: You mustn't renamed this column, as the new field is related stored.

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+
+mrp          / mrp.workorder            / message_last_post (datetime)  : DEL
+mrp          / mrp.workorder            / message_main_attachment_id (many2one): NEW relation: ir.attachment
+mrp          / mrp.workorder            / product_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
+# NOTHING TO DO
+
+mrp          / product.product          / bom_line_ids (one2many)       : NEW relation: mrp.bom.line
+mrp          / product.product          / variant_bom_ids (one2many)    : NEW relation: mrp.bom
+mrp          / product.template         / bom_line_ids (one2many)       : NEW relation: mrp.bom.line
+# NOTHING TO DO
```
pedrobaeza: `NOTHING TO DO: They are one2many with already existing many2one`

addons/mrp/migrations/12.0.2.0/post-migration.py
```diff
@@ -0,0 +1,45 @@
...
+        WHERE wpl.loss_id IS NULL AND wpl.%s = wplt.loss_type
+        """, (AsIs(openupgrade.get_legacy_name('loss_type')), ),
+    )
+
+
+def fill_stock_warehouse_picking_types(env):
+    warehouses = env['stock.warehouse'].search([])
+    for warehouse in warehouses:
+        warehouse.write(
+            warehouse._create_or_update_sequences_and_picking_types())
```
pedrobaeza, line 36: Why are you doing this here? If you want to overwrite all picking types, then do it on `stock` end-migration.

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,120 @@
...
+NEW ir.model.access: mrp.access_uom_category_mrp_manager
+NEW ir.model.access: mrp.access_uom_uom_mrp_manager
+NEW ir.model.access: mrp.access_uom_uom_user
+DEL ir.model.access: mrp.access_mrp_message_mrp_manager
+DEL ir.model.access: mrp.access_mrp_message_mrp_user
+DEL ir.model.access: mrp.access_product_puom_categ
+DEL ir.model.access: mrp.access_product_uom_categ_mrp_manager
+DEL ir.model.access: mrp.access_product_uom_mrp_manager
+DEL ir.model.access: mrp.access_product_uom_user
+DEL ir.sequence: mrp.sequence_mrp_prod (noupdate)
```
pedrobaeza: Why keeping this sequence or this XML-ID?

pedrobaeza (changes requested):

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: and which of the to approaches you recommend?

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: it's a old analysis_work from an old analysis

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
pedrobaeza: If feasible, approach 2 will be desirable for doing things once, and not worrying on each module.

addons/mrp/migrations/12.0.2.0/post-migration.py
MiquelRForgeFlow: to have `pbm_type_id` field filled.

MiquelRForgeFlow: Why in the end-migration of stock and not here?

MiquelRForgeFlow: It breaks in `purchase_stock` if we do this in an end-migration. See https://github.com/OCA/OpenUpgrade/pull/1683 comment.

pedrobaeza: I was asking that for avoiding calling several times to this method if there are other modules adding sequences like this. Seeing the problem, it's OK to do it here, but as always, please put an explanatory comment saying why (like the comment you have made).

pedrobaeza: And also comment about `pbm_type_id`.

addons/mrp/migrations/12.0.2.0/pre-migration.py
MiquelRForgeFlow: Oh, yes, I saw wrong when I did it. I thought it wasn't stored.

MiquelRForgeFlow: maybe `copy_columns`? I think I wanted to save the column because I feared it would change the value in the loading of data because it is a related.

pedrobaeza: No if the column exists. It happens the same as stored computed fields.

addons/mrp/migrations/12.0.2.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: but approach 2 I think is unfeasible. Because, in pre-migration of mail, there are a lot of modules that have not yet loaded and we cannot know it hat moment which models inherit of mail.activity.mixin

pedrobaeza: One possible thing is to look for any column with this name, and rename it. If one is incorrectly renamed, it can be re-renamed again on the module migration itself (in core I don't think it's needed).

MiquelRForgeFlow: and how you get a list of all tables? 🤔

MiquelRForgeFlow: maybe what should only do is just in the end-migration do a recompute of this column for all the models inheriting mail.activity.mixin.

pedrobaeza: You can query certain special tables. Look how openupgradelib `column_exists` is implemented. And about end-migration, it doesn't work. The problem is the existence of the column, not any computation.

MiquelRForgeFlow:
updated. @pedrobaeza could you check again?

pedrobaeza (approved):

PR #1684, opened 2019-01-10
[12.0][MIG] sms

MiquelRForgeFlow:
`#NothingToDo`

It's an auto_install.

--


rvalyi (approved):

odoo/openupgrade/doc/source/modules110-120.rst
```diff
@@ -535,7 +535,7 @@ missing in the new release are marked with |del|.
 +--------------------------------------------+-------------------------------------------------+
 |web_kanban_gauge                            | Nothing to do                                   |
 +--------------------------------------------+-------------------------------------------------+
-| |del| web_planner                          | Done. Merged in web                             |
+| |del| web_planner                          | Nothing to do: Merged in web                    |
```
pedrobaeza: This is correct as it was, as you have done something, which is merge it on `web`. Although it's the same at operational level (both ways get that the module is installed on previous version), we should be correct at semantic level.

pedrobaeza (changes requested):
Let the other thing untouched and it will be good to go.

odoo/openupgrade/doc/source/modules110-120.rst
MiquelRForgeFlow: So, following your logic, I updated the `auth_crypt` case 😏

pedrobaeza (approved):
Yeah, great, the other one doesn't follow the usual semantic. Thanks for checking that.

pedrobaeza:
You can merge when this finishes.

PR #1685, opened 2019-01-10
[12.0][MIG] website

MiquelRForgeFlow:
Migration scripts for `website` module.

Working in progress: Adapt created pages to new bootstrap version.

--


addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,111 @@
...
+        UPDATE website_redirect wr
+        SET active = FALSE
+        WHERE wr.url_from IS NULL OR wr.url_to IS NULL
+        """
+    )
+
+
+def apply_bootstrap_4(view):
+    text = view.arch_db
+    # TO BE FILLED
```
LoisRForgeFlow, line 37: We can complete this in a different PR, to unlock dependencies to this module and allow parallel work?

LoisRForgeFlow (commented):
Migration does not break however you loose your website as the views are not updated to B4. In the other hand there is a lot of dependencies to this module that need to move forward as well

addons/website/migrations/12.0.1.0/post-migration.py
pedrobaeza: Yes, of course!

MiquelRForgeFlow: Yes, that would be nice. I would like to work on this `apply_bootstrap_4` method in a different PR.

pedrobaeza: @Yajo is going to work on that topic next week. Do you have already something on that? I will review remaining PRs also next week.

MiquelRForgeFlow: Ok. Then move forward, do merge and I will open a new PR for this method. I have an initial draft for that script. Then @Yajo can complete it.

LoisRForgeFlow:
@pedrobaeza @mreficent merge then?

addons/website/migrations/12.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,121 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="aboutus" model="ir.ui.view">
```
pedrobaeza: This seems a bit aggressive and I don't know why these records are directly marked as noupdate="1". The technique would be (even to be backported to previous versions):

1. Check if the website view is noupdate="1".
2. If so, check if the view is still valid in current version (this should be done in end-migration).
3. If not, copy and disable it.
4. If not, then restore original view, and mark it somehow for later review.

This is ideal method. Meanwhile, you can simply in this case remove these records for not being forced updated and being treated the same way as the rest of website views.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
+---Fields in module 'website'---
+website      / ir.attachment            / key (char)                    : NEW
```
pedrobaeza, line 2: Mention that this is nothing to do because for now is only used in cropped images for targetting the original one.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
+---Fields in module 'website'---
+website      / ir.attachment            / key (char)                    : NEW 
+website      / ir.attachment            / website_id (many2one)         : NEW relation: website
```
pedrobaeza, line 5: You have to fill this with the website (you can search the first one with the same company) when public=True, or the attachments won't be shown in resource selector.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+website      / ir.ui.view               / page_ids (one2many)           : not a function anymore
+website      / ir.ui.view               / page_ids (one2many)           : is now stored
+website      / ir.ui.view               / website_meta_og_img (char)    : NEW
```
pedrobaeza: Mention that this field is used for linking the image that is going to be showed when linking the page on social media, but leaving it empty gets a default one (company logo), so it's correct to be nothing to do, although it can be improved selecting another, for example, the first image found on the page.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
+---Fields in module 'website'---
+website      / ir.attachment            / key (char)                    : NEW 
+website      / ir.attachment            / website_id (many2one)         : NEW relation: website
+website      / ir.ui.view               / page_ids (one2many)           : not a function anymore
+website      / ir.ui.view               / page_ids (one2many)           : is now stored
+website      / ir.ui.view               / website_meta_og_img (char)    : NEW 
+website      / res.partner              / website_id (many2one)         : NEW relation: website
```
pedrobaeza, line 17: Mention for these 2 fields that there's no way to guess if the user was created from v11 website (signed up), so nothing to do.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+website      / res.users                / website_id (many2one)         : NEW relation: website
+# NOTHING TO DO
+
+website      / website                  / auth_signup_uninvited (selection): NEW selection_keys: ['b2b', 'b2c']
+website      / website                  / country_group_ids (many2many) : NEW relation: res.country.group
+website      / website                  / google_maps_api_key (char)    : NEW
+website      / website                  / social_default_image (binary) : NEW attachment: True
+website      / website                  / social_instagram (char)       : NEW
+website      / website                  / specific_user_account (boolean): NEW
+website      / website                  / theme_id (many2one)           : NEW relation: ir.module.module
```
pedrobaeza: If there is a theme module installed (you can distinguish it by module category), you have to assign it to all existing websites.

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,111 @@
+# Copyright 2018 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_website_company_id(cr):
```
pedrobaeza: Better to do this in pre for avoiding warning messages

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,111 @@
...
+            social_linkedin = c.social_linkedin,
+            social_twitter = c.social_twitter,
+            social_youtube = c.social_youtube
+        FROM res_company c
+        WHERE w.company_id = c.id
+        """
+    )
+
+
+def fill_website_name(cr):
```
pedrobaeza: Better to do this on pre for avoiding warning messages

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+# DONE: post-migration: filled fields in case they are empty
+
+website      / website.menu             / parent_left (integer)         : DEL
+website      / website.menu             / parent_path (char)            : NEW 
+website      / website.menu             / parent_right (integer)        : DEL 
+# DONE: post-migration: called _parent_store_compute method
+
+website      / website.page             / header_color (char)           : NEW
+website      / website.page             / header_overlay (boolean)      : NEW 
+website      / website.page             / is_published (boolean)        : NEW
```
pedrobaeza: Group this one with the lines about `website.page.website_published` field and the work to be done is to rename the column `website_published` > `is_published`.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+# NOTHING TO DO
+
+website      / website.page             / website_id (many2one)         : is now stored
+website      / website.page             / website_ids (many2many)       : DEL relation: website
+# DONE: post-migration: create a copy page for each website in case for pages with more than one website
```
pedrobaeza, line 61: AFAIK, pages are copied any way although you have only one website, so you always work with a copy, not the master one.

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+
+website      / website.page             / website_published (boolean)   : not stored anymore
+website      / website.page             / website_published (boolean)   : now a function
+# NOTHING TO DO
+
+website      / website.redirect         / type (selection)              : now required, default = function
+website      / website.redirect         / url_from (char)               : now required
+website      / website.redirect         / url_to (char)                 : now required
+# DONE: post-migration: filled in case they are empty or put as active = False
```
pedrobaeza, line 71: Please do it on pre for avoiding warnings

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+
+---XML records in module 'website'---
+DEL ir.actions.act_url: website.action_website_homepage (noupdate)
```
pedrobaeza, line 74: Remove it manually

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+NEW ir.actions.act_window: website.action_website_add_features
+NEW ir.actions.act_window: website.action_website_list
+DEL ir.actions.act_window: website.action_module_theme (noupdate)
```
pedrobaeza, line 75: Remove it manually

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+DEL ir.actions.act_window: website.action_module_website (noupdate)
```
pedrobaeza, line 76: Remove it manually

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,153 @@
...
+NEW ir.actions.server: website.ir_actions_server_website_google_analytics
+NEW ir.attachment: website.s_carousel_default_image_1 (noupdate)
```
pedrobaeza, line 84: Put a comment about these files specifying that you don't remove these ones, as they might be used in some places of the website.

pedrobaeza (changes requested):
And please add me in copyrights

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: ?

pedrobaeza: I think the specification is clear enough.

MiquelRForgeFlow: No, it's not clear to me. I have data in a test database v12 where I have ir.attachment records with public=True that don't have website_id :S

addons/website/migrations/12.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: do you mean unlinking manually each of these records?

pedrobaeza: Yes, with the openupgradelib function

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: The question is: one attachment that is shown in v11, is it the same shown on v12? I think the key is this field.

pedrobaeza: Or maybe public=True still control this and this field is for restricting the website visibility? Any way, please put the result of your investigation as comment for later readings.

addons/website/migrations/12.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: done in pre-migration

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: This field was created in this commit: https://github.com/odoo/odoo/commit/c0968e68eda55df624fb94016d036a2af598b10c#diff-c65a40bf26f8b99e5c9fb705f99220bfR16
It's only filled in the `create`, that calls the `get_current_website` method. So, when an attachment is created, it searches if there is a current website from http.request.

But for old attachments, we don't have a way to know to which website could have been linked :S

pedrobaeza: OK, then add this explanation and leave it "Nothing to do"

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: In v12 works differently in multiwebsite. I tested it when I did the `apply_multiwebsite` script.

MiquelRForgeFlow:
Ready. Can you check?

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,84 @@
...
+
+def assign_theme(env):
+    theme_category = env.ref('base.module_category_theme')
+    theme_module = env['ir.module.module'].search(
+        [('state', '=', 'installed'),
+         ('category_id', '=', theme_category.id)],
+        limit=1,
+    )
+    websites = env['website'].search([])
+    websites.write({'theme_id': theme_module.id})
```
pedrobaeza: Do the write only if found

addons/website/migrations/12.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: I have been testing this and the mechanism is:

- If there's a view with `website_id` specified and the same key as the master one, use this.
- If not, use the master one without website but with that key.
- On a fresh DB, you only have master views, but on first touch you make on a page, then a copy is populated with specific website.

Thus, I still prefer the approach of copying all website views for each website, being one or several. Why? Because that way, we can do a second step, which is reset master views to v12 original content, and leave the others untouched, and we will have both of best worlds.

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,77 @@
...
+    fill_website_redirect_type(cr)
+    fill_website_redirect_urls(cr)
+    openupgrade.delete_records_safely_by_xml_id(
+        env, [
+            'website.aboutus',
+            'website.s_image_text_default_image',
+            'website.main_menu',
+            'website.s_cover_default_image',
+            'website.default_website',
+            'website.menu_contactus',
```
pedrobaeza: Why removing this? Are you using this technique instead of loading the noupdate records for website.menu? It's better to update the records due to possible references to this record.

pedrobaeza:
Please consider also the reset of all the website QWeb templates according to my comment on the duplicates.

addons/website/migrations/12.0.1.0/pre-migration.py
MiquelRForgeFlow: But you said me before that you wanted to delete the records instead to update them in `noupdate_changes.xml` 😩

pedrobaeza: No, I said to remove that records on the noupdate_records.xml file, not to remove them on Odoo!

pedrobaeza: Any way, due to the investigation of website views duplication, if you duplicate views before loading noupdate records, we can keep the file as is.

MiquelRForgeFlow: I am so confused now. Tell me what should I do clearly.

pedrobaeza: 1. Duplicate QWeb views for each website, being one or several websites.
2. Load noupdate_records.xml as is

pedrobaeza (approved):
Let's go for the Bootstrapocalypse!

PR #1698, opened 2019-01-21
[9.0] Fix that lot qty done won't be set to total sum of quants qty

paulius-sladkevicius:
Issue is that all `stock.pack.operation.lot` get a total amount of all quant qty, what is wrong, it should show qty done from pack operation.

--

### DIFF
--- a/addons/stock/migrations/9.0.1.1/post-migration.py
+++ b/addons/stock/migrations/9.0.1.1/post-migration.py
@@ -36,18 +36,18 @@ def _migrate_pack_operation(env):
     qty_done on stock.pack.operation for transferred pickings.
     """
     env.cr.execute(
-        "select o.id, o.%(lot_id)s, p.state, sum(q.qty) "
+        "select o.id, o.%(lot_id)s, p.state, o.qty_done "
         "from stock_pack_operation o "
-        "join stock_quant q on q.lot_id=o.%(lot_id)s "
         "join stock_picking p on o.picking_id=p.id "
-        "group by o.id, o.%(lot_id)s, p.state",
+        "where o.%(lot_id)s is not null "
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't see the problem in previous code. Please expose your real case here with the data you have  in v8 and how they are migrated to v9.

paulius-sladkevicius:
Lets say in v8 you have serial number that has N related quants with different qty e.g. total sum 10. After migration to v9 you get `stock.pack.operation.lot` that has 10 qty done while you had in v8 only 1 qty done on a `stock.pack.operation`. Let me know if this gives better overview of issue.

pedrobaeza:
Please put the exact data in each table you have in v8.

paulius-sladkevicius:
Here is one example (it's from v9 db):

```
select openupgrade_legacy_9_0_lot_id, qty_done from stock_pack_operation where id = 6663;
 openupgrade_legacy_9_0_lot_id |  qty_done  
-------------------------------+------------
                          1050 | 15000.0000
(1 row)

select qty from stock_quant where lot_id = 1050;
   qty   
---------
   19125
    9600
    3200
     600
 18462.5
     800
    5200
    5200
    5200
    2000
     600
     200
  2412.5
(13 rows)

select sum(qty) from stock_quant where lot_id = 1050;
  sum  
-------
 72600
(1 row)

select qty, lot_id from stock_pack_operation_lot where operation_id = 6663;
  qty  | lot_id 
-------+--------
 72600 |   1050
```
In a final query you can see that data is wrong, `qty` should be 15000

pedrobaeza (approved):
Sorry for the late review. Confirm proper behavior, so merging.

PR #1700, opened 2019-01-22
[12.0][FIX] openpgrade_records: be able to do analysis from v11

MiquelRForgeFlow:
Trying to perform an analysis of records, I have issues:

1) was blocked by the `_check_credentials` method of res.users. In v11 you have `password` and `password_crypt` fields in res.users, but in v12 this two fields got merged into one (`password`). So, when accessing to the v11 database, the problem raised.

2) in ir_http in web, it's called get_currencies method and breaks. Happens that in v11, `decimal_places` field of res.currency was `store=False`, and in v12 this field is `store=True`.

For the second commit, absolutely secondary, is for when openupgrade open v11 connections. It surprised me that after using this connections, they never get closed. So I thought to add this secondary fix to close them. I am not sure if it's necessary but here it is. I can take it out.

--

### DIFF
--- a/odoo/addons/openupgrade_records/__manifest__.py
+++ b/odoo/addons/openupgrade_records/__manifest__.py
@@ -16,6 +16,7 @@
         'views/install_all_wizard.xml',
         'security/ir.model.access.csv',
     ],
+    'depends': ['base', 'web'],  # web is need for adapt_to_v11.py
     'installable': True,
     'external_dependencies': {
         'python': ['odoorpc', 'openupgradelib'],
--- a/odoo/addons/openupgrade_records/models/__init__.py
+++ b/odoo/addons/openupgrade_records/models/__init__.py
@@ -3,3 +3,4 @@ from . import comparison_config
 from . import analysis_wizard
 from . import generate_records_wizard
 from . import install_all_wizard
+from . import adapt_to_v11
--- a/odoo/addons/openupgrade_records/models/adapt_to_v11.py
+++ b/odoo/addons/openupgrade_records/models/adapt_to_v11.py
@@ -1,0 +1,81 @@
+# coding: utf-8
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import math
+from odoo import models
+from odoo.exceptions import AccessDenied
+from odoo.http import request
+
+
+class Users(models.Model):
+    _inherit = "res.users"
+
+    def _check_credentials(self, password):
+        cr = self.env.cr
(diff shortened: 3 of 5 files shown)


pedrobaeza:
Can you put the error before the fix?

MiquelRForgeFlow:
Of course, see this (using button Test connection):
```python
2019-01-22 20:00:13,471 13653 ERROR v11_OU2 odoo.http: Exception during JSON request handling. 
  File ".../OpenUpgrade/odoo/http.py", line 654, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File ".../OpenUpgrade/odoo/http.py", line 312, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File .../OpenUpgrade/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File ".../OpenUpgrade/odoo/http.py", line 696, in dispatch
    result = self._call_function(**self.params)
  File ".../OpenUpgrade/odoo/http.py", line 345, in _call_function
    return self.endpoint(*args, **kwargs)
  File ".../OpenUpgrade/odoo/http.py", line 939, in __call__
    return self.method(*args, **kw)
  File ".../OpenUpgrade/odoo/http.py", line 517, in response_wrap
    response = f(*args, **kw)
  File ".../OpenUpgrade/addons/web/controllers/main.py", line 811, in authenticate
    request.session.authenticate(db, login, password)
  File ".../12.0/OpenUpgrade/odoo/http.py", line 1039, in authenticate
    uid = odoo.registry(db)['res.users'].authenticate(db, login, password, env)
  File ".../OpenUpgrade/odoo/addons/base/models/res_users.py", line 581, in authenticate
    uid = cls._login(db, login, password)
  File ".../OpenUpgrade/odoo/addons/base/models/res_users.py", line 560, in _login
    user._check_credentials(password)
  File ".../OpenUpgrade/odoo/addons/base/models/res_users.py", line 317, in _check_credentials
    raise AccessDenied()
odoo.exceptions.AccessDenied: Access denied
```

pedrobaeza:
@StefanRijnhart how did you do it in first place the analysis?

MiquelRForgeFlow:
@pedrobaeza here you have an example of superposed force pushes you asked me the other day. 
selection_050
Before this one I did another two amends and force pushes, but they have been "deleted" by this last one.

pedrobaeza:
I see both, although it's true that there were 3 force pushes, but they are indicated:

seleccion_032

For me this is still enough.

MiquelRForgeFlow:
Now I understand that I should use different ports to properly do the analysis. Closed.

PR #1702, opened 2019-01-25
[IMP] stock: Fill missing lots for stock.move.line

pedrobaeza:
On v10, `reserved_quant_id` field on stock.move.operation.link is filled unless manual operations or extra quantities are added on transfer wizard, which technically means that they fall under this part of code:

https://github.com/OCA/OpenUpgrade/blob/10.0/addons/stock/models/stock_picking.py#L759

The SQL query for inserting stock.move.line can't be modified as information is taken from stock.move.operation.link instead of stock.pack.operation/stock.operation.lot as v10 grouped together several stock.move lines with the same product in the same operation record, making impossible to split the lot quantity across each move, so we call this method afterwards for filling missing lots when possible.

Heuristic:

* Find stock.move.operation.link records without reserved quant, but the linked operation has some lots.
* Check the stock.move.line (sml) new records associated to the move. If there's only one sml with the same quantity of stock.pack.operation.
* If found, then we can conclude the lot is that one.
* If not, we try to see that all the stock.move.line of the same product makes the total of the stock.pack.operation.lot for concluding the same.

@Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -1,8 +1,8 @@
-# -*- coding: utf-8 -*-
-# Copyright 2017-2018 Tecnativa - Pedro M. Baeza
+# Copyright 2017-2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.exceptions import UserError
+from odoo.tools import float_compare
 from openupgradelib import openupgrade
 from psycopg2.extensions import AsIs
 
@@ -125,6 +125,11 @@ def create_stock_move_line(env):
       stock.move.line records are created from other source (quants) later
       in `create_stock_move_line_reserved` method.
     * For incoming transfers not yet validated, they are created later.
+
+    Information is taken from stock.move.operation.link instead of
+    stock.pack.operation/stock.operation.lot as v10 grouped together several
+    stock.move lines with the same product in the same operation record, making
+    impossible to split the lot quantity across each move.
     """
     openupgrade.logged_query(
         env.cr, """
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -189,6 +195,64 @@ def create_stock_move_line(env):
...
+            JOIN stock_move_operation_link smol
+                ON smol.operation_id = spo.id
+            JOIN stock_move sm ON sm.id = smol.move_id
+            JOIN stock_pack_operation_lot spol
+                ON spol.operation_id = spo.id
+        WHERE smol.reserved_quant_id IS NULL""",
+    )
+    rounding = env['decimal.precision'].precision_get(
+        'Product Unit of Measure')
+    Move = env['stock.move']
```
MiquelRForgeFlow, line 233: why not `move_model` instead?

addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -189,6 +195,64 @@ def create_stock_move_line(env):
...
 
 
+@openupgrade.logging()
+def fill_missing_lots_for_sml(env):
+    """On v10, `reserved_quant_id` field on stock.move.operation.link is
+    filled unless manual operations or extra quantities are added on transfer
+    wizard, which technically means that they fall under this part of code:
+
+    https://github.com/OCA/OpenUpgrade/blob/10.0/addons/stock/models/
+    stock_picking.py#L759
```
MiquelRForgeFlow, line 205: add something like `on the part of "# 2) then, process the remaining part"`, because the line number may change if the future in we update from upstream again...

MiquelRForgeFlow (approved):
LGTM overall, I won't check on detail. Good catch, interesting the heuristic

pedrobaeza:
The error we are having is totally unrelated:
```
2019-01-25 18:57:07,743 6667 [1;32m[1;49mINFO[0m openupgrade odoo.addons.base.ir.ir_model: Deleting 36@stock.location (stock.stock_location_shop0)
2019-01-25 18:57:07,753 6667 [1;31m[1;49mERROR[0m openupgrade odoo.sql_db: bad query: b'DELETE FROM stock_location WHERE id IN (36)'
ERROR: null value in column "location_dest_id" violates not-null constraint
DETAIL:  Failing row contains (22, null, 2018-03-28 12:21:27.700135, null, 1, null, 25.000, make_to_stock, 25.0, null, 1, 11, 8, 10, 1, null, draft, 25.000, null, null, null, 2018-03-28 12:21:40, null, 1, null, null, f, t, null, 2018-03-28 12:21:40, f, 1, 19, null, [E-COM10] Apple Wireless Keyboard, null, null, null, 2018-03-28 12:25:07.030232, null, 10, null, null, f, null, null, null, null, null, null, null, null, null, null, 0.00, 3, Chic/IN/00001, null, null, null, null, null, null, null, null).
CONTEXT:  SQL statement "UPDATE ONLY "public"."stock_move" SET "location_dest_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "location_dest_id""

2019-01-25 18:57:07,753 6667 [1;33m[1;49mWARNING[0m openupgrade odoo.addons.base.ir.ir_model: Could not delete obsolete record with id: 36 of model stock.location
Please refer to the log message right above
2019-01-25 18:57:07,754 6667 [1;32m[1;49mINFO[0m openupgrade odoo.addons.base.ir.ir_model: Deleting 14@stock.picking.type (stock.chi_picking_type_out)
2019-01-25 18:57:07,757 6667 [1;31m[1;49mERROR[0m openupgrade odoo.sql_db: bad query: b'DELETE FROM stock_picking_type WHERE id IN (14)'
ERROR: null value in column "picking_type_id" violates not-null constraint
DETAIL:  Failing row contains (9, 1, Chic: Stock -> Customers, 20, null, 1, t, 2018-03-28 12:21:27.700135, move, 2018-03-28 12:21:27.700135, null, propagate, null, 9, 36, null, 0, 3, t, make_to_stock, 10, 9, null).
CONTEXT:  SQL statement "UPDATE ONLY "public"."procurement_rule" SET "picking_type_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "picking_type_id""

2019-01-25 18:57:07,757 6667 [1;33m[1;49mWARNING[0m openupgrade odoo.addons.base.ir.ir_model: Could not delete obsolete record with id: 14 of model stock.picking.type
Please refer to the log message right above
2019-01-25 18:57:07,757 6667 [1;32m[1;49mINFO[0m openupgrade odoo.addons.base.ir.ir_model: Deleting 11@stock.picking.type (stock.chi_picking_type_in)
2019-01-25 18:57:07,759 6667 [1;31m[1;49mERROR[0m openupgrade odoo.sql_db: bad query: b'DELETE FROM stock_picking_type WHERE id IN (11)'
ERROR: null value in column "picking_type_id" violates not-null constraint
DETAIL:  Failing row contains (15, 1, Chicago Warehouse:  Buy, 20, null, 1, t, 2018-03-28 12:24:46.760593, buy, 2018-03-28 12:24:46.760593, null, none, null, 36, null, null, 0, 3, t, make_to_stock, 5, 12, null).
CONTEXT:  SQL statement "UPDATE ONLY "public"."procurement_rule" SET "picking_type_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "picking_type_id""
```
@mreficent do you know if in last upstream merge, you found something similar?

addons/stock/migrations/11.0.1.1/post-migration.py
pedrobaeza: Well, for being shorter, and following latest Odoo convention about this (putting the name in upper).

MiquelRForgeFlow:
In the last update upstream I didn't check the log, but we can check now, but I remember the CI was green so maybe that error wasn't in it

addons/stock/migrations/11.0.1.1/post-migration.py
pedrobaeza: I can put the permalink

addons/stock/migrations/11.0.1.1/post-migration.py
MiquelRForgeFlow: but OpenUpgrade should follow OCA convention, not Odoo convention 😁

pedrobaeza: But there's no convention on that one :wink:

sergio-teruel (approved):

pedrobaeza:
And green!

PR #1706, opened 2019-01-28
[11.0][FIX] join procurement_rule and stock_location_path to stock_location_route

hbrunn:
we need this join, otherwise we get the Cartesian product of both tables

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -80,7 +80,8 @@ def create_specific_procurement_rules_from_globals(env):
             INSERT INTO %s
             (%s)
             SELECT %s
-            FROM %s t, stock_location_route slr""", (
+            FROM %s t, stock_location_route slr
+            WHERE t.route_id=slr.id""", (
                 AsIs(table),
                 AsIs(", ".join(dest_columns)),
                 AsIs(", ".join(src_columns)),


pedrobaeza:
@hbrunn what we want is indeed the cartesian product, as these are global rules that applies to all. As global rules don't exist on v11, we take each of the global rule, and copy for each of the stock.location.path, getting N*M records.

If you take a look in pre-migration, you'll see that this is not going to work (or got the same result if null is compared to null), as we only copy in the temp table the rules without route:

https://github.com/OCA/OpenUpgrade/blob/7d92534f1b84b68394ccbb5d822be4a6c105ce96/addons/stock/migrations/11.0.1.1/pre-migration.py#L15

hbrunn:
I have a database where procurement_rule is empty, this then causes integrity errors. Am I fixing a symptom here?

pedrobaeza:
But procurement rules can't be empty if not badly manipulated AFAIK...

hbrunn:
okay, then I keep this only for my database and close the PR

pedrobaeza:
Well, I'm not 100% sure, but I have migrated since now about 10 DBs to v11 without problems. Maybe your DB is migrated also from lower versions and the fix should be there creating that procurement rules?

hbrunn:
it's indeed a v8 database, I'll have a look where this goes wrong

PR #1708, opened 2019-01-29
[12.0][UPD] openupgrade_analysis

MiquelRForgeFlow:
Here we go...

In theory this analysis is correct, but after seeing the changes I am not sure :S

--


addons/base_address_city/migrations/12.0.1.0/openupgrade_analysis.txt
```diff
@@ -1,3 +1,847 @@
...
+base_address_city / report.paperformat       / margin_right (float)          : previously in module anonymization
+base_address_city / report.paperformat       / margin_top (float)            : previously in module anonymization
+base_address_city / report.paperformat       / name (char)                   : previously in module anonymization
+base_address_city / report.paperformat       / orientation (selection)       : previously in module anonymization
+base_address_city / report.paperformat       / page_height (integer)         : previously in module anonymization
+base_address_city / report.paperformat       / page_width (integer)          : previously in module anonymization
+base_address_city / report.paperformat       / report_ids (one2many)         : previously in module anonymization
+base_address_city / res.bank                 / __last_update (datetime)      : previously in module anonymization
+base_address_city / res.bank                 / active (boolean)              : previously in module anonymization
+base_address_city / res.bank                 / bic (char)                    : pr
```
MiquelRForgeFlow, line 846: What happened here in this file? 🤯

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
```diff
@@ -117,178 +117,13 @@ account      / res.company              / invoice_reference_type (selection): NE
 account      / res.company              / qr_code (boolean)             : NEW 
 account      / res.company              / transfer_account_code_prefix (char): NEW 
 ---XML records in module 'account'---
-DEL account.financial.report: account.account_financial_report_assets0 (noupdate)
```
StefanRijnhart, line 120: It looks like all these removed entries should be kept. If this `DEL` entry is removed here, it means that the xml id should still be present in the module in Odoo 12 but I don't see it.

StefanRijnhart:
Looks like something went wrong. Was your comparison configuration correct?

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
MiquelRForgeFlow: Yeah, that' why I suspected it's not correct.

MiquelRForgeFlow:
I don't know. I installed everything without demo data and did the comparison as expected. :S

StefanRijnhart:
Here is the cranky script that I use to create a full analysis between Odoo instances
https://github.com/OCA/OpenUpgrade/wiki/Crude-script-to-create-the-full-analysis-between-two-versions-of-Odoo

MiquelRForgeFlow:
Thanks. I see you use different ports. I always used same port.

MiquelRForgeFlow:
If I get a correct analysis, I will create a new PR.

PR #1711, opened 2019-02-01
[12.0][MIG] sale

MiquelRForgeFlow:
Migration scripts for `sale` module.

--


addons/sale/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_column_renames = {
+    'payment_transaction': [
+        ('sale_order_id', None),
```
rvalyi, line 14: here you are trying to rename the sale_order_id column. However, this column will only exist if you installed the sale_payment module on v11 https://github.com/odoo/odoo/blob/11.0/addons/sale_payment/models/payment.py#L16
Thus this migration you happen in the sale_payment migration module and not in the sale module.
For instance I don't have this sale_payment module installed and then I get this error:
```
2019-02-11 15:08:59,154 126 INFO openupgrade-to-12 OpenUpgrade: sale: pre-migration script called with version 11.0.1.1
2019-02-11 15:08:59,155 126 INFO openupgrade-to-12 OpenUpgrade: table payment_transaction, column sale_order_id: renaming to openupgrade_legacy_12_0_sale_order_id
2019-02-11 15:08:59,155 126 ERROR openupgrade-to-12 odoo.sql_db: bad query: b'ALTER TABLE "payment_transaction" RENAME "sale_order_id" TO "openupgrade_legacy_12_0_sale_order_id"'
ERROR: column "sale_order_id" does not exist                                                    
                                                                                                
2019-02-11 15:08:59,155 126 ERROR openupgrade-to-12 OpenUpgrade: sale: error in migration script sale/migrations/12.0.1.1/pre-migration.py: column "sale_order_id" does not exist
                                                                                                
2019-02-11 15:08:59,155 126 ERROR openupgrade-to-12 OpenUpgrade: column "sale_order_id" does not exist
                                                                                                
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 1445, in wrapped_function      
    if use_env2 else cr, version)                                                               
  File "/odoo/src/addons/sale/migrations/12.0.1.1/pre-migration.py", line 44, in migrate        
    openupgrade.rename_columns(env.cr, _column_renames)                                         
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 387, in rename_columns         
    'ALTER TABLE "%s" RENAME "%s" TO "%s"' % (table, old, new,))
psycopg2.ProgrammingError: column "sale_order_id" does not exist
```

rvalyi (changes requested):

addons/sale/migrations/12.0.1.1/pre-migration.py
rvalyi: and the same happens for the next payment_tx_id column rename in the sale_order table!

addons/sale/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,46 @@
...
+    possible new rows added in the migration"""
+    cr.execute(
+        "ALTER TABLE sale_order_line ADD COLUMN display_type varchar",
+    )
+    openupgrade.logged_query(
+        cr, """
+        INSERT INTO sale_order_line (order_id, layout_category_id,
+            sequence, name, price_unit, product_uom_qty, customer_lead,
+            display_type, create_uid, create_date, write_uid, write_date)
+        SELECT sol.invoice_id, sol.layout_category_id,
```
rvalyi: there is no invoice_id column in the sale_order_line table. You probably meant ```sol.order_id```, right?

rvalyi (dismissed):

cuongnmtm:
Hello guys, i have an issue when migrated to 12.0: 
Sale Orders didn't compute Invoice Status. That because the field `qty_delivered` on `sale.order.line` change way to compute here
(https://github.com/odoo/odoo/commit/3bf8a62e0dd99f7dad42718f895abf4fd53f996f#diff-39d0613187d36f16c541734e80925e9dR880).
So that when migrate, I have to force run these functions to recompute several fields (`_compute_qty_delivered_method`, `_compute_qty_delivered`)

DaoTheKien:
Can anybody help me with this issue? Thanks
image

addons/sale/migrations/12.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,59 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.column_exists(env.cr, 'sale_order', 'payment_tx_id'):
+        # from sale_payment module
+        openupgrade.rename_columns(env.cr, _column_renames)
+    openupgrade.rename_fields(env, _field_renames)
+    if openupgrade.column_exists(env.cr, 'sale_order', 'requested_date'):
+        # from sale_order_dates module
+        openupgrade.drop_columns(env.cr, [('sale_order', 'commitment_date')])
+        openupgrade.rename_fields(env, _field_renames_order_dates)
```
MiquelRForgeFlow, line 72: @pedrobaeza I get an error here. Should I do a new `drop_fields` instead of the `drop_columns`? Or should I just do a `rename_columns` instead of this `rename_fields`? 🤔

MiquelRForgeFlow: solved

MiquelRForgeFlow:
:green_apple:

addons/sale/migrations/12.0.1.1/post-migration.py
```diff
@@ -19,16 +19,82 @@ def type_change_payment_transaction_and_sale_order(env):
...
+    # fill is_expense
+    openupgrade.logged_query(
+        cr, """
+        UPDATE sale_order_line sol
+        SET is_expense = TRUE
+        FROM account_analytic_line aal
+        JOIN product_product pp ON aal.product_id = pp.id
+        LEFT JOIN product_template pt ON pp.product_tmpl_id = pt.id
+        WHERE aal.so_line = sol.id AND aal.amount <= 0
+            AND pt.expense_policy IS NOT NULL AND pt.expense_policy != 'no'
```
pedrobaeza, line 32: Are you sure this is the good condition? As far as I saw on Friday, it's not the same the line that originates the expense than the extra one that are generated after this. These last ones are the ones which need to be put as `is_expense`, but I fear that with this, you are marking both.

addons/sale/migrations/12.0.1.1/post-migration.py
```diff
@@ -19,16 +19,82 @@ def type_change_payment_transaction_and_sale_order(env):
...
+
 def recompute_sale_order_line_invoicing(env):
     sale_lines = env['sale.order.line'].search([])
     sale_lines._compute_untaxed_amount_invoiced()
     sale_lines._compute_untaxed_amount_to_invoice()
 
 
+def recompute_sale_order_line_qty_delivered(env):
+    sale_lines = env['sale.order.line'].search([])
+    sale_lines._compute_qty_delivered()
```
pedrobaeza: With, you are possibly changing the old value, provoking that sales orders that were fully invoiced now are not. Possible cases:

- Manual changes in the field in previous versions.
- Modules that modifies the computation of this field.
- etc

And not only that, doing this on post avoids to have the proper computation according extra modules (for example, `sale_stock`).

We need to go over each of the SO lines that doesn't have manual method in end migration, call the compute method, and if the value is different, then put manual method and restore the previous value.

addons/sale/migrations/12.0.1.1/pre-migration.py
```diff
@@ -3,6 +3,12 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
+_column_copies = {
+    'sale_order_line': [
+        ('qty_delivered', 'qty_delivered_manual', None),
```
pedrobaeza: This should be put as 0 for all the non manual methods. If not, there can be side effects if this is the expected value.

pedrobaeza (changes requested):

addons/sale/migrations/12.0.1.1/pre-migration.py
MiquelRForgeFlow: The qty_delivered_manual for non manual methods will be 0 due to `_inverse_qty_delivered` method.

addons/sale/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: I am sure. Check https://github.com/OCA/OpenUpgrade/blob/12.0/addons/sale/models/analytic.py#L22. I tested and works fine.

addons/sale/migrations/12.0.1.1/pre-migration.py
pedrobaeza: Yes, that's my point, so copying the column is not the solution.

addons/sale/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: done

pedrobaeza (approved):
With my latest changes, I consider it completed

MiquelRForgeFlow (commented):
:+1: good!!

PR #1714, opened 2019-02-04
[FIX] renamed project_task_materials

sergiocorato:
Description of the issue/feature this PR addresses: the module project_task_materials was renamed in project_task_material so is not correctly installed

Current behavior before PR: the module project_task_material in not correctly installed

Desired behavior after PR is merged: the module project_task_material is installed

--


pedrobaeza (approved):
Great that more people is diving into OpenUpgrade

PR #1716, opened 2019-02-04
[12.0][MIG] sale_management

MiquelRForgeFlow:
Migration scripts for `sale_management` module.

--


addons/sale_management/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,66 @@
...
+    ('sale.quote.line', 'sale.order.template.line'),
+    ('sale.quote.option', 'sale.order.template.option'),
+    ('sale.quote.template', 'sale.order.template'),
+]
+
+_table_renames = [
+    ('sale_quote_line', 'sale_order_template_line'),
+    ('sale_quote_option', 'sale_order_template_option'),
+    ('sale_quote_template', 'sale_order_template'),
+]
```
rvalyi, line 37: the sale.order.option, sale.quote.template, sale.quote.line and sale.quote.option models are defined in the v11 website_quote module which may not be installed. In v12, the website_quote is renamed sale_quotation_builder.

In any case none of these two modules may be installed and then these renames just fail.

rvalyi (dismissed):

addons/sale_management/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
+---Fields in module 'sale_management'---
+sale_management / digest.digest            / kpi_all_sale_total (boolean)  : NEW 
+sale_management / sale.order               / sale_order_option_ids (one2many): NEW relation: sale.order.option
+sale_management / sale.order.line          / sale_order_option_ids (one2many): NEW relation: sale.order.option
```
pedrobaeza: This is renamed from `option_line_id`: https://github.com/OCA/OpenUpgrade/blob/243fb574fc37d27b4416ae72e8566500aa2b41df/addons/website_quote/models/sale_order.py#L18

addons/sale_management/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
+---Fields in module 'sale_management'---
+sale_management / digest.digest            / kpi_all_sale_total (boolean)  : NEW 
+sale_management / sale.order               / sale_order_option_ids (one2many): NEW relation: sale.order.option
```
pedrobaeza, line 3: This is renamed from `options`: https://github.com/OCA/OpenUpgrade/blob/243fb574fc37d27b4416ae72e8566500aa2b41df/addons/website_quote/models/sale_order.py#L71

addons/sale_management/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+sale_management / sale.quote.template      / display_name (char)           : previously in module website_quote
+sale_management / sale.quote.template      / mail_template_id (many2one)   : previously in module website_quote
+sale_management / sale.quote.template      / name (char)                   : previously in module website_quote
+sale_management / sale.quote.template      / note (text)                   : previously in module website_quote
+sale_management / sale.quote.template      / number_of_days (integer)      : previously in module website_quote
+# NOTHING TO DO: same fields as in website_quote
+
+sale_management / sale.order.template      / sale_order_template_line_ids (one2many): NEW relation: sale.order.template.line
+sale_management / sale.order.template      / sale_order_template_option_ids (one2many): NEW relation: sale.order.template.option
+# NOTHING TO DO: they are one2many
```
pedrobaeza: Group this with the new name for putting the text of the renaming

addons/sale_management/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,116 @@
...
+_field_renames = [
+    ('sale.order', 'sale_order', 'template_id', 'sale_order_template_id'),
+    ('sale.quote.line', 'sale_quote_line', 'quote_id',
+     'sale_order_template_id'),
+    ('sale.quote.option', 'sale_quote_option', 'template_id',
+     'sale_order_template_id'),
+]
+
+_model_renames = [
+    ('sale.quote.line', 'sale.order.template.line'),
```
pedrobaeza, line 28: This is already in apriori.py file:

https://github.com/OCA/OpenUpgrade/blob/fe23ba62aad4cf7636b9f77a85f01cb5e84e45d2/odoo/addons/openupgrade_records/lib/apriori.py#L65-L67

And I got confused about why again. I have seen already that the other thing is only for analysis, so I think we should add a comment in apriori.py saying so.

addons/sale_management/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,116 @@
...
+        "uom_id", "quantity", "id", "display_name", "create_uid",
+        "create_date", "write_uid", "write_date", "__last_update"),
+}
+
+
+def put_in_correct_module(env, white_list):
+    # avoid ir_model_data_module_name_uniq_index error
+    # because the module of this white fields were sale_quotation_builder
+    for model in white_list:
+        openupgrade.logged_query(
```
pedrobaeza: Use https://github.com/OCA/openupgradelib/blob/e18a4ab83a7bbd9bb3fb76b104c88e3d446c40e4/openupgradelib/openupgrade.py#L1978 instead

pedrobaeza (changes requested):
Little changes and a bit of authorship

addons/sale_management/migrations/12.0.1.0/pre-migration.py
MiquelRForgeFlow: nice

MiquelRForgeFlow: but what about https://github.com/OCA/openupgradelib/blob/e18a4ab83a7bbd9bb3fb76b104c88e3d446c40e4/openupgradelib/openupgrade.py#L1995?

pedrobaeza: Yeah, that was one incorrect assumption by my part. Allowing it in https://github.com/OCA/openupgradelib/commit/103238fe7a197669402510d52e6cb72dbab6a342

addons/sale_management/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,112 @@
...
+        "create_date", "write_uid", "write_date", "__last_update"),
+}
+
+
+def put_in_correct_module(cr, white_list):
+    # avoid ir_model_data_module_name_uniq_index error
+    # because the module of this white fields were sale_quotation_builder
+    for model in white_list:
+        openupgrade.update_module_moved_fields(
+            cr, model, _white_list_fields[model],
```
pedrobaeza: Use here `white_list` for coherence

pedrobaeza (changes requested):
Do you think I deserve co-authorship in files?

addons/sale_management/migrations/12.0.1.0/pre-migration.py
MiquelRForgeFlow: done. sorry, I forgot the thing about the authorship

pedrobaeza (approved):

PR #1717, opened 2019-02-04
[12.0][MIG] point_of_sale

MiquelRForgeFlow:
Migration scripts for `point_of_sale` module.

--


pedrobaeza:
Added a commit for:

- Don't rename main PoS, as people is used to old name (or has already renamed it).
- Try to update UoM of product through code for not failin in other cases.
- Use batch update of pos.order for performance. The method is there for migration purpose.
- XML-ID renaming of `product_uom_categ_unit` already done in base module.

Issue #1665, opened 2018-12-14
[12.0] file base/data/res.lang.csv could not be processed


rvalyi:
Impacted versions: 12.0

Steps to reproduce:
Try to migrate a production DB with the base module installed

Current behavior:
Halt with:
```
2018-12-14 16:20:28,333 68 INFO openupgrade-to-12 odoo.models: Actual recompute of field res.currency.decimal_places for 176 recs.                                                                          
2018-12-14 16:20:28,764 68 INFO openupgrade-to-12 odoo.modules.loading: loading base/data/res.lang.csv                                                                                                       
2018-12-14 16:20:30,339 68 WARNING openupgrade-to-12 odoo.modules.loading: Transient module states were reset                                                                               
2018-12-14 16:20:30,340 68 ERROR openupgrade-to-12 odoo.modules.registry: Failed to load registry                                                                                    
Traceback (most recent call last):                                                                                                                                                                           
  File "/odoo/src/odoo/modules/registry.py", line 86, in new                                                                                                                                                 
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)                                                                                                     
  File "/odoo/src/odoo/modules/loading.py", line 414, in load_modules                                                                                                                           
    report=report, models_to_check=models_to_check, upg_registry=upg_registry)                                                                                                                               
  File "/odoo/src/odoo/modules/loading.py", line 251, in load_module_graph                                                                                                                                   
    load_data(cr, idref, mode, kind='data', package=package, report=report)                                                                                                               
  File "/odoo/src/odoo/modules/loading.py", line 72, in load_data                                                                                                                                            
    tools.convert_file(cr, package.name, filename, idref, mode, noupdate, kind, report)                                                                                                                      
  File "/odoo/src/odoo/tools/convert.py", line 802, in convert_file                                                                                                                                          
    convert_csv_import(cr, module, pathname, fp.read(), idref, mode, noupdate)                                                                                                                              
  File "/odoo/src/odoo/tools/convert.py", line 849, in convert_csv_import                                                                                                               
    raise Exception(_('Module loading %s failed: file %s could not be processed:\n %s') % (module, fname, warning_msg))                                                                                      
Exception: Module loading base failed: file base/data/res.lang.csv could not be processed:                                                                                                                   
 Erreur inconnue pendant l'importation : <class 'odoo.exceptions.UserError'>: ('Le code langue ne peut pas être modifié.', '')                                                                               
[...]

PR #1719, opened 2019-02-04
[MRG] merge upstream 2019-02-05

hbrunn:
I propose this because in v10 the code for the Khmer language was changed:
https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/base/res/res.lang.csv#L38
vs
https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res.lang.csv#L38

So if we run a succession of v10, v11, v12 upgrades, v12 will break because it tries to write the code field and breaks in https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_lang.py#L217


pedrobaeza:
@hbrunn The code change was reverted as it broke also Odoo itself. It was merged in other version (not sure if v11 or v12), but seeing the links, here it's not the change. I would bet you have to put it on pre-migration of the base module for v12.

hbrunn:
ah, I looked at the wrong places, it's `km_KM` there still, but v12 wants to write `km_KH`. Do we add an SQL statement in v12 base's pre migration for this, or how are we going to fix inter version issues?

pedrobaeza:
> Do we add an SQL statement in v12 base's pre migration for this, or how are we going to fix inter version issues?

Yes, a direct SQL update in base pre-migration is the best option for me, and we fix #1665 in that moment.

hbrunn:
ah, didn't see this issue. Will come with a patch soon I think. The problem doesn't arise in the test db as 11 just doesn't have the language at all

MiquelRForgeFlow:
hahaha And why won't you merge this PR anyway? there is a new module added (l10n_be_intrastat_2019), so eventually we will need to update upstream 😏

hbrunn:
oh, if this is helpful anyways, I'll reopen. It's an unchanged merge of odoo/odoo

MiquelRForgeFlow (approved):

pedrobaeza:
The problem here is that we have lost the SHA1 of the last upstream commit for tracking this. May I ask you to do the merge through commit and put it on the merge commit?

hbrunn:
you mean as in rebase?

pedrobaeza:
No, this is something learnt from @StefanRijnhart: instead of doing simply a pull over origin, you make a pull with a pin commit for having that commit reference. He can give more details.

hbrunn:
then I wait here to learn the same thing

StefanRijnhart:
In the PR, you need to mention the commit from upstream that you merged up to (instead of the date). That way, a reviewer can perform the same merge locally and check that the updated code is the same (with a git diff on your branch), and/or that any conflicts are solved in a sane manner.

hbrunn:
so I rewrite this to the commit hash and then we're happy?

StefanRijnhart:
Yes, either you recover the upstream commit that you merged or you start all over again keeping track of that commit, and mention it in the PR description either way.

MiquelRForgeFlow:
Closed by https://github.com/OCA/OpenUpgrade/pull/1754.

PR #1720, opened 2019-02-05
[MIG] website_event_track: Nothing to do

cubells:
cc @Tecnativa


addons/website_event_track/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,51 @@
...
+# Nothing to do
+
+website_event_track / event.track              / speaker_ids (many2many)       : DEL relation: res.partner
+# Nothing to do
+
+website_event_track / event.track              / stage_id (many2one)           : NEW relation: event.track.stage, required: required, req_default: function
+# Nothing to do
+
+website_event_track / event.track              / state (selection)             : DEL required: required, selection_keys: ['announced', 'cancel', 'confirmed', 'draft', 'published', 'refused'], req_default: function
+# Nothing to do
```
MiquelRForgeFlow, line 13: There exists a map between the `state` and the `stage_id`. Each event track that had a state should have now the corresponding stage_id. There exists a stage_id for each state.

addons/website_event_track/migrations/11.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,51 @@
+---Fields in module 'website_event_track'---
+website_event_track / event.track              / activity_ids (one2many)       : NEW relation: mail.activity
+website_event_track / event.track              / kanban_state (selection)      : NEW required: required, selection_keys: ['blocked', 'done', 'normal'], req_default: function
+# Nothing to do
+
+website_event_track / event.track              / speaker_ids (many2many)       : DEL relation: res.partner
+# Nothing to do
```
MiquelRForgeFlow, line 7: if the event track record doesn't have a partner_id, then you should fill it with one of the speaker_ids

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/1874.

pedrobaeza:
Superseeded by #1874

PR #1722, opened 2019-02-06
[12.0][MIG] hr_holidays

MiquelRForgeFlow:
Migration scripts for `hr_holidays` module.

--


DaoTheKien:
I faced this problem when running the script . Can anybody help?
image

MiquelRForgeFlow:
@DaoTheKien thanks, fixed.

addons/hr_holidays/migrations/12.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,41 @@
...
+        ('limit', None),
+    ],
+    'resource_calendar_leaves': [
+        ('holiday_id', None),
+    ]
+ }
+
+_field_renames = [
+    ('hr.holidays', 'hr_holidays', 'number_of_days',
+     openupgrade.get_legacy_name('number_of_days')),
```
pedrobaeza: Rename the column, but not the field

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -0,0 +1,162 @@
...
+            AsIs(openupgrade.get_legacy_name('holidays_id')),
+        ),
+    )
+    # update holiday_id of resource.calendar.leaves:
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE resource_calendar_leaves rcl
+        SET holiday_id = hl.id
+        FROM hr_leave hl
+        WHERE rcl.%s IS NOT NULL AND rcl.%s = hl.%s
```
pedrobaeza: Don't need the `IS NOT NULL` part

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -0,0 +1,162 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+from psycopg2.extensions import AsIs
+
+
+def fill_hr_holidays_dates(cr):
+    openupgrade.logged_query(
+        cr, """
+        UPDATE hr_holidays
+        SET date_from = create_date
```
pedrobaeza: I would put `COALESCE(date_to, create_date)`

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -0,0 +1,162 @@
...
+        WHERE date_from IS NULL
+        """
+    )
+    openupgrade.logged_query(
+        cr, """
+        UPDATE hr_holidays
+        SET date_to = create_date
```
pedrobaeza: I would put `COALESCE(date_from, create_date)`

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,242 @@
...
+*** new hr.leave model ***
+# DONE: post-migration: filled from hr.holidays  (when old type is 'remove')
+
+hr_holidays  / hr.holidays              / date_from (datetime)          : now required, default = function
+hr_holidays  / hr.holidays              / date_to (datetime)            : now required, default = function
+DONE: post-migration: filled if empty
+
+hr_holidays  / hr.holidays              / holiday_status_id (many2one)  : relation is now 'hr.leave.type' ('hr.holidays.status')
+hr_holidays  / hr.holidays              / holiday_type (selection)      : selection_keys is now '['category', 'company', 'department', 'employee']' ('['category', 'employee']')
+hr_holidays  / hr.holidays              / linked_request_ids (one2many) : relation is now 'hr.leave' ('hr.holidays')
```
pedrobaeza, line 14: Should this be reported if we have the renaming clear (that's why the model name appears `hr.holidays` and not `hr.leave`)?

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,242 @@
...
+hr_holidays  / hr.holidays              / parent_id (many2one)          : relation is now 'hr.leave' ('hr.holidays')
+# NOTHING TO DO
+
+hr_holidays  / hr.leave                 / activity_ids (one2many)       : NEW relation: mail.activity
+hr_holidays  / hr.holidays              / message_last_post (datetime)  : DEL
+hr_holidays  / hr.leave                 / message_main_attachment_id (many2one): NEW relation: ir.attachment
+# NOTHING TO DO: fields inherited from mail mixins
+
+hr_holidays  / hr.leave                 / mode_company_id (many2one)    : NEW relation: res.company
+# NOTHING TO DO: used for new 'company' mode in holiday_type
```
pedrobaeza, line 24: Thanks for summarizing the why is nothing to do :clap: :clap:

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,242 @@
...
+
+hr_holidays  / hr.leave                 / request_date_from (date)      : NEW
```
pedrobaeza, line 26: You should fill this one and `request_date_to` at least with the date from `date_from` and `date_to`

addons/hr_holidays/migrations/12.0.1.5/pre-migration.py
pedrobaeza: And then rename the field `number_of_days_display`

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -0,0 +1,162 @@
...
+    # In pre-migration the hr_leave_allocation table still doesn't exist
+    env.cr.execute(
+        """
+        ALTER TABLE hr_leave_allocation
+        ADD COLUMN {} integer;
+        """.format(openupgrade.get_legacy_name('holidays_id'))
+    )
+    openupgrade.logged_query(
+        env.cr, """
+        INSERT INTO hr_leave_allocation (category_id, date_from, date_to,
```
pedrobaeza: You can reuse the same SQL than for `hr_leave` passing table and condition by argument, so you reduce a lot the amount of code.

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,242 @@
...
+hr_holidays  / hr.leave.allocation      / interval_unit (selection)     : NEW selection_keys: ['months', 'weeks', 'years']
+hr_holidays  / hr.leave.allocation      / number_per_interval (float)   : NEW
+hr_holidays  / hr.leave.allocation      / nextcall (date)               : NEW
+hr_holidays  / hr.leave.allocation      / unit_per_interval (selection) : NEW selection_keys: ['days', 'hours']
+# NOTHING TO DO: new features
+
+*** new hr.leave.type model ***
+# DONE: pre-migration: renamed from hr.holidays.status
+
+hr_holidays  / hr.leave.type            / request_unit (selection)      : NEW required: required, selection_keys: ['day', 'hour'], req_default: function
```
pedrobaeza, line 85: You can put that default value = 'day' is correct

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,242 @@
...
+hr_holidays  / hr.leave.type            / sequence (integer)            : NEW
+hr_holidays  / hr.leave.type            / time_type (selection)         : NEW selection_keys: ['leave', 'other']
```
pedrobaeza, line 88: The same about default value

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,242 @@
...
+NEW mail.message.subtype: hr_holidays.mt_leave_refused (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_department_holidays_approved (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_department_holidays_confirmed (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_department_holidays_first_validated (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_department_holidays_refused (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_holidays_approved (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_holidays_confirmed (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_holidays_first_validated (noupdate)
+DEL mail.message.subtype: hr_holidays.mt_holidays_refused (noupdate)
+# DONE: it's preferable to not do the rename and delete the old ones instead
```
pedrobaeza: Why not? You are then mangling all the followers preferences, and the messages with that subtypes becoming dangling.

pedrobaeza (changes requested):

addons/hr_holidays/migrations/12.0.1.5/pre-migration.py
MiquelRForgeFlow: I don't understand. I rename column number_of_days to g_l_n(number_of_days). but then the value in number_of_days_temp should go to number_of_days. What field rename? 🤔

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
MiquelRForgeFlow: I am not sure how to do that in a clean way

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
MiquelRForgeFlow: The problem is that it's a split, not a rename. `mt_holidays_approved` becomes both `mt_leave_approved` and `mt_leave_allocation_approved`. So should be rnamed to the first or to the second? Doesn't make sense.

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
MiquelRForgeFlow: What do you mean? The `one2many` always are `#NothingToDo`

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
MiquelRForgeFlow: but they are not required :S

pedrobaeza: Of course not at DB level and you won't find any error on migration, but it makes an UI inconsistency when cancelling a leave request and not being able due to required by view fields are not present.

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
pedrobaeza: I'm talking about `openupgrade_records` module.

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
pedrobaeza: OK, I'll do it

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
MiquelRForgeFlow:
```yml
ERROR: argument of AND must be type boolean, not type integer
```

pedrobaeza: This shouldn't happen, starting that there shouldn't be any `AND` clause.

addons/hr_holidays/migrations/12.0.1.5/openupgrade_analysis_work.txt
MiquelRForgeFlow: Ah. Despite the model renaming, the `openupgrade_records` always report the model change in the x2x fields.

pedrobaeza: Yes, and this is what I'm proposing to improve for cleaning this file with not needed things to check.

MiquelRForgeFlow: Ok, but I prefer having that in the report

pedrobaeza: Why? This is going to be always "Nothing to do". Remember that the goal of this file is to highlight things that may have something to do.

astirpe:
This is also needed:

```sql
UPDATE mail_message
SET model = 'hr.leave'
WHERE model like 'hr.holidays'
```

MiquelRForgeFlow:
@astirpe thanks, I applied mail_message fix.

astirpe:
@mreficent thanks! It fixed the `mail_message`.

I'm looking at the "FOREIGN KEY" constraints created by this script, regarding field `holiday_id`  in `resource_calendar_leaves`. It seems there are two constraints created, instead of one.

You can see those constraints by executing:

```sql
SELECT *
FROM information_schema.table_constraints
WHERE constraint_name like 'resource_calendar_leaves_holiday_id_fkey%';
```

that gives two results.

astirpe (changes requested):
Also table `ir_act_server` should be updated, since it contains `model_id` and `model_name` still pointing to the old `hr.holidays` model.

MiquelRForgeFlow:
But it doesn't exist any record in `ir_act_server` with `model_name` = `hr.holidays`.

astirpe:
In the database I'm trying to migrate they exist.

MiquelRForgeFlow:
Ok, do you know why in the database you are migrating have an ir_act_server pointing to hr.holidays?

astirpe:
Someone created server actions to be notified by email about the changes of status of a leave. Is a common practice for customers.

Also I'm noticing now that old model `hr.holidays` is still present and active in the system:
go to `Technical` -> `Database structure` -> `Models` and search for `model = 'hr.holidays'`

astirpe:
I see that you commented out the following lines:

```python
_model_renames = [
    ('hr.holidays', 'hr.leave'),
    ...
]

_table_renames = [
    ('hr_holidays', 'hr_leave'),
    ...
]
```

Is there a reason why you commented that code?
To me it seems that the easy approach would be to just rename 'hr_holidays' to 'hr_leave' and extract from there only the allocations, like you were already trying to do I guess. It didn't work? Or did you find issues with that approach?

MiquelRForgeFlow:
I found issues with that model rename approach, I just don't remember which ones.
But FYI, with the model_rename approach, you wouldn't solve your specific cases: in the openupgrade `rename_models` method, you will find a nice "# TODO: signal where the model occurs in references to ir_model" and the `ir_act_server` case is one that is not handled yet.

MiquelRForgeFlow:
@astirpe I applied the rename_model approach for you to try and test 😉 Let me know the results :)

The `ir_act_server` case is handled in https://github.com/OCA/openupgradelib/pull/145.

astirpe:
Thanks! I'm going to test it.
With the rename_model approach, the migration for `project_timesheet_holidays` becomes #NothingToDo, so it could be included in this PR.

astirpe:
Very first test: the `number_of_days` are empty in both `hr.leave` and `hr.leave.allocation`.

astirpe:
Sorry, forget my previous comment. I was wrong. :smiley:

pedrobaeza:
Doing the rename of the model `hr.holidays` to `hr.leave` is not totally correct, as it's a model split, so there are chances that the selected target model name changes references to something that is referring to the other possible model. Said that, as this is a very atypical case, we can keep the renaming, and do the following:

1. There's a chance that there are still foreign keys that are not yet updated (as other v11 modules could add `hr.holidays` references), so we have to remove all the FK constraints. There's a method in openupgradelib called `lift_constraints` that I think serves for this.
2. Transfer records that corresponds to assignations to the other table, also keeping the same ID, so that it's not needed to change any reference.
3. Call the special SQL instruction (something like `SELECT setval()...`) for setting the next number for the id column as max(id) + 1.
4. Remove from original table the records that are now transferred to `hr.leave.allocation`.

@astirpe, about manual records created, being in server actions, etc, you should totally run away from that if you want to migrate without pain. If you need customizations, always include them (although being data), on a custom module, so that you can review them migrating the module. If not, it's not OpenUpgrade duty to fix that things, so you will need to add extra SQL by your own, or migration script in your custom modules.

MiquelRForgeFlow:
@pedrobaeza how should apply point 1? and how I keep same id in point 2 (because "INSERT INTO" does not that)? can you show me an example of point 3? the point 4 is already done.

astirpe:
Another thing: the postgres view `hr_holidays_remaining_leaves_user` does not exist anymore in V12.
It is not automatically removed after the execution of the migration scripts.

pedrobaeza:
The approach of renaming models doesn't serve, as there are too many gaps there of XML-IDs, messages, etc with bad model, so I'm resetting that change. Instead, I'm going to do a raw copy including ID of the records, so no reassignment in all the references is needed, setting later the next sequence in following number, and then manually handling the renaming of known objects.

pedrobaeza:
Finish and waiting CI result

pedrobaeza (approved):

pedroslopez:
It seems this doesn't work when there are no ids passed to `_move_model_in_data`. It produces a syntax error because the query turns out ending in `WHERE res_id IN ()`

MiquelRForgeFlow:
@pedroslopez  Fixed in https://github.com/OCA/OpenUpgrade/pull/1846.

PR #1723, opened 2019-02-06
[12.0][MIG] hr_timesheet

MiquelRForgeFlow:
Migration scripts for `hr_timesheet` module.

--


addons/hr_timesheet/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,44 @@
...
+NEW ir.model.access: hr_timesheet.access_account_analytic_account_user
+NEW ir.model.access: hr_timesheet.access_account_analytic_line_project
+NEW ir.model.access: hr_timesheet.access_uom_uom_hr_timesheet
+DEL ir.model.access: hr_timesheet.access_product_product_user
+DEL ir.model.access: hr_timesheet.access_product_template_timesheet_user
+DEL ir.model.access: hr_timesheet.access_product_uom_hr_timesheet
+NEW ir.ui.view: hr_timesheet.account_analytic_account_view_form_inherit
+NEW ir.ui.view: hr_timesheet.assets_backend
+DEL ir.ui.view: hr_timesheet.timesheet_view_calendar
+NEW uom.uom: uom.product_uom_day (noupdate)
```
pedrobaeza: These 2 updates are curious: they come from noupdate records in module `uom`, but here they are defined in XML as noupdate=0. On runbot, I have seen that the field is filled when you install `hr_timesheet`, no matter the original noupdate (I didn't know this). I suppose in this case it's the same, but I would like your confirmation, and this deserves at least a comment like:
```
NOTHING TO DO: Field `timesheet_widget` updated in noupdate records. Handled by standard update process.

File: odoo/openupgrade/doc/source/modules110-120.rst [hunk 2]
```diff
@@ -447,7 +447,7 @@ missing in the new release are marked with |del|.
 +--------------------------------------------+-------------------------------------------------+
 |project                                     | Done                                            |
 +--------------------------------------------+-------------------------------------------------+
-|project_timesheet_holidays                  |                                                 |
+|project_timesheet_holidays                  | Nothing to do                                   |
```
pedrobaeza, line 450: There's an openupgrade_analysis.txt file for this one.

pedrobaeza (changes requested):

odoo/openupgrade/doc/source/modules110-120.rst
MiquelRForgeFlow: Yes. And? It's a nothing to do.

pedrobaeza: You need to create the `openupgrade_analysis_work.txt` with the proper NOTHING TO DO.

MiquelRForgeFlow: but it's already created :S

pedrobaeza: I didn't see it... have you pushed it now? Any way, if so, we can continue when CIs finishes

addons/hr_timesheet/migrations/12.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: Can you confirm this?

odoo/openupgrade/doc/source/modules110-120.rst
MiquelRForgeFlow: click on the 3rd commit and you should see it :S

pedrobaeza: OK, sorry for the noise.

addons/hr_timesheet/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: Confirmed, just by checking a v12 database.

pedrobaeza: OK, thanks

MiquelRForgeFlow: Wait. I checked now an old oca database migrated and the timesheet_widget was not filled. I will fully recheck properly.

MiquelRForgeFlow: As it doesn't work, I put them in the noupdate_changes.xml

MiquelRForgeFlow:
ready

pedrobaeza (approved):

PR #1731, opened 2019-02-07
[12.0][MIG] base_vat_autocomplete -> partner_autocomplete

MiquelRForgeFlow:
The merge is already done in the apriori.

--


odoo/openupgrade/doc/source/modules110-120.rst
```diff
@@ -93,7 +93,7 @@ missing in the new release are marked with |del|.
 +--------------------------------------------+-------------------------------------------------+
 |base_vat                                    |                                                 |
 +--------------------------------------------+-------------------------------------------------+
-| |del| base_vat_autocomplete                |                                                 |
+| |del| base_vat_autocomplete                | Done. Merge in partner_autocomplete             |
```
pedrobaeza: s/Merge/Merged

pedrobaeza (changes requested):
Why merged instead of renamed?

Put the same sentence in the other part (`partner_autocomplete`). There's nothing to do right now.

MiquelRForgeFlow:
Ready :green_apple:

pedrobaeza:
Some comments are still unattended.

MiquelRForgeFlow:
In the `partner_autocomplete` part it already has a "Nothing to do". I won't put "Nothing. Has merged base_vat_autocomplete." Sounds awful.

It's a merge instead of rename, because the dependencies are changing, and the module partner_autocomplete has many new features.

pedrobaeza:
The text should be: `Done. Renamed from base_vat_autocomplete`, and the operation should be a renaming, not a merge, as it's the best fitting operation. It doesn't matter that "conceptually" new module makes more things. The important thing is the performed SQL operations for getting to this, and speaking conceptually about the migration path, this is a renaming, not a merge: you have `base_vat_autocomplete` on v11, and now you have `partner_autocomplete` in v12. A renaming would be if you have A and B modules in v11, and now there is only module B in v12, so you merge A into B.

MiquelRForgeFlow:
All comments attended.

MiquelRForgeFlow:
@pedrobaeza this one also can be merged, right?

pedrobaeza:
I'm on it...

pedrobaeza (approved):

PR #1734, opened 2019-02-08
[12.0][MIG] repair

MiquelRForgeFlow:
Migration scripts for `repair` module.
Depends of https://github.com/OCA/OpenUpgrade/pull/1683.

--


addons/repair/migrations/12.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="mail_template_repair_quotation" model="mail.template">
+    <field name="model_id" ref="repair.model_repair_order"/>
+    <field name="name">Repair Quotation: Send by email</field>
+    <field name="report_template" ref="action_report_repair_order"/>
+    <field name="body_html" type="xml">
```
StefanRijnhart, line 7: If a template body has changed substantially so that we need to reload it from data, it is usually the case that the body's translations need to be removed for (automatic) reloading as well.

MiquelRForgeFlow: done, thanks!

StefanRijnhart (approved):
Thanks!

PR #1738, opened 2019-02-11
[12.0][MIG] hr_contract

MiquelRForgeFlow:
Migration scripts for `hr_contract` module.

--


addons/hr/migrations/12.0.1.1/pre-migration.py
```diff
@@ -3,11 +3,16 @@
...
     ('hr.employee_root', 'hr.employee_admin'),
 ]
 
 
-@openupgrade.migrate(use_env=False)
-def migrate(cr, version):
-    openupgrade.rename_xmlids(cr, xmlid_renames)
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _field_renames)
```
pedrobaeza: You can't do this if you don't have installed `hr_contract` module, so before calling this method, you have to check if the column exists.

addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -22,6 +21,9 @@ hr           / hr.employee              / study_field (char)            : NEW
 hr           / hr.employee              / study_school (char)           : NEW 
 # NOTHING TO DO
 
+hr           / hr.employee              / km_home_work (integer)        : NEW
+DONE: pre-migration: renamed from vehicle_distance
```
pedrobaeza: Mention that the field comes from `hr_contract` module

addons/hr_contract/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,25 @@
+---Fields in module 'hr_contract'---
+hr_contract  / hr.contract              / active (boolean)              : NEW 
+# DONE: post-migration: filled as True by default
```
pedrobaeza: No need. Its default value is already True, filled when adding the field in the regular update process.

addons/hr_contract/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,25 @@
...
+hr_contract  / hr.employee              / place_of_birth (char)         : module is now 'hr' ('hr_contract')
+# NOTHING TO DO: hr is a dependency of hr_contract
+
+hr_contract  / hr.employee              / vehicle_distance (integer)    : DEL
+# NOTHING TO DO: renamed to km_home_work in 'hr' module
+
+---XML records in module 'hr_contract'---
+DEL ir.model.access: hr_contract.access_hr_contract_type_user
+DEL ir.model.access: hr_contract.access_hr_contract_user
+NEW res.groups: hr_contract.group_hr_contract_manager (noupdate)
```
pedrobaeza: As this new group now restricts some operations, we should make equivalent previous v11 permissions with these ones. As I'm seeing, this means that we should add this group to those users that are HR managers (`hr.group_hr_manager`).

pedrobaeza: As a side note, I think this shouldn't be noupdate, but that's an Odoo bug.

addons/hr_contract/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_hr_contract_active(cr):
+    openupgrade.logged_query(
```
pedrobaeza: Not needed as explained.

pedrobaeza (changes requested):
See comments and I request you as usual my copyright slice as part of the investigation in the changes for the review.

MiquelRForgeFlow:
Updated, comments attended.

addons/hr_contract/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+from psycopg2.extensions import AsIs
+
+
+def add_group_hr_contract_manager(env):
+    hr_manager = env.ref('hr.group_hr_manager').id
+    new_group = env.ref('hr_contract.group_hr_contract_manager').id
```
pedrobaeza: This is simpler to be solved:
```python
env.ref('hr_contract.group_hr_contract_manager').users = [
    (4, x.id) for x in env.ref('hr.group_hr_manager').users
]
```

MiquelRForgeFlow: but this way it's not logged. the query way logs the change :)

pedrobaeza: Don't worry about that in that case. If you want some logging, just put `openupgrade.logging()` decorator in the method, but having the exact "query" is not critical here. Or if you want to be verbose, put an explicit logging saying `Adding "HR contract manager" permission to users %s...`.

MiquelRForgeFlow: ok, replaced. bye bye sexy query 🤧

pedrobaeza: Hehe, sexy but complicated, as in real life :yum:

pedrobaeza (approved):

PR #1739, opened 2019-02-11
[12.0][MIG] sale_timesheet

MiquelRForgeFlow:
`#NothingToDo`

--


pedrobaeza (approved):
Checked that all that new fields are filled with proper values. A little more detail on each of them would be desirable, but not blocking this right now.

davidtranhp:
Should you fill data for the following?
1. project's sale_order_id
2. project task's sale_order_id
3. sale order line's project_id

pedrobaeza:
Not needed AFAIK studying the use for that fields (re-expense of the timesheets), which is a new feature.

PR #1740, opened 2019-02-11
[12.0][MIG] hr_expense

MiquelRForgeFlow:
Migration scripts for `hr_expense` module.

--


pedroslopez:
I'm not sure if it would be an issue with this specific migration, but when trying to delete the `planner_hr_expense` view, it fails because it was being referenced by a web.planner object:

```
2019-05-29 13:01:00,886 27241 ERROR eat12_cilm odoo.sql_db: bad query: b'DELETE FROM ir_ui_view WHERE id IN (2521)'
ERROR: null value in column "view_id" violates not-null constraint
DETAIL:  Failing row contains (6, 761, 1, 2019-05-20 18:17:29.016952, Expense strategy planner, planner_hr_expense, null, 1, 2019-05-20 18:29:42.021054, t, null, null, <p>Send receipts by email to create expenses efficiently.
      ...).
CONTEXT:  SQL statement "UPDATE ONLY "public"."web_planner" SET "view_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "view_id""

2019-05-29 13:01:00,886 27241 WARNING eat12_cilm odoo.addons.base.models.ir_model: Could not delete obsolete record with id: 2521 of model ir.ui.view
Please refer to the log message right above

```

However, like I said, I'm not sure if this should be handled here or if it's a problem with the web.planner migration, since this model is gone in v12.

**EDIT** This has been handled in #1888

pedrobaeza (approved):
I have improved a bit the PR, but only an extra thing

PR #1741, opened 2019-02-11
unnecessary auth_brute_force

ernestotejeda:
https://github.com/OCA/server-auth/issues/38#issuecomment-462308921

Cc @Tecnativa

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -10,6 +10,7 @@ renamed_modules = {
 
 merged_modules = {
     # Odoo
+    'auth_brute_force': 'base',
     'auth_crypt': 'base',
     'account_cash_basis_base_account': 'account',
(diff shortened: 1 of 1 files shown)


odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -10,6 +10,7 @@
 
 merged_modules = {
     # Odoo
+    'auth_brute_force': 'base',
```
MiquelRForgeFlow: but `auth_brute_force` is not from `#Odoo`

please, put in a new `#OCA/server-auth` below like `OCA/web` case

ernestotejeda: thanks @mreficent

MiquelRForgeFlow (commented):
Please, could you squash commits?

pedrobaeza:
Not needed. We can squash on merge.

PR #1742, opened 2019-02-11
[12.0][MIG] l10n_generic_coa

MiquelRForgeFlow:
Migration scripts for `l10n_generic_coa` module.

--


addons/l10n_generic_coa/migrations/12.0.1.1/noupdate_changes.xml
```diff
@@ -0,0 +1,13 @@
...
+<odoo>
+  <record id="conf_cas_interim1" model="account.account.template">
+    <field name="reconcile" eval="True"/>
+  </record>
+  <record id="conf_cas_interim2" model="account.account.template">
+    <field name="reconcile" eval="True"/>
+  </record>
+  <record id="configurable_chart_template" model="account.chart.template">
+    <!--<field name="transfer_account_id"/>-->
+    <field name="transfer_account_code_prefix">1017</field>
```
astirpe, line 12: Shouldn't this value be calculated from the `openupgrade_legacy_12_0_transfer_account_id`? It is possible that the old `transfer_account_id` (defined as `noupdate="1"`) has been manually modified.

astirpe (changes requested):

addons/l10n_generic_coa/migrations/12.0.1.1/noupdate_changes.xml
MiquelRForgeFlow: Right. In fact, it was handled in the `fill_account_chart_template_account_code_prefix` method of post-migration of `account`.

astirpe (approved):
I think it's OK now, thanks.

I see that the calculation of the `transfer_account_code_prefix` was done in the account post-migration script.
But shouldn't line https://github.com/OCA/OpenUpgrade/blob/12.0/addons/account/migrations/12.0.1.1/post-migration.py#L99 use `trailing` instead of `leading`?

Shouldn't be this way:

```
SET transfer_account_code_prefix = trim(trailing '0' from aat.code)
```

?

MiquelRForgeFlow:
Nop, I am sure is `leading`.

astirpe:
`...code_prefix` expects to be the left part of the code, not the right part. So you should remove the right part and keep the left one. The `trailing` seems the one doing the job.

astirpe:
Comment updated

MiquelRForgeFlow:
Oh, yeah, my bad. You are right.

astirpe:
No problem. I will open a PR soon.

MiquelRForgeFlow:
The problem I see is that... There won't be a prefix code ending with `0`, which am I not sure if it's wrong or not. Doing a search in Odoo I see a lot if prefix codes ending with `0`.

astirpe:
Right. I'm thinking to set `transfer_account_code_prefix` the same as `transfer_account_id.code`. It shouldn't be a problem. And the user can manually adjust it later.

MiquelRForgeFlow:
@pedrobaeza can we merge this?

pedrobaeza (approved):

Issue #1624, opened 2018-11-19
[11.0] openupgrade_records: Changed fields not recognized by analysis


pedrobaeza:
On v10, we had these social media links in website:

https://github.com/odoo/odoo/blob/258eaa7a3ce7c25e93f1b9940351335bfcda2aeb/addons/website/models/website.py#L170-L176

But on v11, they are non-stored, related fields:

https://github.com/odoo/odoo/blob/cac91731221b5ac122ad745ac408b406d335e5c7/addons/website/models/website.py#L55-L60

I would expect that the analysis prints that the fields have changed their definition (like other times that I have seen something about "now converted to a function" or something similar).

@Tecnativa

pedrobaeza:
cc @StefanRijnhart

pedrobaeza:
Implemented in #1744

PR #1744, opened 2019-02-12
[12.0][FIX] openupgrade_records: several fixes

MiquelRForgeFlow:
This PR:
-  solves https://github.com/OCA/OpenUpgrade/issues/1624
- reports changes like https://github.com/OCA/OpenUpgrade/pull/1674#pullrequestreview-202486667
- and other fixes

The minor fix commits will be squashed before merge.

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -96,6 +96,13 @@ def report_generic(new, old, attrs, reprs):
                 if new['req_default']:
                     text += ', default = %s' % new['req_default']
                 fieldprint(old, new, None, text, reprs)
+        elif attr == 'stored':
+            if old[attr] != new[attr]:
+                if new['stored']:
+                    text = "is now stored"
+                else:
+                    text = "not stored anymore"
+                fieldprint(old, new, None, text, reprs)
         elif attr == 'isfunction':
             if old[attr] != new[attr]:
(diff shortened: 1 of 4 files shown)


StefanRijnhart:
Looking good! Can you provide a branch that contains the analysis with this change so that we can validate the results? You may need to run the analysis against the older version with this change backported, because you changed the output format.

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -95,35 +95,42 @@ def report_generic(new, old, attrs, reprs):
                 text = "now required"
                 if new['req_default']:
                     text += ', default = %s' % new['req_default']
-                fieldprint(old, new, None, text, reprs)
+                fieldprint(old, new, '', text, reprs)
+        elif attr == 'stored':
+            if old[attr] != new[attr]:
+                if new['stored']:
+                    text = "now is stored"
```
StefanRijnhart: Slight preference for "now stored" (or else "is now stored", but not "now is stored").

StefanRijnhart:
Great work!

pedrobaeza (approved):
Indeed congratulations for the improvements. There's still one missing feature that I would love to have, that is to report added and missing models in the module itself, but this should been done counting references and seeing if it's the last/first one, and that's for another PR of course. Merging this one.

PR #1747, opened 2019-02-14
[ADD] migrate expense's attachments and messages

hbrunn:
this is not perfect because we can't know for which expense line an attachment/message is meant. We might duplicate all of them, but that doesn't seem worth bothering to me.
They will end up on whichever line will be the last in the join.

### DIFF
--- a/addons/hr_expense/migrations/9.0.2.0/post-migration.py
+++ b/addons/hr_expense/migrations/9.0.2.0/post-migration.py
@@ -41,6 +41,22 @@ def hr_expense(env):
             )
         },
     )
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_attachment
+        SET res_model='hr.expense', res_id=hr_expense.id
+        FROM hr_expense
+        WHERE res_model='hr.expense.expense' AND hr_expense.expense_id=res_id
+        """
+    )
+    openupgrade.logged_query(
+        env.cr, """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In v10, the sheet is introduced again, so do you think it's worth to keep any reference to old sheet for doing in OpenUpgrade v10 the restoration?

hbrunn:
I did this actually for a 10 database. 10 looks at the lines' attachments: https://github.com/OCA/OCB/blob/10.0/addons/hr_expense/models/hr_expense.py#L576
so with this, we'll get the attachments nicely on (some) sheet generated for the lines too

pedrobaeza:
OK, my fear was that you have:

* 1 sheet with 10 expenses in v8
* 10 expenses in v9
* 10 sheets with 1 expense in v10

But I don't remember if this was handled already.

Merging meanwhile this.

hbrunn:
it's not, that's exactly how it goes, so all the attachments en up on one of the 10 lines and by this one of the 10 sheets. I am considering to write the code that detects for 10 if we're coming from 8 and then restore the sheets like that, but up to now my customers didn't request that.

pedrobaeza:
OK, I haven't got any customer coming from v8 with expenses (it remains very few right now - only new bounced customers with still v8 with the pending migration-), so no problem by my part either right now (maybe in the future I get one).

PR #1745, opened 2019-02-13
[12.0][UPD] new openupgrade analysis

MiquelRForgeFlow:
New analysis using https://github.com/OCA/OpenUpgrade/pull/1744.

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #1750, opened 2019-02-15
[12.0][UPD] updated docs of already migrated modules

MiquelRForgeFlow:
PR https://github.com/OCA/OpenUpgrade/pull/1745 added new lines is some reports, thus I checked the "work" docs of already migrated modules, trying to update the relevant data and do more scripts if needed.

--

### DIFF
--- a/addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
@@ -24,6 +24,10 @@ hr           / hr.employee              / study_school (char)           : NEW
 hr           / hr.employee              / km_home_work (integer)        : NEW
 DONE: pre-migration: renamed from vehicle_distance that comes from hr_contract module
 
+hr           / hr.employee              / resource_calendar_id (many2one): is now stored
+hr           / hr.employee              / user_id (many2one)            : is now stored
+# DONE: post-migration: filled these fields
+
 hr           / hr.job                   / message_last_post (datetime)  : DEL
 hr           / hr.job                   / message_main_attachment_id (many2one): NEW relation: ir.attachment
 hr           / mail.channel             / subscription_department_ids (many2many): NEW relation: hr.department
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -42,7 +42,7 @@ renamed_models = {
     'product.attribute.line': 'product.template.attribute.line',
     'product.attribute.price': 'product.template.attribute.value',
     'product.uom': 'uom.uom',
-    'product.uom.category': 'uom.category',
+    'product.uom.categ': 'uom.category',
     'sale.quote.line': 'sale.order.template.line',
     'sale.quote.option': 'sale.order.template.option',
     'sale.quote.template': 'sale.order.template',
--- a/addons/analytic/migrations/12.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/analytic/migrations/12.0.1.1/openupgrade_analysis_work.txt
@@ -1,5 +1,6 @@
 ---Fields in module 'analytic'---
 analytic     / account.analytic.account / group_id (many2one)           : NEW relation: account.analytic.group
+analytic     / account.analytic.line    / group_id (many2one)           : NEW relation: account.analytic.group
 # NOTHING TO DO (relation to new model)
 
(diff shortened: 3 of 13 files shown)


addons/crm/migrations/12.0.1.0/post-migration.py
```diff
@@ -29,10 +29,16 @@ def map_crm_team_dashboard_graph_model(cr):
         table='crm_team', write='sql')
 
 
+def fill_crm_lead_expected_revenue(env):
```
pedrobaeza: No need of computing it again. As the SQL column doesn't exist, regular update process creates the column, and compute the value for it.

addons/hr_attendance/migrations/12.0.2.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_hr_employee_last_attendance_id(env):
```
pedrobaeza: The same logic for this: as the SQL column doesn't exist, regular update process creates it and compute the value.

addons/hr/migrations/12.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -24,6 +24,10 @@ hr           / hr.employee              / study_school (char)           : NEW
 hr           / hr.employee              / km_home_work (integer)        : NEW
 DONE: pre-migration: renamed from vehicle_distance that comes from hr_contract module
 
+hr           / hr.employee              / resource_calendar_id (many2one): is now stored
+hr           / hr.employee              / user_id (many2one)            : is now stored
+# DONE: post-migration: filled this fields
```
pedrobaeza: s/this/these

addons/resource/migrations/12.0.1.1/post-migration.py
```diff
@@ -47,13 +47,24 @@ def fill_resource_calendar_hours_per_day(env):
         calendar._onchange_hours_per_day()
 
 
+def fill_resource_test_resource_calendar_id(cr):
```
pedrobaeza: Watch out: This a test only model! It shouldn't exist in regular databases

odoo/addons/base/migrations/12.0.1.3/post-migration.py
```diff
@@ -32,11 +32,17 @@ def update_res_company_onboarding_company_state(env):
     good_companies.write({'base_onboarding_company_state': 'done'})
 
 
+def fill_res_currency_deciaml_places(env):
+    currencies = env['res.currency'].with_context(active_test=False).search([])
+    currencies._compute_decimal_places()
```
pedrobaeza: No need of this

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -41,7 +41,7 @@
     'product.attribute.line': 'product.template.attribute.line',
     'product.attribute.price': 'product.template.attribute.value',
     'product.uom': 'uom.uom',
-    'product.uom.category': 'uom.category',
+    'product.uom.categ': 'uom.category',
```
pedrobaeza, line 45: Isn't this already changed in other PR?

pedrobaeza (changes requested):

odoo/addons/openupgrade_records/lib/apriori.py
MiquelRForgeFlow: It was changed in base pre-migration, but not here.

addons/resource/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: I don't know why, but the table of this `ResourceTest(models.Model)` model is created :S

pedrobaeza: Yes, on demo databases, but what happen on real ones? Any way, you shouldn't care about it although created.

MiquelRForgeFlow: Ok, but it's rare that the openupgrade analysis detected this, because I think it doesn't use demo data.

MiquelRForgeFlow:
all comments attended :)

pedrobaeza (approved):

PR #1762, opened 2019-03-01
[12.0][IMP] openupgrade_records: handle moved/renamed xmlids

MiquelRForgeFlow:
This PR includes 3 things:

 - Detect moved or renamed xmlids in different modules
 - Remove the sickening trailing spaces in the reports
 - New analysis to observe how affect his changes

The `openupgrade_analysis.txt` of `account` module gives a good example of the changes introduced.

--

### DIFF
--- a/addons/project/migrations/12.0.1.1/openupgrade_analysis.txt
+++ b/addons/project/migrations/12.0.1.1/openupgrade_analysis.txt
@@ -3,22 +3,22 @@ project      / account.analytic.account / company_uom_id (many2one)     : module
 project      / account.analytic.account / company_uom_id (many2one)     : relation is now 'uom.uom' ('product.uom')
 project      / account.analytic.account / project_count (integer)       : module is now 'hr_timesheet' ('project')
 project      / account.analytic.account / project_ids (one2many)        : module is now 'hr_timesheet' ('project')
-project      / digest.digest            / kpi_project_task_opened (boolean): NEW 
-project      / project.project          / access_token (char)           : NEW 
+project      / digest.digest            / kpi_project_task_opened (boolean): NEW
+project      / project.project          / access_token (char)           : NEW
 project      / project.project          / analytic_account_id (many2one): module is now 'hr_timesheet' ('project')
-project      / project.project          / company_id (many2one)         : not related anymore
 project      / project.project          / company_id (many2one)         : is now stored
-project      / project.project          / message_last_post (datetime)  : DEL 
+project      / project.project          / company_id (many2one)         : not related anymore
+project      / project.project          / message_last_post (datetime)  : DEL
 project      / project.project          / message_main_attachment_id (many2one): NEW relation: ir.attachment
-project      / project.project          / name (char)                   : not related anymore
 project      / project.project          / name (char)                   : is now stored
-project      / project.project          / partner_id (many2one)         : not related anymore
+project      / project.project          / name (char)                   : not related anymore
 project      / project.project          / partner_id (many2one)         : is now stored
-project      / project.task             / access_token (char)           : NEW 
+project      / project.project          / partner_id (many2one)         : not related anymore
+project      / project.task             / access_token (char)           : NEW
 project      / project.task             / activity_date_deadline (date) : not related anymore
 project      / project.task             / activity_date_deadline (date) : not stored anymore
 project      / project.task             / activity_date_deadline (date) : now a function
-project      / project.task             / message_last_post (datetime)  : DEL 
+project      / project.task             / message_last_post (datetime)  : DEL
 project      / project.task             / message_main_attachment_id (many2one): NEW relation: ir.attachment
 project      / project.task             / remaining_hours (float)       : module is now 'hr_timesheet' ('project')
 project      / project.task             / remaining_hours (float)       : now a function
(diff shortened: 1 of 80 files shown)


addons/project/migrations/12.0.1.1/openupgrade_analysis.txt
```diff
@@ -89,7 +89,7 @@ NEW mail.template: project.rating_project_request_email_template (noupdate)
 DEL mail.template: project.mail_template_data_module_install_project (noupdate)
 DEL mail.template: rating_project.rating_project_request_email_template
 DEL project.task: project.project_task_data_8 (noupdate)
-DEL res.company: base.main_company (noupdate)
+DEL res.company: base.main_company [moved to web module] (noupdate)
```
pedrobaeza: This is still strange that appears as this is not true. What about the approach I told about that ignoring records from other modules that are re-declared (for changing any field) in other modules?

odoo/addons/openupgrade_records/models/openupgrade_record.py
```diff
@@ -38,6 +38,20 @@ class Record(models.Model):
             'openupgrade.attribute', 'record_id',
             readonly=True)
     noupdate = fields.Boolean(readonly=True)
+    prefix = fields.Char(readonly=True, compute='_compute_prefix_and_suffix')
```
pedrobaeza: No need of readonly if they are computed

pedrobaeza:
The risk with this approach is that you have the same suffix in several modules but don't belong to the same record. Are you comparing the model of the record? Even with that, I would put the word "Potentially moved..." for not assuring something that we don't know.

addons/project/migrations/12.0.1.1/openupgrade_analysis.txt
MiquelRForgeFlow: The ignoring is done for the `rename` check. If I do for the `move` check, then the moved groups from `sale` to `account` cannot be detected.

MiquelRForgeFlow:
Ready

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -264,28 +268,68 @@ def match(match_fields, report_fields, warn=False):
...
         key=lambda k: (k['model'], 'old' in k, k['name'])
     )
     for entry in sorted_records:
         content = ''
         if 'old' in entry:
             content = 'DEL %(model)s: %(name)s' % entry
+            if 'moved' in entry:
+                content += ' [potentially moved to %(moved)s module]' % entry
+            elif 'renamed' in entry:
+                content += ' [renamed to %(renamed)s module]' % entry
```
pedrobaeza, line 332: The word `potentially` should be also on this side

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -264,28 +268,68 @@ def match(match_fields, report_fields, warn=False):
...
         elif 'new' in entry:
             content = 'NEW %(model)s: %(name)s' % entry
+            if 'moved' in entry:
+                content += ' [potentially moved from %(moved)s module]' % entry
+            elif 'renamed' in entry:
+                content += ' [renamed from %(renamed)s module]' % entry
```
pedrobaeza, line 338: The word `potentially` should be also on this side

pedrobaeza (commented):
What is the difference between `renamed` and `moved`?

odoo/addons/openupgrade_records/lib/compare.py
MiquelRForgeFlow: the rename is when the prefix (module name) changed. It's difficult to have a wrong rename :S

pedrobaeza: I don't agree. You can have the same suffix for different things.

mreficent: also when it's a rename you should use rename_xmlid method. This is not the case of moved xmlids

pedrobaeza: We are mixing things here. Please answer first the main question outside this comment about the difference between renamed and moved.

MiquelRForgeFlow: but it would be super improbable that in a module a xmlid is removed and then in another module (same model) the same xmlid appears but is used for a different thing

MiquelRForgeFlow:
`renamed`: it's when a xmlid is in another module and the prefix changed. Example:
`stock.incoterms: stock.incoterm_CFR -> account.incoterms: account.incoterm_CFR`

`moved`: it's when a xmlid is in another module but has the same full xmlid: Example:
`res.groups: base.group_something (sale module) -> res.groups: base.group_something` (account module)

odoo/addons/openupgrade_records/lib/compare.py
MiquelRForgeFlow: well. I am lazy and don't want to discuss this. If you are happy with the "potentially renamed", then I will add it. No problem.

MiquelRForgeFlow:
Ready

odoo/addons/openupgrade_records/lib/compare.py
pedrobaeza: Well, it was more a question of understanding the difference of both types. Now I understand, and as you have said, it's very improbable that, having that circumstances, isn't a rename, so you can leave it as before, or change the analysis file according the new text. What you prefer.

MiquelRForgeFlow: ok, changed again haha

PR #1769, opened 2019-03-07
[12.0][FIX] project: some fixes

MiquelRForgeFlow:
- compute `rating_last_value` because you may have a lot of tasks and you may not have `rating_project` module installed because it's not auto_install
- no upload noupdate_changes file if not task records

--

### DIFF
--- a/addons/project/migrations/12.0.1.1/post-migration.py
+++ b/addons/project/migrations/12.0.1.1/post-migration.py
@@ -19,8 +19,14 @@ def fill_project_project_inherits_values(cr):
 @openupgrade.migrate()
 def migrate(env, version):
     fill_project_project_inherits_values(env.cr)
-    openupgrade.load_data(
-        env.cr, 'project', 'migrations/12.0.1.1/noupdate_changes.xml')
+    # openupgrade.load_data(
+    #     env.cr, 'project', 'migrations/12.0.1.1/noupdate_changes.xml',
+    #     mode='init_no_create')
+    # TODO: change again when the 'init_no_create' works correctly
+    task_14 = env.ref('project.project_task_data_14', False)
+    if task_14:
+        openupgrade.load_data(
+            env.cr, 'project', 'migrations/12.0.1.1/noupdate_changes.xml')
     openupgrade.delete_records_safely_by_xml_id(
         env, [
             'project.msg_task_data_14_attach',
--- a/addons/project/migrations/12.0.1.1/pre-migration.py
+++ b/addons/project/migrations/12.0.1.1/pre-migration.py
@@ -9,6 +9,26 @@ column_copies = {
 }
 
 
+def compute_project_task_rating_last_value(env):
+    # The computing process for this value is slow if you have many tasks.
+    # Thus, we compute rating_last_value using SQL because is faster.
+    if not openupgrade.column_exists(
+            env.cr, 'project_task', 'rating_last_value'):
+        openupgrade.logged_query(
+            env.cr, """
+            ALTER TABLE project_task
+            ADD COLUMN rating_last_value DOUBLE PRECISION""",
+        )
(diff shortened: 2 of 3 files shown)


mgielissen:
Fixes the problem:

2019-03-14 21:49:13,340 23483 ERROR copy2_migrated OpenUpgrade: "null value in column "name" violates not-null constraint
DETAIL:  Failing row contains (81, 2019-03-14 21:49:08.353245, 10, null, null, 1, null, null, null, 1, 1, 2019-03-14 21:49:13, null, 1, 0
, null, null, 2019-03-14 21:49:13, null, normal, 2019-03-14 21:49:08.353245, t, null, null, null, null, null, null, null, 2019-03-14 21:4
9:13, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null).
" while parsing /var/tmp/openupgrade/12.0/addons/project/migrations/12.0.1.1/noupdate_changes.xml:3, near
<record id="project_task_data_14" model="project.task" forcecreate="False">
    <field name="user_id" ref="base.user_admin"/>
    <field name="displayed_image_id"/>

addons/project/migrations/12.0.1.1/post-migration.py
```diff
@@ -19,8 +19,10 @@ def fill_project_project_inherits_values(cr):
 @openupgrade.migrate()
 def migrate(env, version):
     fill_project_project_inherits_values(env.cr)
-    openupgrade.load_data(
-        env.cr, 'project', 'migrations/12.0.1.1/noupdate_changes.xml')
+    task_14 = env.ref('project.project_task_data_14', False)
+    if task_14:
+        openupgrade.load_data(
```
pedrobaeza, line 28: This is not the way to do this, but use this parameter instead:

https://github.com/OCA/openupgradelib/blob/b4214ff10ba1ca1fa3f0fd79f049abbba1874d68/openupgradelib/openupgrade.py#L272

that we explicitly added in openupgradelib for these cases

addons/project/migrations/12.0.1.1/pre-migration.py
```diff
@@ -9,6 +9,23 @@
...
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.copy_columns(env.cr, column_copies)
+    if not openupgrade.column_exists(
+            env.cr, 'project_task', 'rating_last_value'):
+        openupgrade.logged_query(
+            env.cr, """
+            ALTER TABLE project_task ADD COLUMN rating_last_value INTEGER""",
```
pedrobaeza: The field is a float one... And why do you need to compute it again? Please always put in the comment in the analysis work for avoiding later questions.

pedrobaeza (changes requested):

MiquelRForgeFlow:
Comments attended.

pedrobaeza:
It seems there's a problem in openupgradelib method (maybe not adapted to v12?). It's also interesting to have all the code inside the method instead of doing the `if` on the main method.

mreficent:
Maybe this error is fixed with https://github.com/OCA/openupgradelib/pull/141.

MiquelRForgeFlow:
Should I revert the 'init_no_create' change to be able to merge this? Then, if in a future the 'init_no_create' works again, then I can put it again doing another PR.

pedrobaeza:
Well, the other code is also not very resilient, as you only compare one of the records if I don't remember bad...

MiquelRForgeFlow:
Yeah, but it's temporarily better than having here this PR open and broken.

pedrobaeza:
OK, let's revert it, but please put the corresponding comment in the code as TODO

MiquelRForgeFlow:
Reverted!

pedrobaeza (approved):

Issue #1771, opened 2019-03-07
[12.0] password_crypt field migration


paulius-sladkevicius:
After migration to v12, the `password` field for `res.users` was different from old field `password_crypt`, so I suppose there is need migration script. What you think?

pedrobaeza:
Yes, it seems so. We should copy `password_crypt` in `password`

MiquelRForgeFlow:
It's not than easy as a copy. We should encrypt the v11 passwords (in case they don't have password_crypt)

pedrobaeza:
Not really. `auth_crypt` is auto-installed. We needed that in previous version, and it was done.

PR #1773, opened 2019-03-07
[12.0][FIX] base (auth_crypt): pass password_crypt

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/issues/1771. (?)

--

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/end-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/end-migration.py
@@ -97,7 +97,19 @@ def fork_off_system_user(env):
         (partner_root.id,))
 
 
+def fill_res_users_password_from_password_crypt(cr):
+    openupgrade.logged_query(
+        cr,
+        """UPDATE res_users
+        SET password = password_crypt
+        WHERE password_crypt IS NOT NULL
+        """,
+    )
+
+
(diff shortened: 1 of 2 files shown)


odoo/addons/base/migrations/12.0.1.3/end-migration.py
```diff
@@ -97,7 +97,19 @@ def fork_off_system_user(env):
...
+        WHERE password_crypt IS NOT NULL
+        """,
+    )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     update_model_terms_translations(env)
     fork_off_system_user(env)
+    if not openupgrade.column_exists(env.cr, 'res_users', 'password_crypt'):
```
pedrobaeza: This should be the reverse condition

pedrobaeza (changes requested):

odoo/addons/base/migrations/12.0.1.3/end-migration.py
MiquelRForgeFlow: right

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #1778, opened 2019-03-12
[12.0][UPD] product_extended: merged module

cubells:
cc @Tecnativa


pedrobaeza:
I'm seeing that this is already handled in https://github.com/OCA/OpenUpgrade/pull/1778/files#diff-6f879554b0311202a31854663851227bR8, as a module renaming, which is the accurate thing.

pedrobaeza:
Documentation updated in https://github.com/OCA/OpenUpgrade/commit/5a05dd0bd32d292bdbf125a7d1586b43dc2aa038

PR #1780, opened 2019-03-18
[12.0][FIX][DOC] new modules

MiquelRForgeFlow:
New modules. They were added in last months.

--


pedrobaeza (approved):
They are not very important, but OK for being complete

PR #1785, opened 2019-03-27
[9.0] crm: assign activity to leads that has action

paulius-sladkevicius:
If it's not assign, issue comes in the v11 activity migration to `mail_activity` if there is no `next_activity_id` it skips such records https://github.com/OCA/OpenUpgrade/blob/11.0/addons/crm/migrations/11.0.1.0/post-migration.py#L78

--


pedrobaeza:
Please check Travis status

paulius-sladkevicius:
@pedrobaeza fixed

pedrobaeza (approved):

PR #1793, opened 2019-04-10
[12.0][FIX] Avoid error is product module is not installed

MiquelRForgeFlow:
Slip :S

--


pedrobaeza (approved):
Weird that you don't have `product` installed, but it's a possibility.

To be merged when CIs finish.

PR #1801, opened 2019-04-22
[FIX] base: Preserve passwords on new admin user

### LINKED ISSUE #1797: [12.0] Admin cannot log in with the old password.
Impacted versions: 12.0

Steps to reproduce: migrate a database from v11.

Current behavior:
Admin cannot log in with the old password. You need to reset admin password through sql query to be able to access the database as admin.

Expected behavior:
Admin can log in with the v11 password.

image

@pedrobaeza @StefanRijnhart @mreficent any clue?

pedrobaeza:
Closes #1797

cc @Tecnativa

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/end-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/end-migration.py
@@ -46,6 +46,16 @@ def fork_off_system_user(env):
         'partner_id': partner_admin.id,
         'login': login,
     })
+    # copy old passwords for not losing them on new admin user
+    crypt = openupgrade.column_exists(env.cr, 'res_users', 'password_crypt')
+    set_query = "SET password = ru2.password "
+    if crypt:
+        set_query += ", password_crypt = ru2.password_crypt "
+    env.cr.execute(
+        "UPDATE res_users ru " + set_query +
+        "FROM res_users ru2 WHERE ru2.id = %s AND ru.id = %s",
+        (user_root.id, user_admin.id),
+    )
     user_root.write({
         'partner_id': partner_root.id,
         'email': 'root@example.com',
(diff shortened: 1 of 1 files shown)


odoo/addons/base/migrations/12.0.1.3/end-migration.py
```diff
@@ -46,6 +46,15 @@ def fork_off_system_user(env):
         'partner_id': partner_admin.id,
         'login': login,
     })
+    # copy old passwords for not losing them on new admin user
+    set_query = "SET password = ru2.password "
+    if openupgrade.column_exists(env.cr, 'res_users', 'password_crypt'):
+        set_query += ", password_crypt = ru2.password_crypt "
```
StefanRijnhart, line 53: Thanks! password_crypt is guaranteed to exist as per 11.0 datamodel, as is also assumed down below in the `migrate` method. Can you set password_crypt to NULL in https://github.com/OCA/OpenUpgrade/pull/1801/files#diff-dc09a830b46c834278b85d839da79ba1L91?

LoisRForgeFlow (approved):
Thanks :+1:

odoo/addons/base/migrations/12.0.1.3/end-migration.py
pedrobaeza: @StefanRijnhart not really, as although `auth_crypt` module, which adds this field, is auto-installed by default, there's a chance that you manually uninstall it afterwards. About the other request, there's no need of doing that, as `password_crypt` is no longer used in v12.

StefanRijnhart: @pedrobaeza the reason I ask to empty `password_crypt` is because later on in the same script, its value is copied to `password`: https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/addons/base/migrations/12.0.1.3/end-migration.py#L100. It is also there where the existence of the column is assumed so your check is useless unless you also implement the check there (or leave it out as I asked you earlier, but I don't really mind which way).

pedrobaeza: OK, I see. Fixed.

StefanRijnhart: Thanks! I see I was wrong about the existence of this column being assumed in fill_res_users_password_from_password_crypt because it is checked before this method is called. Sorry for the confusion.

StefanRijnhart (approved):

MiquelRForgeFlow (approved):
LGTM

PR #1804, opened 2019-04-24
[12.0][MIG] account_bank_statement_import

MiquelRForgeFlow:
`#NothingToDo`

--


addons/account_bank_statement_import/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,5 @@
+---Fields in module 'account_bank_statement_import'---
```
pedrobaeza, line 1: Files swapped. This one is the original one.

pedrobaeza (changes requested):

addons/account_bank_statement_import/migrations/12.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: Right. Amended! 😂

pedrobaeza (approved):

PR #1812, opened 2019-04-30
[12.0][MIG] website: bootstrap (1st attempt)

MiquelRForgeFlow:
Here is a first approximation. It handles the majority of cases.

--

### DIFF
--- a/addons/website/migrations/12.0.1.0/post-migration.py
+++ b/addons/website/migrations/12.0.1.0/post-migration.py
@@ -1,6 +1,7 @@
 # Copyright 2018-19 Eficent <http://www.eficent.com>
 # Copyright 2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from lxml import etree
 from openupgradelib import openupgrade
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
cc @Yajo

yajo (changes requested):
The idea is good, but I see some important improvement points regarding maintainability and errors.

For instance, did you notice that this code could replace what you don't want?:

```python
new_node.attrib['class'] = new_node.attrib['class'].replace('item', 'carousel-item')
```

This would turn `<li class="navbar-item"/>` into `<li class="navbar-carousel-item"/>`, which is not what you want.

So, you need to rewrite this. See this new approach:

1.  Add a `REPLACEMENTS` constant dict where we define a structure similar to:

    ```python
    REPLACEMENTS = {
        # The key is an xpath used to match nodes that need this replacement
        "//*[hasclass('carousel')]/*[hasclass('item')]": {
            # Inside the dict, we have these keys, indicating what to do with the node
            "replace_class": {"item": "carousel-item", ...},

            # More examples of what could be done
            "add_attr": {"role": "navbar-item", ...},
            "add_class": {"btn", ...},
            "remove_attr": {"role", ...},
            "remove_class": {"item", ...},
        },
        ...
    }
    ```

1.  Define a method called `migrate_bootstrap_v4_document`, which does something like this:

    ```python
    def migrate_bootstrap_v4_document(html_string):
        # Important: use HTML parser to avoid possible XML incompatibilities
        document = lxml.html.html5parser.document_fromstring(html_string)
        for expr, instructions in REPLACEMENTS.items():
            nodes = document.xpath(expr)
            for node in nodes:
                migrate_bootstrap_v4_node(node, **instructions)
    ```

1.  Finally, we need the `migrate_bootstrap_v4_node`, also with something similar to:

    ```python
    def migrate_bootstrap_v4_node(node, add_attr=frozendict(), add_class=frozenset(), 
                                  remove_attr=frozenset(), remove_class=frozenset(), 
                                  replace_class=frozendict()):
        # Extract classes as set, to make replacements correctly
        classes = set(node.attrib["class"].split())
        for original, new in replace_class.items():
            if original in classes:
                classes.discard(original)
                classes.add(new)
        classes = classes | add_class ^ remove_class
        node.attrib["class"] = " ".join(classes)
        # Perform attribute replacements
        ...
    ```

That should be enough. It will be less error-prone, and more readable and maintainable.

Thanks!

pedrobaeza:
@Yajo @mreficent remember that this is not the idea. This MUST be in an openupgradelib method, as the same conversion should be done on several fields across all Odoo (any that is HTML and it's filled through website editor - example, blog content field).

pedrobaeza:
@Yajo please share here the full design.

yajo:
> please share here the full design

That's what I've done, isn't it? :thinking:

pedrobaeza:
I mean full design in our internal task. I copy it here. Sorry, it's in Spanish, but I'm on holidays and I don't want to spend more time:
```
- Crear en openupgradelib un archivo openupgrade_12.py para meter ahí las funciones de esto que es específico de la versión.

- Crear una función que sea algo así como `convert_field_to_bs4` que le pasemos como parámetros un modelo, un campo, un posible dominio, y coja los valores de dicho campo y los convierta.

- Esa función llamará a otra de más bajo nivel - `_convert_content_to_bs4` o algo así, con un texto (HTML se supone) como entrada, y que devuelva el texto convertido a HTML BS4. Aquí es donde estará la mayor chicha. Hay que observar la guía https://github.com/odoo/odoo/wiki/Tips-and-tricks:-BS3-to-BS4 para ver cambios obvios de tipo "replace", y luego también observar los bloques base de v11 y los de v12 y detectar patrones para sustituirlos y que se vean igual (salvando las personalizaciones propias que pueden haber, como por ejemplo haber cambiado una imagen, o el texto del título, etc).

- Por último, será llamar a dichos métodos desde los distintos scripts de OpenUpgrade. El primero sería el propio website. Con una conversión del campo `arch_db` de los ir.ui.view que apliquen. Pero luego habrá que también llamarlo para el campo de contenido de los blog post, de los productos de website_sale...
```

yajo:
OK thanks, I think @mreficent has no problem with Spanish :stuck_out_tongue: 

So, @mreficent, the sum of https://github.com/OCA/OpenUpgrade/pull/1812#pullrequestreview-233368134 and https://github.com/OCA/OpenUpgrade/pull/1812#issuecomment-489025511 is the full task. Do you want us to share some efforts?

MiquelRForgeFlow:
> This would turn `<li class="navbar-item"/>` into `<li class="navbar-carousel-item"/>`, which is not what you want.

But why would you have a `navbar-item` inside a `carousel`? 🤔 It doesn't makes sense to me. My code is implying the original source code is correctly written.

@Yajo Yeah, I will do some updates to the code using your review, but obviously I will need more reviews after those changes. To pass this all code to `openupgradelib` just will be the last step.

yajo:
> It doesn't makes sense to me

Expect the unexpected. End users have direct access to an HTML editor. 

Also there might be theme-specific classes involved. For instance a carousel which contains some `<div class="custom-theme-red-item">` would also be broken.

OTOH, you made me realize that there's something more we didn't take into account: Bootstrap components usually can be nested and aren't forced to be direct descendants. There could be an extra `<div>` in between. We need to support those cases... and once they are supported, yes, someone could have a nav inside a carousel. Who knows? :man_shrugging: 

> obviously I will need more reviews after those changes

Don't worry, we're here to help! :blush:

yajo:
Oh sorry, I'm a little bit in a hurry to finish this. I'm going to continue the work from here onwards. Thanks!

MiquelRForgeFlow:
Well, I think we can't pretend to fix ALL possible cases, because all possible cases are infinite. We should pretend just to obtain a generic method that works for OCA migration, and maybe other simple migrations, but not adapted for very specific customized views that someone might have.

yajo:
As I said, I'll continue from here. Thanks for your contribution.

@pedrobaeza Could we please close this? The 1st PR should go for openupgradelib, then I'll open another one here as needed. This is useless as it is right now.

PR #1813, opened 2019-05-01
[FIX] apriori incorrect rename (product_supplierinfo_for_customer_sale)

legalsylvain:

- Openugrapde from 8.0 to 10.0
- during the process 9.0 to 10.0, an error occures
```
IntegrityError: duplicate key value violates unique constraint "name_uniq"
DETAIL:  Key (name)=(product_supplierinfo_for_customer_sale) already exists.
```

This is due to the apriori.py file from openupgrade 10.0 that mentions that the module `product_customer_code_sale` has been renamed into `product_supplierinfo_for_customer_sale`

The problem is that 
- `product_supplierinfo_for_customer_sale` exists in V8, so the module is present in `ir_module_module` table (OCA/product-attribute/8.0) https://github.com/OCA/product-attribute/tree/8.0/product_supplierinfo_for_customer_sale
- `product_customer_code_sale` exists in V8, so the module is also present in `ir_module_module` table. (OCA/sale-workflow/8.0) https://github.com/OCA/sale-workflow/tree/8.0/product_customer_code_sale

I don't know if this is the best solution, but it seems that the current code is not correct, so I disabled the line.

regards.

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -30,7 +30,6 @@ renamed_modules = {
     'sale_delivery_block': 'sale_stock_picking_blocking',
     'sale_delivery_block_proc_group_by_line':
         'sale_stock_picking_blocking_proc_group_by_line',
-    # 'product_customer_code_sale': 'product_supplierinfo_for_customer_sale',
     # OCA/server-tools:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please move it instead to `merged_modules`.

legalsylvain:
done. Thanks for your insight.

pedrobaeza (approved):

PR #1814, opened 2019-05-03
[12.0][MIG] website_img_bg_style: Merge it in website

yajo:
This module has no use in v12, where background image styles are upstream and much better than what this module was.

@Tecnativa

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -36,6 +36,8 @@ merged_modules = {
     # OCA/web
     'web_no_bubble': 'web',
     'web_sheet_full_width': 'web_responsive',
+    # OCA/website
+    'website_img_bg_style': 'website',
 }
 
 renamed_models = {


pedrobaeza:
No transformation is needed for upstream system to handle existing background images?

yajo:
I see nothing much different:

- https://github.com/odoo/odoo/blob/a80f1ddcda416c76760acd920a7114382749a71e/addons/website/views/snippets.xml#L1252
- https://github.com/OCA/website/blob/d7e672bd2f24d708e4d97b82181441b0e974e8b4/website_img_bg_style/views/snippets.xml#L12

It seems like the same features are enabled in the same elements.

pedrobaeza:
OK, you tell me. I see there different styles. If you confirm that backgrounds in previous versions are preserved and with no drawback, then I merge this.

yajo:
Oops, thanks for insisting. This needs some work still.

`website_img_bg_style` was adding some classes to perform its changes, but the new upstream odoo 12.0 feature adds different classes and styles, so we still need to make the translation from previous classes to some new classes/styles.

The best approach will be to leverage the features introduced in #1812, so let's set this as WIP until that other is finished.

Thanks!

yajo:
Oh I have no permissions. Could you label it please?

yajo:
This should happen in v10, where the needed features are already present. Will do, closing this.

PR #1819, opened 2019-05-10
[MIG] website: Handle Bootstrap 3->4 and Less->Scss

yajo:
## Less to Scss

In case a user has customized any Less attachments, these are disabled because 99% of the times it will fail: Syntax from Scss and Less is different, there may not be a `lessc` executable in the Odoo 12 environment, and inherited views would probably fail finding the node to replace from XPath.

Automated resilient conversion between Less and Scss is close to impossible.

To avoid problems, these views are disabled. You are still able to fix and reenable them manually as needed.

## Bootstrap 3 to 4, with new multiwebsite support

In Odoo v12, the standard behavior when a user modifies a website template is to create a modified copy which is specific to the current website (a.k.a. COW 🐮).

To benefit from this new behavior, when migrating a website from Odoo 11 to 12:

1. If a website-specific view exists, it is migrated in place.
2. Website-agnostic views are migrated in a COW 🐮 view, specific for each preexisting website.
3. Webiste-agnostic views are marked as `noupdate=0`, so when starting a new website, you have the updated views as a starting point. To make this work, all this conversion had to be made in the pre-migration stage.
4. The new copies will have a hidden migration column that indicates from which view was it copied.

To make this work, https://github.com/OCA/openupgradelib/pull/148 is needed.

--
@Tecnativa


yajo:
Set as WIP please, I still have to test this.

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -32,45 +32,12 @@ def fill_website_socials(cr):
...
-
-def bootstrap_4_migration(env):
-    pages = env['website.page'].search([])
-    views = pages.mapped('view_id').filtered(
-        lambda v: v.type == 'qweb' and not v.xml_id)
-    for view in views:
-        apply_bootstrap_4(view)
-
-
-def apply_copy_views(env):
```
MiquelRForgeFlow, line 49: I don't get why you delete this method. But if you are sure with this, then please update the `openupgrade_analysis_work.txt` file with your reasons.

pedrobaeza: It was not complete (only website pages were handled, but not the rest of the website views).

yajo: Yes, and the valid logic inside has been moved to pre stage.

For instance, the contactus page is marked as noupdate (so it shouldn't be touched by migration scripts), but [it's included in the noupdate records that should change](https://github.com/OCA/OpenUpgrade/blob/069f5623491fe9b90264b8e5776b724a833a4ec4/addons/website/migrations/12.0.1.0/noupdate_changes.xml#L12) (because of the inherent logic inside the view). So, if we copied+migrated views in the post stage, that one would be already modified and thus lost.

Instead of that, we duplicate and migrate to BS4 in the pre stage, and then the rest of normal migration happens as usual. If i.e. the contactus page for website 1 is broken, you just disable it and revert to its website-agnostic version, just as it came with the module.

yajo:
Ready to review. See https://github.com/OCA/openupgradelib/pull/148#issuecomment-493980364.

MiquelRForgeFlow:
@Yajo Please, rebase

MiquelRForgeFlow (approved):
LGTM

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -32,35 +43,62 @@ def fill_website_socials(cr):
...
+            continue
+        specific_page = env["website.page"].search([
+            ("url", "=", agnostic_page.url),
+            ("website_id", "=", specific_view.website_id.id),
+        ])
+        if not specific_page:
+            specific_page = agnostic_page.copy({
+                "is_published": agnostic_page.is_published,
+                "url": agnostic_page.url,
+                "view_id": specific_view_id,
```
pedrobaeza, line 82: This is missing `"website_id": specific_view.website_id.id`

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -32,35 +43,62 @@ def fill_website_socials(cr):
...
+            })
+        # Create website-specific menu for the copied page
```
pedrobaeza, line 87: If found specif page, you should assign `specific_view_id` to it.

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -32,35 +43,62 @@ def fill_website_socials(cr):
...
+        specific_menu = env["website.menu"].search([
+            ("website_id", "=", specific_page.website_id.id),
+            ("url", "=", specific_page.url),
+        ], limit=1)
+        if specific_menu:
+            if specific_menu.page_id:
```
pedrobaeza, line 93: I think you should assign the specific page nevertheless the previous value

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -32,35 +43,62 @@ def fill_website_socials(cr):
...
+                specific_menu.page_id = specific_page
+        else:
+            agnostic_menu = env["website.menu"].search([
+                ("website_id", "=", False),
+                ("url", "=", specific_page.url),
+            ], limit=1)
+            if agnostic_menu:
+                agnostic_menu.copy({
+                    "website_id": specific_page.website_id.id,
+                    "page_id": agnostic_menu.page_id.id and specific_page.id,
```
pedrobaeza, line 103: This should be:
```suggestion
                    "page_id": specific_page.id or agnostic_menu.page_id.id,
```

```
>>> 3 and 5
5
>>> 3 and False
False
```

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -1,11 +1,194 @@
...
+    # Find updatable views, to be excluded; standard addon update is enough
+    udpatable_ids = env["ir.model.data"].search([
+        ("model", "=", "ir.ui.view"),
+        ("noupdate", "=", False),
+    ]).mapped("res_id")
+    # In Odoo v11, a page can be related to multiple websites, and the related
+    # view doesn't need to be related to a website (although it could be);
+    # in v12, a page is related to a view, which is related (or not) to a sigle
+    # website. Usually it would be impossible to link a page to multiple
+    # websites using UI in v11, so it's not like a
```
pedrobaeza: Comment not finished?

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -1,11 +1,194 @@
...
+        # will be stored in the view; the others will be lost, which isn't a
+        # problem since v11 didn't support multi-websites
+        openupgrade.logged_query(
+            env.cr,
+            "UPDATE ir_ui_view SET website_id = %s WHERE id = %s",
+            (page_website_id, view_id),
+        )
+    # Find views to convert
+    all_views = env['ir.ui.view'].search([
+        ("active", "=", True),
```
pedrobaeza: I think we should convert also non-active views just in case they are activated later.

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -1,11 +1,194 @@
...
+        list(chain.from_iterable(env.cr.fetchall())),
+        prefetch=all_views._prefetch,
+    )
+    for view in views_with_website:
+        new_arch = convert_string_bootstrap_3to4(view.arch_base)
+        new_arch = _apply_v11_dummy_image_replacement(
+            image_replacements,
+            new_arch,
+        )
+        view.arch_base = new_arch
```
pedrobaeza: Why not using `arch_db` field directly for avoiding any kind on manipulation on the content?

pedrobaeza (changes requested):

addons/website/migrations/12.0.1.0/post-migration.py
yajo: Yes, thanks! Fixed.

addons/website/migrations/12.0.1.0/post-migration.py
yajo: Well, that's the case only if the found record was linked to `agnostic_view_id`.

It's weird but website-specific pages can be linked to website-agnostic views legally.

I added this case however, thanks!

addons/website/migrations/12.0.1.0/post-migration.py
yajo: If the menu wasn't pointing to a page before, it's better to leave it as it was. Maybe it was just a "directory" menu.

addons/website/migrations/12.0.1.0/post-migration.py
yajo: If I'm not wrong, is OK as it is :thinking:. The meaning is: "if the agnostic menu was linked to a page, link the specific menu to the specific page".

addons/website/migrations/12.0.1.0/pre-migration.py
yajo: Oops, fixed.

addons/website/migrations/12.0.1.0/pre-migration.py
yajo: OK.

addons/website/migrations/12.0.1.0/pre-migration.py
yajo: Actually I wanted to let Odoo do its stuff, whatever it is. It seems to handle updates in several translations automatically. :man_shrugging:  Well, I didn't want to dig so much, just use the one that is exposed to the user and seems to have a proper inverse method.

yajo (commented):
All attended.

addons/website/migrations/12.0.1.0/pre-migration.py
MiquelRForgeFlow: You should change to `arch_db` here because `arch_base` is not stored. With `arch_db` instead, it will trigger the compute `_compute_arch` that will calculate the `arch`, and then this will trigger the compute `_compute_arch_base` that will calculate the `arch_base` and so on.

pedrobaeza:
@mreficent @Yajo I commented with @JKE-be in Spanish OCA days about the noupdate flag in several website views. They are amending this in odoo/odoo#33919, which is great, but we should modify now this code according this:

- Do an upstream update when merged.
- Remove view update part in noupdate_records.xml and see if this requires a manual reset of the noupdate flag for that views.

addons/website/migrations/12.0.1.0/pre-migration.py
yajo: That seems to be the same as what the inverse method of `arch_base` does, so I see no difference. Anyways I'll change it, you're the pros.

yajo:
All attended, except https://github.com/OCA/OpenUpgrade/pull/1819#issuecomment-499153085, which I understand should be done separately.

pedrobaeza:
@Yajo we have merged now #1886, which contains the noupdate switch, so you can now do the last part for not adding noupdate records that are going to be removed. You have to include in pre-migration script the reset of that ir.model.data records to noupdate=False.

yajo:
Done now.

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -1,11 +1,203 @@
...
+                "datas": b64encode(image_stream.read()),
+            })
+            replacements[xmlid] = new.id
+    return replacements
+
+
+def noupdate_changes(env):
+    """Set some specific noupdate=1 records to noupdate=0"""
+    noupdates = env["ir.model.data"].search([
+        ("model", "=", "website"),
```
MiquelRForgeFlow: `("model", "=", "ir.ui.view"),` 🤔

And not sure if also `("module", "=", "website"),` 🤔

pedrobaeza: Yes, better for restricting (we can have that XML-ID in other modules).

yajo: Thanks, it was a typo. Fixed now.

pedrobaeza (approved):

PR #1822, opened 2019-05-15
[12.0][MIG] website_partner

MiquelRForgeFlow:
Migration scripts for `website_partner` module.

--


addons/website_partner/migrations/12.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,10 @@
+---Fields in module 'website_partner'---
+website_partner / res.partner              / website_meta_og_img (char)    : NEW
+# NOTHING TO DO: used for linking the image that is going to be showed when linking the page on social media. Default one is company logo.
```
pedrobaeza, line 3: Thanks for explaining in a detailed way the nothing to do. I was tempted to say that we should default to partner image, but seeing that Odoo don't do that either, having this unfilled is good.

pedrobaeza (approved):

PR #1846, opened 2019-05-22
[12.0][FIX] hr_holidays: assure ids before renaming models

MiquelRForgeFlow:
Fixes syntax error in `_move_model_in_data` if no `ids` are passed.

--

### DIFF
--- a/addons/hr_holidays/migrations/12.0.1.5/post-migration.py
+++ b/addons/hr_holidays/migrations/12.0.1.5/post-migration.py
@@ -80,7 +80,8 @@ def fill_hr_leave(env):
         RETURNING id""", (AsIs(openupgrade.get_legacy_name('type')), ),
     )
     ids = [x[0] for x in env.cr.fetchall()]
-    _move_model_in_data(env, ids, 'hr.holidays', 'hr.leave')
+    if ids:
+        _move_model_in_data(env, ids, 'hr.holidays', 'hr.leave')
 
 
 def fill_hr_leave_request_dates(cr):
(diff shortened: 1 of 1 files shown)


addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -79,8 +79,9 @@ def fill_hr_leave(env):
         WHERE %s = 'remove'
         RETURNING id""", (AsIs(openupgrade.get_legacy_name('type')), ),
     )
-    ids = [x[0] for x in env.cr.fetchall()]
-    _move_model_in_data(env, ids, 'hr.holidays', 'hr.leave')
+    ids = [x[0] for x in env.cr.fetchall() if x]
```
pedrobaeza: No need of the `if x` part

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -112,8 +113,9 @@ def fill_hr_leave_allocation(env):
         WHERE %s = 'add'
         RETURNING id""", (AsIs(openupgrade.get_legacy_name('type')), ),
     )
-    ids = [x[0] for x in env.cr.fetchall()]
-    _move_model_in_data(env, ids, 'hr.holidays', 'hr.leave.allocation')
+    ids = [x[0] for x in env.cr.fetchall() if x]
```
pedrobaeza: The same

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
MiquelRForgeFlow: amended

pedroslopez:
Looks good to me!

pedrobaeza (approved):

PR #1847, opened 2019-05-22
[12.0][FIX] account: assure origin moves before accessing index

MiquelRForgeFlow:
Fixes error when trying to access index of an empty record.

--


pedroslopez:
Yep, this works. Thanks!

pedrobaeza (approved):

PR #1848, opened 2019-05-23
[11.0][IMP] partner_sector: Better handling of the transition for this module

pedrobaeza:
* Rename tables instead of copying data
* Remove the need to reassign many2one references.

@Tecnativa

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
@@ -25,6 +25,22 @@ model_renames_ir_actions_report = [
 ]
 
 
+def handle_partner_sector(env):
+    if openupgrade.table_exists(env.cr, 'res_partner_sector'):
+        # Module `partner_sector` was installed in 10.0
+        openupgrade.rename_models(
+            env.cr, [('res.partner.sector', 'res.partner.industry')],
+        )
+        openupgrade.rename_tables(
+            env.cr, [(('res_partner_sector', 'res_partner_industry'))]
+        )
+        openupgrade.rename_fields(
+            env, [
+                ('res.partner', 'res_partner', 'sector_id', 'industry_id'),
+            ],
+        )
+
+
 def fill_cron_action_server_pre(env):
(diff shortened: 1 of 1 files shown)


odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -25,6 +25,22 @@
...
+    if openupgrade.table_exists(env.cr, 'res_partner_sector'):
+        # Module `partner_sector` was installed in 10.0
+        openupgrade.rename_models(
+            env.cr, [('res.partner.sector', 'res.partner.industry')],
+        )
+        openupgrade.rename_tables(
+            env.cr, [(('res_partner_sector', 'res_partner_industry'))]
+        )
+        openupgrade.rename_fields(
+            env.cr, [
```
MiquelRForgeFlow: should be `env` instead of `env.cr`

MiquelRForgeFlow (changes requested):
LGTM

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
pedrobaeza: Totally! fixed

MiquelRForgeFlow: but the problem I see is that in the base pre-migration, the `env` still doesn't have the base models if remember correctly. 🤔

pedrobaeza: Uhm, good question... let me check if we need it for something specific.

pedrobaeza: There's no problem:

https://github.com/OCA/openupgradelib/blob/12b1ed9c7d2c354499ba5f89067f908f0f8f94ad/openupgradelib/openupgrade.py#L413-L415

MiquelRForgeFlow: good

MiquelRForgeFlow (approved):

PR #1850, opened 2019-05-24
[12.0][FIX] hr_holidays: Avoid trying to repeat followers

MiquelRForgeFlow:
There is a constraint in `mail.followers` for unique `res_id`, `partner_id`, `res_model`.

Thus, we need to ensure we don't repeat the `res_id` to avoid the constraint error.

--


pedroslopez (approved):
This fixed issues I was having when migrating a database

pedrobaeza (approved):

PR #1851, opened 2019-05-24
[12.0][FIX] account_voucher: assure company_id in old databases

MiquelRForgeFlow:
In odoo v8, the company_id was stored. Thus, maybe some databases
that have been migrating since v8, may still have that column. Thus,
if new records have been created in newer versions, they may have
this field empty.

--

### DIFF
--- a/addons/account_voucher/migrations/12.0.1.0/pre-lol.py
+++ b/addons/account_voucher/migrations/12.0.1.0/pre-lol.py
@@ -1,0 +1,24 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def assure_company(cr):
+    openupgrade.logged_query(
+        cr, """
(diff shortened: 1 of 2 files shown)


addons/account_voucher/migrations/12.0.1.0/pre-lol.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def assure_company(cr):
+    openupgrade.logged_query(
+        cr, """
+        UPDATE account_voucher av
+        SET company_id = aj.company_id
+        FROM account_journal aj
+        WHERE av.journal_id = aj.id
+        """
```
pedrobaeza: Add here `AND av.company_id IS NULL`

pedrobaeza (changes requested):

addons/account_voucher/migrations/12.0.1.0/pre-lol.py
MiquelRForgeFlow: done

pedrobaeza (approved):

PR #1856, opened 2019-05-27
[MIG] website_img_bg_style: Handle module removal

yajo:
This module is removed because the features it added are now upstream.

Applying needed conversion to keep consistent styling and UX.

Replaces #1814.

WIP until merged:

- [x] https://github.com/OCA/openupgradelib/pull/148

--
@Tecnativa

### DIFF
--- a/addons/website/migrations/10.0.1.0/pre-migration.py
+++ b/addons/website/migrations/10.0.1.0/pre-migration.py
@@ -1,7 +1,79 @@
 # -*- coding: utf-8 -*-
 # © 2017 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from itertools import product
+
 from openupgradelib import openupgrade
+from openupgradelib.openupgrade_tools import (
+    convert_html_fragment,
+    convert_html_replacement_class_shortcut as _r,
+)
+
+_BG_SIZES = ("25%", "50%", "75%", "auto")
+REPLACEMENTS = [
+    # Replacements to remove website_img_bg_style
+    _r("bg-center-h bg-bottom", "oe_img_bg o_bg_img_opt_custom",
+       styles={"background-position": "50% 100%"}),
+    _r("bg-center-h bg-top", "oe_img_bg o_bg_img_opt_custom",
+       styles={"background-position": "50% 0%"}),
+    _r("bg-center-v bg-left", "oe_img_bg o_bg_img_opt_custom",
+       styles={"background-position": "0% 50%"}),
+    _r("bg-center-v bg-right", "oe_img_bg o_bg_img_opt_custom",
+       styles={"background-position": "100% 50%"}),
+    _r("bg-left bg-bottom", "oe_img_bg o_bg_img_opt_custom",
+       styles={"background-position": "0% 100%"}),
+    _r("bg-left bg-top", "oe_img_bg o_bg_img_opt_custom",
+       styles={"background-position": "0% 0%"}),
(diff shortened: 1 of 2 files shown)


addons/website/migrations/10.0.1.0/pre-migration.py
```diff
@@ -1,7 +1,79 @@
 # -*- coding: utf-8 -*-
 # © 2017 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from itertools import product
+
 from openupgradelib import openupgrade
+from openupgradelib.openupgrade_tools import (
```
pedrobaeza, line 7: Is this compatible with Python 2?

yajo: Yes, it's `openupgrade_120` that isn't.

pedrobaeza:
It seems there's a incompatibility between versions in `openupgradelib` or here:
```
AttributeError: 'module' object has no attribute 'tools'
```

MiquelRForgeFlow:
it seems to me that is openupgradelib or travis files :S

yajo:
Restarted Travis build, now that https://github.com/OCA/openupgradelib/pull/148 is merged.

pedrobaeza:
@Yajo I already did that. Please check the previous log with the problem.

yajo:
I can see no relation between my patch and those errors, apart from the fact that it's pretty outdated. I rebased; let's see.

pedrobaeza:
I would say that it has to be with the import you are doing, that provokes another imports in the target file, and the order/path mangling in Odoo doesn't fit exactly.

addons/website/migrations/10.0.1.0/pre-migration.py
```diff
@@ -1,7 +1,79 @@
...
+    REPLACEMENTS.append(_r(
+        "bg-size-%s-h bg-size-%s-v" % (h.strip("%"), v.strip("%")),
+        "oe_img_bg o_bg_img_opt_custom",
+        styles={"background-size": "%s %s" % (h, v)}))
+
+
+def update_website_views(env):
+    """Handle removal of website_img_bg_style module."""
+    # Find report views, which should never be converted here
+    report_names = env["ir.actions.report"].search([
```
MiquelRForgeFlow: -> `ir.actions.report.xml`

MiquelRForgeFlow (changes requested):

yajo:
All attended.

MiquelRForgeFlow (approved):

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -35,6 +35,8 @@
     # OCA/stock-logistics-workflow:
     'product_customer_code_picking':
         'product_supplierinfo_for_customer_picking',
+    # OCA/website
+    'website_img_bg_style': 'website',
```
MiquelRForgeFlow: wrong, this should go below, in the `merged_modules` section

MiquelRForgeFlow (dismissed):

odoo/addons/openupgrade_records/lib/apriori.py
pedrobaeza: Please fix this

pedrobaeza (changes requested):

odoo/addons/openupgrade_records/lib/apriori.py
pedrobaeza: I have changed this myself.

pedrobaeza (approved):

PR #1860, opened 2019-05-29
[12.0][account] try to link moves with their reversals

JordiBForgeFlow:
Even if you have account_reversal installed, it may well be that there are old reversals that were created without this module. This PR will try to link all moves with their corresponding reversals, excluding those that were created using module 'account_reversal'.

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/post-migration.py
+++ b/addons/account/migrations/12.0.1.1/post-migration.py
@@ -156,31 +156,32 @@ def fill_account_journal_alias_id(env):
 
 
 def fill_account_move_reverse_entry_id(env):
-    all_moves = env['account.move'].search([])
-    moves_linked_to_reversals = env['account.move']
+    AccountMove = env['account.move']
     if openupgrade.table_exists(env.cr, 'account_move_reverse'):
-        reversal_moves = all_moves.filtered(
-            lambda m: m.auto_reverse and not m.reverse_entry_id)
+        reversal_moves = AccountMove.search([
+            ('auto_reverse', '=', True),
+            ('reverse_entry_id', '=', False),
+        ])
         reversal_moves.write({'reverse_date': '2999-12-31'})
         env['ir.filters'].create({
             'name': 'To be reversed (undetermined date)',
             'model_id': 'account.move',
-            'domain': [('auto_reverse', '=', True),
-                       ('reverse_date', '=', '2999-12-31')],
+            'domain': "[('auto_reverse', '=', True),"
+                      "('reverse_entry_id', '=', False),"
+                      "('reverse_date', '=', '2999-12-31')]",
         })
-        moves_linked_to_reversals += all_moves.filtered(
-            lambda m: m.reverse_entry_id)
+    domain = [('reverse_entry_id', '=', False)]
     installed_langs = env['res.lang'].search([])  # search only active
     for lang in installed_langs:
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (changes requested):
This fix has no sense. There is no `for`. Only there is an `if` and an `else`.

JordiBForgeFlow:
@mreficent the if checks wether you habe the account_reversal module installed. But if you have it installed, and you have prior reversals made without this module, then the old reversals will not be fixed.

This PR fixes that, and I checked that it solves the issue it describes.

addons/account/migrations/12.0.1.1/post-migration.py
```diff
@@ -167,14 +168,17 @@ def fill_account_move_reverse_entry_id(env):
...
     else:
         installed_langs = env['res.lang'].search([])  # search only active
         for lang in installed_langs:
             reversal_text = env['ir.translation']._get_source(
                 name=False, types='code', lang=lang.code,
                 source='reversal of: ')
             reversal_moves = all_moves.filtered(
-                lambda m: m.ref and m.ref.startswith(reversal_text))
+                lambda m: m.ref and m.ref.startswith(reversal_text)
+                and m not in moves_linked_to_reversals)
```
MiquelRForgeFlow: here `moves_linked_to_reversals` will be always empty, so it's irrelevant adding `m not in moves_linked_to_reversals`

addons/account/migrations/12.0.1.1/post-migration.py
```diff
@@ -167,14 +168,17 @@ def fill_account_move_reverse_entry_id(env):
             'domain': [('auto_reverse', '=', True),
                        ('reverse_date', '=', '2999-12-31')],
         })
+        moves_linked_to_reversals += all_moves.filtered(
```
mreficent: and here you are filling `moves_linked_to_reversals` but without any use

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow:
Fixed :+1:

MiquelRForgeFlow (approved):

pedrobaeza:
Added improvement for better performance.

addons/account/migrations/12.0.1.1/post-migration.py
```diff
@@ -156,31 +156,31 @@ def fill_account_journal_alias_id(env):
...
 def fill_account_move_reverse_entry_id(env):
-    all_moves = env['account.move'].search([])
-    moves_linked_to_reversals = env['account.move']
+    AccountMove = env['account.move']
     if openupgrade.table_exists(env.cr, 'account_move_reverse'):
-        reversal_moves = all_moves.filtered(
-            lambda m: m.auto_reverse and not m.reverse_entry_id)
+        reversal_moves = AccountMove.search([
+            ('auto_reverse', '=', True),
+            ('reverse_entry_id', '!=', False),
```
MiquelRForgeFlow: 'reverse_entry_id', **'='**, False

MiquelRForgeFlow (changes requested):

pedrobaeza (approved):

MiquelRForgeFlow:
I just fixed flake8

MiquelRForgeFlow (approved):

Issue #1859, opened 2019-05-29
[12.0] Account payment term multicompany ir.rule disabled


JordiBForgeFlow:
I don't see why the rule 'account.account_payment_term_comp_rule' is explicitly disabled after the migration the post-migration:

https://github.com/OCA/OpenUpgrade/blame/12.0/addons/account/migrations/12.0.1.1/post-migration.py#L8

@mreficent

pedrobaeza:
Check https://github.com/OCA/OpenUpgrade/issues/1448#issuecomment-405007947. It was for preserving the same behavior that you have in previous versions, but maybe it's better to revert to my previous idea of emptying `company_id`.

MiquelRForgeFlow:
I explained explicitly in https://github.com/OCA/OpenUpgrade/pull/1861.

PR #1861, opened 2019-05-29
[12.0][DOC] account: explain why disabling a rule

MiquelRForgeFlow:
Attends https://github.com/OCA/OpenUpgrade/issues/1859.

--


pedrobaeza:
What do you think about reverting my original idea of emptying `company_id` in all existing payment terms? This is for making use of the new feature from now on.

MiquelRForgeFlow:
The problem I see is that the `company_id` in payment terms is **required**.

MiquelRForgeFlow:
@jbeficent could you approve?

pedrobaeza (approved):
Let's move on on this

PR #1871, opened 2019-06-04
[12.0][IMP] openupgrade_records

MiquelRForgeFlow:
This PR includes two improvements in the openupgrade_records reports:

- In new stored records, show if they are computed or related.
- In case of relational fields that have a renamed relation (model), show them as "[Nothing to do]".

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -250,8 +250,11 @@ def compare_sets(old_records, new_records):
             continue
         if column['mode'] == 'create':
             column['mode'] = ''
+        printkeys_plus = printkeys.copy()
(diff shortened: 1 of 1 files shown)


odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -250,8 +252,11 @@ def match(match_fields, report_fields, warn=False):
             continue
         if column['mode'] == 'create':
             column['mode'] = ''
+        printkeys_plus = printkeys
```
pedrobaeza:
```suggestion
        printkeys_plus = printkeys.copy()
```

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #1874, opened 2019-06-05
[11.0][MIG] website_event_track

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/1720.

--


addons/website_event_track/migrations/11.0.1.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+from openupgradelib import openupgrade
+
+
+def fill_event_track_stage_id(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE event_track et
+        SET stage_id = ets.id
+        FROM event_track_stage ets
+        WHERE et.%s = lower(ets.name)
```
MiquelRForgeFlow: pending to check how to do with people that have the translated names for the stages, because `name` field has translations=True.

pedrobaeza: Fixed using XML-IDs

pedrobaeza (approved):

PR #1877, opened 2019-06-11
[12.0][MIG]Avoid account_id non null constraint

OthmaneGhandi:

In account pre-migration.py when we try to insert sections in "account_invoice_line" we got : "account_id" non null constraint, because in V11 the field is always required 

we drop not null constraint from account_id field --> after that the modules in V12 will be loaded and the constraint will be set correctly as defined (depends on display_type)

--


pedrobaeza (approved):
Good found! Thanks

OthmaneGhandi:
> Good found! Thanks

You are welcome Sir

PR #1878, opened 2019-06-11
[12.0][MIG] In Sale : avoid 'product_id' & 'product_uom' non null constraint

OthmaneGhandi:

In **Sale pre-migration.py** when we try to insert sections in "sale_order_line" we got : "product_id" & "product_uom" non null constraint, because in V11 those 2 fields is always required

we drop not null constraint from product_id field & and added a default value for product_uom --> after that the modules in V12 will be loaded and the constraint will be set correctly as defined (depends on display_type)

--

### DIFF
--- a/addons/sale/migrations/12.0.1.1/pre-migration.py
+++ b/addons/sale/migrations/12.0.1.1/pre-migration.py
@@ -51,11 +51,16 @@ def fill_sale_order_line_sections(cr):
     )
     openupgrade.logged_query(
         cr, """
+        ALTER TABLE sale_order_line ALTER COLUMN product_uom DROP not null
+        """,
+    )
+    openupgrade.logged_query(
+        cr, """
         INSERT INTO sale_order_line (order_id, layout_category_id,
-            sequence, name, price_unit, product_uom, product_uom_qty, customer_lead,
+            sequence, name, price_unit, product_uom_qty, customer_lead,
             display_type, create_uid, create_date, write_uid, write_date)
         SELECT sol.order_id, sol.layout_category_id,
-            min(sol.sequence) -1 as sequence, max(slc.name), 0, 1, 0, 0,
+            min(sol.sequence) -1 as sequence, max(slc.name), 0, 0, 0,
             'line_section', min(sol.create_uid), min(sol.create_date),
             min(sol.write_uid), min(sol.write_date)
         FROM sale_order_line sol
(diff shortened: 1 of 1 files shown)


addons/sale/migrations/12.0.1.1/pre-migration.py
```diff
@@ -44,13 +44,18 @@ def fill_sale_order_line_sections(cr):
...
+    )
     openupgrade.logged_query(
         cr, """
         INSERT INTO sale_order_line (order_id, layout_category_id,
-            sequence, name, price_unit, product_uom_qty, customer_lead,
+            sequence, name, price_unit, product_uom, product_uom_qty, customer_lead,
             display_type, create_uid, create_date, write_uid, write_date)
         SELECT sol.order_id, sol.layout_category_id,
-            min(sol.sequence) -1 as sequence, max(slc.name), 0, 0, 0,
+            min(sol.sequence) -1 as sequence, max(slc.name), 0, 1, 0, 0,
```
pedrobaeza: You are harcoding a UoM, which is not correct.

pedrobaeza (changes requested):

pedrobaeza:
Please perform each change in a separate PR, with the explanation for each one.

There's other object (`sale.quotation.template.line` or similar) where this problem might be found.

addons/sale/migrations/12.0.1.1/pre-migration.py
OthmaneGhandi: In this case I think it will be better drop the non null constraint for product_uom too (like product_id), what do you think @pedrobaeza ?

pedrobaeza: Is it required in v12? If not, drop it.

MiquelRForgeFlow: is not required

MiquelRForgeFlow:
please, squash commits, putting the name of the second commit

@pedrobaeza about the sale.quotation.template.line problem, is in the `sale_management`

addons/sale/migrations/12.0.1.1/pre-migration.py
OthmaneGhandi: It's required V11 so when we are trying to add sections to sale order line we get blocked by the non-null constraint 
In V12 it's not required when display_type field is set

pedrobaeza (approved):
Squashing on merge

MiquelRForgeFlow (approved):

Issue #1875, opened 2019-06-07
[12.0-product] Attribute line values are gone with product variants


cvinh:
Impacted versions: 12.0

Steps to reproduce:
Migrate to v12.0 a database with product variants
Current behavior:
attribute line values do not appear
Expected behavior:
attribute line values should be migrated

Video/Screenshot link (optional):

MiquelRForgeFlow:
See https://github.com/OCA/OpenUpgrade/pull/1882.

PR #1882, opened 2019-06-11
[12.0][FIX] product: fix migration of attribute lines

MiquelRForgeFlow:
Solves https://github.com/OCA/OpenUpgrade/issues/1875.

--

### DIFF
--- a/addons/product/migrations/12.0.1.2/end-migration.py
+++ b/addons/product/migrations/12.0.1.2/end-migration.py
@@ -1,0 +1,32 @@
+# Copyright 2019 Eficent <http://www.eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def update_product_template_attribute_value(env):
+    # in v12, for each relation of the product.attribute.value and
+    # product.template.attribute.line, a product.template.attribute.value
+    # should exist. Thus, we determine which lines still doesn't have a
+    # product.template.attribute.value and then we generate them.
+    openupgrade.logged_query(
+        env.cr, """
+        SELECT rel.product_attribute_value_id, ptal.product_tmpl_id
+        FROM product_template_attribute_line ptal
+        JOIN product_attribute_value_product_template_attribute_line_rel
+            rel ON rel.product_template_attribute_line_id = ptal.id
+        LEFT JOIN product_template_attribute_value ptav ON (
+            rel.product_attribute_value_id = ptav.product_attribute_value_id
+            AND ptal.product_tmpl_id = ptav.product_tmpl_id)
+        WHERE ptav.id IS NULL
+        """
+    )
+    env['product.template.attribute.value'].create([{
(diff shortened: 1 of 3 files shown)


addons/product/migrations/12.0.1.2/pre-migration.py
```diff
@@ -24,13 +27,27 @@
 _table_renames = [
     ('product_attribute_line', 'product_template_attribute_line'),
     ('product_attribute_price', 'product_template_attribute_value'),
+    ('product_attribute_line_product_attribute_value_rel',
+     'product_attribute_value_product_template_attribute_line_rel'),
 ]
 
 
+def avoid_new_constraint_in_product_template_attribute_line(cr):
+    openupgrade.logged_query(
```
pedrobaeza, line 37: Why this?

MiquelRForgeFlow: In v11 exists this contrain:
```python
        if any(line.value_ids > line.attribute_id.value_ids for line in self):
            raise ValidationError(_('Error ! You cannot use this attribute with the following value.'))
```
In v12, it has changed to:
```python
        if any(not line.value_ids or line.value_ids > line.attribute_id.value_ids for line in self):
            raise ValidationError(_('You cannot use this attribute with the following value.'))
```

It means that you can only create template attribute lines if and only if they are linked to an attribute value or more. You need to drop those lines that are not linked to any attribute value.

pedrobaeza: Yeah, but you are destructing information with this, so we shouldn't unless it blocks something. In this case, I see it's a Python constraint, so it's not too much problem to let it.

MiquelRForgeFlow: You are not destructing "information" in this case, it's just cleaning unused and useless data. Lines without an attribute value are pointless, that's why odoo added that new constrain.

pedrobaeza: OK, got it

addons/product/migrations/12.0.1.2/end-migration.py
```diff
@@ -0,0 +1,17 @@
...
+
+from openupgradelib import openupgrade
+
+
+def update_product_template_attribute_value(env):
+    # the product.template.attribute.values are recomputed each time a
+    # template.attribute.line changes. The conditions to generate this
+    # template_attribute_value have changed, so we need to recompute them.
+    lines = env['product.template.attribute.line'].search([])
+    lines._update_product_template_attribute_values()
```
pedrobaeza: Can we reconstruct only when needed (when different conditions are met)?

MiquelRForgeFlow: I will check if we can filter them to just the needed cases.

MiquelRForgeFlow: filtered!

pedrobaeza (approved):

PR #1887, opened 2019-06-14
[12.0][MIG] sale_expense: Nothing to do

OthmaneGhandi:
2 fields was added that doesn't impact the migration + some xml views that will be loaded with the module

=> Nothing to do

--


MiquelRForgeFlow:
@OthmaneGhandi please, could you rebase?

MiquelRForgeFlow (approved):
LGTM

pedrobaeza:
@mreficent why rebase is needed?

MiquelRForgeFlow:
Well, it was an wrong comment, after checking the code. It's not needed, that's why approved it.

OthmaneGhandi:
> Well, it was an wrong comment, after checking the code. It's not needed, that's why approved it.

That's good

pedrobaeza (approved):

pedrobaeza:
Thanks for the contribution, @OthmaneGhandi. Good to see new blood in this difficult part of this world. If you can share your experience introducing in it and the found troubles, maybe we can try to improve documentation in that part.

OthmaneGhandi:
@pedrobaeza Thanks for your welcome and your huge effort in the community.
For me I think I would prefer to explain more the use of each migration files (pre-migration, post-migration, end-migration and openupgrade_analysis_work.xml), maybe using a practical example (we all understand very well with examples :smile: )

Beside the documentation, I tried to pick small standard modules and see how the migration was done. Step by step (with real tasks) I will learn more in deep how to use it

pedrobaeza:
OK, we'll see if we can put all together and find time for that (the main problem).

PR #1896, opened 2019-06-21
[12.0][MIG] Avoid 'account.move.line.reconcile' error

OthmaneGhandi:
Current behavior before PR:
You got an error when trying to open Journal Items, related to "'account.move.line.reconcile'" action view
=> The action view should be deleted et the end of the migration 

--


pedroslopez (commented):
Could we use openupgradelib's `delete_records_safely_by_xml_id` here?

OthmaneGhandi:
> `delete_records_safely_by_xml_id`

In this case we should include it in pre-imgration file, because after reloading the modules we lose the xml id of the record. I will try it too

pedrobaeza:
Thanks for discovering the problem, but the patch is not correct, as removing once the XML-ID has been changed deletes the incorrect record. I have pushed the proper fixed in 0452773.

OthmaneGhandi:
Thanks @pedrobaeza

Issue #1204, opened 2017-11-24
Default is only mentioned for required fields


StefanRijnhart:
It will be helpful if the analysis shows the default value for non-required fields as well.

MiquelRForgeFlow:
Then we should debate here if this is a thing we really want. I also think approach does not improve readability, and the value of req_required is mostly just 'function' which is not very informative.

pedrobaeza:
What is my reasoning for needing it? To know when a new field has a default value (not which one), for knowing if you have to check if the default value is correct.

It's true that you have to do similar if no default value is provided for knowing if the business logic should provide a value to field there, but it seems more straight-forward.

What do you think?

MiquelRForgeFlow:
And how many cases are there in which the default value is not correct? 🤔

pedrobaeza:
This was raised from one of these cases, and I remember you were involved also.

MiquelRForgeFlow:
I don't remember 😓

MiquelRForgeFlow:
@pedrobaeza Ok, I updated the PR with a different approach. Now it only shows 'hasdefault' string for those NEW records that have a default. I think this way it's less verbose and more readable than the previous approach. I think that with this way both point of views should be happy.

Issue #213, opened 2015-01-09
Analysis does not detect renamed many2many tables


StefanRijnhart:
Example: property_ids on mrp.bom changed its table from mrp_bom_property_rel to mrp_bom_mrp_property_rel in 8.0

pedrobaeza:
Is this solved?

StefanRijnhart:
no

pedrobaeza:
@mreficent did you handle this already with your improvements?

MiquelRForgeFlow:
I would say no. Should I try to do it?

pedrobaeza:
And do you see it feasible?

MiquelRForgeFlow:
Feasible and done :)

PR #1906, opened 2019-07-01
[12.0][IMP] openupgrade_records: defaults and many2manys tables

MiquelRForgeFlow:
Implements https://github.com/OCA/OpenUpgrade/issues/1204 and https://github.com/OCA/OpenUpgrade/issues/213.

Needs https://github.com/OCA/OpenUpgrade/pull/1907 to be merged.

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -100,7 +100,7 @@ def report_generic(new, old, attrs, reprs):
             if old[attr] != new['required'] and new['required']:
                 text = "now required"
                 if new['req_default']:
-                    text += ', default = %s' % new['req_default']
+                    text += ', req_default: %s' % new['req_default']
                 fieldprint(old, new, '', text, reprs)
         elif attr == 'stored':
             if old[attr] != new[attr]:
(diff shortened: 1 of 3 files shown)


StefanRijnhart:
Great! Can you create a separate branch with the resulting analysis between 11 and 12 with this patch? It would help to review this code.

MiquelRForgeFlow:
@StefanRijnhart done in https://github.com/OCA/OpenUpgrade/pull/1910.

StefanRijnhart:
Nice. I like the side effect that it shows the target model of many2one fields!

What I don't like is that it shows the default values when the field is not required. Back in the day, I chose explicitely only to show the default value for required fields only, because that is where it matters (if there is a sane default value for a new required field, no migration is necessary). Would you consider restoring this behaviour? (I think the new behaviour does not improve readability, and the value of req_required is mostly just 'function' which is not very informative)

MiquelRForgeFlow:
Ok, but then you should close https://github.com/OCA/OpenUpgrade/issues/1204

pedrobaeza:
Stefan, this was the result of a discussion in a pr for knowing if a new field is providing a default value for acting according that, but unfortunately, there's no link to it. I still think it's valuable.

StefanRijnhart:
Haha I opened an issue for that :-P

@mreficent I'll leave it to your own judgment then.

StefanRijnhart (approved):

pedrobaeza (approved):

PR #1913, opened 2019-07-04
[12.0][MIG] fleet

astirpe:
--


addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,161 @@
+---Fields in module 'fleet'---
+fleet        / fleet.vehicle            / activity_ids (one2many)       : NEW relation: mail.activity
+fleet        / fleet.vehicle            / brand_id (many2one)           : NEW relation: fleet.vehicle.model.brand
+fleet        / fleet.vehicle            / first_contract_date (date)    : NEW
```
pedrobaeza, line 8: This is going to be initialized with the day you do the migration, which seems weird. Better to empty it on post-migration

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,161 @@
...
+fleet        / fleet.vehicle            / message_main_attachment_id (many2one): NEW relation: ir.attachment
+fleet        / fleet.vehicle.assignation.log / date_end (date)               : NEW
+fleet        / fleet.vehicle.assignation.log / date_start (date)             : NEW
+fleet        / fleet.vehicle.assignation.log / driver_id (many2one)          : NEW relation: res.partner, required: required
+fleet        / fleet.vehicle.assignation.log / vehicle_id (many2one)         : NEW relation: fleet.vehicle, required: required
+fleet        / fleet.vehicle.log.contract / activity_ids (one2many)       : NEW relation: mail.activity
+fleet        / fleet.vehicle.log.contract / message_last_post (datetime)  : DEL
+fleet        / fleet.vehicle.log.contract / message_main_attachment_id (many2one): NEW relation: ir.attachment
+fleet        / fleet.vehicle.log.contract / user_id (many2one)            : NEW relation: res.users
+# NOTHING TO DO
```
pedrobaeza: It's better to unfold each "NOTHING TO DO" explaining why it's a nothing to do, for future references or for reviewers. You can see an example of these explanations in: https://github.com/OCA/OpenUpgrade/pull/1921/files#diff-66926c746e076b2230abbaa064992a56

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,161 @@
...
+fleet        / fleet.vehicle            / message_last_post (datetime)  : DEL
+fleet        / fleet.vehicle            / message_main_attachment_id (many2one): NEW relation: ir.attachment
+fleet        / fleet.vehicle.assignation.log / date_end (date)               : NEW
+fleet        / fleet.vehicle.assignation.log / date_start (date)             : NEW
+fleet        / fleet.vehicle.assignation.log / driver_id (many2one)          : NEW relation: res.partner, required: required
+fleet        / fleet.vehicle.assignation.log / vehicle_id (many2one)         : NEW relation: fleet.vehicle, required: required
+fleet        / fleet.vehicle.log.contract / activity_ids (one2many)       : NEW relation: mail.activity
+fleet        / fleet.vehicle.log.contract / message_last_post (datetime)  : DEL
+fleet        / fleet.vehicle.log.contract / message_main_attachment_id (many2one): NEW relation: ir.attachment
+fleet        / fleet.vehicle.log.contract / user_id (many2one)            : NEW relation: res.users
```
pedrobaeza, line 35: This field will be initialized to admin. Better to empty it on post-migration

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,161 @@
...
+DEL fleet.vehicle.model: fleet.model_serie6
+DEL fleet.vehicle.model: fleet.model_serie7
+DEL fleet.vehicle.model: fleet.model_seriehybrid
+DEL fleet.vehicle.model: fleet.model_seriem
+DEL fleet.vehicle.model: fleet.model_seriex
+DEL fleet.vehicle.model: fleet.model_seriez4
+DEL fleet.vehicle.model: fleet.model_tt
+DEL fleet.vehicle.model: fleet.model_zafira
+DEL fleet.vehicle.model: fleet.model_zafiratourer
+# Keep those data, do not remove it
```
pedrobaeza: These are supposed to be demo data, but maybe you are using it in your previous database, so they will fail when being auto-removed. With your approach, all of them are kept, but maybe they haven't been used and it's provoking noise, so I suggest to program a loop that goes through each XML-ID and:

* Try to remove it. You can use openupgradelib's `delete_records_safely_by_xml_id`.
* If not able, remove the ir.model.data entry.

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,161 @@
...
+
+DEL fleet.vehicle.tag: fleet.vehicle_tag_break
+DEL fleet.vehicle.tag: fleet.vehicle_tag_compact
+DEL fleet.vehicle.tag: fleet.vehicle_tag_convertible
+DEL fleet.vehicle.tag: fleet.vehicle_tag_junior
+DEL fleet.vehicle.tag: fleet.vehicle_tag_leasing
+DEL fleet.vehicle.tag: fleet.vehicle_tag_purchased
+DEL fleet.vehicle.tag: fleet.vehicle_tag_sedan
+DEL fleet.vehicle.tag: fleet.vehicle_tag_senior
+# Keep those data, do not remove it
```
pedrobaeza: Why don't you group all the entries by logic operation?

pedrobaeza (changes requested):
Can you give me credits as contributor for reviewing deeply the scripts?

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
astirpe: Yes, my customer were using those values.
Technically I could do what you are suggesting, yes. But I don't think is a correct approach: here we are migrating data and those data were existing in the customer's database for years.
Why do you think the unused values are provoking noise? Just because Odoo moved them to demo data? Doesn't sound a good reason to me.

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
astirpe: What do you mean?

pedrobaeza: In the work file, cut these lines and paste them in the same block than the rest with the same comment (`# Keep those data, do not remove it`)

addons/fleet/migrations/12.0.0.1/openupgrade_analysis_work.txt
pedrobaeza: Yes, it is a good reason. OpenUpgrade should preserve at maximum previous data, but trying to be in line with new features / data layout. That's why you should try to remove them, and if not possible, then keep them.

astirpe: > trying to be in line with new features / data layout

This is a better reason :+1:

astirpe: Pedro, I have some problems when using the `delete_records_safely_by_xml_id` in the pre-migration script. It seems failing because the fleet models are not yet loaded in the registry at the moment the function executed. Not sure if I'm doing something wrong.
I would keep the scripts as they are, since I don't have time to continue to develop them. They are working fine by the way.

pedrobaeza: Yeah, they should be done in post-migration. You can do:

- Mark them as noupdate="1", so they are not removed.
- Try to remove them on post. If not possible, remove the ir.model.data record.

astirpe: With `noupdate="1" ` and moving the scripts in post-migration it works better, but still the `delete_records_safely_by_xml_id` raises a not-null exception on the `model_id` field. I could silent the exception and continue to add tricks for going further. But I'm still not fully convinced of this approach.

For example, what about any eventual custom module that adds a not-mandatory many2one field to `fleet.vehicle.model`? In case the to-be-deleted records are used there, no raise will be performed and that information will be lost. 

I prefer to keep the scripts as they are. If someone really wants to remove those records, I would say: go ahead and close this PR.

pedrobaeza: OK, let's move on and let it this way, as always manual deletion can be done.

addons/fleet/migrations/12.0.0.1/pre-migration.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+        """
+    )
+
+
+def keep_old_data_fleet_service_type(cr):
+    openupgrade.logged_query(
+        cr, """
+            DELETE FROM ir_model_data
+            WHERE module like 'fleet'
```
pedrobaeza: Why are you using `like` operator instead of `=`?

astirpe: For no particular reason. I just changed it.

pedrobaeza (approved):

PR #1915, opened 2019-07-09
[12.0][IMP] mail: fast fill add_sign in mail.message

MiquelRForgeFlow:
If you have a lot of messages, filling this new column `add_sign` take much time.

--

### DIFF
--- a/addons/mail/migrations/12.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/mail/migrations/12.0.1.0/openupgrade_analysis_work.txt
@@ -46,6 +46,8 @@ mail         / mail.channel             / moderator_ids (many2many)     : NEW re
 # NOTHING TO DO
 
 mail         / mail.message             / add_sign (boolean)            : NEW
+# DONE: pre-migration: filled this new field using sql query for faster migration
+
 mail         / mail.message             / layout (char)                 : NEW 
 mail         / mail.message             / moderation_status (selection) : NEW selection_keys: ['accepted', 'pending_moderation', 'rejected']
 mail         / mail.message             / moderator_id (many2one)       : NEW relation: res.users
--- a/addons/mail/migrations/12.0.1.0/pre-migration.py
+++ b/addons/mail/migrations/12.0.1.0/pre-migration.py
@@ -14,8 +14,23 @@ _field_renames = [
 ]
 
 
+def fill_mail_message_add_sign_default(env):
+    """Faster way to fill this new field"""
+    openupgrade.logged_query(
+        env.cr, """
+        ALTER TABLE mail_message
+        ADD COLUMN add_sign boolean
+        DEFAULT TRUE""",
+    )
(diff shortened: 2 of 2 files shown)


AaronHForgeFlow:
always True?

MiquelRForgeFlow:
> always True?

By default, it's always True.

MiquelRForgeFlow:
I don't know if this PR provides a faster way. For example, in a database I used, the query lasted **19 minutes** to finally show the `OpenUpgrade: 8397417 rows affected` message.

pedrobaeza:
What about trying to put on pre- the `ALTER TABLE ... ADD...` with default value to see if it's faster?

MiquelRForgeFlow:
Thanks, it solved the problem that way. It was nearly an instant :)

addons/mail/migrations/12.0.1.0/pre-migration.py
```diff
@@ -14,8 +14,20 @@
 ]
 
 
+def fill_mail_message_add_sign_default(env):
+    """Fast way to fill this new field"""
+    if not openupgrade.column_exists(env.cr, 'mail_message', 'add_sign'):
```
pedrobaeza: Why this condition?

addons/mail/migrations/12.0.1.0/pre-migration.py
```diff
@@ -14,8 +14,20 @@
...
+        openupgrade.logged_query(
+            env.cr, """
+            ALTER TABLE mail_message
+            ADD COLUMN add_sign boolean
+            DEFAULT TRUE""",
```
pedrobaeza: Doubt I have now is this default will act for new additions. I think you must remove that default after this one.

pedrobaeza (changes requested):

addons/mail/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -46,6 +46,8 @@ mail         / mail.channel             / moderator_ids (many2many)     : NEW re
 # NOTHING TO DO
 
 mail         / mail.message             / add_sign (boolean)            : NEW
+# DONE: pre-migration: filled this new field using sql query for fast migration
```
pedrobaeza: s/fast/faster

addons/mail/migrations/12.0.1.0/pre-migration.py
MiquelRForgeFlow: how?

pedrobaeza: https://popsql.com/learn-sql/postgresql/how-to-remove-a-default-value-from-a-column-in-postgresql/

MiquelRForgeFlow: thanks, done

pedrobaeza (approved):

PR #1918, opened 2019-07-10
[12.0][FIX] hr_holidays: don't create wrong followers

MiquelRForgeFlow:
If you have a record A with two followers (partners 1 and 2), and you have record B with two followers (partners 2 and 3), then if you do
```python
Followers._insert_followers(
    model, (A+B).ids,
    [1,2,3],
    dict((pid, subtype_ids) for pid in [1,2,3]),
    [], False, check_existing=True, existing_policy='update',
)
```
it will create a follower of partner 3 following record A and a follower of partner 1 following record B. And this is wrong!

--

### DIFF
--- a/addons/hr_holidays/migrations/12.0.1.5/post-migration.py
+++ b/addons/hr_holidays/migrations/12.0.1.5/post-migration.py
@@ -140,32 +140,33 @@ def subscribe_new_subtypes(env):
     """
     subtype_mapping = [
         ('mt_department_holidays_approved',
-         'mt_department_leave_allocation_approved',
-         'mt_department_leave_approved'),
+         ['mt_department_leave_allocation_approved',
+          'mt_department_leave_approved']),
         ('mt_department_holidays_refused',
-         'mt_department_leave_allocation_refused',
-         'mt_department_leave_refused'),
+         ['mt_department_leave_allocation_refused',
+          'mt_department_leave_refused']),
         ('mt_holidays_approved',
-         'mt_leave_allocation_approved',
-         'mt_leave_approved'),
+         ['mt_leave_allocation_approved',
+          'mt_leave_approved']),
         ('mt_holidays_refused',
-         'mt_leave_allocation_refused',
-         'mt_leave_refused'),
+         ['mt_leave_allocation_refused',
+          'mt_leave_refused']),
     ]
-    Followers = env['mail.followers']
-    for old_id, new_id1, new_id2 in subtype_mapping:
-        old_mt = env.ref('hr_holidays.' + old_id)
-        followers = Followers.search([('subtype_ids', '=', old_mt.id)])
-        if followers:
-            new_mt1 = env.ref('hr_holidays.' + new_id1)
-            new_mt2 = env.ref('hr_holidays.' + new_id2)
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
Makes sense :+1:

MiquelRForgeFlow:
@pedrobaeza please review

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -152,21 +152,49 @@ def subscribe_new_subtypes(env):
...
-            )
+    old_ids, new1_ids, new2_ids = [], [], []
+    for old_id, new1_id, new2_id in subtype_mapping:
+        old_ids.append(env.ref('hr_holidays.' + old_id).id)
+        new1_ids.append(env.ref('hr_holidays.' + new1_id).id)
+        new2_ids.append(env.ref('hr_holidays.' + new2_id).id)
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE mail_followers_mail_message_subtype_rel rel
+        SET mail_message_subtype_id = CASE
```
pedrobaeza: Instead of doing this ugly `CASE` that at the end is like executing 4 queries, please do one query per loop simplifying the query

MiquelRForgeFlow: done

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -152,21 +152,36 @@ def subscribe_new_subtypes(env):
...
-                partner_ids,
-                dict((pid, subtype_ids) for pid in partner_ids),
-                [], False, check_existing=True, existing_policy='update',
-            )
+    old_ids, new1_ids, new2_ids = [], [], []
+    for old_id, new1_id, new2_id in subtype_mapping:
+        old_ids.append(env.ref('hr_holidays.' + old_id).id)
+        new1_ids.append(env.ref('hr_holidays.' + new1_id).id)
+        new2_ids.append(env.ref('hr_holidays.' + new2_id).id)
+    for i in range(3):
```
pedrobaeza: Why are you still doing 2 loops with that weird index positioning?

pedrobaeza: Even more... don't do the same SQL (equal except models and ids). Create 2 nested loops that put in the loop variables proper values and just write one SQL.

MiquelRForgeFlow: now there is only 1 loop!

pedrobaeza: OK, but see my second comment: you can save one large SQL doing a nested loop including the model in it as variable

MiquelRForgeFlow: ok, done. superclean!

pedrobaeza: Great, that's the idea! Isn't better to read this code? :smiley:

pedrobaeza (approved):

PR #1921, opened 2019-07-11
[ADD] website_event_track: Migration

yajo:
The only noteworthy thing is the transformation of `website_track` and `website_track_proposal` into stored fields that now behave differently.

--
@Tecnativa


addons/website_event_track/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2019 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import re
+from openupgradelib import openupgrade
+
+
+def _website_track_menu(env):
+    """Transform the tracks menu to new scheme."""
+    events_with_menu = env["event.event"].search([("menu_id", "!=", False)])
```
MiquelRForgeFlow: `with_context(active_test=False)`

addons/website_event_track/migrations/12.0.1.0/post-migration.py
```diff
@@ -0,0 +1,38 @@
...
+    for event in events_with_menu:
+        # In v11, these fields were True if their menus existed
+        website_track = website_track_proposal = False
+        for menu in event.menu_id.child_id:
+            if re.fullmatch(r"/event/.*/track", menu.url):
+                website_track = True
+            elif re.fullmatch(r"/event/.*/track_proposal", menu.url):
+                website_track_proposal = True
+        # Update the flags according to menus presence; this will trigger
+        # the creation of missing website.event.menu records
```
pedrobaeza: This is also creating `website.menu` entries, which should be already created in v11, isn't it? I'm also not sure about the URL being exactly the same.

pedrobaeza (changes requested):
You can put me as contributor, being checked deeply the internals for reviewing this

addons/website_event_track/migrations/12.0.1.0/post-migration.py
yajo: Yes, the URL is the same.

yajo (commented):
All attended.

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (approved):

PR #1923, opened 2019-07-11
[FIX][12.0] Migration of homepage

katyukha:
_Description of the issue/feature this PR addresses:_

Module: website
Homepage was not migrated correctly. Default homepage was shown after migration.

_Current behavior before PR:_

During migration following pages updated with default data:
- Home
- About us
- may be some other

_Desired behavior after PR is merged:_

- Do not owverwrite these pages

--

### DIFF
--- a/addons/website/migrations/12.0.1.0/pre-migration.py
+++ b/addons/website/migrations/12.0.1.0/pre-migration.py
@@ -77,7 +77,7 @@ def bootstrap_4_migration(env):
         ("report_type", "=like", "qweb-%"),
     ]).mapped("report_name")
     # Find updatable views, to be excluded; standard addon update is enough
-    udpatable_ids = env["ir.model.data"].search([
+    updatable_ids = env["ir.model.data"].search([
         ("model", "=", "ir.ui.view"),
         ("noupdate", "=", False),
     ]).mapped("res_id")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
About us is update on post on purpose, but homepage shouldn't be. And how this change prevents this?

katyukha:
It seems that these views becomes updatable, and standard module update overwrite them.
In case they have modifications (for example changes on home page), view updated with default Odoo data.

Before this PR, these views was not touched by `bootstrap_4_migration` function (in pre-migration) and overwritten with default Odoo data before post-migration.

I am not sure if this correct way to fix this bechavior(

pedrobaeza:
@Yajo will check it

katyukha:
Thanks)

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -250,10 +250,10 @@ def migrate(env, version):
...
     openupgrade.rename_fields(env, _field_renames)
     openupgrade.rename_xmlids(cr, _xml_ids_renames)
-    noupdate_changes(env)
     fill_website_company_id(cr)
     fill_website_name(cr)
     fill_website_redirect_type(cr)
     fill_website_redirect_urls(cr)
     disable_less_customizations(env)
     bootstrap_4_migration(env)
+    noupdate_changes(env)
```
yajo, line 259: This was before the BS4 migration to avoid migrating pages that were going to get updated anyways. Please let it where it was before.

The real fix you're looking for is to [remove `homepage` from here](https://github.com/OCA/OpenUpgrade/pull/1923/files#diff-7f69b365074a905b3170de8b9b760af1R55).

yajo (changes requested):

PR #1924, opened 2019-07-11
[12.0][FIX] stock: push_pull rules broccoly

### LINKED ISSUE #1919: [12.0] stock: migration of stock.rule
https://github.com/OCA/OpenUpgrade/blob/7a0eba613ac70809847f1797e1e5458134e0e73f/addons/stock/migrations/12.0.1.1/post-migration.py#L13

@pedrobaeza @mreficent Are you sure about this?, in most of the cases the rules will be just pull, and by adding them the push component you might be introducing unwanted behaviors. You can start to anticipate that with this example:

1. Take standard stock -> customers rule (which in a new v12 db is just pull) and change it to pull n push, which would be the result in a v11 db migrated with this script (not tested but that's what I undertand here):
image

2. Then just receive products into stock:
image

3. And here you have a automatic  (and unwanted) delivery to customers
image

Please, correct me if I did any incorrect assumption. 

Thanks!

MiquelRForgeFlow:
Rules coming from `procurement.rule` (with old action='move') became 'pull'
Rules coming from `stock.location.path` became 'push'
Rules coming from from both became 'pull_push'

Fixes https://github.com/OCA/OpenUpgrade/issues/1919.

--

### DIFF
--- a/addons/stock/migrations/12.0.1.1/post-migration.py
+++ b/addons/stock/migrations/12.0.1.1/post-migration.py
@@ -10,7 +10,7 @@ def map_stock_rule_action(cr):
         cr,
         openupgrade.get_legacy_name('action'),
         'action',
-        [('move', 'pull_push'),
+        [('move', 'pull'),
          ],
         table='stock_rule', write='sql')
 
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/12.0.1.1/post-migration.py
```diff
@@ -101,6 +101,14 @@ def merge_stock_location_path_stock_rule(env):
             [row[1]],
             row[0],
         )
+    pull_push_rule_ids = list(set([r[0] for r in rules_to_merge]))
```
pedrobaeza:
```suggestion
    pull_push_rule_ids = tuple(set([r[0] for r in rules_to_merge]))
```
and remove the other conversion

MiquelRForgeFlow: done

pedrobaeza (approved):

Issue #1783, opened 2019-03-20
v7 -> v8 state error in stock/post-migration.py


bizslink-calon:
```
def migrate_stock_qty(cr, registry):
    """Reprocess stock moves in done state to fill stock.quant."""
    # First set restrict_lot_id so that quants point to correct moves
    with api.Environment.manage():
        env = api.Environment(cr, SUPERUSER_ID, {})
        moves = env['stock.move'].search(
            [('state', 'in', ['assign', 'done'])], order="date")
        for move in moves:
            if move.state == 'assign':
                _move_assign(env, move)
            else:
                _move_done(env, move)
```

error state 'assign', maybe assigned?

pedrobaeza:
Which is the error?

bizslink-calon:
> Which is the error?

'assign' is not in stock_move.state's selection instead of 'assigned'.

pedrobaeza:
You're totally right! Fixed in 8cd47363f42cd03e910caca03bf16ebbee45ee75

bizslink-calon:
> You're totally right! Fixed in [8cd4736](https://github.com/OCA/OpenUpgrade/commit/8cd47363f42cd03e910caca03bf16ebbee45ee75)

You may rewrite this method, somewhere still used 'assign'.

pedrobaeza:
OK, the oversight has been fixed in 6f465f34e91940b3075fb2c1cf99b3a8b28677da

PR #1932, opened 2019-07-16
[8.0] stock: process all done moves before assigned moves

alexis-via:
I'm in the process of migrating a DB with lots of stock moves (about 160 000) from v7 to v12.
I made a test migration in April 2019, and migration from v7 to v8 was working fine.
I made the "real" migration this week-end with up-to-date code (to get the latest fixes/improvements): I started v7 to v8 migration on Friday evening, but it was still running on Monday morning after 55h of processing... so I had to stop it and post-pone the migration to another day. Looking at the logs, it was inside the method migrate_stock_qty() (https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L922) for about 54h and was still in there when I stopped it.
As this migration from v7 to v8 was working fine in April 2019, I had a look at the commits in between. One change from @pedrobaeza caught my attention: this commit https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L922 which was the consequence of this bug report https://github.com/OCA/OpenUpgrade/issues/1783
It seems like a very minor and obvious fix. But it seems strange to find such an obvious fix in v7 to v8 migration in 2019, such a long time after v8 release !
My idea was that I may be impacted by an infinite loop caused by this change ; so I added logs in the methods _move_assign() and _move_done() to confirm this.
But, when reading the source code of this 2 methods, I felt that there may be a logic error in the code: I think that we must process all done moves BEFORE processing the assigned moves, because the assign moves reserve quants generated by the processing of done moves. And, with the current code, done and assigned moves are processed "at the same time" (only ordered by "date"). With the code before the fix introduced by @pedrobaeza on April 4, the _move_assigned() method was not called because of the syntax error ('assign' instead of 'assigned').
I launched a new test migration yesterday afternoon from v7 to v8 (on the same DB as my failed migration of this WE) with the code change of this PR plus some logs. I was expecting to "see" the infinite loop, but it successfully ran until the end after 11h of processing. My interpretation of this is that my code change fixed the infinite loop... but it may be a bit optimistic !

--

### DIFF
--- a/addons/stock/migrations/8.0.1.1/post-migration.py
+++ b/addons/stock/migrations/8.0.1.1/post-migration.py
@@ -929,13 +929,14 @@ def migrate_stock_qty(cr, registry):
 
     with api.Environment.manage():
         env = api.Environment(cr, SUPERUSER_ID, {})
-        moves = env['stock.move'].search(
-            [('state', 'in', ['assigned', 'done'])], order="date")
-        for move in moves:
-            if move.state == 'assigned':
-                _move_assign(env, move)
-            else:
-                _move_done(env, move)
+        done_moves = env['stock.move'].search(
+            [('state', '=', 'done')], order="date")
+        for done_move in done_moves:
+            _move_done(env, done_move)
+        assigned_moves = env['stock.move'].search(
+            [('state', '=', 'assigned')], order="date")
+        for assigned_move in assigned_moves:
+            _move_assign(env, assigned_move)
 
 
 def migrate_stock_production_lot(cr, registry):


alexis-via:
Another thing that I wanted to mention: the _move_assigned() method (https://github.com/OCA/OpenUpgrade/blob/8.0/addons/stock/migrations/8.0.1.1/post-migration.py#L759) takes "move" as argument, but the code of this method uses a "move" variable inside 2 loops (cf line 814 and 825), which is very dangerous: when reading the code, you have to pay attention to the fact that after line 814, "move" is not what has been passed as argument to the method but a temporary variable !!! This code has been written by @jbeficent in this PR https://github.com/OCA/OpenUpgrade/pull/469/commits We should probably change this, shouldn't we ?

MiquelRForgeFlow (approved):
LGTM (it seems logic to me after reading the explanation and seeing the code changes)

pedrobaeza:
I think the idea to mix both kind of moves is for assigning the stock following the proper quant as if the moves were reserved in that moment. Imagine you have a FEFO policy, with these moves:

* 2018-01-01 - IN: 10 units - Expiration date: 30/01/2018
* 2018-01-02 - OUT: 2 units. In the system, the units that have been reserved are the existing ones.
* 2018-01-03 - IN: 2 units - Expiration date: 20/01/2018

With your change, the system will reserve the 2 units from the reception of the 3rd of January instead of the others.

Said that, is this a problem? v7 is too far now for me to say if there the reservation concept goes the same way, but the main principle in OpenUpgrade is to not lose information that was specified in previous version. I don't remember if the reservation concept was present in v7. If not, then I don't see any special problem.

MiquelRForgeFlow:
Well, maybe we could add a previous search to see in which cases it's better to mix the moves and which cases is better to have the separated. Then, split them, and execute the mix method for the mixed case and then the separated method for the separated case.

pedrobaeza:
@alexis-via are you going to comment anything about this or can I close it?

pedrobaeza:
Better option without modifying the flow in #2080, so closing this one.

PR #1933, opened 2019-07-16
[FIX] website: Fix menu & page structure migration

yajo:
- Do not remove the views that have been COW'd in the pre-migration. An odoo code patch needed for that.
- The migration field is now TEXT and stores a JSON with menu info.
- Create pages in pre-migration, since in post-migration they might have disappeared and, after some tests, we were gonna need them anyway.
- Update/recreate website-specific menu items correctly.
- Fix https://github.com/OCA/OpenUpgrade/pull/1923 by respecting `noupdate=1` status of homepage record until the end of the pre-migration. This way, the BS4 migration and website-specification process work as expected.

--
@Tecnativa

### DIFF
--- a/addons/website/migrations/12.0.1.0/post-migration.py
+++ b/addons/website/migrations/12.0.1.0/post-migration.py
@@ -1,6 +1,8 @@
 # Copyright 2018-19 Eficent <http://www.eficent.com>
 # Copyright 2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import json
+
 from psycopg2.extensions import AsIs
 
 from openupgradelib import openupgrade
@@ -47,13 +49,13 @@ def sync_menu_views_pages_websites(env):
     # Main menu and children must be website-agnostic
     main_menu = env.ref('website.main_menu')
     child_menus = env["website.menu"].search([
-        ("id", "child_of", main_menu.id),
+        ("id", "child_of", main_menu.ids),
         ("website_id", "!=", False),
     ])
     child_menus.write({"website_id": False})
-    # Duplicate the main menu for main website
-    website = env["website"].get_current_website()
-    website.copy_menu_hierarchy(main_menu)
+    # Duplicate the main menu for all websites
+    for website in env["website"].search([]):
+        website.copy_menu_hierarchy(main_menu)
     # Find views that were website-specified in pre stage
     col_name = openupgrade.get_legacy_name("bs4_migrated_from")
     env.cr.execute(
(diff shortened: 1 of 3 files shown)


addons/website/models/ir_ui_view.py
```diff
@@ -160,7 +160,14 @@ def unlink(self):
...
         specific_views = self.env['ir.ui.view']
-        if self and self.pool._init:
+
+        # OpenUpgrade: force a False condition to never enter this code block.
+        # This is needed because the update process could correctly unlink
+        # website-agnostic views, but we don't want the website-specific
+        # equivalents to be removed during the migration.
+        if False and self and self.pool._init:
+        # OpenUpgrade: end of modification
+
```
MiquelRForgeFlow, line 170: Why not remove this chunk of code instead of doing this "if False"? 🤔

MiquelRForgeFlow (commented):
Too much changes. Really?

And why that many searches? Cannot we reduce the number of searches and apply more queries instead?

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -64,44 +66,49 @@ def sync_menu_views_pages_websites(env):
...
+                ("url", "=", page.url),
+            ])
+            # If menus exist, it means the agnostic view wasn't removed and
+            # they already contain all the needed information, except that they
+            # are linked to the website-agnostic page. Let's fix that.
+            if menus:
+                menus.write({
+                    "page_id": page.id,
                 })
+            # In case the menus disappeared, it's possibly because the
```
pedrobaeza, line 84: Can we avoid this removal/creation? All of this can affect subsequent modules that extends that records.

addons/website/models/ir_ui_view.py
pedrobaeza: It's done on purpose for knowing exactly what have been changed. Removing code you can only know what happens seeing commit diff.

addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -64,44 +66,49 @@ def sync_menu_views_pages_websites(env):
...
-                ("website_id", "=", False),
-                ("url", "=", specific_page.url),
-            ], limit=1)
-            if agnostic_menu:
-                agnostic_menu.copy({
-                    "website_id": specific_page.website_id.id,
-                    "page_id": agnostic_menu.page_id.id and specific_page.id,
+        for page in specific_view.page_ids:
+            menus = env["website.menu"].search([
+                ("id", "child_of", page.website_id.menu_id),
```
katyukha: if think `.id` is missing here: 
```python
("id", "child_of", page.website_id.menu_id.id),
```

addons/website/migrations/12.0.1.0/post-migration.py
yajo: It's complex.

The problem is that, if a view is removed because it's website-agnostic and not declared by an addon, the cascade FK constraints attached to it will remove the `website.page` records associated to it, which in turn will remove the `website.menu` records attached to it.

~~What we could do is to empty those fields that would produce the cascading effect and store their values in extra columns in the pre-migration. Then, in the post-migration, delete the records if the pointer records are missing and re-link otherwise.~~

:point_up: While I was writing that I realized that it's absurd. What we're doing here is to actually create new website-specific records from the website-agnostic records that existed before. The website-agnostic record removal is just a side effect of something that is no longer gonna be used. The way odoo would work if the user did manual changes would be that. Thus, the relations within the addon would suffer from that same effect in case of extensions.

I could, however, do a `SELECT * FROM website_menu` in the pre-migration to get all fields, including any extended ones, insert the new values from SQL in the post-migration removing the conflicting fields such as `id`. What do you think?

addons/website/models/ir_ui_view.py
yajo: Yes, if we were patching Odoo I'd never do it like this, but I think that this way the diff is reduced so much but it gets obvious that this is a HACK for OpenUpgrade only. Just in case a reader happens to pass through here.

yajo (commented):
> Too much changes. Really?

Well... yes :roll_eyes: 

> And why that many searches? Cannot we reduce the number of searches and apply more queries instead?

The searches are basically among website pages and menus. It is expected that you don't run a wikipedia on Odoo, so there shouldn't be many records to search.

Also, the logic attached to this particular migration is complex enough. Honestly I wanted to make it _work_ before making it _efficient_.

pedrobaeza:
@Yajo is this stabilized?

yajo:
Yes, it behaved properly in a couple of migrations already.

addons/website/migrations/12.0.1.0/post-migration.py
pedrobaeza: Let's keep it this way for now, as the alternative code is not simpler.

pedrobaeza (approved):

PR #1934, opened 2019-07-17
[12.0][IMP] mail: faster end-migration

MiquelRForgeFlow:
A lot FASTER end-migration.

I have a database that without this PR consumes many hours in this end-migration. But with this PR, it's several seconds!!!

--

### DIFF
--- a/addons/mail/migrations/12.0.1.0/end-migration.py
+++ b/addons/mail/migrations/12.0.1.0/end-migration.py
@@ -1,6 +1,7 @@
 # Copyright 2018 Eficent <http://www.eficent.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+from psycopg2.extensions import AsIs
 
 
 def fill_mail_tracking_value_track_sequence(env):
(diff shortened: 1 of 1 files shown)


addons/mail/migrations/12.0.1.0/end-migration.py
```diff
@@ -35,29 +36,26 @@ def fill_mail_thread_message_main_attachment_id(env):
...
-        records = env[model].with_context(active_test=False).search(
-            [('message_main_attachment_id', '=', False)])
-        attachment_obj = env['ir.attachment'].with_context(
-            prefetch_fields=False,
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE %s t
+            SET message_main_attachment_id = ia.id
+            FROM (
+                SELECT id, res_id
```
pedrobaeza, line 50: Why do you need 3 nested select? For sure you can reduce this expression.

MiquelRForgeFlow: Well, if you know how to reduce it, then propose it. By the way, this query is already superfast, I don't see the need to break my brain to find a less nested query.

pedrobaeza: It's more a question of being more readable than performance. It's hard to deduce what is this doing...

MiquelRForgeFlow: I will not break my brain to make it more readable. What matters to me is that the query really does exactly what the previous code did but with efficacy.

But if you want an explanation: the query groups the attachments by their kind of mimetype and res_id. For each res_id, you will have one, two or three of this groups if they exist (and labeled them by 1, 2 or 3), but the query always will select the existing group with the lower label. The `max(id)` is because the order in `ir.attachment ` is defined as `id DESC`.

pedrobaeza: OK, thanks for the explanation. Just put as code comment and it's enough for me.

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #1939, opened 2019-07-18
[12.0][MIG] l10n_be & l10n_be_invoice_bba

MiquelRForgeFlow:
Migration scripts for `l10n_be_invoice_bba` module.

--


addons/account/migrations/12.0.1.1/post-migration.py
```diff
@@ -220,6 +220,42 @@ def set_default_taxes(env):
         })
 
 
+def fill_res_company_invoice_reference_type(cr):
```
pedrobaeza: Why this change here?

MiquelRForgeFlow: Well, I am not really sure of this long query and the other long in the l10n_be_invoice_bba. I was waiting your opinion. I can remove them.

pedrobaeza: But I want to know why is needed to be done in account. Rationale?

MiquelRForgeFlow: Ah, because this part is `account` stuff. We didn't saw when we migrated `account`, the relation between `reference_type` and `invoice_reference_type`.

pedrobaeza: Then it should be a different PR...

MiquelRForgeFlow: done in https://github.com/OCA/OpenUpgrade/pull/1941.

addons/l10n_be_invoice_bba/migrations/12.0.1.2/post-migration.py
```diff
@@ -0,0 +1,64 @@
...
+def map_res_company_invoice_reference_type(cr):
+    openupgrade.logged_query(
+        cr, """
+        UPDATE res_company
+        SET invoice_reference_type = 'structured'
+        WHERE invoice_reference_type = 'bba'"""
+    )
+
+
+def map_res_company_l10n_be_structured_comm(cr):
```
MiquelRForgeFlow: I am not sure of this

MiquelRForgeFlow: Discarded because https://github.com/OCA/OpenUpgrade/pull/1941 was also discarded. Too narrow cases, not worth.

pedrobaeza (approved):

PR #1941, opened 2019-07-22
[12.0][IMP] account: try to fill invoice_reference_type

MiquelRForgeFlow:
Based on https://github.com/odoo/odoo/commit/b5bb5bd421ad5c643f72602f6a0f97a4a0eb8711.

Not sure if this is correct or if we can do something to fill the `invoice_reference_type`. My query is just based on my understanding of the change but I may be wrong.

--


pedrobaeza:
When I checked this, I see the usage of this and it's very limited, having default value for new field as valid for 99% of cases, so :-1: on adding this.

MiquelRForgeFlow:
Ok, no problem. I know it was very narrow.

PR #1943, opened 2019-07-23
[FIX] website: Ensure website-specific children views exist

yajo:
Without this patch, the website designer would be missing the "Customize" menu features in some views.

--
@Tecnativa

### DIFF
--- a/addons/website/migrations/12.0.1.0/post-migration.py
+++ b/addons/website/migrations/12.0.1.0/post-migration.py
@@ -104,6 +104,52 @@ def sync_menu_views_pages_websites(env):
                 })
 
 
+def website_views_add_children(env):
+    """Add all expected children to website-specific views.
+
+    The COW system would duplicate children views when duplicating the parent
+    one, but that system wasn't loaded in the pre-migration stage, where many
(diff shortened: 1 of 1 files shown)


addons/website/migrations/12.0.1.0/post-migration.py
```diff
@@ -104,6 +104,52 @@ def sync_menu_views_pages_websites(env):
...
+                    ("website_id", "=", False),
+                    ("inherit_id.key", "=", parent_view.key),
+                ])
+                # Trigger the COW system in write(), which will make sure that
+                # website-specific children views inherit from their
+                # website-specific parent, creating the missing children
+                # views if needed
+                child_views.write({"inherit_id": parent_view.id})
+            # Skip these views next loop
+            done_views |= todo_views
```
MiquelRForgeFlow, line 150: todo_views.exists()

yajo: In this case it isn't necessary, since `done_views` is only used to filter out ids.

pedrobaeza (approved):

PR #1944, opened 2019-07-24
[FIX] website_seo_redirection: Relocate pages

yajo:
The only controllers that support relocation upstream now are ``website.page``, but they were not relocated with previous patch.

Now, besides redirecting from the old URL, the pages will actually be found in the new URL, preserving old behavior of "relocating controller".

--
@Tecnativa

### DIFF
--- a/addons/website/migrations/11.0.1.0/post-migration.py
+++ b/addons/website/migrations/11.0.1.0/post-migration.py
@@ -79,12 +79,24 @@ def import_website_seo_redirection_data(env):
     """
     if not openupgrade.table_exists(env.cr, 'website_seo_redirection'):
         return
+    # Transfer normal redirections directly
     openupgrade.logged_query(
         env.cr, """
         INSERT INTO website_redirect
         (type, url_from, url_to, active, sequence)
         SELECT '301', wsr.origin, wsr.destination, True, 10
         FROM website_seo_redirection wsr""",
+    )
+    # Find relocated pages and relocate them
+    openupgrade.logged_query(
+        env.cr, """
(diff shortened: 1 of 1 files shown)


addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -79,13 +79,25 @@ def import_website_seo_redirection_data(env):
...
+    # Transfer normal redirections directly
     openupgrade.logged_query(
         env.cr, """
         INSERT INTO website_redirect
         (type, url_from, url_to, active, sequence)
         SELECT '301', wsr.origin, wsr.destination, True, 10
         FROM website_seo_redirection wsr""",
     )
+    # Find relocated pages and relocate them
+    env.cr.execute(
```
MiquelRForgeFlow: Please, do it in only one logged query:
```python
openupgrade.logged_query(
    env.cr, """
    UPDATE website_page wp
    SET url = wsr.url_to
    FROM website_seo_redirection wsr
    WHERE wsr.relocate_controller AND wsr.url_from LIKE '/page/%'
        AND wp.url = wsr.url_from"""
)
```

MiquelRForgeFlow (changes requested):

addons/website/migrations/11.0.1.0/post-migration.py
yajo: Thanks, done. I didn't know that syntax :blush:

MiquelRForgeFlow (approved):
LGTM, I trust you! haha

pedrobaeza (approved):

PR #1950, opened 2019-07-24
[12.0][FIX] website_livechat: convert bootstrap

MiquelRForgeFlow:
Ups!

--


pedrobaeza:
Are you sure that field is editable through website editor?

MiquelRForgeFlow:
Nop, I am not sure. How can I see that?

pedrobaeza:
First check if the field is html. Then, look if the field is added in a QWeb template with editing purposes.

MiquelRForgeFlow:
Yes, it is `html`. In the template, it appears as `<div t-field="channel.website_description" class="oe_structure mt16" />`. Is it ok? 🤔

pedrobaeza:
Yes, it's a `t-field`, so the possibility exists.

/ocabot merge

PR #1954, opened 2019-07-30
[12.0][MIG] mail_bot

### LINKED ISSUE #1953: [12.0] mail_bot migration is renaming former admin user to "OdooBot"
It should rename v12 new System (`__system__`) user.

Current:
ID: 1, name: Sytem, login:`__system__`
ID: XXX, name: OdooBot, login: admin

Expected:
image

cc @mreficent @pedrobaeza

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/issues/1953.

The problem is that until the end-migration of `base`, the `partner_root` is the `partner_admin`. So we need to wait until the end-migration to let the `mail_bot` module to load its data that modifies the `partner_root` info.

--


MiquelRForgeFlow:
It does not work. The migrations scripts doesn't execute because `mail_bot` is a new module, so does an `install` instead of an `update`.

To make it work, we should use a init_hook, a post_init_hook and a register_hook instead. 😅

PR #1955, opened 2019-07-30
[12.0][FIX] base: split v11 admin user into system and admin in the post-migration

LoisRForgeFlow:
Before this the split was done in the end-migration and any other module modifyting
base.user_root will be modifying the previous v11 admin. After that the split is done
in the end-migration of base and you end up with the modifications meant for the
v12  `__system__` user in the v12 `admin` user.

Temptative fix for https://github.com/OCA/OpenUpgrade/issues/1953

@StefanRijnhart does this look good to you? any reason why did this in the end-migration? 

cc @mreficent @pedrobaeza 

--

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/end-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/end-migration.py
@@ -1,6 +1,6 @@
 # © 2018 Opener B.V. (stefan@opener.amsterdam)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from openupgradelib import openupgrade, openupgrade_merge_records
+from openupgradelib import openupgrade
 
 
 def update_model_terms_translations(env):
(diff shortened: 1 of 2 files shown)


StefanRijnhart (approved):
Yes, good point. I don't know why I put it in the end-migration. Maybe it's because I thought I would use the ORM more than I did.

MiquelRForgeFlow (approved):
Because https://github.com/OCA/OpenUpgrade/pull/1954 doesn't work.

pedrobaeza:
One of the reasons for doing this is on end-migration is that when doing `copy` operation on the user, all possible extra fields in `res.users` record are populated, which is not happening here, so you can face problems with some extra modules that rely on that fields, so initially :-1: for me.

StefanRijnhart:
True. Having a required field (not null column in DB) in another module will break this. Possibly solved by performing a copy in SQL fetching all the fields to copy from the DB schema?

LoisRForgeFlow:
After discussing it with @pedrobaeza we reached the conclusion that this will be the best way to tackle the issue: https://github.com/OCA/OpenUpgrade/pull/1956.

PR #1956, opened 2019-07-31
[12.0][FIX] mail_bot: set auto_install to false.

LoisRForgeFlow:
This module is not adding any core functionality, only an
automatic chat that helps to discover funtionalities. Since it
is a new module in v12 it was always being installed by
openupgrade, this causes some problems with admin user split
done in end-migration of base module (see
https://github.com/OCA/OpenUpgrade/issues/1953).

If after running openupgrade you start you migrated database
with upstream Odoo, mail_bot will be installed but this time
without generating issues in the migrations result.

If we found in the future more ocurrences of the issue mentioned above (quite unlikely) we can reopen it, and find a more robust solution, but for the time being this is the more time-efficient solution due to the small scope of the problem.

### DIFF
--- a/addons/mail_bot/__manifest__.py
+++ b/addons/mail_bot/__manifest__.py
@@ -11,7 +11,7 @@
     'depends': ['mail'],
     'installable': True,
     'application': False,
-    'auto_install': True,
+    'auto_install': False,
     'data': [
         'views/assets.xml',
         'views/res_users_views.xml',


pedrobaeza (approved):

astirpe (approved):
:+1:

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
Interesting. I was seeing the branch, thus it was not deleted. I will delete it manually! haha ( well, I cannot because the branches are protected).

pedrobaeza:
@sbidoul FYI

It hasn't been deleted due to the error.

pedrobaeza:
Mistery solved. Branches got auto-protected by the branch protection rule. I have modified for not including these branches.

PR #1958, opened 2019-08-01
[11.0][FIX] account/account_tax_python: Compute correctly invoice line price_total when taxes are Python computed

pedrobaeza:
Don't fail in account post-migration in this case + add the computation in proper place.

cc @Tecnativa


MiquelRForgeFlow (approved):
LGTM

MiquelRForgeFlow:
Put "Done" in doc for `account_tax_python` module.

PR #1959, opened 2019-08-01
[12.0][IMP] mail: remove activity_date_deadline column

MiquelRForgeFlow:
To avoid confusion if someone checks the table later. This field is not stored anymore.

--

### DIFF
--- a/addons/mail/migrations/12.0.1.0/end-migration.py
+++ b/addons/mail/migrations/12.0.1.0/end-migration.py
@@ -64,7 +64,18 @@ def fill_mail_thread_message_main_attachment_id(env):
         )
 
 
+def remove_activity_date_deadline_column(env):
+    activity_mixin_models = [k for k in env.registry if issubclass(
+        type(env[k]), type(env['mail.activity.mixin'])) and env[k]._auto]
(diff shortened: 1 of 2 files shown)


addons/mail/migrations/12.0.1.0/end-migration.py
```diff
@@ -64,7 +64,16 @@ def fill_mail_thread_message_main_attachment_id(env):
         )
 
 
+def remove_activity_date_deadline_column(env):
+    activity_mixin_models = [k for k in env.registry if issubclass(
+        type(env[k]), type(env['mail.activity.mixin'])) and env[k]._auto]
+    _column_renames = {model._table: ('activity_date_deadline', None)
```
pedrobaeza:
```suggestion
    _column_renames = {env[model]._table: ('activity_date_deadline', None)
```

MiquelRForgeFlow: I thought that `model` here is not a string but that `model` is the model class. I was wrong 😑

MiquelRForgeFlow:
updated

pedrobaeza:
Still red

Issue #1952, opened 2019-07-28
[11.0][MRP][MIG] Qty produced in finished product is not correct (When produced products had serial numbers)


OthmaneGhandi:
When the produced products had tracking by serial number the qty produced in finished product is not correct 

> Instead of giving the qty 1 for each product produced (with serial number) we put always the whole quantity 
=> It's like the creation of stock.move.line is not well set for the case of products with serial numbers

https://user-images.githubusercontent.com/6293989/62003604-985c5f00-b119-11e9-8a00-6ea0d32ddc58.png

PR #1963, opened 2019-08-06
[11.0][FIX] mrp: Consider stock move lines from products that have tracking

MiquelRForgeFlow:
The creation of stock move lines in the migration of `stock` and `mrp` modules was (and still is) a headache. I think it will never be perfect.

But I tried. This PR improves it a little bit more.

--

### DIFF
--- a/addons/mrp/migrations/11.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/11.0.2.0/post-migration.py
@@ -120,10 +120,20 @@ def create_stock_move_lines_from_stock_move_lots(env):
             sm.restrict_partner_id,
             sm.picking_id,
             sm.product_id,
-            0,
-            sm.product_uom,
-            0,
-            sm.product_uom_qty as qty_done,
+            CASE WHEN sml.lot_id IS NOT NULL AND NOT sm.is_done
+                 THEN sml.quantity
+            ELSE 0
+            END as product_qty,
+            CASE WHEN sml.lot_id IS NOT NULL THEN spl.product_uom_id
+            ELSE sm.product_uom
+            END as product_uom_id,
+            CASE WHEN sml.lot_id IS NOT NULL AND NOT sm.is_done
+                 THEN sml.quantity
+            ELSE 0
+            END as product_uom_qty,
+            CASE WHEN sml.lot_id IS NOT NULL THEN sml.quantity_done
+            ELSE sm.product_uom_qty
+            END as qty_done,
             sm.name,
             sm.state,
             current_timestamp,
@@ -132,12 +142,11 @@ def create_stock_move_lines_from_stock_move_lots(env):
             TRUE,
             sml.lot_produced_id,
             sml.lot_produced_qty,
-            NULL,
+            COALESCE(sml.production_id, sm.production_id) as production_id,
             sml.workorder_id"""
     from_ = """stock_move sm
         INNER JOIN mrp_production mp ON sm.production_id = mp.id
-        LEFT JOIN stock_move_lots sml ON (sml.move_id = sm.id
-            AND sml.production_id = mp.id)
+        LEFT JOIN stock_move_lots sml ON sml.move_id = sm.id
         LEFT JOIN stock_production_lot spl ON sml.lot_id = spl.id"""
     openupgrade.logged_query(
         env.cr, """
@@ -145,7 +154,8 @@ def create_stock_move_lines_from_stock_move_lots(env):
         )
         SELECT %(select)s
         FROM %(from)s
-        WHERE sm.state NOT IN ('cancel', 'confirmed')
+        WHERE sm.state NOT IN ('cancel') AND (sml.lot_id IS NOT NULL OR
+            sm.state NOT IN ('confirmed'))
             AND sm.id NOT IN (SELECT sq.reservation_id
                               FROM stock_quant sq
                               WHERE sq.reservation_id IS NOT NULL)


pedrobaeza:
Can you explain the change and if this is linked to any issue?

MiquelRForgeFlow:
When you have tracking By Lots (or By Unique Serial Number) like this:

Selection_072

Then, it happens the following:

v10:
Selection_071

v10 migrated v11 (before this PR):
Selection_069

v10 migrated v11 (with this PR):
Selection_070

addons/mrp/migrations/11.0.2.0/post-migration.py
```diff
@@ -120,10 +120,20 @@ def create_stock_move_lines_from_stock_move_lots(env):
             sm.restrict_partner_id,
             sm.picking_id,
             sm.product_id,
-            0,
-            sm.product_uom,
-            0,
-            sm.product_uom_qty as qty_done,
+            CASE WHEN sm.is_done = TRUE THEN 0
```
pedrobaeza:
```suggestion
            CASE WHEN sml.lot_id IS NOT NULL THEN sml.quantity AND NOT sm.is_done THEN sml.quantity
            ELSE 0
```

addons/mrp/migrations/11.0.2.0/post-migration.py
```diff
@@ -120,10 +120,20 @@ def create_stock_move_lines_from_stock_move_lots(env):
...
+                 WHEN sml.lot_id IS NOT NULL THEN sml.quantity
+            ELSE 0
+            END as product_qty,
+            CASE WHEN sml.lot_id IS NOT NULL THEN spl.product_uom_id
+            ELSE sm.product_uom
+            END as product_uom_id,
+            CASE WHEN sm.is_done = TRUE THEN 0
```
pedrobaeza: Do something similar as proposed above

addons/mrp/migrations/11.0.2.0/post-migration.py
```diff
@@ -120,10 +120,20 @@ def create_stock_move_lines_from_stock_move_lots(env):
...
+                 WHEN sml.lot_id IS NOT NULL THEN sml.quantity
+            ELSE 0
+            END as product_uom_qty,
+            CASE WHEN sml.lot_id IS NOT NULL THEN sml.quantity_done
```
pedrobaeza, line 134: Do something similar as proposed above

pedrobaeza (changes requested):

pedrobaeza:
Is this for fixing #1952?

addons/mrp/migrations/11.0.2.0/post-migration.py
MiquelRForgeFlow: but here cannot. Here there is only one WHEN hahaha

MiquelRForgeFlow:
Maybe. I don't know if this fixes that or not. I just wanted to fix my own issue, nor the other ones. But it's possible that the one you mentioned is fixed with this one.

pedrobaeza:
@OthmaneGhandi please try this fix for the problem you mentioned

OthmaneGhandi:
@pedrobaeza & @mreficent  I will test it tomorrow and share with you my results

Thanks guys

OthmaneGhandi:
@pedrobaeza & @mreficent, it solves the issue BUT it has some side effects:

**1/** I think we should keep 
**`WHERE sm.state NOT IN ('cancel', 'confirmed')`** 
instead of 
`WHERE sm.state NOT IN ('cancel')` 

Because with the last one it creates stock.move.line for the confirmed stock.moves (that are not done yet), it's like they considered the quantity was produced, which is not the case.

**2/** I had an issue (I don't know if it's specific only for me or it can happens to others) some stock.move.line was not related to the production record because in stock.move.lots the field production_id was empty (is there cases where produced products can have this field empty in stock.move.lot ?) that's why I tried to use: 

`sm.production_id,` 
instead of 
`sml.production_id` 

but it's not working too :-(

OthmaneGhandi:
It works when you use : 

`from_ = """stock_move sm
        INNER JOIN mrp_production mp ON sm.production_id = mp.id
        LEFT JOIN stock_move_lots sml ON (sml.move_id = sm.id)
        LEFT JOIN stock_production_lot spl ON sml.lot_id = spl.id"""`

**instead of** 

`from_ = """stock_move sm
        INNER JOIN mrp_production mp ON sm.production_id = mp.id
        LEFT JOIN stock_move_lots sml ON (sml.move_id = sm.id
            AND sml.production_id = mp.id)
        LEFT JOIN stock_production_lot spl ON sml.lot_id = spl.id"""`

MiquelRForgeFlow:
@OthmaneGhandi thanks for your suggestions. It helped me a lot. I think both you and I can be happy know with this PR :)

OthmaneGhandi:
@mreficent you are very welcome, big thanks to you too for the great work.

just one remark I saw that you kept **`sml.production_id`** for me it worked with **`sm.production_id`** because some `stock.move.lots` the `production_id` empty : I don't know if it's a common case or is just happened to me

MiquelRForgeFlow:
well, I can put `COALESCE(sml.production_id, sm.production_id)` if your prefer

MiquelRForgeFlow:
@pedrobaeza ready to merge

pedrobaeza (approved):
FYI, there's no need to put `AS field_name`, as the assignment is done positional

MiquelRForgeFlow:
I know. But I like to put it for readability. It you don't put it, sometimes you don't know what corresponds and you have to check the 'insert' part. But in this code, the 'insert' part is separated, several lines above, and it was a bit annoying to check.

pedrobaeza:
OK, not too much problem any way, except longer lines that maybe don't fit in one line.

PR #1965, opened 2019-08-09
[IMP] Do not remove many2many tables

ivantodorovich:
OpenUpgrade has blocks that prevent tables and columns to be dropped, but it's not doing that for tables created for many2many fields.

I'm wondering if this is intentional or just never reported :thinking: 
Btw, if approved, this should be forward-ported and probably backported aswell.

Many2many rel tables should not be dropped

--

### DIFF
--- a/openerp/addons/base/ir/ir_model.py
+++ b/openerp/addons/base/ir/ir_model.py
@@ -634,6 +634,11 @@ class ir_model_relation(Model):
 
         # drop m2m relation tables
         for table in to_drop_table:
+            # OpenUpgrade: do not run the new table cleanup
+            openupgrade.message(
+                cr, 'Unknown', False, False,
+                "Not dropping the many2many table %s", table)
+            continue
             cr.execute('DROP TABLE %s CASCADE'% table,)
             _logger.info('Dropped table %s', table)
 


pedrobaeza:
Indeed this seems something forgotten.

@StefanRijnhart @mreficent @hbrunn what do you think?

MiquelRForgeFlow (changes requested):
No. We avoid dropping tables when migrating, but not when doing uninstall. If you uninstall a module, it's normal that the tables from that module should drop. But, the big question is, when you will do an uninstall with OpenUpgrade?

ivantodorovich:
Ah, that makes sense. :sweat_smile: 

We have a case where, in the migration process, some custom modules will be renamed/merged/removed, but we want to keep the data to reuse it in the new modules.
To do so we're using OpenUpgrade to uninstall them.

pedrobaeza:
@mreficent I don't agree at all, as currently uninstalling doesn't drop columns nor tables either, so this can be a expected behavior also for m2m as well, and it would be consistent and valid for later cleaning with database_cleanup.

MiquelRForgeFlow (approved):
Oh, I see, right. I should have checked previously. Then it's good! 😁

StefanRijnhart (approved):
Indeed! Thank you for fixing this oversight.

PR #1971, opened 2019-08-14
[IMP] website: Translate metadata

yajo:
This also changes the `bs4_migrated_from` column to `INTEGER` and stores additional metadata in `bs4_migration_metadata` because we need an int column to use `copy_fields_multilang()`.

Depends on:

- [x] https://github.com/OCA/openupgradelib/pull/166

@Tecnativa

### DIFF
--- a/addons/website/migrations/12.0.1.0/pre-migration.py
+++ b/addons/website/migrations/12.0.1.0/pre-migration.py
@@ -61,18 +61,55 @@ def noupdate_changes(env):
 
 
 def bootstrap_4_migration(env):
-    """Convert customized website views to Bootstrap 4."""
+    """Convert customized website views to Bootstrap 4 and multiwebsite.
+
+    This process is a bit complex, so here's the big picture:
+
+    In v11, there's no upstream support for multi-websites, but databases
+    can actually have several websites and some records (pages, menus, views)
+    get multi-website features behind the scenes, so we have to treat them
+    like semi-supported. Usually it follows this logic:
+
+    - Website-specific records only appear in their website.
+    - Website-agnostic records appear in all websites.
+    - View changes affect all websites.
+
+    In v12, there's full multi-website support, usually following this logic:
+
+    - Website-specific records override website-agnostic ones.
+    - View changes create a website-specific copy (Copy-On-Write, COW) always.
+
+    This migration tries to make v11 changes look like they were done in v12:
+
+    - Unmodified views (noupdate=0) are not touched.
+    - Website-agnostic modified (noupdate=1) views are copied and
+      marked as noupdate=0. Those copied views are made website-specific
+      and migrated to Bootstrap 4.
+
+    This must be done in the "pre" stage, because the "mid" stage could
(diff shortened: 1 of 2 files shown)


addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -61,17 +61,56 @@ def noupdate_changes(env):
...
+
+    In v12, there's full multi-website support, usually following this logic:
+
+    - Website-specific records override website-agnostic ones.
+    - View changes create a website-specific copy (Copy-On-Write, COW) always.
+
+    This migration tries to make v11 changes look like they were done in v12:
+
+    - Unmodified views (noupdate=0) are not touched.
+    - Modified (noupdate=1) views, if they were website-agnostic, are copied,
```
pedrobaeza: Better doc layout and text:
```
- Modified (noupdate=1) views:
  * if they were website-agnostic, they are copied and made 
    website-specific and migrated to Bootstrap 4.
  * if they were website-agnostic, they are marked as noupdate=0
```

addons/website/migrations/12.0.1.0/pre-migration.py
```diff
@@ -61,17 +61,56 @@ def noupdate_changes(env):
...
+    without any modifications made for other websites.
+    """
     # Preserve old dummy images that have changed
     image_replacements = _preserve_v11_dummy_images(env)
     # Create a column to remember where do new views come from
-    col_name = openupgrade.get_legacy_name("bs4_migrated_from")
+    oldview_id_col = openupgrade.get_legacy_name("bs4_migrated_from")
     table_name = env["ir.ui.view"]._table
-    if not openupgrade.column_exists(env.cr, table_name, col_name):
+    if not openupgrade.column_exists(env.cr, table_name, oldview_id_col):
```
pedrobaeza: Do you think this check is needed? It mustn't exist for sure, or you will have a problem if a module uses that column and you change its content. I would prefer to have an error in that case.

yajo: It's made to allow retrying the migration while you're developing the script. OTOH I'm not retrying like that anymore. Do I remove it then?

addons/website/migrations/12.0.1.0/pre-migration.py
yajo: Both ifs are the same? :thinking:

pedrobaeza: OK, I have read it badly. Then text should be:
```
- Website-agnostic modified (noupdate=1) views are copied and 
  marked as noupdate=0. Copied views are made website-specific
  and migrated to Bootstrap 4.
```

addons/website/migrations/12.0.1.0/pre-migration.py
pedrobaeza: I think so for reducing a bit the number of lines and not giving the impression that this is a usual pattern.

yajo:
Changes applied.

PR #1972, opened 2019-08-14
[FIX] website: Fix homepage and contactus menu URLs

yajo:
The `/page/homepage` menu must point to `/`, and menu items pointing to any view with the `website.` prefix must remove that.

@Tecnativa

### DIFF
--- a/addons/website/migrations/11.0.1.0/post-migration.py
+++ b/addons/website/migrations/11.0.1.0/post-migration.py
@@ -51,6 +51,20 @@ def link_websites_with_pages(env):
     )
 
 
+def add_redirections(env):
+    """Redirect controllers that have been moved."""
+    redirections = (
+        ("/page/contactus", "/contactus"),
+        ("/page/homepage", "/"),
+    )
+    for from_, to in redirections:
+        env["website.redirect"].create({
+            "type": "301",
+            "url_from": from_,
+            "url_to": to,
+        })
+
+
 def add_website_homepages(env):
     # Add homepage for websites
     openupgrade.logged_query(
@@ -66,6 +80,16 @@ def add_website_homepages(env):
             iuv.key = 'website.homepage' AND
             iuv.id = wp.view_id""",
     )
+
+
+def update_menu_urls(env):
+    """Update some menu URLs that must change."""
+    # Fix menus that include the "website." prefix
+    menus = env["website.menu"].search([
+        ("url", "=like", "/page/website.%"),
+    ])
+    for menu in menus:
+        menu.url = menu.url.replace("website.", "", 1)
 
 
 def delete_noupdate_records(env):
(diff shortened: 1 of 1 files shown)


addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -68,6 +68,23 @@ def add_website_homepages(env):
...
+    """Update some menu URLs that must change."""
+    # Fix menus that include the "website." prefix
+    menus = env["website.menu"].search([
+        ("url", "=like", "/page/website.%"),
+    ])
+    for menu in menus:
+        menu.url = menu.url.replace("website.", "", 1)
+    # Fix menus pointing to "/page/homepage"
+    menus = env["website.menu"].search([
+        ("url", "=", "/page/homepage"),
```
pedrobaeza: This must use `=like` the same as above as you can have anchors to homepage (`/page/homepage#anchor`).

pedrobaeza (changes requested):

addons/website/migrations/11.0.1.0/post-migration.py
pedrobaeza: As talked through chat, we need to split this is 2 conditions.

yajo: Fixed, thanks!

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -68,6 +68,28 @@ def add_website_homepages(env):
...
+    ])
+    for menu in menus:
+        menu.url = menu.url.replace("website.", "", 1)
+    # Fix menus pointing to "/page/homepage"
+    menus = env["website.menu"].search([
+        "|", "|", "|", "|", "|",
+        ("url", "=", "/page/homepage"),
+        ("url", "=", "/page/homepage/"),
+        ("url", "=like", "/page/homepage?%"),
+        ("url", "=like", "/page/homepage/?%"),
```
pedrobaeza: For this case you will have as result `//?%`

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -68,6 +68,28 @@ def add_website_homepages(env):
...
+        ("url", "=like", "/page/homepage/#%"),
```
pedrobaeza: For this case you will have as result `//#%`

pedrobaeza (changes requested):

addons/website/migrations/11.0.1.0/post-migration.py
yajo: Odoo always autoremoves the trailing `/` (no matter if the controller exists or not) so there is zero chances that a user copies and pastes this link there, and when using autofill Odoo doesn't provide answers that end with `/`.

Taking that into account, I just removed the searches that finished in `/`. There's really so low chance those exist, and even more in a page such as homepage, which is generated by XML data.

addons/website/migrations/11.0.1.0/post-migration.py
yajo: Fixed also, see above.

yajo:
With previous implementation there were the problems discussed about trailing slashes, and still the links in page bodies would fail.

After all, it's simpler and more effective to just add a 301 redirection, which I added and did the same for contactus controller, which has been moved also.

This looks better now, honestly. Ready to merge IMHO.

PR #1973, opened 2019-08-14
[12.0][FIX] account: transfer_account_code_prefix

astirpe:

As discussed in https://github.com/OCA/OpenUpgrade/pull/1742: the calculation of the transfer_account_code_prefix makes use of `leading` instead of `trailing`. But `...code_prefix` expects to be the left part of the code, not the right part; the `trailing` seems the correct one.

However in Odoo there are a lot of transfer accounts with prefix codes ending with 0, so the use of `trailing` seems to be not correct in any case.

This PR is the proposal to set `transfer_account_code_prefix` the same as `transfer_account_id.code`. It shouldn't be a problem. And the user can manually adjust it later.

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/post-migration.py
+++ b/addons/account/migrations/12.0.1.1/post-migration.py
@@ -96,7 +96,7 @@ def fill_account_chart_template_account_code_prefix(cr):
     openupgrade.logged_query(
         cr, """
         UPDATE account_chart_template act
-        SET transfer_account_code_prefix = trim(leading '0' from aat.code)
+        SET transfer_account_code_prefix = trim(trailing '0' from aat.code)
         FROM account_account_template aat
         WHERE act.%s = aat.id
         """, (AsIs(openupgrade.get_legacy_name('transfer_account_id')), ),


MiquelRForgeFlow:
One question: Is it something bad the ensuring that there aren't leading zeros in the prefix code? 🤔

astirpe:
A prefix code could contain trailing zeros. In the Dutch chart of a account there is an [example](https://github.com/odoo/odoo/blob/12.0/addons/l10n_nl/data/account_chart_template.xml#L10). IMO whatever will be the final solution, the user should check and eventually modify the `transfer_account_code_prefix` manually.

About the leading zeros, they must be kept. Otherwise the code will be not correct.

astirpe:
Comment updated

MiquelRForgeFlow:
Yeah, I know. I was asking for leading zeros. Could a prefix code contain leading zeros? If not, then the PR should be closed. If yes, then your fix is ok. I have found an example of leading zeros in [here](https://github.com/odoo/odoo/blob/12.0/addons/l10n_cr/data/l10n_cr_chart_data.xml#L9), but not sure if it is something correct or not.

astirpe:
I think you found a good example. Leading zeros must be kept.

pedrobaeza:
I still prefer to use `trailing`. If for example your current code is `572000`, and the expected prefix is `5720`, using trailing you will get `572`, which will grant 572001, 572002, and so on, which is perfectly fine and the same as the other prefix.

astirpe:
@pedrobaeza in theory you could have an existing account with code 572100. In this case setting a prefix 5720 rather than 572 makes the difference.

pedrobaeza:
Why? The prefix is used for creating a possible new account with code = prefix + counter, so it is always going to be 0 padded.

PR #1980, opened 2019-08-20
[12.0][IMP] compare_noupdate_xm_records and [FIX] noupdate_changes

### LINKED ISSUE #1978: [12.0] noupdate new records in v12 are not loaded correctly after the migration
**Impacted versions:**

12.0 at least

**Steps to reproduce:**

Upgrade db + update all

**Current behavior:**

New groups that enter in the system between tag noupdate="1" are not being created during the migration.

https://github.com/OCA/OCB/blob/c21dd92e46a0391c7960367d0d08a30fa2b8da3e/addons/account/security/account_security.xml#L204 is not on v12 after the migration

**Expected behavior:**

They should. Is there any way to do it without editing manually the xml files? Am I doing anything wrong?

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/issues/1978.

--


pedrobaeza (approved):
Excellent work!

PR #1982, opened 2019-08-20
[10.0][FIX] several noupdate_changes

MiquelRForgeFlow:
Doing analysis with https://github.com/OCA/OpenUpgrade/pull/1980 update, I found this noupdate_changes.

--


addons/payment_adyen/migrations/10.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="payment.payment_acquirer_adyen" model="payment.acquirer">
```
pedrobaeza: I wouldn't load any of these payment records, as they have been customized for sure.

pedrobaeza (changes requested):

addons/payment_adyen/migrations/10.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: dropped all the payment commits :)

pedrobaeza (approved):

PR #1983, opened 2019-08-20
[ADD] account_asset: Move to OCA account_asset_management (renamed mo…

ghost:

### DIFF
--- a/addons/account/migrations/12.0.1.1/post-migration.py
+++ b/addons/account/migrations/12.0.1.1/post-migration.py
@@ -96,7 +96,7 @@ def fill_account_chart_template_account_code_prefix(cr):
     openupgrade.logged_query(
         cr, """
         UPDATE account_chart_template act
-        SET transfer_account_code_prefix = trim(leading '0' from aat.code)
+        SET transfer_account_code_prefix = trim(trailing '0' from aat.code)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Hi, @apps2grow thanks for this effort for migrating from account_asset to account_asset_management. What belongs here is to put in `apriori.py` the renaming of the module name, but not the renaming of models, that should be done on the module itself pre-migration. And please put it on the merge_modules section, as in 11.0, there's a change you have both modules available in the list.

MiquelRForgeFlow:
I don't understand what happened :S

I did a force push, but not closing :S  And I cannot open :S

MiquelRForgeFlow:
@pedrobaeza the merge is already done in https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/addons/openupgrade_records/lib/apriori.py#L44.

MiquelRForgeFlow:
In fact, the full migration is already done in https://github.com/OCA/account-financial-tools/pull/873! :)

pedrobaeza:
Then this is useless. I will review soon the other PR

ghost:
Thank you for coordinating the openupgrade. :)
I just posted a question on https://github.com/OCA/account-financial-tools/pull/873

PR #1987, opened 2019-08-23
[12.0][IMP] openupgrade_records: detect deleted domains

MiquelRForgeFlow:
The detection of deleted contexts cannot be pushed here, but locally can be done by a little trick.

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -285,51 +285,52 @@ def compare_sets(old_records, new_records):
 def compare_xml_sets(old_records, new_records):
     reprs = collections.defaultdict(list)
 
+    def match(match_fields, match_type='direct'):
+        matched_records = []
+        for column in copy.copy(old_records):
+            found = search(column, new_records, match_fields)
+            if found:
+                old_records.remove(column)
+                new_records.remove(found)
+                if match_type != 'direct':
+                    column['old'] = True
+                    found['new'] = True
+                    column[match_type] = found['module']
+                    found[match_type] = module_map(column['module'])
+                if column['domain'] != found['domain'] and \
+                        column['domain'] != '[]' and \
+                        found['domain'] is False:
+                    column['domain'] = False
+                    found['domain'] = True
+                else:
+                    column['domain'] = False
+                    found['domain'] = False
+                if match_type != 'direct':
+                    matched_records.append(column)
+                    matched_records.append(found)
+                elif match_type == 'direct' and found['domain']:
+                    matched_records.append(found)
+        return matched_records
+
     # direct match
-    match_fields = ['module', 'model', 'name']
-    for column in copy.copy(old_records):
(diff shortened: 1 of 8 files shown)


pedrobaeza:
And what about context?

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -287,11 +287,22 @@ def compare_xml_sets(old_records, new_records):
 
     # direct match
     match_fields = ['module', 'model', 'name']
+    modified_records = []
     for column in copy.copy(old_records):
         found = search(column, new_records, match_fields)
         if found:
             old_records.remove(column)
             new_records.remove(found)
+            if column['domain'] != found['domain'] and \
```
pedrobaeza: Can't all these ifs be simplified a bit? And why you need to do something similar several times? Or at least they can be extracted to a method.

MiquelRForgeFlow:
> And what about context?

I said before:
> The detection of deleted contexts cannot be pushed here, but locally can be done by a little trick.

odoo/addons/openupgrade_records/lib/compare.py
MiquelRForgeFlow: ok

pedrobaeza:
But I don't get what that means... and why it doesn't work the same as with domain?

MiquelRForgeFlow:
Because in the domain case, the implicit default domain is `False` and the explicit domain is `[]`. Thus, we can distinguish between them.
In the case of context, both the implicit and the explicit default context is `{}`. The trick I do locally is to change the default implicit context by `{ }`.

pedrobaeza:
But if I get it correctly, it doesn't matter: what you try to see is if there's a change in that context. It doesn't matter that implicitly an empty context is the same as `{}`. If on N-1 you have empty context and in N `{}` context, you don't need to do anything.

MiquelRForgeFlow:
I don't want to check when there is a change in a context. That would give us a lot of irrelevant data. What I want to detect is only when in old version you have a context different from `{}` and in new version you don't have an explicit context. The problem is: there is no way to distinguish between a context explicit declared as `{}` and a not declared context.

pedrobaeza:
OK, now I see the problem. Thanks for the info.

pedrobaeza:
Is this finished?

StefanRijnhart (approved):

MiquelRForgeFlow:
Yeah, it's finished. It's just that I was waiting for Martin merging my related PRs in odoo to avoid doing merge of the two commits marked as "temporal". Do you prefer to wait or proceed to merge?

pedrobaeza:
Is this detecting missing context then?

As they don't hurt even if fixed upstream, I would say to go on.

MiquelRForgeFlow:
No, this doesn't detect context, as I explained before. But locally I run a detection of the context case, and the purchase commit case was the unique context case I found.

pedrobaeza:
OK, I merge this, but can't you put anything for doing it work that not being done locally?

PR #1990, opened 2019-08-27
[12.0][MIG] mass_mailing_event: Nothing to do

OthmaneGhandi:
There is no fields that were added or changes in xml views

--


MiquelRForgeFlow (approved):

rafaelbn:
This is a mistake, this module should be migrated.

pedrobaeza:
This is not an error. This is an core module that has nothing to do with the OCA module.

rafaelbn:
It's not. Is an error. I'm recording a video. The mistake is to think that the functionality that Odoo introduces is the same as what it had. It's very different

pedrobaeza:
No, the problem is that Odoo has hijacked the module name.

rafaelbn:
OK, we need a new module **name** an migrate it.

The module that shouldn't be migrate to v12 is https://github.com/OCA/event/tree/10.0/event_registration_mass_mailing which is the funcionality Odoo core include not this one mass_mailing_event.

@OthmaneGhandi @MiquelRForgeFlow I explain you the module functionality as it look you didn't know it.

Video explaining the functionally lost:

https://drive.google.com/file/d/1LGe6PXZjHFqKK9eke95Qo0Fht5xbsOnH/view?usp=sharing

From README with example.

1. Create a mass mailing for telling to 1000 partners that a new event is available. <- This is confusing the mass mailing can be sent to mass mailing, contact or whatever. But the destination are never event registration. Odoo 12 make possible to send emails to event registrationg but this module `mass_mailing_event` **don't make that function**.

IMPORTANT: he mass mailing is sent to mass mailing contacts NOT to event registration!! (here is the misunderstanding).

Example:
- You have a Mass Mailing list with 2.500 contacts with a name "EVENT GENERAL LIST"
- Event "OCA CODE SPRINT" with 0 registration
- You send a mass mailing to a Mailing LIST "EVENTS GENERAL LIST" (2.500) - excluding emails that are cancelled or confirmed (0) = 2.500 emails

2. During a week some of them (50) have been registered in the event 

Example:
- 50 contact register to the event
-  You send a mass mailing to a Mailing LIST "EVENTS GENERAL LIST" (2.500) - excluding emails that are cancelled or confirmed (50) = 2.450 emails.

3. Then, a week after the first mass mailing, duplicate it to get a second mass mailing. Relate this one to the event and exclude the registered emails. Change the message body to remember that early bird period is going to expire soon. 

Example:
-354 contacto registered to the event
- You send a mass ailing to a mailing list "EVENTS GENERAL LIST" (2.500) - excluding emails that are cancelled or confirmen (354) = 2.146 emails sent

4. Send the second mass mailing and registered emails are automatically excluded, So it's been only sent to 950 partners the ones who are not registered in the event yet.

Etc. etc.. etc..

**We should migrate functionality to v12 and maintain the module.**

Thanks

pedrobaeza:
@rafaelbn why are you still obstinate on writing here? This has nothing to do with what you want. This PR only analyzes the changes from v11 core module `mass_mailing_event` (https://github.com/OCA/OpenUpgrade/tree/11.0/addons/mass_mailing_event) and the same for v12 (https://github.com/OCA/OpenUpgrade/tree/11.0/addons/mass_mailing_event), and it's nothing to do as the module hasn't changed. What you want to achieve is different.

PR #2001, opened 2019-09-09
[MIG] account_analytic_distribution: Migration scripts

pedrobaeza:
cc @Tecnativa

### DIFF
--- a/addons/analytic/migrations/12.0.1.1/post-migration.py
+++ b/addons/analytic/migrations/12.0.1.1/post-migration.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
+from psycopg2 import sql
 
 
 def fill_account_analytic_line_company_id(cr):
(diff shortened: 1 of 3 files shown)


StefanRijnhart (approved):

pedrobaeza:
@StefanRijnhart check the use I have done of `sql.SQL` method from psycopg2 for building queries. It's very versatile and more secure than `AsIs`. For v13 I will use it massively.

StefanRijnhart:
Yes, I've seen it. I think I wasn't sure if it actually prevents SQL injections when I looked into this module some time ago. Do you have info on that?

pedrobaeza:
using `sql.Identifier()` for adding the column/table names prevents the SQL injection through a check performed on `format` method.

StefanRijnhart:
That's great!

pedrobaeza:
Look this snippet:
```python
>>> query = sql.SQL("SELECT * FROM {table}")
>>> env.cr.execute(query.format(table=sql.Identifier('res_partner')))  # no error
>>> env.cr.execute(query.format(table=sql.Identifier("res_partner; UPDATE res_partner SET name='x' WHERE id=1")))

ERROR: relation "res_partner; UPDATE res_partner SET name='x' WHERE id=1" does not exist
LINE 1: SELECT * FROM "res_partner; UPDATE res_partner SET name='x' ...
                      ^

  File "<console>", line 1, in <module>
  File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 155, in wrapper
  File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 232, in execute
psycopg2.ProgrammingError: relation "res_partner; UPDATE res_partner SET name='x' WHERE id=1" does not exist
LINE 1: SELECT * FROM "res_partner; UPDATE res_partner SET name='x' ...
```

In fact, more than a check itself, it's that it's enclosed for fully qualified query.

PR #2003, opened 2019-09-10
[ADD] stock_dropshipping/sale_purchase: Migration scripts

pedrobaeza:
New module `sale_purchase` is a dependency of `stock_dropshipping` and part of the logic/fields is moved to this one, so both are handled together.

cc @Tecnativa


addons/stock_dropshipping/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_xmlids(env.cr, [
+        ('stock_dropshipping.procurement_rule_drop_shipping',
+         'stock_rule_drop_shipping'),
```
MiquelRForgeFlow: needs the prefix

pedrobaeza: Fixed

pedrobaeza:
I'm seeing that some code is not going to be executed as `sale_purchase` is a new module, so no migration script is executed. I'll change code to other module (probably base), but tomorrow.

pedrobaeza:
OK, I have removed the update of the field moved to other module, as v12 handles it and it's not a translatable field.

MiquelRForgeFlow (approved):

pedrobaeza:
In this project may be more convenient to not use the bot if we have a green CI for not tempting the luck, and also for not having to wait a lot (as the CI is slow).

pedrobaeza:
Any way, thanks for your review!

Issue #468, opened 2016-04-08
[analysis] Removed models are not logged per module


JordiBForgeFlow:
For example, account.analytic.journal disappears in v9, and there's no trace of this in the analysis files, except for many2one references to this model that are deleted.

StefanRijnhart:
They are logged in the general log: https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/addons/base/migrations/9.0.1.3/openupgrade_general_log.txt#L77

It is hard to log them by module, because Odoo does not properly register which module 'owns' a model, although we could log a module in every module in which it is either defined or overridden as this corresponds to the model_ entries in ir_model_data.

pedrobaeza:
Yeah, I think that can be an improvement, although it's very verbose. We can also use a reference count to determine which module is the base (or maybe through _inherit/_name property - the original module that declares the model for the first time is the one that doesn't have _inherit).

pedrobaeza:
@mreficent a challenge for you :wink:

MiquelRForgeFlow:
Each model has the `_original_module` attribute, which shows the module where the model is declared for first time.

pedrobaeza:
OK, if it works properly, then we have the solution!

MiquelRForgeFlow:
With https://github.com/OCA/OpenUpgrade/pull/2004 and https://github.com/OCA/OpenUpgrade/pull/2012 merged, I think we can close this issue.

PR #2004, opened 2019-09-10
[12.0][IMP] openupgrade_records: add compare_model_sets

MiquelRForgeFlow:
- Implements #468.
- Also fixes the "renamed xmlids" search.
- Also sorts the openupgrade_general_log by module.

--


odoo/addons/openupgrade_records/models/generate_records_wizard.py
```diff
@@ -67,4 +67,22 @@ def generate(self):
         self.env.cache.invalidate([
             (self.env['openupgrade.record']._fields['noupdate'], None),
         ])
+
+        # Log model records
+        self.env.cr.execute(
```
pedrobaeza, line 87: Why do you need to log this?

pedrobaeza: Oh, I see that this is implementing #468!! That's a dream come true.

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
```diff
@@ -14,6 +42,18 @@ account      / account.chart.template   / cash_account_code_prefix (char): now r
 account      / account.chart.template   / company_id (many2one)         : DEL relation: res.company
 account      / account.chart.template   / transfer_account_code_prefix (char): NEW required
 account      / account.chart.template   / transfer_account_id (many2one): DEL relation: account.account.template, required
+account      / account.financial.report / account_ids (many2many)       : DEL relation: account.account
```
pedrobaeza, line 45: Why is it here all this extra stuff?

pedrobaeza:
This is so powerful new analysis options and this is near perfection for getting all the possibilities to be covered by OpenUpgrade.

Please split the 2 new analysis features (better renaming detection and model detection) in 2 separate commits.

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
StefanRijnhart: It was removed in https://github.com/odoo/odoo/commit/e51b4f659fc5e4253d3156754ada21166c8d1bf8

StefanRijnhart:
Very nice! Thanks for splitting up the commits. I already ported the framework to 13 and I will cherry pick the framework one.

StefanRijnhart (approved):

pedrobaeza:
@StefanRijnhart Where are you publishing the framework?

StefanRijnhart:
It's here, https://github.com/StefanRijnhart/OpenUpgrade/tree/port/13.0/framework. I think I only need to add the table for the module coverage to the docs. PR this week!

Analysis preview: https://github.com/StefanRijnhart/OpenUpgrade/commit/e9abe833e00e798eb6671a5c2c52fb9eac5eb789

pedrobaeza:
Great! Thanks as always for this initial impulse! Have you started from sass-12.5 branch, isn't it?

StefanRijnhart:
Yes, it's based on Sunday's saas-12.5

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
MiquelRForgeFlow: Because before this stuff was not showed. I decided it was better to show it. Do you like it?

MiquelRForgeFlow:
Updated. Ready to review and merge!

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -273,7 +272,7 @@ def match(match_fields, report_fields, warn=False):
             "# Direct match: %d" % matched_direct,
             "# Found in other module: %d" % matched_other_module,
             "# Found with different type: %d" % matched_other_type,
-            "# Found with different name: %d" % matched_other_name,
+            # "# Found with different name: %d" % matched_other_name,
```
pedrobaeza: Why commented instead of removing it?

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -167,38 +171,33 @@ def clean_records(records):
     matched_direct = 0
     matched_other_module = 0
     matched_other_type = 0
-    matched_other_name = 0
+    # matched_other_name = 0
```
pedrobaeza: Why not removing the line?

odoo/addons/openupgrade_records/lib/compare.py
MiquelRForgeFlow: it's related to https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/addons/openupgrade_records/lib/compare.py#L223L227

odoo/addons/openupgrade_records/lib/compare.py
MiquelRForgeFlow: it's related to https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/addons/openupgrade_records/lib/compare.py#L223L227

pedrobaeza: Yeah, but remove it, not comment it

MiquelRForgeFlow: I prefer not deleting it, as the match part originally was commented and not deleted. Maybe in the future we need this to uncomment the part after some tweak.

addons/product/migrations/12.0.1.2/pre-migration.py
```diff
@@ -51,4 +55,5 @@ def migrate(env, version):
     openupgrade.rename_fields(env, _field_renames)
     openupgrade.rename_models(cr, _model_renames)
     openupgrade.rename_tables(cr, _table_renames)
+    openupgrade.rename_xmlids(cr, xmlid_renames)
```
pedrobaeza, line 58: Then you should remove https://github.com/OCA/OpenUpgrade/blob/e18f323384c2bfcfea2464784fb8c318627fddf5/addons/hr_expense/migrations/12.0.2.0/post-m.py#L32

odoo/addons/openupgrade_records/lib/compare.py
StefanRijnhart: No, it's alright. Just delete it.

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
StefanRijnhart: I for one like it!

pedrobaeza: Well, if the model is removed, no need of showing that all the fields of that model have been removed. This is too verbose and with no practical use. You are already seeing that the model is gone, so the options are 2:

* It has been deprecated/removed. You handle that model removal, not the fields removal.
* It has been renamed. In that case, you have that information also in the model. I wonder though if it's symmetrical: when renamed, are new fields of that new renamed model shown as well on target module? I bet also for removing that verbose information.

odoo/addons/openupgrade_records/lib/compare.py
MiquelRForgeFlow: good!

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
pedrobaeza: @StefanRijnhart for me it's information that is not going to be useful as explained in my previous comment. Can you reconsider?

MiquelRForgeFlow: Well, I prefer showing it, because maybe you have a field moved from a removed model to another model. Then, you may visually detect the moved field if the old field of the old model is showed. The matching is not perfect. The tool doesn't detect some cases, like when the field also changes the name.

As a note, remember that in v13 the "oldname" is removed.

addons/point_of_sale/migrations/12.0.1.0.1/openupgrade_analysis.txt
```diff
@@ -1,3 +1,5 @@
+---Models in module 'point_of_sale'---
+obsolete model pos.discount
```
pedrobaeza, line 2: With current code, why the fields of this model are not logged as `DEL`? (although you know I prefer to not be logged)

MiquelRForgeFlow: because `pos.discount` is transient model

pedrobaeza: Shouldn't we logged that then? Something like:
```
obsolete model pos.discount (transient)
```
or directly not log them, as we don't have to worry about them on migrations.

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -75,8 +75,10 @@
     'hr.holidays': 'hr.leave',
     'hr.holidays.status': 'hr.leave.type',
     'mrp.repair': 'repair.order',
+    'mrp.repair.cancel': 'repair.cancel',
```
pedrobaeza, line 78: Are these manual entries still needed now with new detection tools?

MiquelRForgeFlow: yes, the analysis use them to know the model renames. if not, the analysis will not show any model rename and then, some moved relational fields may not be detected.

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
pedrobaeza: Well, that's a very remote possibility, and at the end, in your manual analysis of the obsolete model you'll detect that linking to the new `ADD` field, so I still vote to not show this. Waiting for @StefanRijnhart's final opinion.

addons/point_of_sale/migrations/12.0.1.0.1/openupgrade_analysis.txt
MiquelRForgeFlow: I prefer do that in another PR. Better for v13 openupgrade.

pedrobaeza: @mreficent you are introducing here this feature, so it's not something to be done later. I vote for removing it. It's as simple as doing that check before printing the text. @StefanRijnhart what do you think?

MiquelRForgeFlow: It's not easy, it will have many new lines of code.

pedrobaeza: Yeah, but if it has to be done, better now than later.

StefanRijnhart: I think current state is good enough. Transient models are logged as obsolete, which is informative, and at the same time the deleted  fields are not logged so it's not too alarmistic.
We can always improve it later.

addons/account/migrations/12.0.1.1/openupgrade_analysis.txt
StefanRijnhart: I agree with @mreficent. Current visual works really well with something like the removed invoice model: https://github.com/StefanRijnhart/OpenUpgrade/blob/tmp/13.0/analysis/addons/account/migrations/13.0.1.1/openupgrade_analysis.txt

pedrobaeza: OK, I don't win any battle, hehe

addons/point_of_sale/migrations/12.0.1.0.1/openupgrade_analysis.txt
pedrobaeza: OK

MiquelRForgeFlow: ok

PR #2009, opened 2019-09-11
[FIX] hr_timesheet: Recompute fields for issues converted to tasks

pedrobaeza:
This module adds several computed stored fields to project.task (for example, effective_hours and remaining_hours). Issues converted to tasks, being inserted by SQL, don't have these fields computed, so we force here its recomputation.

cc @Tecnativa


StefanRijnhart (approved):

yelizariev:
This commit seems to be wrong, because hr_timesheet doesn't depend on project_issue, so origin_issue_column may not exist

yelizariev:
@em230418 got some error with it

pedrobaeza:
See #2035

PR #2011, opened 2019-09-11
Port/13.0/framework

StefanRijnhart:
I removed the `recompute_fields_blacklist` mechanism from https://github.com/OCA/OpenUpgrade/pull/663 as it is incompatible with Odoo 13 computed fields handling and the mechanism was unused since its introduction.

Running tests was refactored in Odoo and the migration test mechanism was affected. I refactored accordingly but I have not yet tested this. I also removed the test_deferred.py mechanism in favour of supporting post-install and other tags in migration test files. Tests of this PR are broken of course and can only be fixed by the migration of the `base` module.

Example of the analysis that this creates: https://github.com/StefanRijnhart/OpenUpgrade/commit/f6217a3f38cf61afbf5a19cf9900a55f001fd03e


StefanRijnhart:
Also added the 12.0 test database here: https://github.com/OCA/OpenUpgrade/releases and updated https://github.com/OCA/OpenUpgrade/wiki/Crude-script-to-create-the-full-analysis-between-two-versions-of-Odoo to use OdooRPC.

addons/stock/__init__.py
```diff
@@ -12,6 +12,11 @@
 
 # TODO: Apply proper fix & remove in master
 def pre_init_hook(cr):
+    # OpenUpgrade: don't uninstall data as this breaks the analysis
+    # Origin of this code is https://github.com/odoo/odoo/issues/22243
```
MiquelRForgeFlow, line 16: I am not sure. I think maybe in v13 is already solved.

odoo/openupgrade/openupgrade_loading.py
```diff
@@ -0,0 +1,285 @@
...
+            'table': v.relation if v.type == 'many2many' else '',
+            'required': v.required and 'required' or '',
+            'stored': v.store and 'stored' or '',
+            'selection_keys': '',
+            'req_default': '',
+            'hasdefault': model._fields[k].default and 'hasdefault' or '',
+            'inherits': '',
+            }
+        if hasattr(v, 'oldname'):
+            properties['oldname'] = v.oldname
```
MiquelRForgeFlow: `oldname` doesn't exist in v13. should be removed everywhere.

odoo/openupgrade/openupgrade_loading.py
```diff
@@ -0,0 +1,285 @@
...
+    if model._inherits:
+        model_registry['_inherits'] = {
+            '_inherits': str(model._inherits)}
+    for k, v in model._fields.items():
+        properties = {
+            'type': typemap.get(v.type, v.type),
+            'isfunction': isfunction(model, k),
+            'isproperty': isproperty(model, k),
+            'isrelated': isrelated(model, k),
+            'relation': v.comodel_name if v.type in (
```
MiquelRForgeFlow:
```suggestion
            'relation': _get_relation(v)
```
call a new method here like:

```python
def _get_relation(self, values):
    if v.type in ('many2many', 'many2one', 'one2many'):
        return v.comodel_name
    elif v.type = 'many2one_reference':
        return v.model_field
    else:
       return  ''
```

odoo/addons/openupgrade_records/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
MiquelRForgeFlow:
```suggestion
.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
```

odoo/addons/openupgrade_records/README.rst
```diff
@@ -0,0 +1,58 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
```
MiquelRForgeFlow:
```suggestion
   :target: https://www.gnu.org/licenses/agpl
```

odoo/addons/openupgrade_records/README.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/openupgrade/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
MiquelRForgeFlow:
```suggestion
* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
```

odoo/addons/openupgrade_records/README.rst
```diff
@@ -0,0 +1,58 @@
...
+* Log removed modules in the module that owned them (#468)
+* Detect renamed many2many tables (#213)
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/openupgrade/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
```
MiquelRForgeFlow:
```suggestion
help us smash it by providing detailed and welcomed feedback.
```

odoo/addons/openupgrade_records/README.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Contributors
+------------
+
+* Stefan Rijnhart <stefan@opener.amsterdam>
+* Holger Brunn <hbrunn@therp.nl>
+* Pedro M. Baeza <pedro.baeza@gmail.com>
+* Ferdinand Gassauer <gass@cc-l-12.chircar.at>
+* Florent Xicluna <florent.xicluna@gmail.com>
+
```
MiquelRForgeFlow, line 45: add me here please

odoo/addons/openupgrade_records/models/generate_records_wizard.py
```diff
@@ -17,6 +17,15 @@ class GenerateWizard(models.TransientModel):
     state = fields.Selection(
         [('init', 'init'), ('ready', 'ready')], default='init')
 
+    def quirk_standard_calendar_attendances(self):
+        """ Introduced in Odoo 13. The reinstallation causes a one2many value
```
MiquelRForgeFlow, line 21: Have you notified odoo of this thing?

MiquelRForgeFlow:
@StefanRijnhart can I force push in this PR to update it a bit? 🙂

StefanRijnhart:
Yes please, go ahead! Agree with all your comments.

odoo/addons/openupgrade_records/models/generate_records_wizard.py
StefanRijnhart: No, I haven't. I don't think reinstallation of modules is officially supported.

MiquelRForgeFlow:
@StefanRijnhart Will you continue to work on this? Will you migrate `base`?

StefanRijnhart:
@mreficent thanks for the updates! No time in the coming week to migrate base I am afraid. I was thinking this could be merged first anyway.

MiquelRForgeFlow:
@StefanRijnhart did you have any problem doing the analysis? I tried but I have a problem when executing the `generate` method, in the line of `Registry.new(...)`, because I don't know why, but a query gets stopped and waiting the resolution of another one (the one from `_compute_session_token` method in `res.users`) that gets blocked by unknown reason. Maybe due the execution of some cron.

MiquelRForgeFlow:
I think the problem was this commit https://github.com/odoo/odoo/commit/be8fc2296b3822baffe0bc68cb4b7efdfe53fa91, fixed in https://github.com/odoo/odoo/commit/eded0294ab1c5dfc6eb268f84d53ada9ba775352.

I will try again.

MiquelRForgeFlow:
Still having same error when trying to do the analysis.

requirements.txt
```diff
@@ -43,3 +43,6 @@ XlsxWriter==0.9.3
 xlwt==1.3.*
 xlrd==1.0.0
 pypiwin32 ; sys_platform == 'win32'
+# Needed for OpenUpgrade
+odoorpc==0.7.0
+openupgradelib>=2.0.0
```
pedrobaeza: Please use for avoiding future problems
```suggestion
git+https://github.com/OCA/openupgradelib.git@master
```

norlinhenrik:
https://github.com/OCA/OpenUpgrade/tree/12.0/odoo/addons/openupgrade_records is not yet in 13.0 and I see in the communication above that there are some issues. I think this should have high priority. We cannot do any migrations to 13.0 without the analysis. Could you please add the label **HELP WANTED**?

StefanRijnhart:
@norlinhenrik I will be trying to fix the blocking error of not being able to create the analysis this week.

MiquelRForgeFlow:
@StefanRijnhart no need, I already have fixed it in local. It's just I don't have pushed yet to this branch some changes.

StefanRijnhart:
What's your fix? Mine is
```--- a/odoo/http.py
+++ b/odoo/http.py
@@ -1024,6 +1024,11 @@ class OpenERPSession(werkzeug.contrib.sessions.Session):
         env = odoo.api.Environment(request.cr, self.uid, self.context)
         # here we check if the session is still valid
         if not security.check_session(self, env):
+            # OpenUpgrade: when asking openupgrade_records to generate records
+            # over jsonrpc, a query on res_users in the call above locks this
+            # table for the sql operations that are triggered by the
+            # reinstallation of the base module
+            env.cr.rollback()
             raise SessionExpiredException("Session expired")
 
     def logout(self, keep_db=False):
```

MiquelRForgeFlow:
Mine was:

```
+++ b/odoo/service/security.py
@@ -20,6 +20,9 @@ def compute_session_token(session, env):
    return self._compute_session_token(session.sid)

def check_session(session, env):
+   # OpenUpgrade HACK: Put early return here for avoiding error doing analysis
+   return True
+   # END HACK
    self = env['res.users'].browse(session.uid)
    expected = self._compute_session_token(session.sid)
    if expected and odoo.tools.misc.consteq(expected, session.session_token):
```

StefanRijnhart:
Heh. Shall I push my version? I rebased on today's Odoo 13.

MiquelRForgeFlow:
Just push the commit, but please try to avoid rebase + push force now if you can, because there have been other changes in the framework.

StefanRijnhart:
Fix pushed

MiquelRForgeFlow:
@pedrobaeza @StefanRijnhart to me this PR is ready. Would you like to merge as it this now, or do prefer to let it open to add in the future the base migration scripts?

yelizariev:
to merge or not to merge
image

StefanRijnhart:
Thanks! Why do you prevent the deletion of dangling XML IDs?

pedrobaeza:
@StefanRijnhart that's one of my requests. It's because that ones are removed via `database_cleanup`, and can be useful for modules that restores a lost functionality. Example, when in 9.0, phone calls were removed from module `crm`, and we extract them into `crm_phonecall`. The phonecall types XML-IDs introduced in data were removed that way, and we had to link them again.

StefanRijnhart:
@pedrobaeza not sure I understand. In your use case you are migrating the XML ids of records that still exist in the database. Did you mean to request to preserve the XML ids of records that were removed from the module data but still exist because noupdate records are not actually removed? Because that is not what was implemented in https://github.com/OCA/OpenUpgrade/pull/2011/commits/cdc470e98791b5e4cbf2f398a2029264e8e8d341

pedrobaeza:
Uhm, they are records `noupdate=0`, but they are not dangling. Please point me to the line with the change to check again. Maybe the case is not valid...

StefanRijnhart:
With dangling XML ids I mean entries in ir_model_data that refer to a res_id that is no longer in the model's table. Like the commit says: 'don't delete xmlids linked to nothing'.

StefanRijnhart:
Did you want to circumvent this line instead: https://github.com/OCA/OpenUpgrade/commit/cdc470e98791b5e4cbf2f398a2029264e8e8d341#diff-bb1e3d0f2e95b26bcd024950824e022fL1977

pedrobaeza:
@mreficent do you have the text I put you about this (it was in some issue/PR/openupgralib thread)? 

> Did you want to circumvent this line instead...

Yeah, it can be. The case I had in mind was the one I have described.

MiquelRForgeFlow:
@StefanRijnhart good catch, maybe I didn't understand correctly the issue @pedrobaeza told me. It was a long time ago. I think it's fine now.

StefanRijnhart:
OK, thanks for the update but the comment is not correct. Database_cleanup will not remove these records, but a regular module update in a regular (non-openupgrade) Odoo instance will remove the XML ids.

StefanRijnhart:
Alright! Github does not let me approve "my own PR", but ready for merge on my part.

MiquelRForgeFlow (approved):
I approve it for you! 😄

pedrobaeza:
Please fix Travis putting Python 3.7 in .travis.yml

pedrobaeza:
This is expected to fail until base migration (and maybe something else is done), so I merge.

PR #2012, opened 2019-09-12
[12.0][UPD] openupgrade_records: add model_type for model records

MiquelRForgeFlow:
Applying https://github.com/OCA/OpenUpgrade/pull/2004#discussion_r323733085 comment.

--

### DIFF
--- a/odoo/addons/openupgrade_records/models/openupgrade_record.py
+++ b/odoo/addons/openupgrade_records/models/openupgrade_record.py
@@ -41,17 +41,39 @@ class Record(models.Model):
     domain = fields.Char(readonly=True)
     prefix = fields.Char(compute='_compute_prefix_and_suffix')
     suffix = fields.Char(compute='_compute_prefix_and_suffix')
-    original_module = fields.Char(compute='_compute_original_module')
+    model_original_module = fields.Char(
+        compute='_compute_model_original_module')
+    model_type = fields.Char(compute='_compute_model_type')
 
     @api.depends('name')
     def _compute_prefix_and_suffix(self):
         for rec in self:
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Can you apply it to analysis files?

MiquelRForgeFlow:
I will do it later, after a new update upstream.

pedrobaeza:
Yeah, but without it, it's difficult to analyze if it's correct or not.

odoo/addons/openupgrade_records/models/openupgrade_record.py
```diff
@@ -41,17 +41,37 @@ class Record(models.Model):
...
+    def _compute_model_type(self):
+        for rec in self:
+            if rec.type == 'model':
+                model = self.env[rec.model]
+                if model._auto and model._transient:
+                    rec.model_type = 'transient'
+                elif model._auto:
+                    rec.model_type = ''
+                else:
+                    rec.model_type = 'abstract'
```
pedrobaeza, line 72: `_auto = False` is not the same as `abstract`, so you must distinguish both cases

MiquelRForgeFlow: Well, and how I call those models that are `_auto=False` but are not defined as `abstract`?

pedrobaeza: I would bet for "sql view"

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #2014, opened 2019-09-12
[12.0] update upstream abc0ecd6965

MiquelRForgeFlow:
Merge remote-tracking branch 'odoo/12.0' into 12.0-update-upstream-abc0ecd6965

Without conflicts

--


MiquelRForgeFlow:
`The job exceeded the maximum log length, and has been terminated.` 🙄

pedrobaeza:
And it's because 2 errors that Odoo has introduced:
```
2019-09-12 15:24:38,474 5664 INFO openupgrade odoo.fields: Field stock.picking.move_ids_without_package depends on non-stored field stock.move.picking_type_entire_packs 

2019-09-12 15:24:38,477 5664 INFO openupgrade odoo.fields: Field stock.picking.move_line_ids_without_package depends on non-stored field stock.move.line.picking_type_entire_packs
```

MiquelRForgeFlow:
I don't think these are errors, because they appear as INFO.

pedrobaeza:
Well, technically they aren't, as it's only information (but denotes something bad, as that depends won't be recomputed correctly), and that's why it appear. The problem is that this message floods totally the log. Look the number of appearances...

MiquelRForgeFlow:
But without this PR, the branch 12.0 also has that information. It's not something new from this PR.

pedrobaeza:
And why it's now so frequent?

MiquelRForgeFlow:
I see as equal frequent. It appears for first time when loading `stock`. And then, it repeats at the same places in the log. It's not something new from this PR.

pedrobaeza:
Yeah, but the problem with Travis is that in this occasion, the number of lines of the log has been exceeded, and it's due to that message.

MiquelRForgeFlow:
I just compared. It's as you say, these two lines over repeated, like spamming.

MiquelRForgeFlow:
Issue done in odoo. Waiting feedback.

pedrobaeza:
Is this working now?

MiquelRForgeFlow:
> Is this working now?

Still same problem.

pedrobaeza:
Where is the line error?

MiquelRForgeFlow:
> Where is the line error?

> `The job exceeded the maximum log length, and has been terminated.`

pedrobaeza:
Ah, OK, so it's the verbose logging thing. Isn't you going to revert those commits?

sswapnesh:
> Ah, OK, so it's the verbose logging thing. Isn't you going to revert those commits?

In https://github.com/odoo/odoo/pull/36885

pedrobaeza:
@sswapnesh are you sure that commit fixes this? I don't think so.

sswapnesh:
My bad, No It not this one

MiquelRForgeFlow:
Replaced by https://github.com/OCA/OpenUpgrade/pull/2044.

PR #2017, opened 2019-09-13
[12.0][UPD] new analysis

MiquelRForgeFlow:
Using https://github.com/OCA/OpenUpgrade/pull/2047 and https://github.com/OCA/OpenUpgrade/pull/2044.

--


addons/sale/migrations/12.0.1.1/openupgrade_analysis.txt
```diff
@@ -57,10 +57,9 @@ sale         / sale.order.line          / qty_delivered_manual (float)  : NEW ha
 sale         / sale.order.line          / qty_delivered_method (selection): NEW selection_keys: ['analytic', 'manual'], isfunction: function, stored
 sale         / sale.order.line          / untaxed_amount_invoiced (float): NEW isfunction: function, stored
 sale         / sale.order.line          / untaxed_amount_to_invoice (float): NEW isfunction: function, stored
-sale_order_dates / sale.order               / commitment_date (datetime)    : not stored anymore
-sale_order_dates / sale.order               / commitment_date (datetime)    : was renamed to expected_date [nothing to do]
+sale_order_dates / sale.order               / commitment_date (datetime)    : not a function anymore
 sale_order_dates / sale.order               / effective_date (date)         : module is now 'sale_stock' ('sale_order_dates')
-sale_order_dates / sale.order               / requested_date (datetime)     : was renamed to commitment_date [nothing to do]
+sale_order_dates / sale.order               / requested_date (datetime)     : DEL
```
pedrobaeza, line 62: Why this change?

MiquelRForgeFlow: I don't know.

pedrobaeza:
Please update this

MiquelRForgeFlow:
@pedrobaeza PR updated. ready to merge.

addons/hr/migrations/12.0.1.1/openupgrade_analysis.txt
```diff
@@ -27,5 +27,6 @@ hr           / mail.channel             / subscription_department_ids (many2many
 ---XML records in module 'hr'---
 NEW hr.employee: hr.employee_admin (noupdate)
 DEL hr.employee: hr.employee_root (noupdate)
+NEW ir.rule: hr.hr_employee_comp_rule (noupdate)
```
pedrobaeza, line 30: Does this changes the behavior of previous version? We may need to disable it when migrating.

addons/sale/migrations/12.0.1.1/openupgrade_analysis.txt
pedrobaeza: You haven't modified code that makes this change, isn't it?

MiquelRForgeFlow: Exactly., thus it surprises me why this changed :S

pedrobaeza: So we accept it and continue?

addons/hr/migrations/12.0.1.1/openupgrade_analysis.txt
mreficent: No problem. In v11, hr.employee didn't have a company_id field.

pedrobaeza: Well, then there's a problem: all companies see all employees, and now there are not going to do it, so we have to query if several companies, and if so, deactivate the rule. And also be sure the field remains empty.

MiquelRForgeFlow: EDIT: In v11, `hr.employee` has `company_id` because does inherit of `resource.mixin`. Thus means, employees have the company of their resource.

pedrobaeza: OK, we can leave it this way then.

addons/sale/migrations/12.0.1.1/openupgrade_analysis.txt
MiquelRForgeFlow: yes

PR #2019, opened 2019-09-14
[12.0][IMP] project: Move fill_project_project_inherits_values to pre

pedrobaeza:
We do this for avoiding the error that appears in the log:
```
odoo.sql_db: bad query: b'ALTER TABLE "project_project" ALTER COLUMN "name" SET NOT NULL'
ERROR: column "name" contains null values
```

Althought that error is temporary, and the constraint is added on a second update of the module, it's better to avoid the annoying error in the log when you check it, and also for having the constraint applied immediately if needed.

cc @Tecnativa


MiquelRForgeFlow (approved):
LGTM

PR #2021, opened 2019-09-16
[FIX] base: Preserve system user email

yajo:
After this migration, all system emails are sent by `__system__` instead of `admin`, so `__system__` should keep its email address to avoid altering the experience of users receiving emails.

The side effect will be that admin user will have no email address, which is not cool, so after migrating you should either update admin's email or remove him.

--
@Tecnativa TT19494

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/end-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/end-migration.py
@@ -58,8 +58,9 @@ def fork_off_system_user(env):
     )
     user_root.write({
         'partner_id': partner_root.id,
-        'email': 'root@example.com',
+        'email': partner_admin.email,
     })
+    partner_admin.email = False
     env.cr.execute(
         """ UPDATE ir_model_data SET res_id = %s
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I have just discovered that we can't have 2 equal emails for partners linked to users (system gives an error), so let's go for switching the email and put in the other this fake email.

yajo:
OK, fixed. It's less arbitrary and just:

1. Let `__system__` have the same email `admin` had.
1. Let the new admin have no email.

pedrobaeza (approved):
Waiting for @StefanRijnhart's approval as well

StefanRijnhart (approved):

pedrobaeza:
FYI I have finally needed to assign an email to old admin user, or we got as well errors with some features requiring it: 87213fb8bab4a83237ce6d73b3c9924ee5ccfa96

Example error (in Spanish):

`El asociado 'x' está suscrito a una o más listas de correo. Se le debe asignar un email.`

PR #2024, opened 2019-09-17
[7.0][ADD] sale_journal: Nothing to do

pedrobaeza:
Migrating directly a very old instance (yes, there are still in the world these ones) to v12, this module is not covered...

cc @Tecnativa


pedrobaeza:
I'm merging this directly as simple.

Issue #2000, opened 2019-09-09
[11.0][STOCK] Missing serial numbers in stock.move.line


OthmaneGhandi:
Impacted versions:

Steps to reproduce:

In outgoing deliveries in V10: you find that n P1 was delivered and the serial numbers were listed in pack_lot_ids **(stock.pack.operation.lot)**

But in V11 : you only have the whole quantity is delivered without any details about the serial numbers of the product in **(stock.move.line)**

@pedrobaeza Thank you for your continued support

pedrobaeza:
This is covered for most cases, so you have to dig a bit more what is the weird case where the lot is not transferred to stock.move.line.

OthmaneGhandi:
@pedrobaeza 
It occurs for me for moves in **done** state.
In **post-migration.py** -> **create_stock_move_line** : you are using data from **stock_pack_operation** and I didn't see a JOIN with **stock_pack_operation_lot** is it normal ?

Am I missing something ?

pedrobaeza:
Lot information is taken from the quant linked to the stock.pack.operation:

https://github.com/OCA/OpenUpgrade/blob/29acc14cee567c597436cc4cbced6c9119c45741/addons/stock/migrations/11.0.1.1/post-migration.py#L198

`stock.pack.operation.lot` table can't be used as it changes cardinality of the SQL, provoking bad sumed quantities. More info in the commit description in https://github.com/OCA/OpenUpgrade/commit/91bba0cf0f4aac58ff812f59f38d7ea28eda7634#diff-fe7bbc3631adade92da541853a857311.

Check why don't you have `reserved_quant_id` filled in stock.move.operation.link.

OthmaneGhandi:
@pedrobaeza Indeed the `reserved_quant_id` is not filled, do you think that is something related to only my database ?

we can't have cases where `reserved_quant_id` is empty and still have serial numbers ?

I didn't see when this field get filled

pedrobaeza:
Maybe your DB was using outdated code (or are you coming from <v10?).

It's filled here:

https://github.com/odoo/odoo/blob/c7c68bab842aba5300f3a9deec28df6e17b40559/addons/stock/models/stock_picking.py#L645

and I studied the code some time ago and there's only one possibility of not being filled, and it's due to corrupted data (but not on normal flows).

alexis-via:
I confirm this very important bug. I've been badly hit by this on a migration of a medical project with a majority of products are tracked by lots.
[...]

PR #2030, opened 2019-09-20
[11.0][STOCK] Fix missing serial numbers in stock.move.line

OthmaneGhandi:

This PR try to fix my reported issue here #2000 about missing serial numbers in stock.move.line
@pedrobaeza what do you think about this PR 

Is my approach correct or I'm missing something ?

Thanks

--

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -112,11 +112,119 @@ def set_partially_available_state(env):
         WHERE partially_available AND state != 'done'"""
     )
 
+@openupgrade.logging()
+def fill_stock_operation_lot_with_move(env):
+    """link stock_pack_operation_lot to the appropriate move_id so we can later use it to retrieve serial numbers in stock.move.line"""
+    if not openupgrade.column_exists(env.cr, 'stock_pack_operation_lot', 'move_id'):
+        openupgrade.add_fields(env, [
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But the problem seems to be in your DB as stated. It's not a general problem. Or is it? If so, please explain the steps to reproduce the problem in regular DB for not having stock_move_operation_link populated.

OthmaneGhandi:
I reproduced these steps : **create a delivery order** (using a product that is tracked by serial number)
**I clicked on Mark as do** -> **reserve** and after that I clicked on the button (see picture) (specified inside which serial numbers to use: 3 serial numbers in my example)

serial numbers

**After that I validated the transfer => done state**

Current result : 
In `stock.pack.operation.lot` : I have the 3 serial numbers (3 records)
In `stock.pack.operation.link`: I have ONE record with qty 3
NB: In `stock.pack.operation.link`  the field `reserved_quant_id` is empty (it was filled before arriving to the done state)
=> That's why I will not get my 3 records with serial numbers with just checking `stock.pack.operation.link `and its connection with `stock.quant`

@pedrobaeza

pedrobaeza:
Oh, well, so the picking is in state reserved? Yeah, in that case you lose that reservation. There's no other way.

pedrobaeza:
Closing this as no further answer and no easy solution (this PR will have several side effects) and it's not the usual thing that you keep serial numbers put but not transferred.

PR #2031, opened 2019-09-20
[12.0][FIX] sale_timesheet: Set qty_delivered_method

pedrobaeza:
Value 'timesheet' is added by sale_timesheet module, and the field is new on v12 and thus, the sale.order.line records that qualify for it don't have proper value. We fill it by SQL here for avoiding to trigger other dependent computed fields.

cc @Tecnativa

### DIFF
--- a/addons/sale/migrations/12.0.1.1/post-migration.py
+++ b/addons/sale/migrations/12.0.1.1/post-migration.py
@@ -51,7 +51,7 @@ def fill_sale_order_line_qty_delivered_method(cr):
             SET qty_delivered_method = 'stock_move'
             FROM product_product pp
             LEFT JOIN product_template pt ON pp.product_tmpl_id = pt.id
-            WHERE sol.product_id = pp.id AND NOT sol.is_expense
+            WHERE sol.product_id = pp.id AND sol.is_expense IS NOT TRUE
                 AND pt.type IN ('consu', 'product')
             """
         )
@@ -62,7 +62,7 @@ def fill_sale_order_line_qty_delivered_method(cr):
         SET qty_delivered_method = 'timesheet'
         FROM product_product pp
         LEFT JOIN product_template pt ON pp.product_tmpl_id = pt.id
-        WHERE sol.product_id = pp.id AND NOT sol.is_expense
+        WHERE sol.product_id = pp.id AND sol.is_expense IS NOT TRUE
             AND pt.type = 'service' AND pt.service_type = 'timesheet'
         """
     )
--- a/addons/sale_timesheet/migrations/12.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/sale_timesheet/migrations/12.0.1.0/openupgrade_analysis_work.txt
@@ -19,8 +19,10 @@ sale_timesheet / project.sale.line.employee.map / sale_line_id (many2one)       
 sale_timesheet / project.task             / billable_type (selection)     : NEW selection_keys: ['employee_rate', 'no', 'task_rate']
 sale_timesheet / project.task             / sale_order_id (many2one)      : NEW relation: sale.order
 sale_timesheet / sale.order.line          / project_id (many2one)         : NEW relation: project.project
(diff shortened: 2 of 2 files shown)


carlosdauden (approved):

MiquelRForgeFlow (commented):
What about https://github.com/OCA/OpenUpgrade/blob/12.0/addons/sale/migrations/12.0.1.1/post-migration.py#L58L68?

pedrobaeza:
Well, it's not working... but I didn't remember that the other part exists. The only thing I can say is that this works now. We can remove that part if you want.

MiquelRForgeFlow:
Can you check **why** the other part doesn't work but this works? Also, can you check if the sale_stock part (qty_delivered_method = 'stock_move') also doesn't work if we need to do a similar like this?

pedrobaeza:
I'm not able to investigate it right now.

rafaelbn (approved):
:+1:

sergiocorato:
~Travis is green even if there is an errore [here](https://travis-ci.org/OCA/OpenUpgrade/builds/587449666#L3772) (I found it because in my fork it was [red](https://travis-ci.com/sergiocorato/OpenUpgrade/jobs/248185451#L3608))~ (not relevant)

StefanRijnhart:
Given that it's the exact same query, it seems dubious that the new version works whereas the existing version does not. I'm suspecting the criterium on `NOT sol.expense` which works for values that are SQL FALSE but not for SQL NULL (and note that false values are indeed left to NULL here: https://github.com/OCA/OpenUpgrade/blob/12.0/addons/sale/migrations/12.0.1.1/post-migration.py#L22-L27) May I suggest to replace with `sol.expense IS NOT TRUE` in either version to see if that works?

MiquelRForgeFlow:
Oh, I think you may have a point @StefanRijnhart

pedrobaeza:
I don't get that. What I know is that having the code in this PR, I get this working, and previously it wasn't, so how can it be that with what you are talking about the `is_expense` = NULL?

StefanRijnhart:
Just a hypothesis based on what I gather from the code. I'll have time to investigate later this week.

StefanRijnhart:
As much as I hate to say that I was right ;-) This PR does not fix the problem. It does not set the qty_delivered_method on any line until I modify the query from `NOT is_expense` to `is_expense IS NOT TRUE`. Compare the number of affected rows in both variants:

```
2019-12-20 12:34:57,076 9024 INFO sale_timesheet11 OpenUpgrade: sale_timesheet: post-migration script called with version 11.0.1.0
2019-12-20 12:34:57,077 9024 DEBUG sale_timesheet11 OpenUpgrade: 0 rows affected after 0:00:00.000678 running
        UPDATE sale_order_line sol
        SET qty_delivered_method = 'timesheet'
        FROM product_product pp,
            product_template pt
        WHERE pp.id = sol.product_id
            AND pt.id = pp.product_tmpl_id
            AND NOT sol.is_expense
            AND pt.type = 'service'
            AND pt.service_type = 'timesheet'
```

```
2019-12-20 12:37:09,452 9413 INFO sale_timesheet11 OpenUpgrade: sale_timesheet: post-migration script called with version 11.0.1.0
2019-12-20 12:37:09,454 9413 DEBUG sale_timesheet11 OpenUpgrade: 6 rows affected after 0:00:00.000935 running
        UPDATE sale_order_line sol
        SET qty_delivered_method = 'timesheet'
        FROM product_product pp,
            product_template pt
        WHERE pp.id = sol.product_id
            AND pt.id = pp.product_tmpl_id
            AND sol.is_expense IS NOT TRUE
            AND pt.type = 'service'
            AND pt.service_type = 'timesheet'
```

I suggest you fix the query in https://github.com/OCA/OpenUpgrade/blob/12.0/addons/sale/migrations/12.0.1.1/post-migration.py#L65 instead of adding this new query.

pedrobaeza:
@StefanRijnhart better to be right now in PR phase than later! Thanks for hunting down the problem. I have adapted the code to what you point, so now it should be good.

StefanRijnhart (approved):

PR #2032, opened 2019-09-20
[12.0][IMP] openupgrade_records: consider merged models

MiquelRForgeFlow:
This way, we can keep track also for merged models.

You can check https://github.com/OCA/OpenUpgrade/pull/2017 for what this fix provides. Basically, the 'stock.location.path' appears also as a renamed model to 'stock.rule'.

It will be more useful in v13, as there are more merged models to keep track.

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -19,8 +19,13 @@ def module_map(module):
         module, apriori.merged_modules.get(module, module))
 
 
-def model_map(model):
+def model_rename_map(model):
     return apriori.renamed_models.get(model, model)
+
+
+def model_map(model):
+    return apriori.renamed_models.get(
+        model, apriori.merged_models.get(model, model))
 
 
 def inv_model_map(model):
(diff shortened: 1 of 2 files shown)


StefanRijnhart (approved):

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -50,7 +55,7 @@ def compare_records(dict_old, dict_new, fields):
             if module_map(dict_old['module']) != dict_new['module']:
                 return False
         elif field == 'model':
-            if model_map(dict_old['model']) != dict_new['model']:
+            if model_rename_map(dict_old['model']) != dict_new['model']:
```
pedrobaeza, line 58: Why are you changing here the called method?

MiquelRForgeFlow: Well, because this one should still do the same as before. Another option should be not changing this one, but changing the other ones but it would imply more line diff.

pedrobaeza: Something doesn't fit me or I don't have enough information, because if you change this, why changing also the method `model_map` for fallbacking to `merged_models` mapping?

MiquelRForgeFlow: Because in the other cases where this method is called, you want to include the merged_models thing. It's only in this line where you don't want the merged_models thing being included.

pedrobaeza: OK, now go it, thanks.

pedrobaeza (approved):
You can proceed when green. Please update analysis files.

MiquelRForgeFlow:
Finally!

/ocabot merge

PR #2033, opened 2019-09-23
[FIX] website: Add aboutus and website.-prefixed redirections

yajo:
These redirections were missing.

--
@Tecnativa TT19453


pedrobaeza:
Isn't `website.` part stripped in another part of the script? Or another version?

yajo:
It's in this same version, and the real problem is not that it's stripped or not, the real problem is that maybe a web crawler or a random `<a>` element in any page/blog post/product/etc could contain a link to those URLs and become broken. By adding the redirection you fix those problems.

pedrobaeza (approved):
OK, got it

PR #2035, opened 2019-09-23
[FIX] hr_timesheet: Recompute fields for issues, only if issue_id exists

norlinhenrik:
Issue when account.analytic.account issue_id does not exist:

Current behavior before PR:
hr_timesheet post-migration will fail on recompute_tasks_from_issues_fields

hr_timesheet post-migration will succeed.

--

### DIFF
--- a/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
@@ -24,7 +24,7 @@ def migrate_project_issue_sheet(env):
     from issues.
     """
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
-    if not origin_issue_column:
+    if not openupgrade.column_exists(env.cr, 'account_analytic_line', origin_issue_column):
         return
     openupgrade.logged_query(
         env.cr, """
@@ -42,7 +42,7 @@ def recompute_tasks_from_issues_fields(env):
     its recomputation.
     """
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
-    if not origin_issue_column:
+    if not openupgrade.column_exists(env.cr, 'project_task', origin_issue_column):
         return
     env.cr.execute("SELECT id FROM project_task WHERE %s IS NOT NULL",
                    (AsIs(origin_issue_column), ))
--- a/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
@@ -42,6 +42,9 @@ def recompute_tasks_from_issues_fields(env):
     being inserted by SQL, don't have these fields computed, so we force here
     its recomputation.
     """
+    if not openupgrade.column_exists(
+            env.cr, 'account_analytic_line', 'issue_id'):
+        return
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
     env.cr.execute("SELECT id FROM project_task WHERE %s IS NOT NULL",
                    (AsIs(origin_issue_column), ))
--- a/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
@@ -23,10 +23,9 @@ def migrate_project_issue_sheet(env):
     and then fill `task_id` field in analytic lines with the tasks created
     from issues.
     """
-    if not openupgrade.column_exists(
-            env.cr, 'account_analytic_line', 'issue_id'):
-        return
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
+    if not origin_issue_column:
+        return
     openupgrade.logged_query(
         env.cr, """
         UPDATE account_analytic_line aal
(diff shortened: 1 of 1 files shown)


addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -42,6 +42,9 @@ def recompute_tasks_from_issues_fields(env):
     being inserted by SQL, don't have these fields computed, so we force here
     its recomputation.
     """
+    if not openupgrade.column_exists(
+            env.cr, 'account_analytic_line', 'issue_id'):
```
pedrobaeza: Please use `origin_issue_column`

pedrobaeza (changes requested):
Commit from a valid CLA

addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
norlinhenrik: The PR works well when issue modules have not been installed. The `if` statement is copied from the method `migrate_project_issue_sheet` just above. Feel free to change it if desired. I don't know how to use `origin_issue_column`.

pedrobaeza: @norlinhenrik `origin_issue_column` it's a variable just below. It's because the method is touching that column, not the other, so for consistency, you should check for that column.

addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -23,10 +23,9 @@ def migrate_project_issue_sheet(env):
     and then fill `task_id` field in analytic lines with the tasks created
     from issues.
     """
-    if not openupgrade.column_exists(
-            env.cr, 'account_analytic_line', 'issue_id'):
-        return
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
+    if not origin_issue_column:
```
pedrobaeza:
```suggestion
    if not openupgrade.column_exists(env.cr, 'account_analytic_line', origin_issue_column):
```

addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -43,6 +42,8 @@ def recompute_tasks_from_issues_fields(env):
     its recomputation.
     """
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
+    if not origin_issue_column:
```
pedrobaeza:
```suggestion
    if not openupgrade.column_exists(env.cr, 'project_task', origin_issue_column):
```

addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
norlinhenrik: :+1:

addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
norlinhenrik: :+1:

norlinhenrik:
@pedrobaeza I tried to approve the changes, but "Pull request authors can't approve their own pull request". Can you or someone else approve? Do I need to manually do these changes in my code?

pedrobaeza:
You don't have to approve it, but commit suggestion, and better add them in batch from "Files changed" tab. Any way, just do the changes if you have troubles with this option.

em230418:
This patch (based on this PR with @pedrobaeza suggestions)
```diff
diff --git a/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py b/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
index 46f351784..715fbef02 100644
--- a/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
@@ -23,10 +23,9 @@ def migrate_project_issue_sheet(env):
     and then fill `task_id` field in analytic lines with the tasks created
     from issues.
     """
-    if not openupgrade.column_exists(
-            env.cr, 'account_analytic_line', 'issue_id'):
-        return
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
+    if not openupgrade.column_exists(env.cr, 'account_analytic_line', origin_issue_column):
+        return
     openupgrade.logged_query(
         env.cr, """
         UPDATE account_analytic_line aal
@@ -43,6 +42,8 @@ def recompute_tasks_from_issues_fields(env):
     its recomputation.
     """
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
+    if not openupgrade.column_exists(env.cr, 'project_task', origin_issue_column):
+        return
     env.cr.execute("SELECT id FROM project_task WHERE %s IS NOT NULL",
                    (AsIs(origin_issue_column), ))
     tasks = env['project.task'].browse([x[0] for x in env.cr.fetchall()])
```
Fixes bug, where migrating fails on sql query:
```sql
SELECT id FROM project_task WHERE openupgrade_legacy_11_0_origin_issue_id IS NOT NULL
```
That comes from 
https://github.com/OCA/OpenUpgrade/blob/0ccfe463f488518ce1c72f3e749ecc5be4e14a08/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py#L45-L47

addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
```diff
@@ -23,10 +23,9 @@ def migrate_project_issue_sheet(env):
     and then fill `task_id` field in analytic lines with the tasks created
     from issues.
     """
-    if not openupgrade.column_exists(
-            env.cr, 'account_analytic_line', 'issue_id'):
-        return
     origin_issue_column = openupgrade.get_legacy_name('origin_issue_id')
+    if not openupgrade.column_exists(env.cr, 'account_analytic_line', origin_issue_column):
```
pedrobaeza: This column overpasses 80 col

pedrobaeza (changes requested):
And please squash all commits in one

norlinhenrik:
Because of little disk space I have deleted the local repo and done the later commits in the Github UI. Does the OpenUpgrade repo allow squash merging?

pedrobaeza:
Yeah, our bot doesn't still support to squash, but I'm thinking now that in this specific repo, the bot is not the more convenient to merge. Sorry for the extra work. Merging now.

PR #2043, opened 2019-09-27
[12.0][FIX] calendar: Don't remove used types

pedrobaeza:
Some data calendar event types have been removed in v12, and we remove them if it's safe to do it, but there's no constraint that avoids to be removed even if the types are used, so we switch to this algorithm for manually detecting if used.

cc @Tecnativa TT19531


rafaelbn (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

PR #2053, opened 2019-10-08
[12.0][IMP] openupgrade_records: show noupdate xmlid changes between versions

MiquelRForgeFlow:
This way, although the xmlid is the same between versions, if it has a noupdate change, from True to False, or from False to True, it will show the two old and new xmlids in order to detect this cases.

Updating a module doesn't change the noupdate flag from `ir.model.data`. It has to be changed manually in the migration scripts.

--


pedrobaeza:
I see no direct message is put that the noupdate has been switched, so it will be difficult to recognize it, isn't it?

MiquelRForgeFlow:
> so it will be difficult to recognize it, isn't it?

I don't think so. But let's wait for the update analysis and then say me again that 😏

MiquelRForgeFlow:
> direct message is put that the noupdate has been switched

Now it's put.

odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -304,10 +304,16 @@ def match(match_fields, match_type='direct'):
                 else:
                     column['domain'] = False
                     found['domain'] = False
+                column['noupdate_switched'] = False
+                if column['noupdate'] != found['noupdate']:
+                    found['noupdate_switched'] = True
+                else:
+                    found['noupdate_switched'] = False
```
pedrobaeza:
```suggestion
                    found['noupdate_switched'] = column['noupdate'] != found['noupdate']
```

MiquelRForgeFlow: right 😁

pedrobaeza (approved):

pedrobaeza (approved):

PR #2058, opened 2019-10-15
[IMP] sale,sale_management,account: add empty sections

pedrobaeza:
If you have a sales order/sales template/invoice in v11 with this layout:

- Section 1
  * Line 1
  * Line 2
- Section 2
  * Line 3
* Line 4

It ends up with these lines:

- Section 1
- Line 1
- Line 2
- Section 2
- Line 3
- Line 4

so Line 4 seems that belongs to Section 2.

For avoiding this, we add an extra blank section for lines without v11 layout category.
Also for avoiding noise, we remove that section immediately for sales orders that have
no layout category in any of their lines.

Final result:

- Section 1
- Line 1
- Line 2
- Section 2
- Line 3
-
- Line 4

@Tecnativa TT20064


MiquelRForgeFlow (approved):
code LGTM, your explanation makes sense

PR #2067, opened 2019-10-24
[IMP] base: Continue if a partner fails VAT validation

yajo:
This method took about 8h in a database with 280000 partners. It's disgusting when all that time is wasted just because one partner failed VAT validation, and it's even more disgusting when you don't know which one failed.

Now, in case a partner cannot sync its commercial fields, that won't abort the whole migration. Instead, in the log you'll be able to see the whole exception message and the partner ID for each one of those that failed. At least, you'll be able to let them as they are, or to fix all migration problems at once before retrying the migration.

--
@Tecnativa TT18838

### DIFF
--- a/openerp/openupgrade/deferred_70.py
+++ b/openerp/openupgrade/deferred_70.py
@@ -28,6 +28,7 @@ being run multiple times on the same database.
 """
 
 import logging
+from openupgradelib.openupgrade import savepoint
 from openerp import SUPERUSER_ID
 
 logger = logging.getLogger("OpenUpgrade")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
VAT check was already in 6.1? If so, we can't simply bypass this. And I'm not sure either, as you are leaving partners without syncing information, provoking data inconsistency.

yajo:
However the VAT check implementation must be different, because I had VAT errors when migrating to v7 with VATs that were present in v6.1. Also, it's possible that a partner had a wrong vat before `base_vat` was installed in v6.1, and then that would block the migration.

We still have the option to raise a final exception if some record failed in the process. That wouldn't leave inconsistencies, but at least you wouldn't need to remigrate after each failure because you'd have all failures details in the logs.

pedrobaeza:
Explore the option to pass a context for avoiding the vat check, and if not, the option for aborting at the end seems OK for me.

yajo:
I added both.

addons/base_vat/base_vat.py
```diff
@@ -115,6 +115,9 @@ def button_check_vat(self, cr, uid, ids, context=None):
         return True
 
     def check_vat(self, cr, uid, ids, context=None):
+        context = context or {}
+        if context.get("openupgrade_disable_vat_validation"):
```
pedrobaeza: We don't modify modules code in OpenUpgrade. I'm not sure if allowing this exception... @mreficent what do you think?

MiquelRForgeFlow: We don't have clients at v6.1, so do as you like. But if you want this, at least add a comment saying it's an openupgrade fix.

yajo: I think that when it's necessary it must be allowed. Otherwise, separating OpenUpgrade form OCB wouldn't make sense.

yajo: Comment added.

yajo:
I disabled the workaround. Now it is just adding the logs for you to fix those problems.

pedrobaeza (approved):

PR #2068, opened 2019-10-24
[FIX] workflow: Index m2o cols with ondelete

yajo:
Without this patch, executing `openupgrade.delete_model_workflow()` (as in `procurement`'s pre-migration) is sloooooow.

Much better now.

--
@Tecnativa TT18838


yajo:
Oops, I shouldn't modify addons code; closinig, will do better later.

PR #2069, opened 2019-10-24
[12.0][IMP] OpenUpgrade: don't delete xmlids linked to nothing

MiquelRForgeFlow:
The xmlids linked to nothing should not be deleted, as maybe they could be used later.

--

### DIFF
--- a/odoo/addons/base/models/ir_model.py
+++ b/odoo/addons/base/models/ir_model.py
@@ -1715,6 +1715,9 @@ class IrModelData(models.Model):
                                 res_id, model)
                         # /OpenUpgrade
                     else:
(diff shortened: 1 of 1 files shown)


odoo/addons/base/models/ir_model.py
```diff
@@ -1715,6 +1715,9 @@ def _process_end(self, modules):
                                 res_id, model)
                         # /OpenUpgrade
                     else:
+                        # OpenUpgrade: don't delete xmlids, use data_cleanup instead
```
pedrobaeza:
```suggestion
                        # OpenUpgrade: don't delete xmlids. Use database_cleanup instead
```

pedrobaeza (approved):

GSLabIt (approved):

PR #2089, opened 2019-11-21
[IMP] Improve delete_quants_for_consumable_performance

hugosantosred:
Description of the issue/feature this PR addresses:  Call first drop_slow_constraint just before delete_quants_for_consumable for improved delete performance
 
Current behavior before PR: If there any consumable quants in the database when migrating to v11 the database call that deletes that quants is slow.

Desired behavior after PR is merged: Using drop_slow_constraint before delete_quants_for_consumable improves the deletion performance

--


pedrobaeza (approved):
Indeed it was conceived for that, but focus on end-migration quant removal. I haven't noticed till now because my DBs don't have too much consumables. Thanks for noting!

PR #2093, opened 2019-11-28
[11.0][ADD] mass_mailing_event: Nothing to do

pedrobaeza:
cc @Tecnativa TT17665


MiquelRForgeFlow (approved):
Normal, this module is new in v12.

PR #2100, opened 2019-12-02
[9.0][IMP] sale: Avoid costly ORM computations

pedrobaeza:
This optimizes the speed of migration of sale module in 2 ways:

- Replicating the compute algorithm for `price_tax`, `price_total` and `price_subtotal` in sale.order.line and storing it via SQL for avoiding the triggering of other dependent computed fields.
- Emulating the logic of the rest of the computed fields through SQL.

cc @Tecnativa TT18838

### DIFF
--- a/addons/sale/migrations/9.0.1.0/post-migration.py
+++ b/addons/sale/migrations/9.0.1.0/post-migration.py
@@ -4,6 +4,7 @@
 # © 2016 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
 # © 2016 Opener B.V. - Stefan Rijnhart
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from pprint import pformat
 
 from openupgradelib import openupgrade
 from openerp import api, SUPERUSER_ID
@@ -85,14 +86,51 @@ def set_so_line_computed_rest(env):
     """Emulate the computation of the rest of the computed fields through
     equivalente SQL querys.
     """
+    uom_precision = env['decimal.precision'].precision_get('Product Unit of Measure')
     openupgrade.logged_query(
         env.cr, """
         UPDATE sale_order_line
         SET price_reduce = CASE
-            WHEN abs(product_uom_qty) > 0.00001
+            -- ROUND() missing in original code, but it's a bug
+            WHEN ROUND(product_uom_qty, %(uom_precision)s) != 0
                 THEN price_subtotal / product_uom_qty
             ELSE 0.0
-        END""")
+        END""",
+        {"uom_precision": uom_precision})
+    # Fail if there are UoM mismatches
+    openupgrade.logged_query(
+        env.cr,
+        """
+            SELECT
+                rel.order_line_id,
+                sol_u.id AS order_line_uom_id,
+                sol_u.category_id AS order_line_uom_category_id,
+                sol.product_id AS order_line_product_id,
+                rel.invoice_line_id,
+                ail_u.id AS invoice_line_uom_id,
+                ail_u.category_id AS invoice_line_uom_category_id,
+                ail_pp.id AS invoice_line_product_id,
+                ail_pt.name AS invoice_line_product_template_name
+            FROM sale_order_line_invoice_rel rel
+                JOIN sale_order_line sol ON
+                    rel.order_line_id = sol.id
+                JOIN account_invoice_line ail ON
+                    rel.invoice_line_id = ail.id
+                JOIN product_product ail_pp ON ail_pp.id = ail.product_id
+                JOIN product_template ail_pt ON ail_pt.id = ail_pp.product_tmpl_id
+                JOIN account_invoice ai ON ai.id = ail.invoice_id
+                JOIN product_uom ail_u on (ail_u.id = ail.uom_id)
+                JOIN product_uom sol_u on (sol_u.id = sol.product_uom)
+            WHERE
+                ail_u.category_id != sol_u.category_id
+        """
+    )
+    uom_category_mismatches = env.cr.dictfetchall()
+    if uom_category_mismatches:
+        raise Exception(
+            "Found these mismatching UoM in related sale.order.line "
+            "and account.invoice.line records: %s" %
+            pformat(uom_category_mismatches))
     openupgrade.logged_query(
         env.cr, """
         UPDATE sale_order_line sol
@@ -114,7 +152,7 @@ def set_so_line_computed_rest(env):
                 JOIN account_invoice ai ON ai.id = ail.invoice_id
                 JOIN product_uom u on (u.id = ail.uom_id)
                 JOIN product_uom u2 on (u2.id = sol2.product_uom)
-            WHERE ai.state != ('cancel')
+            WHERE ai.state != 'cancel'
             GROUP BY rel.order_line_id
         ) AS sub
         WHERE sol.id = sub.order_line_id""")
@@ -139,36 +177,52 @@ def set_so_line_computed_rest(env):
         env.cr, """
         UPDATE sale_order_line sol
         SET invoice_status = CASE
-            WHEN sol.state IN ('sale', 'done') THEN 'no'
-            WHEN abs(sol.qty_to_invoice) > 0.00001 THEN 'to invoice'
+            WHEN sol.state NOT IN ('sale', 'done') THEN 'no'
+            WHEN ROUND(sol.qty_to_invoice, %(uom_precision)s) != 0 THEN 'to invoice'
             WHEN sol.state = 'sale' AND pt.invoice_policy = 'order' AND
-                (sol.qty_delivered - sol.product_uom_qty) > 0.00001
+                ROUND(sol.qty_delivered, %(uom_precision)s) > ROUND(sol.product_uom_qty, %(uom_precision)s)
                     THEN 'upselling'
-            WHEN sol.qty_invoiced >= sol.product_uom_qty THEN 'invoiced'
+            WHEN ROUND(sol.qty_invoiced, %(uom_precision)s) >= ROUND(sol.product_uom_qty, %(uom_precision)s)
+                THEN 'invoiced'
             ELSE 'no'
         END
         FROM sale_order_line sol2
             JOIN product_product pp ON pp.id = sol2.product_id
             JOIN product_template pt ON pt.id = pp.product_tmpl_id
-        WHERE sol2.id = sol.id""")
+        WHERE sol2.id = sol.id""",
+        {"uom_precision": uom_precision})
     openupgrade.logged_query(
         env.cr, """
         UPDATE sale_order so
         SET invoice_status = CASE
             WHEN so.state NOT IN ('sale', 'done') THEN 'no'
-            WHEN id = ANY(
-                SELECT order_id FROM sale_order_line sol
-                WHERE sol.order_id = so.id
-                    AND sol.invoice_status = 'to invoice'
-            ) THEN 'to invoice'
-            WHEN id = ALL(SELECT order_id FROM sale_order_line sol
-                WHERE sol.order_id = so.id
-                    AND sol.invoice_status IN ('invoiced', 'upselling'))
-                THEN 'upselling'
-            WHEN id = ALL(SELECT order_id FROM sale_order_line sol
-                WHERE sol.order_id = so.id
-                    AND sol.invoice_status = 'invoiced')
+            -- Check if at least there's one line to invoice
+            WHEN 'to invoice' = ANY(
+                    SELECT invoice_status FROM sale_order_line sol
+                    WHERE sol.order_id = so.id
+                )
+                THEN 'to invoice'
+            -- Check if all lines are invoiced
+            WHEN 'invoiced' = ALL(
+                    SELECT invoice_status
+                    FROM sale_order_line sol
+                    WHERE sol.order_id = so.id
+                )
                 THEN 'invoiced'
+            WHEN
+                -- Check if all lines are either invoiced or upselling
+                0 = (
+                    SELECT COUNT(*)
+                    FROM sale_order_line sol
+                    WHERE sol.order_id = so.id
+                    AND sol.invoice_status NOT IN ('invoiced', 'upselling')
+                -- ... and that at least one is upselling
+                ) AND 'upselling' = ANY(
+                    SELECT invoice_status
+                    FROM sale_order_line sol
+                    WHERE sol.order_id = so.id
+                )
+                THEN 'upselling'
             ELSE 'no'
         END""")
     openupgrade.logged_query(
--- a/addons/sale/migrations/9.0.1.0/post-migration.py
+++ b/addons/sale/migrations/9.0.1.0/post-migration.py
@@ -55,6 +55,131 @@ def set_crm_team_message_types(env):
             (4, env.ref('sale.mt_salesteam_invoice_created').id)]})
 
 
+@openupgrade.logging()
+def set_so_line_amounts(env):
+    """We replicate here the code of the compute function and set the values
+    finally via SQL for avoiding the trigger of the rest of the computed
+    fields that depends on these fields.
+    """
+    lines = env['sale.order.line'].search([])
+    for line in lines:
+        price = line.price_unit * (1 - (line.discount or 0.0) / 100.0)
+        taxes = line.tax_id.compute_all(
+            price, line.order_id.currency_id, line.product_uom_qty,
+            product=line.product_id, partner=line.order_id.partner_id)
+        env.cr.execute("""
+            UPDATE sale_order_line
+            SET price_tax = %s,
+                price_total = %s,
+                price_subtotal = %s
+            WHERE id = %s""", (
+                taxes['total_included'] - taxes['total_excluded'],
+                taxes['total_included'],
+                taxes['total_excluded'],
+                line.id,
+            )
+        )
+
+
+def set_so_line_computed_rest(env):
+    """Emulate the computation of the rest of the computed fields through
+    equivalente SQL querys.
+    """
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line
+        SET price_reduce = CASE
+            WHEN abs(product_uom_qty) > 0.00001
+                THEN price_subtotal / product_uom_qty
+            ELSE 0.0
+        END""")
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET qty_invoiced = sub.qty
+        FROM (
+            SELECT rel.order_line_id,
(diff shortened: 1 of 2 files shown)


addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+                END) AS qty
+            FROM sale_order_line_invoice_rel rel
+                JOIN sale_order_line sol2 ON
+                    rel.order_line_id = sol2.id
+                JOIN account_invoice_line ail ON
+                    rel.invoice_line_id = ail.id
+                JOIN account_invoice ai ON ai.id = ail.invoice_id
+                JOIN product_uom u on (u.id = ail.uom_id)
+                JOIN product_uom u2 on (u2.id = sol2.product_uom)
+            WHERE ai.state != ('cancel')
```
yajo: I know it's the same but feels weird like it is:

```suggestion
            WHERE ai.state != 'cancel'
```

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+            ELSE 0.0
+        END""")
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET qty_invoiced = sub.qty
+        FROM (
+            SELECT rel.order_line_id,
+                SUM(CASE
+                    WHEN ai.type = 'out_invoice'
```
yajo, line 141:
```suggestion
                    WHEN ai.type = 'out_invoice' AND u.category_id = u2.category_id
```

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+                        THEN ail.quantity / u.factor * u2.factor
+                    WHEN ai.type = 'out_refund'
```
yajo, line 143:
```suggestion
                    WHEN ai.type = 'out_refund' AND u.category_id = u2.category_id
```

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+                        THEN -ail.quantity / u.factor * u2.factor
+                    ELSE 0.0
```
yajo, line 145:
```suggestion
                    WHEN ai.type IN ('out_invoice', 'out_refund') AND u.category_id != u2.category_id
                        THEN ail.quantity
                    ELSE 0.0
```

This suggestion is in combination with the 2 above, to match this code: https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/product/product.py#L120-L124

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+        FROM sale_order_line sol2
+            JOIN sale_order so ON so.id = sol2.order_id
+            JOIN product_product pp ON pp.id = sol2.product_id
+            JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        WHERE sol2.id = sol.id""")
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET invoice_status = CASE
+            WHEN sol.state IN ('sale', 'done') THEN 'no'
```
yajo:
```suggestion
            WHEN sol.state NOT IN ('sale', 'done') THEN 'no'
```

Little big bug: https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L493-L494

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+
+def set_so_line_computed_rest(env):
+    """Emulate the computation of the rest of the computed fields through
+    equivalente SQL querys.
+    """
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line
+        SET price_reduce = CASE
+            WHEN abs(product_uom_qty) > 0.00001
```
yajo:
```suggestion
            WHEN product_uom_qty != 0
```

Looking at upstream function, it doesn't apply any rounding protection, so I guess you shouldn't here also:

https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L560

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+            JOIN sale_order so ON so.id = sol2.order_id
+            JOIN product_product pp ON pp.id = sol2.product_id
+            JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        WHERE sol2.id = sol.id""")
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET invoice_status = CASE
+            WHEN sol.state IN ('sale', 'done') THEN 'no'
+            WHEN abs(sol.qty_to_invoice) > 0.00001 THEN 'to invoice'
```
yajo: Obtain precision before with this code somewhere above:

```py
uom_precision = self.env['decimal.precision'].precision_get('Product Unit of Measure')
```

Then apply this suggestion and later add `uom_precision` as query arg:

```suggestion
            WHEN ROUND(sol.qty_to_invoice, %(uom_precision)s) != 0 THEN 'to invoice'
```

This will be more exactly equivalent to:
https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L495-L496

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+            WHEN sol.state = 'sale' AND pt.invoice_policy = 'order' AND
+                (sol.qty_delivered - sol.product_uom_qty) > 0.00001
```
yajo: Following above comment:

```suggestion
                ROUND(sol.qty_delivered, %(uom_precision)s) > ROUND(sol.product_uom_qty, %(uom_precision)s)
```

Ref code: https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L497-L499

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+                    THEN 'upselling'
+            WHEN sol.qty_invoiced >= sol.product_uom_qty THEN 'invoiced'
```
yajo: This one has a little bug also compared to: https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L500-L501

Change to fix it:

```suggestion
            WHEN ROUND(sol.qty_invoiced, %(uom_precision)s) = ROUND(sol.product_uom_qty, %(uom_precision)s) THEN 'invoiced'
```

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +55,139 @@ def set_crm_team_message_types(env):
...
+                    AND sol.invoice_status = 'to invoice'
+            ) THEN 'to invoice'
+            WHEN id = ALL(SELECT order_id FROM sale_order_line sol
+                WHERE sol.order_id = so.id
+                    AND sol.invoice_status IN ('invoiced', 'upselling'))
+                THEN 'upselling'
+            WHEN id = ALL(SELECT order_id FROM sale_order_line sol
+                WHERE sol.order_id = so.id
+                    AND sol.invoice_status = 'invoiced')
+                THEN 'invoiced'
```
yajo: According to reference code, you have to switch the order of these 2 conditions:

https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L78-L81

yajo (changes requested):
In-depth code review, comparing with replaced python code.

---

I have been investigating a little bit about the rounding. I have checked docs from Postgres 8.2 and 12 (although links below are for 12).

According to [PostgreSQL docs](https://www.postgresql.org/docs/12/functions-math.html) (I checked both 8.2 and 12.0 docs):


> Function | Return Type | Description | Example | Result
> -- | -- | -- | -- | --
> round(v           numeric, s int) | numeric | round to s decimal           places | round(42.4382, 2) | 42.44

You can see that when you call `ROUND(float_value, decimals)`, the return type will always be `NUMERIC`. So, [about that type the docs say that **it is exact**](https://www.postgresql.org/docs/12/datatype-numeric.html), so we can trust that `ROUND()` will return what we expect.

Given those facts, you could replace the `ABS(float_value) > 0.00001` trick with `ROUND(product_uom_qty, %s) != 0` and it should work better, being more readable and letting you obtain the exact expected decimal precision in all cases.

Some suggestions below are based on this.

---

OTOH, the result of this in a DB with 641599 sale order lines is going down from ~12h migrating `sale` to ~25min. Big kudos! :tada:

yajo:
OK I applied all my own suggestions!



Ready to review.

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -85,11 +85,12 @@ def set_so_line_computed_rest(env):
     """Emulate the computation of the rest of the computed fields through
     equivalente SQL querys.
     """
+    uom_precision = self.env['decimal.precision'].precision_get('Product Unit of Measure')
     openupgrade.logged_query(
         env.cr, """
         UPDATE sale_order_line
         SET price_reduce = CASE
-            WHEN abs(product_uom_qty) > 0.00001
+            WHEN product_uom_qty != 0
```
pedrobaeza: Shouldn't you round this as well?

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -100,10 +101,12 @@ def set_so_line_computed_rest(env):
         FROM (
             SELECT rel.order_line_id,
                 SUM(CASE
-                    WHEN ai.type = 'out_invoice'
+                    WHEN ai.type = 'out_invoice' AND u.category_id = u2.category_id
```
pedrobaeza: This is not something reflected on ORM code and trying to prevent an error that should raise any way in other places (incompatible UoMs).

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -139,18 +142,19 @@ def set_so_line_computed_rest(env):
...
-            WHEN sol.state IN ('sale', 'done') THEN 'no'
-            WHEN abs(sol.qty_to_invoice) > 0.00001 THEN 'to invoice'
+            WHEN sol.state NOT IN ('sale', 'done') THEN 'no'
+            WHEN ROUND(sol.qty_to_invoice, %(uom_precision)s) != 0 THEN 'to invoice'
             WHEN sol.state = 'sale' AND pt.invoice_policy = 'order' AND
-                (sol.qty_delivered - sol.product_uom_qty) > 0.00001
+                ROUND(sol.qty_delivered, %(uom_precision)s) > ROUND(sol.product_uom_qty, %(uom_precision)s)
                     THEN 'upselling'
-            WHEN sol.qty_invoiced >= sol.product_uom_qty THEN 'invoiced'
+            WHEN ROUND(sol.qty_invoiced, %(uom_precision)s) = ROUND(sol.product_uom_qty, %(uom_precision)s) THEN 'invoiced'
```
pedrobaeza: This should be `>=`, not `=`

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -139,18 +142,19 @@ def set_so_line_computed_rest(env):
...
             ELSE 'no'
         END
         FROM sale_order_line sol2
             JOIN product_product pp ON pp.id = sol2.product_id
             JOIN product_template pt ON pt.id = pp.product_tmpl_id
-        WHERE sol2.id = sol.id""")
+        WHERE sol2.id = sol.id""",
+        {"uom_precision": uom_precision})
```
pedrobaeza, line 193: Should this be handled by psycopg2 argument substitution or directly by Python?

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -161,14 +165,14 @@ def set_so_line_computed_rest(env):
...
             ) THEN 'to invoice'
-            WHEN id = ALL(SELECT order_id FROM sale_order_line sol
-                WHERE sol.order_id = so.id
-                    AND sol.invoice_status IN ('invoiced', 'upselling'))
-                THEN 'upselling'
             WHEN id = ALL(SELECT order_id FROM sale_order_line sol
                 WHERE sol.order_id = so.id
                     AND sol.invoice_status = 'invoiced')
                 THEN 'invoiced'
+            WHEN id = ALL(SELECT order_id FROM sale_order_line sol
```
pedrobaeza: This is sorted this way on purpose, so you are altering the logic. Take into account that `ALL` works for the select inside, so it doesn't work the same as in python.

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: I'm trying to imitate the python method as close as possible, and it doesn't round: https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/sale/sale.py#L560

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: The ORM has this code:

https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/product/product.py#L120-L124

It contemplates a situation where it's possible for incompatible UoM categories to land.

Indeed this context key that is required wouldn't exist during migration and it would fail instead. In any case we were losing also the exception with previous implementation.

So do I leave it like it is, or do I add the check and raise an exception when incompatible UoMs are found? (I'd have to search for them via SQL).

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: The result is an integer, so it should be OK as it is AFAIK. See:

https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/decimal_precision/decimal_precision.py#L10-L15

https://github.com/OCA/OpenUpgrade/blob/bd1a1d7b04b1d5ea5d94b931443d61fc71d810a2/addons/decimal_precision/decimal_precision.py#L24-L28

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: Now that you say that, I'm seeing both of us were wrong. Fixed now, thanks!

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: Oops, fixed.

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: According to private conversation, adding `ROUND()` because it's a bug upstream, fixed only in future versions.

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: According to private conversation, adding exception.

yajo (approved):

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +56,186 @@ def set_crm_team_message_types(env):
...
+                line.id,
+            )
+        )
+
+
+def set_so_line_computed_rest(env):
+    """Emulate the computation of the rest of the computed fields through
+    equivalente SQL querys.
+    """
+    uom_precision = self.env['decimal.precision'].precision_get('Product Unit of Measure')
```
pedrobaeza:
```suggestion
    uom_precision = env['decimal.precision'].precision_get('Product Unit of Measure')
```

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +56,186 @@ def set_crm_team_message_types(env):
...
+        END""",
+        {"uom_precision": uom_precision})
+    # Fail if there are UoM mismatches
+    openupgrade.logged_query(
+        env.cr,
+        """
+            SELECT
+                rel.order_line_id,
+                sol_u.id AS order_line_uom_id,
+                sol_u.category_id AS order_line_uom_category_id,
```
pedrobaeza, line 107: I think you don't need to put the categories, and the SQL is vastly simplified. I would put maybe sales order number or product name though

addons/sale/migrations/9.0.1.0/post-migration.py
```diff
@@ -55,12 +56,186 @@ def set_crm_team_message_types(env):
...
+            JOIN sale_order so ON so.id = sol2.order_id
+            JOIN product_product pp ON pp.id = sol2.product_id
+            JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        WHERE sol2.id = sol.id""")
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE sale_order_line sol
+        SET invoice_status = CASE
+            WHEN sol.state NOT IN ('sale', 'done') THEN 'no'
+            WHEN ROUND(sol.qty_to_invoice, %(uom_precision)s) != 0 THEN 'to invoice'
```
pedrobaeza, line 181: flake8 is checked in migration scripts AFAIK, but with 120 cols limit.

addons/sale/migrations/9.0.1.0/post-migration.py
yajo: I prefer to keep the categories because it's actually gonna be the data that triggers this exception.

OTOH I'll add the other data.

yajo:
All attended.

PR #2102, opened 2019-12-04
[12.0][MIG] website_sale_wishlist: Migration script

sergio-teruel:
cc @Tecnativa TT19371
@pedrobaeza @Yajo Can you review?


addons/website_sale_wishlist/migrations/12.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,5 @@
+---Models in module 'website_sale_wishlist'---
+---Fields in module 'website_sale_wishlist'---
+website_sale_wishlist / product.wishlist         / session (char)                : DEL
```
pedrobaeza, line 3: As this has been removed, you have to remove the entries that doesn't have `partner_id` or the `partner_id` is a public user (not sure if this is a real case - @Yajo can you say?).

addons/website_sale_wishlist/migrations/12.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="product_wishlist_rule" model="ir.rule">
+    <field name="domain_force">["|", ('partner_id','=', user.partner_id.id), "&amp;", ("partner_id", "=", False), ("session", "=", user.current_session)]</field>
```
pedrobaeza: You have applied the script with the paths reversed. The correct one is:
```suggestion
    <field name="domain_force">[('partner_id','=', user.partner_id.id)]</field>
```

addons/website_sale_wishlist/migrations/12.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,7 @@
...
+    <field name="groups" eval="[(4, ref('base.group_portal')),(4, ref('base.group_public')), (4, ref('base.group_user'))]"/>
```
pedrobaeza: If you apply the script as is (with the paths correctly sorted), you will get:
```xml
   <field name="groups" eval="[(4, ref('base.group_portal')), (4, ref('base.group_user'))]"/>
```
but this won't make the desire effect, as we want to remove the public group from this rule, and `(4, x)` operations only add without replacing, and thus you need to change the expression to:
```xml
   <field name="groups" eval="[(3, ref('base.group_public')), ]"/>
```

pedrobaeza (changes requested):

addons/website_sale_wishlist/migrations/12.0.1.0/openupgrade_analysis_work.txt
yajo: I see, it seems Odoo stores now `wishlist_ids` into the session object itself.

It's weird, I don't know how will they garbage-collect them now...

Well, anyway it seems like session-related wishlist products will be lost, so I guess the best way is to do what you say in a pre-migration, but also add `WHERE session IS NOT NULL`.

addons/website_sale_wishlist/migrations/12.0.1.0/noupdate_changes.xml
MiquelRForgeFlow: an alternative is:
```xml
<field name="groups" eval="[(5,), (4, ref('base.group_portal')), (4, ref('base.group_user'))]"/>
```

sergio-teruel:
Ok... Changes done!!

addons/website_sale_wishlist/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    group_public = env.ref('base.group_public')
+    partners = group_public.users.mapped('partner_id')
+    sql = """
+    DELETE from product_wishlist
+    WHERE session IS NOT NULL AND (partner_id IS NULL OR parter_id in %s)
```
pedrobaeza:
```suggestion
    WHERE session IS NOT NULL AND (partner_id IS NULL OR partner_id in %s)
```

pedrobaeza (changes requested):

sergio-teruel:
Changes done!!

pedrobaeza (approved):

pedrobaeza (changes requested):
This is giving an error:

https://travis-ci.org/OCA/OpenUpgrade/builds/622925588#L8307

I think this is probably as you are using an intermediate variable.

addons/website_sale_wishlist/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    group_public = env.ref('base.group_public')
+    partners = group_public.users.mapped('partner_id')
```
yajo: I think you should add here also this, just in case there are multiple websites or they have different public users:

```suggestion
    partners = group_public.users.mapped('partner_id') | env['website'].search([]).mapped('partner_id')
```

addons/website_sale_wishlist/migrations/12.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+    openupgrade.logged_query(
+        env.cr, """DELETE from product_wishlist WHERE session IS NOT NULL AND
+                   (partner_id IS NULL OR partner_id in %s)""",
+        tuple(partners.ids)
```
yajo: AFAIK you're transforming the list into a tuple, instead of using a tuple with the 1st element being a list:

```suggestion
        (partners.ids,)
```

yajo (changes requested):

sergio-teruel:
Changes done

yajo (approved):
@pedrobaeza please update your review

pedrobaeza (approved):

PR #2104, opened 2019-12-05
[IMP] purchase: Optimize migration in SQL

yajo:
There are plenty of new computed stored fields in `purchase`.

Here I imitate ORM methods in SQL to make them much faster to migrate. Inspired in https://github.com/OCA/OpenUpgrade/pull/2100 but adapted for purchase.

@Tecnativa TT18838

--

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,11 +43,12 @@ script:
     - cat openerp/addons/base/migrations/*/tests/*.yml addons/*/migrations/*/tests/*.yml > ../test_data80.yml
     # Roundtrip to previous release to update the test database with additional test data
     # Loading and committing test data without triggering post tests needs https://github.com/odoo/odoo/pull/13146. This is now included in OpenUpgrade.
+    - old_commit="$(git rev-parse HEAD)"
     - git fetch --depth 2 origin 8.0
     - git reset --hard `git ls-remote |grep refs/heads/8.0 |awk '{print $1}'`
     - pip install -q -r requirements.txt
     - $ODOO --database=$DB --test-file=`readlink -f ../test_data80.yml` --test-commit --stop-after-init
-    - git reset -q --hard $TRAVIS_COMMIT
+    - git reset -q --hard $old_commit
     # Install Python requirements of target release
     - pip install -q -r requirements.txt
     # don't use pypi's openupgradelib, but the one from source to avoid choking
--- a/addons/purchase/migrations/9.0.1.2/end-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/end-migration.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Tecnativa - Jairo Llopis
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.logging()
+def deferred_pol_qty_received_compute(env):
+    """Complete migration of purchase.order.line#qty_received
+
+    Skip records that were optimized in SQL in post-migration.
+    """
+    # Execute https://bit.ly/2PgUKaT for matching records
+    pol_todo = env["purchase.order.line"].search([
+        ("order_id.state", "in", ['purchase', 'done']),
+        ("product_id.type", "in", ['consu', 'product']),
+    ])
+    pol_todo._compute_qty_received()
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    deferred_pol_qty_received_compute(env)
--- a/addons/purchase/migrations/9.0.1.2/end-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/end-migration.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2019 Tecnativa - Jairo Llopis
+# Copyright 2019 Tecnativa - Pedro M. Baeza
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from openupgradelib import openupgrade
@@ -7,14 +8,16 @@ from openupgradelib import openupgrade
 
 @openupgrade.logging()
 def deferred_pol_qty_received_compute(env):
-    """Complete migration of purchase.order.line#qty_received
-
-    Skip records that were optimized in SQL in post-migration.
+    """Complete migration of purchase.order.line#qty_received for products
+    with BoMs. We don't go further in initial filtering, as this number should
+    be low and it's not going to be heavy to be computed.
     """
-    # Execute https://bit.ly/2PgUKaT for matching records
+    if 'mrp.bom' not in env:
+        return  # mrp not installed
     pol_todo = env["purchase.order.line"].search([
         ("order_id.state", "in", ['purchase', 'done']),
         ("product_id.type", "in", ['consu', 'product']),
+        ("product_id.product_tmpl_id.bom_ids", "!=", False),
     ])
     pol_todo._compute_qty_received()
 
--- a/addons/purchase/migrations/9.0.1.2/post-migration.py
+++ b/addons/purchase/migrations/9.0.1.2/post-migration.py
@@ -3,6 +3,8 @@
 # © 2015 Serpent Consulting Services Pvt. Ltd.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from pprint import pformat
+
 from openupgradelib import openupgrade
 from openerp import api, SUPERUSER_ID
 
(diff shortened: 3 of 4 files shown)


addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -92,9 +94,182 @@ def account_properties(cr):
...
+                taxes['total_excluded'],
+                line.id,
+            )
+        )
+
+
+@openupgrade.logging()
+def set_po_line_computed_rest(env):
+    """Emulate the computation of the rest of the computed fields through
+    equivalente SQL queries.
```
pedrobaeza:
```suggestion
    equivalent SQL queries.
```

addons/purchase/migrations/9.0.1.2/end-migration.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@openupgrade.logging()
+def deferred_pol_qty_received_compute(env):
+    """Complete migration of purchase.order.line#qty_received
+
+    Skip records that were optimized in SQL in post-migration.
+    """
+    # Execute https://bit.ly/2PgUKaT for matching records
+    pol_todo = env["purchase.order.line"].search([
```
pedrobaeza, line 17: This search is not taking into account BoMs, so it's basically recomputing almost everything.

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -92,9 +94,182 @@ def account_properties(cr):
...
+def set_po_line_amounts(env):
+    """We replicate here the code of the compute function and set the values
+    finally via SQL for avoiding the trigger of the rest of the computed
+    fields that depends on these fields.
+
+    Replaces code from https://bit.ly/2rgpTTV
+    """
+    lines = env['purchase.order.line'].search([])
+    for line in lines:
+        price = line.price_unit * (1 - (line.discount or 0.0) / 100.0)
```
pedrobaeza: There's no discount in core in purchase. Use directly `line.price_unit`

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -92,9 +94,182 @@ def account_properties(cr):
...
+    # Replace https://bit.ly/363UnqR
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE purchase_order_line pol
+            SET qty_invoiced = sub.qty
+            FROM (
+                SELECT
+                    pol2.id,
+                    ail.quantity / pol2_u.factor * ail_u.factor * (
```
pedrobaeza: This should have a `SUM` operator

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -92,9 +94,182 @@ def account_properties(cr):
...
+                        ON po2.id = pol1.order_id
+                    -- pol2 cotains lines to invoice
+                    LEFT JOIN purchse_order_line pol2
+                        ON po2.id = pol2.order_id AND
+                        ROUND(pol2.qty_invoiced, %(uom_precision)s) < ROUND(pol2.product_qty, %(uom_precision)s)
+                    -- pol3 contains invoiced lines
+                    LEFT JOIN purchse_order_line pol3
+                        ON po2.id = pol3.order_id AND
+                        ROUND(pol3.qty_invoiced, %(uom_precision)s) >= ROUND(pol3.product_qty, %(uom_precision)s)
+            ) AS sub
```
pedrobaeza, line 281: Subqueries are usually slower. The approach in sale is not valid here?

pedrobaeza (changes requested):

addons/purchase/migrations/9.0.1.2/post-migration.py
yajo: They are slower when in the `WHERE`, but in the `FROM` it shouldn't be a problem.

addons/purchase/migrations/9.0.1.2/end-migration.py
yajo: Yes, I'm just moving the thing to the end, where mrp will be loaded already; I'm not actually optimizing this part.

yajo: Last test migration took 13 min. here.

IMHO it's not worth the effort to optimize this part more.

.travis.yml
```diff
@@ -43,11 +43,12 @@ script:
     - cat openerp/addons/base/migrations/*/tests/*.yml addons/*/migrations/*/tests/*.yml > ../test_data80.yml
     # Roundtrip to previous release to update the test database with additional test data
     # Loading and committing test data without triggering post tests needs https://github.com/odoo/odoo/pull/13146. This is now included in OpenUpgrade.
+    - old_commit="$(git rev-parse HEAD)"
```
pedrobaeza, line 46: @StefanRijnhart @mreficent FYI a possible workaround for avoiding the problem we have with the Travis glitch

mreficent:
conflicts

.travis.yml
MiquelRForgeFlow: good to know

MiquelRForgeFlow: should we implement it in the other versions?

pedrobaeza: We are experimenting in this branch, but if everything goes alright, yes, we will propagate it to all branches.

pedrobaeza (approved):

pedrobaeza:
@Yajo please review my latest optimization and we can proceed to merge.

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -194,30 +196,32 @@ def set_po_line_computed_rest(env):
...
-        env.cr,
-        """
-            UPDATE purchase_order_line pol
-            SET qty_received = pol.product_qty
+        env.cr, """
+        UPDATE purchase_order_line pol
+        SET qty_received = sub.qty_received
+        FROM (
+            SELECT CASE
+                WHEN MIN(po.state) NOT IN ('purchase', 'done') THEN 0.0
```
MiquelRForgeFlow, line 226: what does `MIN(po.state)` mean?

pedrobaeza: It's because we are grouping at other level, so we will have the same `po.state` across several potential records, so we need this for making it work, but no other consequences.

MiquelRForgeFlow: ok

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -92,9 +96,188 @@ def account_properties(cr):
...
+                ail_u.category_id != pol_u.category_id
+        """
+    )
+    uom_category_mismatches = env.cr.dictfetchall()
+    if uom_category_mismatches:
+        raise Exception(
+            "Found these mismatching UoM in related purchase.order.line "
+            "and account.invoice.line records: %s" %
+            pformat(uom_category_mismatches))
+    # Replace https://bit.ly/363UnqR
```
StefanRijnhart: I really find the original links, even if they are too long much more readable. Maybe mark them with noqa or split them up on a slash forward?

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -92,9 +96,188 @@ def account_properties(cr):
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE purchase_order_line pol
+            SET qty_invoiced = sub.qty
+            FROM (
+                SELECT
```
StefanRijnhart, line 170: I know this should already be a huge improvement over the ORM, but in terms of SQL subqueries are relatively expensive. Did you consider using a temporary table instead of the subquery? See https://github.com/VanMoof/OpenUpgrade/commit/2d26cdf118a226855193cfdce94c0131841c4925#diff-d4a6069f9ea8152872ee07aa6bd111fbR480 for an example.

pedrobaeza: The subquery in this case is needed to be populated only once, so I don't think we are going to get a lot of improvement. We will execute the script tomorrow and tell you back what are the obtained times and the number of records.

StefanRijnhart: Sure! This is non blocking on my part.

yajo: Subqueries are expensive when in `SELECT` or `WHERE`, because they are (or could be, depending on postgres optimizations and configuration) executed row by row.

When they're in the `FROM` part, they're not so expensive.

addons/purchase/migrations/9.0.1.2/post-migration.py
```diff
@@ -194,30 +196,32 @@ def set_po_line_computed_rest(env):
...
+        SET qty_received = sub.qty_received
+        FROM (
+            SELECT CASE
+                WHEN MIN(po.state) NOT IN ('purchase', 'done') THEN 0.0
+                WHEN MIN(po.state) IN ('purchase', 'done')
+                    AND MIN(pt.type) NOT IN ('consu', 'product')
+                    THEN MIN(pol2.product_qty)
+                ELSE SUM(CASE
+                    WHEN sm.state IS NULL OR sm.state != 'done' THEN 0.0
+                    ELSE COALESCE(sm.product_uom_qty / u.factor * u2.factor, 0)
```
yajo, line 232: You're not checking that both UoMs belong to the same category, right? That's a behavioral change from: https://github.com/OCA/OpenUpgrade/blob/075cfa226b29819cae3ba1558fd8b3954c31aff0/addons/product/product.py#L122

Migration must fail in that case and provide a useful traceback. You can borrow some ideas from my code a little bit more above.

addons/purchase/migrations/9.0.1.2/end-migration.py
```diff
@@ -1,20 +1,21 @@
...
+    """Complete migration of purchase.order.line#qty_received for products
+    with BoMs. We don't go further in initial filtering, as this number should
+    be low and it's not going to be heavy to be computed.
     """
-    # Execute https://bit.ly/2PgUKaT for matching records
+    if 'mrp.bom' not in env:
+        return  # mrp not installed
     pol_todo = env["purchase.order.line"].search([
-        ("order_id.state", "in", ['purchase', 'done']),
-        ("product_id.type", "in", ['consu', 'product']),
```
yajo: You shouldn't delete these lines; they are there to exclude records that were already computed in post migration, and AFAICS that are gonna hit these lines without need if bom is installed: https://github.com/OCA/OpenUpgrade/blob/195b61a948dff1c64279eb68f054433b34006b51/addons/purchase/purchase.py#L493-L498

addons/purchase/migrations/9.0.1.2/post-migration.py
pedrobaeza: Replaced by full URLs

addons/purchase/migrations/9.0.1.2/post-migration.py
pedrobaeza: Included check

addons/purchase/migrations/9.0.1.2/end-migration.py
pedrobaeza: Included again

yajo (commented):
Seems good overall.

StefanRijnhart (approved):

addons/purchase/migrations/9.0.1.2/post-migration.py
pedrobaeza: On our real DB, this is the result:
```
109182 rows affected after 0:00:07.030084
```
What means 7 seconds for more than 100k records, which legitimates that this subquery usage is sane and no need to bother with temporary tables.

StefanRijnhart: Absolutely! Thanks for the feedback.

.travis.yml
pedrobaeza: Cherry-picked in all branches after successful tests in 9.0 branch.

yajo: Happy to help :blush:

PR #2108, opened 2019-12-08
[IMP] Set default values for multiple columns of the same table

StefanRijnhart:
...in the same query

Setting the default value for a single field takes just as long as setting
the default values for multiple fields at the same time, so this is a win for
large tables with multiple new fields with a default value. Results in queries
like

UPDATE "account_move_line" SET "amount_residual" = 0.0,
    "amount_residual_currency" = 0.0,
    "debit_cash_basis" = 0.0,
    "balance_cash_basis" = 0.0,
    "credit_cash_basis" = 0.0,
    "balance" = 0.0;

### DIFF
--- a/openerp/models.py
+++ b/openerp/models.py
@@ -37,6 +37,7 @@ from operator import itemgetter
 import babel.dates
 import dateutil.relativedelta
 import psycopg2
+from psycopg2 import sql
 from lxml import etree
 
 import openerp
@@ -2428,8 +2429,7 @@ class BaseModel(object):
         # (re-)create the FK
         self._m2o_add_foreign_key_checked(source_field, dest_model, ondelete)
 
-
-    def _set_default_value_on_column(self, cr, column_name, context=None):
+    def _set_default_value_on_column(self, cr, column_name, dont_write=False, context=None):
         # ideally, we should use default_get(), but it fails due to ir.values
         # not being ready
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you share improvement numbers? I'm a bit reluctant to introduce this kind of changes as it's modifying core and can have side effects. This is also benefiting only this version.

StefanRijnhart:
The clause above takes 5 minutes to update 6 fields in one go. Before this change, it would take 6 * 5 minutes (2.2 million records). So in only this particular case it saves me already 25 minutes.

StefanRijnhart:
Of course, if only we had more tests a green flag from Travis would be so much more meaningful...

legalsylvain (approved):
Thanks. Code review / no test.

pedrobaeza:
Have you benchmarked about setting the default in column creation? I have recently added support for this "initialization value" in openupgradelib: https://github.com/OCA/openupgradelib/pull/205

This is for avoiding a patch on OpenUpgrade itself that must be done on each version and control this through migration scripts instead.

Any way, this can be an improvement on Odoo itself for master version.

StefanRijnhart:
Thanks, that's a much better approach
```
migtest=> alter table account_move_line add column test NUMERIC;
migtest=> update account_move_line set test = 0.0;
Time: 346114.279 ms (05:46.114)
migtest=> alter table account_move_line drop column test;
migtest=> alter table account_move_line add column test NUMERIC default 0.0;
ALTER TABLE
Time: 63904.258 ms (01:03.904)
```

PR #2109, opened 2019-12-08
[IMP] use less resources when assigning taxes to move lines

StefanRijnhart:
Apply chunking for lower memory usage, and avoid move validation.
Fixes MemoryError in some configurations


pedrobaeza (approved):
It seems we are both optimizing the migration of this version

rvalyi:
kudos for these optimizations, we are still migrating several Odoo v8 and it's really appreciated!

StefanRijnhart:
> It seems we are both optimizing the migration of this version

Good! I'll be submitting my proposal for account_full_reconcile next.

pedrobaeza:
We have been working on sale part in #2100 and now in purchase in #2104

PR #2118, opened 2019-12-13
[IMP] Do not drop indexes

StefanRijnhart:
Odoo will drop any index that is defined in any other module than the module
that adds the column itself. Such indexes were supposedly added by customization
modules because they were needed in real life scenarios. Typically, removing
such indexes are harmful to the performance of the migration, plus it adds the
cost of re-adding the indexes on the upgrade of the module that defines them.

### DIFF
--- a/openerp/models.py
+++ b/openerp/models.py
@@ -2662,10 +2662,16 @@ class BaseModel(object):
                                         "Use a search view instead if you simply want to make the field searchable."
                                     _schema.warning(msg, self._table, f._type, k)
                             if res2 and not f.select:
-                                cr.execute('DROP INDEX "%s_%s_index"' % (self._table, k))
-                                cr.commit()
-                                msg = "Table '%s': dropping index for column '%s' of type '%s' as it is not required anymore"
-                                _schema.debug(msg, self._table, k, f._type)
+                                # OpenUpgrade: do not drop indexes
+                                # Odoo will drop any index that is defined in any other module than the module
+                                # that adds the column itself. Such indexes were supposedly added by customization
+                                # modules because they were needed in real life scenarios. Typically, removing
+                                # such indexes are harmful to the performance of the migration, plus it adds the
+                                # cost of re-adding the indexes on the upgrade of the module that defines them.
+                                # cr.execute('DROP INDEX "%s_%s_index"' % (self._table, k))
+                                # cr.commit()
+                                msg = "Table '%s': not dropping index for column '%s'"
+                                _schema.debug(msg, self._table, k)
 
                             if isinstance(f, fields.many2one) or (isinstance(f, fields.function) and f._type == 'many2one' and f.store):
                                 dest_model = self.pool[f._obj]


StefanRijnhart:
@pedrobaeza do you remember discussing deleting account_account with type = 'view' or setting them inactive? That provides a very practical example for this. Deletion takes 7 minutes on this particular database with 6 different administrations that I'm working on. I can fix this by creating index for the following FK constraints (that come in handy anyway, probably), but Odoo would simply wipe them before they get used:

```
explain analyze delete from account_account where openupgrade_legacy_9_0_type = 'view';
-- ... --
 Trigger for constraint account_analytic_line_general_account_id_fkey: time=64023.175 calls=766
  Trigger for constraint account_invoice_account_id_fkey: time=200484.992 calls=766
   Trigger for constraint account_invoice_line_account_id_fkey: time=84118.895 calls=766
 Trigger for constraint account_invoice_tax_account_id_fkey: time=24405.092 calls=766
  Trigger for constraint account_voucher_account_id_fkey: time=2785.443 calls=766
 Trigger for constraint account_voucher_line_account_id_fkey: time=9066.677 calls=766
 Trigger for constraint account_voucher_writeoff_acc_id_fkey: time=2446.123 calls=766
  Trigger for constraint account_payment_writeoff_account_id_fkey: time=1883.579 calls=766

 Execution time: 418242.878 ms
```
But otherwise Odoo dropping indexes is insane anyway of course.

pedrobaeza:
I see... so is this also happening in latest version that the index is dropped and re-added when an index is defined in other module than the original field?

The only risk I see with this patch is that we don't drop indexes that disappear from one version to the next one, keeping them and draining performance for no gain in migrated databases. It's not the usual thing, but it can happen.

Any way, being the `select` value evaluated through Python, isn't the proper one in all the inheritance chain when server is running? If not, can we delay the index dropping to a point where it's clear if the index has been removed in all the chain?

pedrobaeza:
See also https://github.com/OCA/openupgradelib/pull/189 and #2116 for an example where indexes are added on the fly for a certain delete specific operation that improves it a lot.

StefanRijnhart:
The value of `select`/`index` is evaluated on a module per module basis, which is why indexes are constantly dropped and re-added. You are right that indexes might remain where they are dropped in the code. I think this is totally harmless and very rare at the same time. I run this change on all my production instances because it is harmful to have indexes being dropped inadvertently, or even being dropped and then rebuild during the same load cycle. On my *really* big instance, some indexes take hours to rebuild.

You are right too that the proper way is to evaluate all indexes when all modules are loaded. I consider such a change out of scope for OpenUpgrade though.

pedrobaeza:
Then what to do? Let current behavior and apply for your use case this patch as hotfix?

I have found also this related Odoo issue: https://github.com/odoo/odoo/issues/24958

pedrobaeza:
Then what to do? Let current behavior and apply for your use case this patch as hotfix?

I have found also this related Odoo issue: https://github.com/odoo/odoo/issues/24958

StefanRijnhart:
No, I think the intermediate dropping and re-adding of indexes is harmful in any case, and I think it would be a good thing if this was merged.
Thanks for the pointer! I'll check it out but it will not be feasible to backport it by the looks of it.

pedrobaeza:
@Yajo what do you think about this?

yajo (approved):
I think this PR makes sense. After all, the recommended operation is to migrate all needed versions with OU and then apply a normal upgrade with normal odoo, which will do its things anyways.

Indexes are quite important in migrations AFAICS, where many batch operations and searches are done.

PR #2115, opened 2019-12-13
[IMP] account: Faster computation of cash basis

yajo:
Translates slow fields to SQL:

- matched_percentage
- debit_cash_basis
- credit_cash_basis
- balance_cash_basis

@Tecnativa TT18838

PR #2119, opened 2019-12-14
[FIX] Regression in matched_percentage calculation causing division by zero

StefanRijnhart:
Main problem is that BETWEEN is inclusive, which allows for a rounding of 1
as is the case for JPY. The LOG of 1 is 0 which causes a division by zero.
Instead, we can just rely on NULL to invalidate numeric comparisons and check
if the result stands up to the numeric comparison 'smaller than 1'.

Regression of https://github.com/OCA/OpenUpgrade/pull/2115

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -990,7 +990,7 @@ def fast_compute_cash_basis(env):
                         WHEN 0.0 = ROUND(
                             SUM(ABS(COALESCE(aml.debit, 0.0) - COALESCE(aml.credit, 0.0))),
                             CASE
-                                WHEN MAX(COALESCE(rc.rounding, 0)) BETWEEN 0 AND 1
+                                WHEN MAX(rc.rounding) < 1
                                 THEN CEIL(LOG(1 / MAX(rc.rounding)))
                                 ELSE 0
                             END::INTEGER


pedrobaeza:
If `rc.rounding` is NULL, then `MAX(rc.rounding) < 1` is not evaluated and goes to the `ELSE` part?

StefanRijnhart:
That's it!
```
update res_partner set company_id = null;
UPDATE 4

select CASE WHEN MAX(company_id) < 2 THEN 'yes' ELSE 'no' END from res_partner;
 case 
------
 no
(1 row)

update res_partner set company_id = 1 where id = 1;
UPDATE 1

select CASE WHEN MAX(company_id) < 2 THEN 'yes' ELSE 'no' END from res_partner;
 case 
------
 yes
(1 row)
```

pedrobaeza:
But this is not what I was talking about:
```
db=# SELECT CASE WHEN MAX(NULL) < 1 THEN 1 ELSE 0 END;
ERROR:  operator does not exist: text < integer
LINE 1: SELECT CASE WHEN MAX(NULL) < 1 THEN 1 ELSE 0 END;
                                   ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```
I mean that you have removed the `COALESCE` and there's a possibility of `rc.rounding` to be NULL.

StefanRijnhart:
See the first line of the SQL snippet, where I set company_id NULL so that MAX is computed over an array of NULL values.

pedrobaeza:
OK, right, my fault.

pedrobaeza (approved):

addons/account/migrations/9.0.1.1/post-migration.py
```diff
@@ -990,7 +990,7 @@ def fast_compute_cash_basis(env):
                         WHEN 0.0 = ROUND(
                             SUM(ABS(COALESCE(aml.debit, 0.0) - COALESCE(aml.credit, 0.0))),
                             CASE
-                                WHEN MAX(COALESCE(rc.rounding, 0)) BETWEEN 0 AND 1
+                                WHEN MAX(rc.rounding) < 1
```
yajo, line 993: If you don't check this is bigger than zero, you could fall in a division by zero error, isn't it?

StefanRijnhart: That would trigger a division by zero indeed. But is `0.0` a valid value for currency rounding? It is not a value that is present in a regular database (and default for new currencies is 0.01)

```
odoo/openerp/addons/base/res$ grep rounding res_currency_data.xml  |sort|uniq
            <field name="rounding">0.0001</field>
            <field name="rounding">0.01</field>
            <field name="rounding">1.00</field>
            <field name="rounding">1</field>
```

PR #2120, opened 2019-12-14
[IMP] Create pack operation lots in a single query

StefanRijnhart:
Replacing ORM creation by SQL brings this stanza back from 16 minutes to 3 seconds to create some 100.000 records.

Also, take care that the correct quantities are set. Earlier version uses `qty_done`, which is only setup a couple of lines later (and even then is only setup for operations of done pickings): https://github.com/OCA/OpenUpgrade/blob/9.0/addons/stock/migrations/9.0.1.1/post-migration.py#L59. All quantities came out zero that way.

### DIFF
--- a/addons/stock/migrations/9.0.1.1/post-migration.py
+++ b/addons/stock/migrations/9.0.1.1/post-migration.py
@@ -35,21 +35,23 @@ def _migrate_pack_operation(env):
     mark pickings that need to recreate pack operations, and update new field
     qty_done on stock.pack.operation for transferred pickings.
     """
-    env.cr.execute(
-        "select o.id, o.%(lot_id)s, p.state, o.qty_done "
-        "from stock_pack_operation o "
-        "join stock_picking p on o.picking_id=p.id "
-        "where o.%(lot_id)s is not null "
-        "group by o.id, o.%(lot_id)s, p.state, o.qty_done",
+    openupgrade.logged_query(
+        env.cr,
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
cc @paulius-sladkevicius

paulius-sladkevicius:
@StefanRijnhart speed, speed, love it! LGTM

addons/stock/migrations/9.0.1.1/post-migration.py
```diff
@@ -35,21 +35,23 @@ def _migrate_pack_operation(env):
...
+        env.cr,
+        """ INSERT INTO stock_pack_operation_lot (
+            id, lot_id, operation_id, qty_todo, qty,
+            create_uid, write_uid, create_date, write_date)
+        SELECT
+            nextval('stock_pack_operation_lot_id_seq'),
+            %(lot_id)s, o.id,
+            CASE WHEN p.state != 'done' THEN o.product_qty ELSE 0 END,
+            CASE WHEN p.state = 'done' THEN o.product_qty ELSE 0 END,
+            1, 1, now() at time zone 'UTC', now() at time zone 'UTC'
```
pedrobaeza: Better to use here the same user as the stock_pack_operation, and even the dates (although this one is more debatable)

StefanRijnhart: Done!

pedrobaeza (approved):

PR #2128, opened 2019-12-16
[12.0][UPD] openupgrade_analysis

MiquelRForgeFlow:
Nothing to do.

--


addons/l10n_es/migrations/12.0.4.0/openupgrade_analysis.txt
```diff
@@ -2,3 +2,4 @@
 ---Fields in module 'l10n_es'---
 ---XML records in module 'l10n_es'---
 DEL account.account.template: l10n_es.account_common_572999
+DEL account.tax.group: l10n_es.tax_group_iva_10-5 (noupdate)
```
MiquelRForgeFlow: This appears because the improvement in v11 has not been forwardported yet to v12.

pedrobaeza: Well, maybe it's worth to wait for it.

MiquelRForgeFlow: Yeah, I will wait for it.

MiquelRForgeFlow: Resolved.

odoo/addons/base/migrations/12.0.1.3/openupgrade_general_log.txt
```diff
@@ -166,7 +166,7 @@ new model website.sale.payment.acquirer.onboarding.wizard [module website_sale]
 # Found in other module: 156
 # Found with different type: 2
 # In obsolete models: 214
-# New columns: 1369
+# New columns: 1370
```
pedrobaeza, line 169: Which is the new column?

odoo/addons/base/migrations/12.0.1.3/openupgrade_analysis.txt
```diff
@@ -121,6 +121,7 @@ NEW res.country.state: base.state_mn_29
 NEW res.country.state: base.state_mn_32
 NEW res.country.state: base.state_mn_34
 NEW res.country.state: base.state_mn_35
+NEW res.currency: base.MRU
```
pedrobaeza, line 124: Why it appears as new? Then why have we put it in noupdate_records.xml

MiquelRForgeFlow: What we have put in `noupdate_records.xml` is the change in a `res.country` record.

odoo/addons/base/migrations/12.0.1.3/openupgrade_general_log.txt
MiquelRForgeFlow: I didn't update this. I haven't done a new analysis. Just updated manually the previous. I think it's not important this number. Later, in a future analysis, this will be changed automatically.

PR #2132, opened 2019-12-17
[FIX][10.0] Prevent duplicate move lines when correcting old reconciliations

thomaspaulb:

Odoo 10.0 [creates account.payment records](https://github.com/OCA/OCB/blob/10.0/addons/account/models/account_bank_statement.py#L911) on reconcilation from bank statement since [this commit](https://github.com/OCA/OCB/commit/3d48f2a375f22fc3da0f013041cfe0ce9ba6fb13). Odoo 9.0 and before [did not do this](https://github.com/OCA/OCB/blob/9.0/addons/account/models/account_bank_statement.py#L826).

On cancelling a reconciliation on a bank statement, Odoo 10.0 uses the [payment_id stored on account.move.line](https://github.com/OCA/OCB/blob/10.0/addons/account/models/account_bank_statement.py#L426) to find the move lines to delete. However for the bank statement lines created in 9.0 or earlier, there is no linked payment_id and as such those move lines are not found and not deleted. When the reconcilation is done again, duplicate move lines result and the ledger goes faulty.

With this PR, account.payment records are created for each reconciled bank statement line, so that corrections can be made.

Current behavior before PR:

Cancelling a pre-migration bank statement line reconciliation and making it again results in duplicate move lines.

Cancelling a pre-migration bank statement line reconciliation and making it again results in old move lines being deleted and replaced by new ones.

--

### DIFF
--- a/addons/account/migrations/10.0.1.1/post-migration.py
+++ b/addons/account/migrations/10.0.1.1/post-migration.py
@@ -2,6 +2,7 @@
 # Copyright 2017 Therp BV
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _
 from openupgradelib import openupgrade
 
 
@@ -120,7 +121,7 @@ def migrate(env, version):
     # gather non-zero bank statement lines that have journal entries,
     # but no linked payments, meaning they are reconciled, but no
     # payment was created.
-    cr.execute('''
+    openupgrade.logged_query(cr, '''
         select l.id
         from account_bank_statement_line l
         join account_move m
@@ -161,8 +162,8 @@ def migrate(env, version):
             state='reconciled',
             amount=abs(total),
             communication=communication,
-            name=st_line.statement_id.name or
-                _('Bank Statement %s') % st_line.date
+            name=st_line.statement_id.name or _(
+                'Bank Statement %s') % st_line.date
         ))
 
         # link this new payment to the statement line
@@ -171,4 +172,3 @@ def migrate(env, version):
             ('name', '=', st_line.move_name),
         ])
         move.line_ids.write(dict(payment_id=payment.id))
-
--- a/addons/account/migrations/10.0.1.1/post-migration.py
+++ b/addons/account/migrations/10.0.1.1/post-migration.py
@@ -116,3 +116,59 @@ def migrate(env, version):
     openupgrade.load_data(
         cr, 'account', 'migrations/10.0.1.1/noupdate_changes.xml',
     )
+
+    # gather non-zero bank statement lines that have journal entries,
+    # but no linked payments, meaning they are reconciled, but no
+    # payment was created.
+    cr.execute('''
+        select l.id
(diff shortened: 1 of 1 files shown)


thomaspaulb:
In case you are post-migration and duplicate move lines were already generated, this is the query to remove them. Dropping it here because I see no other obvious place to do so.

```
delete from account_move where id in (
select m.id
from account_move_line l, account_move m, account_account a,
    account_journal j, account_bank_statement_line sl
where sl.openupgrade_legacy_9_0_journal_entry_id = m.id
    and j.id = l.journal_id and l.move_id = m.id 
    and l.account_id = a.id and m.statement_line_id is null
    and m.statement_line_id is distinct from sl.id
)
```

addons/account/migrations/10.0.1.1/post-migration.py
```diff
@@ -116,3 +117,58 @@ def migrate(env, version):
...
+    ''')
+    statement_lines = env['account.bank.statement.line'].browse([
+        row[0] for row in cr.fetchall()])
+
+    # create payments for these
+    for st_line in statement_lines:
+        total = st_line.amount
+        journal = st_line.journal_id
+        currency = journal.currency_id or st_line.company_id.currency_id
+        payment_methods = (total > 0) and \
```
StefanRijnhart: As a rule of thumb, use `if ... else` in cases like these (`journal.inbound_payment_method_ids if total > 0 else outbound_payment_method_ids`) to prevent `journal.outbound_payment_method_ids` kicking in when `journal.inbound_payment_method_ids` is falsy even when total < 0.

addons/account/migrations/10.0.1.1/post-migration.py
```diff
@@ -116,3 +117,58 @@ def migrate(env, version):
...
+        having (
+            sum(case when ap.id is not null then 1 else 0 end) = 0
+            and sum(case when ml.id is not null then 1 else 0 end) > 0
+        )
+    ''')
+    statement_lines = env['account.bank.statement.line'].browse([
+        row[0] for row in cr.fetchall()])
+
+    # create payments for these
+    for st_line in statement_lines:
```
StefanRijnhart: Can you apply https://doc.therp.nl/openupgrade/API.html#openupgradelib.openupgrade.chunked to the statement_lines recordset? This improves performance and prevents memory errors on large operations.

StefanRijnhart:
Thanks for the fix and the detailed description!

addons/account/migrations/10.0.1.1/post-migration.py
thomaspaulb: Nicely spotted, thanks.

addons/account/migrations/10.0.1.1/post-migration.py
thomaspaulb: I have dumbly added another nesting of `for chunk in statement_lines: for st_line in chunk:`. Is that enough or have I overlooked something?

StefanRijnhart: That's it, thanks!

StefanRijnhart (approved):

PR #2133, opened 2019-12-18
[FIX] utm: Rename `crm.tracking.*` to `utm.*`

yajo:
- Done in `base` because `utm` is a new module and migration must be done before it is installed.
- Removed some migration from `crm` which is buggy after the new paradigm.

@Tecnativa TT18838


MiquelRForgeFlow (approved):

pedrobaeza:
Remove a leftover, so let's merge it when ready:

/ocabot merge

PR #2134, opened 2019-12-18
[12.0] update upstream e7984c4968c

MiquelRForgeFlow:
Update to include fix in `l10n_es`.

--

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
+++ b/odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
@@ -171,4 +171,7 @@ System</span>]]></field>
   <record id="dm" model="res.country">
     <field eval="1767" name="phone_code"/>
   </record>
+  <record id="mr" model="res.country">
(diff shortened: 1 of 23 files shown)


odoo/addons/base/migrations/12.0.1.3/noupdate_changes.xml
```diff
@@ -171,4 +171,7 @@ System</span>]]></field>
   <record id="dm" model="res.country">
     <field eval="1767" name="phone_code"/>
   </record>
+  <record id="mr" model="res.country">
+    <field name="currency_id" ref="MRU"/>
```
pedrobaeza, line 175: Isn't there any constraint that prevents this?

MiquelRForgeFlow: I don't think so :S

pedrobaeza: Yeah, and Travis would complain about it, so good to go.

PR #2136, opened 2019-12-19
[FIX] base: Drop FKs for procurement.order

yajo:
Running [this v12 migration code][1] in a DB that was migrated from v10 and that had many procurement orders takes days. That's because the FK was set without an index.

At the end of the day, this isn't useful because that table is no longer used, so I'm dropping its FKs here.

WIP until merged:

- [x] https://github.com/OCA/openupgradelib/pull/195

[1]: https://github.com/OCA/OpenUpgrade/blob/bdf26ceb11834fd45ca4b4577532d6f5ee2ca4e4/addons/sale/migrations/12.0.1.1/pre-migration.py#L93-L104

@Tecnativa TT18838

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
@@ -24,6 +24,10 @@ model_renames_ir_actions_report = [
     ('ir.actions.report.xml', 'ir.actions.report')
 ]
 
+_obsolete_tables = (
+    "procurement_order",
+)
+
 
 def handle_partner_sector(env):
     if openupgrade.table_exists(env.cr, 'res_partner_sector'):
@@ -63,6 +67,7 @@ def fill_cron_action_server_pre(env):
 
 @openupgrade.migrate()
 def migrate(env, version):
+    openupgrade.remove_tables_fks(env.cr, _obsolete_tables)
     openupgrade.update_module_names(
         env.cr, apriori.renamed_modules.items()
     )
--- a/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
@@ -25,7 +25,36 @@ model_renames_ir_actions_report = [
 ]
 
 _obsolete_tables = (
+    "base_action_rule",
+    "base_action_rule_lead_test",
+    "base_action_rule_line_test",
+    "crm_activity",
+    "hr_timesheet_sheet_sheet",
+    "ir_values",
+    "marketing_campaign",
+    "marketing_campaign_activity",
+    "marketing_campaign_segment",
+    "marketing_campaign_transition",
+    "marketing_campaign_workitem",
     "procurement_order",
+    "project_issue",
+    "report",
+    "res_font",
+    "stock_move_lots",
+    "stock_move_operation_link",
+    "stock_pack_operation",
+    "stock_pack_operation_lot",
+    "stock_picking_wave",
+    "subscription_document",
+    "subscription_document_fields",
+    "subscription_subscription",
+    "subscription_subscription_history",
+    "wkf",
+    "wkf_activity",
+    "wkf_instance",
+    "wkf_transition",
+    "wkf_triggers",
+    "wkf_workitem",
 )
 
 
--- a/addons/stock/migrations/11.0.1.1/pre-migration.py
+++ b/addons/stock/migrations/11.0.1.1/pre-migration.py
@@ -37,17 +37,6 @@ def delete_quants_for_consumable(env):
     )
 
 
-def drop_slow_constraint(env):
-    """Removing this constraint, that doesn't affect new data structure, as
-    it belongs to an obsolete model, we get tons of more performance on
-    quant removal.
-    """
-    openupgrade.logged_query(
-        env.cr, "ALTER TABLE stock_move_operation_link DROP CONSTRAINT "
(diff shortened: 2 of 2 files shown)


pedrobaeza (commented):
I wouldn't complicate the code adding a method, a list. Simply add this line in `def migrate`:
```python
openupgrade.remove_table_fks(env.cr, 'procurement_order')
```

yajo:
I did it like this because I see many obsolete models:
https://github.com/OCA/OpenUpgrade/blob/a6dbb655b5a6cfb750bf14993029e1bcb6df392a/odoo/addons/base/migrations/11.0.1.3/openupgrade_general_log.txt#L41-L71

I suspect that many of them will have to do this sooner or later, so the method is to have a placehloder to start including those things.

pedrobaeza:
OK for the list, but don't create a method and call the openupgradelib call in the main migrate method. Should openupgradelib method admit a list?

yajo:
> Should openupgradelib method admit a list?

It could be a solution.

pedrobaeza:
I'm thinking that this should be removed in the scripts of the module that previously contained the table. If not, we will have an error.

yajo:
In case the table is absent, for this case it would be the same as if the table wouldn't had any FKs. It shouldn't error AFAICS in the openupgradelib method implementation.

yajo:
Oh and AFAIK the `procurement` module no longer exists in this version...

pedrobaeza:
What about directly include here all the obsolete models?

yajo:
I guess it makes sense, although I just wanted to keep a smaller patch. Some of the obsolete models are just renamed I guess, and in those cases this shouldn't be needed.

pedrobaeza:
Good point, but I think we should check what are renamed and add the rest here instead of doing it in 2 steps. Also with the merge of #2148 we minimize the need of this one.

yajo:
Done now.

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -24,6 +24,39 @@
     ('ir.actions.report.xml', 'ir.actions.report')
 ]
 
+_obsolete_tables = (
+    "base_action_rule",
+    "base_action_rule_lead_test",
+    "base_action_rule_line_test",
+    "crm_activity",
+    "hr_timesheet_sheet_sheet",
```
MiquelRForgeFlow, line 32:
```suggestion
```
This table is renamed to 'hr_timesheet_sheet' in `hr_timesheet_sheet` module. I think tables like this one shouldn't be in this list.

MiquelRForgeFlow:
Also delete `drop_slow_constraint(env)` method in `stock` pre-migration.

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
MiquelRForgeFlow: Basically, I think you should delete from this list:

base_action_rule
base_action_rule_lead_test
base_action_rule_line_test
hr_timesheet_sheet_sheet
stock_picking_wave

Not sure if I left something.

yajo: What do you think @pedrobaeza ?

pedrobaeza: @Yajo if other module defines the same m2o/m2m field, the FK is re-added by ORM. If so, then I think full list is correct and if you install for example `hr_timesheet_sheet` module, everything will be set as it was. If not, you don't have a table with FKs consuming resources when doing operations in linked tables.

pedrobaeza:
Any way, this is red for now.

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
yajo: Yes, the only problem would be pre-migration scripts, if any, because they could rely on a missing FK.

I think I'm just gonna remove the last commit, as this PR is drifting out of scope.

pedrobaeza: I don't think that's the case, and I prefer to remove this by default and if any pre-migration should rely on that, then re-add it (which I don't think there's any case). Removing this, we are losing potentially performance for a non clear need of restoration.

yajo: If this were the 1st commit in this migration, yes... But to be sure we're not breaking anything here, I should:

1. Check which modules defined these tables in the previous version.
1. Check if those are properly migrated in this version.
1. Check no pre-migration script relies on an FK for these tables.
1. Remove the tables from the list or fix the affected premigration scripts.

So, when this PR started as a simple mechanism to improve performance of the procurement migration, it would become a complex monster that would potentially affect the whole migration process.

I agree it would be good to drop all those FKs, but at the point where we're at, and with the implied risk of introducing new bugs, let me insist that we stay pragmatic and just fix the issue at hand, please.

pedrobaeza: I save you all the time: you can do it without risk. I can assure you, so let's go this way.

pedrobaeza:
Just fix Travis and it's good to go.

yajo:
Let's see if it goes :heavy_check_mark: now.

MiquelRForgeFlow:
@Yajo it's green. Would you mind to do the same thing for v12?

pedrobaeza (approved):

Issue #2144, opened 2019-12-23
[12.0] Error A user cannot have both Tax B2B and Tax B2C, migrating account module. 


GSLabIt:
Impacted versions: 12.0

Steps to reproduce:
1. in a v 11.0 db have Website Tax as B2C, so that Portal and/or Public Group will have B2C Tax.
2. start migrate to v 12.0 using `-u all`
3. error on account migration due to: 
https://github.com/OCA/OpenUpgrade/blob/bdf26ceb11834fd45ca4b4577532d6f5ee2ca4e4/addons/account/models/res_users.py#L24

Current behavior:
exception is raised because user has both B2B and B2C tax at same time. This seems to be caused by:
https://github.com/OCA/OpenUpgrade/blob/bdf26ceb11834fd45ca4b4577532d6f5ee2ca4e4/addons/account/migrations/12.0.1.1/noupdate_changes.xml#L108

as it force `account.group_show_line_subtotals_tax_excluded` to all users in groups Portal and Public.

Expected behavior: 
Do not force `account.group_show_line_subtotals_tax_excluded` in groups Portal and Public during migration.

Video/Screenshot link (optional):

StefanRijnhart:
I see the problem. However, can you clarify "in a v 11.0 db have Website Tax as B2C" and how that leads to account.group_show_line_subtotals_tax_included being an implied group of Portal and Public given that this group is also introduced in Odoo 12?

GSLabIt:
@StefanRijnhart will try to better explain the issue. 

On `website` settings, when u have `website_sale` installed, you can select if product will be shown as Tax Excluded or Tax Included:

<img width="1279" alt="Schermata 2020-01-03 alle 15 14 26" src="https://user-images.githubusercontent.com/11652845/71727815-bef69280-2e3b-11ea-9be9-1930f885b24e.png">

Setting option to `Tax Included` will change implied group for `Portal` and `Public` groups to `Tax Display B2C`, see following pics:

<img width="1270" alt="Schermata 2020-01-03 alle 15 16 45" src="https://user-images.githubusercontent.com/11652845/71727952-1d237580-2e3c-11ea-9bb1-39bb1f386559.png">
<img width="1272" alt="Schermata 2020-01-03 alle 15 16 25" src="https://user-images.githubusercontent.com/11652845/71727953-1d237580-2e3c-11ea-90c8-f4a584443a2d.png">

[...]

PR #2146, opened 2019-12-23
[12.0][FIX] account migration, do not force account.group_show_line_subtotals_tax_excluded

GSLabIt:
See https://github.com/OCA/OpenUpgrade/issues/2144

Current behavior before PR:

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/noupdate_changes.xml
+++ b/addons/account/migrations/12.0.1.1/noupdate_changes.xml
@@ -104,12 +104,6 @@
   <record id="data_account_type_direct_costs" model="account.account.type">
     <field name="internal_group">expense</field>
   </record>
-  <record id="base.group_portal" model="res.groups">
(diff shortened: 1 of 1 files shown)


GSLabIt:
@pedrobaeza could u double check and review this one? or tag appropriate reviewer? big thx

StefanRijnhart:
Showing the total excluding tax is the default for new installations of Odoo 12.0 (https://github.com/odoo/odoo/blob/12.0/addons/account/security/account_security.xml#L77-L85) and it was the standard behaviour of older version if I am not mistaken. So I'm not convinced of the fix. Then again, I'm not entirely clear on the problem, see my comment on the ticket.

StefanRijnhart:
Agreed with the fix now, after reading your clarification on the ticket. The setting is dealt with by renaming the existing groups, and the noupdate data that is removed int his PR is redundant and causing inconsistencies indeed. Thanks for the fix!

StefanRijnhart (approved):

PR #2147, opened 2019-12-28
[FIX] missing report type and tree type

sergiocorato:


pedrobaeza:
I think we should directly remove that checks, as they block the migration and can have valid values depending on the extra modules in the DB, so adding more and more values is not the ideal situation.

@legalsylvain you added it in 91f7d18d4d993f8ab143305fc00eeaffd7d0c923. Can you explain the reasoning behind this?

legalsylvain:
As far as I remember (5 years ago, certainly for a migration from 5 to 6 or 7...) i had a database with obsoletes values, and need a warning to know it.
I think that we should improve the code instead of removing it, introducing dynamic list but not a blocking point.
I'll check the Pr to know the discussion arpund this commit.

legalsylvain (approved):

pedrobaeza:
Odoo already checks this, indicating that a selection value is not valid, so I still prefer to remove this.

StefanRijnhart:
It just logs an error doesn't it? Is there much harm in that? Could be informative for developers. Maybe change to WARN?

StefanRijnhart (approved):

sergiocorato:
Yes, it is only a log

StefanRijnhart:
@pedrobaeza please reconsider. It's only a harmless log message, not a block.

pedrobaeza:
Well, I don't like but I'm not blocking it either, as I haven't pushed "Changes requested". If you still prefer to keep it, go ahead, as this is an old version.

StefanRijnhart:
@pedrobaeza ok, thanks for confirming.

/ocabot merge

PR #2148, opened 2019-12-28
[12.0][PERF] Improve perf for processing Layout on sale and invoice

sebastienbeau:
Adding and removing layout section on sale_order_line and invoice_line is very very long

Current behavior before PR:

The current migration process add layout on every sale and invoice and then remove all not need layout section. The sale and invoice that do not need them are the sale/invoice that do not have line with layout_category

Layout section are only insert for sale and invoice that need it.
In my case as I migrate from v8 so I do not have any layout on any sale invoice and this change same me more then 18 hours

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/pre-migration.py
+++ b/addons/account/migrations/12.0.1.1/pre-migration.py
@@ -99,21 +99,13 @@ def fill_account_invoice_line_sections(cr):
             min(ail.write_uid), min(ail.write_date)
         FROM account_invoice_line ail
         LEFT JOIN sale_layout_category slc ON slc.id = ail.layout_category_id
+        WHERE ail.invoice_id in (
+            SELECT invoice_id
+            FROM account_invoice_line
+            WHERE layout_category_id IS NOT NULL)
         GROUP BY invoice_id, layout_category_id
         ORDER BY invoice_id, layout_category_id, sequence
         """
-    )
-    # We remove recently created account.invoice.line for sections on invoices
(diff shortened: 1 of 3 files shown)


pedrobaeza:
The reason for the slowness is due to a FK constraint to the old table `procurement_order`. We are handling that in #2136.

The addition + removal of the layout lines was added in #2058 with the rationale written in that PR.

Any way, if we have an alternative that doesn't need that, it's welcome. Can you share the new query times in comparison with the old ones (adding all the lines)? Not having a test DB with a deep usage of section lines makes the test not very significant, but at least to know if that subquery is also very time consuming.

pedrobaeza:
Ah, and please add the same subquery in `sale_mangement` module.

sebastienbeau:
Hi Pedro

Adding all line takes around 1 minute and 30 seconds, but the deletion was really really long (more then 18 hours I stop before). After my PR the insert was done in less then 2s ! (more exactly : 01.31s).

I have added the change on sale management.

Thanks for the link to the procurement order issue, I understand more why the delete was so long

pedrobaeza:
There's an error in the latest SQL:
```   
ERROR: column "order_id" does not exist
LINE 12:             SELECT order_id
```

yajo (approved):

sebastienbeau:
Thanks for the merge and for your time for the review !

pedrobaeza:
Thanks for fixing this :wink: 

Happy New Year!

PR #2149, opened 2019-12-30
[12.0][FIX]fill purchase_order_user_id before it is autofilled

AaronHForgeFlow:
Purchase representative is not filled by the migration scripts because Odoo fills it automatically when upgrading the module

Current behavior before PR: 
Purchase representative in purchase order is always administrator. 

Purchase representative is the user who created the purchase order

--

### DIFF
--- a/addons/purchase/migrations/12.0.1.2/pre-migration.py
+++ b/addons/purchase/migrations/12.0.1.2/pre-migration.py
@@ -54,7 +54,23 @@ def precompute_pol_product_uom_qty(env):
     )
 
 
+def fill_purchase_order_user_id(cr):
+    if not openupgrade.column_exists(cr, 'purchase_order', 'user_id'):
+        # Done here because user_id is filled automatically when upgrading
+        cr.execute(
+            """
+            ALTER TABLE purchase_order ADD COLUMN user_id integer;
+            """)
+    openupgrade.logged_query(
+        cr, """
+        UPDATE purchase_order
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
cc @MiquelRForgeFlow

pedrobaeza:
Check Travis status

addons/purchase/migrations/12.0.1.2/pre-migration.py
```diff
@@ -53,8 +53,23 @@ def precompute_pol_product_uom_qty(env):
             AND pol.product_uom != pt.uom_id""",
     )
```
MiquelRForgeFlow, line 55: add blank line to solve flake8

AaronHForgeFlow: I cannot believe the problem was that minor thing. :sob:

MiquelRForgeFlow (approved):
LGTM, but I think you could update also the `openupgrade_analysis_work.txt` to mark that the script is in pre-migration instead of post-migration.

AaronHForgeFlow:
> LGTM, but I think you could update also the `openupgrade_analysis_work.txt` to mark that the script is in pre-migration instead of post-migration.

Sure. I just did.

AaronHForgeFlow:
cc @pedrobaeza

pedrobaeza (approved):

PR #2155, opened 2020-01-07
[FIX] l10n_es: migrate account.type records

yajo:
In both Odoo core and OCA/l10n-spain implementations of `l10n_es` module, there were some `account.type` records declared that are now absent in v11.

If somebody used `account.chart.update` to apply any of the `account.chart.template`s that this module supplies, existing `account.account` records would be linked to these records that will be now deleted.

To avoid that problem, I upload a CSV file which contains the mapping between the removed `account.type` records, the account code, and the new equivalent type in v11. Then, I use that mapping file to update `account.account` records in the pre-migration script.

@Tecnativa TT18838


yajo:
Added a fallback mode to accounts without an exact equivalent in the template.

yajo:
:heavy_check_mark: finally. It should be ready to merge.

pedrobaeza:
Change in https://github.com/OCA/OpenUpgrade/blob/11.0/odoo/openupgrade/doc/source/modules100-110.rst `Nothing to do` by `Done`

pedrobaeza (approved):

PR #2156, opened 2020-01-08
[ADD] base_status: merge with base

yajo:
Module `base_status` in v7 declared just a couple of mixins and translations.

It is garbage, basically. Merging into base to remove it automatically and stop raising warnings after migrations, like:

    WARNING prod odoo.modules.graph: module base_status: not installable, skipped

@Tecnativa TT18838

### DIFF
--- a/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
+++ b/openerp/addons/base/migrations/8.0.1.3/pre-migration.py
@@ -37,6 +37,7 @@ def cleanup_modules(cr):
     openupgrade.update_module_names(
         cr, [
             ('account_report_company', 'account'),
+            ('base_status', 'base'),
             # from OCA/product-attribute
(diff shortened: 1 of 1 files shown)


pedrobaeza:
OK, I agree,

/ocabot merge

openerp/addons/openupgrade_records/lib/apriori.py
```diff
@@ -4,6 +4,7 @@
 
 renamed_modules = {
     'base_calendar': 'calendar',
+    'base_status': 'base',
```
pedrobaeza: This is not renamed, but merged!!

pedrobaeza (changes requested):

openerp/addons/openupgrade_records/lib/apriori.py
yajo: Oops... fixed.

PR #2157, opened 2020-01-10
[12.0][IMP] account: Avoid warning on transfer_account_code_prefix column

pedrobaeza:
We pre-create it and fill it with default values for that.

@Tecnativa TT21209

### DIFF
--- a/addons/account/migrations/12.0.1.1/pre-migration.py
+++ b/addons/account/migrations/12.0.1.1/pre-migration.py
@@ -117,6 +117,17 @@ def fill_account_invoice_line_sections(cr):
     )
 
 
+def prefill_account_chart_template_transfer_account_prefix(env):
+    """We do this for avoiding a temporary warning on null values."""
+    openupgrade.add_fields(
+        env, [('transfer_account_code_prefix', 'account.chart.template',
+               'account_chart_template', 'char', False, 'account')],
+    )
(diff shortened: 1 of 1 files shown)


addons/account/migrations/12.0.1.1/pre-migration.py
```diff
@@ -117,6 +117,17 @@ def fill_account_invoice_line_sections(cr):
...
 
+def prefill_account_chart_template_transfer_account_prefix(env):
+    """We do this for avoiding a temporary warning on null values."""
+    openupgrade.add_fields(
+        env, [('transfer_account_code_prefix', 'account.chart.template',
+               'account_chart_template', 'char', False, 'account')],
+    )
+    openupgrade.logged_query(
+        env.cr, "UPDATE account_chart_template "
+                "SET transfer_account_code_prefix = 'OUB'")
```
MiquelRForgeFlow, line 128:
```suggestion
                "SET transfer_account_code_prefix = '-OU'")
```
maybe?

MiquelRForgeFlow (approved):
LGTM

addons/account/migrations/12.0.1.1/pre-migration.py
pedrobaeza: This is the same than the rest of prefixes in post.

PR #2158, opened 2020-01-12
[ADD] rename partner_street_number to base_address_extended

hbrunn:
I think we can see [`base_address_extended`](https://github.com/OCA/OCB/blob/11.0/addons/base_address_extended/models/base_address_extended.py) as the successor of `partner_street_number` even though the layout looks somewhat different and the implementation is different. The column names are the same though, so this doesn't need an extra migration.


pedrobaeza (approved):
OK for me. Waiting a second opinion.

cubells (approved):
Ok for me too.

PR #2160, opened 2020-01-16
[9.0] stock: Fine tuning tracking by serial assignation

pedrobaeza:
This PR contains 2 commits for the same part:

1. If you have this stock distribution of one lot:

   - Location 1: quant with 2.0
   - Location 2: quant with -1.0
    
   previous query considers this as a serial number, as the sum is 1.0, which is not correct. This PR adds an extra condition for checking if the number of quants is 1, reducing this possibility.

2. Previous query assumes that product template IDs are the same than the product variants IDs, which is not correct when you have variants.
    
   With this new query, that's taken into account, although there's the risk that one variant has lots with one quantity and others not, and this query assigns serial in that case.

@Tecnativa TT18601


StefanRijnhart (approved):
Thanks! Can we merge this or do you want to wait for the requested reviews?

pedrobaeza:
Not needed if you have already validated, so merging.

PR #2166, opened 2020-01-21
[FIX] sale: Avoid costly qty_delivered computation

pedrobaeza:
In v11, `qty_delivered` was a regular field that was filled by business logic. Now in v12, it's a computed (stored) field, which means that it's automatically computed according some conditions.

As there's a chance that the stored value differs from the v12 computation result (for example, if you have extension modules that modifies the logic), we provided a trick for computing on the fly the value for v12, and if it differs from the stored one, we switch the delivered method to manual and set the previous value.

This has proven to be extremely slow with large quantity of sales order lines (which is logic, as it requires to create a virtual record per record, and compute by ORM the result).

That's why this approach is disabled in this commit, given the facts:

* It's unlikely that the stored quantity differs from the v12 computed quantity. If you had v11 modules that alters this result, you should replicate them on v12.
* Even on that case, the stored delivered quantity is preserved while no change in dependant fields is triggered, so for most of the records, the value will be untouched.

cc @Tecnativa TT19503


yajo (approved):

MiquelRForgeFlow (commented):
I am not opposed. I understand sometimes we should prioritize optimization from accuracy.

What I am not sure is if your use case it's something very specific or something to be applied in general. Until now, this query had not presented slowness problems. I mean, usually there are not many databases with that many sale order lines, right?

JordiBForgeFlow:
@MiquelRForgeFlow I agree with @pedrobaeza. We are experiencing serious bottlenecks due to this. So :+1: on my side.

MiquelRForgeFlow (approved):

pedrobaeza:
That totally depends on the company's sector, but it's very normal to have sales order lines. A lot or few depends on this, but we should scale properly, so let's try this and see if we find inaccuracies (I don't think so).

Merging then.

PR #2167, opened 2020-01-21
[10.0] res.country.state migration now works on manually-created records that don't have XMLID

alexis-via:
--

### DIFF
--- a/odoo/addons/base/migrations/10.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/10.0.1.3/pre-migration.py
@@ -85,33 +85,54 @@ def ensure_country_state_id_on_existing_records(cr):
                 }
             )
             found_id = cr.fetchone()
-            if not found_id:
-                continue
-            # fourth: if found, ensure it has the same xmlid as the csv record
-            openupgrade.logged_query(
-                cr,
-                """
-                UPDATE ir_model_data
-                SET name = '%(data_name)s', module = 'base'
-                WHERE id = %(data_id)s AND model = 'res.country.state'
-                """ % {
-                    'data_name': data_name,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why do you want to do this?

alexis-via:
Well... I want to do this to make the migration work... it's as simple as that :) Without this, it just doesn't work when the res.country.states have been created manually (so they don't have an XMLID).

pedrobaeza:
I still don't have the problem. What is the error?

StefanRijnhart:
Reading through the code, it looks like the original code only worked if state records already had an XML id (either from a custom or a localization module, or an __export__ id). This change then adds an XMLID if necessary (and not just update an existing XMLID). Looks good to me!

StefanRijnhart (approved):

pedrobaeza:
Yeah, but my question is in which part later the XML-ID is needed? This is only for avoid duplications when loading later core states AFAIK, so there's no added value on adding XML-IDs for non existing states.

alexis-via:
Yes, without this PR, Odoo will generate new res.country.state upon upgrade, so the upgrade will fail on the sql constraint unique(country_id, code), cf https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_country.py#L101

pedrobaeza:
Where in the code Odoo generates that states?

StefanRijnhart:
Here: https://github.com/OCA/OpenUpgrade/blob/10.0/odoo/addons/base/migrations/10.0.1.3/openupgrade_analysis_work.txt#L73-L658

pedrobaeza (approved):
OK, now I see the workflow. Thanks for the improvement.

Issue #1752, opened 2019-02-18
[9.0] hr_recruitment migration raises IntegrityError


metaminux:
Impacted versions: Migrate from 8.0 to 9.0

Steps to reproduce:
Migrate a database in v8 with modules 'hr_recruitment' installed and hr_recruitment_source records exist with id > 5 ("utm" addon is defining 5 sources)

Current behavior:
```
2019-02-15 20:19:09,985 628 INFO database_8_migrated openerp.sql_db: bad query: ALTER TABLE "hr_applicant" ADD FOREIGN KEY ("source_id") REFERENCES "utm_source" ON DELETE set null
2019-02-15 20:19:09,986 628 ERROR database_8_migrated openerp.modules.registry: Failed to load registry
Traceback (most recent call last):
  File "/var/tmp/openupgrade/9.0/server/openerp/modules/registry.py", line 390, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/var/tmp/openupgrade/9.0/server/openerp/modules/loading.py", line 446, in load_modules
    force, status, report, loaded_modules, update_module, models_to_check, upg_registry)
  File "/var/tmp/openupgrade/9.0/server/openerp/modules/loading.py", line 331, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check, upg_registry=upg_registry
  File "/var/tmp/openupgrade/9.0/server/openerp/modules/loading.py", line 178, in load_module_graph
    init_module_models(cr, package.name, models)
  File "/var/tmp/openupgrade/9.0/server/openerp/modules/module.py", line 307, in init_module_models
    obj._auto_end(cr, {'module': module_name})
  File "/var/tmp/openupgrade/9.0/server/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/var/tmp/openupgrade/9.0/server/openerp/models.py", line 2762, in _auto_end
    cr.execute('ALTER TABLE "%s" ADD FOREIGN KEY ("%s") REFERENCES "%s" ON DELETE %s' % (t, k, r, d))
  File "/var/tmp/openupgrade/9.0/server/openerp/sql_db.py", line 154, in wrapper
    return f(self, *args, **kwargs)
  File "/var/tmp/openupgrade/9.0/server/openerp/sql_db.py", line 233, in execute
    res = self._obj.execute(query, params)
[...]

PR #2168, opened 2020-01-21
[9.0][FIX] bug #1752 proper migration of source_id on hr.applicant

alexis-via:
More details on bug report #1752 

--

### DIFF
--- a/addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
+++ b/addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
@@ -2,7 +2,7 @@
 # © 2016 Tecnativa <vicent.cubells@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
-from openerp import fields
+from openerp import api, fields, SUPERUSER_ID
 from datetime import timedelta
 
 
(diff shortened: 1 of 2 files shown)


addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
```diff
@@ -27,22 +27,18 @@ def update_applicant_availability(cr):
...
-                   "SELECT name "
-                   "FROM %s "
-                   "RETURNING id",
-                   (openupgrade.get_legacy_name('hr_recruitment_source'),))
-        new_id = cr.fetchone()[0]
-        cr.execute("UPDATE hr_applicant "
-                   "SET source_id = %s "
-                   "WHERE source_id = %s",
-                   (new_id, old_id[0]))
+    env = api.Environment(cr, SUPERUSER_ID, {})
```
pedrobaeza, line 30: Use this as faster operation and usual way to do things here in OpenUpgrade:
```python
column = openupgrade.get_legacy_name('hr_recruitment_source')
openupgrade.logged_query(cr, "ALTER TABLE utm_source ADD %s INT4", (column, ))
openupgrade.logged_query(
    cr, """
    INSERT INTO utm_source
    (create_uid, create_date, write_uid, write_date, name, %s)
    SELECT create_uid, create_date, write_uid, write_date, name, id
    FROM hr_recruitment_source
    """, (column, ),
)
openupgrade.logged_query(
    cr, """
    UPDATE hr_applicant ha SET source_id = us.id
    FROM utm_source us WHERE us.%s = ha.%s
    """, (column, openupgrade.get_legacy_name('source_id')),
)
```
Maybe for avoiding flake8 you need to put `AsIs` in the names of fields.

StefanRijnhart: The number of recruitement sources is likely to be limited in almost any case isn't it? Using the ORM seems fine to me.

StefanRijnhart (approved):

addons/hr_recruitment/migrations/9.0.1.0/post-migration.py
pedrobaeza: But then the previous code was OK except the renaming of the column. No need to make a rewriting of something that is not better.

StefanRijnhart: The perfect is the enemy of the good. The new code is not worse than the old code. In fact it's better because it contains the bugfix. So I'm asking you to reconsider so we can merge this fix.

pedrobaeza: I have given the needed code, so in this case I prefer to have the "perfect" solution. If not, I can do the PR with my proposal.

StefanRijnhart: That's the true optimistic merging style that I still favour! Merge quickly, and create a follow-up PR to fix what could be improved. Proceeding with the merge

/ocabot merge

pedrobaeza:
My suggested version is in #2214

PR #2181, opened 2020-01-28
[MIG] account_banking_payment_export: merge with account_payment_order

yajo:
A module was merged into another.

@Tecnativa TT18838

### DIFF
--- a/openerp/addons/openupgrade_records/lib/apriori.py
+++ b/openerp/addons/openupgrade_records/lib/apriori.py
@@ -126,6 +126,8 @@ merged_modules = [
     ('sale_fiscal_position_update', 'sale'),
     ('sale_documents_comments', 'sale_comment_propagation'),
     # from OCA/bank-payment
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is already handled in https://github.com/OCA/bank-payment/blob/9.0/account_payment_order/migrations/9.0.1.0.0/pre-migration.py#L220 and https://github.com/OCA/bank-payment/blob/9.0/account_payment_order/migrations/9.0.1.0.0/pre-migration.py#L227, so there's something you are not doing well in your migration.

yajo:
I think this is still needed. That code you are linking is the migration script from account_payment_order itself, but that module wouldn't be installed in a v8 because it didn't exist. As such, that script would never get executed.

pedrobaeza:
In OpenUpgrade v9, the module `account_payment` is renamed to `account_payment_order`, and the rest of the "magic" happens.

PR #2184, opened 2020-01-30
[FIX] sale_stock: column product_id is ambiguous

cubells:
An error is raised on migrating from 8.0 to 9.0 because sql query has a product_id column ambiguous.


pedrobaeza (approved):
OK, this is because you have a custom module that adds `product_id` in `stock_location` table, but I agree having a fully qualified column name is always better, so no problem in adding this.

PR #2200, opened 2020-02-21
[9.0][FIX] set currency_id to account.configurable_chart_template

legalsylvain:

the Configurable Chart of Account is : 
in 8.0 : 
-> without currency_id
-> defined in ``account`` module
-> noupdate="1"

in 9.0
-> with currency_id (because it is required)
-> defined in ``l10n_fr_coa`` module
-> noupdate="1"

When upgrading from 8.0 to 9.0 it generates the following error : 

``bad query: ALTER TABLE "account_chart_template" ALTER COLUMN "currency_id" SET NOT NULL
ERROR: column "currency_id" contains null values``

No error.

--

### DIFF
--- a/addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt
@@ -145,6 +145,8 @@ account      / account.chart.template   / company_id (many2one)         : NEW re
 
 
 account      / account.chart.template   / currency_id (many2one)        : now required
+# In pre-migration script, set currency to account.configurable_chart_template data
+# as it is marked as noupdate="1"
 
 account      / account.chart.template   / expense_currency_exchange_account_id (many2one): NEW relation: account.account.template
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think what we have to do is to remove completely that outdated chart template that is no used by any company any way.

legalsylvain:
Hi @pedrobaeza thanks for your answer. Do you propose to have delete commands in the same premigration script ? 

kind regards.

pedrobaeza:
Yes, directly remove all the records belonging to that chart template. Look that in post-migration, an exception must be done as well for this case:

https://github.com/OCA/OpenUpgrade/blob/ccde97eabda64f9e9a4c3be82cbba95e0916fa4f/addons/account/migrations/9.0.1.1/post-migration.py#L83

pedrobaeza:
@legalsylvain are you going to complete this?

pedrobaeza:
Let's merge this although the proposed solution would be better.

PR #2203, opened 2020-02-21
[10.0][IMP] Performance of payment creation for statement lines

StefanRijnhart:
Takes 1 minute and 50 seconds instead of 25 minutes for the ORM version.


pedrobaeza (approved):
Seems OK in general with code review and trusting in the contributor

PR #2207, opened 2020-02-27
[FIX] Get migration tests from module/migration/tests

kos94ok-3D:
--

### DIFF
--- a/odoo/modules/loading.py
+++ b/odoo/modules/loading.py
@@ -290,7 +290,7 @@ def load_module_graph(cr, graph, status=None, perform_checks=True,
 
             # OpenUpgrade: run tests
             if os.environ.get('OPENUPGRADE_TESTS') and package.name is not None:
-                prefix = '.migrations.%s' % adapt_version(package.data['version'])
+                prefix = '.migrations'
                 registry.openupgrade_test_prefixes[package.name] = prefix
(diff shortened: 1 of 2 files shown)


kos94ok-3D:
@StefanRijnhart
Can you review it?

StefanRijnhart:
Oh, but I like having the tests in migrations/\<version\>. Why does it need to change?

kos94ok-3D:
importlib.import_module can not correctly interpret the version folder, so we can not have a structure:
<pre>
Module
    | migration
        | 13.0.1.3
            | tests
</pre>
But we can have:
<pre>
Module
    | migration
        | tests
</pre>
and use tags in order to filtered tests
Rewrite for migration folder?

StefanRijnhart:
Thanks for the explanation! But in order to be able to put tests in module/migration/tests, don't you need to keep at least `.migrations` as the prefix instead of dropping the prefix at all?

kos94ok-3D:
Is it ok?

StefanRijnhart:
Ah thanks for the update. If we have tests in <module>/migrations, then I don't think we need specific tags. Or do you think that will be useful someway?

kos94ok-3D:
Just try to do something new.
I don't need in it.

StefanRijnhart (approved):

PR #2208, opened 2020-02-28
[11.0][IMP] Performance: account post-migration

AdriaGForgeFlow:
This PR tries to improve the computation of the price_total in account invoice lines for those lines that have multiple taxes which should be minority but in some cases it is not.

Current behavior before PR:
Migration in this step took ~ 5 hours.

`2020-01-22 18:03:57,767 21851 DEBUG OpenUpgrade: Compute the rest of the account.invoice.linetotals: 6507506`
`2020-01-22 23:02:47,607 21851 DEBUG OpenUpgrade: 4 rows affected after 0:00:00.000357 running `

Migration in this step takes now ~ 1 hour.

`2020-02-17 10:37:42,970 11933 DEBUG OpenUpgrade: Compute the rest of the account.invoice.linetotals: 6507506`
`2020-02-17 11:42:52,819 11933 INFO OpenUpgrade: account: loading migrations/11.0.1.1/noupdate_changes.xml `

--

### DIFF
--- a/addons/account/migrations/11.0.1.1/post-migration.py
+++ b/addons/account/migrations/11.0.1.1/post-migration.py
@@ -91,21 +91,66 @@ def fill_account_invoice_line_total(env):
     rest_lines = line_obj.search([]) - empty_lines - simple_lines
     openupgrade.logger.debug("Compute the rest of the account.invoice.line"
                              "totals: %s" % len(rest_lines))
-    for line in rest_lines:
+
+    # We group invoice lines that have the same price_unit, discount,
+    # currency_id, quantity, product_id and invoice_line_tax_ids since they
+    # will have the same price_total
+    env.cr.execute("""
+        SELECT
+                STRING_AGG(id::CHARACTER varying, ',') id,
+                price_unit, discount, currency_id, quantity,
+                product_id, tax_id
+        FROM (
+            SELECT aml.*, ailt.tax_id AS tax_id
+            FROM account_invoice_line AS aml
+            LEFT JOIN (
+                    SELECT invoice_line_id,
+                        STRING_AGG(tax_id::CHARACTER varying, ',') tax_id
+                    FROM account_invoice_line_tax
+                    GROUP BY invoice_line_id) AS ailt
+                ON aml.id = ailt.invoice_line_id
+            WHERE aml.id IN %s
+                AND currency_id IS NOT NULL
+                AND product_id IS NOT null
+                AND tax_id IS NOT NULL) AS sub
+        GROUP BY price_unit, discount, currency_id, quantity, product_id,
+                tax_id;
+    """, (tuple(rest_lines.ids),))
+
+    # We iterate over every group of account invoice lines where:
+    # row[0] : (invoice_line_id_1, invoice_line_id_2, ...) invoice_line_ids
+    # row[1] : price_unit
+    # row[2] : discount
+    # row[3] : currency_id
+    # row[4] : quantity
+    # row[5] : product_id
+    # row[6] : (tax_id_1, tax_id_2, ...) invoice_line_tax_ids
+    for row in env.cr.fetchall():
+        # This has been extracted from `_compute_price` method and adapted
+        price = row[1] * (1 - (row[2] or 0.0) / 100.0)
+        invoice_line_tax_ids = env["account.tax"].browse(
+            [int(line_id) for line_id in row[6].split(",")])
+
         # avoid error on taxes with other type of computation ('code' for
         # example, provided by module `account_tax_python`). We will need to
         # add the computation on the corresponding module post-migration.
         types = ['percent', 'fixed', 'group', 'division']
-        if any(x.amount_type not in types for x in line.invoice_line_tax_ids):
+        if any(x.amount_type not in types for x in invoice_line_tax_ids):
             continue
-        # This has been extracted from `_compute_price` method
-        currency = line.invoice_id and line.invoice_id.currency_id or None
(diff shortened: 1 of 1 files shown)


addons/account/migrations/11.0.1.1/post-migration.py
```diff
@@ -91,21 +91,47 @@ def fill_account_invoice_line_total(env):
...
-        if any(x.amount_type not in types for x in line.invoice_line_tax_ids):
-            continue
-        # This has been extracted from `_compute_price` method
-        currency = line.invoice_id and line.invoice_id.currency_id or None
-        price = line.price_unit * (1 - (line.discount or 0.0) / 100.0)
-        taxes = line.invoice_line_tax_ids.compute_all(
-            price, currency, line.quantity, product=line.product_id,
-            partner=line.invoice_id.partner_id,
+
+    openupgrade.logged_query(
```
pedrobaeza: Why doing all of this? Isn't enough to write `price_total` by SQL with previous code?

AdriaGForgeFlow: The problem here was the for loop over 6 million invoice lines that was taking too long when in fact a grouping of those lines can be done. If you find lines that share the same invoice_line_tax_ids, price, currency, quantity and product, for all those invoice lines the price_total will be the same. That is the reason why the first SQL query to group and then the second to compute the price_total for every group. The for loop now is way smaller than before.

pedrobaeza: But that's very unlikely to happen. If the problem is the loop over a great number, then simply use chunk method in openupgradelib for splitting it. I consider this code very ugly and hard to debug/analyze. And a select never uses `loggued_query`!

Please try what I suggest to see the performance.

AdriaGForgeFlow: I do not think what you propose would have any effect at all. I don't see the difference between iterating once over 6 million records than 100 times over 60000 records. Overall time remains the same IMO but maybe I'm missing something here.
Regarding the code I know that is not very clean, I have added comments to make it more understandable.

pedrobaeza: Why don't you try with both suggestions (chunk + update by SQL)? Please have faith :angel:

AdriaGForgeFlow: Yes, I am going to try with your approach

StefanRijnhart: Wait, I like this approach! I happen to be working on a migration in which this can drastically lower the duration as well due to the uniformity of the invoice lines.

pedrobaeza: Let's see any way the numbers with what I have been talking.

AdriaGForgeFlow:
I did a PoC in a smaller database (not to wait too much on the results):

Using [this approach](https://github.com/ForgeFlow/OpenUpgrade/blob/11.0-imp-performance-account-post_migration/addons/account/migrations/11.0.1.1/post-migration.py) (the one in this PR) the duration of the computation of price_total is 5 seconds:
`2020-02-28 11:23:00,137 30712 DEBUG OpenUpgrade: Compute the rest of the account.invoice.linetotals: 227319`
`...`
`2020-02-28 11:23:05,013 30712 DEBUG OpenUpgrade: Compute finished`

Using [this other approach](https://github.com/ForgeFlow/OpenUpgrade/blob/11.0-imp-performance-account-post_migration_2/addons/account/migrations/11.0.1.1/post-migration.py) the duration of the computation of price_total is 4 minutes:
`2020-02-28 11:59:38,435 8797 DEBUG OpenUpgrade: Compute the rest of the account.invoice.linetotals: 227319`
`...`
`2020-02-28 12:03:37,248 8797 DEBUG OpenUpgrade: Compute finished`

pedrobaeza:
OK, so it seems it's interesting. I'm not able to review in deep, but if @StefanRijnhart says that he is interested, I suppose he can review it.

My fear is that the write on the field through SQL doesn't triggered other computes that are needed.

addons/account/migrations/11.0.1.1/post-migration.py
```diff
@@ -91,21 +91,66 @@ def fill_account_invoice_line_total(env):
...
     openupgrade.logger.debug("Compute the rest of the account.invoice.line"
                              "totals: %s" % len(rest_lines))
-    for line in rest_lines:
+
+    # We group invoice lines that have the same price_unit, discount,
+    # currency_id, quantity, product_id and invoice_line_tax_ids since they
+    # will have the same price_total
+    env.cr.execute("""
+        SELECT
+                STRING_AGG(id::CHARACTER varying, ',') id,
```
gdgellatly, line 100: Does uom not need consideration here? I don't know, it just seems it would

gdgellatly: Also, under what circumstances does the product matter, I get that it is used further down, but I don't think it does? I guess if there are a lot of lines for only a few products. But if there a only a few lines and lots of products then maybe not so good. Oh I answered my own uom question the same way. Clearly you don't need uom for the same reason I think you might nott need product.

gdgellatly: I had a look and it seems to me product is only used for 'code' type computation.

pedrobaeza:
As this is debatable (UoM thing for example), let's close it and also as this is not the regular case. If more interest comes, it can be reopen attending the comments.

PR #2210, opened 2020-03-03
[ADD] Additions to the base migration

StefanRijnhart:


kos94ok-3D:
Hi, @StefanRijnhart .
I forgot to rename a field in my PR
<pre>
_field_renames = [
    ('ir.server.object.lines', 'ir_server_object_lines', 'type', 'evaluation_type'),
]
=======================================================================
base         / ir.server.object.lines   / evaluation_type (selection)   : NEW required, selection_keys: ['equation', 'reference', 'value'], req_default: function, hasdefault
base         / ir.server.object.lines   / type (selection)              : DEL required, selection_keys: ['equation', 'reference', 'value'], req_default: function
</pre>
Please add it in your one.
And what do you think about add it in post-migration:
<pre>
def fix_binding_fields(env):
    """
    Adapt binding_type and binding_view_types fields values.
    Useful for manually created actions.
    """
    model_name_lists = [
        'ir.actions.act_url',
        'ir.actions.act_window',
        'ir.actions.act_window_close',
        'ir.actions.actions',
        'ir.actions.client',
        'ir.actions.report',
        'ir.actions.server',
    ]
    for model_name in model_name_lists:
        records = env[model_name].search([('binding_type', '=', 'action_form_only')])
        vals = {
            'binding_type': 'action',
            'binding_view_types': "'form'",
        }
        records.write(vals)
</pre>

StefanRijnhart:
Field renamed, thanks. I don't see action_form_only used anywhere in Odoo 12. What should it have been used for?

StefanRijnhart:
Ah, I see someting in convert.py. Will have a look

StefanRijnhart:
'action_form_only' now migrated correctly. Odoo happens to use table inheritance here (and only here, I believe), see https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/data/base_data.sql#L9-L13. Therefore, we just have to adjust a single table.

Thanks, that was a good catch!

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
```diff
@@ -463,9 +465,16 @@
...
+    'ir_attachment': [
+        ('name', None),
+        ('datas_fname', 'name'),
+    ],
     'res_lang': [
         ('week_start', None),
     ],
+    'ir_server_object_lines': [
+        ('type', 'evaluation_type'),
+    ],
```
kos94ok-3D: I think better use openupgrade.rename_fields than openupgrade.rename_columns in order to keep translation

kos94ok-3D (commented):
rename_fields or rename_columns

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
StefanRijnhart: Agreed! Code adapted. I also removed the use_env=True argument to the migrate decorator as it has been the default for some time now.

kos94ok-3D (approved):
Code review and functionality test.
Thank you for a great job.

StefanRijnhart:
Thank you for the review!

legalsylvain (approved):
LGTM. Thanks. Code review. No test.

PR #2218, opened 2020-03-13
[11.0] update upstream d9b2605eadc

MiquelRForgeFlow:
Updated.

No conflicts.

--


pedrobaeza:
Why so many changes in noupdate file? Have you modified something on the script that gets the file? Maybe you can use old one in this case for keeping a minimal diff?

MiquelRForgeFlow:
@pedrobaeza I don't want to compare record by record. Don't care the amount diff, just changing `res.country` and `res.currency` things up to day. The no_update_changes generator script was changed a bit, yes, in order to have always a unique sorted output.

pedrobaeza:
As said, I prefer that you keep the old version for 11.0 and 12.0. The new sorted one must be used in >=13.0 only.

MiquelRForgeFlow:
@pedrobaeza done!

PR #2219, opened 2020-03-13
[12.0] update upstream 1341cc6ab92

MiquelRForgeFlow:
Updated.

No conflicts.

--


pedrobaeza:
The same about the diff on noupdate file.

pedrobaeza:
I suppose there's no diff now in noupdate, isn't it?

MiquelRForgeFlow:
> I suppose there's no diff now in noupdate, isn't it?

Exactly 😅

PR #2223, opened 2020-03-13
[13.0][IMP] base

MiquelRForgeFlow:
Some improvements.

--

### DIFF
--- a/odoo/addons/base/migrations/13.0.1.3/post-migration.py
+++ b/odoo/addons/base/migrations/13.0.1.3/post-migration.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Andrii Skrypka
 # Copyright 2020 Opener B.V. (stefan@opener.amsterdam)
 # Copyright 2019-2020 Tecnativa - Pedro M. Baeza
+# Copyright 2020 ForgeFlow
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import pytz
 from psycopg2 import sql
@@ -103,3 +104,9 @@ def migrate(env, version):
         env.cr, 'base', 'migrations/13.0.1.3/noupdate_changes.xml')
     fill_ir_attachment_legacy_name(env)
     delete_noupdate_records(env)
+    # Activate back the noupdate flag on the group
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_model_data SET noupdate=True
+        WHERE  module='base' AND name='group_user'""",
+    )
--- a/odoo/addons/base/migrations/13.0.1.3/post-migration.py
+++ b/odoo/addons/base/migrations/13.0.1.3/post-migration.py
@@ -1,15 +1,12 @@
 # Copyright 2020 Andrii Skrypka
 # Copyright 2020 Opener B.V. (stefan@opener.amsterdam)
+# Copyright 2019-2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-
+import pytz
 from psycopg2 import sql
+from datetime import datetime
 from openupgradelib import openupgrade
-
-
-def fix_res_lang_url_code(env):
-    """ Url code is required. """
-    for lang in env['res.lang'].with_context(active_test=False).search([('url_code', '=', False)]):
-        lang.url_code = lang.iso_code or lang.code
+from odoo import fields
 
 
 def fix_res_partner_image(env):
@@ -23,30 +20,17 @@ def fix_res_partner_image(env):
         ResPartner.browse(attachment.res_id).image_1920 = attachment.datas
 
 
-def restore_res_lang_week_start(env):
-    legacy_name = openupgrade.get_legacy_name('week_start')
-    openupgrade.logged_query(
+def res_lang_week_start_map_values(env):
+    openupgrade.map_values(
         env.cr,
-        """
-        SELECT id, %s
-        FROM res_lang
-        WHERE %s IS NOT NULL
-        """ % (legacy_name, legacy_name),
+        openupgrade.get_legacy_name("week_start"),
+        "week_start",
+        [(1, "1"), (2, "2"), (3, "3"), (4, "4"), (5, "5"), (6, "6"), (7, "7")],
+        table="res_lang",
     )
-    grouped_by_week_start = {}
-    for record in env.cr.fetchall():
-        grouped_by_week_start.setdefault(record[1], []).append(record[0])
-    for week_start, lang_ids in grouped_by_week_start.items():
-        env['res.lang'].browse(lang_ids).write({'week_start': str(week_start)})
 
 
-@openupgrade.migrate()
-def migrate(env, version):
-    fix_res_lang_url_code(env)
-    restore_res_lang_week_start(env)
-    openupgrade.load_data(env.cr, 'base', 'migrations/13.0.1.3/noupdate_changes.xml')
-    fix_res_partner_image(env)
-
+def ir_actions_binding_type_views(env):
     # Populate missing binding_model_id values in ir_act_window
     openupgrade.logged_query(
         env.cr,
@@ -62,6 +46,15 @@ def migrate(env, version):
         """ UPDATE ir_actions
         SET binding_view_types = 'form', binding_type = 'action'
         WHERE binding_type = 'action_form_only' """)
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_act_window
+        SET binding_view_types = 'list' WHERE {column}""".format(
+            column=openupgrade.get_legacy_name('multi')),
+    )
+
+
+def fill_ir_attachment_legacy_name(env):
     # Set new name column from old name column if missing
     openupgrade.logged_query(
         env.cr,
@@ -71,3 +64,42 @@ def migrate(env, version):
         AND COALESCE({legacy_name}, '') != '' """).format(
             legacy_name=sql.Identifier(
                 openupgrade.get_legacy_name('name'))))
+
+
+def fill_ir_cron_lastcall(env):
+    """Fill with current date for avoiding notifs of past recurring events."""
+    now = fields.Datetime.context_timestamp(env["ir.cron"], datetime.now())
+    openupgrade.logged_query(
+        env.cr,
+        "UPDATE ir_cron SET lastcall = %s",
+        (fields.Datetime.to_string(now.astimezone(pytz.UTC)),),
+    )
+
+
+def fill_arch_prev(env):
+    """This field is a fallback to previous architecture in case of failure,
+    so we should initially put the same as the current value.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        "UPDATE ir_ui_view SET arch_prev = arch_db")
+
+
+def delete_noupdate_records(env):
+    openupgrade.delete_records_safely_by_xml_id(
+        env, ["base.res_partner_rule", "base.module_website_form_editor",
+              "base.SDD"]
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    res_lang_week_start_map_values(env)
+    fix_res_partner_image(env)
+    ir_actions_binding_type_views(env)
+    fill_ir_cron_lastcall(env)
+    fill_arch_prev(env)
+    openupgrade.load_data(
+        env.cr, 'base', 'migrations/13.0.1.3/noupdate_changes.xml')
+    fill_ir_attachment_legacy_name(env)
+    delete_noupdate_records(env)
--- a/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Andrii Skrypka
 # Copyright 2020 Opener B.V. (stefan@opener.amsterdam)
 # Copyright 2019-2020 Tecnativa - Pedro M. Baeza
+# Copyright 2020 ForgeFlow
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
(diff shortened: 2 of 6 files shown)


pedrobaeza:
Wow, thanks for recovering and adapting my experimental branch that I started before OCA RfQ got frozen! It seems there are still things to do in base. I will check later because I don't remember what I did and see why the Travis error (for sure is a mistake as I didn't test anything in that moment).

MiquelRForgeFlow:
> Wow, thanks for recovering and adapting

Obviously I would do that

> see why the Travis error

There were some errors but they are already fixed. Now it's green :green_apple:

kos94ok-3D:
Why do you keep only two image sizes?
Are others not important?

pedrobaeza:
@kos94ok-3D please make the comment inline for knowing the context. Without remembering too much, I would say that the rest were not compatible (the expected size).

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
```diff
@@ -494,18 +82,393 @@ def fix_lang_table(cr):
...
+
+def rename_image_attachments(env):
+    for old, new in [("image", "image_1920"), ("image_medium", "image_128")]:
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE ir_attachment SET res_field = %s
+            WHERE res_field = %s AND res_model = 'res.partner'""",
+            (new, old),
+        )
```
kos94ok-3D: Here we just rename field by SQL and Odoo won`t recompute other related fields.
So I used ORM in post-migration.
Before migration
изображение
After migration
изображение

kos94ok-3D: My function which was removed in this commit
<pre>
def fix_res_partner_image(env):
    ResPartner = env['res.partner']
    attachments = env['ir.attachment'].search([
        ('res_model', '=', 'res.partner'),
        ('res_field', '=', 'image'),
        ('res_id', '!=', False),
    ])
    for attachment in attachments:
        ResPartner.browse(attachment.res_id).image_1920 = attachment.datas
</pre>

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
```diff
@@ -494,18 +82,393 @@ def fix_lang_table(cr):
...
+            ("base.module_category_human_resources",
+             "base.module_category_human_resources_employees"),
+            ("base.module_category_manufacturing",
+             "base.module_category_manufacturing_manufacturing"),
+            ("base.module_category_mass_mailing",
+             "base.module_category_marketing_email_marketing"),
+            ("base.module_category_payment_acquirer",
+             "base.module_category_accounting_payment"),
+            ("base.module_category_point_of_sale",
+             "base.module_category_sales_point_of_sale"),
```
kos94ok-3D: The category was moved to a point_of_sale module

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
MiquelRForgeFlow: @kos94ok-3D If we use your function, should we add the image_medium -> image_128 rename also?

kos94ok-3D: enough only image_1920, other fields will be recomputed automatically

MiquelRForgeFlow: thanks, I will take into account for future image renames in other models :+1:

pedrobaeza: In general, we should avoid ORM computations that are very costly. For this case, I don't know if it can be avoidable, but I will check.

pedrobaeza:
Hey, @MiquelRForgeFlow I prefer that you don't change that code until I verify it.

MiquelRForgeFlow:
Too late, but I can redo it if you want, no problem. Just say it.

odoo/addons/base/models/ir_model.py
```diff
@@ -1994,7 +1994,9 @@ def _process_end(self, modules):
                     # Note: A regular module update in a regular (non-openupgrade) Odoo instance will remove this xmlids
                     record = self.env[model].browse(res_id)
                     if record.exists():
-                        continue
+                        if model != 'ir.ui.view':  # delete views always
+                            continue
```
StefanRijnhart: ~~Your commit message mentions "delete obsolete xmlids in case of views". But if I understand correctly, this code prevents the actual record (actions, menu items, record rules, access rules) from being deleted if they no longer occur in the module (except views). That would not be good, would it? Or let's say I'm curious why you'd want to do this.~~ Looks like I misunderstood. Maybe I messed up the framework port? I'll have a look.

StefanRijnhart: Can you have a look at what happened here: https://github.com/OCA/OpenUpgrade/pull/2011/commits/0ddea82b61cab21f6202b2560a748eae6991cd5f. I think we should reinstate the deletion of all records that was removed in this code.

MiquelRForgeFlow: why?

StefanRijnhart: It seems to me that otherwise, data that was removed from the modules (actions, menu items, record rules, access rules) are not removed from the migrated database.

MiquelRForgeFlow: In https://github.com/OCA/OpenUpgrade/commit/0ddea82b61cab21f6202b2560a748eae6991cd5f says "`# Note: A regular module update in a regular (non-openupgrade) Odoo instance will remove this xmlids`"

StefanRijnhart: I'm talking about the records. https://github.com/OCA/OpenUpgrade/commit/0ddea82b61cab21f6202b2560a748eae6991cd5f puts a `continue` before the actual record is unlinked.

Admittedly, it's confusing. The original Odoo code makes the distinction between existing *records* and non-existing records. When removing an existing record, its XMLID is deleted automatically. If the record does not exist, Odoo takes extra care to remove the dangling XMLID. We want to avoid the latter in OpenUpgrade, but not the former.

MiquelRForgeFlow: I don't know what you are trying to say, sorry. This is confusing. It was already discussed in https://github.com/OCA/OpenUpgrade/pull/2011#issuecomment-559075224.

StefanRijnhart: Yes, it was discussed I believe precisely for the same reason, and I don't understand why the faulty code got through after all (well, I made a mistake when I approved).

Let's have a look at a simple case:

In Odoo 13, a menu item was removed: https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/base/migrations/13.0.1.3/openupgrade_analysis.txt#L153

This menu item in Odoo 12 was not in a noupdate statement: https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/wizard/base_language_install_views.xml#L39-L43

OpenUpgrade, like Odoo itself, depends on such records to be removed automatically when the module is upgraded. The code that does that was removed/disabled by you in https://github.com/OCA/OpenUpgrade/commit/0ddea82b61cab21f6202b2560a748eae6991cd5f.

That means, that on an upgraded database, the menu item is still present:
```
odoo12base=> select latest_version from ir_module_module where name = 'base';
 latest_version 
----------------
 13.0.1.3
(1 row)

odoo12base=> select module, name, noupdate, res_id from ir_model_data
where name = 'menu_view_base_language_install';
 module |              name               | noupdate | res_id 
--------+---------------------------------+----------+--------
 base   | menu_view_base_language_install | f        |     49
(1 row)

odoo12base=> select name from ir_ui_menu where id = 49;
        name        
--------------------
 Load a Translation
(1 row)
```

I believe this is wrong. Do you agree?

MiquelRForgeFlow: Not really, because after the migration, you will run an "update all" in a regular (non-openupgrade) Odoo instance that will remove that menu item.

StefanRijnhart: Then we disagree. Why change this behaviour in OpenUpgrade 13.0? What are the advantages? Note that you are already making the exception for views in this PR.

MiquelRForgeFlow: Well, I am not hard on this. Just done as a request by @pedrobaeza . The reason is explained in this comment: https://github.com/OCA/OpenUpgrade/pull/2011#issuecomment-559071019

And yes, I made an exception for views, because if I didn't, then later in the "update all" some deprecated views (before  deletion) are validated and break.

pedrobaeza: @StefanRijnhart the goal behind this is double:

* When you concatenate several migrations (for example doing a migration from 12 to 14). There are a lot of records that can be reused in later version, and you don't lose possible extra data attached to that records. Not so frequent.
* And the main one: if there's something that disappears in core, but you restore it later in a custom module. This way, you don't lose that records and possible extra data.

StefanRijnhart: Indeed, if you leave the records around they break the consistency of the data. This goes for views as much as for other data types I'm sure.

@pedrobaeza I doubt you have a concrete case because this is a huge, experimental change in behaviour.

pedrobaeza: I had the old `crm_phonecall` case linked by @MiquelRForgeFlow, but if you think it's so conflictive, let's revert it.

StefanRijnhart: I would very much appreciate that. We can always reintroduce this new approach when there is a good case in OpenUpgrade 13. If ou13 is more mature then we can estimate the consequences much better.

MiquelRForgeFlow: done

odoo/addons/base/migrations/13.0.1.3/post-migration.py
```diff
@@ -1,52 +1,58 @@
...
-@openupgrade.migrate()
-def migrate(env, version):
-    fix_res_lang_url_code(env)
-    restore_res_lang_week_start(env)
-    openupgrade.load_data(env.cr, 'base', 'migrations/13.0.1.3/noupdate_changes.xml')
-    fix_res_partner_image(env)
 
+def ir_actions_binding_type_views(env):
+    # Convert obsolete 'action_form_only' binding type to binding_view_types
+    for table in ACTIONS_TABLES:
```
StefanRijnhart: The previous version only updated ir_actions on purpose because the other tables inherit from it on the SQL level. Did that not work as I expected?

odoo/addons/base/migrations/13.0.1.3/post-migration.py
```diff
@@ -56,18 +62,59 @@ def migrate(env, version):
...
+    # Set new name column from old name column if missing
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_attachment
         SET name = {legacy_name}
         WHERE COALESCE(name, '') = ''
-        AND COALESCE({legacy_name}, '') != '' """).format(
-            legacy_name=sql.Identifier(
-                openupgrade.get_legacy_name('name'))))
+        AND COALESCE({legacy_name}, '') != '' """.format(
```
StefanRijnhart: I'm very interested in using the composite queries that Pedro has introduced me to. Why do you replace it by ordinary string replacement? Did I not do it right first time?

MiquelRForgeFlow: Yes, you did it right, and I can revert it if all of you want, but it happens that I cannot see a reason to use the composite one instead of the ordinary one.

odoo/addons/base/migrations/13.0.1.3/post-migration.py
MiquelRForgeFlow: I only checked that the other actions have the same columns, so I thought that better update those columns by query instead of let odoo handle it.

StefanRijnhart:
This PR is full of arbitrary changes to the existing version. It seems to me that you just replaced the exsiting version with the one that you just had lying around already. Can you please do this again and only pick the functionality that was missing?

pedrobaeza:
@MiquelRForgeFlow can you check what @StefanRijnhart says. I haven't compared both versions.

MiquelRForgeFlow:
@StefanRijnhart just tell me what changes you don't like and I will revert them.

pedrobaeza:
@MiquelRForgeFlow I think it's better that you recheck my code and only add strictly missing changes. Other cosmetic or superfluous change should be avoided in favor of diff friendly.

MiquelRForgeFlow:
I didn't now cosmetic changes were unwelcome here. Maybe when I have time I'll check again.

pedrobaeza:
@MiquelRForgeFlow this has been the guideline always. Only migrating a module or doing a huge refactoring can be an occasion for that, but not this time, and more as we come late to the party when the module has been already covered by 2 persons. Doing them more difficult to review what's new and what's not is not gentle. It would be tolerable if you split that possible cosmetic changes in one commit, and the actual changes in another.

MiquelRForgeFlow:
@StefanRijnhart @pedrobaeza all "cosmetic" changes have been removed. Anything else? Can we proceed?

odoo/addons/base/migrations/13.0.1.3/post-migration.py
pedrobaeza: Indeed it's that way:

https://github.com/OCA/OpenUpgrade/blob/d607b670588e96917f89d8b6221bcc65e48345f3/odoo/addons/base/data/base_data.sql#L9-L13

so I'm reverting this change.

StefanRijnhart: @pedrobaeza do I understand correctly that this code will be reverted only to update ir_actions?

pedrobaeza: Yes, pushing in a few minutes

StefanRijnhart:
Thanks for understanding. I'm not necessarily against cosmetic changes but I feel it's more welkoming for new contributors if their first contribution is not immediately overwritten by a new version. And I think it's important that the openupgrade contributors show that they know how table inheritance works ;-)

pedrobaeza:
Pushed the version reverting `ir_actions` thing. It should be now OK for merging. @StefanRijnhart do you agree?

StefanRijnhart:
@pedrobaeza yes, thanks!

PR #2224, opened 2020-03-18
[11.0][FIX] base: avoid unique_name_per_day sql constrain

MiquelRForgeFlow:
https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_currency.py#L235
This constrain is new in v11.

As a result, in v10 you could have two records with the same name, currency_id, company_id and not violate the constrain, but in v11 the name is converted from datetime to date, and then it fails.

For example, in v10 you have two records with:

2020-01-01 09:00:00 | EUR | My Company
2020-01-01 10:00:00 | EUR | My Company

When you migrate to v11:

2020-01-01 | EUR | My Company
2020-01-01 | EUR | My Company

Then the constrain fails here.

--

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/pre-migration.py
@@ -8,13 +8,9 @@ from odoo.addons.openupgrade_records.lib import apriori
 from openupgradelib import openupgrade
 
 
-# backup of
-# - datetime field because it changes to date field
-column_copies = {
-    'res_currency_rate': [('name', None, None)],
-}
 column_renames = {
     'ir_actions': [('usage', None)],
+    'res_currency_rate': [('name', None)],
 }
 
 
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Make sense. Thanks ! 

(even if in the new odoo design, I'm don't understand this constraint. It is not possible to have two rate per day ? Well. Anyway, another story).

MiquelRForgeFlow:
Also fixed now the tz from first user's company to create_uid user, as it makes more sense.

JordiBForgeFlow:
@AdriaGForgeFlow

MiquelRForgeFlow:
Now it's green :green_apple:

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -94,6 +90,41 @@ def fill_cron_action_server_pre(env):
     )
 
 
+def set_currency_rate_dates(env):
+    """Set currency rate date per company's most popular timezone.
```
pedrobaeza: This docstring is not accurate now. It should be by creation user timezone

odoo/addons/base/migrations/11.0.1.3/pre-migration.py
```diff
@@ -94,6 +90,41 @@ def fill_cron_action_server_pre(env):
...
+        WHERE rcr.create_uid = ru.id
+        """, (
+            AsIs(openupgrade.get_legacy_name('name')),
+        ),
+    )
+    # Now delete duplicated currency rates due to 'unique_name_per_day'
+    # sql constrain and name type changed from datetime to date"""
+    openupgrade.logged_query(
+        cr, """
+        DELETE
```
pedrobaeza: Use window function that is more performant:
```
DELETE FROM res_currency_rate
WHERE id IN (
    SELECT id
    FROM (
        SELECT id, row_number() over (partition BY name, currency_id, company_id ORDER BY id) AS rnum
        FROM res_currency_rate
    ) t
    WHERE t.rnum > 1);
```

pedrobaeza (approved):

AdriaGForgeFlow (approved):

PR #2236, opened 2020-03-20
[12.0] base: Revert "[IMP] OpenUpgrade: don't delete xmlids linked to nothing"

MiquelRForgeFlow:
This reverts commit 93d0f7ea6d1655458f9034ac1328714b29395ac6.

It was long ago, I didn't remember. And that was wrong.

--

### DIFF
--- a/odoo/addons/base/models/ir_model.py
+++ b/odoo/addons/base/models/ir_model.py
@@ -1728,9 +1728,6 @@ class IrModelData(models.Model):
                                 res_id, model)
                         # /OpenUpgrade
                     else:
-                        # OpenUpgrade: don't delete xmlids. Use database_cleanup instead
-                        continue
-                        # /OpenUpgrade
                         bad_imd_ids.append(id)
         if bad_imd_ids:
             self.browse(bad_imd_ids).unlink()


pedrobaeza:
But this is not the same as we are debating in the other place. This is for XML-IDs, not for records themselves.

MiquelRForgeFlow:
We debated this in https://github.com/OCA/OpenUpgrade/pull/2011. In that PR, I initially proposed apply the change that now I am trying to revert. In that PR, Stefan said that there was no point to preserve dangling XML ids (entries in ir_model_data that refer to a res_id that is no longer in the model's table, or simply, xmlids linked to nothing) so I didn't apply it finally. Thus, that's why I am reverting it here.

pedrobaeza:
Check https://github.com/OCA/OpenUpgrade/pull/2223#discussion_r395517388 where the pointed commit is not this one (but one with same name): 0ddea82b61cab21f6202b2560a748eae6991cd5f

MiquelRForgeFlow:
I am not speaking of https://github.com/OCA/OpenUpgrade/pull/2223#discussion_r395517388 nor https://github.com/OCA/OpenUpgrade/commit/0ddea82b61cab21f6202b2560a748eae6991cd5f. Do not relate one thing with the other or we will never get out of this stupid loop.

What I am saying is that in https://github.com/OCA/OpenUpgrade/pull/2011 I initially proposed the change I am trying to revert now in this PR. Stefan pointed that there was no sense to save xmlids linked to nothing and thus pointed to the correct solution that I finally implemented in that branch (https://github.com/OCA/OpenUpgrade/commit/0ddea82b61cab21f6202b2560a748eae6991cd5f) that now in https://github.com/OCA/OpenUpgrade/pull/2223 we have reverted.

pedrobaeza:
OK, I have a little mess with all of that:

/ocabot merge

StefanRijnhart:
Ok, I think I do agree with removing this because indeed it does not make any sense to preserve XMLID's of which the record is already removed, like you say.

Note that https://github.com/OCA/OpenUpgrade/commit/cdc470e98791b5e4cbf2f398a2029264e8e8d341#diff-bb1e3d0f2e95b26bcd024950824e022fR1955 preserves noupdate records and their XML ids. I think that should cover the important cases.

pedrobaeza:
Uhm, isn't that what this PR is removing? I confess I'm pretty lost right now

StefanRijnhart:
Odoo by default removes XML ids for which the record no longer exists. Miquel introduced a change earlier on so that those dangling XML ids were preserved. This PR reverts that change.

pedrobaeza:
Then should we continue with this PR or not according your opinion?

StefanRijnhart:
As you guessed, yes please continue ;-)

PR #2247, opened 2020-03-20
[13.0][ADD] uom module

kos94ok-3D:
--


legalsylvain (approved):
Make sense. Thanks !

addons/uom/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,9 @@
+---Models in module 'uom'---
+---Fields in module 'uom'---
+uom          / uom.category             / measure_type (selection)      : selection_keys is now '['length', 'unit', 'volume', 'weight', 'working_time']' ('['length', 'time', 'unit', 'volume', 'weight']')
+# Done in post-migration script
```
pedrobaeza: You have to rename the variable to `None` in pre and then map on post. And remember the format for these annotations I talked about in the other issue.

pedrobaeza (changes requested):

addons/uom/migrations/13.0.1.0/openupgrade_analysis_work.txt
kos94ok-3D: Is it ok?

pedrobaeza:
Sorry, I told you wrong... We need to copy column, not to rename it.

pedrobaeza:
Let me explain the why of this for not getting the impression of picky choice: if you let the incorrect value, any ORM operation writing that records can break the update process due to the incorrect selection value. About the copying process instead of direct mapping, it's for not losing the original value. In this case, being a direct renaming, it's not very critical, but in other cases a value is mapped to other existing value, and there's no way of differentiating between the original ones and the recently mapped ones.

I have also told you bad that the mapping is on post. It must be on pre. Sorry for not being very focused these days.

kos94ok-3D:
@pedrobaeza 
What about it?
изображение

pedrobaeza:
Not in this case as explained above about the risk, and there's nothing counter-productive in doing it in pre.

kos94ok-3D:
@pedrobaeza Done

addons/uom/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,9 @@
+---Models in module 'uom'---
+---Fields in module 'uom'---
+uom          / uom.category             / measure_type (selection)      : selection_keys is now '['length', 'unit', 'volume', 'weight', 'working_time']' ('['length', 'time', 'unit', 'volume', 'weight']')
+# DONE: pre-migration: renamed `measure_type` to None then mapped 'time' value to 'working_time'
```
pedrobaeza:
```suggestion
# DONE: pre-migration: copy `measure_type` and then mapped 'time' value to 'working_time'
```

pedrobaeza (approved):

StefanRijnhart:
Not sure why the Travis job is not marked as success for the gitbot PRs

/ocabot merge

PR #2254, opened 2020-03-25
[12.0][FIX] base: add remove_tables_fks

MiquelRForgeFlow:
I wanted to do it several days ago, and now I make it before going to lunch =P

--

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/pre-migration.py
@@ -70,6 +70,10 @@ xmlid_renames = [
     ('stock.incoterm_FOB', 'account.incoterm_FOB'),
 ]
 
+_obsolete_tables = (
+    "stock_location_path",
+)
+
 
 def switch_noupdate_flag(env):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't this be on `stock` module instead?

MiquelRForgeFlow:
Why? It just happens it's only one table, related to `stock`, but there could be more tables and ones unrelated to `stock`. If in the future we want to add more tables, for example, many2many tables, then it will be easy to add them in this method instead of add again the method. It's the same done in https://github.com/OCA/OpenUpgrade/commit/565ab92573f0848afab9270522ee342a45bb5e53, that later was filled with https://github.com/OCA/OpenUpgrade/commit/ddb89a8bc48a629d5072a92e8e2758bb2c6f80d3.

pedrobaeza (approved):
Yeah, there are 2 conflicting goals here:

- Have all obsolete models in the same place.
- Reduce noise in the log. I'm analyzing the openupgradelib method and indeed there's no noise if the module is not installed, so OK to put it here.

PR #2256, opened 2020-03-27
[FIX] Commit module state to database before migrating next module

### LINKED ISSUE #2255: Question about OpenUpgrade migration process
Still in the process of migrating a database from 11.0 to 12.0 I just go an issue that made me asking for the migration process.

At the first attempt to migrate [A], I got an exception raised from UoM `_check_category_reference_uniqueness` because I had two entries set as reference for the same category.
After fixing the DB manualy, I made a new attempt to migrate [B], but a new exception is raised about a non existing column from the resource module pre-migration script.

After watching logs, I can see that the resource module has already been migrated during [A] (pre and post migration scripts are executed) and looking at the database schema, I can see that the column `openupgrade_legacy_12_0_tz` is already there.

So my question is :
- Does the migration process is made using transaction ?
- Why the resource module version has not been written during [A] since the exception is raised from another module migration (uom) ?

[A: migrate-11-07_26__120_Upgrade_dec_database_to_OpenERP_Odoo_12_via_openupgrade.log](https://github.com/OCA/OpenUpgrade/files/4381673/migrate-11-07_26__120_Upgrade_dec_database_to_OpenERP_Odoo_12_via_openupgrade.log)
[B: migrate-14-14_45__120_Upgrade_dec_database_to_OpenERP_Odoo_12_via_openupgrade.log](https://github.com/OCA/OpenUpgrade/files/4381676/migrate-14-14_45__120_Upgrade_dec_database_to_OpenERP_Odoo_12_via_openupgrade.log)

ypapouin:
fixes #2255
--

### DIFF
--- a/odoo/modules/loading.py
+++ b/odoo/modules/loading.py
@@ -303,6 +303,10 @@ def load_module_graph(cr, graph, status=None, perform_checks=True,
             ver = adapt_version(package.data['version'])
             # Set new modules and dependencies
             module.write({'state': 'installed', 'latest_version': ver})
+            # OpenUpgrade: commit module_n state and version immediatly
+            # to avoid invalid database state if module_n+1 raises an
+            # exception
+            cr.commit_org()
 
             package.load_state = package.state
             package.load_version = package.installed_version


StefanRijnhart (approved):
Thanks!

pedrobaeza:
Is this needed in v13?

StefanRijnhart:
Porting as we speak

StefanRijnhart:
https://github.com/OCA/OpenUpgrade/pull/2257

PR #2262, opened 2020-03-30
[13.0][MIG] product

MiquelRForgeFlow:
Migration of `product`.

--


addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,80 @@
...
+def rename_image_attachments(env):
+    for old, new in [("image_variant", "image_variant_1024")]:
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE ir_attachment SET res_field = %s
+            WHERE res_field = %s AND res_model = 'product.product'""",
+            (new, old),
+        )
+    for old, new in [("image", "image_1024"), ("image_medium", "image_128")]:
```
MiquelRForgeFlow: @kos94ok-3D do you think we should this through ORM?

kos94ok-3D: Yes, I do.
And for product.product model too, because in v13 fields are stored.
In v12
image
In v13
image
@pedrobaeza What do you think about it? Do you found another solution?

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
+---Models in module 'product'---
+obsolete model product.price.history
+# NOTHING TO DO: totally removed, replaced in a weird way by 'stock.valuation.layer' in stock_account
```
pedrobaeza: I think we need to convert one into the other. Not reviewed deeply any way.

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+
+---Fields in module 'product'---
+product      / product.attribute        / product_tmpl_ids (many2many)  : NEW relation: product.template, isfunction: function, stored
+# NOTHING TO DO: is a new compute, so we could precreate and fill to speed up, but not needed because usually there are few records
```
pedrobaeza, line 7: I don't agree. I have a customer with 2k products coming from the same template, so if there's easy pre-fill, let's do it now.

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+
+product      / product.attribute.value  / pav_attribute_line_ids (many2many): NEW relation: product.template.attribute.line
+# NOTHING TO DO: New many2many
```
pedrobaeza:
```suggestion
# NOTHING TO DO: Inverse many2many of an existing m2m table
```

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+product      / product.template         / weight_uom_name (char)        : now a function
+product      / product.template.attribute.value / product_tmpl_id (many2one)    : now related
+# NOTHING TO DO
+
+product      / product.pricelist.item   / active (boolean)              : NEW isrelated: related, stored
+# DONE: pre-migration: precreated and filled column to speed up
+
+product      / product.pricelist.item   / compute_price (selection)     : now required, req_default: function
+product      / product.pricelist.item   / pricelist_id (many2one)       : now required, req_default: function
+# DONE: post-migration: filled the field when empty
```
pedrobaeza: This should be done on pre-migration instead for avoiding a temporary warning (and there's no reason for not doing it there).

addons/product/migrations/13.0.1.2/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        env.cr, """
+        UPDATE product_pricelist_item
+        SET compute_price = 'fixed'
+        WHERE compute_price IS NULL
+        """,
+    )
+    # fill pricelist_id
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE product_pricelist_item ppi
```
pedrobaeza: You must remove these records instead, as they are not being taken into account in previous versions.

addons/product/migrations/13.0.1.2/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        SET attribute_id = ptal.attribute_id
+        FROM product_template_attribute_line ptal
+        WHERE ptav.attribute_line_id = ptal.id""",
+    )
+
+
+def fill_product_variant_combination_table(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE product_variant_combination pvc
```
pedrobaeza: This can't be updated if there's no previous records on it (it's a new table), and no previous table renaming has been done. Why not renaming `product_attribute_value_product_product_rel` to `product_variant_combination` and the corresponding columns?

addons/product/migrations/13.0.1.2/pre-migration.py
pedrobaeza: For avoiding excessive computation I would call `_compute_image_XXX` of the missing images sizes only, but renaming those that can be reused. Note that this is what is probably going to happen on the missing fields. We have to check how ORM behaves in attachment case.

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+
+product      / product.template         / has_configurable_attributes (boolean): NEW isfunction: function, stored
+# NOTHING TO DO: we could try to precreate and fill in pre-migration to speed up, but it a priori is not needed.
+
+product      / product.template.attribute.line / active (boolean)              : NEW hasdefault
+product      / product.template.attribute.value / ptav_active (boolean)         : NEW hasdefault
+# NOTHING TO DO: New field that has default, will be automatically filled
+
+product      / product.template.attribute.value / attribute_id (many2one)       : is now stored
+# DONE: post-migration: filled field (done in post-migration because it's a related to a new field)
```
pedrobaeza, line 79: You should pre-create the column for avoiding that Odoo tries to fill it by ORM with its related value (still blank).

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+
+product      / product.template.attribute.line / product_template_value_ids (many2many): is now stored
+product      / product.template.attribute.line / product_template_value_ids (many2many): not a function anymore
+product      / product.template.attribute.line / product_template_value_ids (many2many): type is now 'one2many' ('many2many')
+product      / product.template.attribute.value / attribute_line_id (many2one)  : NEW relation: product.template.attribute.line, required
+# DONE: post-migration: filled field (using the same compute method of v12)
```
pedrobaeza: I don't get the comment "using the same compute method of v12". It's an SQL, not a compute. I suppose you mean the same approach than in v12, but in v12 where?

addons/product/migrations/13.0.1.2/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_product_pricelist_item_prices(env):
+    # fill compute_price
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE product_pricelist_item
+        SET compute_price = 'fixed'
```
pedrobaeza: I think a better value is "Formula" (not sure technical value), although this shouldn't happen in records done by UI.

pedrobaeza (changes requested):
I can perform the changes if you prefer.

addons/product/migrations/13.0.1.2/pre-migration.py
MiquelRForgeFlow: well, if I didn't misunderstand from what we did in `base`, the point of doing the rename by orm is that then the missing image fields are recomputed automatically by resizing the existing ones, right?

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
MiquelRForgeFlow: but this should be done in `stock_account`, not here

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
MiquelRForgeFlow: ok

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
MiquelRForgeFlow: yeah, I meant "the same approach method"

addons/product/migrations/13.0.1.2/post-migration.py
MiquelRForgeFlow: why do you think so?

addons/product/migrations/13.0.1.2/post-migration.py
MiquelRForgeFlow: We can't do that rename you suggest because they are different models. The `product_attribute_value_product_product_rel` is for product.attribute.value and `product_variant_combination` is for product.template.attribute.value. The problem is that in v12, the relation table for product.template.attribute.value and product.product is a nonstored computed one.

pedrobaeza: But then we should insert, not update. I will check in deep.

addons/product/migrations/13.0.1.2/post-migration.py
pedrobaeza: Yeah, they are at the end the same valid, and it's unlikely to happen.

addons/product/migrations/13.0.1.2/post-migration.py
MiquelRForgeFlow: Yeah, probably should be a insert. The script the same, but with an insert instead of update. It seems a silly mistake.

pedrobaeza:
I will perform the changes and upload them here, so you can continue with the rest, @MiquelRForgeFlow .

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
pedrobaeza: As it means to pre-create the m2m table, columns, indexes, etc, let's keep it for now as computed and will see later performance bottlenecks.

addons/product/migrations/13.0.1.2/pre-migration.py
pedrobaeza: Yes, we need to reprocess them after checking with a demo migration

pedrobaeza (approved):

PR #2265, opened 2020-04-01
[13.0][MIG] calendar

MiquelRForgeFlow:
Migration of `calendar`.

--


addons/calendar/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,16 @@
+---Models in module 'calendar'---
+---Fields in module 'calendar'---
+calendar     / calendar.alarm           / alarm_type (selection)        : NEW required, selection_keys: ['email', 'notification'], req_default: function, hasdefault
+calendar     / calendar.alarm           / type (selection)              : DEL required, selection_keys: ['email', 'notification'], req_default: function
+# DONE: pre-migration: renamed fields
+
+calendar     / calendar.event           / event_tz (selection)          : NEW selection_keys: function, hasdefault
+# NOTHING TO DO: New field that has default, will be automatically filled
```
kos94ok-3D, line 8: I don`t know why but I have a null value in an event_tz column after migration.
Do you have the same problem?

MiquelRForgeFlow: I will check. But null value in this field means that `self.env.context.get('tz') or self.user_id.tz` is falsy 🤔

MiquelRForgeFlow: Well, the field is not required, so no serious problem if it has null values.

kos94ok-3D: You are right:
self.env.context.get('tz') == None
and
self.user_id == res.users() => self.user_id.tz == False

kos94ok-3D (approved):

addons/calendar/migrations/13.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: I'm pushing a revision making sure this field is empty for all for preserving v12 behavior.

pedrobaeza (approved):
With my added changes

PR #2276, opened 2020-04-08
[13.0][MIG] sales_team

Yasaie:
Migration of `sales_team`
--


kos94ok-3D:
Why not add a migration state?

MiquelRForgeFlow (approved):

addons/sales_team/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,8 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
```
MiquelRForgeFlow:
```suggestion
@openupgrade.migrate()
```
by default is already `use_env=True`

kos94ok-3D (approved):

PR #2278, opened 2020-04-08
[13.0][MIG] payment

Yasaie:
Migration of `payment`
--


addons/payment/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
...
+    ],
+}
+
+_field_renames = [
+    ('payment.acquirer', 'payment_acquirer', 'image_medium', 'image_128'),
+    ('payment.acquirer', 'payment_acquirer', 'environment', 'state'),
+]
+
+_xmlid_renames = [
+    ('payment.payment_acquirer_ogone', 'payment_acquirer_ingenico'),
```
MiquelRForgeFlow:
```suggestion
    ('payment.payment_acquirer_ogone', 'payment.payment_acquirer_ingenico'),
```

addons/payment/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,47 @@
+---Models in module 'payment'---
+new model payment.link.wizard [transient]
+# NOTHING TO DO
+
+---Fields in module 'payment'---
+payment      / account.invoice          / transaction_ids (many2many)   : DEL relation: payment.transaction
+payment      / account.move             / transaction_ids (many2many)   : NEW relation: payment.transaction
+# NOTHING TO DO: many2many table didn't changed
```
MiquelRForgeFlow: but you should update the invoice_id values of this account_invoice_transaction_rel table.

addons/payment/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+from openupgradelib import openupgrade
+
+unlink_by_xmlid = [
+    'payment.payment_acquirer_custom'
+]
+
+
+def map_payment_acquirer_state(cr):
+    openupgrade.map_values(
```
MiquelRForgeFlow: don't use map_values in this case, because I think it's a conditional case: map 'prod' option to 'enabled' if 'website_published = True , else map 'prod' to 'disabled'

addons/payment/migrations/13.0.1.0/noupdate_changes.xml
```diff
@@ -19,37 +19,37 @@
...
     <field name="display_as">Credit Card (powered by Authorize)</field>
     <field name="image_128" type="base64" file="payment_authorize/static/src/img/authorize_icon.png"/>
   </record>
   <record id="payment_acquirer_buckaroo" model="payment.acquirer">
-    <field name="image" type="base64" file="payment_buckaroo/static/src/img/buckaroo_icon.png"/>
+<!--    <field name="image" type="base64" file="payment_buckaroo/static/src/img/buckaroo_icon.png"/>-->
     <field name="display_as">Credit Card (powered by Buckaroo)</field>
     <field name="image_128" type="base64" file="payment_buckaroo/static/src/img/buckaroo_icon.png"/>
   </record>
   <record id="payment_acquirer_paypal" model="payment.acquirer">
```
MiquelRForgeFlow, line 51: you have to manually add the noupdate changes for the payment_acquirer_ingenico. An easy way to do that, is to manually rename the payment_acquirer_ogone to payment_acquirer_ingenico in a local v12 and then run the noupdate generator script in v13

addons/payment/migrations/13.0.1.0/openupgrade_analysis_work.txt
Yasaie: why? and how should i do that?

addons/payment/migrations/13.0.1.0/noupdate_changes.xml
Yasaie: thanks. done

addons/payment/migrations/13.0.1.0/post-migration.py
Yasaie: done

addons/payment/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,35 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+_column_copies = {
+    'payment_acquirer': [
+        ('website_published', None, None),
```
MiquelRForgeFlow: put in field_renames

addons/payment/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
...
+        SET state = 'enabled'
+        WHERE state = 'prod' AND {website_published} = TRUE""".format(
+            website_published=openupgrade.get_legacy_name('website_published')
+        )
+    )
+    openupgrade.logged_query(
+        cr, """
+        UPDATE payment_acquirer
+        SET state = 'disabled'
+        WHERE state = 'prod' AND {website_published} = FALSE""".format(
```
MiquelRForgeFlow:
```suggestion
        SET state = CASE WHEN {website_published} = TRUE THEN 'enabled'
            ELSE 'disabled' END
        WHERE state = 'prod'""".format(
```

addons/payment/migrations/13.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: Because the invoice_id points to a different place. Before it was pointing to some id in account_invoice and now should point to some id in account_move, and thus probably not the same id.

add in post-migration
```python
def update_account_invoice_transaction_rel_table(env):
    openupgrade.logged_query(
        env.cr, """
        UPDATE account_invoice_transaction_rel aitr
        SET invoice_id = am.id
        FROM account_move am
        WHERE am.old_invoice_id = aitr.old_invoice_id""",
    )
```

obviously, this only makes sense if my PR of account module gets merged as it is now. that's why we, the people here in this repo, try to not make many PRs in advance, because some dependencies become key to proceed and move forward 😉

Yasaie: Is this because that table was merged and ids changed? 
Do I have to copy in old_invoice_id in account_invoice_transaction_rel and account_move tables in pre-migration first?

MiquelRForgeFlow: yes. and no need.

MiquelRForgeFlow (approved):
LGTM

addons/payment/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,54 @@
...
+# DONE: Pre-migration renamed to image_128
+
+payment      / payment.acquirer         / module_state (selection)      : is now stored
+payment      / payment.acquirer         / post_msg (html)               : DEL
+payment      / payment.acquirer         / specific_countries (boolean)  : DEL
+payment      / payment.acquirer         / website_published (boolean)   : DEL
+payment      / payment.acquirer         / error_msg (html)              : DEL
+payment      / payment.token            / company_id (many2one)         : NEW relation: res.company, isrelated: related, stored
+payment      / payment.transaction      / invoice_ids (many2many)       : relation is now 'account.move' ('account.invoice') [nothing to do]
+# NOTHING TO DO: many2many table didn't changed
```
kos94ok-3D: You updated link in post-migration so change comment

addons/payment/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+
+def update_account_invoice_transaction_rel_table(cr):
+    openupgrade.logged_query(
+        cr, """
+        UPDATE account_invoice_transaction_rel aitr
+        SET invoice_id = am.id
+        FROM account_move am
+        WHERE am.old_invoice_id = aitr.old_invoice_id
```
gdgellatly: @MiquelRForgeFlow You have approved this, but I get error so maybe I am doing something wrong.  THere appears to be no column `aitr.old_invoice_id` but rather just `aitr.invoice_id`. It would make sense looking at query but maybe not.

addons/payment/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,54 @@
...
+
+payment      / payment.acquirer         / auth_msg (html)               : NEW hasdefault
+payment      / payment.acquirer         / check_validity (boolean)      : NEW
+payment      / payment.acquirer         / color (integer)               : NEW isfunction: function, stored
+payment      / payment.acquirer         / display_as (char)             : NEW
+# NOTHING TO DO
+
+payment      / payment.acquirer         / environment (selection)       : DEL required, selection_keys: ['prod', 'test'], req_default: function
+payment      / payment.acquirer         / state (selection)             : NEW required, selection_keys: ['disabled', 'enabled', 'test'], req_default: function, hasdefault
+# DONE: Pre-migration renamed to state
```
kos94ok-3D: 1. move `payment      / payment.acquirer         / website_published (boolean)   : DEL` here
2. add comment like this:
Post-migration: 
  set `enabled` where environment == prod and website_published True
  set `disabled` where environment == prod and website_published False

addons/payment/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,54 @@
...
+payment      / payment.acquirer         / image_small (binary)          : DEL attachment: True
+# DONE: Pre-migration renamed to image_128
+
+payment      / payment.acquirer         / module_state (selection)      : is now stored
+payment      / payment.acquirer         / post_msg (html)               : DEL
+payment      / payment.acquirer         / specific_countries (boolean)  : DEL
+payment      / payment.acquirer         / website_published (boolean)   : DEL
+payment      / payment.acquirer         / error_msg (html)              : DEL
+payment      / payment.token            / company_id (many2one)         : NEW relation: res.company, isrelated: related, stored
+# NOTHING TO DO: many2many table didn't changed
```
kos94ok-3D: Does need it? `many2many table didn't changed`

kos94ok-3D (commented):
Small fixes in analysis_work

pedrobaeza:
I have added extra needed patches.

PR #2279, opened 2020-04-08
[13.0][FIX] utm: pre-migration 

astirpe:
This fixes an error I get when I try to re-run the migration scripts after something has gone wrong.

--

### DIFF
--- a/addons/utm/migrations/13.0.1.0/pre-migration.py
+++ b/addons/utm/migrations/13.0.1.0/pre-migration.py
@@ -48,18 +48,18 @@ def move_mailing_campaign_to_utm_campaign(env):
 @openupgrade.migrate()
 def migrate(env, version):
     cr = env.cr
+    openupgrade.rename_models(cr, _model_renames)
     if openupgrade.table_exists(cr, 'mail_mass_mailing_stage'):
-        openupgrade.rename_models(cr, _model_renames)
         openupgrade.rename_tables(cr, _table_renames)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not correct. If you are re-running something, it should be rolled-back. If not, then you should start from beginning again.

astirpe:
Ok but it saved a lot of time to me, especially in this early stage. Feel free to close this PR

pedrobaeza:
The only reading here is if the commit after migrating each module is working properly. @StefanRijnhart can it be incorrect, committing changes in migration scripts but not committing module installed version?

MiquelRForgeFlow:
This is wrong. @astirpe the problem you may have is that after a failed migration, the following migration is trying to migrate an already migrated module.

astirpe:
The error I got was because the script failed somewhere and the `rename_models` rolled-back but the `rename_tables` did not. Re-executing the script, the `rename_models` was not executed anymore because not any table with `mail_mass_mailing_stage` name was found: this caused an error.

According to you all, the only way to go on is to restart all over again since the rollback didn't work properly.
For me this PR worked fine and I saved some good time.

I realize you don't see any advantage in this change, so I close the PR and save extra time.

pedrobaeza:
@astirpe that's the weird thing: everything should be rolled-back. The commit level should be per module, and if everything is in the same module, everything should be rolled-back.

PR #2280, opened 2020-04-08
[13.0][MIG] project

Yasaie:
Migration of `project`
--


addons/project/migrations/13.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,40 @@
...
+project      / project.task             / notes (text)                  : DEL
+project      / project.task.type        / legend_priority (char)        : DEL
+# NOTHING TO DO
+
+---XML records in module 'project'---
+NEW ir.actions.act_window: project.portal_share_action
+NEW ir.model.access: project.access_account_analytic_account_manager [renamed from hr_timesheet module]
+NEW ir.model.access: project.access_account_analytic_account_user [renamed from hr_timesheet module]
+NEW ir.model.access: project.access_account_analytic_line_project [renamed from hr_timesheet module]
+# NOTHING TO DO: Handels in hr_timesheet
```
MiquelRForgeFlow: no, you have to handle the xmlid renames in this module.

addons/project/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+from openupgradelib import openupgrade
+
+
+def fill_project_task_company_id(cr):
+    openupgrade.logged_query(
+        cr, """
+        UPDATE project_task pt
+        SET company_id = c.id
+        FROM (SELECT id FROM res_company ORDER BY id LIMIT 1) AS c
+        WHERE pt.company_id IS NULL
```
MiquelRForgeFlow:
```suggestion
        SET company_id = ru.company_id
        FROM res_users ru
        WHERE ru.id = pt.create_uid AND pt.company_id IS NULL
```

addons/project/migrations/13.0.1.1/openupgrade_analysis_work.txt
Yasaie: it's `renamed from` so might be not exist. i guess they should rename when it's `renamed to`

gdgellatly: @Yasaie That is what @MiquelRForgeFlow is saying. You need to rename it pre-migration if it already exists. Otherwise it will create new ones. The `rename_xmlids` will take care if it doesn't exist.

addons/project/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,19 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_project_task_company_id(cr):
+    openupgrade.logged_query(
+        cr, """
```
gdgellatly: Hi,

Are we sure on this? To me this should be the last resort if all the other related fields with company ids are NULL.

company_id of project should be first choice, then assigned user, then assigned manager.

But also, I don't know the scope for adding notes/docs but telling people to set company correctly first for tasks pre-migration.

addons/project/migrations/13.0.1.1/openupgrade_analysis_work.txt
Yasaie: @gdgellatly I mean in case of `DEL` of xml we need to do actions like `rename_xmlids` otherwise if it's already renamed in `hr_timesheet` nothing to do unless it will create a new one.
I think this more correct way

addons/project/migrations/13.0.1.1/post-migration.py
Yasaie: You are right `project.project` is default `company_id` and I changed it.
But I have no idea about it should be in `pre-migration` or `post-migration`.
I thing fill should be in `pre-migration`

addons/project/migrations/13.0.1.1/openupgrade_analysis_work.txt
gdgellatly: @Yasaie I think you are misunderstanding what this means because it is obvious you must handle it here.

hr_timesheet depends on project. The xmlid has moved from hr_timesheet to project. If you do not rename it here pre-migration and hr_timesheet is installed you will duplicate.

addons/project/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: it's ok as it is now

MiquelRForgeFlow (approved):

addons/project/migrations/13.0.1.1/post-migration.py
kos94ok-3D: It has to be in pre-migration because in post-migration all tasks already have company_id.
And I don`t see the query for tasks without project.
<pre>

    openupgrade.logged_query(
        cr, """
        UPDATE project_task pt
        SET company_id = ru.company_id
        FROM res_users ru
        WHERE pt.project_id IS NULL AND pt.company_id IS NULL AND pt.create_uid = ru.id
        """
    )
</pre>

kos94ok-3D (commented):
Multi-company issues

kos94ok-3D (approved):

pedrobaeza:
Made some adjustments, so merging now.

kos94ok-3D:
@pedrobaeza What to do with task without project?

pedrobaeza:
Ouch, very good question... restoring through direct commit. That's the problem though for not putting code comments about the why.

pedrobaeza:
Restored in 4301baad445d50a71b45b0fd2926bdd686bd1fb8 with proper comment.

kos94ok-3D:
Thanks

pedrobaeza:
Thanks to you for being vigilant :smiley:

PR #2281, opened 2020-04-08
[13.0][MIG] event

Yasaie:
Migration of `event`
--


addons/event/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def fill_event_mail_notification_type(cr):
```
MiquelRForgeFlow: Delete this method, it will do nothing. During the update of the module (between the pre-migration and the post-migration), both notification_type columns will be created and automatically filled by their default value.

addons/event/migrations/13.0.1.0/noupdate_changes.xml
```diff
@@ -1,18 +1,11 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="event_event_company_rule" model="ir.rule">
-    <field name="domain_force">['|',
-                                            ('company_id', '=', False),
-                                            ('company_id', 'in', company_ids),
-                                        ]
+    <field name="domain_force">['|', ('company_id', '=', False), ('company_id', 'in', company_ids),]
```
MiquelRForgeFlow: don't do this kind of cosmetic change, please.

addons/event/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+        SET notification_type = 'mail'
+        WHERE notification_type IS NULL
+        """
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    fill_event_mail_notification_type(env.cr)
+    openupgrade.load_data(env.cr, 'event', 'migrations/13.0.1.0/noupdate_changes.xml')
```
MiquelRForgeFlow, line 9: you should delete the record translations of event_reminder and event_subscription templates.

MiquelRForgeFlow (changes requested):

addons/event/migrations/13.0.1.0/post-migration.py
Yasaie: does always odoo fill default values of new fields?

Yasaie:
@MiquelRForgeFlow Thanks for reviews. changes made

addons/event/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, 'event', 'migrations/13.0.1.0/noupdate_changes.xml')
+    openupgrade.delete_record_translations(
+        env.cr, 'event', [
+            'event_reminder',
+            'event_subscription',
+        ],
+    )
```
MiquelRForgeFlow:
```suggestion
    )

```

addons/event/migrations/13.0.1.0/post-migration.py
MiquelRForgeFlow: yes. if a field has default, and the field is new, then odoo fill the new column with the default.

MiquelRForgeFlow (approved):

kos94ok-3D (approved):

Abranes (approved):

StefanRijnhart:
Merged manually because ocabot is blocked by Github actions failing on coverage push

PR #2284, opened 2020-04-08
[13.0][MIG] gamification

Yasaie:
Migration of `gamification`
--


addons/gamification/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
...
+
+_unlink_by_xmlid = [
+    'gamification.mt_badge_granted',
+]
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(env, _unlink_by_xmlid)
+    openupgrade.load_data(env.cr, 'gamification', 'migrations/13.0.1.0/noupdate_changes.xml')
```
MiquelRForgeFlow, line 28: detete record translations of email_template_badge_received and simple_report_template templates

addons/gamification/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+
+_field_renames = [
+    ('gamification.badge', 'gamification_badge', 'image', 'image_1920'),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _field_renames)
+    # Fix image of gamification.badge after renaming column to image_1920
```
MiquelRForgeFlow: better do it in post-migration using orm like done in product migration.

Yasaie: I just used the way that mail migration was:
```python
# addons/mail/migrations/13.0.1.0/pre-migration.py

_field_renames = [
    ('mail.channel', 'mail_channel', 'image', 'image_128'),
    ('mail.message', 'mail_message', 'layout', 'email_layout_xmlid'),
    ('mail.message', 'mail_message', 'needaction_partner_ids', 'notified_partner_ids'),
    ('mail.notification', 'mail_message_res_partner_needaction_rel', 'email_status', 'notification_status'),
    ('mail.tracking.value', 'mail_tracking_value', 'track_sequence', 'tracking_sequence'),
]


@openupgrade.migrate()
def migrate(env, version):
    openupgrade.rename_models(env.cr, _model_renames)
    # Fix image of mail.channel after renaming column to image_128
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE ir_attachment
        SET res_field = 'image_128'
        WHERE res_field = 'image' and res_model = 'mail.channel'
        """,
    )
```

I this wrong?

MiquelRForgeFlow: mail.channel doesn't inherit from the image.mixin, that's why not needed

only models which inherit from image.mixin need the image rename in post-migration

MiquelRForgeFlow (approved):

StefanRijnhart (approved):

PR #2285, opened 2020-04-08
[13.0][MIG] hr

Yasaie:
Migration of `hr`
--


addons/hr/migrations/13.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,90 @@
...
+NEW hr.plan.activity.type: hr.onboarding_training (noupdate)
+NEW ir.actions.act_window: hr.hr_departure_wizard_action
+NEW ir.actions.act_window: hr.hr_employee_action_from_user
+NEW ir.actions.act_window: hr.hr_employee_public_action
+NEW ir.actions.act_window: hr.hr_plan_action
+NEW ir.actions.act_window: hr.hr_plan_activity_type_action
+NEW ir.actions.act_window: hr.plan_wizard_action
+NEW ir.actions.act_window: hr.res_users_action_my
+DEL ir.actions.act_window: hr.hr_employee_action_subordinate_hierachy
+NEW ir.actions.report: hr.hr_employee_print_badge [renamed from hr_attendance module]
```
MiquelRForgeFlow, line 63: you should handle in this module the xmlid rename

addons/hr/migrations/13.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+
+_field_renames = [
+    ('hr.employee', 'hr_employee', 'image', 'image_1920'),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _field_renames)
+    # Fix image of hr.employee after renaming column to image_1920
```
MiquelRForgeFlow: do it by orm in post-migration because is image.mixin

addons/hr/migrations/13.0.1.1/openupgrade_analysis_work.txt
Yasaie: It's `renamed from` it could be not exist. Isn't it better done in `renamed to` side?

MiquelRForgeFlow: if not exist, no problem, the xmlid rename don't break in that case

the rename always is done in destiny module. so if the rename is hr_attendance -> hr, then corresponds to hr module have he xmlid rename script.

Yasaie: Thanks, done

addons/hr/migrations/13.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020 Payam Yasaie <https://www.tashilgostar.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_xmlid_renames = [
+    ('hr.hr_employee_print_badge', 'hr.hr_employee_print_badge')
```
MiquelRForgeFlow:
```suggestion
    ('hr_attendance.hr_employee_print_badge', 'hr.hr_employee_print_badge')
```

MiquelRForgeFlow (approved):

pedrobaeza:
I have optimized the migration scripts with 2 patches.

PR #2290, opened 2020-04-11
[13.0][ADD] hr_recruitment

Yasaie:
Migration of `hr_recruitment`
--


addons/hr_recruitment/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,27 @@
+---Models in module 'hr_recruitment'---
+---Fields in module 'hr_recruitment'---
+hr_recruitment / hr.applicant             / email_cc (text)               : type is now 'char' ('text')
```
MiquelRForgeFlow, line 3: in pre-migration:

```python
openupgrade.logged_query(
        env.cr, """
        ALTER TABLE hr_applicant
        ALTER COLUMN email_cc
        TYPE varchar"""
    )
```

although maybe not needed, I have not tested if it is done automatically without losing data

addons/hr_recruitment/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+        'job_ids', openupgrade.get_legacy_name('job_id')
+    )
+    openupgrade.load_data(env.cr, 'hr_recruitment', 'migrations/13.0.1.0/noupdate_changes.xml')
+    openupgrade.delete_record_translations(
+        env.cr, 'hr_recruitment', [
+            'email_template_data_applicant_congratulations',
+            'email_template_data_applicant_interest',
+            'email_template_data_applicant_refuse',
+        ],
+    )
```
MiquelRForgeFlow:
```suggestion
    )

```

addons/hr_recruitment/migrations/13.0.1.0/openupgrade_analysis_work.txt
Yasaie: I added the code in case if didn't do it automatic

MiquelRForgeFlow (approved):

addons/hr_recruitment/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    ],
+}
+
+
+def alter_hr_applicant_email_cc_type(cr):
+    openupgrade.logged_query(
+        cr, """
+        ALTER TABLE hr_applicant
+        ALTER COLUMN email_cc
+        TYPE varchar
```
StefanRijnhart: You can see here that casting Text fields to Char is supported by Odoo natively
https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L1546-L1547

pedrobaeza: Removed and documentation updated

pedrobaeza (approved):

PR #2291, opened 2020-04-11
[13.0][MIG] hr_timesheet

Yasaie:
Migration of `hr_timesheet`
--


addons/hr_timesheet/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,33 @@
...
+---XML records in module 'hr_timesheet'---
+DEL ir.actions.act_window: hr_timesheet.project_task_action_view_timesheet
+NEW ir.actions.act_window.view: hr_timesheet.timesheet_action_view_report_by_employee_graph
+NEW ir.actions.act_window.view: hr_timesheet.timesheet_action_view_report_by_project_graph
+NEW ir.actions.act_window.view: hr_timesheet.timesheet_action_view_report_by_task_graph
+# NOTHING TO DO
+DEL ir.model.access: hr_timesheet.access_account_analytic_account_manager [renamed to project module]
+DEL ir.model.access: hr_timesheet.access_account_analytic_account_user [renamed to project module]
+DEL ir.model.access: hr_timesheet.access_account_analytic_line_project [renamed to project module]
+# DONE: pre-migration renamed
```
MiquelRForgeFlow: No, this should be done in `project` migration

MiquelRForgeFlow (approved):

gdgellatly:
```
import odoo
from odoo.modules.db import create_categories

i = 'hr_timesheet'
info = odoo.modules.load_information_from_description_file(i)
categories = info['category'].split('/')
create_categories(env.cr, categories)
```

I had to run this in shell to make this module migrate propoerly afterwards

pedrobaeza:
@gdgellatly I would say that your problem was that the initial category gathering was not done in `base` module or you didn't have latest code.

gdgellatly:
@pedrobaeza Well if that is the case I am definitely not the only one as since reporting I have had a few private mails asking for assistance in running this particular script.

pedrobaeza:
But I don't get what is the problem you are having. A crash?

gdgellatly:
@pedrobaeza I think it might happen only in multi version migrations, because that is where it is affecting us and everyone who has asked. In my particular case it was 11 to 13. 

It occurs when doing a -u all after migration and back on main odoo code. Its like a category is on no update or doesn't get created. It was 2 months ago, I don't remember too well the exact issue, and as we just do that in our end-migration script now,  I never see the error again.

pedrobaeza:
OK, then I prefer to have specific steps to reproduce it and the exact error via an issue for analyzing it. Let's keep it as is for now.

PR #2292, opened 2020-04-11
[13.0][MIG] l10n_generic_coa

Yasaie:
Migration of `l10n_generic_coa`
--


addons/l10n_generic_coa/migrations/13.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,35 @@
...
+    ('l10n_generic_coa.a_capital', 'l10n_generic_coa.capital'),
+    ('l10n_generic_coa.a_dividends', 'l10n_generic_coa.dividends'),
+    ('l10n_generic_coa.a_expense_finance', 'l10n_generic_coa.expense_finance'),
+    ('l10n_generic_coa.a_expense_invest', 'l10n_generic_coa.expense_invest'),
+    ('l10n_generic_coa.a_salary_expense', 'l10n_generic_coa.expense_salary'),
+    ('l10n_generic_coa.cas', 'l10n_generic_coa.non_current_liabilities'),
+    ('l10n_generic_coa.conf_a_expense', 'l10n_generic_coa.expense'),
+    ('l10n_generic_coa.conf_a_pay', 'l10n_generic_coa.payable'),
+    ('l10n_generic_coa.conf_a_recv', 'l10n_generic_coa.receivable'),
+    ('l10n_generic_coa.conf_a_reserve_and_surplus', 'l10n_generic_coa.tax_payable'),
```
MiquelRForgeFlow, line 15: are you sure about this one? I don't know

Yasaie: I checked those and were the same just codes changed.

MiquelRForgeFlow: Ok, nice.

MiquelRForgeFlow (approved):

kos94ok-3D:
missing 
<pre>
openupgrade.set_xml_ids_noupdate_value(
        env,
        "l10n_generic_coa",
        [
            "configurable_chart_template",
            "purchase_tax_template",
            "sale_tax_template",
        ],
        False,
    )
</pre>

addons/l10n_generic_coa/migrations/13.0.1.1/pre-migration.py
kos94ok-3D: Hi, @Yasaie 
I did not found where this account used in Odoo, so I`m not sure that it the same.
I think someone could use this account in the case which not related to tax_payable.
If you do not provide an example where Odoo12 use it, I think better remove this line.

StefanRijnhart (approved):
Thanks! I see you have also fixed @kos94ok-3D's comment, but you did not post a comment about the update so he did not see it. But as such, ready for merge.

PR #2293, opened 2020-04-11
[13.0][MIG] website_mass_mailing

Yasaie:
Migration of `website_mass_mailing`
--


addons/website_mass_mailing/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+_column_copies = {
+    'mailing_list': [
+        ('popup_content', None, None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, _column_copies)
```
MiquelRForgeFlow: I think you can do rename_columns instead.

StefanRijnhart: @Yasaie please have a look at this comment ^

StefanRijnhart:
Please rebase to have Github run the CI instead of Travis.

addons/website_mass_mailing/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    WHERE {popup_content} IS NOT NULL
+    """.format(popup_content=openupgrade.get_legacy_name('popup_content'))
+    env.cr.execute(sql)
+    rows = env.cr.fetchall()
+
+    for row in rows:
+        popup.create({
+            'mailing_list_id': row[0],
+            'popup_content': row[1]
+        })
```
MiquelRForgeFlow: Also it better a create in batch:

```suggestion
    to_create = []
    for row in rows:
        to_create.append({
            'mailing_list_id': row[0],
            'popup_content': row[1]
        })
    if to_create:
        popup.create(to_create)
```

addons/website_mass_mailing/migrations/13.0.1.0/pre-migration.py
MiquelRForgeFlow: I don't remember why I said this. Maybe I was wrong here.

EDIT: Yeah, I remember. Because `popup_content` column doesn't exist anymore in that model.

pedrobaeza: This is not needed indeed.

pedrobaeza:
Pushed a new version with one SQL to get the popup in original language in one shot, and also handle translations, that are important as well. Merging if all OK on CIs.

PR #2305, opened 2020-04-15
[13.0][MIG] survey

Yasaie:
## Migration of `survey`

> Depends on `gamification`

--


gdgellatly:
Receive the following error migrating a database with only mis-builder as third party module.
```
psycopg2.IntegrityError: duplicate key value violates unique constraint "ir_model_fields_name_unique"
DETAIL:  Key (model, name)=(survey.question, survey_id) already exists.
```

addons/survey/migrations/13.0.3.0/pre-migration.py
```diff
@@ -0,0 +1,95 @@
...
+    ('survey.email_template_survey', 'survey.mail_template_user_input_invite'),
+]
+
+
+def move_survey_page_to_survey_question(env):
+    openupgrade.add_fields(env, _field_adds)
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE survey_question sq
+        SET title = sp.title AND survey_id = sp.survey_id AND is_page = TRUE
```
gdgellatly: incorrect syntax use `,` not `AND` to avoid

```
psycopg2.ProgrammingError: argument of AND must be type boolean, not type character varying
LINE 3:         SET title = sp.title AND survey_id = sp.survey_id AN...
```

addons/survey/migrations/13.0.3.0/pre-migration.py
```diff
@@ -0,0 +1,95 @@
...
+        FROM survey_page sp
+        WHERE sq.page_id = sp.id
+        """
+    )
+
+
+def update_survey_question_page_id(cr):
+    replaces = [
+        ('survey_user_input', 'last_displayed_page_id'),
+        ('survey_user_input_line', 'page_id'),
```
gdgellatly: I think this field no longer exists, so causes error on update. Not sure, database I have has no records for this table.

Yasaie: You were right. it's a related field now, thanks

MiquelRForgeFlow (approved):

gdgellatly:
@MiquelRForgeFlow It might pay to check. Just testing a migration and the field access_token is being set on surveys as it is a new field, however all surveys are receiving an identical access token, which is then causing constraints not to be applied and singleton errors on accessing via respondent link.

PR #2301, opened 2020-04-13
[13.0][MIG] stock

Yasaie:
Migration of `stock`
--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2306, opened 2020-04-15
[13.0][MIG] stock

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2301.

--


Yasaie:
I have a question I like to know, what is `end-migration.py` and when to use it?

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,322 @@
...
+    openupgrade.logged_query(
+        cr, """
+        UPDATE stock_putaway_rule spr
+        SET company_id = ru.company_id
+        FROM res_users ru
+        WHERE ru.id = spr.create_uid AND spr.company_id IS NULL"""
+    )
+    # stock.production.lot
+    openupgrade.logged_query(
+        cr, """
```
gdgellatly, line 76: I really don't like this from a multicompany perspective. I think it needs more thought. Something more like

`SELECT company_id from stock_move_line sml WHERE sml.lot_id = spl.id and sml.company_id IS NOT NULL LIMIT 1` 

but even then you are will have cases where stock has transferred between companies, so you are going to need to do more advanced things like split the lot in two and rewrite some records.

MiquelRForgeFlow: not sure, neither agree nor disagree

MiquelRForgeFlow:
> I have a question I like to know, what is end-migration.py and when to use it?

the end-migration scripts, as their name suggests, are executed after all post-migration scripts of all modules have been executed.

and they are used when you have something that should be done in post-migration but can't be done in post-migration because other following modules need to use something about related. Just corner cases.

AaronHForgeFlow (approved):
:+1: tested in real database. No issues found in the result.

Only small thing, old production location will remain with old quants. and multiple production locations are created for companies. But as long the stock there does not count for the quantities available it is ok to to be like that.

gdgellatly:
I just tested on a real database where initial stock was initially imported by admin user in 9 companies. It is making all lots wrong company. In fact the company it chose doesn't even have a single stock move.

MiquelRForgeFlow:
@gdgellatly As you suggested, the lots now check the company of the move lines. If this doesn't work or is not enough, then please suggest me another way to find the correct company for the lots.

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,329 @@
...
+        WHERE spl.package_id = sqp.id AND spl.company_id IS NULL"""
+    )
+    # stock.scrap
+    openupgrade.logged_query(
+        cr, """
+        UPDATE stock_scrap ss
+        SET company_id = ru.company_id
+        FROM res_users ru
+        WHERE ru.id = ss.create_uid AND ss.company_id IS NULL"""
+    )
```
kos94ok-3D:
```
    openupgrade.logged_query(
        cr, """
        UPDATE stock_scrap ss
        SET company_id = ru.company_id
        FROM stock_move sm
        WHERE sm.id = ss.move_id AND ss.company_id IS NULL"""
    )
```

kos94ok-3D (changes requested):
The Scrap order and related stock move have to have the same company

gdgellatly:
> @gdgellatly As you suggested, the lots now check the company of the move lines. If this doesn't work or is not enough, then please suggest me another way to find the correct company for the lots.

@MiquelRForgeFlow Thanks. I checked for this database, it is only 250 or so lots of 120000 that have been used multicompany so we can split manually. Besides the nature of lots transferred between companies is that they are no longer available in at least one of those companies.

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,339 @@
...
+    # although later the 'stock.sequence_stock_scrap' sequence will be deleted,
+    # we need to nullify its code (if having one) here
+    # because we want a new autogenerated sequence
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_sequence seq
+        SET code = NULL
+        FROM ir_model_data imd
+        WHERE imd.res_id = seq.id AND imd.module = 'stock'
+            AND imd.model = 'sequence_stock_scrap'"""
```
kos94ok-3D: AND imd.name = 'sequence_stock_scrap'

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,339 @@
...
+            imd.name IN ('location_inventory', 'location_production') AND
+             ip.value_reference = 'stock.location,' || sl.id"""
+    )
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE stock_location sl
+        SET company_id = NULL
+        FROM ir_model_data imd
+        WHERE imd.res_id = sl.id AND imd.module = 'stock'
+            AND imd.model = 'stock_location_scrapped'"""
```
kos94ok-3D: AND imd.name = 'stock_location_scrapped'

kos94ok-3D (changes requested):
wrong conditional (used 'model' instead of 'name')

kos94ok-3D (approved):

addons/stock/migrations/13.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,66 @@
...
+    ],
+}
+
+_model_renames = [
+    ('report.stock.forecast', 'report.stock.quantity'),
+    ('stock.fixed.putaway.strat', 'stock.putaway.rule'),
+]
+
+_table_renames = [
+    ('stock_fixed_putaway_strat', 'stock_putaway_rule'),
```
kos94ok-3D, line 18: I think it is a bad idea to rename the table because:
 1. you have empty column location_in_id
 2. not all rows
Example:
 You have two locations with the same putaway so after migration, you have only one record in table stock_putaway_rule instead of two records

kos94ok-3D (commented):
If you don`t have time to fix it I can create PR in your fork

addons/stock/migrations/13.0.1.1/pre-migration.py
MiquelRForgeFlow: and what do you propose instead?

kos94ok-3D: you can join stock_fixed_putaway_strat and stock_location by putway and insert in stock_putaway_rule

MiquelRForgeFlow: done

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def fill_stock_putaway_rule_location_in_id(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE stock_putaway_rule spr
+        SET location_in_id = sl.id
+        FROM stock_location sl
+        JOIN product_putaway pp ON sl.putaway_strategy_id = pp.id
+        WHERE spr.putaway_id = pp.id AND spr.location_in_id IS NULL"""
```
kos94ok-3D: <pre>
FROM stock_location sl
  WHERE spr.putaway_id = sl.putaway_strategy_id AND spr.location_in_id IS NULL
</pre>

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,364 @@
...
+        INSERT INTO stock_putaway_rule (product_id, category_id,
+            location_in_id, location_out_id, sequence, company_id,
+            create_uid, create_date, write_uid, write_date)
+        SELECT spr.product_id, spr.category_id, sl.id, spr.location_out_id,
+            spr.sequence, spr.company_id, spr.create_uid, spr.create_date,
+            spr.write_uid, spr.write_date
+        FROM stock_putaway_rule spr
+        JOIN product_putaway pp ON spr.putaway_id = pp.id
+        JOIN stock_location sl ON sl.putaway_strategy_id = pp.id
+        WHERE spr.location_in_id != sl.id"""
```
kos94ok-3D: <pre>
FROM stock_putaway_rule spr
  JOIN stock_location sl
    ON sl.putaway_strategy_id = spr.putaway_id and spr.location_in_id != sl.id
</pre>

kos94ok-3D (changes requested):
Performance improve

kos94ok-3D (approved):

dennybiasiolli (approved):

SolvoSupport:
Any news about this pull? it seens ready to merge, I tested in multi company, multi warehouse env and it's seens to migrate ok

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,363 @@
...
+    openupgrade.m2o_to_x2m(
+        env.cr,
+        env['stock.inventory'], 'stock_inventory',
+        'location_ids', openupgrade.get_legacy_name('location_id')
+    )
+    openupgrade.m2o_to_x2m(
+        env.cr,
+        env['stock.inventory'], 'stock_inventory',
+        'product_ids', openupgrade.get_legacy_name('product_id')
+    )
```
kos94ok-3D, line 234: Check that stock_inventory_preparation_filter module is installed and convert `category_id`, `lot_id` fields
Related to OCA/stock-logistics-warehouse#959

MiquelRForgeFlow: done

kos94ok-3D:
@MiquelRForgeFlow  and @pedrobaeza What do you think about it https://github.com/kos94ok-3D/OpenUpgrade/commit/1546d73b684df7ed68a0d22082f2e25e5ec884c3#?

MiquelRForgeFlow:
@kos94ok-3D is it not handled already? what is your code adding new? I don't understand.

kos94ok-3D:
We do not use multi-company so we enough update exist locations instead of creating new (location_inventory, location_production, stock_location_scrapped)

Abranes (approved):

StefanRijnhart:
Love the collaboration on this PR.

@AaronHForgeFlow's comment 'multiple production locations are created for companies' sounds disconcerting but it looks like that may have been solved in the mean time.

One of my setups will suffer from the multi-company lot issue, but in our case it will probably make sense to add some customizations to enable multi-company lots explicitly, lifting the company ir.rule and changing the constraint to enforce uniqueness across companies. Would that make sense to you @gdgellatly?

@kos94ok-3D: that single-company shortcut, I can imagine it is a lot quicker than updating all the location ids throughout the database. Do you think it can be generalized, assigning the old locations to the company with the lowest id, and then excluding that company_id when assigning the new location ids? It might save some time and disk space even on multi company databases. Maybe we can do it after merging this one?

StefanRijnhart (approved):
Excellent work!

kos94ok-3D:
Hi, @StefanRijnhart .
I`m not sure that it will be correct for multi-company because an old location can have stock.move for more than one company, 
so I think we can get an access error.

StefanRijnhart:
> an old location can have stock.move for more than one company

Yes, but we will still be passing through https://github.com/OCA/OpenUpgrade/pull/2306/files#diff-23ca17a5c4c2204907213a7d5651c694fb2c55deb5e58a45cab51270acd76ee4R339 to rewrite the location ids, but excluding that company with the lowest id because that company was assigned to the existing locations. Does that make sense?

StefanRijnhart:
But let's get this one merged first! @MiquelRForgeFlow, are we ready?

addons/stock/migrations/13.0.1.1/end-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 ForgeFlow <https://www.forgeflow.com>
+# Copyright 2020 Tecnativa - Pedro M. Baezaw
```
pankk:
```suggestion
# Copyright 2020 Tecnativa - Pedro M. Baeza
```
I was just admiring this work when I saw this.

addons/stock/migrations/13.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,79 @@
...
+
+_xmlid_renames = [
+    # mail.template
+    ('delivery.mail_template_data_delivery_confirmation', 'stock.mail_template_data_delivery_confirmation'),
+    # ir.model.access
+    ('stock.access_stock_fixed_putaway_strat', 'stock.access_stock_putaway_manager'),
+    ('stock.access_stock_fixed_putaway_user', 'stock.access_stock_putaway_all'),
+    ('stock.access_stock_forecast_user', 'stock.access_report_stock_quantity'),
+    # ir.actions.report
+    ('stock_zebra.label_barcode_product_product', 'stock.label_barcode_product_product'),
```
pedrobaeza: These lines are not needed, as the module merge already performs this:

https://github.com/OCA/OpenUpgrade/blob/80a6e24d63ea80c61746d11b4124a5217aee3ccc/odoo/addons/openupgrade_records/lib/apriori.py#L30

addons/stock/migrations/13.0.1.1/end-migration.py
MiquelRForgeFlow: done

kos94ok-3D:
any news about ready for merge?

StefanRijnhart:
All comments honoured, so let's go.

/ocabot merge patch

MiquelRForgeFlow:
?

StefanRijnhart:
Does ocabot not work here, or would it have worked with nobump?

pedrobaeza:
I think the problem was that the merge was failed, but there was no feedback of that, and now the CI is red. I haven't had time for checking the underlying problem.

PR #2310, opened 2020-04-22
[13.0][MIG] purchase

MiquelRForgeFlow:
Migration of `purchase`.

--


addons/purchase/migrations/13.0.1.2/post-migration.py
```diff
@@ -0,0 +1,72 @@
...
+        openupgrade.logged_query(
+            cr, """
+            UPDATE purchase_order_line pol
+            SET qty_received_method = 'stock_moves'
+            FROM product_product pp
+            LEFT JOIN product_template pt ON pp.product_tmpl_id = pt.id
+            WHERE pol.product_id = pp.id AND pol.display_type IS NULL
+                AND pt.type IN ('consu', 'product')
+            """
+        )
```
kos94ok-3D: <pre>SET qty_received_method = 'stock_moves', qty_received_manual = 0.0</pre>

kos94ok-3D (commented):
I think need to keep a behave written in the `_inverse_qty_received` function
изображение

addons/purchase/migrations/13.0.1.2/post-migration.py
MiquelRForgeFlow: I set it to NULL instead, ok?

kos94ok-3D (approved):

addons/purchase/migrations/13.0.1.2/post-migration.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def move_fields_from_invoice_to_moves(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_move_line aml
+        SET purchase_line_id = ail.purchase_line_id
+        FROM account_invoice_line ail
+        WHERE aml.old_invoice_line_id = ail.id"""
+    )
```
legalsylvain, line 14: I think adding ``and ail.purchase_line_id is not null`` could make the request faster, avoiding false in many cases.

addons/purchase/migrations/13.0.1.2/post-migration.py
```diff
@@ -0,0 +1,70 @@
...
+        JOIN account_move am ON am.old_invoice_id = ai.id
+        WHERE ampor.{} = ai.id
+        """.format(openupgrade.get_legacy_name('account_invoice_id'))
+    )
+
+
+def fill_purchase_order_line_qty_received_method(cr):
+    if openupgrade.column_exists(cr, 'purchase_order', 'picking_type_id'):
+        # purchase_stock is installed
+        # set qty_delivered_method = 'stock_moves'
```
legalsylvain: There is a function to test if a module is installed.

legalsylvain (commented):
Some minor remarks. Lgtm otherwise. Thanks !

addons/purchase/migrations/13.0.1.2/post-migration.py
MiquelRForgeFlow: Is it better that function than using that column_exists?

kos94ok-3D: It is not always possible, because the module can be merged into another

Abranes (approved):

addons/purchase/migrations/13.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,45 @@
...
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, _column_copies)
+    openupgrade.rename_columns(env.cr, _column_renames)
+    openupgrade.rename_tables(env.cr, _table_renames)
+    openupgrade.rename_xmlids(env.cr, _xmlid_renames)
+    openupgrade.logged_query(
+        env.cr, """
+        ALTER TABLE purchase_order_line
+        ADD COLUMN qty_received_method varchar""",
+    )
```
ThomasBinsfeld, line 34: `rename_columns` and `rename_tables` preserve the `NOT NULL` on the legacy `account_invoice_id` column. That can cause some problems. I think the `NOT NULL` should be dropped here.

```suggestion
    )
    openupgrade.logged_query(
        env.cr, """
        ALTER TABLE account_move_purchase_order_rel
        ALTER COLUMN openupgrade_legacy_13_0_account_invoice_id
        DROP NOT NULL""",
    )
```

MiquelRForgeFlow: Now rename is avoided.

bishalgit:
@MiquelRForgeFlow  Should I review this one? Why is this still not merged?

abdounasser202:
I think the migration of this module blocks the one from sale_purchase

MiquelRForgeFlow:
> Should I review this one?

Reviews are always welcomed.

> Why is this still not merged?

Because @pedrobaeza still has not reviewed it.

pedrobaeza:
Well, this one is heavy and some experienced reviewer is a good choice for not merging something not completed. Anyway, right now the CI is red, so it must be solved before merging. I will review this meanwhile. @MiquelRForgeFlow please check.

MiquelRForgeFlow:
@pedrobaeza The error in CI is due to purchase_stock, which should not be taken into account, as it will follow this PR.

kos94ok-3D:
@MiquelRForgeFlow  Runbot is red so maybe better move code from https://github.com/OCA/OpenUpgrade/pull/2314 here.
@pedrobaeza Do you plan to review it until summer?

pedrobaeza:
I have a local branch with the needed changes.

pedrobaeza (approved):
I have reviewed and performed latest fine-tune

PR #2311, opened 2020-04-22
[FIX] website: handle dangling root menus

yajo:
We could find in a DB a dangling root menu that was not being used.

After migration, that menu became website's menu, breaking menu layout.

To avoid that problem, make sure that any website-specific-but-unused root menu is transformed to website-agnostic, so the main website menu selected by `env['website']._compute_menu()` is definitely the menu we create later in this function.

@Tecnativa TT23336


pedrobaeza:
Migration script is not correct.

pedrobaeza (approved):

PR #2313, opened 2020-04-23
[ADD] rename account_bank_statement_import_camt to the OCA version

hbrunn:
I think it's safe to assume that an OpenUpgrade user will want to use the [OCA version](https://github.com/OCA/bank-statement-import/tree/11.0/account_bank_statement_import_camt_oca) of this addon.


pedrobaeza (approved):

StefanRijnhart:
Just curious @hbrunn, did you put this in `merged_modules` on purpose (instead of in `renamed_modules`)? If so, what's the benefit?

hbrunn:
as I don't remember anything specific, I assume that was a mistake

PR #2318, opened 2020-04-24
[FIX] account: migration 8->9, wrong sql to determine tax in account.invoice.tax

jonerikceberio:
 [FIX] account: migration 8->9, the sql to determine the tax id for account.invoice.tax, in the case of several taxes found for the combination of (base_code_id, tax_code_id) is missing additional parenthesis, as everything after the OR is grouped together.

Current behavior before PR:
The result is that several taxes are found, and therefore are not updated / migrated
Only one tax is found, and therefore is updated in the account.invoice.tax

--


ivantodorovich (approved):
Looks good

pedrobaeza (approved):

PR #2319, opened 2020-04-24
[13.0][MIG] snailmail

MiquelRForgeFlow:
Migration of `snailmail`.

--


kos94ok-3D (approved):

addons/snailmail/migrations/13.0.0.1/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+_column_copies = {
+    'snailmail_letter': [
+        ('state', None, None),
+    ],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, _column_copies)
```
bishalgit, line 14: Can you please explain me whey we do pre-migration for state here if there no change in field? I have also seen in other modules' migration the same process. I am new maybe I am missing something.

bishalgit (approved):
I approve this PR.

addons/snailmail/migrations/13.0.0.1/pre-migration.py
MiquelRForgeFlow: because later the copied column is used in the post-migration script. the 'draft' option has disappeared from state field.

PR #2321, opened 2020-04-24
[13.0][MIG] im_livechat

MiquelRForgeFlow:
Migration of `im_livechat`.

--


addons/im_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,24 @@
...
+im_livechat  / im_livechat.channel      / rating_ids (one2many)         : NEW relation: rating.rating
+im_livechat  / mail.channel             / country_id (many2one)         : NEW relation: res.country
+im_livechat  / res.users                / livechat_username (char)      : NEW
+# NOTHING TO DO (tentative)
+
+im_livechat  / mail.channel             / livechat_operator_id (many2one): NEW relation: res.partner
+# DONE: pre-migration: filled with the user_ids of im_livechat.channel
+
+---XML records in module 'im_livechat'---
+NEW im_livechat.channel: im_livechat.im_livechat_channel_data
```
kos94ok-3D, line 18: renamed from `website_livechat.im_livechat_channel_data_website`
https://github.com/odoo/odoo/commit/318893d6ed941f81bbbea230a087a3574cc97dcb

kos94ok-3D (changes requested):

kos94ok-3D (approved):

addons/im_livechat/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+_field_renames = [
+    ('im_livechat.channel', 'im_livechat_channel',
```
pedrobaeza, line 7: You need to do more than renaming the field, as this is an attachment field, so there will be attachments with the incorrect `res_field`, so you will lose all the existing ones.

pedrobaeza (changes requested):

addons/im_livechat/migrations/13.0.1.0/pre-migration.py
MiquelRForgeFlow: but I think that is done in the method below :S

pedrobaeza: You are right, sorry. I didn't remember we included such condition.

pedrobaeza (approved):

PR #2302, opened 2020-04-13
[13.0][MIG] point_of_sale

Yasaie:
Migration of `point_of_sale`
--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2151, opened 2020-01-03
[10.0][PERF] Perf improvement when migrating stock module

sebastienbeau:
Description of the issue/feature this PR addresses:

Adding related field with the ORM is long

Current behavior before PR:

On my database adding the fields stock_move.scrapped and stock_production_lot.product_uom_id
take 4 hours. Around 3.5 million of stock move and 1.3 million of stock production lot

Desired behavior after PR is merged:

It only take 30 min

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2326, opened 2020-04-27
[13.0][MIG] point_of_sale

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2302.

--


addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,185 @@
...
+        vals_list = []
+        for journal in env['account.journal'].browse(journal_ids):
+            vals = {
+                'name': 'Cash',
+                'cash_journal_id': journal.id,
+                'is_cash_count': True,
+                'company_id': journal.company_id.id,
+            }
+            vals_list += [vals]
+        env['pos.payment.method'].create(vals_list)
```
sbidoul: I have an issue here because `pos.payment.method.receivable_account_id` is required and it's default value comes from a new field `res.company.account_default_pos_receivable_account_id`.

addons/point_of_sale/migrations/13.0.1.0.1/noupdate_changes.xml
```diff
@@ -1,19 +1,19 @@
...
 <odoo>
-  <record id="pos_config_main" model="pos.config">
+<!--  <record id="pos_config_main" model="pos.config">
     <field name="picking_type_id"/>
-  </record>
+  </record>-->
   <record id="product_product_consumable" model="product.product">
     <field name="default_code">DISC</field>
-    <field name="image" type="base64" file="point_of_sale/static/img/product_product_49-image.jpg"/>
+    <field name="image_1920" type="base64" file="point_of_sale/static/img/product_product_49-image.jpg"/>
```
sbidoul: Is this image file missing from this PR?

MiquelRForgeFlow: deleted

addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
MiquelRForgeFlow: moved method to end-migration because `res.company.account_default_pos_receivable_account_id` is filled in end-migration of `account`.

pedrobaeza: Please check this that I have fixed all of this:

https://github.com/Tecnativa/OpenUpgrade/commit/146a30fe48197ac0c190039efdf03cc471df5245
https://github.com/Tecnativa/OpenUpgrade/commit/238da181379def6b8f8f12aa0932aa16f248b027

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
```diff
@@ -0,0 +1,91 @@
...
+    env.cr.execute("""
+        SELECT DISTINCT journal_id
+        FROM pos_config_journal_rel
+    """)
+    journal_ids = [aj[0] for aj in env.cr.fetchall()]
+    if journal_ids:
+        vals_list = []
+        for journal in env['account.journal'].browse(journal_ids):
+            vals = {
+                'name': 'Cash',
```
legalsylvain: name seems hard coded. Should not be ``journal.name`` ?

addons/point_of_sale/migrations/13.0.1.0.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,180 @@
...
+# DONE: post-migration: filled empty cases
+
+point_of_sale / pos.config               / module_pos_hr (boolean)       : NEW
+# DONE: post-migration: marked as True if hr is installed, because the new pos_hr module has auto_install = True
+
+point_of_sale / pos.config               / group_by (boolean)            : DEL
+point_of_sale / pos.config               / journal_ids (many2many)       : DEL relation: account.journal
+point_of_sale / pos.config               / payment_method_ids (many2many): NEW relation: pos.payment.method, hasdefault
+point_of_sale / account.journal          / pos_payment_method_ids (one2many): NEW relation: pos.payment.method
+# DONE: end-migration: for each journal in journal_ids, created a pos.payment.method record with cash_journal_id pointing the corresponding journal
```
legalsylvain, line 47: why do you make ``pos.payment`` and ``pos.payment.method`` in the ``end-migration`` part ? why not in classical ``post-migration``. 
This will make all the values unavailable during the end of the update.

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
```diff
@@ -0,0 +1,91 @@
...
+                list(st_lines_ids_dict)):
+            vals = {
+                'name': st_line.name,
+                'pos_order_id': st_line.pos_statement_id.id,
+                'amount': st_line.amount,
+                'payment_method_id': st_lines_ids_dict[st_line.id],
+                'payment_date': st_line.create_date,
+            }
+            vals_list += [vals]
+        env['pos.payment'].create(vals_list)
```
legalsylvain, line 89: I guess that such insert should be done by SQL, and not by ORM. In a database with some millions of records, it will take few seconds by SQL, and hours by ORM.

addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,99 @@
...
+def fill_pos_session_move_id(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE pos_session ps
+        SET move_id = po.{move}
+        FROM pos_order po
+        WHERE po.session_id = ps.id AND po.{move} IS NOT NULL
+            AND ps.move_id IS NULL AND ps.state = 'closed'
+        """.format(move=openupgrade.get_legacy_name('account_move'))
+    )
```
legalsylvain, line 49: If understood correctly (maybe I'm wrong ...)

- in V12, ``pos_order.account_move`` is a reference to the account_move generated by a pos_order when closing the session.

- in V13 ``pos_session.move_id`` is a reference to the global sale account_move generated when closing the session. 

it doesn't seems to be the same concept. don't you think ?

legalsylvain (changes requested):
Hi @MiquelRForgeFlow. Thanks for this work. some points inline.

Also a french question : 
 
- When we have ``l10n_fr_pos_cert`` installed, a hash is computed for each ``pos.order``.

- in V12 The hash is based on the field ``statement_ids`` (``comodel_name = account.bank.statement.line``) https://github.com/odoo/odoo/blob/12.0/addons/l10n_fr_pos_cert/models/pos.py#L57

- in V13 The hash is based on the field ``payment_ids`` (``comodel_name = pos.payment``) https://github.com/odoo/odoo/blob/13.0/addons/l10n_fr_pos_cert/models/pos.py#L40

(this change is totally crappy)

At this point, if we run a migration script from 12.0 to 13.0, the inalterability chain will be broken for odoo instance with ``l10n_fr_pos_cert`` installed.

I don't see manner to fix it, except to add a ``if isinstalled('l10n_fr_pos_cert')`` insert in ``pos_payment.id``, the value of the ``bank_statement_line.id``.

 What do you think ? 
@alexis-via, @vdewulf, @PierrickBrun : that can interest you.

addons/point_of_sale/migrations/13.0.1.0.1/openupgrade_analysis_work.txt
legalsylvain: Note. following that remark (https://github.com/OCA/OpenUpgrade/pull/2326#pullrequestreview-414629022), I think that ``pos_payment_method.receivable_account_id`` could be populated with default property, that are required in V12 to make the Point of Sale working. (https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/models/pos_order.py#L1005) without waiting for the ``end-migration`` step.

MiquelRForgeFlow: For having proper defaults, it must be done in end-migration after account module end-migration execution filling default accounts.

pedrobaeza: What is the problem with doing this in end-migration? I remember to need to move this because if not, we got errors.

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
pedrobaeza: I don't remember if it's needed for something special, but each approach has it cons and pros. Do you have real measures for your statement?

addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
MiquelRForgeFlow: The old `po.invoice_id` is the new `po.account_move`, so the new `ps.move_id` has to be obtained in some way from the old `po.account_move`.

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
legalsylvain: Well, when the migration is about table that can be big, SQL should be prefered. I don't have measures for this case, because I didn't tried to migrate my db with > 2,2 millions of records, but yes, introducing SQL has significant benefice, regarding duration. exemple : 
- https://github.com/OCA/OpenUpgrade/pull/2208 (from 5 hours to 1 hours)
- https://github.com/OCA/OpenUpgrade/pull/2151 (from 4 hours to 30 minutes)
- https://github.com/OCA/OpenUpgrade/pull/2148 (from 18 hours to 2 seconds)
- https://github.com/OCA/OpenUpgrade/pull/1934 (from hours to seconds)
- etc.

MiquelRForgeFlow:
@legalsylvain about your `l10n_fr_pos_cert` issue, I think it should be handled doing migration scripts in `l10n_fr_pos_cert`.

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
pedrobaeza: Yes, but unless you have measures proving that this is significantly slow, we shouldn't do any effort of going to SQL, that needs to handle by hand everything.

addons/account/migrations/13.0.1.1/end-migration.py
```diff
@@ -11,7 +11,21 @@ def add_default_account_data(env):
     Need to be on end-migration for making sure all chart of templates have
     their data loaded.
     """
+    def fill_chart_defaults(chart_tmpl):
```
pedrobaeza: Please get info recursively in a similar way of core.

MiquelRForgeFlow: what do you mean? it's already a recursive method

pedrobaeza: And why isolate it for this and not for the rest of values? You are also ignoring direct values in first CoA. You are forcing them to have a parent.

MiquelRForgeFlow: which other values? I am not forcing anything, see `if not chart_tmpl[key] and chart_tmpl.parent_id[key]:` condition.

pedrobaeza: And all under `if chart_tmpl.parent_id:`, so not having parent, it will never reach that code. And again: why not putting everything together in the same flow for not repeating things? DRY principle.

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
legalsylvain: > but unless you have measures proving that this is significantly slow

I don't understand you on that point. It seems to me quite obvious that it will be _very_ slower. Let's think backwards. Are there cases where creating millions of records through the ORM takes less or as much time as an SQL query ?

pedrobaeza: My fear is that doing by SQL, there were things not computed and this can affect. I always try for sure to do things on SQL, but on this code that is not mine, I wouldn't touch it unless performance drain is critical or it's very clear that there won't be difference on resulting data.

addons/account/migrations/13.0.1.1/end-migration.py
MiquelRForgeFlow: Well, at this moment my only interest is in making green the CI. Later, once CI is green, I can try to do a more dry code.

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
legalsylvain: > My fear is that doing by SQL, there were things not computed and this can affect

That's another point, that I can understand.
Well on that current topic, this new model is quite simple, and I really don't fear side effect, even If I don't use 13.0 for the time being (See : https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/models/pos_payment.py)

But well, if it's not possible to do it in that PR, let's go ahead. I'll improve it if someone hasn't already done so, when I'll work on the migration to 16.

pedrobaeza: Yeah, seen like that, it doesn't seem to be very complicated, but as said, I haven't touched it. Maybe when doing my review, I decide to convert it for the same reasons as you want. Are you serious about not doing the jump till v16?

addons/point_of_sale/migrations/13.0.1.0.1/end-migration.py
```diff
@@ -0,0 +1,67 @@
...
+
+def create_pos_payments(env):
+    env.cr.execute("""
+        SELECT DISTINCT absl.id, ppm.id
+        FROM account_bank_statement abs
+        JOIN pos_session ps ON abs.pos_session_id = ps.id
+        JOIN account_bank_statement_line absl ON absl.statement_id = abs.id
+        LEFT JOIN account_journal aj ON abs.journal_id = aj.id
+        JOIN pos_payment_method ppm ON ppm.cash_journal_id = aj.id
+    """)
```
etobella, line 76: I think you need to add
` WHERE  absl.pos_statement_id is not NULL`

Also, in order to avoid the browse, you can return also `absl.create_date`, `absl.amount` and `absl.name` WDYT?

MiquelRForgeFlow: done

jguenat (commented):
Functional review
Great work! The migration went well except for the pos.payment.method.
I don't understand de idea behind calling the post_install_pos_localisation() and not trying to migrate payment methods ?
The result is a loss of all custom payment method history in the pos.order payments tab.
I guess it's not a big deal as account.move.lines are what really matter and custom payment methods can be reconfigured manually post migration.

etobella:
@jguenat A PR exists fixing this kinds of issues I also found https://github.com/ForgeFlow/OpenUpgrade/pull/24

jguenat (approved):
Functional review
Worked fine on my database, tested it a little bit in V13 after migration (not going live as I have other modules to migrate yet).

pedrobaeza:
This one has been long standing and all things seem to be honored, so I finally merge it. Possible extra things can be handled in later PRs.

PR #2327, opened 2020-04-27
[10.0][IMP] hr_expense: Rescue sheets from v8

pedrobaeza:
If you come from v8 and haven't cleaned up table, it's possible to rescue previous expense sheets and use them in this version.

@Tecnativa TT23685

### DIFF
--- a/addons/hr_expense/migrations/10.0.2.0/pre-migration.py
+++ b/addons/hr_expense/migrations/10.0.2.0/pre-migration.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2017 Tecnativa - Vicent Cubells
+# Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from openupgradelib import openupgrade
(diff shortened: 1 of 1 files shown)


addons/hr_expense/migrations/10.0.2.0/pre-migration.py
```diff
@@ -11,9 +12,50 @@
...
+    ])
+    openupgrade.logged_query(
+        env.cr, """UPDATE hr_expense_sheet hes
+        SET accounting_date = am.date
+        FROM account_move am
+        WHERE am.id = hes.account_move_id""",
+    )
+    openupgrade.map_values(
+        env.cr, 'openupgrade_legacy_9_0_state', 'state', [
+            ('draft', 'submit'),
```
MiquelRForgeFlow, line 45: maybe the ones in draft could be deleted, as it's not the same a confirmed (submited) one that a draft which was not even confirmed.

pedrobaeza: Well, not really, but we don't want to lose those, and more having in v12 again the draft state. I have just pushed directly a little patch for it:

https://github.com/OCA/OpenUpgrade/commit/7abe19fa2220bb27b92e3d26901dfa762b018377

so I would go for this way.

MiquelRForgeFlow: then don't delete them but don't change their state and left them in draft although they are in an invalid state.

if not, imagine the following: coming from v8, with record in draft, then migrate to v10, and then stay in v10 production during some months because client doesn't want to migrate too fast to v12. then in this draft record in "submit" state, there are some changes, and maybe also the state also change no another state. then, few months later you migrate to v12 and then this record becomes again in 'draft' state! that doesn't makes sense!!!

And why would you want to make a stop in v10 instead of migrating from v8 to v12 in one swoop? well, I could give you several reasons.

pedrobaeza: Why they are in an invalid state? Being submitted doesn't mean they are anything more than that. Semantically, if you go only from v8 to v10, you mustn't lose that expenses groups, but auto-submit them.

MiquelRForgeFlow: ok, I see now that in the patch in v12 you only change to draft the ones that are still in submit (and were old draft).

MiquelRForgeFlow (approved):

addons/hr_expense/migrations/10.0.2.0/pre-migration.py
pedrobaeza: Yes, that's the safeguard part of the comment in the commit :wink:

PR #2331, opened 2020-05-03
10.0 fix point of sale bad invoice journal

legalsylvain:
The V10 introduce on ``pos.config`` a new journal field ``invoice_journal_id``

The default function ``_default_invoice_journal`` is used to populate this new column.  ([see](https://github.com/OCA/OpenUpgrade/blob/10.0/addons/point_of_sale/models/pos_config.py#L38))
the journal ID will be wrong in a multi company context.

**Fix**
Added a pre-migration script to add the field, if not exist. (the field can exist if the module V8 [pos_invoice_journal](https://github.com/OCA/pos/tree/8.0/pos_invoice_journal) is installed)

CC : original authors : @JordiBForgeFlow, @MiquelRForgeFlow 

--

### DIFF
--- a/addons/point_of_sale/migrations/10.0.1.0.1/openupgrade_analysis_work.txt
+++ b/addons/point_of_sale/migrations/10.0.1.0.1/openupgrade_analysis_work.txt
@@ -13,7 +13,7 @@ point_of_sale / pos.config               / state (selection)             : DEL r
 point_of_sale / pos.config               / default_fiscal_position_id (many2one): NEW relation: account.fiscal.position
 point_of_sale / pos.config               / iface_fullscreen (boolean)    : DEL
 point_of_sale / pos.config               / invoice_journal_id (many2one) : NEW relation: account.journal
-# NOTHING TO DO
+# Done: pre-migration. We set the invoice_journal_id with the journal_id value, because, default function return bad journal in multicompany context. We populate the value only if the column doesn't exist. (that is the case if the OCA module pos_invoice_journal is installed in previous version)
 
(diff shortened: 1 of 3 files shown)


addons/point_of_sale/migrations/10.0.1.0.1/openupgrade_analysis_work.txt
```diff
@@ -13,7 +13,7 @@ point_of_sale / pos.config               / state (selection)             : DEL r
 point_of_sale / pos.config               / default_fiscal_position_id (many2one): NEW relation: account.fiscal.position
 point_of_sale / pos.config               / iface_fullscreen (boolean)    : DEL
 point_of_sale / pos.config               / invoice_journal_id (many2one) : NEW relation: account.journal
-# NOTHING TO DO
+# Done: pre-migration. We set the invoice_journal_id with the journal_id value, because, default function return bad journal in multicompany context. We populate the value only if the column doesn't exist. (that is the case if the OCA module pos_invoice_journal is installed in previous version)
```
MiquelRForgeFlow, line 16: Also, shouldn't we add the merge of `pos_invoice_journal` into `point_of_sale`, in v9 or v10 OpenUpgrade apriori file?

legalsylvain: In V10, indeed, because the feature is introduced in V10 in the core module ``point_of_sale``.

Done ! Thanks a lot for the improvment.

legalsylvain:
Hi @MiquelRForgeFlow. could you update your review ? 

thansk !

MiquelRForgeFlow (approved):

PR #2334, opened 2020-05-08
[ADD] Rename l10n_nl_intrastat

StefanRijnhart:


pedrobaeza:
I suppose it's correct, but I don't know this localization module. I think @hbrunn's quick confirmation serves for giving a go.

StefanRijnhart:
Here's a reference: https://github.com/OCA/l10n-netherlands/issues/118#issuecomment-396529112

pedrobaeza:
OK, then I assume extra migration scripts will be needed, but on the target module.

pedrobaeza (approved):

StefanRijnhart:
Added a flake8 fix

pedrobaeza:
Can you squash both?

StefanRijnhart:
Done

Issue #2336, opened 2020-05-14
[11.0] Sale line linked to project is of another partner sale order


sergiocorato:
I'm investigating why this happen (in 2 out of 3 migration)

https://github.com/OCA/OpenUpgrade/blob/98e6d57611c4ccc1a6b4b9a075b1ab8080149390/addons/sale_timesheet/migrations/11.0.1.0/post-migration.py#L122

sergiocorato:
I add PR for solution

pedrobaeza:
Well, it's good habit to link both, but I'm closing this having the PR.

PR #2337, opened 2020-05-15
[11.0][FIX] link project with analytic account id to fix sol search

sergiocorato:
Description of the issue/feature this PR addresses: #2336

Current behavior before PR:

--


pedrobaeza:
This seems your specific use case, but I think others don't fit with strict partner assignation on project/analytic account.

sergiocorato:
> This seems your specific use case, but I think others don't fit with strict partner assignation on project/analytic account.

@pedrobaeza with the last commit it works in this migration, perhaps it was a specific case, but I have a second specific case (all not 'mine' installations but acquired)

pedrobaeza (approved):
This sounds better indeed and can be the cause of some failing and others not, as the ID for both AA/project may be the same if you only create projects.

PR #2342, opened 2020-05-30
[12.0][IMP] populate account.fiscal.year with account.fiscalyear data 

legalsylvain:
- if we come from a 8.0 database, the fiscal years are present in the table ``account_fiscalyear``. (model ``account.fiscalyear``).
- V12.0 reintroduce a similar model ``account.fiscal.year`` in the table ``account_fiscal_year``.

So, this PR improve the V12 migration script, populating new table with the old datas, if the old table is present.

CC : @StefanRijnhart, @pedrobaeza, @MiquelRForgeFlow

Thanks for your review ! 

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/account/migrations/12.0.1.1/openupgrade_analysis_work.txt
@@ -32,7 +32,8 @@ account      / account.fiscal.year      / company_id (many2one)         : NEW re
 account      / account.fiscal.year      / date_from (date)              : NEW required: required
 account      / account.fiscal.year      / date_to (date)                : NEW required: required
 account      / account.fiscal.year      / name (char)                   : NEW required: required
-# NOTHING TO DO (new model)
+# Done: post-migration: populate with old account.fiscalyear model
+# if we come from a 8.0 database.
 
 account      / account.group            / parent_left (integer)         : DEL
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I prefer to not have that entries if the fiscal year is equal to the natural one. Is there any possibility of checking that?

legalsylvain:
Hi @pedrobaeza. thanks for your answer. It seems, you don't like a lot this V12 model -;)

Well, I could add a test if month != 12 and day != 31 (for ``date_from`` and ``date_to`` fields) but, in fact, we cant know if user want to keep or not the data, even if the fiscal year is the "natural one".

My point is to keep all the data. Then, if user don't want to use them, there are free to drop them very easely.
- Keep there data will not hurt anything, (AFAIK)
- it's only a few records in the database. (not a space problem IMO)
- for users that want to use some OCA modules like [account_move_fiscal_year](https://github.com/OCA/account-financial-tools/tree/12.0/account_move_fiscal_year), it avoids to recreate all the fiscal years. CC : @benwillig, @bjeficent

Do you think it's acceptable ?

sbidoul (approved):
Sounds reasonable to me.

pedrobaeza:
OK, I suppose that getting fiscal year dates doesn't take more time having this table than not having it:

/ocabot merge nobump

PR #2298, opened 2020-04-13
[13.0][MIG] mrp

Yasaie:
Migration of `mrp`
--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2346, opened 2020-06-03
[13.0][MIG] mrp

kos94ok-3D:
Supersedes #2298.

--


addons/mrp/migrations/13.0.2.0/post-migration.py
```diff
@@ -0,0 +1,151 @@
...
+        wh['manufacture_mto_pull_id'] = env['stock.rule'].create(values).id
+
+
+def fill_mrp_workorder_product_uom_id(cr):
+    openupgrade.logged_query(
+        cr, """
+        UPDATE mrp_workorder mw
+        SET product_uom_id = mp.product_uom_id
+        FROM mrp_production mp
+        WHERE mw.production_id = mp.id AND mw.product_uom_id IS NULL
```
dennybiasiolli, line 221: If you want to read `uom_id` field directly from `product_template` table, you can act as follows:
```sql
UPDATE mrp_workorder w
SET    product_uom_id = pt.uom_id
FROM   product_product pp
JOIN   product_template pt ON pp.product_tmpl_id = pt.id
WHERE  w.product_id = pp.id
AND    w.product_uom_id IS NULL;
```

dennybiasiolli (approved):

addons/mrp/migrations/13.0.2.0/post-migration.py
kos94ok-3D: @dennybiasiolli 
I don't know why you thought that.
I set product uom according to odoo12
See: https://github.com/odoo/odoo/blob/3ca2a1903da2431bbf72d6112d659d79d470f249/addons/mrp/models/mrp_workorder.py#L39

dennybiasiolli: You're right, I think it was "cleaner" to read `product_uom_id` directly from `product_template`, but looking at the link you provided it's cleaner your way ;)

addons/mrp/migrations/13.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,84 @@
...
+          END
+        WHERE reservation_state in ('partially_available', 'none')
+        """
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.remove_tables_fks(env.cr, 'mrp_bom_line_product_attribute_value_rel')
+    openupgrade.rename_models(env.cr, _model_renames)
```
dennybiasiolli, line 92: The code breaks here if `mrp_subproduct` was not installed in 12.0.
I suggest to add
```python
    if not openupgrade.table_exists(env.cr, 'mrp_subproduct'):
        create_model_table(env.cr, 'mrp_subproduct', 'Byproduct')
```
before this instruction, remembering to add an import for `create_model_table` to the top of the file:
`from odoo.tools import create_model_table`

dennybiasiolli (changes requested):
Update break if `mrp_subproduct` was not installed in 12.0.
Take a look to the suggested solution

addons/mrp/migrations/13.0.2.0/pre-migration.py
kos94ok-3D: @pedrobaeza I think it has to been done in rename_tables of openupgradelib instead.
What do you think?

kos94ok-3D: @dennybiasiolli  Is OK now?

dennybiasiolli: It's perfect, thanks!

dennybiasiolli (approved):

addons/mrp/migrations/13.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,90 @@
...
+
+def mapped_reservation_state(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE mrp_production
+        SET reservation_state = CASE
+          WHEN reservation_state = 'partially_available' THEN 'assigned'
+          WHEN reservation_state = 'none' THEN NULL
+          END
+        WHERE reservation_state in ('partially_available', 'none')
```
dennybiasiolli: please change this line to `WHERE reservation_state in ('partially_available', 'confirmed')`.
Without this fix every mrp_production record in this state gives error when trying to start/stop working on its workorders.

dennybiasiolli (changes requested):

addons/mrp/migrations/13.0.2.0/pre-migration.py
kos94ok-3D: `reservation_state` was `availability` in v12, `availability` did not have value `confirmed`

dennybiasiolli: ok right, so the error is in `mrp_workorder.production_availability`, the field is defined as 
```
    production_availability = fields.Selection(
        'Stock Availability', readonly=True,
        related='production_id.reservation_state', store=True,
        help='Technical: used in views and domains only.')
```

In this case we should add another migration query, because it contains wrong values:

```
    openupgrade.logged_query(
        env.cr, """
        UPDATE mrp_workorder
        SET production_availability = CASE
          WHEN production_availability = 'partially_available' THEN 'assigned'
          WHEN production_availability = 'none' THEN NULL
          END
        WHERE production_availability in ('partially_available', 'none')
        """
    )
```

kos94ok-3D: Try with changes from https://github.com/kos94ok-3D/OpenUpgrade/commit/dc1f2321d42ecf7be7f7340a8e757c65f818b8e0

dennybiasiolli: Thanks, I'm trying right now, I'll let you know if it works

dennybiasiolli: @kos94ok-3D it keeps breaking when I try to start a workorder. Launching this query solves the problem:

```sql
UPDATE mrp_workorder
SET production_availability = CASE
  WHEN production_availability = 'partially_available' THEN 'assigned'
  WHEN production_availability = 'none' THEN NULL
  END
WHERE production_availability in ('partially_available', 'none')
```

kos94ok-3D: Add at the end of function
<pre>
    # production_availability of mrp.workorder because it related and store=True
    openupgrade.logged_query(
        env.cr, """
        UPDATE mrp_workorder wo
        SET production_availability = mo.reservation_state
        FROM mrp_production mo
        WHERE wo.production_id = mo.id
        """
    )
</pre>

dennybiasiolli: Yes, with this I think that all works perfectly

addons/mrp/migrations/13.0.2.0/post-migration.py
```diff
@@ -0,0 +1,217 @@
...
+    SET date_planned_start = mo.date_planned_start, date_planned_finished = mo.date_planned_finished
+    FROM mrp_production mo
+    WHERE wo.state NOT IN ('done', 'cancel') AND wo.production_id = mo.id
+        AND wo.date_planned_finished IS NULL AND wo.date_planned_start IS NULL
+    """)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    mapped_reservation_state
```
dennybiasiolli: maybe `mapped_reservation_state(env)` ?

dennybiasiolli (changes requested):

kos94ok-3D:
Hi, @MiquelRForgeFlow.
Can you remove the `Blocked by dependency` label?

MiquelRForgeFlow:
done

addons/mrp/migrations/13.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,128 @@
...
+        ('lot_produced_id', None),
+    ],
+}
+
+_column_copies = {
+    'mrp_production': [
+        ('availability', 'reservation_state', None),
+    ],
+    'mrp_bom_line_product_template_attribute_value_rel': [
+        ('product_attribute_value_id', 'product_template_attribute_value_id', None),
```
LoisRForgeFlow: After migration `product_attribute_value_id` remains in the schema as a required column, this makes imposible to write the field "Apply on variants" on a BoM.

image

LoisRForgeFlow: Odoo logs:
```
bad query: INSERT INTO mrp_bom_line_product_template_attribute_value_rel (mrp_bom_line_id, product_template_attribute_value_id) VALUES (14535, 13596) ON CONFLICT DO NOTHING
ERROR: null value in column "product_attribute_value_id" violates not-null constraint
DETAIL:  Failing row contains (14535, null, 13596).
```

MiquelRForgeFlow: solved

kos94ok-3D:
Rebase and diff
изображение

MiquelRForgeFlow:
@kos94ok-3D if you don't mind, I added a commit to move the filling of the m2m table to post-migration. This way the old table still exists after the migration (and I need that table to test some things).

MiquelRForgeFlow (approved):
Tested last changes. To me it's good this PR to merge.

addons/mrp/migrations/13.0.2.0/post-migration.py
```diff
@@ -272,6 +272,40 @@ def fill_planned_datetime(env):
...
+    A part from the usual value for raw moves in mrp productions,
+    now this field has a value for delivery moves generated from a kit.
+    See _prepare_phantom_move_values method, that has bom_line_id field."""
+    openupgrade.logged_query(
+        env.cr, """
+    UPDATE stock_move sm0
+    SET bom_line_id = mbl0.id
+    FROM (
+        SELECT moves.id, moves.product_id,
+            (array_agg(moves.bom_id))[1] as bom_id
```
kos94ok-3D, line 286: Why not use min or max?
```suggestion
        SELECT moves.id, moves.product_id,
            min(moves.bom_id) as bom_id

```

MiquelRForgeFlow: Because I'm not interested in the min or max. I select the first in the selected order, that's why I put `ORDER BY sm.id, mb.sequence, mb.product_id`.

MiquelRForgeFlow: Selection_129

pedrobaeza:
Not checked at all, but reviewed by some of you and run in a DB without errors, so merging for having a working version.

PR #2350, opened 2020-06-15
[FIX] mail: remove admin from mail.alias

yajo:
~~Since Administrator is not going to be used anymore, it is best to archive that partner, just like it is done with that user. If not doing so, accessing the user will display a warning about this.~~

When a mail.alias record belongs to admin, it will try to send mails in name of a user that most likely has wrong data in it, and is actually garbage. That can produce a problem where aliases create records but don't notify followers, or even worse, that same problem in loop (depending possibly on if you use IMAP4 or POP3 to fetch mails).

@Tecnativa TT24347

### DIFF
--- a/addons/mail/migrations/12.0.1.0/post-migration.py
+++ b/addons/mail/migrations/12.0.1.0/post-migration.py
@@ -31,9 +31,25 @@ def fill_mail_notification_mail_id(cr):
     )
 
 
+def remove_admin_alias_owner(cr):
+    """Admin's aliases would fail.
+
+    Admin is going to be disabled later in base end-migration.
+    When that happens, any aliases belonging to him would fail.
+
+    By removing the alias owner, we let Odoo decide the default, which will
+    be __system__ usually, the new admin replacement.
+    """
+    openupgrade.logged_query(
+        cr,
+        "UPDATE mail_alias SET alias_user_id = NULL WHERE alias_user_id = 1",
+    )
+
+
 @openupgrade.migrate(use_env=False)
 def migrate(cr, version):
     fill_mail_blacklist_res_partner(cr)
     fill_mail_notification_mail_id(cr)
+    remove_admin_alias_owner(cr)
     openupgrade.load_data(
         cr, 'mail', 'migrations/12.0.1.0/noupdate_changes.xml')


pedrobaeza:
@StefanRijnhart you did this part. Can you please check?

StefanRijnhart:
I don't understand why you say that the Administrator is not going to be used anymore. I believe it is, but its privileges will now be based on group membership instead of having a magic ID (that is the prerogative of the inactive root user).

addons/mail/migrations/12.0.1.0/post-migration.py
```diff
@@ -31,9 +31,25 @@ def fill_mail_notification_mail_id(cr):
...
+
+    Admin is going to be disabled later in base end-migration.
+    When that happens, any aliases belonging to him would fail.
+
+    By removing the alias owner, we let Odoo decide the default, which will
+    be __system__ usually, the new admin replacement.
+    """
+    openupgrade.logged_query(
+        cr,
+        "UPDATE mail_alias SET alias_user_id = NULL WHERE alias_user_id = 1",
```
StefanRijnhart, line 45: Without this query, ownership of the aliases would be transferred to the (new) Admin user in the end migration. I see system aliases do not have an owner in new Odoo 12 databases but can have the Admin user in older databases so I think resetting the alias_user_id here is correct.

odoo/addons/base/migrations/12.0.1.3/end-migration.py
```diff
@@ -104,11 +104,11 @@ def fork_off_system_user(env):
...
         (user_root.id, ),
     )
-    # Ensure also partner_root is inactive
+    # Ensure also partner_root and partner_admin are inactive
     env.cr.execute(
         """ UPDATE res_partner
-            SET active = FALSE WHERE id = %s """,
-        (partner_root.id,))
+            SET active = FALSE WHERE id IN %s """,
+        ((partner_root.id, partner_admin.id),))
```
StefanRijnhart: I think setting the admin user's partner to inactive is incorrect. On new Odoo 12 databases there is an active admin user with an active partner.

yajo: Oh I see. It's weird that's not just demo data. I guess this could have been a side effect of seeing a warning in the admin user form, but yes probably this isn't the real fix to the problem.

I'll revert this.

yajo:
All solved.

ivantodorovich (approved):

StefanRijnhart (approved):
Thanks for the update!

PR #2352, opened 2020-06-17
[12.0][FIX] hr_holidays: Re-assign holidays to resource.calendar.leaves

pedrobaeza:
They were disconnected due to the column renaming, but can't be done before or we will get a FK constraint error.

@Tecnativa TT24299

### DIFF
--- a/addons/hr_holidays/migrations/12.0.1.5/post-migration.py
+++ b/addons/hr_holidays/migrations/12.0.1.5/post-migration.py
@@ -2,6 +2,7 @@
 # Copyright 2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+from psycopg2 import sql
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

addons/hr_holidays/migrations/12.0.1.5/post-migration.py
```diff
@@ -84,6 +85,19 @@ def fill_hr_leave(env):
         _move_model_in_data(env, ids, 'hr.holidays', 'hr.leave')
 
 
+def restore_resource_calendar_leave_link(env):
+    """We set this now from the renamed column as can't be done previously
+    or we will get fkey violation.
+    """
+    openupgrade.logged_query(
+        env.cr, sql.SQL(
+            "UPDATE resource_calendar_leave SET holiday_id = {}"
```
MiquelRForgeFlow:
```suggestion
            "UPDATE resource_calendar_leaves SET holiday_id = {}"
```

PR #2275, opened 2020-04-07
[13.0][MIG] account

MiquelRForgeFlow:
Migration of `account`.

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2359, opened 2020-07-02
[11.0][IMP] mrp: perf issue in consume sml

sebastienbeau:

Insertion is really really long (more then one day), I have more then 1 millions of mrp.production order

Current behavior before PR:

The table stock_move_line_consume_rel is empty (new table), but we do a check to not insert the record if already exist.
The issue is before inserting a record a check is done, so inserting record is really long for postgresql (version 10)

Remove this check as the table is empty (new table on V11)
And the distinct will already remove duplicated entry

@MiquelRForgeFlow can you give your feedback please.

I am re-running the migration so I will have stats soon

--

### DIFF
--- a/addons/mrp/migrations/11.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/11.0.2.0/post-migration.py
@@ -197,12 +197,6 @@ def fill_stock_move_line_consume_rel(cr):
             ON sml1.move_id = sqmr1.move_id
         INNER JOIN stock_move_line sml2
             ON sml2.move_id = sqmr2.move_id
-        WHERE NOT EXISTS (
-            SELECT *
-            FROM stock_move_line_consume_rel
-            WHERE consume_line_id = sml1.id
-                AND produce_line_id = sml2.id
-        )
         """
     )
 


pedrobaeza:
@sebastienbeau please get used to proper commit/PR messages

MiquelRForgeFlow:
@sebastienbeau it would be better if you could provide example data with before/after migration with/without patch to be able compare and understand the patch.

sebastienbeau:
@pedrobaeza sorry I will take care

sebastienbeau:
@MiquelRForgeFlow You want me to
- add a first commit with a test like that https://github.com/OCA/OpenUpgrade/tree/11.0/addons/account/migrations/11.0.1.1/tests
- add my fix and check that nothing have change
Right ?

MiquelRForgeFlow:
@sebastienbeau no need of that complex thing.

just show image of data in a comment, like https://github.com/OCA/OpenUpgrade/pull/2275#issuecomment-628255038 comment.

sebastienbeau:
@pedrobaeza the milestone should be 11.0

pedrobaeza:
Right, changed. That's why I ask you to put full info in commit/PR title :wink:

sebastienbeau:
thanks ;)

addons/mrp/migrations/11.0.2.0/post-migration.py
```diff
@@ -197,12 +197,6 @@ def fill_stock_move_line_consume_rel(cr):
             ON sml1.move_id = sqmr1.move_id
         INNER JOIN stock_move_line sml2
             ON sml2.move_id = sqmr2.move_id
-        WHERE NOT EXISTS (
```
StefanRijnhart, line 200: You can replace with `ON CONFLICT DO NOTHING`

pedrobaeza: Which PG version introduces this syntax? Maybe the version is too high for this Odoo version

StefanRijnhart:
```
insert into res_groups_users_rel values(2, 3);
ERROR:  duplicate key value violates unique constraint "res_groups_users_rel_pkey"
DETAIL:  Key (gid, uid)=(2, 3) already exists.
insert into res_groups_users_rel values(2, 3) ON CONFLICT DO NOTHING;
INSERT 0 0
```

pedrobaeza: It was introduced in PG 9.5:

https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.5#INSERT_..._ON_CONFLICT_DO_NOTHING.2FUPDATE_.28.22UPSERT.22.29

What is the minimum PG version for v11?

StefanRijnhart: It's introduced in 9.5

StefanRijnhart: I think that's entirely reasonable to expect people to run their migration on. Our travis config uses 9.6.

pedrobaeza: OK, no problem for me.

pedrobaeza:
@sebastienbeau can you attende Stefan's comment?

PR #2360, opened 2020-07-08
Add sponsors to README

simahawk:
The board got 1k from Odoo Italia community after we launched the funding campaign.

We'd like to give all sponsors a bit of recognition.

As many people/companies already put a lot of effort here, maybe we should add more items to the list (suggestions?).

Maybe there's a better place or an additional place where we want to put this (suggestions?).

@OCA/openupgrade-maintainers can you provide your feedback?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,10 @@ OpenUpgrade
 ===========
 
 This <a href="https://odoo-community.org">OCA</a> project aims to provide an Open Source upgrade path for <a href="https://github.com/odoo/odoo">Odoo</a>. It is hosted at <a href="https://github.com/oca/openupgrade">GitHub</a>. For documentation, see <a href="https://doc.therp.nl/openupgrade">here</a>.
+
+Sponsors
+--------
+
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Instead of here, I would put them at the same level than OCA general sponsors.

simahawk:
@pedrobaeza thanks for you feedback.
Regarding sponsorship presence on the website: we'll do that (should be there already indeed).
At the same time it would be nice to see some attribution on specific projects.
Maybe we can handle this in the future at website level w/ a relation to supported projects/modules but for now this seems a cheap option. No?

pedrobaeza:
But that way visibility is going to be asymmetric: one general sponsor only appears on website and OU sponsor in 2 places? I think all should be handled in the website, not splitting this around GitHub.

simahawk:
I see your point and somehow you are right. The board will establish some rules to handle such cases better. Discussion ongoing.
For now I keep this on hold.

PR #2365, opened 2020-07-16
[10.0][FIX] hr_timesheet: Create project for all possible timesheets

pedrobaeza:
If you have disabled "use_timesheets" in an analytic account, but you have already introduced timesheets for it, when migrating, you get a problem, as these timesheets won't have any linked project.

Changing the SQL for creating missing projects, basing it also in the existing timesheets, apart from the existing criteria, fixes the problem.

cc @Tecnativa

### DIFF
--- a/addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/10.0.1.0/post-migration.py
@@ -22,13 +22,14 @@ def migrate_missing_projects(env):
     """Create a project and link it to the orphaned analytic accounts"""
     env.cr.execute(
         """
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@bodi000 please review

StefanRijnhart (approved):

StefanRijnhart:
Having read the conversation on #2362, I agree with this alternatie. @bodi000 about not renaming the one field while the other field is renamed, you are right about the consistency but it's a bit late in the 10.0 development cycle to do so now.

MiquelRForgeFlow (approved):
Very special case, but ok.

bodi000:
fine with me. I thought the second version was cleaner, but I got your point :-)

Best

PR #2367, opened 2020-07-22
[11.0][FIX] base: Adjust res.partner.category colors

pedrobaeza:
Colors have changed over versions, so we try to preserve the most similar ones.

Selección_032

Selección_034

@Tecnativa TT17617

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/post-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/post-migration.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2017 bloopark systems (<http://bloopark.de>)
+# Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import json
 from odoo.tools import pickle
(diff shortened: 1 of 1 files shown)


chienandalu (approved):
:+1: Looks good :)

ivantodorovich (approved):

yajo:
> Selección_032
> 
> Selección_034



:laughing: Jokes apart, good job!

BTW, would this affect the whole Odoo framework everywhere a given record uses colors? Should this go then into the version-specific file of openupgradelib?

pedrobaeza:
Not sure if these colors are applied in other places. Do you know other cases?

yajo:
A quick search of `color_field` in the v11 codebase yields a few results:

- account.invoice.line.analytic_tag_ids
- crm.lead.tag_ids
- fleet.vehicle.tag_ids
- hr.employee.category_ids
- hr.applicant.categ_ids
- note.note.tag_ids
- project.task.tag_ids

... and a few more. Maybe some of these apply to the same models, I didn't dig that much.

But it seems to me like this could fit fine in openupgradelib then. If you prefer, we can leave that for when this is actually needed? :thinking: But it's good to keep it in mind.

PR #2370, opened 2020-07-29
[11.0][FIX] website: set website.page url by view key, view xmlid

hbrunn:
I wonder how this could have worked before? In v10, the page controller [requests a template](https://github.com/OCA/OCB/blob/10.0/addons/website/controllers/main.py#L121) which then [requests a view](https://github.com/OCA/OCB/blob/10.0/addons/website/models/website.py#L425) which [searches for the key](https://github.com/OCA/OCB/blob/10.0/addons/website/models/ir_ui_view.py#L81) and [falls back to the xmlid](https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_ui_view.py#L959).

@karan-dreambits can you explain in which cases the linking by name works? I'm under the impression it will only ever work if the name happens to be the same as the key, in which case we can delete this part of the query.

### DIFF
--- a/addons/website/migrations/11.0.1.0/post-migration.py
+++ b/addons/website/migrations/11.0.1.0/post-migration.py
@@ -16,7 +16,13 @@ def fill_website_pages(env):
         INSERT INTO website_page
             (url, view_id, website_indexed, website_published, create_uid,
              create_date, write_uid, write_date)
-        SELECT COALESCE(wm.url, '/page/' || iuv.name), iuv.id, TRUE, TRUE,
+        SELECT COALESCE(
+                wm.url,
+                '/page/' || COALESCE(
+                    substring(iuv.key from 9), imd.name, iuv.name
+                )
+            ),
+            iuv.id, TRUE, TRUE,
             iuv.create_uid,
             iuv.create_date, iuv.write_uid, iuv.write_date
         FROM
@@ -28,6 +34,10 @@ def fill_website_pages(env):
             AND (wm.website_id = iuv.website_id
             OR wm.website_id IS NULL
             OR iuv.website_id IS NULL)
+        LEFT JOIN
+            ir_model_data imd ON imd.module='website'
+            AND imd.model='ir.ui.view'
+            AND imd.res_id=iuv.id
         WHERE
             wp.id IS NULL
             AND iuv.page = TRUE""",


MiquelRForgeFlow:
Your patch is considering that all views are from website module, which have keys with "website._anything_" format. But I think maybe this is wrong, because you may have website_page linked views from other modules, like website_crm, website_hr_recruitment or website_event, which have different key format.

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -16,7 +16,13 @@ def fill_website_pages(env):
         INSERT INTO website_page
             (url, view_id, website_indexed, website_published, create_uid,
              create_date, write_uid, write_date)
-        SELECT COALESCE(wm.url, '/page/' || iuv.name), iuv.id, TRUE, TRUE,
+        SELECT COALESCE(
+                wm.url,
+                '/page/' || COALESCE(
+                    substring(iuv.key from 9), imd.name, iuv.name
```
MiquelRForgeFlow, line 22:
```suggestion
                    substring(iuv.key from (strpos(iuv.key, '.')+1)), imd.name, iuv.name
```

addons/website/migrations/11.0.1.0/post-migration.py
```diff
@@ -28,6 +34,10 @@ def fill_website_pages(env):
             AND (wm.website_id = iuv.website_id
             OR wm.website_id IS NULL
             OR iuv.website_id IS NULL)
+        LEFT JOIN
+            ir_model_data imd ON imd.module='website'
+            AND imd.model='ir.ui.view'
+            AND imd.res_id=iuv.id
```
MiquelRForgeFlow, line 40:
```suggestion
            ir_model_data imd ON (imd.model='ir.ui.view' AND imd.res_id=iuv.id)
```

MiquelRForgeFlow (commented):
I think with this suggestions, we can solve the problem I pointed before.

addons/website/migrations/11.0.1.0/post-migration.py
hbrunn: thanks, but I don't believe this fixes the issue you mentioned. when we had an xmlid as key in v10, it was a full xmlid in the url. So if we want to cover this case too, I think we better add another left join that handles the non-website xmlids/keys

pedrobaeza:
Is this going to be finished?

PR #2372, opened 2020-08-03
[FIX] utm: avoid issue '"utm_stage" does not exist'

MiquelRForgeFlow:
Only affected databases without mass_mailing, because those cannot obtain "utm_stage" table from "mail_mass_mailing_stage".

_utm.stage_ model is created after _utm.campaign_ model, but stage_id field of _utm.campaign_ relates to _utm.stage_. When installing the module, it doesn't break. But for unkown reason, when migrating, it breaks.

Thus, as an easy solution, we define the _utm.stage_ model before _utm.campaign_ model.

Another solution could be create the utm_stage table manually in pre-migration.

--

### DIFF
--- a/addons/utm/migrations/13.0.1.0/pre-migration.py
+++ b/addons/utm/migrations/13.0.1.0/pre-migration.py
@@ -1,6 +1,7 @@
 # Copyright 2020 ForgeFlow <http://www.forgeflow.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+from odoo.tools import create_model_table
 
 _model_renames = [
     ('mail.mass_mailing.stage', 'utm.stage'),
@@ -63,3 +64,5 @@ def migrate(env, version):
             ],
             False,
         )
+    else:
+        create_model_table(cr, "utm_stage", "Campaign Stage")


pedrobaeza:
As talked privately, I don't think this is the ideal solution, and other thing should be on the background. Anyway, your patch is declaring 2 times the same model with `_name`. That's incorrect.

AaronHForgeFlow (commented):
Functionally speaking, it works :)

legalsylvain:
Hi. Thanks for the patch. As pedro said, I don't think it's a good idea to add code in modules. The second option you mention is more in the spirit of Openupgrade.  (Create table in premigration script, if not exist) and not more complex.

StefanRijnhart (approved):
I can live with this

pedrobaeza:
See my comment in https://github.com/OCA/OpenUpgrade/pull/2392#issuecomment-671989842

StefanRijnhart:
@pedrobaeza that's true, triggering the default will be avoided if the column is precreated. Probably the best solution.

dennybiasiolli:
@pedrobaeza @StefanRijnhart another solution could be moving `class UtmStage` defininition on top of `class UtmCampaign` definition.
The tables would be created "top-down" and everything works fine.
Now I'm trying the "create column" solution, I'll let you know in a moment.

dennybiasiolli:
@pedrobaeza @StefanRijnhart I think `stage_id` column it's already created [at this line](https://github.com/OCA/OpenUpgrade/blob/13.0/addons/utm/migrations/13.0.1.0/pre-migration.py#L29) in `addons/utm/migrations/13.0.1.0/pre-migration.py`, so maybe the best solution it' the moving of `class UtmStage` on top of `class UtmCampaign` definition.

MiquelRForgeFlow:
@dennybiasiolli what you are saying is what this PR already is doing.

@pedrobaeza the column cannot be pre-created because the table doesn't exist.

dennybiasiolli:
Something like this does not solve the problem:

```git
diff --git a/addons/utm/migrations/13.0.1.0/pre-migration.py b/addons/utm/migrations/13.0.1.0/pre-migration.py
index d0bd85cf48a..a8eff443974 100644
--- a/addons/utm/migrations/13.0.1.0/pre-migration.py
+++ b/addons/utm/migrations/13.0.1.0/pre-migration.py
@@ -45,6 +45,12 @@ def move_mailing_campaign_to_utm_campaign(env):
     )


+def add_empty_stage_id_to_utm_campaign(env):
+    openupgrade.add_fields(env, [
+        ("stage_id", "utm.campaign", "utm_campaign", "integer", False, "utm"),
+    ])
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     cr = env.cr
@@ -63,3 +69,5 @@ def migrate(env, version):
             ],
             False,
         )
+    else:
+        add_empty_stage_id_to_utm_campaign(env)
```

I've already made a PR to odoo repo (https://github.com/odoo/odoo/pull/55795) in order to fix this at the source, but I can provide a PR here too, what do you think?

dennybiasiolli:
@MiquelRForgeFlow I was thinking about something like this: https://github.com/OCA/OpenUpgrade/pull/2393
Without adding new code, but only "moving" the class definitions

pedrobaeza:
@MiquelRForgeFlow if the problem is the table, pre-create the table as well. It's one SQL...

MiquelRForgeFlow:
@dennybiasiolli you are not taken into account that this repo is getting updated from upstream from time to time, and we like to do the minimum changes in odoo code to avoid conflicts.

@pedrobaeza ok, done. I just didn't want to create tables this way to avoid unexpected issues.

pedrobaeza:
Let me share with you why I'm so insistent with not modifying core source code: we are doing experiments with new features with migration addons path for avoiding to maintain full Odoo source code, and let OpenUpgrade only for pure migration scripts. There are still challenges like the changes in loading modules code, but it would be great to avoid to do upstream merges.

pedrobaeza (approved):

dennybiasiolli:
I'm trying right now the repo with the updated code, but it's not working. The newly created `utm_stage` table is missing `sequence` column.

```
psycopg2.errors.UndefinedColumn: column utm_stage.sequence does not exist
LINE 1: SELECT "utm_stage".id FROM "utm_stage" ORDER BY "utm_stage"....
```

After adding that column in pre-migration script everything works fine.

PR #2374, opened 2020-08-04
[13.0][MIG] website_blog

MiquelRForgeFlow:
Migration of website_blog.

--


addons/website_blog/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr, "website_blog", "migrations/13.0.1.0/noupdate_changes.xml")
```
yajo, line 9: This is wrong. You have to remove `name` and `subtitle` from that file:

https://github.com/OCA/OpenUpgrade/blob/1acf10ec2cf2839bd3ff3fa066142143a93ca86b/addons/website_blog/migrations/13.0.1.0/noupdate_changes.xml#L4-L5

yajo (changes requested):

addons/website_blog/migrations/13.0.1.0/post-migration.py
yajo: The fix is in https://github.com/OCA/OpenUpgrade/pull/2418.

yajo (approved):
Since the fix for https://github.com/OCA/OpenUpgrade/pull/2374/files#r505465263 is in https://github.com/OCA/OpenUpgrade/pull/2418, I approve this :blush:

Abranes (approved):

pedrobaeza:
@Yajo why do you merge this if the dependencies are not merged?

yajo:
Well it was approved, the code seems fine, and I didn't notice that label. (Actually I didn't know such a thing existed).

Anyways does this merge harm?

pedrobaeza:
Yes, it can produce problems, as this is installing all the dependencies in Travis and thus some conflicts can arise, although Travis right now is green. Please don't do that next time or ask not being you a usual contributor here.

addons/website_blog/migrations/13.0.1.0/noupdate_changes.xml
```diff
@@ -1,8 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="blog_blog_1" model="blog.blog">
-    <field name="name">Our blog</field>
-    <field name="subtitle">We are a team of passionate people whose goal is to improve everyone's life.</field>
     <field name="cover_properties">{"background-image": "url('/website_blog/static/src/img/cover_5.jpg')", "resize_class": "o_record_has_cover cover_mid", "background-color": "oe_black", "opacity": "0.4"}</field>
```
pedrobaeza, line 4: This shouldn't be changed this way, and I'm doubting right now if this record can be deleted and thus this will lead to problems.

addons/website_blog/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,71 @@
+---Models in module 'website_blog'---
+---Fields in module 'website_blog'---
+website_blog / blog.blog                / content (html)                : NEW
+website_blog / blog.blog                / cover_properties (text)       : NEW hasdefault
+# NOTHING TO DO
```
pedrobaeza, line 5: This is not nothing to do, but we should adapt previous properties for the cover to this new field.

pedrobaeza:
@MiquelRForgeFlow @Yajo I have detected that the migration scripts are not fully correct and it should be fixed. Please check.

yajo:
All concerns solved in https://github.com/OCA/OpenUpgrade/pull/2496

Issue #1879, opened 2019-06-11
end-migration scripts of already migrated modules are not executed after resuming a failing migration


pedrobaeza:
**Impacted versions:** All

**Steps to reproduce:** Migrate a DB, complete several module migrations (with end migration scripts), and fail it.

**Current behavior:** Resuming the migration after fixing the problem, end migration scripts for the already migrated modules are not executed.

**Expected behavior:** They should be.

It's worth to note that Odoo is also contemplating to allow to resume migrations in https://github.com/odoo/odoo/pull/32634, and they have faced also the end-migration problem.

@Tecnativa

pedrobaeza:
@hbrunn @StefanRijnhart what do you think about this?

hbrunn:
I'd wait to see what upstream does and backport that if it's not totally convoluted

StefanRijnhart:
Agree with @hbrunn

PR #2376, opened 2020-08-04
[13.0][MIG] stock_account

MiquelRForgeFlow:
Migration of stock_account.

--


addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+                FROM product_price_history
+                GROUP BY company_id, product_id
+                ORDER BY id
+                ) pph
+            JOIN product_price_history pph2 ON pph.id = pph2.id
+            JOIN product_product pp ON pph.product_id = pp.id
+            JOIN product_template pt ON pp.product_tmpl_id = pt.id
+            WHERE pp.id IN %s""", (tuple(product_ids), ),
+        )
+        # NOTE: It seems to be incomplete (without the link of the stock move)
```
kos94ok-3D: Useless records for Inventory Valuation report without filled `quantity` and `value` fields
Inventory Report
изображение
Inventory Valuation
изображение

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_stock_valuation_layer(env):
+    openupgrade.logged_query(
```
AaronHForgeFlow: Why not doing this in the pre-migration? In any case I prefer to check the column exists just in case we want to attempt to migrate after a fail on a previous try.

pedrobaeza: In any fail try, changes are rolled-back, but don't do that, or you will lose end-migration scripts: #1879

MiquelRForgeFlow: Well, I think he meant for testing purposes, you may want to continue from the failed point than start from zero.

pedrobaeza: But then as said, the whole module changes are rolled-back. Other advice, always run half-migrated DBs with -u all.

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,333 @@
...
+    domain = []
+    if svl_vals.get('stock_move_id'):
+        domain = expression.AND([domain, [('stock_move_id', '=', svl_vals['stock_move_id'])]])
+    if svl_vals.get('old_product_price_history_id'):
+        env.cr.execute("""
+            SELECT create_date
+            FROM product_price_history
+            WHERE id = %s
+        """, (svl_vals['old_product_price_history_id'],))
+        create_date = env.cr.fetchone()[0]
```
kos94ok-3D: Is it need to add `LIMIT 1` here?

MiquelRForgeFlow: I think it doesn't matter in this case, because `fetchone` already limits to one. The `select` is not logged, so also don't care if the output is short or long. But I can add it if you want.

kos94ok-3D: OK, no problem.
I wanted just ask about it.

StefanRijnhart: There is always an index on the `id` column in Odoo, so when selecting a record by that column there is no benefit in adding a LIMIT.

MiquelRForgeFlow:
Ready @pedrobaeza

StefanRijnhart:
What a data model nightmare creating these valuation records. Have you got any statistics about the duration of this process, iterating over every done stock move?

MiquelRForgeFlow:
I have added the `tracking_disable=True` to speed up a bit.

StefanRijnhart:
Is the performance now reasonable? Have you got an indication?

MiquelRForgeFlow:
I will tell you numbers later 😉

MiquelRForgeFlow:
@StefanRijnhart A client bbdd with 258500 stock moves (221880 done moves) -> 7 hours 22 min (duration of this post-migration)

kos94ok-3D:
@MiquelRForgeFlow  Is `product.product` call tracking?

MiquelRForgeFlow:
No, because I disabled tracking as you can see in the code.

kos94ok-3D:
I ask why you added it

MiquelRForgeFlow:
Because the tracking is not needed and can generate side issues.

kos94ok-3D:
Ok, What do you think about use SQL instead of call `_compute_value_svl`.
I call  `_compute_value_svl` because the product doesn`t have in cache correct value when I create SVL by SQL

MiquelRForgeFlow:
I don't know. Maybe doable by putting some explicit flush() calls to avoid the cache issue?

kos94ok-3D:
I tried but don`t work

kos94ok-3D:
@MiquelRForgeFlow It has to save ~30% time:https://github.com/kos94ok-3D/OpenUpgrade/commit/8c5405f15b934a544d8552c1472d4d1020a6415c

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,334 @@
...
+    if len(account_moves) > 1:
+        return env['account.move']
+    return account_moves
+
+
+def _compute_value_svl(env, product, company):
+    # We call this function because the product doesn't have in cache correct value
+    # when we create SVL by SQL
+    # use sql instead of ORM because it saves ~30% time
+    openupgrade.logged_query(env.cr, """
```
kos94ok-3D: Oh, I mistook. We can use `env.cr.execute(` in order to not spam in logs

MiquelRForgeFlow: done

kos94ok-3D:
Hi, @pedrobaeza. Do you have a planned date for review it?

pedrobaeza:
@StefanRijnhart @kos94ok-3D we have added a new algorithm for populating the stock.valuation.layer. It's a lot faster and we get proper results, as previously we didn't get them.

Please tell us if you see it correct.

kos94ok-3D:
I have an unexpected result in my DB which was created especially for test this module.
I need some time for investigating it.
I can attach here DB if needed. It takes 3.5 MB

pedrobaeza:
Well, we need more info to judge.

kos94ok-3D:
For all products I do the same action:
1. Create with price
2.  Update price on product form
3. Inventory initial
4. Receipt product
5. Update price on product form
6. Delivery product

Result with your algorithms
изображение

Result with my algorithims
изображение

pedrobaeza:
Thanks for the extra info. Can you provide the DB as well to check?

kos94ok-3D:
login: andrew
password: 1111
[mig_stock_account_v12.zip](https://github.com/OCA/OpenUpgrade/files/6213226/mig_stock_account_v12.zip)

pedrobaeza:
Thanks, my colleague @carlosdauden will check and tell you back.

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+                if product.cost_method in ("average", "fifo") and not float_is_zero(
+                        previous_qty, precision_digits=precision_uom):
+                    while h_index < len(history_lines):
+                        price_history_rec = history_lines[h_index]
+                        if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
+                            svl_vals = _prepare_man_svl_vals(
+                                price_history_rec, previous_price, previous_qty, company, product)
+                            svl_man_vals_list.append(svl_vals)
+                            previous_price = price_history_rec["cost"]
+                        h_index += 1
```
kos94ok-3D: The first that I found, it should be out the loop:
```suggestion
            # Add manual adjusts after last move
            if product.cost_method in ("average", "fifo") and not float_is_zero(
                    previous_qty, precision_digits=precision_uom):
                while h_index < len(history_lines):
                    price_history_rec = history_lines[h_index]
                    if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
                        svl_vals = _prepare_man_svl_vals(
                            price_history_rec, previous_price, previous_qty, company, product)
                        svl_man_vals_list.append(svl_vals)
                        previous_price = price_history_rec["cost"]
                    h_index += 1
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+                if product.cost_method in ("average", "fifo") and not float_is_zero(
+                        previous_qty, precision_digits=precision_uom):
+                    while h_index < len(history_lines) and history_lines[h_index]["datetime"] < move["date"]:
+                        price_history_rec = history_lines[h_index]
+                        if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
+                            svl_vals = _prepare_man_svl_vals(
+                                price_history_rec, previous_price, previous_qty, company, product)
+                            svl_man_vals_list.append(svl_vals)
+                            previous_price = price_history_rec["cost"]
+                        h_index += 1
```
kos94ok-3D: I do not know why do you not do it for stadard price.
If we create product and change price on product form when do not have qty on stock, we not need create svl but have to set previous price
```suggestion
                # Add manual adjusts
                have_qty = not float_is_zero(previous_qty, precision_digits=precision_uom)
                while h_index < len(history_lines) and history_lines[h_index]["datetime"] < move["date"]:
                    price_history_rec = history_lines[h_index]
                    if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
                        if have_qty:
                            svl_vals = _prepare_man_svl_vals(
                                price_history_rec, previous_price, previous_qty, company, product)
                            svl_man_vals_list.append(svl_vals)
                        previous_price = price_history_rec["cost"]
                    h_index += 1
```
Do the same for above comment

kos94ok-3D (changes requested):
Is not excellent but better

kos94ok-3D:
new result
изображение

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: Of course, that was the idea, that tab was added by mistake at some point. Sorry

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: I agree with your approach, I had added that condition because Odoo uses it (somewhere :sweat_smile:), but without analyzing it in depth because what I urgently needed to solve was the average cost method.

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+
+_logger = logging.getLogger(__name__)
+
+# Declare global variant to avoid that it is passed between methods
+precision_price = 0
+
+
+def _prepare_common_svl_vals(move, product):
+    return {
+        "create_uid": move["write_uid"],
```
kos94ok-3D, line 22: why not move["create_uid"]?

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+        "remaining_value": 0.0,
+    })
+    return vals
+
+
+def _prepare_man_svl_vals(price_history_rec, previous_price, quantity, company, product):
+    diff = price_history_rec["cost"] - previous_price
+    value = float_round(diff * quantity, precision_digits=precision_price)
+    svl_vals = {
+        "create_uid": price_history_rec["write_uid"],
```
kos94ok-3D, line 71: why not price_history_rec["create_uid"]?

kos94ok-3D (commented):
Just question

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+    }
+    return svl_vals
+
+
+def get_product_price_history(env, company_id, product_id):
+    env.cr.execute("""
+        SELECT id, company_id, product_id, datetime, cost, create_uid, write_uid, write_date
+        FROM product_price_history
+        WHERE company_id = %s AND product_id = %s
+        ORDER BY company_id, product_id, datetime
```
kos94ok-3D:
```suggestion
        SELECT id, company_id, product_id, datetime, cost, create_uid, write_uid, write_date
        FROM product_price_history
        WHERE company_id = %s AND product_id = %s
        ORDER BY datetime, id
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+               WHEN sl.usage = 'internal' AND sld.usage <> 'internal' THEN 'out'
+               WHEN sl.usage = 'supplier' AND sld.usage = 'customer' THEN 'dropship'
+               WHEN sl.usage = 'customer' AND sld.usage = 'supplier' THEN 'dropship_return'
+               ELSE 'other'
+          END AS move_type
+        FROM stock_move sm LEFT JOIN stock_location sl ON sl.id = sm.location_id
+            LEFT JOIN stock_location sld ON sld.id = sm.location_dest_id
+        WHERE sm.company_id = %s AND sm.product_id = %s
+            AND state = 'done'
+        ORDER BY sm.company_id, sm.product_id, sm.date, sm.id
```
kos94ok-3D:
```suggestion
        SELECT sm.id, sm.company_id, sm.product_id, sm.date, sm.product_qty, sm.reference,
          COALESCE(sm.price_unit, 0.0) AS price_unit,
          sm.create_uid, sm.create_date, sm.write_uid, sm.write_date,
          CASE WHEN sl.usage <> 'internal' AND sld.usage = 'internal' THEN 'in'
               WHEN sl.usage = 'internal' AND sld.usage <> 'internal' THEN 'out'
               WHEN sl.usage = 'supplier' AND sld.usage = 'customer' THEN 'dropship'
               WHEN sl.usage = 'customer' AND sld.usage = 'supplier' THEN 'dropship_return'
               ELSE 'other'
          END AS move_type
        FROM stock_move sm LEFT JOIN stock_location sl ON sl.id = sm.location_id
            LEFT JOIN stock_location sld ON sld.id = sm.location_dest_id
        WHERE sm.company_id = %s AND sm.product_id = %s
            AND state = 'done'
        ORDER BY sm.date, sm.id
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+                            (previous_price * previous_qty + move["price_unit"] * move["product_qty"]) / total_qty,
+                            precision_digits=precision_price)
+                    svl_vals = _prepare_in_svl_vals(
+                        move, move["product_qty"], move["price_unit"], product, is_dropship)
+                    svl_in_vals_list.append(svl_vals)
+                    previous_qty = total_qty
+                # Add out svl
+                if move["move_type"] == "out" or is_dropship:
+                    qty = move["product_qty"]
+                    if product.cost_method in("average", "fifo") and not is_dropship:
```
kos94ok-3D:
```suggestion
                    if product.cost_method in ("average", "fifo") and not is_dropship:
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+                        svl_vals = _prepare_man_svl_vals(
+                            price_history_rec, previous_price, previous_qty, company, product)
+                        svl_man_vals_list.append(svl_vals)
+                        previous_price = price_history_rec["cost"]
+                    h_index += 1
+            # svl_vals_list = svl_in_vals_list + svl_man_vals_list + svl_out_vals_list
+            # svl_vals_list = sorted(svl_vals_list, key=lambda k: (k["create_date"]))
+            # _logger.info("To create {}".format(len(svl_vals_list)))
+            # if svl_vals_list:
+            #     query_insert(env.cr, "stock_valuation_layer", svl_vals_list)
```
kos94ok-3D:
```suggestion

```

kos94ok-3D (changes requested):
Cosmetic changes

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+                have_qty = not float_is_zero(previous_qty, precision_digits=precision_uom)
+                while h_index < len(history_lines) and history_lines[h_index]["datetime"] < move["date"]:
+                    price_history_rec = history_lines[h_index]
+                    if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
+                        if have_qty:
+                            svl_vals = _prepare_man_svl_vals(
+                                price_history_rec, previous_price, previous_qty, company, product)
+                            svl_man_vals_list.append(svl_vals)
+                        previous_price = price_history_rec["cost"]
+                    h_index += 1
```
kos94ok-3D:
```suggestion
                if product.cost_method in ("average", "standard"):
                    # useless for Fifo because we have price unit in stock.move
                    # Add manual adjusts
                    have_qty = not float_is_zero(previous_qty, precision_digits=precision_uom)
                    while h_index < len(history_lines) and history_lines[h_index]["datetime"] < move["date"]:
                        price_history_rec = history_lines[h_index]
                        if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
                            if have_qty:
                                svl_vals = _prepare_man_svl_vals(
                                    price_history_rec, previous_price, previous_qty, company, product)
                                svl_man_vals_list.append(svl_vals)
                            previous_price = price_history_rec["cost"]
                        h_index += 1
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,221 @@
...
+            if product.cost_method in ("average", "fifo") and not float_is_zero(
+                    previous_qty, precision_digits=precision_uom):
+                while h_index < len(history_lines):
+                    price_history_rec = history_lines[h_index]
+                    if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
+                        svl_vals = _prepare_man_svl_vals(
+                            price_history_rec, previous_price, previous_qty, company, product)
+                        svl_man_vals_list.append(svl_vals)
+                        previous_price = price_history_rec["cost"]
+                    h_index += 1
```
kos94ok-3D:
```suggestion
            if product.cost_method in ("average", "standard") and not float_is_zero(
                    previous_qty, precision_digits=precision_uom):
                # useless for Fifo because we have price unit on product form
                while h_index < len(history_lines):
                    price_history_rec = history_lines[h_index]
                    if float_compare(price_history_rec["cost"], previous_price, precision_digits=precision_price):
                        svl_vals = _prepare_man_svl_vals(
                            price_history_rec, previous_price, previous_qty, company, product)
                        svl_man_vals_list.append(svl_vals)
                        previous_price = price_history_rec["cost"]
                    h_index += 1
```

kos94ok-3D (changes requested):
Product price history not needed for FIFO because we already have the correct price in stock.move

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,224 @@
...
+        ORDER BY sm.company_id, sm.product_id, sm.date, sm.id
+    """, (company_id, product_id))
+    return env.cr.dictfetchall()
+
+
+@openupgrade.logging()
+def generate_stock_valuation_layer(env):
+    company_obj = env["res.company"]
+    product_obj = env["product.product"]
+    precision_price = env["decimal.precision"].precision_get("Product Price")
```
carlosdauden, line 157:
```suggestion
    # Needed to modify global variable
    global precision_price
    precision_price = env["decimal.precision"].precision_get("Product Price")
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: :+1:

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: :+1:

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: The svl record is created when the status of the movement changes to done, so the most common is that the user who validates the movement is the last one to modify the record.

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: Usually the product.price.history records are not modified and create_uid is the same as write_uid, but if it were modified it would normally be to change the price, so I think the correct user is write_uid.

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: :+1:

carlosdauden (approved):

carlosdauden (changes requested):
https://github.com/OCA/OpenUpgrade/pull/2376/files#r602941588

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,224 @@
...
+    return env.cr.dictfetchall()
+
+
+def get_stock_moves(env, company_id, product_id):
+    env.cr.execute("""
+        SELECT sm.id, sm.company_id, sm.product_id, sm.date, sm.product_qty, sm.reference,
+          COALESCE(sm.price_unit, 0.0) AS price_unit,
+          sm.create_uid, sm.create_date, sm.write_uid, sm.write_date,
+          CASE WHEN sl.usage <> 'internal' AND sld.usage = 'internal' THEN 'in'
+               WHEN sl.usage = 'internal' AND sld.usage <> 'internal' THEN 'out'
```
carlosdauden:
```suggestion
          CASE WHEN (sl.usage <> 'internal' AND (sl.usage <> 'transit' OR sl.company_id <> sm.company_id))
                AND (sld.usage = 'internal' OR (sld.usage = 'transit' AND sld.company_id = sm.company_id))
                THEN 'in'
               WHEN (sl.usage = 'internal' OR (sl.usage = 'transit' AND sl.company_id = sm.company_id))
                AND (sld.usage <> 'internal' AND (sld.usage <> 'transit' OR sld.company_id <> sm.company_id))
                THEN 'out'
```

carlosdauden (changes requested):

addons/stock_account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: WTF

MiquelRForgeFlow:
All comments attended

addons/stock_account/migrations/13.0.1.1/post-migration.py
carlosdauden: https://github.com/odoo/odoo/blob/982514b00b6f75760c07ae47eb78d407e3c200c1/addons/stock_account/models/stock_location.py#L25-L32
:sweat_smile:

addons/stock_account/migrations/13.0.1.1/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,79 @@
+---Models in module 'stock_account'---
+new model stock.valuation.layer
+# DONE: post-migration: try to use data from product.price.history in some way (?)
+
+---Fields in module 'stock_account'---
+stock_account / account.move.line        / is_anglo_saxon_line (boolean) : NEW
+# TODO (?): is there a way to know which lines are anglosaxon??
```
MiquelRForgeFlow, line 7: maybe we could check if aml.company_id.anglo_saxon_accounting is true

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,225 @@
...
+    precision_price = env["decimal.precision"].precision_get("Product Price")
+    precision_uom = env["decimal.precision"].precision_get(
+        "Product Unit of Measure"
+    )
+    companies = company_obj.search([])
+    products = product_obj.with_context(active_test=False).search([("type", "in", ("product", "consu"))])
+    all_svl_list = []
+    for product in products:
+        for company in companies:
+            _logger.info("Product:{} {} Company:{}".format(product.id, product.name, company.name))
```
MiquelRForgeFlow: is needed to log this? I have a bbdd with a fucking lot of variants, and this log is spamming

pedrobaeza: Yes, don't log on loops. I'm preparing a progressbar for not needing to know "where I am".

carlosdauden: Ok, remove it.
I added it while I was testing the performance, but at the current speed of the process (milliseconds) it doesn't make sense.

kos94ok-3D (commented):
I think we have to fill it
изображение
It was done in my version:
изображение
And will be good if we link SVL to Price history

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,224 @@
...
+        "create_date": move["date"],
+        "write_uid": move["write_uid"],
+        "write_date": move["date"],
+        "stock_move_id": move["id"],
+        "company_id": move["company_id"],
+        "product_id": move["product_id"],
+        "description": move["reference"] and "%s - %s" % (move["reference"], product.name) or product.name,
+        "unit_cost": 0.0,
+        "remaining_qty": 0.0,
+        "remaining_value": 0.0,
```
carlosdauden, line 33:
```suggestion
        "remaining_value": 0.0,
        "old_product_price_history_id": price_history_rec["id"],
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,224 @@
...
+            all_svl_list.extend(svl_in_vals_list + svl_out_vals_list + svl_man_vals_list)
+    if all_svl_list:
+        all_svl_list = sorted(all_svl_list, key=lambda k: (k["create_date"]))
+        _logger.info("To create {} svl records".format(len(all_svl_list)))
+        query_insert(env.cr, "stock_valuation_layer", all_svl_list)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    generate_stock_valuation_layer(env)
```
carlosdauden, line 252:
```suggestion
    openupgrade.logged_query(
        env.cr, """
        ALTER TABLE stock_valuation_layer
        ADD COLUMN old_product_price_history_id integer""",
    )
    generate_stock_valuation_layer(env)
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: but `_prepare_common_svl_vals` doesn't have `price_history_rec`

AaronHForgeFlow:
@kos94ok-3D How did you manage to get the account move information to introduce it in the layer? I think it would be great if we can preserve that information.

addons/stock_account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: I mean, I am not sure how I should add `price_history_rec` in the `generate_stock_valuation_layer` method for in / out svl cases. Should I use a `while` like is done in the other cases or what?

kos94ok-3D:
> @kos94ok-3D How did you manage to get the account move information to introduce it in the layer? I think it would be great if we can preserve that information.

<pre>
def _get_related_account_move(env, svl_vals):
    """ Return Account move related to Stock Valuation Layer"""
    domain = []
    if svl_vals.get('stock_move_id'):
        domain = expression.AND([domain, [('stock_move_id', '=', svl_vals['stock_move_id'])]])
    if svl_vals.get('old_product_price_history_id'):
        env.cr.execute("""
            SELECT create_date
            FROM product_price_history
            WHERE id = %s
        """, (svl_vals['old_product_price_history_id'],))
        create_date = env.cr.fetchone()[0]
        domain = expression.AND([domain, [('create_date', '=', create_date)]])
    if not domain:
        return env['account.move']
    account_moves = env['account.move'].search(domain)
    if len(account_moves) > 1:
        return env['account.move']
    return account_moves


</pre>

addons/stock_account/migrations/13.0.1.1/post-migration.py
kos94ok-3D: @MiquelRForgeFlow I think he mistook and it has to be in `_prepare_man_svl_vals`

MiquelRForgeFlow: thanks

carlosdauden: Yes sorry, it was an error passing the line from my editor to the comment in the PR

MiquelRForgeFlow:
@kos94ok-3D If we add your account move filling method, it will affect badly the performance or not?

kos94ok-3D:
@MiquelRForgeFlow I think my function needs to rewrite to SQL, I sent it just as an example.

MiquelRForgeFlow:
@kos94ok-3D @AaronHForgeFlow Added the account move information in a neat way :)

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,253 @@
...
+        "write_uid": move["write_uid"],
+        "write_date": move["date"],
+        "stock_move_id": move["id"],
+        "company_id": move["company_id"],
+        "product_id": move["product_id"],
+        "description": move["reference"] and "%s - %s" % (move["reference"], product.name) or product.name,
+        "unit_cost": 0.0,
+        "remaining_qty": 0.0,
+        "remaining_value": 0.0,
+        "account_move_id": move["account_move_id"],
```
kos94ok-3D, line 36:
```suggestion
        "account_move_id": move["account_move_id"],
        "old_product_price_history_id": None,
```

kos94ok-3D (changes requested):
Required because `old_product_price_history_id` not filled if first row does not content it
изображение

addons/stock_account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: Done

kos94ok-3D:
It is ok for me

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -0,0 +1,256 @@
...
+    companies = company_obj.search([])
+    products = product_obj.with_context(active_test=False).search([("type", "in", ("product", "consu"))])
+    all_svl_list = []
+    for product in products:
+        for company in companies:
+            history_lines = get_product_price_history(env, company.id, product.id)
+            moves = get_stock_moves(env, company.id, product.id)
+            svl_in_vals_list = []
+            svl_out_vals_list = []
+            svl_man_vals_list = []
```
AaronHForgeFlow, line 170: Can you explain what does this list of layer contain? for the svl_in_vals_list and svl_out_vals_list is clear the purpose but this one is unclear for me. Thanks in advance!

Edit: Ok I see it is for the manual update of the cost price.  Sorry for asking too soon.

AaronHForgeFlow:
For the fifo case the quantities of the layers is well migrated for the test  I have performed so far. However, for the outgoing stock moves, the value in the layers is not correct. For example in standard Odoo 13:

```
layer_in_1: qty: 10, unit_value: 25, total_value: 250
layer_in_2: qty: 10, unit_value: 15, total_value: 150
```
the layer out takes 10 units from layer_in_1 and 2 units from layer_in_2:
`layer_out_1: qty: 12 unit_value: 23.33 total_value: 280`

In a migrated database:
```
layer_in_1: qty: 10, unit_value: 25, total_value: 250
layer_in_2: qty: 10, unit_value: 15, total_value: 150
```

`layer_out_1: qty: 12 unit_value: 20 total_value: 240`

I will propose the changes that fixes this use case here.

Thanks all for the work.

AaronHForgeFlow (approved):
I think we can merge this now, and then if any issue appears create another PR for that case.

kos94ok-3D:
@AaronHForgeFlow Is it correct?
image

AaronHForgeFlow:
@kos94ok-3D I see, I am checking. Is that the Unit Value or the Total Value?

PR #2379, opened 2020-08-04
[13.0][MIG] stock_dropshipping

MiquelRForgeFlow:
Migration of stock_dropshipping.

--


addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -175,8 +175,9 @@ def fill_stock_picking_type_sequence_code(env):
     openupgrade.logged_query(
         env.cr, """
         UPDATE stock_picking_type
-        SET sequence_code = CASE WHEN code = 'incoming' THEN 'IN'
-          WHEN code = 'outgoing' THEN 'OUT'
+        SET sequence_code = CASE WHEN code = 'outgoing' THEN 'OUT'
+          WHEN code = 'incoming' AND name LIKE '%Dropship%' THEN 'DS'
```
pedrobaeza: Do this on its own module.

MiquelRForgeFlow: But if we do that way, then for the rest of the code in the post-migration of stock and other modules before reaching stock_dropshipping, the sequence_code of that kind of picking type will be wrong, and I am not sure it will generate side effects or not.

pedrobaeza: sequence_code is not used on the rest of the script or code, as you shouldn't generate any new picking AFAIK.

MiquelRForgeFlow: Done

addons/stock_dropshipping/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def update_stock_picking_type_sequence_code_dropship(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE stock_picking_type
+        SET sequence_code = 'DS'
```
pedrobaeza: You should use the existing sequence code instead of hardcoding it, and check another way of finding dropshipping picking types than the name, as you may rename it.

pedrobaeza (commented):
Can you please explain the general strategy here? I see a lot of code with barely indications and I sincerely don't understand why all of this is needed.

addons/stock_dropshipping/migrations/13.0.1.0/post-migration.py
MiquelRForgeFlow: I don't know how to find them in another way. Also, I put 'DS' because that is the one Odoo uses: https://github.com/OCA/OpenUpgrade/blob/13.0/addons/stock_dropshipping/models/res_company.py#L55

MiquelRForgeFlow:
The general strategy here is similar to https://github.com/OCA/OpenUpgrade/blob/13.0/addons/stock/migrations/13.0.1.1/post-migration.py#L234, which it's to generate dropship sequences for each company an adapt existing models that point to default sequence to the new sequence, and similar with dropship picking types and stock rules that are dependent of those sequences.

pedrobaeza:
I have rebased, and added on top these changes:
   
- Don't change current preferences on picking type code/sequence prefix.
- Keep current record for first company and duplicate it for the rest.
- Save mapping of records for later references.
- Use smart foreign key reference replacing for extra created records.

This depends on https://github.com/OCA/openupgradelib/pull/246, but I have pushed an extra commit that should be removed when merged for testing if all goes good. I think this strategy can be used also in `stock` script for simplifying things and covering more cases.
    
@Tecnativa TT28115

addons/stock/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: THIS SHOULD BE FIXED AGAIN :dancers:

pedrobaeza: Yes, we should remove that replacing also from here, as we should respect the user preferences. Can we delay this change until merging this PR, or should we do it all in once?

MiquelRForgeFlow: I think we should do it in the same PR.

pedrobaeza: Done in 7be23b7. Is what you want?

MiquelRForgeFlow: Yes :+1:

pedrobaeza (approved):

PR #2384, opened 2020-08-05
[13.0][MIG] payment_paypal

MiquelRForgeFlow:
Migration of payment_paypal.

--


addons/payment_paypal/migrations/13.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def switch_noupdate_records(env):
+    openupgrade.set_xml_ids_noupdate_value(
```
jcdrubay: In version 12 and 13, both are no update

https://github.com/odoo/odoo/blob/12.0/addons/payment_paypal/views/payment_paypal_templates.xml#L3
https://github.com/odoo/odoo/blob/13.0/addons/payment_paypal/views/payment_paypal_templates.xml#L3

Do we still need this?

pedrobaeza: @MiquelRForgeFlow can you attend this?

MiquelRForgeFlow: Well, the noupdate_analysis says: `ir.ui.view: payment_paypal.paypal_form (noupdate) (noupdate switched)`, so it's obvious why I called the switch method here. It seems true that in both versions this record is noupdate and that the analysis is wrong. I don't know why. But the switch method is doing nothing wrong, it's just a bit superfluous. I don't think there is a strong reason to remove it.

pedrobaeza: The reason is to not do anything not needed for not confusing code readers. You should remove as well the line in `openupgrade_analysis_work.txt`.

MiquelRForgeFlow: Shouldn't it be more interesting in finding why the analysis is wrong?

pedrobaeza: Of course, but for this PR, the actions are the described above.

MiquelRForgeFlow: Done. I just hope I don't have to add this again later.

pedrobaeza (approved):

PR #2387, opened 2020-08-05
[13.0][MIG] membership

MiquelRForgeFlow:
Migration of membership.

--


pedrobaeza:
I have fixed the FK constraint moving things to pre-migration and preserving previous value, so it's good to go.

PR #2388, opened 2020-08-05
[13.0][MIG] sale_timesheet

MiquelRForgeFlow:
Migration of sale_timesheet.

--


pedrobaeza:
Fixed a problem renaming a column that needs its data instead of copying it.

rigo1985:
Hello @pedrobaeza @MiquelRForgeFlow 

I have a problem when performing a migration from o12 to o13 with this module. (sale_timesheet)

The error it gives me is the following:
openupgrade

This happens because when migrating it takes the invoices from the account.invoice of v12 (ex: id = 78) and when passing it to o13 it tries to find it with id 78 in 'account.move' when it is id 9.

account.invoice o12
acc_invoice_odoo12

account.move o13
acc_move_odoo_13

My question is how can I map the ids of the account.invoice v12 to the id account.move of v13

pedrobaeza:
This is handled here: https://github.com/OCA/OpenUpgrade/pull/2388/files#diff-f875757c386c55637585ccb63b0959c8b31d91090cf492bf258e478314e7bb1aR10

so you have a previous inconsistent DB.

PR #612, opened 2016-08-08
[ADD] Migration of sale_stock

StefanRijnhart:

PR #2389, opened 2020-08-06
[9.0][IMP] Correct qty_delivered for phantom bom lines

StefanRijnhart:

### DIFF
--- a/addons/sale_mrp/migrations/9.0.1.0/post-migrate.py
+++ b/addons/sale_mrp/migrations/9.0.1.0/post-migrate.py
@@ -1,6 +1,52 @@
 # -*- coding: utf-8 -*-
 # Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import logging
+from openupgradelib import openupgrade
+from openerp import api, SUPERUSER_ID
 
-# TODO: Set delivered_qty in sale order lines with bom
-# See https://github.com/OCA/OpenUpgrade/pull/1410
+
+def phantom_bom_qty_delivered(env):
+    """ Apply Odoo 12 logic to determine if sale lines with phantom bom
+    products have been delivered: this is the case if there *are*
+    stock moves from non-cancelled pickings, and all of these moves are
+    in 'done' state. """
+    logger = logging.getLogger('sale_mrp.migrations.9.0.1.1')
+    logger.info(
+        'Updating qty delivered on sale lines with phantom bom products')
+    openupgrade.logged_query(
+        env.cr,
+        """
+        WITH lines AS (
+            SELECT DISTINCT(sol.id) FROM sale_order_line sol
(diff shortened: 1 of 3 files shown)


addons/sale_mrp/migrations/9.0.1.0/post-migrate.py
```diff
@@ -1,6 +1,55 @@
...
+    logger.info(
+        'Updating qty delivered on %s sale lines with phantom bom products',
+        len(delivered))
+    sale_lines = env['sale.order.line'].browse(delivered.keys())
+    for sale_line in openupgrade.chunked(sale_lines):
+        qty_delivered = (
+            sale_line.product_uom_qty
+            if delivered[sale_line.id] else 0.0)
+        if sale_line.qty_delivered != qty_delivered:
+            sale_line.qty_delivered = qty_delivered
```
pedrobaeza: Should this be done through ORM?

StefanRijnhart: Yes because it ensures correct values for qty_to_invoice. I see the migration of `sale` in this edition now uses SQL to do that, and I guess we could reuse those queries but for me only a relatively small number of sale lines have phantom bom products so don't see the need to do the extra effort and run the additional risks.

StefanRijnhart: How does that work btw if the migration of sale sets `qty_to_invoice` based on `qty_delivered` and only afterwards `qty_delivered` is set (using SQL) in the migration of sale_stock?

StefanRijnhart: Let's ask the original author :-O https://github.com/OCA/OpenUpgrade/pull/612

StefanRijnhart: I think I assumed that we would trigger a recompute of qty_to_invoice in the deferred step, and it ended up in the sale module. Shall I refactor it like that? I can then refactor the above to use SQL.

pedrobaeza: OK, I see that you are touching that part. Let me know when you finish.

StefanRijnhart: Well, that was solved by a simple bit of copy/paste. The only hypothetical downside is if any 3rd party or custom module depends in its own migration on correct values for sale order line's qty_to_invoice or sale order's invoice_status as they won't have been computed yet. It will be impossible to ensure this does not exist, but it will be rare I think.

StefanRijnhart: Hold on, got to add a check if sale is actually installed.

StefanRijnhart: OK, that's it

legalsylvain (approved):
Code review. No test.

pedrobaeza (approved):
Code review (check the move of the code + new SQL for phantom BoMs)

PR #2391, opened 2020-08-06
[13.0][UPT]l10n_es

AaronHForgeFlow:
l10n_es_account_invoice_sequence introduced different numbering between invoices and account moves using a specific sequence selected in the journal. In v13 invoice number is lost. It is best to keep the sequence of the invoice in this case. We should trust that sequence, as long as all localization reports used that sequence.


MiquelRForgeFlow:
Included in https://github.com/OCA/OpenUpgrade/pull/2419.

Issue #2317, opened 2020-04-23
[12.0->13.0][utm] relation "utm_stage" does not exist


khaeusler:
Hello,

If only utm is installed in a v12 DB, the migration will fail with this error

```
DEBUG KH_5 odoo.modules.loading: loading module utm (18/134) 
INFO KH_5 odoo.modules.migration: module utm: Running migration [>13.0.1.0] pre-migration 
INFO KH_5 OpenUpgrade: utm: pre-migration script called with version 12.0.1.0 
INFO KH_5 odoo.modules.registry: module utm: creating or updating database tables 
DEBUG KH_5 odoo.schema: Table 'utm_medium': dropped index 'utm_medium_name_index' 
DEBUG KH_5 odoo.schema: Table 'utm_medium': dropped index 'utm_medium_active_index' 
DEBUG KH_5 odoo.schema: Table 'utm_medium': dropped index 'utm_medium_create_uid_index' 
DEBUG KH_5 odoo.schema: Table 'utm_medium': dropped index 'utm_medium_create_date_index' 
DEBUG KH_5 odoo.schema: Table 'utm_medium': dropped index 'utm_medium_write_uid_index' 
DEBUG KH_5 odoo.schema: Table 'utm_medium': dropped index 'utm_medium_write_date_index' 
DEBUG KH_5 odoo.schema: Table 'utm_campaign': dropped index 'utm_campaign_name_index' 
DEBUG KH_5 odoo.schema: Table 'utm_campaign': added column 'user_id' of type int4 
DEBUG KH_5 odoo.models: Table 'utm_campaign': setting default value of new column user_id to 1 
DEBUG KH_5 odoo.schema: Table 'utm_campaign': column 'user_id': added constraint NOT NULL 
DEBUG KH_5 odoo.schema: Table 'utm_campaign': dropped index 'utm_campaign_user_id_index' 
DEBUG KH_5 odoo.schema: Table 'utm_campaign': added column 'stage_id' of type int4 
ERROR KH_5 odoo.sql_db: bad query: SELECT "utm_stage".id FROM "utm_stage" ORDER BY "utm_stage"."sequence"   limit 1
ERROR: relation "utm_stage" does not exist
LINE 1: SELECT "utm_stage".id FROM "utm_stage" ORDER BY "utm_stage"....
                                   ^
 
WARNING KH_5 odoo.modules.loading: Transient module states were reset 
ERROR KH_5 odoo.modules.registry: Failed to load registry 
[...]

PR #2392, opened 2020-08-11
[FIX] relation "utm_stage" does not exist [12.0->13.0]

dennybiasiolli:
Ref. #2317

If only utm is installed in a v12 DB, the migration will fail with this error:
`relation "utm_stage" does not exist`

--


pedrobaeza:
There's another opened PR for this: #2372 

Anyway, none of both solutions are my favorite, as it's touching the addon, not the scripts. I think that a creation of this column by SQL in pre will avoid the trigger of the default.

Anyway, closing and let's continue in the other PR.

dennybiasiolli:
Thanks @pedrobaeza , I didn't think about adding the column in pre_migration, it's a good idea. I'll work on this and I will provide a new PR ASAP.

PR #2393, opened 2020-08-12
[FIX] utm: fix utm.campaign model creation

dennybiasiolli:
Before this fix, when trying to upgrade from Odoo 12.0 without
`utm.stage` table created, the creation follows a "top-down" logic,
creating `utm.campaign` with `stage_id` column referencing `utm.stage`
table in `stage_id` field default value:

```
    stage_id = fields.Many2one('utm.stage', string='Stage',
    ondelete='restrict', required=True,
            default=lambda self: self.env['utm.stage'].search([],
            limit=1),
                    group_expand='_group_expand_stage_ids')
```

This commit prevents errors like
`psycopg2.errors.UndefinedTable: relation "utm_stage" does not exist`

Ref #2372

--

### DIFF
--- a/addons/utm/models/utm.py
+++ b/addons/utm/models/utm.py
@@ -13,35 +13,6 @@ class UtmMedium(models.Model):
     name = fields.Char(string='Medium Name', required=True)
     active = fields.Boolean(default=True)
 
-
-class UtmCampaign(models.Model):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Sorry, but we want to avoid such code diff in Odoo code, so better to continue in the other PR.

PR #2399, opened 2020-08-27
[10.0] [FIX] Regression: im_livechat: records in pre migration might not exist

StefanRijnhart:
Records might not exist, especially if im_livechat was not installed
before but was merged into im_odoo_support.

Regression of https://github.com/OCA/openupgradelib/commit/3945526d1

Fixes `ValueError: External ID not found in the system: im_livechat.ir_action_client_open_livechat_menu`

### DIFF
--- a/addons/im_livechat/migrations/10.0.1.0/pre-migration.py
+++ b/addons/im_livechat/migrations/10.0.1.0/pre-migration.py
@@ -11,5 +11,4 @@ xml_ids = [
 
 @openupgrade.migrate()
 def migrate(env, version):
-    for xml_id in xml_ids:
-        env.ref(xml_id).unlink()
+    openupgrade.delete_records_safely_by_xml_id(env, xml_ids)


pedrobaeza:
There's a method for doing this in openupgradelib that won't raise this problem: `delete_records_safely_by_xml_id`

https://github.com/OCA/openupgradelib/blob/3945526d1377533a0aa39d20be88edf2a03f3145/openupgradelib/openupgrade.py#L2408

StefanRijnhart:
Are you telling me I should be using an obscure wrapper method instead of the regular Odoo API.
Sorry, but you broke my migration with a direct commit on openupgradelib. Not in a good mood.

pedrobaeza:
Yes. It's not obscure. It's the way to standardize the deletion of records without errors, and this problem comes from not using it in first place.

StefanRijnhart:
ok I'll change it

pedrobaeza:
Take into account that even without the change in openupgradelib, your DB might have that record removed manually, and thus it will fail the same. Well, there are nuances in this, as if the record is `noupdate=0`, a good practice is to completely update the DB in current version before migrating, so the record will appear again, but using this wrapper prevents to know which is the noupdate status, and this is usually done for noupdate=1 records anyway.

pedrobaeza (approved):

PR #2400, opened 2020-08-28
[DOC] Tip about pip version (no more import error message)

zeroheure:
Explain why there OpenUpgrade can't find libopenupgrade : install libopenupgrade with the right pip version.

--

### DIFF
--- a/odoo/openupgrade/doc/source/migration_details.rst
+++ b/odoo/openupgrade/doc/source/migration_details.rst
@@ -46,6 +46,11 @@ General Tips
 
     $ pip install git+git://github.com/OCA/openupgradelib.git
 
+* You must use the appropriate *pip* version, it is specific to the version of Python that will run OpenUpgrade (for a detailed explanation see https://realpython.com/lessons/why-cant-python-find-my-modules/). Otherwise you will have import error like :
+
+   $ ImportError: No module named <package_name>
+ 
+
 Configuration options
 +++++++++++++++++++++
 
--- a/odoo/openupgrade/doc/source/migration_details.rst
+++ b/odoo/openupgrade/doc/source/migration_details.rst
@@ -44,11 +44,7 @@ General Tips
 
 * When installing the openupgradelib make sure you check out the latest version from github to get the latest updates and fixes::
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
More than that, I would only change upper line putting `pip/pip3` instead of just `pip`

zeroheure:
Yes, I understand the point, but you are a python dev. While I can see in bug reports here and elsewhere that those new to Python _doesn't know_ that pip can be very specific to a version. 

From a sysadmin like me, the web page linked in my PR was a very important explanation to what can happened. Actualy I was facing the problem of import not finding the openupgradelib. I've read many report with answers that never explained what happened, but only how to fix.

pedrobaeza:
Honestly filling this guide with things that belongs to other guides (like "how to install Python libraries through pip") is not the approach to follow. Giving the tip of using `pip` or `pip3`, you can try both and don't make this documentation for what is not intended, as others can ask about including "how to deploy OpenUpgrade", "how to install Ubuntu for putting OpenUpgrade", etc.

Do you get my whole point?

zeroheure:
I got it. I will change the PR.

But again, OpenUpgrade is also used by sysadmin. May be we should add a short paragraph into Introduction saying that a bit of Python and Postgresql knowledge is required to use it ?

pedrobaeza:
OK about adding the section Prerequisites. Anyway, I think pip stuff is also duty of sysadmins. Please also squash the current 2 commits into one with a meaningful commit message.

zeroheure:
My idea was also to help fighting stupid bug reports saying "OpenUpgrade can't find libopenupgrade" ;-)

pedrobaeza:
Yeah, I know, and that's why I originally add the `pip` sentence about using latest GitHub version.

pedrobaeza:
Are you going to make the requested changes?

zeroheure:
Yes I will. I am curently very busy, sorry for the delay.

pedrobaeza:
Directly commited in 776383da1c3d4a6a3a4a39ade77c25bc99334bcb for not having this PR more time opened.

PR #2403, opened 2020-09-15
[12.0][IMP] analytic: ensure analytic line name is not null 

dreispt:


addons/analytic/migrations/12.0.1.1/pre-migration.py
```diff
@@ -20,8 +20,19 @@ def migrate_account_analytic_distribution_pre(cr):
             cr, [('account_analytic_distribution', None)],
         )
 
+def fill_analytic_line_empty_names(cr):
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE account_analytic_line
+        SET name = '(No description)')
```
rvalyi: there is an extra ) at the end of the line

rvalyi (changes requested):

pedrobaeza:
And how can this happen in standard? Isn't this field required in previous version?

pedrobaeza:
My question remains unanswered, as this seems a bad DB state that shouldn't be fixed at OpenUpgrade level

dreispt:
@pedrobaeza You're right. I checked that, and indeed the field is supposed to be required on v11.
@rvalyi Thanks for the review.

PR #2408, opened 2020-10-03
[ADD] base_address_extended: post_init_hook to parse street name correctly

hbrunn:
base_address_extended [autoinstalls](https://github.com/OCA/OCB/blob/11.0/addons/l10n_nl/__manifest__.py#L16) for users of ie l10n_nl, but as country data is loaded only after the fields are computed, street names are parsed wrongly.

I think we also should PR much more countries because the [existing configuration](https://github.com/OCA/OCB/blob/14.0/addons/base_address_extended/data/base_address_extended_data.xml) leaves nearly all countries to the [default](https://github.com/OCA/OCB/blob/14.0/addons/base_address_extended/models/res_partner.py#L69), which is wrong for most countries I'm aware of.

### DIFF
--- a/addons/base_address_extended/__init__.py
+++ b/addons/base_address_extended/__init__.py
@@ -2,3 +2,13 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from . import models
+from odoo import api, SUPERUSER_ID
+
+
+def _post_init_hook(cr, registry):
+    """Recalculate street fields now the countries have their street format set"""
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    partners = env['res.partner'].with_context(active_test=False).search([])
+    for field in ['street_name', 'street_number', 'street_number2']:
+        partners._recompute_todo(partners._fields[field])
+    partners.recompute()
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I don't like to do such things outside migration scripts. I think you should use end-migration of `l10n_nl` or a more general one doing that stuff.

hbrunn:
I'll propose this to upstream, because I realize this is actually a bug in the installation procedure of the module - everyone will get trouble when installing it. Will come to OpenUpgrade then with the next refresh from OCB

pedrobaeza:
OK, I'm afraid there's a problem: 11.0 is no longer supported at Odoo, and we refresh from odoo/odoo, not from OCB, so this is needed anyway.

hbrunn:
I see this has been addressed in [v13](https://github.com/odoo/odoo/blob/13.0/addons/base_address_extended/__init__.py#L10) basically the same way it's done here, but this code will never run for a migration from <11 to something >12, so I suggest to add this here anyways.

It will still be wrong for most countries as of https://github.com/odoo/odoo/issues/28590

pedrobaeza:
But this is not something of OpenUpgrade itself, isn't it? Shouldn't this go on OCB instead?

pedrobaeza:
Forget it, I see the auto-install thing, so let's merge it.

PR #2411, opened 2020-10-06
[13.0][UPD] openupgrade analysis

MiquelRForgeFlow:
Because of https://github.com/OCA/OpenUpgrade/pull/2409.

--


addons/account/migrations/13.0.1.1/openupgrade_analysis.txt
```diff
@@ -182,6 +182,7 @@ account      / account.move             / message_follower_ids (one2many): NEW r
 account      / account.move             / message_ids (one2many)        : NEW relation: mail.message
 account      / account.move             / message_main_attachment_id (many2one): NEW relation: ir.attachment
 account      / account.move             / partner_id (many2one)         : not a function anymore
+account      / account.move             / reversal_move_id (one2many)   : NEW relation: account.move
```
pedrobaeza, line 185: It seems we finally have this field. We should populate it.

addons/base_address_extended/migrations/13.0.1.0/openupgrade_analysis.txt
```diff
@@ -1,4 +1,4 @@
 ---Models in module 'base_address_extended'---
 ---Fields in module 'base_address_extended'---
 ---XML records in module 'base_address_extended'---
----nothing has changed in this module--
+NEW res.country: base.de (noupdate)
```
pedrobaeza, line 4: ?

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
```diff
@@ -58,24 +58,8 @@ DEL account.account.tag: l10n_es.mod_303_36_37
...
-DEL account.fiscal.position.tax.template: l10n_es.fptt_not_subject_tai_ventas_10s
-DEL account.fiscal.position.tax.template: l10n_es.fptt_not_subject_tai_ventas_21b
-DEL account.fiscal.position.tax.template: l10n_es.fptt_not_subject_tai_ventas_21isp
-DEL account.fiscal.position.tax.template: l10n_es.fptt_not_subject_tai_ventas_21s
-DEL account.fiscal.position.tax.template: l10n_es.fptt_not_subject_tai_ventas_4b
-DEL account.fiscal.position.tax.template: l10n_es.fptt_not_subject_tai_ventas_4s
-DEL account.fiscal.position.template: l10n_es.fp_not_subject_tai
-DEL account.tax.template: l10n_es.account_tax_template_p_iva0_ns_b
+NEW account.fiscal.position.template: l10n_es.fp_intra_private
+NEW account.tax.template: l10n_es.account_tax_template_s_iva0_ns
```
pedrobaeza: I think this is not real, but outdated 12.0 branch

addons/sale/migrations/13.0.1.1/openupgrade_analysis.txt
```diff
@@ -77,6 +76,7 @@ DEL ir.ui.view: sale.sale_product_configurator_view_form
 DEL ir.ui.view: sale.variants_tree_view
 DEL ir.ui.view: sale.view_order_form_editable_list
 NEW mail.template: sale.mail_template_sale_confirmation (noupdate)
+NEW res.groups: base.group_multi_currency
```
pedrobaeza, line 79: ?

addons/utm/migrations/13.0.1.0/openupgrade_analysis.txt
```diff
@@ -14,7 +10,11 @@ utm          / mail.mass_mailing.tag    / __last_update (datetime)      : previo
 utm          / mail.mass_mailing.tag    / color (integer)               : previously in module mass_mailing
 utm          / mail.mass_mailing.tag    / display_name (char)           : previously in module mass_mailing
 utm          / mail.mass_mailing.tag    / name (char)                   : previously in module mass_mailing
+utm          / utm.campaign             / color (integer)               : NEW
```
pedrobaeza, line 13: No more `previously in`?

odoo/addons/base/migrations/13.0.1.3/openupgrade_analysis.txt
```diff
@@ -303,6 +302,21 @@ NEW res.country.state: base.state_cn_XJ
 NEW res.country.state: base.state_cn_XZ
 NEW res.country.state: base.state_cn_YN
 NEW res.country.state: base.state_cn_ZJ
+NEW res.country.state: base.state_ee_37
```
pedrobaeza, line 305: Are you sure these are new states?

MiquelRForgeFlow: hey, I used last openupgrade v12 and v13 for real. Thus, what do you want me to do in these comments of yours?

pedrobaeza: OK, just confirm that you are using in both parts the latest version, for not adding changes that are not real (for example, these states may be added both in v12 and v13).

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
pedrobaeza: I'm pretty sure this one is due to an outdated 12.0 branch

addons/base_address_extended/migrations/13.0.1.0/openupgrade_analysis.txt
MiquelRForgeFlow: https://github.com/odoo/odoo/commit/2b674478ad50ca6c6caffb9a6efc69e265571508

pedrobaeza: Yeah, but the question is that this shouldn't be recognized as a change, as the module is not the same

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
MiquelRForgeFlow: Comparing openupgrade v12 with odoo 12.0, I see that there aren't commits https://github.com/odoo/odoo/commit/624cd93f07d40e6b6d1ecaaa5810505eb1c8a6d4 and https://github.com/odoo/odoo/commit/207207d9dc1c2e0c18c6f3b48078314ced8b38cb.

addons/sale/migrations/13.0.1.1/openupgrade_analysis.txt
MiquelRForgeFlow: https://github.com/odoo/odoo/commit/f862f9650395b02f6a43c57cdf232a309fa6948e

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
pedrobaeza: Here you have the same record for v12:

https://github.com/odoo/odoo/blob/12.0/addons/l10n_es/data/account_tax_data.xml#L786

addons/utm/migrations/13.0.1.0/openupgrade_analysis.txt
MiquelRForgeFlow: Yeah, it seems now it failed to recognize they were from mass_mailing module. I think it's due to the apriori renames.

odoo/addons/base/migrations/13.0.1.3/openupgrade_analysis.txt
MiquelRForgeFlow: https://github.com/odoo/odoo/commit/882fa61fd3c3e0e332fb62ecb92a6c4416864321

addons/base_address_extended/migrations/13.0.1.0/openupgrade_analysis.txt
MiquelRForgeFlow: The entry in this analysis means that a new record in the module with xmlid "base.de" is created (or modified in case it already exists in a module dependency). And it has always been this way. If you want it to be different, then propose how would you like to be the entry and then I may study how to implement it.

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
MiquelRForgeFlow: Yeah, but the commit https://github.com/odoo/odoo/commit/624cd93f07d40e6b6d1ecaaa5810505eb1c8a6d4 was added in odoo v12 after the last update upstream of openupgrade v12.

addons/base_address_extended/migrations/13.0.1.0/openupgrade_analysis.txt
pedrobaeza: Is this detected by compare noupdate script? If so, then I propose to change the label `ADD` to something like `UPD` or similar.

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
pedrobaeza: So this is why I'm saying that 12.0 branch (of OpenUpgrade) is outdated... When you are migrating from 12.0 to 13.0, you are not using the outdated OpenUpgrade 12.0 branch for the previous DB version, but Odoo/OCB one.

MiquelRForgeFlow: Yeah, one thing is when you are migrating. But oher thing is to do an analysis. And the analysis is done between openupgrades, always. 😅
And yesterday I didn't do an update upstream for v12 because I did it 20 days ago, which I thought it wouldn't be necessary to update again so recently.

addons/base_address_extended/migrations/13.0.1.0/openupgrade_analysis.txt
MiquelRForgeFlow: Well, I will create an issue, and in the future try to do it.

MiquelRForgeFlow: https://github.com/OCA/OpenUpgrade/issues/2412

addons/l10n_es/migrations/13.0.4.0/openupgrade_analysis.txt
pedrobaeza: But you have removed only one line. At least the other NEW line is also incorrect. Please update 12.0 OpenUpgrade branch or check correctly the changes.

MiquelRForgeFlow: Wait, I searched in odoo 12.0 and I didn't find that fp_intra_private template. So I thought it's really new in v13, thus I didn't delete that line. I am wrong?

pedrobaeza: OK, you are right. I checked bad.

addons/account/migrations/13.0.1.1/openupgrade_analysis.txt
MiquelRForgeFlow: What do you mean by populate? It's a one2many

pedrobaeza: Oh, really, with such name?! Then nothing to do.

PR #2413, opened 2020-10-07
[11.0][UPD] Fix cost_method from default values

ieski:
Fix cost_method from default values

Description of the issue/feature this PR addresses: Replacing "real" with "fifo" in the ir_default table

Current behavior before PR: Error while creating new product because "real" value returns to "fifo"

Desired behavior after PR is merged: It expects the "real" value to change with "fifo" and while the new product is created, the cost_method is waiting to arrive fifo.

--

### DIFF
--- a/addons/stock_account/migrations/11.0.1.1/pre-migration.py
+++ b/addons/stock_account/migrations/11.0.1.1/pre-migration.py
@@ -15,7 +15,8 @@ def update_fifo_cost_method(env):
         SET json_value = '"fifo"'
         WHERE field_id in (SELECT id FROM ir_model_fields 
                          WHERE model = 'product.template' 
-                             AND name = 'cost_method')
+                             AND name = 'cost_method'
+                             AND json_value = '"real"')
         """,
     )
     openupgrade.logged_query(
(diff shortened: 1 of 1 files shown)


addons/stock_account/migrations/11.0.1.1/pre-migration.py
```diff
@@ -9,6 +9,15 @@ def update_fifo_cost_method(env):
     """The real costing method doesn't exist anymore, so the most similar one
     is 'average'. Values mapped according this.
     """
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE ir_default
+        SET json_value = '"fifo"'
+        WHERE field_id in (SELECT id FROM ir_model_fields 
+                         WHERE model = 'product.template' 
+                             AND name = 'cost_method')
```
StefanRijnhart: You are not checking the current default. What if it is set to something else but `"real"`?

Travis noticed trailing whitepace in your query.

ieski: Right thanks

StefanRijnhart (approved):
Thanks!

PR #2414, opened 2020-10-09
[13.0][IMP] website_logo: merged in website

ernestotejeda:
cc @Tecnativa TT25961

The website_logo module of OCA/website is no longer necessary, its functionalities are already included in the odoo website module in V13.0.

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -46,6 +46,7 @@ merged_modules = {
     'web_widget_many2many_tags_multi_selection': 'web',
     # OCA/website
     'website_canonical_url': 'website',
+    'website_logo': 'website',
 }
 
 # only used here for openupgrade_records analysis:


pedrobaeza:
You need to rename field/column of the old module for preserving the logos you had.

ernestotejeda:
The new field for the website logo that is added in the website module is the same as it was in website_logo:

- website_logo: https://github.com/OCA/website/blob/fae710bbeeb65de3b38aef5199f4c96fe4e27bde/website_logo/models/website.py#L7-L7
- website: https://github.com/OCA/OCB/blob/2d77eef588c51b3644d164796f34ef5dbf33b464/addons/website/models/website.py#L87-L87

The one who changed is the one from res.config.settings:
- website_logo: https://github.com/OCA/website/blob/fae710bbeeb65de3b38aef5199f4c96fe4e27bde/website_logo/models/res_config.py#L10-L10
- website: https://github.com/OCA/OCB/blob/2d77eef588c51b3644d164796f34ef5dbf33b464/addons/website/models/res_config_settings.py#L21-L21

I have to do something else anyway?

pedrobaeza:
Great then! Nothing more to do, as res.config.settings is a transient model.

PR #2416, opened 2020-10-15
[11.0] [FIX] Overeager procurement rule expansion

StefanRijnhart:
When expanding what it considers 'global' procurement rules and location
paths without a route, the stock migration incorrectly also expands items
that already did have a route.

I'm also proposing a different treatment of inactive items, instead of
expanding them in a huge amount of just more inactive items.

Also, by not removing the original records from the database we preserve the
referential links from stock moves and procurement orders to the procurement
rules.

--

### DIFF
--- a/addons/stock/migrations/11.0.1.1/pre-migration.py
+++ b/addons/stock/migrations/11.0.1.1/pre-migration.py
@@ -2,21 +2,106 @@
 # Copyright 2017-2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from psycopg2.extensions import AsIs
 from openupgradelib import openupgrade
 
 
-def copy_global_rules(env):
-    """Copy global rules to another table and remove them from main one."""
+def _get_dummy_route_id(env, default):
+    """ Return an inactive dummy route. Create on demand """
+    if default:
+        return default
+    env.cr.execute(
+        """ INSERT INTO stock_location_route
+        (create_date, create_uid, name, active)
+        VALUES(now() at time zone 'UTC', 1,
+        'OpenUpgrade 11.0 Dummy Route for inactive rules & paths',
+        False) RETURNING id """)
+    return env.cr.fetchone()[0]
+
+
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
BTW in our setup application management have deleted many routes they deemed incorrect or outdated. That left a lot of dangling, unused procurement rules. I'll be asking them to set those rules to inactive before the migration, and then this code will prevent the expansion.

addons/stock/migrations/11.0.1.1/pre-migration.py
```diff
@@ -2,21 +2,104 @@
...
+    to assign to inactive records without a route.
+
+    Note that rules and paths are left dangling if you unlink their route.
+    Before starting the migration, check to see which of your 'global'
+    rules need to be set to inactive to prevent the unexpected expansion
+    of such rules and paths.
+    """
+    dummy_route_id = False
+    env.cr.execute(
+        "SELECT id FROM stock_location_route WHERE active ORDER BY id")
```
hbrunn: shouldn't you first order by [`sequence`](https://github.com/OCA/OpenUpgrade/blob/10.0/addons/stock/models/stock_location.py#L118)?

StefanRijnhart: Ah, the ordering is a remainder from an earlier attempt in which I needed to know in the post script which route I assigned in the pre script. It is arbitrary but in the current version I might just as well leave it out (and I might just as well leave it in, which I think I will).

StefanRijnhart: But I will add a limit

StefanRijnhart: Added a limit, removed the ordering after all and added a clarifying comment.

hbrunn (approved):

MiquelRForgeFlow:
@StefanRijnhart @hbrunn could you check https://github.com/OCA/OpenUpgrade/pull/2502? I think my approach it's more readable and cleaner.

StefanRijnhart:
Absolutely!

PR #2418, opened 2020-10-15
[FIX] website_blog: respect name and subtitle

yajo:
This shouldn't be included as a noupdate change. It should be respected as it was, instead.

This requires merging https://github.com/OCA/OpenUpgrade/pull/2374 to be actually useful.

@Tecnativa TT23136


MiquelRForgeFlow:
Included in https://github.com/OCA/OpenUpgrade/pull/2374.

PR #2419, opened 2020-10-19
[13.0][MIG] l10n_es

MiquelRForgeFlow:
Migration of l10n_es.

--


odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -37,6 +37,7 @@
     # OCA/account-reconcile
     'account_set_reconcilable': 'account',
     # OCA/l10n-spain
+    'l10n_es_account_invoice_sequence': 'account',
```
MiquelRForgeFlow: delete this, as is already merged in l10n_es (repeated key)

addons/l10n_es/migrations/13.0.4.0/pre-migration.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def set_account_move_number_to_invoice_number(env):
+    """l10n_es_account_invoice_sequence introduced different numbering between
+    invoices and account moves.
+
```
MiquelRForgeFlow:
```suggestion

```

odoo/addons/openupgrade_records/lib/apriori.py
pedrobaeza: I'm seeing, thanks.

MiquelRForgeFlow (commented):
:+1:

PR #2420, opened 2020-10-19
[FIX] point_of_sale : populate tax_ids if pos_pricelist was installed

legalsylvain:
when migrating from 8.0 to 9.0 (or more), taxes on each lines are lost

Current behavior before PR:
Lost data

Data correctly populated. (only if pos_pricelist was installed)

--

### DIFF
--- a/addons/point_of_sale/migrations/9.0.1.0.1/pre-migration.py
+++ b/addons/point_of_sale/migrations/9.0.1.0.1/pre-migration.py
@@ -33,6 +33,27 @@ column_renames = {
 }
 
 
+def populate_pos_order_tax_ids(cr):
+    if not openupgrade.is_module_installed(cr, "pos_pricelist"):
+        return
+
+    openupgrade.logged_query(
+        cr,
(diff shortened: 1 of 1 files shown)


addons/point_of_sale/migrations/9.0.1.0.1/pre-migration.py
```diff
@@ -33,6 +38,30 @@
 }
 
 
+def populate_pos_order_tax_ids(cr):
+    if not openupgrade.is_module_installed(cr, "pos_pricelist"):
+        _logger.error(
```
pedrobaeza: This is not an error condition, so please don't log as such. And even more, no there's a return, so no table is populated at all. I think you should simply `return`.

legalsylvain: Thanks for your review. I saw it as an error, because having pos_pricelist installed, I was loosing all my taxes. But in a regular usage, it is not the case.

PR updated.

pedrobaeza (approved):

PR #2421, opened 2020-10-20
[ADD] base_suspend_security: Merged on base

joao-p-marques:
Functionality now available through `with_user()`

See https://github.com/OCA/server-backend/issues/44

TT25937

--


pedrobaeza:
Just a note: it's not through `with_user`, but with `sudo`, which doesn't alter the user, only the permissions.

joao-p-marques:
Ok, thank you.

I was reading the code and thought this would be it.

PR #2432, opened 2020-11-03
[IMP] hr_skills: Rename hr_skill (OCA) to hr_skills (Odoo)

victoralmau:
Rename addon `hr_skills` (OCA) to `hr_skill` (Odoo)

Please @pedrobaeza review it

@Tecnativa TT26586

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -7,7 +7,9 @@ renamed_modules = {
     'crm_reveal': 'crm_iap_lead',
     'document': 'attachment_indexation',
     'payment_ogone': 'payment_ingenico',
-    # OCA/...
+    # OCA/hr
+    # TODO: Transform possible data
+    'hr_skill': 'hr_skills'
 }
 
 merged_modules = {


odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -7,7 +7,8 @@
     'crm_reveal': 'crm_iap_lead',
     'document': 'attachment_indexation',
     'payment_ogone': 'payment_ingenico',
-    # OCA/...
+    # OCA/hr
+    'hr_skill': 'hr_skills'
```
MiquelRForgeFlow, line 12:
```suggestion
    'hr_skills': 'hr_skill',
```
according to what your PR and commit message say

pedrobaeza: It's the contrary... so the title should be incorrect.

pedrobaeza, line 12: This is missing the commented TODO:
```suggestion
    # TODO: Transform possible data
    'hr_skill': 'hr_skills'
```

MiquelRForgeFlow:
could you also edit commit message?

PR #2439, opened 2020-11-03
[13.0][MIG] hr_expense

MiquelRForgeFlow:
Migration of hr_expense.

--


pedrobaeza (approved):
Done the following changes:

- The company of the user is not a reliable source for filling missing company in the expenses, as the user may switch the current company, and a user may create expenses for other employees. Use employee instead.
- Don't rename ir.rule XML-IDs, but remove existing one and let Odoo to create the new ones, as the content is not exactly the same, so this better than rename + update the changes.
- Try to remove product. Although it's present in sale_expense, it's as demo data, so for regular DBs, if not used, that product shouldn't be present.

PR #2441, opened 2020-11-06
[13.0][FIX] account: don't update date on nondraft move lines

MiquelRForgeFlow:
If not, some things could change fiscal period, and we need to avoid that.

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -201,7 +201,7 @@ def migration_invoice_moves(env):
         SET exclude_from_invoice_tab = FALSE, sequence = ail.sequence,
         price_unit = ail.price_unit, discount = ail.discount, price_subtotal = ail.price_subtotal,
         price_total = ail.price_total, display_type = ail.display_type,
-        is_rounding_line = ail.is_rounding_line, old_invoice_line_id = ail.id, date = ai.date_invoice,
+        is_rounding_line = ail.is_rounding_line, old_invoice_line_id = ail.id,
         create_uid = ail.create_uid, create_date = ail.create_date
         FROM account_invoice_line ail
             JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
@@ -223,7 +223,7 @@ def migration_invoice_moves(env):
             env.cr, """
             UPDATE account_move_line aml
             SET exclude_from_invoice_tab = TRUE, sequence = ail.sequence,
-            name = '(OLD GROUPED ITEM)' || aml.name, date = ai.date_invoice,
+            name = '(OLD GROUPED ITEM)' || aml.name,
             create_uid = ail.create_uid, create_date = ail.create_date
             FROM account_invoice_line ail
                 JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -288,7 +288,7 @@ def migration_invoice_moves(env):
                 THEN ail.currency_id ELSE NULL END, ail.partner_id, ail.uom_id,
                 ail.product_id, ail.account_analytic_id, ail.display_type,
                 ail.is_rounding_line, COALESCE(ai.move_id, am.id), ail.id, ai.date_invoice,
-                ail.create_uid, ail.create_date, ail.write_uid, ail.write_date
+                COALESCE(ai.date, ail.create_uid), ail.create_date, ail.write_uid, ail.write_date
```
pedrobaeza: The coalesce should be in the previous field, not here

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -324,7 +324,7 @@ def migration_invoice_moves(env):
         THEN ail.currency_id ELSE NULL END, ail.partner_id, ail.uom_id,
         ail.product_id, ail.account_analytic_id, ail.display_type,
         ail.is_rounding_line, COALESCE(ai.move_id, am.id), ail.id, ai.date_invoice,
-        ail.create_uid, ail.create_date, ail.write_uid, ail.write_date
+        COALESCE(ai.date, ail.create_uid), ail.create_date, ail.write_uid, ail.write_date
```
pedrobaeza: The same

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -360,7 +360,7 @@ def migration_invoice_moves(env):
         ait.amount, ait.currency_id, ait.base, ait.tax_id,
         ait.account_analytic_id, COALESCE(ai.move_id, am.id),
         ait.id, TRUE, am.state, 1.0, ai.commercial_partner_id, ai.date_invoice,
-        ait.create_uid, ait.create_date, ait.write_uid, ait.write_date
+        COALESCE(ai.date, ait.create_uid), ait.create_date, ait.write_uid, ait.write_date
```
pedrobaeza: The same

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -497,11 +497,11 @@ def migration_voucher_moves(env):
         INSERT INTO account_move_line (company_id, account_id,
         sequence, name, quantity, price_unit, price_subtotal,
         product_id, analytic_account_id, move_id, old_voucher_line_id,
-        create_uid, create_date, write_uid, write_date)
+        date, create_uid, create_date, write_uid, write_date)
         SELECT avl.company_id, avl.account_id, avl.sequence, avl.name,
         avl.quantity, avl.price_unit, avl.price_subtotal, avl.product_id,
         avl.account_analytic_id, COALESCE(av.move_id, am.id), avl.id,
-        avl.create_uid, avl.create_date, avl.write_uid, avl.write_date
+        COALESCE(am.date, avl.create_uid), avl.create_uid, avl.create_date, avl.write_uid, avl.write_date
```
pedrobaeza: The same

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: Yes, sorry, my mistake, for going too fast.

pedrobaeza (approved):

PR #2444, opened 2020-11-18
[13.0][FIX+IMP] account: Several fixes

pedrobaeza:
* [FIX] When invoicing to a contact, the partner in journal items is the commercial partner.
* [IMP] Perform 2 rounds with different criteria for improving the matching.
* [FIX] Reverse signed amounts for vendor bills/refunds. This has changed from v12 to v13.

@Tecnativa

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -147,9 +147,16 @@ def migration_invoice_moves(env):
         ai.reference, ai.comment, ai.type, ai.journal_id, ai.company_id,
         ai.currency_id, ai.partner_id, ai.commercial_partner_id,
         ai.amount_untaxed, ai.amount_tax, ai.amount_total, ai.residual,
-        ai.amount_untaxed_signed, ai.amount_tax_company_signed, COALESCE(
-        ai.amount_total_company_signed, ai.amount_total_signed),
-        COALESCE(ai.residual_company_signed, ai.residual_signed),
+        CASE WHEN ai.type IN ('in_invoice', 'in_refund')
+        THEN -ai.amount_untaxed_signed ELSE ai.amount_untaxed_signed END,
+        CASE WHEN ai.type IN ('in_invoice', 'in_refund')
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I have decided to move the name comparison also to the stricter match, as most of the lines will be matched with this one, and for the rest, there's still the risk that users modify this field, as it's one of the visible ones in journal items view, so it doesn't bring too much value, and it's better to have multiple match than none if ever.

pedrobaeza:
Still WIP though...

pedrobaeza:
Definitively removed the name comparison in both matches, as the generated name in the journal items has changed across Odoo versions and thus, there's no match for old lines. This one is ready to be reviewed.

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -206,17 +221,30 @@ def migration_invoice_moves(env):
         FROM account_invoice_line ail
             JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
             JOIN account_move am ON ail.invoice_id = am.old_invoice_id
-        WHERE am.id = aml.move_id AND ail.company_id = aml.company_id AND ail.account_id = aml.account_id
-            AND ail.partner_id = aml.partner_id
-            AND ail.name = aml.name
```
MiquelRForgeFlow, line 211: what happened to this name condition?

MiquelRForgeFlow (approved):

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow, line 211: Maybe you could change it to three criterias instead of two, and putting the name in the more strictly one?

pedrobaeza:
No, as said, the name is different in journal items across Odoo versions, so it will fail a lot in the stricter matching, and thus, there's more risk to incorrect matching if we let all these items are matched in the second pass. I'm going to merge and let's try this approach. Tried in 2 DBs, this is working well.

PR #2447, opened 2020-11-23
[13.0][account][fix] Improve the mapping of invoice line data to move lines.

JordiBForgeFlow:
Do not map tax related lines or receivable/payable line to invoice lines.

As a result of this:
- Before this fix the tax_ids was assigned to all the move lines relating to an invoice, resulting in incorrect tax reporting results. Now the tax_ids is assigned only to the move lines that is really applicable.

cc @MiquelRForgeFlow 

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -222,8 +222,15 @@ def migration_invoice_moves(env):
             JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
             JOIN account_move am ON ail.invoice_id = am.old_invoice_id
         """
+    # We assign to the move lines information from the matching invoice line.
+    # Everything that has a tax_line_id (originator tax) is by definition originating
+    # from the invoice taxes, not from invoice lines.
+    # The move lines that have a receivable or payable account (the same as the one
+    # from the invoice) are not associated to any invoice line.
     minimal_where = """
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -222,8 +222,15 @@ def migration_invoice_moves(env):
...
         """
+    # We assign to the move lines information from the matching invoice line.
+    # Everything that has a tax_line_id (originator tax) is by definition originating
+    # from the invoice taxes, not from invoice lines.
+    # The move lines that have a receivable or payable account (the same as the one
+    # from the invoice) are not associated to any invoice line.
     minimal_where = """
         WHERE am.id = aml.move_id
+            AND aml.tax_line_id IS NULL
+            AND aml.account_id <> ai.account_id
```
MiquelRForgeFlow, line 233: why this line?

JordiBForgeFlow: Because there's one move line for the receivable/payable that is not linked to a particular invoice line, that represents the total to pay/receive for this vendor/customer.

JordiBForgeFlow:
Tested after real database migration. All the tax reports now appear correct, calculating the right base and tax amounts.
The Balance Sheet and P&L are right as well.

pedrobaeza (approved):
All the problems come from not using products in invoice lines, but it's a legit case, so I agree with these constraints for avoiding "over-matching"

PR #2448, opened 2020-11-23
[IMP] Change Travis by Github Actions

etobella:

Delete Travis and use Github Actions

Current behavior before PR:

No CI was executed

Github Actions CI is executed

@pedrobaeza 

--


etobella:
image
It must be checked if actions are enabled for this repository or for OCA organization

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,40 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: OpenUpgrade documentation
+
+on:
+  push:
+    branches: ["11.0", "12.0", "13.0"]
```
pedrobaeza: Why these branches if we are pushing this to a specific branch?

pedrobaeza:
I can't enable that as I don't have access to Settings page in this project.

legalsylvain:
Hi @etobella 

Thanks a lot to investigate that way ! 
Don't you think it should be better a global talk, regarding the OCA ? we are facing issue with Travis currently. (with a new limit of 10.000 minutes) I think that the OCA have to decide some things regarding CI, but I prefer to have a global analysis, instead of having local manual switches in some repositories.
Don't you think ?

Ref : 

https://odoo-community.org/groups/contributors-15/contributors-178949?mode=thread&date_begin=&date_end=

Last message from @sbidoul : 

> The board has decided to activate the $249/month Travis plan, which gives us 5 concurrent builds.

> We will be evaluating options in the coming weeks

> In the meantime, Travis is back and should work as before.

pedrobaeza:
@legalsylvain this repository is very specific and have a totally different Travis setup and test workflow, so this can be switched (and in fact, we have been talking about this since a bit), so it's normal that we start doing the switch without a global discussion.

sbidoul:
I agree this repo can move to GitHub actions independently of the regular addons repos.

GitHub actions are enabled globally and for this repo so there must be another reason why it does not run.

etobella:
Thanks @sbidoul I will check why this is not executed

etobella:
This is weird, because with the same configuration I was able to use it properly on my repo:

https://github.com/tegin/OpenUpgrade/pull/2

Here the actions are not being generated

.github/workflows/flake.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Flake8 Script
+        run: |
+          flake8 odoo/openupgrade --max-line-length=120
+          flake8 odoo/addons/openupgrade* --max-line-length=120 --filename=__init__.py --ignore=F401
+          flake8 odoo/addons/openupgrade* --max-line-length=120 --exclude=__init__.py
+          # only flake8 migration scripts from the openupgrade project, presumably
+          # identifiable by using the openupgrade helpers
+          flake8 --max-line-length=120 scripts $(find . \( -name 'pre-*.py' -or -name 'post-*.py' \) -exec grep -q openupgrade {} \; -print)
+          flake8 odoo/addons/*/migrations/*/tests/ --max-line-length=120
+          flake8 addons/*/migrations/*/tests/ --max-line-length=120
```
MiquelRForgeFlow: add ending blank line

pedrobaeza:
My suspect is that this is not in the main branch, so it's not being triggered. I'm going to merge and if not, we can revert.

pedrobaeza:
And here we have it: https://github.com/OCA/OpenUpgrade/actions

pedrobaeza:
Testing if it's working properly in #2449

PR #2455, opened 2020-11-30
[13.0][FIX] utm: avoid foreign_key error in utm_tag_rel table

MiquelRForgeFlow:
To avoid violating obsolete constraint, we delete the constrains. Thus, we avoid doing the rename of the table and instead we refill it post-migration the new table.

--

### DIFF
--- a/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
@@ -556,6 +556,7 @@ _obsolete_tables = (
     "account_voucher",
     "account_voucher_line",
     "lunch_order_line",
+    "mail_mass_mailing_campaign"
     "slide_category",
     "survey_page",
     "account_analytic_tag_account_invoice_line_rel",
@@ -567,6 +568,7 @@ _obsolete_tables = (
     "sale_order_line_invoice_rel",
     "summary_dept_rel",
     "project_task_assign_so_line_rel",
+    "mail_mass_mailing_tag_rel",
 )
 
 
--- a/addons/utm/migrations/13.0.1.0/post-migration.py
+++ b/addons/utm/migrations/13.0.1.0/post-migration.py
@@ -1,0 +1,19 @@
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def update_utm_tag_rel(env):
+    openupgrade.logged_query(
+        env.cr, """
+        INSERT INTO utm_tag_rel (tag_id, campaign_id)
+        SELECT rel.tag_id, mmc.campaign_id
+        FROM mail_mass_mailing_tag_rel rel
+        JOIN mail_mass_mailing_campaign mmc ON mmc.id = rel.campaign_id"""
+    )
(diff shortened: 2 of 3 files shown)


HviorForgeFlow (approved):

cubells (approved):
:+1: 

LGTM

Code review

pedrobaeza:
And how this helps on the issue?

MiquelRForgeFlow:
Well, if the foreign_key contraint of that table is deleted, then there won't be any error in the update query of the utm pre-migration.

MiquelRForgeFlow:
Ah, wait.

pedrobaeza:
OK, this is a bit obscured...

But the links are updated on pre: https://github.com/OCA/OpenUpgrade/blob/b2bb2d85a07e9fd2608192fb5fd583f632829a4d/addons/utm/migrations/13.0.1.0/pre-migration.py#L40

so this shouldn't happen.

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
```diff
@@ -567,6 +568,7 @@
     "sale_order_line_invoice_rel",
     "summary_dept_rel",
     "project_task_assign_so_line_rel",
+    "mail_mass_mailing_tag_rel",
```
pedrobaeza, line 571: Please put this on utm, not here.

MiquelRForgeFlow: why? I mean, it's the same. These are obsolete tables.

MiquelRForgeFlow: Otherwise, I should move every table in that dictionary to each module.

pedrobaeza: Well, it's not the same that the model disappears than this is a functional change belonging to that module. Anyway, I would prefer if it's the case to have all the FK dropping in its respective module if proceeds.

MiquelRForgeFlow: But the model in this case also disappears. There is no functional change. The model mail.mass.mailing.campaign is not renamed and thus becomes obsolete, the table mail_mass_mailing_campaign becomes obsolete and the mail_mass_mailing_tag_rel the same. Don't see the difference.

MiquelRForgeFlow: > to have all the FK dropping in its respective module if proceeds.

In this PR or in another?

pedrobaeza: But if the model disappears, and we don't change anything, why do we need to drop FK constraints?

MiquelRForgeFlow: Quoting @Yajo in two cases:

A) "to avoid unknown cascade behaviors to slow down or break the DB structure."
B) "in a DB that was migrated from v10 and that had many procurement orders takes days. That's because the FK was set without an index."

pedrobaeza: OK, I remember this now, thanks. Maybe a code comment can serve for refreshing this. Let's keep it here for not moving more.

pedrobaeza (approved):

PR #2456, opened 2020-11-30
[12.0][FIX] stock: company_id of stock rule should be the same as the warehouse

MiquelRForgeFlow:
This is not literally an "issue" in v12, but it's good to have data correctness.

--

### DIFF
--- a/addons/stock/migrations/12.0.1.1/post-migration.py
+++ b/addons/stock/migrations/12.0.1.1/post-migration.py
@@ -112,6 +112,16 @@ def merge_stock_location_path_stock_rule(env):
         )
 
 
+def assure_stock_rule_company_is_correct(env):
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE stock_rule sr
+        SET company_id = sw.company_id
+        FROM stock_warehouse sw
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In which cases this is not happening? Is due to bad manual configuration or migrated DB or migrated DB with current script-set?

pedrobaeza:
Anyway, please change it only when it's different (and take care about the comparison with NULL).

MiquelRForgeFlow:
> In which cases this is not happening? Is due to bad manual configuration or migrated DB or migrated DB with current script-set?

I just found migrating a DB, and found check in v13. I think the problem is due to company_id having the default=lambda self: self.env.user.company_id. Thus, if rules are created with sudo or during migration, then they will have the company of the admin.

MiquelRForgeFlow:
> and take care about the comparison with NULL

Is not enough adding "`AND sw.company_id != sr.company_id`"?

pedrobaeza:
It seems not all comparisons with possible values are working well with NULL.

MiquelRForgeFlow:
Closing in favor of https://github.com/OCA/OpenUpgrade/pull/2456 due to wrong remote.

PR #2457, opened 2020-11-30
[12.0][FIX] stock: company_id of stock rule should be the same as the warehouse

MiquelRForgeFlow:
This is not literally an "issue" in v12, but it's good to have data correctness.

I just found migrating a DB, and found check in v13. I think the problem is due to company_id having the default=lambda self: self.env.user.company_id. Thus, if rules are created with sudo or during migration, then they will have the company of the admin.

--

### DIFF
--- a/addons/stock/migrations/12.0.1.1/post-migration.py
+++ b/addons/stock/migrations/12.0.1.1/post-migration.py
@@ -112,6 +112,16 @@ def merge_stock_location_path_stock_rule(env):
         )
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

MiquelRForgeFlow:
I added it in stock pre-migration v13, thus maybe it's not needed here.

pedrobaeza:
Then do we close this?

PR #2464, opened 2020-12-03
[13.0][UPD] noupdate_changes.xml files up to day

MiquelRForgeFlow:
Updating.

--


addons/account/migrations/13.0.1.1/noupdate_changes.xml
```diff
@@ -35,6 +35,9 @@
   <record id="account_reconcile_model_template_comp_rule" model="ir.rule">
     <field name="domain_force">['|',('company_id','=',False),('company_id', 'in', company_ids)]</field>
   </record>
+  <record id="action_open_settings" model="ir.actions.act_window">
```
pedrobaeza, line 38: Is this really a noupdate relevant thing?

MiquelRForgeFlow: Relevant or not, it appears. I can add commented if you prefer.

pedrobaeza: Well, I'm just surprised that's noupdate. Let's keep it.

pedrobaeza (approved):
Thanks for watching on these things that I don't have into account

PR #2470, opened 2020-12-04
[12.0][FIX] stock: consider also inactive warehouses

MiquelRForgeFlow:
If not created picking types, then in v13 we can cry.

--


pedrobaeza (approved):
:sob:

JordiBForgeFlow (approved):

PR #2471, opened 2020-12-07
[FIX] mrp + openupgrade_records : do not create column when installing mrp

legalsylvain:
due to recent commit in odoo 13.0 (https://github.com/odoo/odoo/commit/549de5601efdad3ebdb03feb8beb1e898f80113a)
 that has been recently introduced in OpenUpgrade (probably [here](https://github.com/OCA/OpenUpgrade/pull/2460)), the use of openugrade_records is not working any more, because odoo fails to recreate existing column.

this trivial patch fixes that bug.

CC : @StefanRijnhart 

--

### DIFF
--- a/addons/mrp/__init__.py
+++ b/addons/mrp/__init__.py
@@ -14,16 +14,19 @@ def _pre_init_mrp(cr):
           leads to "Out of Memory" crashes
         - stock.move.line.done_move is a stored+related on the former... 
         - Also set the default value for unit_factor in the same UPDATE query to save some SQL constraint checks"""
-    cr.execute("""ALTER TABLE "stock_move" ADD COLUMN "unit_factor" float;""")
-    cr.execute("""ALTER TABLE "stock_move" ADD COLUMN "is_done" bool;""")
-    cr.execute("""ALTER TABLE "stock_move_line" ADD COLUMN "done_move" bool;""")
-    cr.execute("""UPDATE stock_move
-                     SET is_done=COALESCE(state in ('done', 'cancel'), FALSE),
-                         unit_factor=1.0;""")
-    cr.execute("""UPDATE stock_move_line
-                     SET done_move=sm.is_done
-                    FROM stock_move sm
-                   WHERE move_id=sm.id;""")
+    # Openupgrade: the following lines will generate an errors, when running
+    # openupgrade_records.
+    # cr.execute("""ALTER TABLE "stock_move" ADD COLUMN "unit_factor" float;""")
+    # cr.execute("""ALTER TABLE "stock_move" ADD COLUMN "is_done" bool;""")
+    # cr.execute("""ALTER TABLE "stock_move_line" ADD COLUMN "done_move" bool;""")
+    # cr.execute("""UPDATE stock_move
+    #                  SET is_done=COALESCE(state in ('done', 'cancel'), FALSE),
+    #                      unit_factor=1.0;""")
+    # cr.execute("""UPDATE stock_move_line
+    #                  SET done_move=sm.is_done
+    #                 FROM stock_move sm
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In general, we should avoid this kind of patches, and more with the new framework that doesn't include the addons. Isn't there any other option?

legalsylvain:
Hi @pedrobaeza : thanks for your review.

> In general, we should avoid this kind of patches, and more with the new framework that doesn't include the addons

We are talking about the 13.0 branch. So this is the "old framework".

> Isn't there any other option?

I don't know. I only saw that there were other similar patches in openupgrade, for the time being. when I cherry picked the diff in openupgrade to generate sandalone upgrade_analysis modules, I saw that patches : (i didn't introduced) :

https://github.com/legalsylvain/server-tools/commit/c37e442d73b5ab24e3363bf0a0b73467e87f703c#diff-b409a07ffc8468a0d270617f999d45cddb5e7b52d560ca2492e82b878ee85d59

pedrobaeza:
Yeah, I have never liked that kind of patches, as they are very prone to be erroneous when Odoo changes upstream, so that's why I say if we are able to patch that at analysis level, not hardcoded one, in advance where we have v14>v15 with pure framework without the rest of the source code. Of course monkey patches can be done as the one that you show me, but that should be the last resort.

legalsylvain:
Hi. well, I just restored a patch that has been recently erased. Indeed, could be done in ``openupgrade_records`` module I guess, like this ``resource`` patch : https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/openupgrade_records/models/generate_records_wizard.py#L20, droppping column, if exists.

I have no time for that, but I'll review your PR, if needed.

pedrobaeza:
I think @StefanRijnhart removed it as not needed anymore, but maybe he wasn't taken into account this analysis. I'm not able as well to do such "alternative patch" now.

StefanRijnhart:
Sorry, I'm not aware I removed anything in OpenUpgrade 13. It's true this pre_init_script breaks the analysis. I am not sure what is wrong with changing the code in place as is done in this PR for this edition given that it's clearly marked as an OpenUpgrade specific change.

pedrobaeza:
The potential problem is that when updating upstream (or in v14, if we monkey-patch these things), that we are not aware of a possible bug that we are introducing. For example, in this case, further code can be added upstream at the end of the hook, and that code is needed and make things to go wrong. I know it's a remote possibility, but I have already experimented that remote possibilities, and they are a headache when happening.

addons/mrp/__init__.py
```diff
@@ -14,6 +14,9 @@ def _pre_init_mrp(cr):
           leads to "Out of Memory" crashes
         - stock.move.line.done_move is a stored+related on the former... 
         - Also set the default value for unit_factor in the same UPDATE query to save some SQL constraint checks"""
+    # Openupgrade: the following lines will generate an errors, when running
+    # openupgrade_records.
+    return
```
pedrobaeza: Can you comment conflicting instruction instead of putting a return?

legalsylvain:
Hi @pedrobaeza. well new design should reduce the problem. 
- we won't risk crushing the patches anymore, updating upstream 
- as the patches are now splitted in two distinct projects, patches will only impact ``upgrade_analysis`` module, and not the openupgrade regular process. 

However, you're right and current mrp patch could be changed into another function like [this one](https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/openupgrade_records/models/generate_records_wizard.py#L20). 
I'll add it in the roadmap file of the ``upgrade_analysis`` module.

pedrobaeza:
Thanks!

PR #2472, opened 2020-12-07
[13.0][ADD] procurement_jit: Nothing to do

pedrobaeza:
@Tecnativa TT21638


pedrobaeza:
Self-merging as trivial

kos94ok-3D:
Is it not a problem that it depends on sale_stock?

pedrobaeza:
Not right now, although you are right that I haven't checked that stuff. This is installing all the dependency chain, but nothing blocks the CI. I will go with such dependency chain this week anyway. Thanks for noticing.

PR #2473, opened 2020-12-07
[13.0][ADD] google_calendar: Nothing to do

pedrobaeza:
@Tecnativa TT19739


pedrobaeza:
Self-merging as well as nothing to do (and this time the dependency chain is assured, hehe).

PR #2474, opened 2020-12-07
[13.0][ADD] product_margin: Nothing to do

pedrobaeza:
@Tecnativa TT25943


pedrobaeza:
Other trivial one merging.

PR #2480, opened 2020-12-11
[12.0][FIX] account: reference_type is used in account_payment_order module

MiquelRForgeFlow:
If not, reference_type defined in account_payment_order become empty after migration. See https://github.com/OCA/bank-payment/blob/12.0/account_payment_order/models/account_invoice.py#L19

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/pre-migration.py
+++ b/addons/account/migrations/12.0.1.1/pre-migration.py
@@ -10,6 +10,9 @@ _column_copies = {
     'account_payment_term_line': [
         ('option', None, None),
     ],
+    'account_invoice': [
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why do we need then to copy it? I guess it was renamed for avoiding later errors, or it was just a "preservation" technique?

MiquelRForgeFlow:
I think I did it as a preservation technique.

pedrobaeza:
Then don't copy and simply let it as is.

MiquelRForgeFlow:
But if I do that, maybe I am breaking some custom migration that is using the legacy named field...

pedrobaeza:
OK, good point.

PR #2481, opened 2020-12-11
[ADD][13.0] Support generating noupdate_changes from upgrade_records in 14.0

StefanRijnhart:
This change allows to generate noupdate_changes in the later version (see https://github.com/legalsylvain/server-tools/pull/7/files)

### DIFF
--- a/odoo/addons/openupgrade_records/models/openupgrade_record.py
+++ b/odoo/addons/openupgrade_records/models/openupgrade_record.py
@@ -1,9 +1,12 @@
-# coding: utf-8
 # Copyright 2011-2015 Therp BV <https://therp.nl>
-# Copyright 2016 Opener B.V. <https://opener.am>
+# Copyright 2016-2020 Opener B.V. <https://opener.am>
+# Copyright 2019 Eficent <https://eficent.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
+import ast
+import os
 from odoo import api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.modules.module import get_module_path, MANIFEST_NAMES
 
 
 class Attribute(models.Model):
@@ -111,3 +114,40 @@ class Record(models.Model):
                 dict([(x.name, x.value) for x in record.attribute_ids]))
             data.append(repre)
         return data
+
+    @api.model
+    def list_modules(self):
+        """ Return the set of covered modules """
+        self.env.cr.execute(
+            """SELECT DISTINCT(module) FROM openupgrade_record
+            ORDER BY module""")
+        return [module for module, in self.env.cr.fetchall()]
+
+    @staticmethod
+    def _read_manifest(addon_dir):
+        for manifest_name in MANIFEST_NAMES:
+            if os.access(os.path.join(addon_dir, manifest_name), os.R_OK):
+                with open(os.path.join(addon_dir, manifest_name), 'r') as f:
+                    manifest_string = f.read()
+                    return ast.literal_eval(manifest_string)
+        raise ValidationError('No manifest found in %s' % addon_dir)
+
+    @api.model
+    def get_xml_records(self, module):
+        """ Return all XML records from the given module """
+        addon_dir = get_module_path(module)
+        manifest = self._read_manifest(addon_dir)
+        # The order of the keys are important.
+        # Load files in the same order as in
+        # module/loading.py:load_module_graph
+        files = []
+        for key in ['init_xml', 'update_xml', 'data']:
+            if not manifest.get(key):
+                continue
+            for xml_file in manifest[key]:
+                if not xml_file.lower().endswith('.xml'):
+                    continue
+                parts = xml_file.split('/')
+                with open(os.path.join(addon_dir, *parts), 'r') as xml_handle:
+                    files.append(xml_handle.read())
+        return files
--- a/odoo/addons/openupgrade_records/models/openupgrade_record.py
+++ b/odoo/addons/openupgrade_records/models/openupgrade_record.py
@@ -148,6 +148,9 @@ class Record(models.Model):
                 if not xml_file.lower().endswith('.xml'):
                     continue
                 parts = xml_file.split('/')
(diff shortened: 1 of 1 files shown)


odoo/addons/openupgrade_records/models/openupgrade_record.py
```diff
@@ -111,3 +114,40 @@ def field_dump(self):
...
+        return [module for module, in self.env.cr.fetchall()]
+
+    @staticmethod
+    def _read_manifest(addon_dir):
+        for manifest_name in MANIFEST_NAMES:
+            if os.access(os.path.join(addon_dir, manifest_name), os.R_OK):
+                with open(os.path.join(addon_dir, manifest_name), 'r') as f:
+                    manifest_string = f.read()
+                    return ast.literal_eval(manifest_string)
+        raise ValidationError('No manifest found in %s' % addon_dir)
```
pedrobaeza, line 133: Why not simply `return {}` for ignoring invalid folders? At the end, they are harmless for this purpose, and we avoid possible crashes.

pedrobaeza: Stefan, what do you think about this?

StefanRijnhart: Sorry for the delay! About your question, the method is only called on the list of modules that are returned by the method `list_modules` above (https://github.com/legalsylvain/server-tools/blob/14.0-ADD-upgrade_analysis/upgrade_analysis/models/upgrade_analysis.py#L462-L466), so this is a classic case of "should not happen". As such, I think we can leave it like this.

pedrobaeza: OK

odoo/addons/openupgrade_records/models/openupgrade_record.py
```diff
@@ -111,3 +114,40 @@ def field_dump(self):
...
+        files = []
+        for key in ['init_xml', 'update_xml', 'data']:
+            if not manifest.get(key):
+                continue
+            for xml_file in manifest[key]:
+                if not xml_file.lower().endswith('.xml'):
+                    continue
+                parts = xml_file.split('/')
+                with open(os.path.join(addon_dir, *parts), 'r') as xml_handle:
+                    files.append(xml_handle.read())
```
MiquelRForgeFlow:
```suggestion
                paths.append(os.path.join(addon_dir, *parts))
```
See https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/models/upgrade_record.py#L170.

StefanRijnhart: I still think there is a problem here. You changed the API to return paths instead of XML files. But it looks like this https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/models/upgrade_analysis.py#L472-L473 gives you local data even if the paths are retrieved from the remote.

MiquelRForgeFlow: Then tell what is wrong in `_parse_paths` method. It should work regardless of what path is passed.

StefanRijnhart: It looks like they will be interpreted as local paths, from the instance that runs the analysis (14.0 in this case). They are being passed straigt to etree here: https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/models/upgrade_analysis.py#L423. How can that return the remote XML data?

-- edit -- fixed link

MiquelRForgeFlow: So, how we should handle that? Should we put a `_parse_paths` method in v13 and then in the upgrade_analysis call to this remote `_parse_paths` for the remote paths?

StefanRijnhart: You have to reinstate that xml files are passed instead of paths, and then solve the UTF-8 error differently.

MiquelRForgeFlow: Would you like to make a PR in server-tools to amend that in a way that also solves the utf-8 error? I am lost with etrees and encodings.

StefanRijnhart: I don't really have the time right now. It looks like the problem could be in this line: https://github.com/OCA/server-tools/pull/2057/files#diff-e768c0bad8381e3d5a9cef3f3193890e5438592d5e9da9486c523bb55b19805aL422 because performs the utf-8 encoding explicitely. I suggest to catch encoding exceptions there and just skip that file. It's only l10n_ma/data/l10n_ma_chart_data.xml that encodes with `<?xml version="1.0" encoding="ISO-8859-15"?>` and I think that's just wrong in itself.

MiquelRForgeFlow: Ok, I will do that.

MiquelRForgeFlow:
```suggestion
                try:
                    with open(os.path.join(addon_dir, *parts), 'r') as xml_handle:
                        files.append(xml_handle.read())
                except UnicodeDecodeError:
                    continue
```

MiquelRForgeFlow (changes requested):

odoo/addons/openupgrade_records/models/openupgrade_record.py
MiquelRForgeFlow: I did in local this `files` approach instead of `paths` approach, and the noupdate data was still wrong :S

StefanRijnhart:
Fixing CI here: https://github.com/OCA/openupgradelib/pull/236

StefanRijnhart:
CI fixed with
```
2021-04-20 17:06:43,958 6073 DEBUG openupgrade OpenUpgrade: Deleting record for XML-ID survey.stage_closed 
2021-04-20 17:06:43,960 6073 INFO openupgrade OpenUpgrade: Error deleting XML-ID survey.stage_closed: KeyError('survey.stage') 
```

MiquelRForgeFlow (approved):

PR #2488, opened 2020-12-22
[13.0][FIX] fix migration with product with twice the same attribute

sebastienbeau:

 For some product you can have twice the same attribute
 Example you have a bi-color product you can set
 attribute line 1: color => white, black, yellow
 attribute line 2: color => green, blue, yellow
 if both attribute line have same value for the color it's a litle more
 complexe to fill the value as you have two product_template_attribute_value
 with the same attribute_value "yellow"

Current behavior before PR:

Migration fail

Migration work

--

### DIFF
--- a/addons/product/migrations/13.0.1.2/post-migration.py
+++ b/addons/product/migrations/13.0.1.2/post-migration.py
@@ -4,6 +4,42 @@
 from openupgradelib import openupgrade
 
 
+def fill_product_template_attribute_value_attribute_line_id_complexe_case(env):
+    # For some product you can have twice the same attribute
+    # for example you have a bi-color product
+    # you can set
+    # attribute line 1: color => white, black, yellow
+    # attribute line 2: color => green, blue, yellow
+    # if both attibute line have same value for the color it's a litle more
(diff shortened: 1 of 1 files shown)


kos94ok-3D:
Hi, @sebastienbeau.
It`s not allowed in Odoo 11, so maybe do it there?

pedrobaeza:
Yeah, that my same question: how do you get to that incorrect configuration?

sebastienbeau:
In my case, the data was coming from version 8, so it was possible in V8.
And in V13/V14 it's possible to do it.

MiquelRForgeFlow:
If two product variants have the same attributes, then we should merge those variants into one record.

MiquelRForgeFlow:
As I commented before about merging variants, I have done a PR with that approach, which I think is better. Please, could you check https://github.com/OCA/OpenUpgrade/pull/2513?

kos94ok-3D:
HI, @sebastienbeau. Could you confirm that https://github.com/OCA/OpenUpgrade/pull/2513 fixed or not your problem?

kos94ok-3D:
@sebastienbeau Any news?

Issue #2487, opened 2020-12-22
[13.0] account group_show_line_subtotals_tax_included / excluded


estraka:
Impacted versions:
13.0 

Steps to reproduce:

I have configured that the users see the subtotals with taxes included (group_show_line_subtotals_tax_included), and when I execute the migration, it fails me because the users cannot have both options marked.

ERROR:
```
2020-12-22 10:55:48,970 1 CRITICAL CadenetMig odoo.service.server: Failed to initialize database `CadenetMig`. 
Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 717, in parse
    self._tag_root(de)
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 679, in _tag_root
    f(rec)
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 582, in _tag_record
    record = model._load_records([data], self.mode == 'update')
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4139, in _load_records
    data['record']._load_records_write(data['values'])
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4077, in _load_records_write
    self.write(values)
  File "/opt/odoo/auto/addons/mail/models/res_users.py", line 177, in write
    write_res = super(res_groups_mail_channel, self).write(vals)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/res_users.py", line 1048, in write
    res = super(GroupsView, self).write(values)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/res_users.py", line 981, in write
    self._check_one_user_type()
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/res_users.py", line 112, in _check_one_user_type
    self.users._check_one_user_type()
  File "/opt/odoo/auto/addons/account/models/res_users.py", line 24, in _check_one_user_type
    raise ValidationError(_("A user cannot have both Tax B2B and Tax B2C.\n"
odoo.exceptions.ValidationError: ('Un usuario no puede tener dos tipos de impuestos B2B (Empresa a Empresa) y B2C (Empresa a Cliente).\nDebe ir a Configuración general y elegir Mostrar precios de los productos.\n con "Impuestos incluidos" o en el modo "Sin impuestos" (o cambie entre los modo si ya está en el deseado).', None)

During handling of the above exception, another exception occurred:

[...]

PR #2489, opened 2020-12-28
Fix bug: account group_show_line_subtotals_tax_included / excluded #2487

estraka:

When a migration is executed and the system and users are confirmed to show prices including taxes, the process fails due to validation:
Constraint in V13 : https://github.com/odoo/odoo/blob/13.0/addons/account/models/res_users.py#L13

Origin of the fault
no_update.xml file in OpenUpgrade : https://github.com/OCA/OpenUpgrade/blob/13.0/addons/account/migrations/13.0.1.1/noupdate_changes.xml#L41https://github.com/OCA/OpenUpgrade/blob/13.0/addons/account/migrations/13.0.1.1/post-migration.py#L1024

Current behavior before PR:

To fix the problem in a generic way, I remove the membership of account.group_show_line_subtotals_tax_included for the users member of base.group_user, then call this function before that call and delete the relationship between account.group_show_line_subtotals_tax_**included**  and base.group_user in res_groups_implied_rel table because new relationship between base.group_user and account.group_show_line_subtotals_tax_**excluded** will be created in noupdate_changes.xml file.

The process of migration not fails.

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/noupdate_changes.xml
+++ b/addons/account/migrations/13.0.1.1/noupdate_changes.xml
@@ -38,9 +38,11 @@
   <record id="action_open_settings" model="ir.actions.act_window">
     <field name="context" eval="{'module': 'general_settings', 'bin_size': False}"/>
   </record>
+  <!--
   <record id="base.group_user" model="res.groups">
     <field name="implied_ids" eval="[(4, ref('account.group_show_line_subtotals_tax_excluded'))]"/>
   </record>
+  -->
   <record id="data_unaffected_earnings" model="account.account.type">
     <field name="internal_group">equity</field>
   </record>


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -996,6 +996,42 @@ def assign_account_tags_to_move_lines(env):
     )
 
 
+def fill_invoice_ids(env):
```
pedrobaeza: Why are you doing this unrelated change?

pedrobaeza (changes requested):
I think the simplest solution is to not change groups for base user commenting this line:

https://github.com/OCA/OpenUpgrade/blob/050399414ab70667ca837b90d34205a51f978859/addons/account/migrations/13.0.1.1/noupdate_changes.xml#L42

addons/account/migrations/13.0.1.1/post-migration.py
estraka: That is an error. I saw it on github ... but I didn't realize that I had added it, I thought that I would have put some space in front of the def and it would detect it as a change ... I absolutely did not remember having put them, I would make a mistake in a: " CTRL + V "The truth is that I am new to the whole process (python, odoo, git, github ...) and I am struggling to adapt ...
Apologies for my mistake.

pedrobaeza: No worries. These things happen. Just amend it and go on.

pedrobaeza (approved):

StefanRijnhart (approved):

StefanRijnhart:
Thanks for the fix! This bit of noupdate data is not new in 13 (https://github.com/odoo/odoo/blob/12.0/addons/account/security/account_security.xml#L83-L84) and should not be force-loaded here.

PR #2299, opened 2020-04-13
[13.0][MIG] website

Yasaie:
Migration of `website`
--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr
@Tecnativa TT23136

PR #2496, opened 2021-01-20
[FIX] website_blog: fix blog and post covers

yajo:
v13 introduces `blog.blog` covers. Thus, it's better to keep the default value for `blog_blog_1` instead of setting an url to it. The default will be a solid color instead.

Also, `blog.post` covers behave similarly, but use different classes, so here I provide the required transformation.

@Tecnativa TT23206

### DIFF
--- a/addons/website_blog/migrations/13.0.1.0/post-migration.py
+++ b/addons/website_blog/migrations/13.0.1.0/post-migration.py
@@ -1,12 +1,47 @@
 # Copyright 2020 ForgeFlow <http://www.forgeflow.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import json
+
 from openupgradelib import openupgrade
+
+
+def _preserve_post_covers(env):
+    """Preserve how blog post covers display.
+
+    Performs these transformations to preserve blog post cover sizes:
+
+    * Full size:
+      .cover.container-fluid.cover_full ➡ .o_record_has_cover.cover_full
+    * Medium size:
+      .cover.container-fluid.cover_narrow ➡ .o_record_has_cover.cover_mid
+    * Narrow size:
+      .cover.container.cover_narrow ➡ .o_record_has_cover.cover_auto
+    """
+    posts = env["blog.post"].search([("cover_properties", "!=", False)])
+    for post in posts:
+        properties = json.loads(post.cover_properties)
+        resize_classes = set(properties.get("resize_class", "").split())
+        if "cover" in resize_classes:
+            resize_classes.add("o_record_has_cover")
+            # Mid
+            if {"container-fluid", "cover_narrow"} <= resize_classes:
+                resize_classes.add("cover_mid")
+            # Narrow
+            elif {"container", "cover_narrow"} <= resize_classes:
+                resize_classes.add("cover_auto")
+            resize_classes.difference_update({"cover", "container", "container-fluid"})
+            post.write(
+                {
+                    "cover_properties": json.dumps(
+                        dict(properties, resize_class=list(resize_classes))
+                    )
+                }
+            )
 
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.load_data(
-        env.cr, "website_blog", "migrations/13.0.1.0/noupdate_changes.xml")
+    _preserve_post_covers(env)
     openupgrade.delete_records_safely_by_xml_id(
         env, [
             "website_blog.blog_post_cover_01",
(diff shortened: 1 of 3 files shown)


MiquelRForgeFlow (approved):
LGTM, but I prefer if you don't delete the noupdate_changes file (and just comment all its content).

yajo:
OK, changed.

StefanRijnhart:
@pedrobaeza here we can see what's wrong with CI:
```
coveralls.exception.CoverallsException: Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
resubmitting with id None-1488390468035714388
```
Any idea how to fix this?

pedrobaeza:
I'm afraid not... It seems some kind of API deprecation. A quick fix can be to disable coveralls at all, as that information is not very valuable on OU.

MiquelRForgeFlow:
@Yajo FYI we are removing the coveralls in https://github.com/OCA/OpenUpgrade/pull/2498

yajo:
Thanks, rebased.

pedrobaeza:
`website` is not merged AFAIK.

pedrobaeza:
Indeed: #2299 Aborting

MiquelRForgeFlow:
but website_blog was already merged, no need to abort

pedrobaeza:
Wasn't it reverted? Anyway, let's go for merging website PR.

MiquelRForgeFlow:
No, it wasn't reverted because there wasn't any problem :S

yajo:
I know I'm "new" here, but IMHO there's not much benefit on marking good PRs as stale just because their dependencies aren't migrated yet. In this case, `website_blog` doesn't depend on anything specific from the `website` migration apart from the migration itself.

I mean, is there anything that wasn't broken before and that will be broken now by merging this? If not, then the most reasonable thing to do for me is merge. This way we clear the TODO list.

addons/website_blog/migrations/13.0.1.0/post-migration.py
```diff
@@ -1,12 +1,47 @@
...
+            if {"container-fluid", "cover_narrow"} <= resize_classes:
+                resize_classes.add("cover_mid")
+            # Narrow
+            elif {"container", "cover_narrow"} <= resize_classes:
+                resize_classes.add("cover_auto")
+            resize_classes.difference_update({"cover", "container", "container-fluid"})
+            post.write(
+                {
+                    "cover_properties": json.dumps(
+                        dict(properties, resize_class=resize_classes)
```
MiquelRForgeFlow:
```
2021-02-03 04:23:01,369 1 ERROR - OpenUpgrade: Object of type 'set' is not JSON serializable 
  File "/usr/local/lib/python3.6/site-packages/openupgradelib/openupgrade.py", line 1873, in wrapped_function
    if use_env2 else cr, version)
  File "/opt/odoo/auto/addons/website_blog/migrations/13.0.1.0/post-migration.py", line 44, in migrate
    _preserve_post_covers(env)
  File "/opt/odoo/auto/addons/website_blog/migrations/13.0.1.0/post-migration.py", line 36, in _preserve_post_covers
    dict(properties, resize_class=resize_classes)
  File "/usr/local/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'set' is not JSON serializable
2021-02-03 04:23:01,372 1 ERROR - odoo.modules.loading: Error executing post migration script for module website_blog
: Object of type 'set' is not JSON serializable 
```

yajo: Fixed, thanks!

pedrobaeza:
@Yajo the reason for not merging migration scripts for a module that depends on other not yet merged is because OpenUpgrade mark that modules as installed on tests, making that whole dependency chain is installed, which can provoke an error on data improperly set because they are not transformed, or improper interactions with other modules. That's luckily not the case, but we should keep this consistency for avoiding any problem in the main branch.

pedrobaeza:
@MiquelRForgeFlow you have been so quick, as I was checking if everything is correct, and ocabot seems to not work yet, so I was about to merge when green.

MiquelRForgeFlow:
ah, I thought ocabot already works :S

pedrobaeza:
Not sure, anyway, it's not adding so much extra value in OpenUpgrade, as there are no extra commits added. The only valuable thing is that it rebases PR branch over current main branch, but there's unlikely any bad interaction for not having a rebased branch (except on dependencies, but that's better to do a manual rebase and check in the PR, as I have done here).

pedrobaeza:
What I have done for future PRs is the option of auto-merge:

https://github.blog/changelog/2020-12-16-pull-request-auto-merge-public-beta/

so we don't have to be watching.

PR #2499, opened 2021-01-21
[12.0][FIX] Don't reset module states on exception

StefanRijnhart:
This is in all Odoo versions from 10.0 on. Whenever an uncaught exception occurs, module states are reset from `to upgrade` to `installed`. I think this is harmful in migration context.

--

### DIFF
--- a/odoo/modules/registry.py
+++ b/odoo/modules/registry.py
@@ -85,7 +85,9 @@ class Registry(Mapping):
                     try:
                         odoo.modules.load_modules(registry._db, force_demo, status, update_module)
                     except Exception:
-                        odoo.modules.reset_modules_state(db_name)
+                        # OpenUpgrade: don't reset module states so that the migration
+                        # can be continued after fixing the error
+                        # odoo.modules.reset_modules_state(db_name)
                         raise
                 except Exception:
                     _logger.exception('Failed to load registry')


pedrobaeza:
Is this really needed? AFAIK, the transaction is rolled-back, so everything is reset to its initial state, being `to upgrade`.

StefanRijnhart:
The module is upgraded in a transaction that is rolled back, but this is outside that transaction apparently.

StefanRijnhart:
I mean, it's easily fixed manually with an SQL query resetting the state of 'installed' modules with a lower major version number, but if you don't know that it's quite misleading.

StefanRijnhart:
Feel free to leave this open until someone confirms this behaviour. I'm a little surprised that I encountered this behaviour only now.

pedrobaeza:
Well, I don't use such flow, as I do:

* odoo -u all
* If error, I can do:
  *  odoo -u all again fixing the problem
  * restore DB and odoo -u all

But it's probably what you say then.

MiquelRForgeFlow:
> Well, I don't use such flow, as I do:

Me too. How would this flow would change if we merge this PR?

StefanRijnhart:
I do
* odoo -d db -u all --stop-after-init

In case of an error (when developing or debugging), with this fix you can

* odoo -d db --stop-after-init

This will continue the upgrade process from where it left off, starting with the module that caused the error.
Obviously, that way the migration will not run the end-migration scripts of the modules that were already upgraded before the error, so in case of a live migration you always need to restore the database.

pedrobaeza:
OK, thanks for the explanations, and it's good to have this alternative for not having to update the already migrated modules again. About end migration scripts, both flows are affected. Then this change affects the other flow?

StefanRijnhart:
No, you can still run `-u all` on a database that already has marked some modules for upgrading. The result will still be that all other installed modules are marked for upgrading as well.

pedrobaeza (approved):

StefanRijnhart:
Thanks. I'll port this to other versions.

pedrobaeza:
It should be put in 13.0 as well (and other versions?)

MiquelRForgeFlow:
@StefanRijnhart

StefanRijnhart:
I answered Pedro's question before he posted it ;-) And it's on my todo list. I assumed there is no hurry?

MiquelRForgeFlow:
Hahaha, sorry, I didn't read correctly. The order of messages are important!

No, there is no hurry. Just to be sure this doesn't get forgotten.

MiquelRForgeFlow:
@StefanRijnhart I tested this. The end-migrations are not a problem, because I can execute them using shell. The problem is later, when you update the migrated db with odoo, that the obsolete views are breaking because they were not deleted. It means that the cleanup process neither was executed. Is there a way to execute manually the cleanup process?

PR #2500, opened 2021-01-21
[13.0] [FIX] Account: populate the journal_id field of move lines

ThomasBinsfeld:
The journal_id of SQL created move lines should not be let empty.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -314,13 +314,13 @@ def migration_invoice_moves(env):
         if ail_ids:
             openupgrade.logged_query(
                 env.cr, """
-                INSERT INTO account_move_line (company_id, account_id,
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):

MiquelRForgeFlow (approved):
The field is related, so I thought it would be filled automatically.

pedrobaeza:
The ocabot flow is not working with GitHub actions. cc @sbidoul

sbidoul:
@pedrobaeza that's because it's configured to [trigger only on 13.0 branches](https://github.com/OCA/OpenUpgrade/blob/2cd778116acb0ddbb08d7ffd9a3db18971e6ed86/.github/workflows/test.yml#L8), and therefore not the `13.0-ocabot-merge...` branch. I think gha accept wildcards for branches (to be verified), so `13.0*` there may work.

pedrobaeza:
Thanks for the insights. I have changed it to that, so let's see next PR if it works.

PR #2502, opened 2021-01-22
[11.0][FIX] stock: use a global route to avoid creating too many rules

MiquelRForgeFlow:
If your bbdd has too many companies, and your global rules don't use several warehouses, then when transforming that global rule into specific rules for each route, you are creating many unnecessary rules. To avoid that, the proper way should be create a global route and then update the global rules by using that route.

--

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -59,34 +59,21 @@ def product_assign_responsible(env):
 
 @openupgrade.logging()
 def create_specific_procurement_rules_from_globals(env):
-    """Create one record per route for the global rules found in previous
-    version.
-    """
-    for table in ['procurement_rule', 'stock_location_path']:
-        env.cr.execute(
-            """SELECT column_name
-            FROM information_schema.columns
-            WHERE table_name = %s
-                AND column_name != 'id'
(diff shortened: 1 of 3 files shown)


addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -59,32 +59,22 @@ def product_assign_responsible(env):
 
 @openupgrade.logging()
 def create_specific_procurement_rules_from_globals(env):
-    """Create one record per route for the global rules found in previous
-    version.
-    """
+    """Update global rules by linking them to a global route."""
+    warehouses = env["stock.warehouse"].with_context(active_test=False).search(
+        [])
+    global_route = env["stock.location.route"].create({
```
StefanRijnhart: I think this route gets created when there are no dangling rules or paths. Maybe you can fetch those first, and only create this route if necessary. You can refactor out the SQL queries by writing pull_ids and push_ids directly at creation time.

MiquelRForgeFlow: I think it's done! :)

StefanRijnhart (approved):
Great, thanks!

hbrunn (approved):

LoisRForgeFlow (approved):
Tested, this is much better than before!

PR #2512, opened 2021-02-01
[13.0][FIX] account: read fields when computing balance for draft invoices

MiquelRForgeFlow:
Add some stored fields.

--


kos94ok-3D (approved):
It just prefetches fields once, so OK for me.

pedrobaeza:
Why not reading all fields for avoiding this problem with other modules that add fields to this computation?

kos94ok-3D:
I just tried to improve some performance and didn't think that it will be a problem.

MiquelRForgeFlow:
@pedrobaeza now it reads all fields

pedrobaeza (approved):

MiquelRForgeFlow:
ocabot still not working :S

PR #2513, opened 2021-02-02
[13.0][FIX] product: merge product variants with same attributes

MiquelRForgeFlow:
Make use of merge_records tool to merge records.

--

### DIFF
--- a/addons/product/migrations/13.0.1.2/pre-migration.py
+++ b/addons/product/migrations/13.0.1.2/pre-migration.py
@@ -1,7 +1,7 @@
 # Copyright 2020 ForgeFLow <http://www.forgeflow.com>
 # Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from openupgradelib import openupgrade
+from openupgradelib import openupgrade, openupgrade_merge_records
 from odoo.tools import sql
 
 _xmlid_renames = [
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
I can retarget to v11 branch if you prefer.

kos94ok-3D:
We have not the constrain for combination_indices in pre-migration so maybe better to calculate combination_indices and then group product by product template and combination_indices?

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -123,6 +124,36 @@ def insert_missing_product_template_attribute_value(env):
...
+        JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        JOIN templates ON (templates.id = pt.id
+            AND templates.combination_indices = pp.combination_indices)"""
+    )
+    templates = {}
+    for product_id, template_id in env.cr.fetchall():
+        if template_id in templates:
+            templates[template_id] += [product_id]
+        else:
+            templates[template_id] = [product_id]
```
kos94ok-3D: I think  it is looks better.
```suggestion
         templates.setdefault(template_id, []).append(product_id)
```

kos94ok-3D (approved):

kos94ok-3D:
Hi, @MiquelRForgeFlow Could you rebase it?

kos94ok-3D:
Maybe need to be added in a commit? https://github.com/odoo/odoo/pull/32086
And said that it was added in Odoo v12 6 months after the release of Odoo v13.

kos94ok-3D (approved):
Tested on a real DB

pedrobaeza (commented):
Shouldn't you update combination_indices after the merge?

MiquelRForgeFlow:
> Shouldn't you update combination_indices after the merge?

No, why? The combination_indices are the same, they don't change.

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -161,6 +161,33 @@ def insert_missing_product_template_attribute_value(env):
...
+        JOIN templates ON (templates.id = pt.id
+            AND templates.combination_indices = pp.combination_indices)"""
+    )
+    templates = {}
+    for product_id, template_id in env.cr.fetchall():
+        templates.setdefault(template_id, []).append(product_id)
+    for template_id in templates:
+        openupgrade_merge_records.merge_records(
+            env, 'product.product', templates[template_id][1:],
+            templates[template_id][0], field_spec=None, method='sql',
```
MiquelRForgeFlow, line 187: The question now is: should I put a field_spec?

pedrobaeza: It depends if the standard fields require it or default merge values are enough

MiquelRForgeFlow: Yeah, my best option is `{'default_code': 'first_not_null', 'active': 'or', 'barcode': 'first_not_null', 'volume': 'max', 'weight': 'max'}`. My concern is for custom modules that add float/char/boolean fields, which would need to add more things in this dict. Maybe we could add in openupgradelib a functionality like if the dict contains `'other_fields': 'preserve'`, then every other field is excluded and preserved.

MiquelRForgeFlow: https://github.com/OCA/openupgradelib/pull/270

pedrobaeza:
My take on this is that merging complete products is something that shouldn't be corrected at OpenUpgrade level. If they have done such ugly thing, then it must be fixed before doing the migration.

MiquelRForgeFlow:
Ok (don't delete branch)

cvinh:
FYI this PR fixed our issue where we have a product with same attributes
@MiquelRForgeFlow thanks for not having deleted it

PR #2514, opened 2021-02-02
[13.0][FIX] account: post-migration: simple speed up of some queries

MiquelRForgeFlow:
Passing a big tuple as a variable slows down a lot a query. Thus, we should try to avoid that.

Sometimes we seem blind people in front of some obvious bad code hahaha 🤦‍♂️

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -253,45 +253,25 @@ def migration_invoice_moves(env):
     )
     aml_ids += tuple(x[0] for x in env.cr.fetchall())
     # 2st: exclude from invoice_tab the grouped ones, and create a new separated ones
-    if aml_ids:
-        openupgrade.logged_query(
-            env.cr, """
-            UPDATE account_move_line aml
-            SET exclude_from_invoice_tab = TRUE, sequence = ail.sequence,
-            name = '(OLD GROUPED ITEM)' || aml.name,
-            create_uid = ail.create_uid, create_date = ail.create_date
-            FROM account_invoice_line ail
-                JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
-                JOIN account_move am ON ail.invoice_id = am.old_invoice_id
-            WHERE am.id = aml.move_id AND ail.company_id = aml.company_id AND ail.account_id = aml.account_id
-                AND ail.partner_id = aml.partner_id
-                AND ((ail.product_id IS NULL AND aml.product_id IS NULL) OR ail.product_id = aml.product_id)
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -253,45 +253,25 @@ def migration_invoice_moves(env):
...
-            WHERE am.id = aml.move_id AND ail.company_id = aml.company_id AND ail.account_id = aml.account_id
-                AND ail.partner_id = aml.partner_id
-                AND ((ail.product_id IS NULL AND aml.product_id IS NULL) OR ail.product_id = aml.product_id)
-                AND ((ail.uom_id IS NULL AND aml.product_uom_id IS NULL) OR ail.uom_id = aml.product_uom_id)
-                AND ((ail.account_analytic_id IS NULL AND aml.analytic_account_id IS NULL)
-                    OR ail.account_analytic_id = aml.analytic_account_id)
-                AND aml.tax_line_id IS NULL
-            RETURNING aml.id""",
-        )
+    openupgrade.logged_query(
```
pedrobaeza, line 256: Is there any possibility that this query includes rows that have been included in the previous ones, updating opposite information like `exclude_from_invoice_tab`?

MiquelRForgeFlow: no, because previous ones have the `old_invoice_line_id` filled, and this query only runs where it's unfilled.

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -343,9 +323,8 @@ def migration_invoice_moves(env):
             SET exclude_from_invoice_tab = TRUE
             FROM account_move_line aml2
```
pedrobaeza: No need here of the `FROM` part and you can optimize for not re-writing some lines if not needed:
```python
            """
            UPDATE account_move_line aml
            SET exclude_from_invoice_tab = TRUE
            WHERE aml.old_invoice_line_id IS NULL
                AND NOT exclude_from_invoice_tab"""
```

MiquelRForgeFlow: maybe last line should say `AND NOT aml.exclude_from_invoice_tab or aml.exclude_from_invoice_tab IS NULL`

pedrobaeza: A boolean field can have `NULL`?

MiquelRForgeFlow: if not filled?

pedrobaeza: OK, just remove then that condition.
```python
            """
            UPDATE account_move_line aml
            SET exclude_from_invoice_tab = TRUE
            WHERE aml.old_invoice_line_id IS NULL
            """
```

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #2519, opened 2021-02-03
[IMP] stock: adapt records instead of re-create

kos94ok-3D:
We do not use multi-company so we enough update exist locations instead of creating new (location_inventory, location_production, stock_location_scrapped)

--


kos94ok-3D:
Waiting for review: @pedrobaeza, @MiquelRForgeFlow, @StefanRijnhart

StefanRijnhart:
Reiterating my comment from https://github.com/OCA/OpenUpgrade/pull/2306
> that single-company shortcut, I can imagine it is a lot quicker than updating all the location ids throughout the database. Do you think it can be generalized, assigning the old locations to the company with the lowest id, and then excluding that company_id when assigning the new location ids? It might save some time and disk space even on multi company databases.

kos94ok-3D:
@StefanRijnhart I`m not sure that I understood you.
Did you say about it? [link](https://github.com/kos94ok-3D/OpenUpgrade/compare/13.0...kos94ok-3D:13.0-stock-upgrade_record?expand=1)

StefanRijnhart:
This code makes an exception for the case where there is only one company in the database. I think multicompany setups could benefit from a refactored version of this code that preserved existing locations to use with the company with the lowest ID.

kos94ok-3D:
In the previous message, I hide the link: https://github.com/kos94ok-3D/OpenUpgrade/compare/13.0...kos94ok-3D:13.0-stock-upgrade_record?expand=1

kos94ok-3D:
Hi, @StefanRijnhart. I think you didn`t have the notification about the previous message.

StefanRijnhart:
Ah thanks. That comparison you linked to is an alternative that implements my suggestion? That's great! Are you happy with the result or do you doubt its usefulness?

kos94ok-3D:
I'm not against it, so I waiting for a responce from @MiquelRForgeFlow and @pedrobaeza.

StefanRijnhart:
In that comparison, don't you need to keep the bit where the newly created locations are assigned to the pickings, moves etc (except for the pickings, moves, etc from the main company)?

kos94ok-3D:
That was need only for assign new locaitions which was created instead of `stock.location_inventory`, `stock.location_production`, `stock.stock_location_scrapped`

StefanRijnhart:
> That was need only for assign new locaitions which was created instead of stock.location_inventory, stock.location_production, stock.stock_location_scrapped

But for each of the companies that is not main_company, those locations are still created and need to be assigned throughout the stock system. The main idea is that where currently the stock migration scripts generate and assign new locations for all companies, I am suggesting to generate and assign new locations for all companies except the main_company.

MiquelRForgeFlow:
same opinion as @StefanRijnhart

kos94ok-3D:
@MiquelRForgeFlow , @StefanRijnhart I see, check that same link.

StefanRijnhart:
You're quick :hot_face:!

This is very much the idea. Can we agree to use that code for this PR? We can then continue the review here (one thing to note is that you don't have to check if main_company is defined, because it always is).

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -344,24 +355,35 @@ def map_stock_locations(env):
...
     fill_stock_inventory_start_empty(env)
     fill_stock_inventory_fields(env)
     map_stock_location_usage(env)
     fill_stock_picking_type_sequence_code(env)
-    handle_stock_scrap_sequence(env)
-    map_stock_locations(env)
+    handle_stock_scrap_sequence(env, main_company)
+    map_stock_locations(env, main_company)
     convert_many2one_stock_inventory_product_and_location(env)
     openupgrade.delete_records_safely_by_xml_id(env, _unlink_by_xmlid)
```
MiquelRForgeFlow: delete this and also delete the end-migration

kos94ok-3D: Done

MiquelRForgeFlow:
The code looks good

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -246,47 +243,52 @@ def convert_many2one_stock_inventory_product_and_location(env):
...
-    # although later the 'stock.sequence_stock_scrap' sequence will be deleted,
-    # we need to nullify its code (if having one) here
-    # because we want a new autogenerated sequence
+def handle_stock_scrap_sequence(env, main_company):
+    # assign on main company
     openupgrade.logged_query(
         env.cr, """
         UPDATE ir_sequence seq
-        SET code = NULL
+        SET company_id = %s, name = %s || ' ' || seq.name
```
MiquelRForgeFlow:
```suggestion
        SET company_id = %s, name = %s || ' Sequence scrap'
```

MiquelRForgeFlow (approved):
LGTM

addons/stock/migrations/13.0.1.1/post-migration.py
kos94ok-3D: It is not important for me so OK

MiquelRForgeFlow:
@pedrobaeza

pedrobaeza:
Well, I'm not following this, so a better reviewer would be @StefanRijnhart

StefanRijnhart (approved):
Great, thanks!

PR #2520, opened 2021-02-03
[11.0][IMP] rating: prepopulate stored field on mail_message

StefanRijnhart:
An old optimization I had lying around. Prevents a compute method from kicking in on all messages in the system (https://github.com/odoo/odoo/blob/11.0/addons/rating/models/mail_message.py#L15)

### DIFF
--- a/addons/rating/migrations/11.0.1.0/pre-migration.py
+++ b/addons/rating/migrations/11.0.1.0/pre-migration.py
@@ -14,3 +14,14 @@ COLUMN_COPIES = {
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.copy_columns(env.cr, COLUMN_COPIES)
+
+    # Precreate and populate mail_message's new stored field rating_value
+    openupgrade.add_fields(env, [
+        ("rating_value", "mail.message",
+         "mail_message", "float", False, "rating", 0)])
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE mail_message mm
+        SET rating_value = rr.rating
+        FROM rating_rating rr
+        WHERE rr.message_id = mm.id""")


legalsylvain (approved):
Code review. Looks ok. Thanks.

pedrobaeza:
Please check Travis

StefanRijnhart:
Thanks for noting, now green!

addons/rating/migrations/11.0.1.0/pre-migration.py
```diff
@@ -14,3 +14,18 @@
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.copy_columns(env.cr, COLUMN_COPIES)
+
+    # Precreate and populate mail_message's new stored field rating_value
+    env.cr.execute(
```
pedrobaeza: Maybe you can use the openupgradelib function that saves you also to handle the default value and thus the latest query:
```python
openupgrade.add_fields(env, [("rating_value", "mail.message", "mail_message", "float", False, "rating", 0)])
```
but I don't want to be picky, so if you don't want to change this, we can merge it as is.

StefanRijnhart: Yes, alright.

pedrobaeza (approved):

pedrobaeza:
FYI, we don't have to care anymore about these optimizations in 12.0+ since this PR:

https://github.com/odoo/odoo/pull/65232

It was something needed not only for migrations, but for new modules that are installed containing this kind of fields and with lots of data already in the DB.

StefanRijnhart:
Cool, but this was a computed field, not a related field.

pedrobaeza:
OK, then we will need the optimization in such cases.

PR #2524, opened 2021-02-04
[13.0][account][fix] do not create new accounts if they are going to conflict with existing ones.

JordiBForgeFlow:

Current behavior before PR:

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/end-migration.py
+++ b/addons/account/migrations/13.0.1.1/end-migration.py
@@ -40,11 +40,22 @@ def add_default_account_data(env):
                 if code_main > 0 and code_main <= chart.code_digits:
                     code_acc = str(code_acc) + (
                         str("0" * (chart.code_digits - code_main)))
-                tax_template_ref = {}
-                vals = chart._get_account_vals(
-                    company, account_template, code_acc, tax_template_ref)
-                account = chart._create_records_with_xmlid(
-                    "account.account", [(account_template, vals)], company)
+                account = env['account.account'].search([
+                    ('code', '=', code_acc),
+                    ('company_id', '=', company.id),
+                ], limit=1)
+                # If there's already an account with the same code and company don't create it
+                # as it is not allowed by "account_account_code_company_uniq" constraint.
+                if not account:
+                    tax_template_ref = {}
+                    vals = chart._get_account_vals(
+                        company, account_template, code_acc, tax_template_ref)
+                    account = chart._create_records_with_xmlid(
+                        "account.account", [(account_template, vals)], company)
+                else:
+                    # we add xmlid for such existing account
+                    env["ir.model.data"].sudo()._update_xmlids(
+                        [dict(xml_id=xml_id, noupdate=True, record=account)])
             company[company_field] = account.id
 
 


JordiBForgeFlow:
cc @MiquelRForgeFlow

MiquelRForgeFlow (approved):
LGTM

addons/account/migrations/13.0.1.1/end-migration.py
```diff
@@ -36,6 +36,13 @@ def add_default_account_data(env):
                 if code_main > 0 and code_main <= chart.code_digits:
                     code_acc = str(code_acc) + (
                         str("0" * (chart.code_digits - code_main)))
+                if env['account.account'].search([
+                    ('code', '=', code_acc),
+                    ('company_id', '=', company.id),
+                ]):
+                    # If there's already an account with the same code and company don't create it
+                    # as it is not allowed by a constraint.
+                    continue
```
legalsylvain: Hi. Don't you have to affect  company[company_field] to the found account ? or maybe I missed something ?

legalsylvain (changes requested):

pedrobaeza (changes requested):
This is a fault by your part, not executing `account_chart_update` in previous version for matching existing accounts for regenerating XML-IDs. We can argue about the tolerance of OpenUpgrade on this, but then you are not adding the account but still letting the non-existing XML-ID, which can cause other problems. I vote for not adding this and forces to act according the other steps.

AaronHForgeFlow:
I try to migrate a database with the generic chart of accounts installed on two companies. No any other account and any change to the chart of accounts. But without this PR I guess this error:

```
File "/opt/odoo/auto/addons/account/migrations/13.0.1.1/end-migration.py", line 43, in add_default_account_data
    "account.account", [(account_template, vals)], company)
  File "/opt/odoo/auto/addons/account/models/chart_template.py", line 602, in _create_records_with_xmlid
    return self.env[model]._load_records(data_list)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4171, in _load_records
    records = self._load_records_create([data['values'] for data in to_create])
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4085, in _load_records_create
    return self.create(values)
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 336, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3813, in create
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3919, in _create
    cr.execute(query, params)
  File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 173, in wrapper
  File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 250, in execute
psycopg2.IntegrityError: duplicate key value violates unique constraint "account_account_code_company_uniq"
DETAIL:  Key (code, company_id)=(101300, 3) already exists.
```
I think OU should prevent this.

pedrobaeza:
There's no generic CoA since v8, so this should be cleaned before, not here.

addons/account/migrations/13.0.1.1/end-migration.py
MiquelRForgeFlow: resolved

pedrobaeza:
Although I still think that this should be fixed outside OpenUpgrade, your are proposing a half solution, as the XML-ID is not created for such existing account.

MiquelRForgeFlow:
@pedrobaeza it's ready :)

pedrobaeza (approved):

PR #2525, opened 2021-02-04
[13.0][MIG] sale_expense

bishalgit:
Migration of sale_expense

--


addons/sale_expense/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,18 @@
...
+# DONE: post-migration: try to delete
+
+NEW ir.ui.view: sale_expense.assets_backend
+NEW ir.ui.view: sale_expense.hr_expense_form_view_inherit_account_manager
+NEW ir.ui.view: sale_expense.hr_expense_form_view_inherit_saleman
+NEW ir.ui.view: sale_expense.qunit_suite
+# NOTHING TO DO: new xmls
+
+DEL product.product: hr_expense.product_product_fixed_cost (noupdate)
+# NOTHING TO DO: moved from one data file to another
```
pedrobaeza: Moved from where? Seeing this, it seems it has to be removed as well

MiquelRForgeFlow: yeah, it has to be removed as well

bishalgit: This data has been moved to demo file. So yes it can be removed. But should I remove from hr_expense instead? I will also migrate hr_expense to fix this then.

pedrobaeza: @MiquelRForgeFlow why it hasn't been recognized as moved from other module?

MiquelRForgeFlow: because demo data is not used in analysis

MiquelRForgeFlow: @bishalgit the hr_expense is already migrated in https://github.com/OCA/OpenUpgrade/pull/2439, which you can review

pedrobaeza: OK, I see, I would let then `hr_expense` and try here to remove it: if not possible, it has been used already.

bishalgit: @MiquelRForgeFlow and @pedrobaeza I will try to delete here as well then. Thanks

MiquelRForgeFlow (approved):

pedrobaeza:
The change of the line in the documentation is needed for both documenting this migration and for activating the check in the CI.

bishalgit:
@pedrobaeza It is all good now. Can you review now?

pedrobaeza (approved):

Issue #2339, opened 2020-05-19
[13.0] duplicate key when renaming needaction_partner_ids to notified_partner_ids


sbidoul:
Impacted versions: 13.0 (testing with all open 13.0 PR aggregated)

Steps to reproduce: Run openupgrade 8->9, 9->10, ..., 12-13.

Current behavior: 

In the 12->13 step:

```
bad query: 
            UPDATE ir_model_fields
            SET name = 'notified_partner_ids'
            WHERE name = 'needaction_partner_ids'
                AND model = 'mail.message'
            
ERROR: duplicate key value violates unique constraint "ir_model_fields_name_unique"
DETAIL:  Key (model, name)=(mail.message, notified_partner_ids) already exists.
```

This is apparently because `notified_partner_ids` was a thing in v8, then disappeared, then reappeared in v13.

Not sure if this is something I'm doing wrong (do I need to do some cleanup between steps?) or if it's something that has to / can be fixed. I can try to provide a fix if you think it's useful and give me a hint on how to approach this.

pedrobaeza:
Indeed this kind of things can happen when you migrate in batch all versions. I suggest to patch this in openupgradelib catching that error and deleting the old row in that case, so we have a patch for all possible collisions.

pedrobaeza:
I have implemented the suggestion in OCA/openupgradelib@38f0d13, but renaming instead of deleting, for being more conservative.

pedrobaeza:
Another iteration has been needed in https://github.com/OCA/openupgradelib/commit/351034f8e18ff49768748c6995ad2137a3ac9b1a (+ a syntax fix) for fully handling this, but now tested and working.

StefanRijnhart:
Working on an alternative fix as the previous fix would lead to dropping ir.property records if migration scripts were not careful and renamed the field twice (which is not unlikely for property fields, apparently). In the mean time, this particular case can be solved manually by removing the ir.model.fields record, and its ir.model.data entry manually before executing the 13.0 migration.

FranciscoBigWings:
Hello. I am facing this error when upgrading from V12 to V13. Was this fixed through Openupgrade or should I manually remove the record as suggested above?

Thanks

FranciscoBigWings:
Still getting this error saturday 27th. Impacted migration V12 to V13
- Original database is from V8 so I understand the error come from then.
- V12 is working stable in production enviroment since more than a year ago
- Update all to database V12 before migration to apply latest changes
[...]

PR #2531, opened 2021-02-04
[9.0][FIX] Don't (also) insert child taxes as parent taxes

StefanRijnhart:
After using OpenUpgrade 9.0, parent-child relationship between account.tax records are perfectly circular. There was code to populate the many2many tables in the correct way and code to populate them the wrong way around.

Trying to update the taxes in the UI raises a recursion error.

image

--

### DIFF
--- a/addons/account/migrations/9.0.1.1/post-migration.py
+++ b/addons/account/migrations/9.0.1.1/post-migration.py
@@ -135,17 +135,17 @@ def account_templates(env):
 
 
 def parent_id_to_m2m(cr):
-    cr.execute(
-        'insert into account_tax_template_filiation_rel '
-        '(parent_tax, child_tax) '
-        'select id, parent_id from account_tax_template '
-        'where parent_id is not null'
-    )
-    cr.execute(
-        'insert into account_tax_filiation_rel '
-        '(parent_tax, child_tax) '
-        'select id, parent_id from account_tax where parent_id is not null'
-    )
+    # Get parent_id from tax template and insert children_tax_ids (m2o => m2m)
+    cr.execute("""
+    INSERT INTO account_tax_template_filiation_rel (parent_tax, child_tax)
+    SELECT parent_id, id from account_tax_template WHERE parent_id IS NOT NULL
+    """)
+
+    # Get parent_id from tax and insert children_tax_ids (o2m => m2m)
+    cr.execute("""
+    INSERT INTO account_tax_filiation_rel (parent_tax, child_tax)
+    SELECT parent_id, id from account_tax WHERE parent_id IS NOT NULL
+    """)
 
 
 def parent_id_to_tag(env, model, tags_field='tag_ids', recursive=False):
@@ -1285,18 +1285,6 @@ def migrate(env, version):
     # Set display_on_footer to False
     cr.execute("""
     UPDATE account_journal SET display_on_footer = False
-    """)
-
-    # Logic to move from child_ids to children_tax_ids (o2m => m2m)
-    cr.execute("""
-    INSERT INTO account_tax_filiation_rel (parent_tax, child_tax)
-    SELECT parent_id, id from account_tax WHERE parent_id IS NOT NULL
-    """)
-
-    # Get parent_id and insert it into children_tax_ids (m2o => m2m)
-    cr.execute("""
-    INSERT INTO account_tax_template_filiation_rel (parent_tax, child_tax)
-    SELECT parent_id, id from account_tax_template WHERE parent_id IS NOT NULL
     """)
 
     # In v8, if child_depend == True, then in v9, set amount_type='group'


hbrunn (approved):

StefanRijnhart:
Test `addons/account/migrations/9.0.1.1/tests/test_product_property.py` is broken, just like in the previous merge from the 28th of November. In the merge before that, on 26th of October 2020, the test succeeded. I don't see any relevant changes in those PRs or in OpenUpgradelib for that matter. I'll come back to this.

legalsylvain (approved):

MiquelRForgeFlow (approved):
LGTM, it's that kind of "common sense" PR

StefanRijnhart:
The field from the test was renamed twice, which became a problem after this fix in openupgradelib through one of @pedrobaeza's direct commits (or rather 3: https://github.com/OCA/openupgradelib/commits/01e27a9). I'm proposing an improvement here: https://github.com/OCA/openupgradelib/pull/227. I like to think that someone would have suggested somethig like this if the original change had gone through a proper PR but maybe that's wishful thinking.

I'm fixing the programming error of renaming the field twice in the second commit in this PR.

StefanRijnhart:
Also, this should be an encouragement for all of us to write some migration tests sometimes.

hbrunn:
@StefanRijnhart thank you for the fix. @pedrobaeza I share Stefan's concern that you sidestep the process in your effort to move on fast. Can we start a discussion how to combine speed and reviews? If used sparsely, I could do priority reviews when tagged a certain string or some other filter

pedrobaeza:
Don't want to be in a defensive way, but honestly would you anticipate these things through a PR in openupgradelib? If you see dates, that was done on a Saturday, performing a migration, and due to the "fragmentation" of things (Python library, OCA modules, OpenUpgrade), preparing an environment for the temporary pull requests does things harder, and more when this was something already detected in the past (see #2339). The commits are well documented about the whys and the how.

Sorry for the extra time you have needed for this PR that is unrelated to what you are contributing. If you have suspected in any moment that is due to that, you can tell me and I will spend the needed time for fixing it. You know I'm responsible with any of my actions.

Said that, `rename_property` is doing only half of the operations needed on a renaming (no export profiles adjustment, no filters change, etc). What was the original error or calling both? Maybe we can call `rename_property` from `rename_fields` if we detect that is a property.

StefanRijnhart:
The error was that the original method to rename properties (like the one from openupgradelib) renames the ir_model_fields. Then, `openupgradelib.rename_fields`, which also renames the ir_model_fields, was called on all fields from the module including the property fields. As of the 5th of December, this method is no longer reentrant. The ir_model_fields of the new field was found to exist and was deleted, removing entries from ir_property with it as per cascade.

Agreed that the `rename_property` method is lacking. I've adapted my second commit here so that `rename_fields` can still be called.

I regularly use a custom openupgradelib when I need an unmerged bugfix by installing it from my own fork so I don't see how production pressure should allow unreviewed changes, although I agree with you that this issue might not have been picked up in any review.

StefanRijnhart:
I'm removing the second commit, which fixes the way that properties from the account module are renamed as it's a very hairy issue in itself. I'm working on a fix for that separately.  I think it would be wise not to run a final migration using openupgrade 9.0 until the field renaming issue is fixed.

pedrobaeza:
@StefanRijnhart I have migrated a v8 database to v9 with current main branch code without problems. Why is this failing though?

StefanRijnhart:
This PR fails in Travis because of the terrible situation with the ir.property records that are currently being dropped due to the change in `rename_fields`. That can only be fixed properly by fixing the way that Odoo 9, 10 and 11 treat ir.model.fields and their ir.model.data entries when loading databases (I'm actively working on that, but evenings only. See https://github.com/OCA/OpenUpgrade/pull/2535 for what that is going to look like in 9.0). 
So you might have lost your product account settings. And you will have circular tax hierarchy but that is only a problem when you edit certain values of the taxes in the UI (e.g. when you change the order of the child taxes).

StefanRijnhart:
I would feel much better if we could agree to revert those commits from 5th of december in openupgradelib (in the mean time). That would give an error migrating from 8.0 to 13.0 but the workaround is to remove the notified_partner_ids field entry and its ir_model_data entry manually before executing the 13.0 migration.

StefanRijnhart:
https://github.com/OCA/openupgradelib/pull/229 made this PR green, so with three approvals I'm merging.

pedrobaeza:
@StefanRijnhart do you have any query for fixing the problem on migrated DBs with the previous code?

StefanRijnhart:
That would be
```
delete from account_tax_filiation_rel rel
where exists(
    select * from account_tax at
    where at.id = rel.parent_tax and at.parent_id = rel.child_tax);
```

pedrobaeza:
Thank you very much! Very helpful.

PR #2533, opened 2021-02-05
[11.0][FIX] base: end-migration: merge_config_settings

MiquelRForgeFlow:
Example how I found this problem:

In v10, in a custom module, you have a model `xxx.config.settings`, with a field defined as `lalala = field.Selection(..., default = "A", required=True)`.

After migrating, there exists a default record in `ir_default` table that corresponds to that default value A, but this record, its `field_id` is pointing to the `ir_models_fields` table, in a record that still has the `model` = `xxx.config.settings`. Thus, when a user enter the Settings, do something and save, it gets the information pop up that the field `lalala` is still unfilled and requires to fill it up (it means it doesn't detect the default).

So, the solution to this example is: in ir_models_fields table, renaming "xxx.config.settings" to "res.config.settings".

But as we are at it, we do a merge of the setting models and then check all the possible cases where may happen this of inconsistent obsolete data.

--

### DIFF
--- a/odoo/addons/base/migrations/11.0.1.3/end-migration.py
+++ b/odoo/addons/base/migrations/11.0.1.3/end-migration.py
@@ -1,9 +1,106 @@
 # Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from openupgradelib import openupgrade
+from openupgradelib.openupgrade_merge_records import merge_records
+
+
+def merge_config_settings(env):
+    env.cr.execute("""
+        SELECT id, model
+        FROM ir_model
+        WHERE model LIKE '%.config.settings' AND transient
+            AND model != 'res.config.settings'""")
+    list_models = {x[0]: x[1] for x in env.cr.fetchall()}
+    if list_models:
+        env.cr.execute("""
+            SELECT id FROM ir_model
+            WHERE model = 'res.config.settings'""")
+        setts_id = env.cr.fetchone()[0]
+        for model_id, model_name in list_models.items():
+            _update_referenced_data(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't really get the use case, and I think this should be handled on such custom module, not in a generic way here if there's a chance this provoke other side effects.

odoo/addons/base/migrations/11.0.1.3/end-migration.py
```diff
@@ -1,9 +1,109 @@
...
+                )
+            )
+        for rename in rename_model_id:
+            openupgrade.logged_query(
+                env.cr, """
+                UPDATE {table}
+                SET {field} = {new_value}
+                WHERE {field} = {old_value}
+                """.format(
+                    table=rename[0], field=rename[2],
```
kos94ok-3D: rename have just two elements or I miss something?

MiquelRForgeFlow:
> I don't really get the use case

What thing of my example do you not understand? :S

MiquelRForgeFlow:
> there's a chance this provoke other side effects.

Also, I have just put `delete=False` on the merge to assure there is no side effects.

pedrobaeza:
What is the relation between `ir_default` and the old field.

MiquelRForgeFlow:
> What is the relation between ir_default and the old field.

Which old field? :S

pedrobaeza:
> Which old field? :S

Well, the base of this PR, as there's a field `lalala` in `xxx.config.settings`, but how this relates to the new model?

MiquelRForgeFlow:
> lalala in xxx.config.settings, but how this relates to the new model?

Ah, that was implicit in my example, sorry. I mean, in v11, in the python code, it's obvious you would have renamed the xxx.config.settings to res.config.settings, and you would have lalala field in the res.config.settings.

pedrobaeza:
But they are transient models. Why do you need to perform this difficult renaming?

PR #2534, opened 2021-02-06
[13.0][MIG] website_livechat

AMhadhbi:
Nothing to do

--


addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,25 @@
+---Models in module 'website_livechat'---
+---Fields in module 'website_livechat'---
+website_livechat / mail.channel             / livechat_active (boolean)     : NEW
+website_livechat / mail.channel             / livechat_visitor_id (many2one): NEW relation: website.visitor
+website_livechat / website.visitor          / livechat_operator_id (many2one): NEW relation: res.partner, isfunction: function, stored
+website_livechat / website.visitor          / mail_channel_ids (one2many)   : NEW relation: mail.channel
+# Nothing to do
+---XML records in module 'website_livechat'---
+DEL im_livechat.channel: website_livechat.im_livechat_channel_data_website (noupdate)
```
pedrobaeza, line 9: This is in fact renamed to `im_livechat.im_livechat_channel_data` (but in `im_livechat`), so it's interesting to mark it.

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,25 @@
+---Models in module 'website_livechat'---
+---Fields in module 'website_livechat'---
+website_livechat / mail.channel             / livechat_active (boolean)     : NEW
+website_livechat / mail.channel             / livechat_visitor_id (many2one): NEW relation: website.visitor
+website_livechat / website.visitor          / livechat_operator_id (many2one): NEW relation: res.partner, isfunction: function, stored
```
pedrobaeza, line 5: I think you should fill this in a similar manner as https://github.com/OCA/OpenUpgrade/pull/2321/files#diff-731049fb90a841c7cf2ae3ec345b644ab76cffd65af083c32f9ac5dbbbba08b4R15

pedrobaeza (changes requested):

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
AMhadhbi: I marked as : # DONE in im_livechat : renamed to im_livechat.im_livechat_channel_data, right?

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
AMhadhbi: @pedrobaeza i added this function
Screenshot 2021-02-10 at 20 52 34
website.visitor is new model so why we need to update it? i think there is no changes needed! or should i insert into table website_visitor with livechat_operator_id and livechat_visitor_id
Thanks.

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: I would say that `# NOTHING TO DO: Renamed to im_livechat.im_livechat_channel_data in im_livechat`

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
pedrobaeza: OK, I see, thanks for the insight. No need of doing anything indeed, as it's a new model, so no records in it, and guessing them is not suitable.

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
AMhadhbi: Done.

addons/website_livechat/migrations/13.0.1.0/openupgrade_analysis_work.txt
AMhadhbi: Thanks for your feedback.

pedrobaeza (approved):

PR #2537, opened 2021-02-07
[11.0][OU-FIX] #2339, purge obsolete models and fields

### LINKED ISSUE #2339: [13.0] duplicate key when renaming needaction_partner_ids to notified_partner_ids

StefanRijnhart:
Fixes #2339

* add loaded models and fields to the set of loaded XMLIDs
* purge models and fields with noupdate NULL instead of FALSE

Also

* Fix update_fields_xmlid searching for fields by ir.model.fields instead of
the actual model that the field belongs to
* fix KeyError in case of a missing model in ir.model.fields' unlink
* remove coverage as it is giving out red marks to PRs for no reason
* check reference count before deleting a record after module upgrade
* remove ir.model.relation from ir.model's unlink

Also missing was marking the XMLIDs of fields and models as loaded in the
first place.

another.

### DIFF
--- a/odoo/addons/base/ir/ir_model.py
+++ b/odoo/addons/base/ir/ir_model.py
@@ -837,15 +837,16 @@ class IrModelFields(models.Model):
         # Given that we arrive here in order of inheritance, we simply check
         # if the field's xmlid belongs to a module already loaded, and if not,
         # update the record with the correct module name.
+        model = self.env[field.model_name]
         self.env.cr.execute(
             "SELECT f.*, d.module, d.id as xmlid_id, d.name as xmlid "
             "FROM ir_model_fields f LEFT JOIN ir_model_data d "
-            "ON f.id=d.res_id and d.model='ir.model.fields' WHERE f.model=%s",
-            (field.model_name,))
+            "ON f.id=d.res_id and d.model='ir.model.fields' WHERE f.name=%s AND f.model=%s",
+            (field.name, field.model_name,))
         for rec in self.env.cr.dictfetchall():
             if 'module' in self.env.context and\
                     rec['module'] and\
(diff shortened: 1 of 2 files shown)


odoo/addons/base/ir/ir_model.py
```diff
@@ -829,7 +841,7 @@ def _reflect_field(self, field):
             "SELECT f.*, d.module, d.id as xmlid_id, d.name as xmlid "
             "FROM ir_model_fields f LEFT JOIN ir_model_data d "
             "ON f.id=d.res_id and d.model='ir.model.fields' WHERE f.model=%s",
-            (self._name,))
+            (field.model_name,))
```
StefanRijnhart: @pedrobaeza I think this is the source of the problems you observed with fields not being moved to their new home module properly. My mistake when porting the framework to 11.0. Sorry for the inconvenience!

pedrobaeza: No worries. The important thing is that it gets fixed in the best possible way.

pedrobaeza (approved):

StefanRijnhart:
Observed data loss so I'm investigating.

StefanRijnhart:
The observed data loss was a dropped field definition of res.partner::property_stock_customer and the related ir.property entries, amongst similar problems. The problem was due to a number of additional problems due to incorrect porting of this method after it was moved from the field's related model to ir.model.fields itself in Odoo 11.0. Fixed in the additional commit. The problem was specific to OpenUpgrade 11.0 and does not occur in the other branches related to #2339.

hbrunn (approved):

PR #2538, opened 2021-02-07
[12.0][OU-FIX] #2339, set noupdate of ir.model.fields XMLIDs to FALSE

### LINKED ISSUE #2339: [13.0] duplicate key when renaming needaction_partner_ids to notified_partner_ids

StefanRijnhart:
Fixes #2339

Also

* add loaded models to the set of loaded XMLIDs
* purge models with noupdate NULL instead of FALSE
* Remove redundant update_fields_xmlid method and add tests
* fixes broken migration tests mechanism
* fix unreleased savepoint
* remove coverage as it is giving out red marks to PRs for no reason
* remove ir.model.relation from ir.model's unlink

Also missing was marking the XMLIDs of fields and models as loaded in the
first place.

another.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,8 @@ script:
 # only flake8 migration scripts from the openupgrade project, presumably
 # identifyable by using the openupgrade helpers
     - flake8 --max-line-length=120 scripts $(find . \( -name 'pre-*.py' -or -name 'post-*.py' \) -exec grep -q openupgrade {} \; -print)
-    - flake8 odoo/addons/*/migrations/*/tests/ --max-line-length=120
-    - flake8 addons/*/migrations/*/tests/ --max-line-length=120
+    - flake8 odoo/addons/*/migrations/tests/ --max-line-length=120 --exclude=__init__.py
+    - flake8 addons/*/migrations/tests/ --max-line-length=120 --exclude=__init__.py
     - npm install -g less less-plugin-clean-css
     - createdb $DB
     - wget -q -O- https://github.com/OCA/OpenUpgrade/releases/download/databases/11.0.psql | pg_restore -d $DB --no-owner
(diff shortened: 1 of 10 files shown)


MiquelRForgeFlow:
@StefanRijnhart It seems good :S if these chain of PRs is ready, then let's merge them.

One question. Does v14 needs this kind of fix?

StefanRijnhart:
Thanks! All PRs are ready indeed and I'm running my migrations with this code currently. As this set of branches basically backports functionality from Odoo 13, no changes are necessary beyond that release.

MiquelRForgeFlow:
@pedrobaeza let's do merges then?

pedrobaeza:
OK, I haven't checked with detail, but having tests and coming from you, it should be OK.

pedrobaeza:
> As this set of branches basically backports functionality from Odoo 13, no changes are necessary beyond that release.

Then what about #2539 that was for 13.0?

StefanRijnhart:
> Then what about #2539 that was for 13.0?

Fixing earlier OpenUpgrade customizations, including the removal of `update_field_xmlid` which was already taken care of by core Odoo, plus setting `noupdate` to `False` instead of `NULL` for `ir.model` entries in `ir_model_data` to align with the default in Odoo 13.0

pedrobaeza:
So that's not needed in v14?

StefanRijnhart:
No, because it's the default in Odoo 13 already so after we aligned preexisting entries in OpenUpgrade 13, Odoo can take care of removing models itself, like it could take care of removing fields itself in Odoo 12.

pedrobaeza:
Understood. One less thing to care about. Thanks for the explanations.

pedrobaeza:
And of course for the work :smiley:

PR #2539, opened 2021-02-07
[13.0][OU-FIX] #2339, set noupdate of ir.model XMLIDs to FALSE

### LINKED ISSUE #2339: [13.0] duplicate key when renaming needaction_partner_ids to notified_partner_ids

StefanRijnhart:
Fixes #2339

Also
    
* Remove redundant update_fields_xmlid method and add tests
* fixes broken migration tests mechanism
* fix unreleased savepoint

Also missing was marking the XMLIDs of fields and models as `loaded` in the
first place.

another.


MiquelRForgeFlow:
> However, this mechanism was not applied to field and model entries because
> their XMLIDs were created with noupdate NULL instead of FALSE and as such
> excluded in the query to gather all obsolete data records.

That's not true in v13, because https://github.com/odoo/odoo/pull/32881 fixed that in v13 and newer.

StefanRijnhart:
@MiquelRForgeFlow the commit message is largely the same for all 6 related branches. I mention version 13.0 explicitely further down as the release in which all of that is working indeed, and the scope of this particular PR is limited to setting noupdate to false where it is still null, plus testing.

StefanRijnhart:
@MiquelRForgeFlow OK, I went all the way and described the actual list of changes per PR. I hope you like it better now!

MiquelRForgeFlow:
Good! Very nice! :+1:

hbrunn (approved):

PR #2543, opened 2021-02-09
[13.0][ADD] sale_stock: Nothing to do

pedrobaeza:
@Tecnativa TT21675


addons/sale_stock/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,17 @@
+---Models in module 'sale_stock'---
+---Fields in module 'sale_stock'---
+sale_stock   / account.invoice          / incoterms_id (many2one)       : DEL relation: account.incoterms
+# NOTHING TO DO: This field was duplicated in the v12 DB schema, and it was hidden in views, so no possible values in it
```
kos94ok-3D, line 4: Now a model `account.move` and an `invoice_incoterm_id` field so was done in an `account` module

pedrobaeza: Not really. That's another field. An upcoming v12 patch will "merge" the values of both fields where it belongs.

pedrobaeza: See #2544

kos94ok-3D: Oh, I see. Thanks.

kos94ok-3D (approved):

MiquelRForgeFlow (approved):

pedrobaeza:
Again I have forgotten about the dependency... but I'm reviewing it.

PR #2544, opened 2021-02-09
[12.0][FIX] sale_stock: Move incoterms_id content to incoterm_id

pedrobaeza:
There were both fields in this version, one coming from account, and the other being here, but this one was hidden for solving the inconsistency, but it can contain values from previous versions, so we move them to the valid field meanwhile there's no other value in the target field.

@Tecnativa TT21675


MiquelRForgeFlow (approved):

kos94ok-3D (approved):

AaronHForgeFlow (approved):
LGTM!

PR #2546, opened 2021-02-09
[13.0][IMP-OU] account: improve performance creating amount_tax_company_signed

MiquelRForgeFlow:
I think this is kind of obvious: if the tax is 0.0, then don't need to calculate it's value with convert(tax) * sign

--


pedrobaeza (approved):

pedrobaeza:
But this is `IMP`, not `FIX`, as it was previously working

kos94ok-3D (approved):

addons/account/migrations/13.0.1.1/pre-migration.py
```diff
@@ -105,6 +106,12 @@ def create_account_invoice_amount_tax_company_signed(env):
                 SET amount_tax_company_signed = %s WHERE id = %s""",
                 (company_tax, _id),
             )
+        openupgrade.logged_query(
```
MiquelRForgeFlow, line 109: Oh, shit, this query is stupid Y_Y

PR #2554, opened 2021-02-10
[13.0][MIG] website_form

bishalgit:
Nothing to do in this module.

--


addons/website_form/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,21 @@
...
+NEW ir.ui.view: website_form.ir_model_fields_view
+NEW ir.ui.view: website_form.ir_model_view
+NEW ir.ui.view: website_form.register_s_website_form
+NEW ir.ui.view: website_form.remove_external_snippets
+NEW ir.ui.view: website_form.s_website_form
+NEW ir.ui.view: website_form.snippet_options
+# NOTHING TO DO: new xmls
+
+NEW website.page: website_form.contactus_thanks [renamed from website_crm module]
+# NOTHING TO DO: noupdate=0: moved from website_crm to website_form
```
MiquelRForgeFlow, line 21: No, here you need in pre-migration call the rename_xmlids method

MiquelRForgeFlow (changes requested):

addons/website_form/migrations/13.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: sorry, I said nothing

pedrobaeza: This requires some work for not losing custom "Thank you" pages, due to the COW (Create On Write) system. You have to:

- Effectively rename XML-IDs on pre. Although they are noupdate=0, there are some references to the view ID on the COW code, so keeping the same ID will help. This also helps with subsequent inheritance of the regular views.
- Update through a query all the views with the same old `key` field to the new key.

pedrobaeza (changes requested):

addons/website_form/migrations/13.0.1.0/openupgrade_analysis_work.txt
bishalgit: Done

bishalgit:
@pedrobaeza I have made the changes. Can you review it now?

pedrobaeza:
The query for renaming existing ir.ui.views with old `key` to the new one is not yet performed.

bishalgit:
> The query for renaming existing ir.ui.views with old `key` to the new one is not yet performed.

So renaming is not enough? Then should I also rename `key` as well?

pedrobaeza:
As said, there are extra views that are created called "COW views", that are different from the one you are renaming.

pedrobaeza:
As I can't push to corporate branches, I supersede this in #2582 for adding what I'm talking about the key.

PR #2557, opened 2021-02-11
[FIX] sale: don't change prefix of sale sequence

MiquelRForgeFlow:
Sale Orders should maintain name of SOxxxx.

--


addons/sale/migrations/13.0.1.1/noupdate_changes.xml
```diff
@@ -46,6 +46,6 @@
   </record>
   <record id="seq_sale_order" model="ir.sequence">
     <field name="padding">5</field>
```
pedrobaeza: Don't change this as well. Each company put the padding they want.

MiquelRForgeFlow: Done

pedrobaeza (approved):

PR #2558, opened 2021-02-12
[12.0][FIX] barcodes: assure nomenclature_id

MiquelRForgeFlow:
I don't understand why I didn't do it before.

--

### DIFF
--- a/addons/barcodes/migrations/12.0.2.0/pre-migration.py
+++ b/addons/barcodes/migrations/12.0.2.0/pre-migration.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Eficent <http://www.eficent.com>
+# Copyright 2018-21 ForgeFlow <http://www.forgeflow.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
@@ -6,6 +6,7 @@ from openupgradelib import openupgrade
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     cr = env.cr
+    # pre-creation for speed up
     if not openupgrade.column_exists(cr, 'res_company', 'nomenclature_id'):
         openupgrade.add_fields(
             env, [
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
make sense.

marcelsavegnago (approved):
LGTM

addons/barcodes/migrations/12.0.2.0/pre-migration.py
```diff
@@ -14,15 +15,31 @@ def migrate(env, version):
...
-                """
+                    AND spt.barcode_nomenclature_id IS NOT NULL"""
+            )
+
+        cr.execute(
+            """
+            SELECT bn.id
+            FROM ir_model_data imd
+            JOIN barcode_nomenclature bn ON (imd.res_id = bn.id AND imd.model = 'barcode.nomenclature')
+            WHERE imd.module = 'barcodes' AND imd.name = 'default_barcode_nomenclature'"""
```
pedrobaeza: A more elegant SQL is:
```suggestion
            SELECT imd.res_id
            FROM ir_model_data imd
            WHERE imd.model = 'barcode.nomenclature'
                AND imd.module = 'barcodes'
                AND imd.name = 'default_barcode_nomenclature'"""
```
but you can use totally safe in pre-migration this:

```python
env["ir.model.data"].xmlid_to_res_id(
    "barcodes.default_barcode_nomenclature",
    raise_if_not_found=False,
)
```

MiquelRForgeFlow: No, that give an error, because model `barcode.nomenclature` still is not loaded in the env. I just checked that in the beginning, before doing the force pushes.

pedrobaeza: That method is not loading the record. Just returning the `ID`. See the implementation here: https://github.com/odoo/odoo/blob/7aaad8480dde97b2d04b4148e4e54beb57b7f141/odoo/addons/base/models/ir_model.py#L1409

MiquelRForgeFlow: Oh, sorry, I was confused with `xmlid_to_object` method.

StefanRijnhart (approved):

PR #2561, opened 2021-02-13
[13.0][FIX] stock: must delete model data of location_procurement

MiquelRForgeFlow:
In https://github.com/OCA/OpenUpgrade/pull/2519, the location_procurement was forgotten. @kos94ok-3D 

--


kos94ok-3D (approved):

pedrobaeza:
Shouldn't you include such location in the other part of the scripts?

MiquelRForgeFlow:
>  in the other part of the scripts?

Which part? There is nothing we can do with this location

pedrobaeza:
I mean the part `DONE: post-migration: mapped them to new locations`

MiquelRForgeFlow:
Well, in this case this location cannot be mapped to something else :S

pedrobaeza:
Isn't there any replacement for it?

MiquelRForgeFlow:
No, as far I know.

Issue #2412, opened 2020-10-07
[openupgrade analysis] Change NEW record to UPD record for already existing records


MiquelRForgeFlow:
When a module modifies an existing record of a module dependency, it appears in the openupgrade analysis as NEW record of that module, which gives confusion. It wold be better if in this cases, it says UPD instead.

PR #2565, opened 2021-02-15
[13.0][UPD] openupgrade analysis

MiquelRForgeFlow:
- First commit is a normal analysis.
- Second commit implements https://github.com/OCA/OpenUpgrade/issues/2412.
- Third commit is another normal analysis but using previous commit.

UPD indicator means it's not the creation of the record, but an update of the record. So, DEL(UPD) means a deletion of that update in that module and NEW(UPD) means new update of that record in that module.

--


odoo/addons/openupgrade_records/lib/compare.py
```diff
@@ -325,21 +376,37 @@ def match(match_fields, match_type='direct'):
...
+                    potential = 'potentially '
+                entry['potent'] = potential
+                content += ' [%(potent)smoved to %(moved)s module]' % entry
             elif 'renamed' in entry:
                 content += ' [renamed to %(renamed)s module]' % entry
         elif 'new' in entry:
-            content = 'NEW %(model)s: %(name)s' % entry
+            label = 'NEW'
+            if 'upd' in entry:
+                label += '(UPD)'
```
pedrobaeza: I see weird to see `NEW(UPD)`. Better just `UPD`:
```suggestion
                label = 'UPD'
```

MiquelRForgeFlow: And what about the `DEL(UPD)` cases?

StefanRijnhart:
Given that the records need to be managed in the modules that own them rather in the modules that override them, and that we also have the noupdate data xml for every module, what is it that these lines add to the analyis (in the old or the new format)? Can we perhaps just leave them out?

pedrobaeza:
@MiquelRForgeFlow I see what Stefan says as a good proposal.

MiquelRForgeFlow:
@StefanRijnhart @pedrobaeza Done!

StefanRijnhart (approved):
Nice, thanks!

MiquelRForgeFlow:
@StefanRijnhart @pedrobaeza I have added more improvements in the analysis as you can see in the commits. The comparison of m2m columns has proved to be useful!!

StefanRijnhart:
Very nice, thank you!

pedrobaeza:
I suppose we need to add this to the recently migrated https://github.com/OCA/server-tools/pull/1941

addons/mass_mailing/migrations/13.0.2.0/openupgrade_analysis_work.txt
```diff
@@ -45,6 +45,10 @@ mass_mailing / mail.mass_mailing.list_contact_rel / list_id (many2one)
 mass_mailing / mail.mass_mailing.list   / contact_ids (many2many)       : relation is now 'mailing.contact' ('mail.mass_mailing.contact') [nothing to do]
 # NOTHING TO DO: Renamed models, but IDs don't change
 
+mass_mailing / mail.mass_mailing        / contact_list_ids (many2many)  : column1 is now 'mailing_mailing_id' ('mail_mass_mailing_id')
```
pedrobaeza, line 48: For not having to look for the relation table, I think it would be good to include it here in the line where is mentioned the change about the columns.

pedrobaeza (commented):
You can squash last 4 commits by pairs.

addons/mass_mailing/migrations/13.0.2.0/openupgrade_analysis_work.txt
StefanRijnhart: Not sure if that context is actually available.

MiquelRForgeFlow: done

pedrobaeza (approved):

Issue #1733, opened 2019-02-07
Migration scripts for v13


pedrobaeza:
`(del)` means that the module has been removed in this version
`(add)` means that the module has been added in this version
`>` means that it has been identified that the module has been renamed to the name below
`<` means that it has been identified that the module has been rename from the name below
`->` means that the module has been merged in the module below
`<-` means that the module is the target where the module below has been merged 

OCA has funded the development of the migration scripts that are installed in their Odoo instance. They are marked as `(OCA project)` for being recognized and expect them to be done in the near future if not yet done.

- [x] account (OCA project) - By @MiquelRForgeFlow / @kos94ok-3D / @pedrobaeza - #2275 
- [x] account_analytic_default - By @pedrobaeza - #2542 
- [ ] (add) account_analytic_default_hr_expense
- [ ] (add) account_analytic_default_purchase
- [x] account_bank_statement_import (OCA project) - By @pedrobaeza - Nothing to do
- [x] (del) account_cancel -> account (OCA project)
- [ ] account_check_printing
- [ ] (add) account_debit_note
- [x] account_facturx (OCA project) - By @MiquelRForgeFlow - #2434 
- [ ] account_lock
- [x] account_payment (OCA project) - By @pedrobaeza - #2517 
- [ ] account_tax_python
- [ ] account_test
- [x] (del) account_voucher (OCA project) -> account
- [x] analytic (OCA project) - By @MiquelRForgeFlow - #2259
- [ ] association (OCA project)
- [x] (add) attachment_indexation < document 
- [x] auth_ldap - Nothing to do
- [x] auth_oauth - Nothing to do (OCA project)
- [x] auth_password_policy - Nothing to do
- [x] auth_password_policy_signup - By @kos94ok-3D - #2230 
- [x] auth_signup - By @kos94ok-3D - #2230 
- [x] barcodes - By @kos94ok-3D - #2232 
- [x] base - By @kos94ok-3D, @StefanRijnhart, @MiquelRForgeFlow and @pedrobaeza 
- [x] base_address_city - Nothing to do
- [x] base_address_extended - Nothing to do
- [x] base_automation (OCA project) - By @MiquelRForgeFlow - #2437 
- [x] base_gengo - By @kos94ok-3D - #2233 
- [x] base_geolocalize - By @kos94ok-3D - #2234 
- [x] base_iban (OCA project) - By @MiquelRForgeFlow - #2435 
- [x] base_import - Nothing to do
[...]

PR #2567, opened 2021-02-16
[13.0][ADD] website_rating

bishalgit:
Nothing to do.

--


sergio-teruel:
Hi @bishalgit Mybe you can set publisher_id and publisher_datetime in a post-migration like this

UPDATE rating_rating
SET publisher_id = rp.id,
        publisher_datetime = rr.write_date
FROM rating_rating rr
        LEFT JOIN res_users ru ON rr.write_uid = ru.id
        LEFT JOIN res_partner rp ON ru.partner_id = rp.id

What do you think?

pedrobaeza:
It's a new feature for allowing to publish an official reply from the publisher, so it's nothing to do, although this explanation is good to be added to the analysis work file.

bishalgit:
@pedrobaeza the documentation in [issue 1733](https://github.com/OCA/OpenUpgrade/issues/1733#) has not been updated. Can you check it again?

PR #2568, opened 2021-02-16
[13.0][MIG] website_partner: Migration script

sergio-teruel:
cc @Tecnativa TT19853


legalsylvain (approved):

pedrobaeza:
Next time, please don't rename the migration folder to other version number. It must match the version in the manifest, whatever it is (yeah, I know that `0.1` is an odd version, but moving scripts to `1.0` will make this to not work, being higher than the target manifest version).

Anyway, there's nothing to do here in fact. Although the fields have been moved from one module to another, in any case that there's something to do, it must be done on `website`, not here, as in the dependency chain, the other is loaded first, so any effect this would have, it's not valid, as the other module will act first. But in a discussion with @StefanRijnhart in https://github.com/OCA/openupgradelib/commit/6a4360b353986cd2e606b7d1dcc2abfa74f683a6#commitcomment-46485030, we saw that Odoo already handles this kind of field module re-allocation correctly in core, and that's why it's nothing to do.

PR #2569, opened 2021-02-16
[13.0][MIG] website_payment: Migration script

sergio-teruel:
cc @Tecnativa TT19854

Nothing to do


legalsylvain (approved):

pedrobaeza:
Being a bit picky, the view XML-ID renaming should be covered for the case where you have manual views added directly in DB inheriting the original one, so when the final cleaning happens at the end of the upgrade process (with regular Odoo), that view can't be removed due to the not null constraint in `inherit_id` field, breaking the process. I know it's very remote possibility that you have such manual views, but better to prevent it, as it has an easy solution. I have pushed a version with this XML-ID renaming.

PR #2570, opened 2021-02-16
[13.0][MIG] account_check_printing: Migration script

sergio-teruel:
cc @Tecnativa TT27460


legalsylvain (approved):
LGTM. 

naive question : what happens if you don't handle the changes (from integer to char) ? the value will be lost and not converted implicitely ? 

could be great to have openupgradelib function to make the conversion more easy.

``openupgradelib.map_integer_to_char(env, 'account_journal', 'check_next_number')``

to do : 

```
    openupgrade.logged_query(
        env.cr, """
        UPDATE account_journal
        SET check_next_number = '' || %s
        WHERE %s IS NOT NULL
        """, (openupgrade.get_legacy_name('check_next_number'),
              openupgrade.get_legacy_name('check_next_number'))
    )
```

don't you think ?

sergio-teruel:
@legalsylvain Yes... I agree with you.
I not sure if I would be convert column type directly like this
https://github.com/OCA/OpenUpgrade/blob/13.0/addons/account/migrations/13.0.1.1/pre-migration.py#L60

pedrobaeza:
I function in openupgradelib can be interesting, but this happens in so few occasions...

@sergio-teruel you pushed a file here from `website_crm`. I have already removed it. Please take care also to see if there's noupdate changes, like in this case (the files are already populated for all OpenUpgrade modules). Finally, check the usual syntax we use for the work file:

- Each changeset with an extra empty line for better recognition, as there's no syntax highlighting in this file.
- DONE / NOTHING TO DO in upper for the same recognition.
- Mention if the change is on pre, post or end migration.
- If a changeset requires changes in several stages, mention them in separate lines.

Pushed the resulting changes. Waiting to the results for merging.

PR #2574, opened 2021-02-17
[13.0][MIG] website_sale_management: Migration script

sergio-teruel:
cc @Tecnativa TT22094

Nothing to do. In this module not exists openupgrade_analysis.txt file...


legalsylvain (commented):
Hi @sergio-teruel. Did you ``meld`` the code ? maybe the absence of the file is an error ?

sergio-teruel:
@legalsylvain I always create my new branch from upstream. I am seeing that in v13.0 the file not exits.
You can see here
https://github.com/OCA/OpenUpgrade/tree/13.0/addons/website_sale_management

legalsylvain (approved):
I'm ok with that ! Just saying that maybe, there was an error in Openupgrade (analysis file was missing), and maybe there was work to do.
but I quickly meld V12 and V13 folders and there is no diff. (except in ``l10n`` folder)

image

Just weird that the openupgrade_analysis.txt is missing...

pedrobaeza (approved):

PR #2579, opened 2021-02-17
[13.0][OU-FIX] account: consider anglo_saxon_accounting

MiquelRForgeFlow:
Anglo Saxon accounting creates additional moves lines that should have `exclude_from_invoice_tab = TRUE`.

See https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/account_move.py#L49 and https://github.com/odoo/odoo/blob/13.0/addons/stock_account/models/account_move.py#L109. (you can find equivalent methods in previous versions)

Example of additional move lines: https://github.com/odoo/odoo/blob/12.0/addons/stock_account/models/product.py#L343

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -221,6 +221,7 @@ def migration_invoice_moves(env):
         FROM account_invoice_line ail
             JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
             JOIN account_move am ON ail.invoice_id = am.old_invoice_id
+            JOIN res_company rc ON ai.company_id = rc.id
         """
     # We assign to the move lines information from the matching invoice line.
     # Everything that has a tax_line_id (originator tax) is by definition originating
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):

sergio-teruel (approved):

carlosdauden (approved):

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -249,6 +250,8 @@ def migration_invoice_moves(env):
     openupgrade.logged_query(
         env.cr, query + minimal_where + """
             AND aml.old_invoice_line_id IS NULL
+            AND (rc.anglo_saxon_accounting = FALSE
+                OR rc.anglo_saxon_accounting IS NULL)
```
pedrobaeza: A more elegant way is:
```suggestion
            AND rc.anglo_saxon_accounting IS DISTINCT FROM True
```

MiquelRForgeFlow: `a IS DISTINCT FROM b` is doing `((a <> b OR a IS NULL OR b IS NULL) AND NOT (a IS NULL AND b IS NULL))`, thus performance-wise it should not be recommended to use that when you have b = True, as you are introducing silly comparisons.

pedrobaeza: That's not true with constants, and the planner avoid the "silly" parts.

MiquelRForgeFlow: Good to know! Done.

PR #2583, opened 2021-02-20
[13.0][FIX] website: COW resetting must be done on end-migration

pedrobaeza:
So that other inherited modules like `website_form`, `website_crm` or so get benefit from it.

@Tecnativa TT19845


gdgellatly (approved):

cubells (approved):
:+1: 

LGTM

PR #2586, opened 2021-02-21
[13.0][FIX+IMP] account: No error on follower duplication + less verbose on such queries

pedrobaeza:
* If a move from an invoice has one follower, and the invoice has the same one, when doing the conversion, the scripts rose an error due to follower duplication. Current query avoids this at the cost of a bit less performance.
* We don't log such transformation queries, as they don't add too much value for debugging (if an error raises like this one, you will see the offending query in log).

@Tecnativa TT19674

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -1,6 +1,8 @@
 # Copyright 2020 ForgeFlow <http://www.forgeflow.com>
+# Copyright 2021 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+from psycopg2 import sql
 
 
 def fill_account_reconcile_model_second_analytic_tag_rel_table(env):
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -580,17 +582,25 @@ def _move_model_in_data(env, ids_map, old_model, new_model):
...
-                WHERE {field1} = '{old_value1}' AND {field2} = {old_value2}
-                """.format(
-                    table=rename[0], field1=rename[1], field2=rename[2],
-                    old_value1=old_model, new_value1=new_model,
-                    old_value2=old_id, new_value2=new_id,
-                )
-            )
+                SET {field1} = %(new_value1)s, {field2} = %(new_value2)s
+                WHERE {field1} = %(old_value1)s AND {field2} = %(old_value2)s"""
+            if rename[0] == 'mail_followers':
```
MiquelRForgeFlow, line 589: To avoid the loss of performance, maybe you could do it by not adding this specific part here, and then later, in the end of `migration_voucher_moves(env)` script, do a cleanup of repeated followers?

MiquelRForgeFlow (approved):

addons/account/migrations/13.0.1.1/post-migration.py
pedrobaeza: That can't be possible, as the unique constraint will raise then and break the upgrade process.

PR #2588, opened 2021-02-22
[13.0][OU-MIG] website_event

MiquelRForgeFlow:
`# NOTHING TO DO
`

--


jcdrubay (approved):

addons/website_event/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,35 @@
+---Models in module 'website_event'---
+---Fields in module 'website_event'---
+website_event / event.event              / cover_properties (text)       : NEW hasdefault
```
pedrobaeza, line 3: Are the default cover properties the equivalent visualization to what Odoo did in v12, or you should change them to mimic the previous behavior for not having visual differences?

MiquelRForgeFlow: I don't know.

pedrobaeza: Then you should investigate it. If not, the PR is not complete.

MiquelRForgeFlow: I investigated and I think it's fine as it is, nothing to do. That thing it's more like of a new feature, only used for background-images.

pedrobaeza: Then a blog cover in v12 is seen the same in v13, no matter the properties set in v12?

MiquelRForgeFlow:
```python
event_cover_properties = json.loads(self.cover_properties)
# background-image might contain single quotes eg `url('/my/url')`
res['default_opengraph']['og:image'] = res['default_twitter']['twitter:image'] = event_cover_properties.get('background-image', 'none')[4:-1].strip("'")
```

that's the only use of event cover_properties, the twitter image, and in v12 I don't see that was filled, so it's pretty something new.

pedrobaeza (approved):
OK, thanks for confirming

PR #2589, opened 2021-02-23
[12.0][RFR] mrp: force set products as consumable

StefanRijnhart:
Keeping up to date with the latest and the greatest (https://github.com/odoo/odoo/commit/3d34d5838)

--

### DIFF
--- a/addons/mrp/migrations/12.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/12.0.2.0/post-migration.py
@@ -36,6 +36,36 @@ def fill_stock_warehouse_picking_types(env):
             warehouse._create_or_update_sequences_and_picking_types())
 
 
+def force_set_products_consumable(env):
+    """Set products are only allowed to be consumable
+    as per https://github.com/odoo/odoo/commit/3d34d5838.
+    Also delete any quants.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DELETE FROM stock_quant WHERE product_id IN (
+            SELECT pp.id FROM product_product pp
+            JOIN product_template pt ON pt.id = pp.product_tmpl_id
+            WHERE pt.type = 'product' AND EXISTS(
+                SELECT * FROM mrp_bom mb
+                WHERE mb.type = 'phantom' AND mb.active
+                    AND (mb.product_tmpl_id = pt.id OR pp.id = mb.product_id)))
+        """)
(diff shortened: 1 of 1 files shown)


addons/mrp/migrations/12.0.2.0/post-migration.py
```diff
@@ -36,6 +36,23 @@ def fill_stock_warehouse_picking_types(env):
...
+        """
+        UPDATE product_template SET type = 'consu'
+        WHERE id IN (
+            SELECT pt.id FROM product_product pp
+            JOIN product_template pt ON pt.id = pp.product_tmpl_id
+            WHERE pt.type = 'product' AND EXISTS(
+                SELECT * FROM mrp_bom mb
+                WHERE mb.type = 'phantom' AND mb.active
+                    AND (mb.product_tmpl_id = pt.id OR pp.id = mb.product_id)))
+        """)
```
pedrobaeza, line 66: For only writing those needed (it can be also an indicator of what was wrong configured):
```suggestion
        AND type != 'consu'""")
```

pedrobaeza (commented):
It's a good improvement, but a bit late (by Odoo, not by you). I think quants should be removed for that possible products, don't you think?

addons/mrp/migrations/12.0.2.0/post-migration.py
StefanRijnhart: I'm reverting your suggestion because I already had pt.type = 'product' in the sub clause.

StefanRijnhart:
Removing quants as well. Thanks for the suggestion!

addons/mrp/migrations/12.0.2.0/post-migration.py
pedrobaeza: Is it possible to have previously the product as service?

StefanRijnhart: If anybody configures their set products as services they probably have a good reason to do so, and in that case they may override the [method that enforces the product type](https://github.com/odoo/odoo/commit/3d34d5838#diff-cadd7a2530955f21c93bbe4372640d0bbb4afce512c9a2b0b6f0dbbc96585096R83), I imagine ;-)

pedrobaeza (approved):
Please squash commits

StefanRijnhart:
Squashed

PR #2594, opened 2021-02-24
[13.0][OU-FIX] mass_mailing: update mailing_domain

MiquelRForgeFlow:
Updating mailing_domain filter.

--

### DIFF
--- a/addons/mass_mailing/migrations/13.0.2.0/post-migration.py
+++ b/addons/mass_mailing/migrations/13.0.2.0/post-migration.py
@@ -8,7 +8,26 @@ _unlink_by_xmlid = [
 ]
 
 
+def update_mailing_domain_filter(env):
+    if openupgrade.table_exists(env.cr, 'account_invoice'):
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE mailing_mailing
+            SET mailing_domain = replace(mailing_domain,
+                '''customer''', '''customer_rank''')
+            WHERE mailing_domain LIKE '%''customer''%'""",
+        )
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE mailing_mailing
+            SET mailing_domain = replace(mailing_domain,
+                '''supplier''', '''supplier_rank''')
+            WHERE mailing_domain LIKE '%''supplier''%'""",
+        )
+
+
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     openupgrade.delete_records_safely_by_xml_id(env, _unlink_by_xmlid)
+    update_mailing_domain_filter(env)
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):

pedrobaeza:
This seems very specific, as this field renaming is one of a lot. A better approach is to include this as part of the `rename_fields` method in openupgradelib.

Anyway, I would say that this is not even correct, as previous field was boolean, and the new one is an integer.

MiquelRForgeFlow:
> A better approach is to include this as part of the rename_fields method in openupgradelib.

how? :S

StefanRijnhart:
Ouch I did not think about the field type. That would leave the domain invalid even after this update.

And rename_fields would not have helped as it is not called on the change from supplier to supplier_rank (for the same reason that the type is different I imagine).

StefanRijnhart:
Not sure what such domains look like but maybe you can try to deserialize into a list and replace by member? Maybe do that in the module that defines these fields in case mass_mailing is installed? And maybe do that for other domains as well?

MiquelRForgeFlow:
I don't see the problem. If in that domain you have "`... ('customer', '=', True) ...`", and you change that by "`... ('customer_rank', '=', True) ...`", then it's still valid, right? I mean, you don't bother if customer_rank is 1, 2 or any value. You only care that it's not 0.

pedrobaeza:
@MiquelRForgeFlow tell that to the ORM, who is the one that should allow that. I don't know right now if it won't fail.

Anyway, I still insist this is very specific, and that domain is not a default that is added by the module (as we needed in v12). I wouldn't go for this here.

About `rename_fields`, you can add to the routines of the renamings to search into this field if the table and column exists. I know that this doesn't work for this specific change, but it's a good improvement for the rest and for avoiding support tickets after migration.

StefanRijnhart:
Odoo itself uses [('supplier_rank','>', 0)]. It seems wise to adhere to that or it will bite some people later on when they start assigning negative ranking (using it as some kind of credit score).

MiquelRForgeFlow:
> Maybe do that in the module that defines these fields in case mass_mailing is installed?

Why? I think the good logic is doing it here in mass_mailing, that's where the field you are modifying is.

> And maybe do that for other domains as well?

Which other domains? If there are others in other modules, then we should attend them in their corresponding modules.

> or it will bite some people later on when they start assigning negative ranking

But negative values don't make sense. This ranks values are counting things. Always.

> tell that to the ORM

What? that doesn't fail. integer ~ boolean comparison don't give errors.

> this is very specific

Yeah, that's a bit true. I can close the PR if you want and use the branch as my custom code, no problem.

MiquelRForgeFlow:
Added in my docker =)

PR #2602, opened 2021-03-09
[13.0][IMP] stock: Populate lot company in a better way

pedrobaeza:
No need to do such costly operation, but going to stock.move info.

As fallback for manually created lots, we get the company of the product or the user's company.

@Tecnativa TT27487

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -72,12 +72,19 @@ def fill_company_id(cr):
     openupgrade.logged_query(
         cr, """
         UPDATE stock_production_lot spl
-        SET company_id = COALESCE(
-            (SELECT sml.company_id FROM stock_move_line sml
-             WHERE sml.lot_id = spl.id AND sml.company_id IS NOT NULL
-             LIMIT 1), ru.company_id)
-        FROM res_users ru
-        WHERE ru.id = spl.create_uid AND spl.company_id IS NULL"""
+        SET company_id = sm.company_id
+        FROM stock_move_line sml
+        JOIN stock_move sm ON sm.id = sml.move_id
+        WHERE sml.lot_id = spl.id AND spl.company_id IS NULL"""
+    )
+    openupgrade.logged_query(
+         cr, """
+         UPDATE stock_production_lot spl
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
LGTM. (code review) 
thanks !

jcdrubay (approved):

pedrobaeza:
I would like to have @MiquelRForgeFlow's opinion to see why he did the other way.

legalsylvain:
Ow sorry, @pedrobaeza.
I just deleted the branch  ``13.0-ocabot-merge-pr-2602-by-legalsylvain-bump-nobump``. 
this must have aborted the bot process.

pedrobaeza:
Thanks.

Another info: I would say that the big doesn't work in 13.0 branch and don't bring too much value in the OU flow, so we merge directly when proceed.

MiquelRForgeFlow (approved):
I think I did that way to assure that if one cannot find a suitable company through sml, one could use the company of the create_uid, and I did that in one query because I was not thinking in the performance. Yeah, it's clever to use the stock.move relation.

pedrobaeza:
I have added as fallback for manually created lots, the company of the product or the user's company.

For reference, before this patch, this query needed 40+ min in a DB, and now it takes 10 seconds.

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -72,12 +72,19 @@ def fill_company_id(cr):
...
+        WHERE sml.lot_id = spl.id AND spl.company_id IS NULL"""
+    )
+    openupgrade.logged_query(
+         cr, """
+         UPDATE stock_production_lot spl
+         SET company_id = COALESCE(pt.company_id, ru.company_id)
+         FROM res_users ru
+         JOIN product_product pp ON pp.id = spl.product_id
+         JOIN product_template pt ON pt.id = pp.product_tmpl_id
+         WHERE ru.id = spl.create_uid AND spl.company_id IS NULL"""
```
MiquelRForgeFlow:
```suggestion
         FROM res_users ru, product_product pp
         JOIN product_template pt ON pt.id = pp.product_tmpl_id
         WHERE pp.id = spl.product_id AND ru.id = spl.create_uid AND spl.company_id IS NULL"""
```

PR #2604, opened 2021-03-10
[11.0][FIX] base: noupdate data from report module

MiquelRForgeFlow:
The `compare_noupdate_xml_records` tool doesn't consider merged modules. The module `report` was merged on `base`, and it had some noupdate data.

--


pedrobaeza (approved):
These values are in the edge of things to be preserved if custom needs, but not too much people customized them, so let's put the new sane default

Issue #2605, opened 2021-03-10
[RFC] Allow a step by step migration during Manual migration development process


jcdrubay:
# Current behavior:

The manual migration described in https://doc.therp.nl/openupgrade/migration_details.html says:

> Expect to repeat the process several times as you encounter errors, clean up your data, and try again.

But running the execution of the `--update all` is quite long making the trial and error process really time consuming.

# Expected behavior:

Be able to run the upgrade restarting from an "intermediate" step just before the previous failure.

From the documentation

> There are three migration stages at which you can have migration scripts executed: before the module is loaded (`pre` stage), after the module is loaded into the database (`post` stage) and when all the modules are loaded (`end` stage). 

Ideally we should be able to run with the following options:

- `--update all`: run the full upgrade
- `--update account:post`: run the upgrade starting from the post stage of the module account, then run the `end` stage for all modules
- `--update account:end`: run the `end` stage for account module and all of the following modules

To be able to achieve this, we would also need to:

- [A] be able to know in which order OpenUpgrade would upgrade all the modules.
- [B] at the end of the successful execution of the upgrade of each module, commit the change in the database
- [C] each migration script must be executable multiple times without getting a different result then if it is executed only one time

This feature [B] should be activated if we are using the option `--dev openupgrade`.

rvalyi:
Hello @jcdrubay there is probably room for improvement in this trial and error workflow as you say.
That being said I have been successful by trying `--update=all` first and then when it fails on a given module some_module, I try to fix the scripts and do a `--update=<some_module>`. I don't know if there is a better way but usually this enables me to fix the some_module scripts. After it is fixed however I go back to my dump and run with `--update=all` because just doing `--update=<some_module>` and letting dependent modules is likely to lead to parial and incomplete migration.

Another hint tro speed things up: I use a Porstgres conf with filesync=off for OpenUpgrade which gives a significant boost (at the cost of zero security with the durability of your data, that is never do that in production).

legalsylvain:
Hi @jcdrubay. 

Regarding ``update account:post`` I don't see the interest. update are done automatically in the correct order. If it breaks during account, you can re-run your migration, and the still updated modules (like ``base``) will not be upgraded again. (just updated)
The only thing i'm not sure is regarding the ``end`` option. 
You can maybe check that PR is related. https://github.com/OCA/OpenUpgrade/pull/2499

kind regards.

jcdrubay:
[...]

PR #2606, opened 2021-03-11
[14.0][DOC]Add details and recommendations to the page Running the migration

jcdrubay:

Adds recommendation in the part `2. Check coverage of the migration scripts for your installed modules`

and extra advice in the part `5. Run and write and run and ... the Upgrade`

--

### DIFF
--- a/doc/source/migration_details.rst
+++ b/doc/source/migration_details.rst
@@ -5,36 +5,126 @@ Check out the code manually and upgrade your database by calling odoo-bin,
 (or openerp-server) directly. You will want to do this when you are working on
 developing migration scripts for uncovered modules.
 
-1. Make the `openupgrade_framework` and the `openupgrade_scripts` modules
-   available in the addons path in the Odoo instance of the new version.
+1. Get the code from OpenUpgrade and dependencies
+*************************************************
 
-   Or, for older versions: check out the OpenUpgrade source code from Github
-   for the branches you need. Each branch migrates to its version from the
-   previous version, so branch 13.0 migrates from 12.0 to 13.0. If you are
-   migrating across multiple versions, you need to run each version of
-   OpenUpgrade in order. Skipping versions is not supported.
+Make the `openupgrade_framework` and the `openupgrade_scripts` modules
+available in the addons path in the Odoo instance of the new version.
 
-2. Check if there are migration scripts provided for the set of modules that
-   are installed in your Odoo database. If there are modules for which no
-   migration scripts have been developed yet, your migration may fail or the
-   integrity of your database may me lacking. Check the module coverage in
-   this documentation.
+Or, for older versions: check out the OpenUpgrade source code from Github
+for the branches you need. Each branch migrates to its version from the
+previous version, so branch 13.0 migrates from 12.0 to 13.0. If you are
+migrating across multiple versions, you need to run each version of
+OpenUpgrade in order. Skipping versions is not supported.
 
-3. Decide which database you are going to upgrade. You absolutely *must* make a
-   backup of your live database before you start this process!
+2. Check coverage of the migration scripts for your installed modules
+*********************************************************************
 
-4. Edit the configuration files and command line parameters to point to the
-   database you are going to upgrade. Run Odoo with the *--update all
-   --stop-after-init --load=web,openupgrade_framework* flags.
+Check if there are migration scripts provided for the set of modules that
+are installed in your Odoo database. If there are modules for which no
+migration scripts have been developed yet, your migration may fail or the
+integrity of your database may me lacking. Check the module coverage in
+this documentation.
 
-   For versions earlier than 14.0 that are running the OpenUpgrade fork rather
-   than Odoo itself, you do not pass the `load` parameter.
+At this stage, for some of the uncovered modules, you might want to uninstall
+them from your database and re-install it later on after the migration.
+This option might reduce the workload of the migration and can be applicable
+for "ux" modules like ``web_widget_image_download`` if you did not
+override this modules.
 
-5. Run the upgrade and check for errors. You will probably learn a lot about
-   your data and have to do some manual clean up before and after the upgrade.
-   Expect to repeat the process several times as you encounter errors, clean up
-   your data, and try again. If necessary, ask for help or report bugs on
-   Github.
+
+3. Make a copy the database to migrate
+**************************************
+
+Decide which database you are going to upgrade. You absolutely *must* make a
+backup of your live database before you start this process!
+
+4. Adjust your Odoo configuration
+*********************************
+
+Edit the configuration files and command line parameters to point to the
+database you are going to upgrade. Run Odoo with the ``--update all
+--stop-after-init --load=web,openupgrade_framework`` flags.
+
+For versions earlier than 14.0 that are running the OpenUpgrade fork rather
+than Odoo itself, you do not pass the `load` parameter.
+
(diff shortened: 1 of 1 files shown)


doc/source/migration_details.rst
```diff
@@ -5,36 +5,126 @@ Check out the code manually and upgrade your database by calling odoo-bin,
...
+are installed in your Odoo database. If there are modules for which no
+migration scripts have been developed yet, your migration may fail or the
+integrity of your database may me lacking. Check the module coverage in
+this documentation.
+
+At this stage, for some of the uncovered modules, you might want to uninstall
+them from your database and re-install it later on after the migration.
+This option might reduce the workload of the migration and can be applicable
+for "ux" modules like ``web_widget_image_download`` if you did not
+override this modules.
```
StefanRijnhart, line 33: This advice seems dangerous. Uninstalling modules from Odoo can cause data loss.

doc/source/migration_details.rst
```diff
@@ -5,36 +5,126 @@ Check out the code manually and upgrade your database by calling odoo-bin,
...
+You can also refer to the following documents:
+
+- `The Odoo Migration Manager
+  <https://doc.therp.nl/openupgrade/migrationmanager.html>`_
+- `OpenUpgrade API <https://doc.therp.nl/openupgrade/API.html>`_
+- the code of OpenUpgrade API
+  `openupgrade.py <https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py>`_
+- `Development FAQ <https://doc.therp.nl/openupgrade/devfaq.html>`_
+
+The Trial and error process for the development of your scripts
```
StefanRijnhart, line 76: The stanzas below are great advice but maybe belong to a `development` section rather than a `runtime` section?

doc/source/migration_details.rst
```diff
@@ -5,36 +5,126 @@ Check out the code manually and upgrade your database by calling odoo-bin,
...
+Try `--update=all` first and then when it fails on a given module
+some_module, try to fix the scripts and do a ``--update=<some_module>``
+
+This enables fixing and testing the some_module scripts
+
+After it is fixed go back to ``--update=all`` because just doing
+``--update=<some_module>`` and letting dependent modules is likely to lead
+to partial and incomplete migration.
+
+``fsync = off`` in your postgresql.conf
```
StefanRijnhart, line 114: I consider this development advice, or at least `severely expert` mode as well because it is incredibly dangerous and I would very much hesitate to encourage people to do this, given that they might use the same database server for migration as for production. We already encounter a lot of cases where less experienced people get lost trying to work with openupgrade. Having them modify database settings like this could increase the collateral damage.

doc/source/migration_details.rst
jcdrubay: Could we place this in a section / page "Dangerous and optional actions with possible benefits"?

Inside that section we could also add the part relate to ``fsync = off``

Or would you prefer to remove it completely?

doc/source/migration_details.rst
jcdrubay: Maybe a new page under "https://doc.therp.nl/openupgrade/development.html" ?

doc/source/migration_details.rst
jcdrubay: Maybe a new page under "https://doc.therp.nl/openupgrade/development.html" ?

pedrobaeza: For me the fsync thing should never be mentioned. If you need it, it's because OpenUpgrade is not well optimized, so we have to focus on that instead.

legalsylvain:
Hi. 
I have an intermediate point of view regarding the ``fsync`` debate. 

- @StefanRijnhart, indeed, this option should not be disabled on production server. But it is, in this PR, very explicit in this text : ``that is never do that in production``. In fact, it is like the whole openupgrade project, that should not be used on production.

- @pedrobaeza : the fact is that OpenUpgrade is not well optimized. Most of the time, migration script are proposed, and then, in a second time, optimized, but between that both steps, it can take several months or even years. One of the reason is that Openupgrade is supported for a part, by the OCA, that pay for upgrading the odoo-community.org instance, that is not a very big instance, so no need to optimize the migration script at this first step. (<1k ``account.move.line``, <10k ``res.partner``, <10k ``product.product``, no stock). Another reason is that ORM does a lot of the work natively, without migration scripts, but can be very slow. (typical example : new computed stored field).  So adding a pre-migration script makes the migration faster for big instance.

Openupgrade can be improved, but in the meantime, if an option can speed up a huge migration, I think it should be mentionned.

- In the other hand, during my last migration, I tried this option and It doesn't solved the problem I had, regarding the fact that the migration was too long. (I stopped the migration before the end, after 36 hours.). I so disabled some scripts (mainly one), by postponing treatments until later. @rvalyi, or @sebastienbeau : I know you talked me about that option. Could you provide specific examples of performance improvements. (what server configuration, what database size, what migration times with and without the ``fsync`` option?)

As an example, the last migration I did was a 8-> 12 migration, the last year. 
- It was a 10 year old database (originally running on OpenERP 5) with a size of 36 GB. (2M ``account.move.line``, 12M ``stock.move``)
- the dedicated server had 96GB of memory and 500GB of SSD. (full spec available [here](https://www.scaleway.com/en/dedibox/pro/pro-4-l/))
- the full migration process took 14 hours. (with disabled scripts). (I didn't disabled ``fsync``).

doc/source/migration_details.rst
StefanRijnhart: Yes, sounds good!

doc/source/migration_details.rst
StefanRijnhart: @jcdrubay yes, for development purposes I myself run a partial migration and then execute
`update ir_module_module set state = 'to remove' where state IN ('to upgrade', 'to install') returning name`
I then run Odoo with `-u dummy --stop-after-init` to uninstall the modules that could not be upgraded yet. That way, you get a partially migrated database that you can login to and perform UI testing. So mentioning this under the `development` section sounds good as long as it is clear that this step does involve data loss.

doc/source/migration_details.rst
StefanRijnhart: Pedro's comment https://github.com/OCA/OpenUpgrade/pull/2606#discussion_r593771503 applies to this section.

rvalyi: yeah, I was mentionning the fsync=off hack for experts only, I don't advise putting fsync=off inside the general OpenUpgrade documentation either. Otherwise some people would end up using the database with that configuration later, get some energy cut and loose their data and cry after us.

legalsylvain: Hi @rvalyi. Sorry I misunderstood [your comment](https://github.com/OCA/OpenUpgrade/issues/2605#issuecomment-795319768) so.
I abstain on that point.
regards.

jcdrubay:
Thanks for all the feedback. I will split this PR into smaller items trying to take into account all feedback.

jcdrubay:
A first PR for the migration_details page here with some small adjustments:
https://github.com/OCA/OpenUpgrade/pull/2621

jcdrubay:
A second PR for the development page here:
https://github.com/OCA/OpenUpgrade/pull/2624

jcdrubay:
Final PR with the dangerous tips:
https://github.com/OCA/OpenUpgrade/pull/2638

PR #2611, opened 2021-03-16
[10.0][IMP] Set invoices paid

StefanRijnhart:
In Odoo 9.0, an invoice was moved from state 'open' to state 'paid'
on reconciliation of the move line that was linked as the trigger of
payment subflow in the invoice workflow. This flow usually prevented
zero amount invoices from getting into state `paid` because zero amount
move lines are usually not reconciled. The invoice would be marked as
`reconciled` correctly, because there was no residual amount.
In Odoo 10.0, any invoice that is marked as `reconciled` will be moved to
state `paid` as per `action_invoice_paid` that is triggered from the
model's `_write` method<sup>1</sup>. Preemtively set the correct state using SQL here
for performance reasons.

<sup>1</sup> https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L365

--

### DIFF
--- a/addons/account/migrations/10.0.1.1/pre-migration.py
+++ b/addons/account/migrations/10.0.1.1/pre-migration.py
@@ -1,7 +1,26 @@
 # -*- coding: utf-8 -*-
 # © 2017 Therp BV
+# © 2021 Opener B.V.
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

MiquelRForgeFlow (approved):
LGTM

PR #2613, opened 2021-03-17
[13.0][MIG] stock: Recompute complete_name in stock locations.

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -319,6 +319,16 @@ def map_stock_locations(env, main_company):
     """)
 
 
+def recompute_stock_location_complete_name(env):
+    # In 12.0, the displayed name of the locations was obtained by the
+    # name_get method. This method limited the location path up to a parent
+    # location of type view is found. In 13.0, complete_name is used for
+    # this purpose, and this one used to be stored with full path in v12.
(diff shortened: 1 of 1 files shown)


addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -319,6 +319,11 @@ def map_stock_locations(env, main_company):
     """)
 
 
+def recompute_stock_location_comple_name(env):
```
legalsylvain: typo : ``recompute_stock_location_comple_name`` --> ``recompute_stock_location_complete_name``

legalsylvain, line 322: typo. complete_name.

legalsylvain (commented):
Hi. thanks for the patch. could you mention why complete_name should be recomputed ? 

thanks !

sergio-teruel:
Hi @legalsylvain Method comment added....
Explanation:
In 12.0, the displayed name of the locations was obtained by the name_get method. This method limited the location path up to a parent location of type view is found. In 13.0, complete_name is used for this purpose, and this one used to be stored with full path in v12. We need to recompute it.

chienandalu (approved):

StefanRijnhart (approved):

kos94ok-3D (approved):

PR #2614, opened 2021-03-17
[13.0][OU-FIX] account: avoid two or more move lines pointing to same invoice line

MiquelRForgeFlow:
If we don't fix this, we could hypothetically have several move lines pointing to the same invoice line (because two invoice lines can have all the same values except id and sequence), and thus several moves lines with the same sequence.

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -238,24 +238,42 @@ def migration_invoice_moves(env):
             AND ((ail.product_id IS NULL AND aml.product_id IS NULL) OR ail.product_id = aml.product_id)
             AND ((ail.uom_id IS NULL AND aml.product_uom_id IS NULL) OR ail.uom_id = aml.product_uom_id)
         """
-    # Try first with a stricter criteria for matching invoice lines with account move lines
-    openupgrade.logged_query(
-        env.cr, query + minimal_where + """
-            AND ail.account_id = aml.account_id
-            AND ai.commercial_partner_id = aml.partner_id
-            AND ((ail.account_analytic_id IS NULL AND aml.analytic_account_id IS NULL)
-                OR ail.account_analytic_id = aml.analytic_account_id)
-        RETURNING aml.id""",
-    )
-    aml_ids = tuple(x[0] for x in env.cr.fetchall())
-    # Try now with a more relaxed criteria, as it's possible that users change some data on amls
-    openupgrade.logged_query(
-        env.cr, query + minimal_where + """
-            AND aml.old_invoice_line_id IS NULL
-            AND rc.anglo_saxon_accounting IS DISTINCT FROM TRUE
-        RETURNING aml.id""",
-    )
-    aml_ids += tuple(x[0] for x in env.cr.fetchall())
+    # Doing loop to fix where two or more move lines points to the same old invoice line
+    while True:
+        # Try first with a stricter criteria for matching invoice lines with account move lines
+        openupgrade.logged_query(
+            env.cr, query + """
+            LEFT JOIN account_move_line aml2 ON aml2.old_invoice_line_id = ail.id
+            """ + minimal_where + """
+                AND ail.account_id = aml.account_id
+                AND ai.commercial_partner_id = aml.partner_id
+                AND ((ail.account_analytic_id IS NULL AND aml.analytic_account_id IS NULL)
+                    OR ail.account_analytic_id = aml.analytic_account_id)
+                AND aml.old_invoice_line_id IS NULL AND aml2.id IS NULL
+            RETURNING aml.id""",
+        )
+        # Try now with a more relaxed criteria, as it's possible that users change some data on amls
+        openupgrade.logged_query(
+            env.cr, query + """
+                LEFT JOIN account_move_line aml2 ON aml2.old_invoice_line_id = ail.id
+                """ + minimal_where + """
+                AND aml.old_invoice_line_id IS NULL AND aml2.id IS NULL
+                AND rc.anglo_saxon_accounting IS DISTINCT FROM TRUE
+            RETURNING aml.id""",
+        )
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE account_move_line aml
+            SET old_invoice_line_id = NULL
+            FROM (SELECT id, ROW_NUMBER() OVER (PARTITION BY old_invoice_line_id ORDER BY id) repeated_invoice_line
+                  FROM account_move_line
+                  WHERE old_invoice_line_id IS NOT NULL) aml2
+            WHERE aml.id = aml2.id AND aml2.repeated_invoice_line > 1
+            RETURNING aml.id""",
+        )
+        wrong_aml_ids = tuple(x[0] for x in env.cr.fetchall())
+        if not wrong_aml_ids:
+            break
     # 2st: exclude from invoice_tab the grouped ones, and create a new separated ones
     openupgrade.logged_query(
         env.cr, """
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -248,10 +248,40 @@ def migration_invoice_moves(env):
...
+                AND aml.old_invoice_line_id IS NULL AND aml2.id IS NULL
+            RETURNING aml.id""",
+        )
+        aml_ids += tuple(x[0] for x in env.cr.fetchall())
     # Try now with a more relaxed criteria, as it's possible that users change some data on amls
     openupgrade.logged_query(
-        env.cr, query + minimal_where + """
-            AND aml.old_invoice_line_id IS NULL
+        env.cr, query + """
+                LEFT JOIN account_move_line aml2 ON aml2.old_invoice_line_id = ail.id
```
pedrobaeza, line 258: Why changing this?

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -248,10 +248,40 @@ def migration_invoice_moves(env):
         RETURNING aml.id""",
     )
     aml_ids = tuple(x[0] for x in env.cr.fetchall())
+    # Fix (exceptional) cases where two or more move lines points to the same old invoice line
+    while True:
```
pedrobaeza, line 242: Shouldn't this loop be in a more wider scope, for applying all possible matches each time?

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: to ensure new matching move lines are not pointing to already pointed invoice lines :S

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: ah, you mean to include the query of relaxed criteria here, right?

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -248,14 +248,62 @@ def migration_invoice_moves(env):
...
+        aml_ids += tuple(x[0] for x in env.cr.fetchall())
     # Try now with a more relaxed criteria, as it's possible that users change some data on amls
-    openupgrade.logged_query(
-        env.cr, query + minimal_where + """
-            AND aml.old_invoice_line_id IS NULL
-            AND rc.anglo_saxon_accounting IS DISTINCT FROM TRUE
-        RETURNING aml.id""",
-    )
-    aml_ids += tuple(x[0] for x in env.cr.fetchall())
+    while True:
```
pedrobaeza: I don't think we should include 2 loops, but all in the same one

MiquelRForgeFlow: Why? The first one is for stricter criteria and the second one is for lesser criteria. If we mix both, it could lead to undesired (although remote) cases. Performance is unaffected by doing separated or mixed way.

pedrobaeza: There's no risk of repeat lines if you also include the join criteria on the second main SQL, but we don't repeat some code and don't do even less readable than it is. Please try to put only one loop.

MiquelRForgeFlow: Now there is only one loop.

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -238,24 +238,63 @@ def migration_invoice_moves(env):
...
+                        OR ail.account_analytic_id = aml.analytic_account_id)
+                    AND aml.old_invoice_line_id IS NULL AND aml2.id IS NULL
+                RETURNING aml.id""",
+            )
+            aml_ids += tuple(x[0] for x in env.cr.fetchall())
+            openupgrade.logged_query(
+                env.cr, """
+                UPDATE account_move_line aml
+                SET old_invoice_line_id = NULL
+                FROM (SELECT id, rank() OVER (PARTITION BY old_invoice_line_id ORDER BY id) repeated_invoice_line
```
pedrobaeza: You are still doing 2 times this query, and this is what I want to avoid. It should be executed only once per loop, as it's very bad at performance. Instead of `rank()`, maybe `row_number()` performs better.

MiquelRForgeFlow: You want to avoid to execute this twice per loop just for performance. But I am more interested in assure that there is not a wrong matched invoice line. If we execute the two matching queries consecutively, it could happen that the one of lesser criteria matches one wrongly.

pedrobaeza: I don't think so with the `LEFT JOIN` with already matched lines. Take into account that this case is minority and I'm not facing it, so for being accepted, this should be optimized the maximum.

MiquelRForgeFlow: changes applied. I hope now it fits your taste :)

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -238,24 +238,46 @@ def migration_invoice_moves(env):
...
+            FROM (SELECT id, ROW_NUMBER() OVER (PARTITION BY old_invoice_line_id ORDER BY id) repeated_invoice_line
+                  FROM account_move_line
+                  WHERE old_invoice_line_id IS NOT NULL) aml2
+            WHERE aml.id = aml2.id AND aml2.repeated_invoice_line > 1
+            RETURNING aml.id""",
+        )
+        wrong_aml_ids = tuple(x[0] for x in env.cr.fetchall())
+        if not wrong_aml_ids:
+            break
+        aml_ids = tuple(x for x in aml_ids if x not in wrong_aml_ids)
```
pedrobaeza: Although see my other suggestion:
```suggestion
        aml_ids = tuple(set(aml_ids) - set(wrong_aml_ids))
```

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -238,24 +238,46 @@ def migration_invoice_moves(env):
...
-    aml_ids = tuple(x[0] for x in env.cr.fetchall())
-    # Try now with a more relaxed criteria, as it's possible that users change some data on amls
-    openupgrade.logged_query(
-        env.cr, query + minimal_where + """
-            AND aml.old_invoice_line_id IS NULL
-            AND rc.anglo_saxon_accounting IS DISTINCT FROM TRUE
-        RETURNING aml.id""",
-    )
-    aml_ids += tuple(x[0] for x in env.cr.fetchall())
+    aml_ids = tuple()
```
pedrobaeza: I don't think we need this variable, as it's only used in one `if` in:

https://github.com/OCA/OpenUpgrade/blob/8d6b578ba656d39c407bcf803f8251b1f2030044/addons/account/migrations/13.0.1.1/post-migration.py#L345

and I think we can run that query nevertheless it, and if not, we can just have a flag for activating it.

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: but your suggestion is more inefficient as far I know

pedrobaeza: No, sets operations are more efficient than list comprehension

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: but we need it because inside the while we are using `aml_ids +=`

pedrobaeza: I don't see its usage. Please check again.

MiquelRForgeFlow: lines 256, 266 and 280

pedrobaeza: That aren't read operations, but write ones, and thus, useless. We can remove them all. That's my point.

MiquelRForgeFlow: ok, done

pedrobaeza (approved):

PR #2617, opened 2021-03-17
[13.0][FIX] workflows: update python libraries

MiquelRForgeFlow:
`python-psycopg2` and `python-werkzeug` don't exist anymore.

--


pedrobaeza (approved):
Thanks for investigating

PR #2622, opened 2021-03-19
[13.0][FIX] mass_mailing: Subject is now filled

pedrobaeza:
In v12, the related utm.source.name was used as subject. Now they are 2 different fields, and we need to fill the new one with the
content of the previous one.

@Tecnativa TT28794

### DIFF
--- a/addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
+++ b/addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
@@ -72,6 +72,29 @@ _xmlid_renames = [
 ]
 
 
+def fill_mailing_subject(env):
+    openupgrade.logged_query(env.cr, "ALTER TABLE mailing_mailing ADD name VARCHAR")
+    openupgrade.logged_query(
+        env.cr, """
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):

chienandalu (approved):

BartvandenHout:
v12 has 'customer' in res.partner. Many of the mailing_domain of the database I just OpenUpgraded have that as a selection which is faulty selection in v13.

addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
```diff
@@ -72,6 +72,29 @@
 ]
 
 
+def fill_mailing_subject(env):
```
jcdrubay, line 75: Is it normal that this function never gets called?
At least, it's not called by this file.

pedrobaeza: Ouch, my bad. Thanks for checking. Added in 40745746543e553ac4b554d58d69c6cd60a4d86d

PR #2624, opened 2021-03-19
[14.0][DOC]Add guidelines for the development process of migration scripts

jcdrubay:
--

### DIFF
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -1,8 +1,128 @@
 Migration script development
 ++++++++++++++++++++++++++++
 
+Overview
+--------
+
+The execution of the migration scripts is triggered by the "Migration Manager":
+:doc:`migrationmanager`.
+
+As explained with more detail in the page :doc:`migrationmanager`, the migration scripts
+of a module are usually the combination of 3 python files (might be less if some
+steps are not necessary):
+
+- pre-migration.py
+- post-migration.py
+- end-migration.py
+
+Since version 14, these files can be found here:
+``openupgrade-framework/openupgrade_scripts/scripts/<module-name>``
+
+Each of those files will contain a function `migrate` which is called by the
+Migration Manager.
+
+Then, the function migrate will call other functions. Each of these functions
+will execute a task of the migration, these functions are the one which
+needs most of the work from developers. These functions are usually declared in the
+same file.
+
+Luckily, many pre-existing functions already exist within `openupgradelib` reducing
+significantly the work of development.
+See `OpenUpgrade API <https://doc.therp.nl/openupgrade/API.html>`_
+
+For many modules, a developer will not even need to write any function, but will
+"simply" need to call pre-existing functions from `openupgradelib` with the appropriate
+arguments. The main complexities becomes to learn what are the functions available in
+`openupgradelib` and then to select appropriatly the arguments (usually according to
+the `openupgrade_analysis.txt`, see below).
+
+For instance, the migration to version 13 of mass_mailing did not require any custom
+function: https://github.com/OCA/OpenUpgrade/pull/2273/files
+
+After this introduction, which highlight how simple the migration scripts can be,
+do not underestimate the power of OpenUpgrade.
+The migration of the account module to version 13 is the kind of situation where
+the full power of OpenUpgrade was unleashed to successfully overcome the
+"acc-pocalypse":
+https://github.com/OCA/OpenUpgrade/pull/2275/files.
+
+
+Learn from existing migration scrips
+------------------------------------
+
+
+Since version 14, the migration scripts are located in:
+``OpenUpgrade/openupgrade-framework/openupgrade_scripts/scripts/<module-name>``
+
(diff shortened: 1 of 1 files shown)


doc/source/development.rst
```diff
@@ -1,8 +1,130 @@
...
+
+The Trial and error process for the development of your scripts
+---------------------------------------------------------------
+
+Basically, this is the happening during the step when you try to run the upgrade
+described in :doc:`migration_details`:
+
+- [A] get the copy of your database in old version as `db-upgrade`, it is
+  easy to do through the database manager of your old odoo
+  `http://localhost:8012/web/database/manager`
```
MiquelRForgeFlow:
```suggestion
  `http://localhost:8069/web/database/manager`
```
odoo uses that port as default

doc/source/development.rst
```diff
@@ -1,8 +1,130 @@
...
+- [B] run `odoo -d db-upgrade -u all --stop-after-init`
+- [C] In case of error, fix the error (adding or editing migration
+  scripts), then rerun **without** the `--upgrade all`:
+  `odoo -d db-upgrade --stop-after-init`. This will continue the upgrade
+  process from where it left off, starting with the module that caused the
+  error. That way, the migration will not run the end-migration scripts of
+  the modules that were already upgraded before the error, so in case of a
+  live migration you always need to restore the database. But running this way
+  will help you save a lot of time in the upgrade process.
```
MiquelRForgeFlow:
```suggestion
- [C] In case of error, fix the error (adding or editing migration
  scripts), then rerun **without** the `--upgrade all`:
  `odoo -d db-upgrade --stop-after-init`. This will continue the upgrade
  process from where it left off, starting with the module that caused the
  error. That way, the migration will not run the end-migration scripts of
  the modules that were already upgraded before the error. But in case of a
  live migration you always need to restore the database. This way of running
  is only done for testing purposes and will help you save a lot of time
  in the upgrade process.
```

doc/source/development.rst
```diff
@@ -1,8 +1,130 @@
...
+
+Note that for the time being [C] is not available in all versions
+(see explanations here: https://github.com/OCA/OpenUpgrade/pull/2499).
+It is available for 12 and 13:
+https://github.com/OCA/OpenUpgrade/pulls?q=is%3Apr+%5BFIX%5D+reset+exception
+
+Whenever facing unexpected errors, you might want to restart from [A] as
+the step [C] might leave your database in an inconsistent state.
```
MiquelRForgeFlow:
```suggestion
the step [C] will leave your database in an inconsistent state.
```

doc/source/development.rst
```diff
@@ -1,8 +1,130 @@
...
+  `odoo -d db-upgrade --stop-after-init`. This will continue the upgrade
+  process from where it left off, starting with the module that caused the
+  error. That way, the migration will not run the end-migration scripts of
+  the modules that were already upgraded before the error, so in case of a
+  live migration you always need to restore the database. But running this way
+  will help you save a lot of time in the upgrade process.
+
+Note that for the time being [C] is not available in all versions
+(see explanations here: https://github.com/OCA/OpenUpgrade/pull/2499).
+It is available for 12 and 13:
```
MiquelRForgeFlow:
```suggestion
It is available only for 12 and 13:
```

doc/source/development.rst
```diff
@@ -1,8 +1,130 @@
...
+https://github.com/OCA/OpenUpgrade/pulls?q=is%3Apr+%5BFIX%5D+reset+exception
+
+Whenever facing unexpected errors, you might want to restart from [A] as
+the step [C] might leave your database in an inconsistent state.
+
+Restart the upgrade of the failed module instead of upgrading all systematically
+................................................................................
+
+**As an alternative to the step [C] mentioned above...**
```
MiquelRForgeFlow, line 99: I think the other method should be the alternative method, and this one should be the usual method, because the above method only applies to v12 and v13, and this are the docs for v14 , and v14 doesn't have enabled that way at this moment. Could you change the order?

doc/source/development.rst
```diff
@@ -1,8 +1,130 @@
...
+
+Try `--update=all` first and then when it fails on a given module
```
MiquelRForgeFlow:
```suggestion
Try **with** `--update=all` first and then when it fails on a given module
```

doc/source/development.rst
jcdrubay: ok, I will do this change of order.

jcdrubay: it's done @MiquelRForgeFlow

MiquelRForgeFlow (approved):
LGTM

Please squash

pedrobaeza (approved):

marcelsavegnago (approved):

jcdrubay:
@maintainers: Is it possible to merge this PR? thanks

jcdrubay:
@maintainers: Is it possible to merge this PR? thanks :)

PR #2629, opened 2021-03-25
[14.0][UPD] openupgrade_scripts: new update to include improvements

MiquelRForgeFlow:
This PR includes:
- Minor fixes.
- New analysis, making use of https://github.com/OCA/server-tools/pull/2053.

--


openupgrade_framework/readme/CONFIGURE.rst
```diff
@@ -1,4 +1,4 @@
-* call your odoo instance with the option ``--load=web,openupgrade_framework``
+* call your odoo instance with the option ``--load=base,web,openupgrade_framework``
```
legalsylvain, line 1: Why adding base ?

legalsylvain (commented):
One minor question. LGTM otherwise, thanks for the update.

openupgrade_framework/readme/CONFIGURE.rst
MiquelRForgeFlow: If we don't include it, /jsonrpc route doesn't work. Then, If I use for example odoorpc to do the analysis, it will be impossible. Remember that the default of --load is `base,web`, so in any case, you should respect that and add modules, but not removing.

rvalyi (approved):
LGTM

PR #2631, opened 2021-03-26
[14.0][UPD] openupgrade_scripts: update noupdate_changes data

MiquelRForgeFlow:
Up to today.

Requires:

- [x] https://github.com/OCA/server-tools/pull/2071
- [x] https://github.com/OCA/OpenUpgrade/pull/2481

--

### DIFF
--- a/openupgrade_scripts/scripts/auth_oauth/14.0.1.0/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/auth_oauth/14.0.1.0/noupdate_changes.xml
@@ -1,0 +1,6 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+  <record id="provider_google" model="auth.oauth.provider">
+    <field name="css_class">fa fa-fw fa-google</field>
+  </record>
+</odoo>
--- a/openupgrade_scripts/scripts/base/14.0.1.3/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/noupdate_changes.xml
@@ -1,5 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
+  <record id="PEN" model="res.currency">
+    <field name="currency_subunit_label">Centimos</field>
+    <field name="currency_unit_label">Soles</field>
+  </record>
   <record id="ao" model="res.country">
     <field name="zip_required">0</field>
   </record>
--- a/openupgrade_scripts/scripts/event_sale/14.0.1.2/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/event_sale/14.0.1.2/noupdate_changes.xml
@@ -1,6 +1,7 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="product_product_event" model="product.product">
+    <field name="default_code"/>
     <field name="description_sale" eval="False"/>
   </record>
 </odoo>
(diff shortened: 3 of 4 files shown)


pedrobaeza:
Why these great changes? They were totally incorrect previously?

MiquelRForgeFlow:
That seems to be. Maybe when the older noupdate data analysis was done, there were some PRs pending to be merged in v14. Also now the merged modules are taken into account. Also, I am suspicious the previous noupdate data analysis it was done using the old script, because the upgrade_analysis module has an improvement that excludes to print the fields that don't exist anymore (because that was code that always ended being commented).

If you are suspicious of some changes, you can check manually if they are true.

pedrobaeza:
I will check some random ones, as I'm surprised that some mail templates are not changed, but my question was for knowing your general overall.

openupgrade_scripts/scripts/account/14.0.1.1/noupdate_changes.xml
```diff
@@ -1,115 +1,18 @@
...
-  <record id="account_comp_rule" model="ir.rule">
-    <field name="global"/>
+  <record id="base.group_user" model="res.groups">
+    <field name="implied_ids" eval="[(4, ref('account.group_show_line_subtotals_tax_excluded'))]"/>
   </record>
-  <record id="account_fiscal_position_comp_rule" model="ir.rule">
-    <field name="global"/>
-  </record>
-  <record id="account_invoice_line_rule_portal" model="ir.rule">
-    <field name="domain_force">[('move_id.move_type', 'in', ('out_invoice', 'out_refund', 'in_invoice', 'in_refund')), ('move_id.message_partner_ids','child_of',[user.commercial_partner_id.id])]</field>
```
StefanRijnhart: For sure it is a mistake that sections like these are removed. This rule is updated with the renaming of field `type` to `move_type`. Compare 13 https://github.com/odoo/odoo/blob/13.0/addons/account/security/account_security.xml#L209 with 14 https://github.com/odoo/odoo/blob/14.0/addons/account/security/account_security.xml#L214

MiquelRForgeFlow: Oh, then you did it manually, right?

StefanRijnhart: No, I did not edit the files after generating them last time.

MiquelRForgeFlow: It's very strange :S

MiquelRForgeFlow: Sure this means there is a glitch somewhere.

StefanRijnhart: Possibly related: https://github.com/OCA/OpenUpgrade/pull/2481#discussion_r616535367

MiquelRForgeFlow: Fuck, that's not the problem! haha I did a new run with `files` instead of `paths` (adding the UnicodeDecodeError exception) and I still get this wrong noupdate changes. I will soon do a more detailed debug and check the root of the bug.

StefanRijnhart: Fascinating! Looking forward for this mystery to unravel.

MiquelRForgeFlow: Solved the mystery!! :dancers:

StefanRijnhart (approved):

StefanRijnhart:
Won't you tell us what it was?

MiquelRForgeFlow:
Yes, let me do the fix in server-tools.

MiquelRForgeFlow:
Explanation in https://github.com/OCA/server-tools/pull/2071.

PR #2632, opened 2021-03-26
[13.0][UPD] openupgrade_records: retrocompatibility with upgrade_analysis (v14)

MiquelRForgeFlow:
If you try to run an analysis with upgrade_analysis comparing v13 and v14, happens that the noupdata computations break because they don't exist in v13.

--

### DIFF
--- a/odoo/addons/openupgrade_records/models/openupgrade_record.py
+++ b/odoo/addons/openupgrade_records/models/openupgrade_record.py
@@ -3,7 +3,13 @@
 # Copyright 2016 Opener B.V. <https://opener.am>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import ast
+import os
+
 from odoo import api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.modules.module import MANIFEST_NAMES, get_module_path
+from odoo.tools.translate import _
 
 
 class Attribute(models.Model):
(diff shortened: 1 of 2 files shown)


odoo/addons/openupgrade_records/models/openupgrade_record.py
```diff
@@ -119,3 +125,43 @@ def field_dump(self):
...
+        for manifest_name in MANIFEST_NAMES:
+            if os.access(os.path.join(addon_dir, manifest_name), os.R_OK):
+                with open(os.path.join(addon_dir, manifest_name), "r") as f:
+                    manifest_string = f.read()
+                    return ast.literal_eval(manifest_string)
+        raise ValidationError(_("No manifest found in %s" % addon_dir))
+
+    # backward compatibility from upgrade_analysis (v14)
+    @api.model
+    def get_xml_records(self, module):
```
pedrobaeza, line 151: Where do you use this?

MiquelRForgeFlow: It's called from upgrade_analysis module when executing the analysis v13/v14.

pedrobaeza (approved):

pedrobaeza:
Is this finally ready?

MiquelRForgeFlow:
Of course. I hope that the problem in https://github.com/OCA/OpenUpgrade/pull/2631 is not related with this code. I haven't had time yet to check https://github.com/OCA/OpenUpgrade/pull/2631 properly.

pedrobaeza:
Well, I prefer that you check if related before merging.

StefanRijnhart:
This is a duplicate of https://github.com/OCA/OpenUpgrade/pull/2481, except that one returns XML files and this one returns *paths* to XML files. Could that be the reason of the corruption in https://github.com/OCA/OpenUpgrade/pull/2631?

MiquelRForgeFlow:
@StefanRijnhart Sorry, I didn't see you already have a PR doing this. Stupid me. I will close then. This one returns _paths_ because that's what `upgrade_analysis` module is doing now, see https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/models/upgrade_record.py#L170. I don't think that's the cause of the corruption of #2631, but I still have pending to check the cause in detail.

PR #2634, opened 2021-03-28
[12.0][IMP] account: consider account_fiscal_year from OCA/account-financial-tools

MiquelRForgeFlow:
In case of having that module, it's good to consider it for filling the new model.

--

### DIFF
--- a/addons/account/migrations/12.0.1.1/post-migration.py
+++ b/addons/account/migrations/12.0.1.1/post-migration.py
@@ -235,6 +235,27 @@ def populate_fiscal_years(env):
                 create_date, create_uid, write_date, write_uid
             FROM account_fiscalyear;
             """)
+    elif openupgrade.table_exists(env.cr, 'date_range_type') and \
+            openupgrade.column_exists(env.cr, 'date_range_type', 'fiscal_year'):
+        query = """INSERT INTO account_fiscal_year (
+                name, company_id, date_from, date_to,
+                create_date, create_uid, write_date, write_uid
+            )
+            SELECT
+                dr.name, {},
+                dr.date_start, dr.date_end,
+                dr.create_date, dr.create_uid, dr.write_date, dr.write_uid
+            FROM date_range dr
+            JOIN date_range_type drt ON dr.type_id = drt.id"""
+        openupgrade.logged_query(
+            env.cr, query.format("COALESCE(dr.company_id, drt.company_id)") + """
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Not using it, but seems correct

pedrobaeza:
I have retargeted the PR but it seems there are conflicts.

MiquelRForgeFlow:
Solved (it was due to the branch retargeting)

addons/account/migrations/12.0.1.1/post-migration.py
```diff
@@ -221,7 +221,28 @@ def set_default_taxes(env):
 
 
 def populate_fiscal_years(env):
-    if openupgrade.table_exists(env.cr, 'account_fiscalyear'):
```
legalsylvain: Better to keep this test, and if is false test if date_rangetype.

Otherwise. Lgtm. Thanks.

legalsylvain (changes requested):

addons/account/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: Done

addons/account/migrations/12.0.1.1/post-migration.py
```diff
@@ -235,6 +235,37 @@ def populate_fiscal_years(env):
...
+            )
+            SELECT
+                dr.name, COALESCE(dr.company_id, drt.company_id),
+                dr.date_start, dr.date_end,
+                dr.create_date, dr.create_uid, dr.write_date, dr.write_uid
+            FROM date_range dr
+            JOIN date_range_type drt ON dr.type_id = drt.id
+            WHERE drt.fiscal_year AND COALESCE(dr.company_id, drt.company_id) IS NOT NULL
+            """)
+        for company in env["res.company"].search([]):
```
pedrobaeza: Don't do loop but do JOIN with res_company in the SQL query, and please reuse the common query part putting it in a string variable for more readable code.

legalsylvain (approved):

addons/account/migrations/12.0.1.1/post-migration.py
MiquelRForgeFlow: How to do the join of res_company?

pedrobaeza: With a CROSS JOIN

MiquelRForgeFlow: Done :)

PR #2635, opened 2021-03-29
Revert "[13.0][OU-FIX] account: avoid two or more move lines pointing to same invoice line"

pedrobaeza:
Reverts OCA/OpenUpgrade#2614

It seems this is really affecting the performance of these operations. I'm going to do a test with this reverted, and if so, I'm going to proceed, and you will need to find another method or keep it custom, @MiquelRForgeFlow 

@Tecnativa TT26343


pedrobaeza:
Now from more than 30 hours without result to 13 minutes, so this is definitively affecting badly the performance. I merge, and let's see if another way can be found for the other problem, very rare anyway.

PR #2638, opened 2021-04-03
[DOC] Add page dangerous tips

jcdrubay:
Based on some discussions from this PR

https://github.com/OCA/OpenUpgrade/pull/2606

--

### DIFF
--- a/doc/source/dangerous_tips.rst
+++ b/doc/source/dangerous_tips.rst
@@ -1,0 +1,72 @@
+Dangerous Tips
+++++++++++++++
+
+Only for experts in need for desperate ideas to increase the development process.
+
+Disclaimer / Warning
+--------------------
+
+If you see this page, it must be after multiple considerations whether it should or
+should not be published within OpenUpgrade official documentation.
+
+Why not publish it?
+
+Because **the content of this page might lead you to big unrecoverable mess in your
+data.**
+
+Why still publish it?
+
+Because you might be working on a very heavy migration, with a very large size database,
+and if you know what you are doing, this page will help you saving time during the
+development process of migration scripts.
+
+
+Dangerous tip: uninstall modules which are not yet covered
+----------------------------------------------------------
+
+For some of the uncovered modules, you might want to uninstall
+them from your database and re-install them later on after the migration.
+
+This might reduce the workload of the migration and can be applicable
+for "ux" modules, like ``web_widget_image_download``.
+
+But, if you uninstall the wrong modules, **you might end up in losing data.**
+
+To uninstall module, you can execute:
+
+::
+
+   UPDATE ir_module_module
+   SET state = 'to remove'
+   WHERE
+       state IN ('to upgrade', 'to install')
+       AND name in ('module_to_uninstall_1', 'module_to_uninstall_2')
+   RETURNING name
+
+Then run Odoo with ``-u dummy --stop-after-init`` to uninstall the modules
+that could not be upgraded yet.
+
+That way, you get a partially migrated database that you can login to and
+perform UI testing.
+
+
+Dangerous tip: ``fsync = off`` in your postgresql.conf
+------------------------------------------------------
+
+Use the Postgres configuration ``fsync=off`` for OpenUpgrade boost your migration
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Hi @jcdrubay. 

thanks a lot for your many PRs regarding documentation.

This looks great to me.

Don't you have to add a reference to this new page, in the index.rst (or in other place) ?

thanks.

jcdrubay:
@legalsylvain yes, thanks for pointing this out.
Can I wait for the PR related to development documentation to be merged and then add the reference to this page [here](https://github.com/OCA/OpenUpgrade/pull/2624/files#diff-90955bcdeabf45a682bf96a4088456afdb134cdcfa75be5979b247210c43b2e0R121)?

jcdrubay:
The reference to this page has been added. Do you think it's ready for merge?

doc/source/dangerous_tips.rst
```diff
@@ -0,0 +1,72 @@
...
+   RETURNING name
+
+Then run Odoo with ``-u dummy --stop-after-init`` to uninstall the modules
+that could not be upgraded yet.
+
+That way, you get a partially migrated database that you can login to and
+perform UI testing.
+
+
+Dangerous tip: ``fsync = off`` in your postgresql.conf
```
StefanRijnhart, line 53: Still too dangerous for my taste, sorry. Don't mention this at all, please.

pedrobaeza: Yes, I agree with Stefan.

legalsylvain:
Closing, 
- because those dangerous tips don't seem to be consensual 
- and because this PR is not against the new dedicated ``documentation`` branch.
Feel free to create a new PR.

note, regarding the first tips, I think that a more general advice could be to uninstall useless + obsolete + not migrated modules _before_ running an openupgrade. don't you think ?

jcdrubay:
Hi @legalsylvain  I am fine with the closing of the PR. 

A more general advice page could be interesting, but there is much more to write inside.

I have done some documentation internal to Komit "Project Management for Major Version Upgrades". I could try to make this page more generic. However, I am pretty sure that it would be impossible to have everyone to agree on one process. 

So, if I do this page, what do you think if we do it as an "example", rather than a guideline?

Here is what I have in mind to publish, and then I have some details for some of the lines in the table
image
image

legalsylvain:
Hi @jcdrubay. 
Thanks for sharing internal documentation. Do you have other text / graphics related to openupgrade ?
For me that information are missing in the current documentation. It is not technical, but very interesting.

@bguillot, @etobella what is your point of view here ?

Note : @jcdrubay i'm currently working on openupgrade documentation refactoring, so if you are OK, i can adapt the content you wrote and make a PR. (With your crédit of course)

jcdrubay:
Hi @legalsylvain ,

I am glad that you find these interesting. Feel free to modify and re-use this content.

I did not find more graphs, but the following table could be interesting as does an introduction to the different tools when upgrading. We have been working with each of these and for each project we consider all options. 

I am not sure if you want to mention this or not, especially that "Native Upgrade" isn't reliable or systematic (but we saved tremendous time and successfully did early major upgrades with this method since version 13). 


image

Finally, there are 2 more option is to:

- not do data migration: in some projects, it makes sense and then we just get out of all the worries of data upgrade.
- do basic data upgrade with export / import through Excel files with native Odoo => could reduce the work significantly if accepting to migrate only a few data

 If we don't write these 2 options in the doc of OpenUpgrade, is there a better place to document this option in OCA documentation?

In the past, I have also been trying to work on OCA CM which had the "crazy" ambition to revolutionize the way OCA contributions are made. I believe there is a great potential, but I don't have the energy to lead the change required for this. Anyway, this started when I witnessed the complexity to collect all info while working on OpenUpgrade so I believe that it's worth being mentionned:
https://oca-cm-15.komit.link/

I would be happy talking about it or anything else and to share any code (privately at this stage, I already did so in the past). If an interest for OCA is confirmed I could also rewrite some of the code so it would not be "Komit code style" like but more following "OCA standards".

PR #2640, opened 2021-04-03
[13.0][IMP] account: Add progressbar to _move_model_in_data 

pedrobaeza:
Requires:

- [ ] https://github.com/OCA/openupgradelib/pull/233

@Tecnativa TT19674

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -572,7 +572,8 @@ def migration_voucher_moves(env):
     )
 
 
-def _move_model_in_data(env, ids_map, old_model, new_model):
+@openupgrade.progress()
+def _move_model_in_data(env, id_map, old_model, new_model):
     renames = [
         ('mail_message', 'model', 'res_id'),
         ('mail_followers', 'res_model', 'res_id'),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As the progress-bar thing can take some time to be discussed, we can extract that commit to a separate PR if you consider so for speeding up the rest.

kos94ok-3D:
login: andrew
password: 1111
[mig_account_v12.zip](https://github.com/OCA/OpenUpgrade/files/6258865/mig_account_v12.zip)

I try to migrate DB that above.
I opened invoice customer and vendor bill and see nothing

pedrobaeza:
@kos94ok-3D that has nothing to do with this patch. I have migrated already 3 DBs with this patch (and without) and invoices are shown.

kos94ok-3D:
Could you try my DB and then check? There only 4 or 5 invoices

kos94ok-3D:
I found that post-migration script did not start. I do not understand why.
<pre>ERROR mig13 odoo.modules.migration: module account: Each post-migration file must have a "migrate(cr, installed_version)" function</pre>

pedrobaeza:
Indeed this is the log from Travis as well:

https://github.com/OCA/OpenUpgrade/pull/2640/checks?check_run_id=2259629368#step:12:4729

but this is a log from today of a migrated DB with the same PR + openupgradelib modification:
```
. . . 
2021-04-05 12:44:38,569 1 [1;32m[1;49mINFO[0m prod odoo.modules.loading: loading account/report/account_hash_integrity_templates.xml 
2021-04-05 12:44:38,579 1 [1;32m[1;49mINFO[0m prod odoo.modules.migration: module account: Running migration [13.0.1.1>] post-migration 
2021-04-05 12:44:38,584 1 [1;32m[1;49mINFO[0m prod OpenUpgrade: account: post-migration script called with version 12.0.1.1 
. . .
```

pedrobaeza:
The problem is due to the new decorator doesn't exist yet in the library, then the file fails to load, and gives such error. It's normal in this PR, but in your installation, you must install the library specific from the other PR.

kos94ok-3D:
I tried to install openupgradelib from your branch and tried to install the latest from openpgrade with adding your change but it does not work.
Maybe does it not work with python 3.8?

pedrobaeza:
Yes, probably, but that's out of the scope of this. Use the recommended version, which is Python 3.6.

MiquelRForgeFlow:
I would like to merge this without the first commit, and then, when the openupgradelib PR is merged, assuring it works for several python versions, then you can do another PR to add that decorator.

pedrobaeza:
@MiquelRForgeFlow extracted the performance round in #2643

pedrobaeza:
Left this one only for the progressbar.

PR #2644, opened 2021-04-06
[12.0][FIX] account: Don't disable payment term multi-company rule

pedrobaeza:
This rule is introduced one version after introducing the `company_id` field in `account.payment.term`. At this level, we can have payment terms created for one company, but used in documents of other company, so it doesn't serve to disable the rule for avoiding the mess, and more as in v13, the company consistency is checked per document.

Let's keep the rule enabled and get the yell before going to v13 then.

@Tecnativa TT29056 TT31655


MiquelRForgeFlow:
```yml
psycopg2.IntegrityError: null value in column "company_id" violates not-null constraint
```

pedrobaeza:
Reword the commit/PR message and simply not disable rule, as there's not easy solution to this problem and must be solved manually.

MiquelRForgeFlow (approved):

PR #2645, opened 2021-04-07
[11.0][FIX] stock: set use_propagation_minimum_delta to TRUE

MiquelRForgeFlow:
We set this new parameter to true to maintain the same behavior of v10 (as it should be expected).

--

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -439,6 +439,13 @@ def create_stock_move_line_from_inventory_moves(env):
     )
 
 
+def fill_config_parameter_use_propagation_minimum_delta(env):
+    """This method sets to True the new parameter use_propagation_minimum_delta
+    to maintain same behavior of v10"""
+    env["ir.config_parameter"].set_param(
+        'stock.use_propagation_minimum_delta', 'True')
+
+
 def recompute_stock_move_line_qty_different_uom(env):
     """Re-compute product_qty for those lines where product UoM != line UoM."""
     env.cr.execute(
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow (approved):

pedrobaeza:
I don't see where is this used in code.

MiquelRForgeFlow:
This parameter lets set the `propagation_minimum_delta` field. See https://github.com/OCA/OpenUpgrade/blob/11.0/addons/stock/views/res_config_settings_views.xml#L288. The field was visible in v10 because this parameter didn't exist yet.

MiquelRForgeFlow:
We also need this because in the v13 migration, we use this parameter to set the `propagate_date`.

addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -439,6 +439,12 @@ def create_stock_move_line_from_inventory_moves(env):
     )
 
 
+def fill_config_parameter_use_propagation_minimum_delta(env):
+    # new parameter, we set to true to maintain same behavior of v10
```
pedrobaeza: Why don't you put this as docstring?

MiquelRForgeFlow: Why should I? I don't care, but I will do if you care haha

pedrobaeza: Each tool is for each usage. If you are documenting the aim of the method, use docstring. If it's a clarification about specific line/s of code, the use comments.

MiquelRForgeFlow: Ok, done.

PR #2646, opened 2021-04-07
[11.0][FIX] stock: global route should have maximum sequence

MiquelRForgeFlow:
We don't want this route to be the first one being evaluated when searching rules in the common standard logic.

Is 999999999 enough big? In fact, the biggest one should be 2^31-1, but we don't need to put that, right?

--


pedrobaeza (approved):
OK, global routes are a bad idea, but if you have some...

PR #2648, opened 2021-04-08
[13.0][FIX] stock: get correctly boolean value from ir_config_parameter

MiquelRForgeFlow:
If the value is '1' or 't' or 'True', or anything like this, should be True. And '0', or 'f', or 'False' should be False.

--

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -128,7 +128,7 @@ def fill_propagate_date_minimum_delta(env):
         FROM ir_config_parameter icp
         WHERE sm.propagate_date IS NULL
             AND icp.key = 'stock.use_propagation_minimum_delta'
-            AND icp.value = 'True'"""
+            AND icp.value::boolean"""
     )
     # stock rule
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
makes sense

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -146,7 +146,7 @@ def fill_propagate_date_minimum_delta(env):
         FROM ir_config_parameter icp
         WHERE sr.propagate_date IS NULL
             AND icp.key = 'stock.use_propagation_minimum_delta'
-            AND icp.value != 'True'"""
+            AND icp.value::boolean IS DISTINCT FROM TRUE"""
```
pedrobaeza: Or it can be `NULL`?
```suggestion
            AND NOT icp.value::boolean"""
```

MiquelRForgeFlow: Yeah, right. Theoretically, it can be null if it's set with set_param(key, NULL), but that never happens haha

pedrobaeza (approved):

PR #2651, opened 2021-04-12
[13.0][OU-FIX] account: set to zero some null monetary values

MiquelRForgeFlow:
If we let balance be null in some cases, then errors happen. When accessing partners, the account invoice report is used to compute the invoice totals, and when the price_subtotal for a partner is null we get an error (the balance is used to set the price_subtotal).

NOTE: The method _compute_balance doesn't help in this case because for monetary values, odoo compares previous value with new one to do the write. So if previous was 0.0 and the new one is NULL, then the null converts to 0.0, and then as both are 0.0, thus the write is not done. 

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -348,10 +348,9 @@ def migration_invoice_moves(env):
         FROM account_invoice_line ail
             JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state IN ('draft', 'cancel')
             LEFT JOIN res_company rc ON ail.company_id = rc.id
-        LEFT JOIN account_move am ON am.old_invoice_id = ai.id
-        LEFT JOIN account_account aa ON ai.account_id = aa.id
-        WHERE am.id IS NOT NULL AND
-            aa.internal_type in ('receivable', 'payable')""",
+        JOIN account_move am ON am.old_invoice_id = ai.id
+        JOIN account_account aa ON ai.account_id = aa.id
+        WHERE aa.internal_type in ('receivable', 'payable')""",
     )
     # Not Draft or Cancel Invoice Taxes
     openupgrade.logged_query(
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -447,6 +449,13 @@ def migration_invoice_moves(env):
             WHERE debit IS NULL
             """,
     )
+    openupgrade.logged_query(
```
pedrobaeza: Aren't we able to inject this value on the UPDATE/INSERT queries?

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -346,10 +346,9 @@ def migration_invoice_moves(env):
         FROM account_invoice_line ail
             JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state IN ('draft', 'cancel')
             LEFT JOIN res_company rc ON ail.company_id = rc.id
-        LEFT JOIN account_move am ON am.old_invoice_id = ai.id
-        LEFT JOIN account_account aa ON ai.account_id = aa.id
-        WHERE am.id IS NOT NULL AND
-            aa.internal_type in ('receivable', 'payable')""",
+        JOIN account_move am ON am.old_invoice_id = ai.id
```
pedrobaeza, line 351: This is a IMP separate thing

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: well, there are several "insert into" queries. do you want to put that in all of them instead?

pedrobaeza: Yes. Remember my last IMP PR to account, where this kind of queries were spending 13 minutes for 2.2M records, so it's worth to include it in the others.

pedrobaeza (approved):

JordiMForgeFlow (approved):

PR #2653, opened 2021-04-13
[12.0][FIX] mrp: don't set kit products as consumables

MiquelRForgeFlow:
Only version 12.0 forces kits as consumables. In v13, they are not forced, like they were in v11.

--

### DIFF
--- a/addons/mrp/migrations/12.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/12.0.2.0/post-migration.py
@@ -36,36 +36,6 @@ def fill_stock_warehouse_picking_types(env):
             warehouse._create_or_update_sequences_and_picking_types())
 
 
-def force_set_products_consumable(env):
-    """Set products are only allowed to be consumable
-    as per https://github.com/odoo/odoo/commit/3d34d5838.
-    Also delete any quants.
-    """
-    openupgrade.logged_query(
-        env.cr,
-        """
-        DELETE FROM stock_quant WHERE product_id IN (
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Have you reviewed FW-port PRs for that thing? It's probably that the patch is not yet merged, but I think the idea is that they forbid for each supported version this thing.

pedrobaeza:
@chienandalu are you aware of the state of the fw-port?

AaronHForgeFlow:
actually that change was reverted in v12: https://github.com/odoo/odoo/commit/6c4623a46cbe35c18017c59335b3a3c563ae5fbf

Perhaps we should not convert products even if target version is 12.0

pedrobaeza:
Yes, then I think it should be reverted this patch as well. cc @StefanRijnhart

MiquelRForgeFlow:
Revert done.

StefanRijnhart (approved):
Thanks!

MiquelRForgeFlow:
Here I did the update upstream to include the odoo revert: https://github.com/OCA/OpenUpgrade/pull/2655

chienandalu:
> actually that change was reverted in v12: [odoo/odoo@6c4623a](https://github.com/odoo/odoo/commit/6c4623a46cbe35c18017c59335b3a3c563ae5fbf)
> 
> Perhaps we should not convert products even if target version is 12.0

The funny thing is that it was a dirty trick to patch another problem that it was fixed in other way in later versions :upside_down_face:

PR #2654, opened 2021-04-14
[9.0] Merge account_banking_payment_transfer with account_payment_order

rconjour:
Merge account_banking_payment_transfer with account_payment_order like described in: https://github.com/OCA/bank-payment/issues/222

--


pedrobaeza:
This is already done in https://github.com/OCA/bank-payment/blob/72b738ccaf65286150415526dc26f0d63418031c/account_payment_order/migrations/9.0.1.0.0/pre-migration.py#L237, and can't be done unconditionally here.

PR #2657, opened 2021-04-15
[13.0][ADD] website_megamenu: Merged on website

CarlosRoca13:
@Tecnativa TT26952

ping @pedrobaeza @Tardo

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -79,6 +79,7 @@ merged_modules = {
     'website_canonical_url': 'website',
     'website_form_builder': 'website_form',
     'website_logo': 'website',
+    'website_megamenu': 'website',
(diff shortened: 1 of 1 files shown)


rconjour (approved):

pedrobaeza:
Is there anything more to do or the resulting HTML is the same parsed as block on v13?

CarlosRoca13:
@pedrobaeza The only difference is that in v13 the class o_default_snippet_text is added to all texts. But without this they will be working fine

pedrobaeza:
Uhm, not sure it won't have side effects. I think that class is used for auto-selecting text when selecting the block. @Yajo can you confirm?

pedrobaeza:
@Yajo can you please confirm?

Tardo:
As far as I could see in the code, the 'o_default_snippet_text' class is added by the editor.

yajo (approved):
Indeed, that's its use.

PR #2658, opened 2021-04-15
[13.0] Merge base_vat_sanitized with base_vat

rconjour:
[13.0] Merge base_vat_sanitized with base_vat

See description: https://github.com/OCA/partner-contact/tree/13.0/base_vat_sanitized

--

### DIFF
--- a/addons/base_vat/migrations/13.0.1.0/pre-migrate.py
+++ b/addons/base_vat/migrations/13.0.1.0/pre-migrate.py
@@ -1,0 +1,18 @@
+# Copyright 2021 Demolium BV <http://www.demolium.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def merge_base_vat_sanitized_column(env):
+    openupgrade.logged_query(
+        env.cr,
+        "UPDATE res_partner "
+        "SET vat = sanitized_vat "
+        "WHERE vat != sanitized_vat")
+
+
+@openupgrade.migrate()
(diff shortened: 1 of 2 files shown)


pedrobaeza:
In which conditions the sanitized vat is written in the same vat field? I can't reproduce such conditions.

rconjour:
@pedrobaeza it is possible to have databases where the `vat` field is not sanitized, and the vat_sanitized is filled, so it would be appropiate to move the value of `vat_sanitized` to the vat field since we merge `base_vat_sanitized` with `base_vat`

pedrobaeza:
Yeah, but the merge makes sense if Odoo now sanitizes the field. If not, `base_vat_sanitized` is still needed.

rconjour:
Odoo sanitizes the field in `_fix_vat_number()`: https://github.com/odoo/odoo/blob/9ebc6f2bfd9868398bb55bb782338beab9afcb2e/addons/base_vat/models/res_partner.py#L466-L474

This function is triggered on the create and write of a partner:
https://github.com/odoo/odoo/blob/9ebc6f2bfd9868398bb55bb782338beab9afcb2e/addons/base_vat/models/res_partner.py#L480
https://github.com/odoo/odoo/blob/9ebc6f2bfd9868398bb55bb782338beab9afcb2e/addons/base_vat/models/res_partner.py#L486

pedrobaeza:
Yeah, you are right. Thanks for the insights. The problem is that there's an issue with such sanitizing:

https://github.com/odoo/odoo/issues/69333

but the merge (and deprecation of the other module) is correct.

Issue #2661, opened 2021-04-16
[RFC] OCA modules migration scripts host


pedrobaeza:
Due to the new design of OpenUpgrade in v14, we can host all the migration scripts here (Odoo core modules and OCA modules) in this repository, as they will be used the same as for the core modules. Let's balance the pros and cons:

*This is summarizing whole thread*

**Pros of putting everything together in OCA/OpenUpgrade repo**

- [1] Everything together for better discoverability.
- [2] Avoid noise on the business logic repository.
- [3] We can test such migration scripts with a centralized strategy and CI (something that we don't have currently and it's impossible to be done if not centralized).
- [4] On the same manner, we don't hammer other repos CIs testing the business logic for not testing at all a PR adding migration scripts.

**Cons of putting everything together in OCA/OpenUpgrade repo:**

- [1] The scripts can't be used by other upgrade mechanisms, like the enterprise one. I think this is not totally true (but I can't confirm for sure), as enterprise process will upgrade core modules, and then you can perform another update, and you are able to indicate this repo as migration scripts source, and only not yet updated modules will be run.
 
  I also don't think this is necessarily a cons, as some migration scripts are very coupled to the design decisions done on OpenUpgrade, so they will miserably fail when used with other migration process. For example, on v13, we put on `account.move` the field `old_invoice_id` for pointing to the legacy `account.invoice` record that links to the move. The rest of the modules that added fields on `account.invoice` and now are moved to `account.move`, contains migration scripts based on this assumption. You then decide if using this as migration path or not for your migration process.

- [2] This repository will get more dirty. True, but we can organize folders this way:

   \openupgrade_framework
   \openupgrade_scripts
   \oca_scripts

   and host in the last one the migration for OCA modules, and thus we split conveniently the number of contained folders.

- [3] When migrating a module to v14, you may want to add migration scripts at the same time, and with this structure, you are forced to do 2 PRs, one in the module repo, and one here. This one can't be avoided, but for me it's not a great trade-off.

- [4] Pointed by @JordiBForgeFlow: The closer the migration code is to where the actual code is, the better. Answer by @pedrobaeza: We have such isolation though on Odoo core/OpenUpgrade main migration scripts now on v14.

[...]

PR #2662, opened 2021-04-16
14.0 imp documentation

legalsylvain:
### First commit : Fix missing fetch

- trivial patch to fetch the file https://github.com/OCA/OpenUpgrade/raw/13.0/odoo/openupgrade/doc/source/modules120-130.rst that was missing.

### Second commit : Improve coverage visibility

- consider that the coverage of the current version (``modules130-140.rst``) is now in the ``ROADMAP.rst`` file of the ``openupgrade_scripts`` module. With that change, the coverage will be present on app stores, on pypi, and in the readme of the ``openupgrade_scripts`` on github. I think that it is an important information that should be very visible (a lot of issues are opened because users doesn't know the coverage of the current version)
- add another line in the ``fetch_coverage.sh`` file to get the file ``modules130-140.rst`` up to date.
- add a ``DEVELOP.rst`` section to explicit what to write in the ``ROADMAP.rst`` file. (possible values)

Hope you'll see that changes as relevant.

kind regards.

CC : @pedrobaeza, @StefanRijnhart

--


jcdrubay (approved):

openupgrade_scripts/readme/ROADMAP.rst
```diff
@@ -0,0 +1,778 @@
...
++--------------------------------------------+-------------------------------------------------+
+| |del| website_theme_install                | Merged into website                             |
++--------------------------------------------+-------------------------------------------------+
+|website_twitter                             |                                                 |
++--------------------------------------------+-------------------------------------------------+
+
+OCA modules
++++++++++++
+
+Here you will find the coverage of OpenUpgrade for other OCA modules that has
```
pedrobaeza, line 777: This section is not maintained at all, so better to remove it, and link with current discussion at #2661

openupgrade_scripts/readme/ROADMAP.rst
```diff
@@ -0,0 +1,778 @@
+Module coverage 13.0 -> 14.0
```
StefanRijnhart, line 1: There is already https://github.com/OCA/OpenUpgrade/blob/14.0/doc/source/modules130-140.rst. I don't find moving it here very intuitively. In the roadmap, we can link to this doc instead.

legalsylvain: > There is already https://github.com/OCA/OpenUpgrade/blob/14.0/doc/source/modules130-140.rst. 

the idea is to have the master file in the readme folder of the script module. and then to have it also up to date in the documentation, because of the wget [here](https://github.com/OCA/OpenUpgrade/pull/2662/files#diff-491a891e27f244dd6c91aa9da9d99a0422b08ecfab3f25a09765a2b9180910d2R10) **as all the other coverage files** (of the other revision)

So, finally
- ``/doc/`` is for all the general documentation, which should not move much
- ``openupgrade_scripts`` is for all the scripts, and i consider this coverage file as the roadmap of this module, so when we propose a script, we change only files of a single folder.

> In the roadmap, we can link to this doc instead.

Indeed, if you don't like this move, we can just consider to add a link to the coverage section. It's better than nothing.

legalsylvain:
Hi all, do you know when (and where) 

- `` doc/fetch_coverage.sh`` is executed ? I don't understand very well how all the coverage files are updated, specially files of other revisions

- how the website is configured ? the current coverage page https://doc.therp.nl/openupgrade/status.html seems to point on very obsolete page
- not on the [V14 page](https://github.com/OCA/OpenUpgrade/blob/14.0/doc/source/status.rst)
- not on the [V13 page](https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/openupgrade/doc/source/status.rst)
- but on a altered version of [V12 page](https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/openupgrade/doc/source/status.rst)

StefanRijnhart:
Those two issues are related. Hosting the documentation there is a legacy issue from the time that OpenUpgrade itself was hosted on the Therp GitHub account. Now that @hbrunn has left Therp as well, it is not being updated anymore and we should host it under the OCA umbrella. When we set that up, we can configure a build job to fetch the coverage files and push the documentation.

openupgrade_scripts/readme/ROADMAP.rst
legalsylvain: Hi @pedrobaeza. 
- I just copied the file, not changed the content.
- AFAIK, this part is for Odoo features that moved in OCA modules (or the inverse). See, for the V12 coverage file : https://doc.therp.nl/openupgrade/modules110-120.html#oca-modules

pedrobaeza: Yes, that's why I'm saying to remove it from both parts.

pedrobaeza:
Any news on this, Sylvain?

openupgrade_scripts/readme/ROADMAP.rst
MiquelRForgeFlow: I vote for the link option. It's not good to maintain duplicated .rst files.

MiquelRForgeFlow:
Sylvain, note that we have added a functional documentation github workflow in https://github.com/OCA/OpenUpgrade/pull/2744.

legalsylvain:
Well we should before fix what Stefan said. The website is not updated since years, that is a pity. Don't know exactly how to move forward, though. New website named doc-openupgrade.odoo-community.org ?

StefanRijnhart:
Maybe we should commit the HTML in a Github job to the &lt;latest&gt; branch and host it with Github pages. It will then be available at https://OCA.github.io/OpenUpgrade if I understand correctly.

pedrobaeza:
I vote for moving to GitHub pages. Any of you have experience doing that?

StefanRijnhart:
No, but I can try.

pedrobaeza:
That would be awesome

StefanRijnhart:
See https://github.com/OCA/OpenUpgrade/pull/2765. ~~But I need to look into the gitbot thing.~~

tmotyl:
Hi
I see that docs are now rendered fine using github pages. 

What are the next steps with this patch? I wanted to push some fixes to docs I found when reading, but first maybe its better to merge outstanding patches related to documentation so I dont have to fix sth which is already fixed.

StefanRijnhart:
I believe the main objective of this PR was to increase the exposure of the module coverage. Given the discussion above about how to do so, and the subsequent implementation of live documentation, I think this PR can be closed.

PR #2663, opened 2021-04-17
[13.0][FIX] stock_account: wrong sign for svl value of out moves

kos94ok-3D:
image

--


kos94ok-3D:
@AaronHForgeFlow Can you check this?

kos94ok-3D:
https://github.com/OCA/OpenUpgrade/pull/2376

AaronHForgeFlow:
@kos94ok-3D I see, I am checking. Is that the Unit Value or the Total Value?

AaronHForgeFlow:
I think you are correct and I am making a PR to fix that. Still checking.

kos94ok-3D:
@AaronHForgeFlow screen from runbot Odoo13
There are we see that price unit price is always positive, but in stock_move we can have negative
image
image

PR #2666, opened 2021-04-19
[14.0][MIG] uom

RaulOForgeFlow:
Migration scripts. 
--


MiquelRForgeFlow (approved):

rvalyi (approved):

pedrobaeza:
Added a correction for some uom roundings, so merging.

openupgrade_scripts/scripts/uom/14.0.1.0/noupdate_changes.xml
```diff
@@ -39,7 +39,7 @@
     <field name="name">qt (US)</field>
   </record>
   <record id="product_uom_unit" model="uom.uom">
-    <field name="rounding"/>
+    <field name="rounding" eval="0.01"/>
```
gfcapalbo, line 42: violates   _check_category_reference_uniqueness 

https://github.com/OCA/OCB/blob/13.0/addons/uom/models/uom_uom.py#L83-L94

File "/home/gcapalbo/k240/migrate_therp/build-14.0/parts/odoo/addons/uom/models/uom_uom.py", line 85, in _check_category_reference_uniqueness
    raise ValidationError(_("UoM category %s should only have one reference unit of measure.") % (self.env['uom.category'].browse(uom_data['category_id']).name,))
odoo.exceptions.ValidationError: UoM categorie Unitheeft slechts één referentie maateenheid.

MiquelRForgeFlow: But this error doesn't have any relation with the rounding. This constrain already existed in v13, and your problem seems more a problem of your db data, which has something wrong and you should solve manually before migrating.

PR #2669, opened 2021-04-20
[13.0][FIX] account: consider mail.alias relation to invoices

MiquelRForgeFlow:
.

--

### DIFF
--- a/addons/account/migrations/13.0.1.1/end-migration.py
+++ b/addons/account/migrations/13.0.1.1/end-migration.py
@@ -44,6 +44,16 @@ def add_default_account_data(env):
             company[company_field] = account.id
 
 
+def update_mail_alias_for_moves(env):
+    # Done here just in case alias customizations exist
+    journals = env["account.journal"].with_context(
+        active_test=False).search([("alias_id", "!=", False)])
+    for journal in journals:
+        journal.alias_id.unlink()
+        journal._update_mail_alias({"alias_name": journal.alias_name})
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     add_default_account_data(env)
+    update_mail_alias_for_moves(env)


pedrobaeza:
I'm afraid this doesn't serve, as the alias also contains some applied default values that will make this fail anyway. We need to call the recreation of the mail.alias by ORM for each journal on post, and remove this bulk renaming.

MiquelRForgeFlow:
How to call the recreation of the mail.alias?

pedrobaeza:
Don't know yet. Check how is done in the code.

kos94ok-3D:
@MiquelRForgeFlow Maybe it will help you. account journal have `_update_mail_alias` function

MiquelRForgeFlow:
Thanks @kos94ok-3D , the `_update_mail_alias` method helped me understand what needed to be updated. Also I tested it in a migrated db of a client :)

pedrobaeza:
This is still not changing the default values in alias.

pedrobaeza:
Just call `_update_mail_alias` in all journals with alias_name.

MiquelRForgeFlow:
In `_update_mail_alias` method, the alias gets updated with:

```
    'alias_defaults': {'type': type == 'purchase' and 'in_invoice' or 'out_invoice', 'company_id': self.company_id.id, 'journal_id': self.id},
    'alias_parent_thread_id': self.id,
    'alias_name': alias_name,
```
As far I can see, none of these values should change or be changed.

pedrobaeza:
Needed code:

```python
journals = env["account.journal"].search([("alias_name", "!=", False)])
for journal in journals:
    journal.alias_id.unlink()
    journal._update_mail_alias({"alias_name": journal.alias_name})
```

pedrobaeza:
In previous alias, `company_id` wasn't present.

pedrobaeza:
And other customization may add more values.

pedrobaeza:
I'm thinking this should be run on `end-migration` due to my last sentence.

MiquelRForgeFlow:
> In previous alias, company_id wasn't present.

False, I checked in v12, and code is the same. That's why my migrated multicompany db has alias names with the company.

> And other customization may add more values.

Ok, that's your unique valid point. Fuck customizations.

addons/account/migrations/13.0.1.1/end-migration.py
```diff
@@ -44,6 +44,15 @@ def add_default_account_data(env):
             company[company_field] = account.id
 
 
+def update_mail_alias_for_moves(env):
+    # Done here just in case alias customizations exist
+    journals = env["account.journal"].search([("alias_id", "!=", False)])
```
MiquelRForgeFlow:
```suggestion
    journals = env["account.journal"].with_context(active_test=False).search([("alias_id", "!=", False)])
```
?

pedrobaeza:
About the company, sorry, I meant `journal_id` value.

addons/account/migrations/13.0.1.1/end-migration.py
pedrobaeza: Yes, I have just checked that disable journals still preserve the alias.

addons/account/migrations/13.0.1.1/end-migration.py
```diff
@@ -44,6 +44,16 @@ def add_default_account_data(env):
             company[company_field] = account.id
 
 
+def update_mail_alias_for_moves(env):
+    # Done here just in case alias customizations exist
+    journals = env["account.journal"].with_context(
+        active_test=False).search([("alias_id", "!=", False)])
```
pedrobaeza, line 50: As it's the important data

```suggestion
        active_test=False).search([("alias_name", "!=", False)])
```

pedrobaeza: Ah, no, forget it

pedrobaeza (approved):

PR #2671, opened 2021-04-22
[13.0][FIX] mass_mailing: fill campaign_id in some models

MiquelRForgeFlow:
The model `mail.mass_mailing.campaign` was merged into `utm.campaign`. There were some models that had a many2one field for each of those models (`mass_mailing_campaign_id` and `campaign_id` correspondingly). But can happen in some cases that `campaign_id` was empty but `mass_mailing_campaign_id` was filled, thus we need to fill `campaign_id`. Also there was a wrong field rename related to this issue.

--

### DIFF
--- a/addons/mass_mailing/migrations/13.0.2.0/post-migration.py
+++ b/addons/mass_mailing/migrations/13.0.2.0/post-migration.py
@@ -8,7 +8,20 @@ _unlink_by_xmlid = [
 ]
 
 
+def fill_several_campaign_id(env):
+    tables = ["mailing_mailing", "mailing_trace", "link_tracker", "link_tracker_click"]
+    for table in tables:
+        openupgrade.logged_query(
+            env.cr, """
+            UPDATE {} AS tt
+            SET campaign_id = mmc.campaign_id
+            FROM mail_mass_mailing_campaign mmc
+            WHERE tt.mass_mailing_campaign_id = mmc.id
+                AND tt.campaign_id IS NULL""".format(table))
+
+
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
+    fill_several_campaign_id(env)
     openupgrade.delete_records_safely_by_xml_id(env, _unlink_by_xmlid)
     openupgrade.load_data(env.cr, 'mass_mailing', 'migrations/13.0.2.0/noupdate_changes.xml')
--- a/addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
+++ b/addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
@@ -41,7 +41,6 @@ _table_renames = [
 
 _field_renames = [
     ('link.tracker.click', 'link_tracker_click', 'mail_stat_id', 'mailing_trace_id'),
-    ('mailing.trace', 'mailing_trace', 'mass_mailing_campaign_id', 'campaign_id'),
     ('mail.mail', 'mail_mail', 'statistics_ids', 'mailing_trace_ids'),
     ('utm.campaign', 'utm_campaign', 'mass_mailing_ids', 'mailing_mail_ids'),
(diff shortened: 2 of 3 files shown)


addons/mass_mailing/migrations/13.0.2.0/post-migration.py
```diff
@@ -8,7 +8,39 @@
 ]
 
 
+def fill_several_campaign_id(env):
+    openupgrade.logged_query(
+        env.cr, """
+    UPDATE mailing_mailing mm
```
pedrobaeza: It seems the same query is repeated 4 times changing the table. This worths a loop iterating on the table name for a clearer code.

addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
```diff
@@ -41,7 +41,6 @@
 
 _field_renames = [
     ('link.tracker.click', 'link_tracker_click', 'mail_stat_id', 'mailing_trace_id'),
-    ('mailing.trace', 'mailing_trace', 'mass_mailing_campaign_id', 'campaign_id'),
```
pedrobaeza, line 44: Why is this incorrect? `campaign_id` doesn't exist before v13, so the renaming is good for me.

pedrobaeza (changes requested):
I also don't get the duality you are mentioning. Please put a practical example. "There are some cases" doesn't serve.

addons/mass_mailing/migrations/13.0.2.0/pre-migration.py
MiquelRForgeFlow: The problem here is you are renaming a field `mass_mailing_campaign_id` (of model `mail.mass_mailing.campaign`) to field campaign_id (of model `utm.campaign`), which obviously that's wrong as the ids may not correspond.

MiquelRForgeFlow:
> I also don't get the duality you are mentioning. Please put a practical example. "There are some cases doesn't serve".

In v12:
- model `mail.mass_mailing` had both fields mass_mailing_campaign_id and campaign_id
- model `link.tracker` had both fields mass_mailing_campaign_id and campaign_id

Also:
- model `mail.mail.statistics` had field mass_mailing_campaign_id
- model `link.tracker.click` had field mass_mailing_campaign_id

Those four mass_mailing_campaign_id fields point to `mail.mass_mailing.campaign` model, a model which has campaign_id field pointing to `utm.campaign` model. Thus, the post-migration script I added, assures the campaign_id field of those 4 models are filled pointing to `utm.campaign` model.

addons/mass_mailing/migrations/13.0.2.0/post-migration.py
MiquelRForgeFlow: done

addons/mass_mailing/migrations/13.0.2.0/post-migration.py
```diff
@@ -8,7 +8,20 @@
 ]
 
 
+def fill_several_campaign_id(env):
+    tables = ["mailing_mailing", "mailing_trace", "link_tracker", "link_tracker_click"]
```
pedrobaeza, line 12: And aren't the rest of the models (except `mailing.trace`) having already filled `campaign_id`?

MiquelRForgeFlow: It may be already filled, or it may not be filled, the cases vary. That's way I am using "`AND tt.campaign_id IS NULL`". I mean, they having old field `mass_mailing_campaign_id` filled doesn't guarantee they had also filled old field `campaign_id`.

pedrobaeza: OK, I buy it :)

pedrobaeza (approved):

PR #2672, opened 2021-04-22
[13.0][FIX] product: still missing attributes and relations

MiquelRForgeFlow:
Explained in code, see each commit. It's a continuation of https://github.com/OCA/OpenUpgrade/pull/2633.

--

### DIFF
--- a/addons/mrp/migrations/13.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/13.0.2.0/post-migration.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Payam Yasaie <https://www.tashilgostar.com>
 # Copyright 2020 Andrii Skrypka <andrijskrypa@ukr.net>
+# Copyright 2020 ForgeFlow <http://www.forgeflow.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
 
@@ -14,11 +15,13 @@ def fill_bom_product_template_attribute_value(env):
         (mrp_bom_line_id, product_template_attribute_value_id)
     SELECT mbl_pav_rel.mrp_bom_line_id, ptav.id
     FROM mrp_bom_line_product_attribute_value_rel mbl_pav_rel
+    JOIN product_attribute_value pav ON pav.id = mbl_pav_rel.product_attribute_value_id
     JOIN mrp_bom_line mbl ON mbl.id = mbl_pav_rel.mrp_bom_line_id
     JOIN mrp_bom mb ON mbl.bom_id = mb.id
     JOIN product_template_attribute_value ptav ON (
             mb.product_tmpl_id = ptav.product_tmpl_id
-            AND mbl_pav_rel.product_attribute_value_id = ptav.product_attribute_value_id)
+            AND pav.id = ptav.product_attribute_value_id
+            AND pav.attribute_id = ptav.attribute_id)
     GROUP BY mbl_pav_rel.mrp_bom_line_id, ptav.id
     """)
 
--- a/addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
+++ b/addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
@@ -5,6 +5,7 @@ obsolete model product.price.history
 ---Fields in module 'product'---
 product      / product.attribute        / product_tmpl_ids (many2many)  : NEW relation: product.template, isfunction: function, stored
 # NOTHING TO DO: is a new compute, so we could precreate and fill to speed up, but not needed because usually there are few records
+# (we assure the field attribute_line_id of ptav is filled, as it is used in the compute)
 
 product      / product.attribute.value  / pav_attribute_line_ids (many2many): NEW relation: product.template.attribute.line
 # NOTHING TO DO: Inverse many2many of an existing m2m table
@@ -76,8 +77,7 @@ product      / product.template.attribute.value / ptav_active (boolean)         
 # DONE: pre-migration: Insert missing records for variants with attributes that currently are not on the template definition
 
 product      / product.template.attribute.value / attribute_id (many2one)       : is now stored
-# DONE: pre-migration: added SQL column for avoiding the ORM computation
-# DONE: post-migration: filled field (done in post-migration because it's a related to a new field)
+# DONE: pre-migration: filled field (done in pre to let product_attribute_product_template_rel be calculated and filled in load).
 
 product      / product.template.attribute.line / product_template_value_ids (many2many): is now stored
 product      / product.template.attribute.line / product_template_value_ids (many2many): not a function anymore
--- a/addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
+++ b/addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
@@ -83,7 +83,7 @@ product      / product.template.attribute.line / product_template_value_ids (man
 product      / product.template.attribute.line / product_template_value_ids (many2many): not a function anymore
 product      / product.template.attribute.line / product_template_value_ids (many2many): type is now 'one2many' ('many2many')
 product      / product.template.attribute.value / attribute_line_id (many2one)  : NEW relation: product.template.attribute.line, required
-# DONE: post-migration: filled field (using the same approach than in OU v12)
+# DONE: pre-migration: filled required field (using the same approach than in OU v12)
 
 ---XML records in module 'product'---
 NEW decimal.precision: product.decimal_volume (noupdate)
--- a/addons/product/migrations/13.0.1.2/pre-migration.py
+++ b/addons/product/migrations/13.0.1.2/pre-migration.py
@@ -2,6 +2,7 @@
 # Copyright 2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+from odoo.tools import sql
 
 _xmlid_renames = [
     ('sale.group_discount_per_so_line', 'product.group_discount_per_so_line'),
@@ -174,6 +175,10 @@ def migrate(env, version):
     openupgrade.rename_xmlids(cr, _xmlid_renames)
     fill_product_pricelist_item_prices(env)
     fill_product_pricelist_item_active_default(env)
+    sql.create_index(
+        env.cr, 'product_template_attribute_value_ou_migration_idx',
+        "product_template_attribute_value",
+        ['product_attribute_value_id', 'product_tmpl_id'])
     insert_missing_product_template_attribute_line(env)
     insert_missing_product_template_attribute_value(env)
     calculate_product_product_combination_indices(env)
--- a/addons/product/migrations/13.0.1.2/pre-migration.py
+++ b/addons/product/migrations/13.0.1.2/pre-migration.py
@@ -72,6 +72,23 @@ def insert_missing_product_template_attribute_line(env):
         WHERE ptal.id IS NULL
         GROUP BY pav.attribute_id, pp.product_tmpl_id""",
     )
+    if openupgrade.table_exists(env.cr, "mrp_bom_line"):
+        # make use of mrp_bom_line_product_attribute_value_rel
+        openupgrade.logged_query(
+            env.cr, """
+            INSERT INTO product_template_attribute_line
+                (active, product_tmpl_id, attribute_id)
+            SELECT False, mb.product_tmpl_id, pav.attribute_id
+            FROM mrp_bom_line_product_attribute_value_rel mblpavr
+            JOIN mrp_bom_line mbl ON mbl.id = mblpavr.mrp_bom_line_id
+            JOIN mrp_bom mb ON mbl.bom_id = mb.id
+            JOIN product_attribute_value pav ON pav.id = mblpavr.product_attribute_value_id
+            LEFT JOIN product_template_attribute_line ptal
+                ON ptal.product_tmpl_id = mb.product_tmpl_id
+                    AND ptal.attribute_id = pav.attribute_id
+            WHERE ptal.id IS NULL
+            GROUP BY pav.attribute_id, mb.product_tmpl_id"""
+        )
 
 
 def insert_missing_product_template_attribute_value(env):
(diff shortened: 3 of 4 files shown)


pedrobaeza:
I don't see enough explanations neither on the PR message nor the commit messages. Please do the same as I did in #2633.

MiquelRForgeFlow:
Yeah, yesterday I did it a bit fast.

The second commit is easy to explain. After migration of a client that has mrp, there were some bom lines that had lost their attribute, just the same way/problem you had previously with products that you solved previously. And example for this case should be a copy pasta of the ones you put but changing products by bom lines. And that's why the queries are very similar.

About the first commit it's a bit tricky. It happens that the new archived attributes you created, after migration, they still had null attribute_line_id, which is a required field. The problem lies in that when you created the archived attribute lines, you should also fill the m2m table between PAV and PTAL, and that's what the commit basically does. It also includes two things: 1) in postmigration, first check for active ptals and then for inactive ptals, always trying to find ptals because we want all attribute_line_id filled. 2) in pre-migration, we begin to fill attribute_line_id for the new archived ptavs (using the previous m2m table between PAV and PTAL), and thus in post-migration this ones won't need have attribute_line_id filled.

MiquelRForgeFlow:
> I don't see enough explanations neither on the PR message nor the commit messages

Done. Now you can check commit by commit what each minor change is about.

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -174,7 +174,7 @@ def migrate(env, version):
     openupgrade.rename_xmlids(cr, _xmlid_renames)
     fill_product_pricelist_item_prices(env)
     fill_product_pricelist_item_active_default(env)
+    calculate_product_product_combination_indices(env)
     insert_missing_product_template_attribute_line(env)
     insert_missing_product_template_attribute_value(env)
-    calculate_product_product_combination_indices(env)
```
pedrobaeza, line 179: If the ptvas were archived on v13, they has combination_indices, isn't it? Why not filling such ones then?

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
```diff
@@ -83,7 +83,7 @@ product      / product.template.attribute.line / product_template_value_ids (man
 product      / product.template.attribute.line / product_template_value_ids (many2many): not a function anymore
 product      / product.template.attribute.line / product_template_value_ids (many2many): type is now 'one2many' ('many2many')
 product      / product.template.attribute.value / attribute_line_id (many2one)  : NEW relation: product.template.attribute.line, required
-# DONE: post-migration: filled field (using the same approach than in OU v12)
+# DONE: pre-migration: filled required field (using the same approach than in OU v12)
```
pedrobaeza, line 86: If I don't remember bad, if you this, there are missing values that are fill lately.

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -125,18 +125,32 @@ def insert_missing_product_template_attribute_value(env):
     )])
     openupgrade.logged_query(
         env.cr,
-        """
+        "ALTER TABLE product_template_attribute_value "
+        "ADD COLUMN attribute_line_id INT4",
+    )
+    fill_product_template_attribute_value_attribute_line_id(env)
```
pedrobaeza: Please don't nest this high level methods.

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -88,6 +88,39 @@ def insert_missing_product_template_attribute_line(env):
         AND insert_ptal.attribute_id = pav.attribute_id)
     GROUP BY pav.id, insert_ptal.id""",
     )
+    if openupgrade.table_exists(env.cr, "mrp_bom_line"):
```
pedrobaeza, line 75: I don't get the relation between product_template_attribute_line and mrp boms, neither the code has any reference to this.

pedrobaeza (changes requested):
I still miss an explanation of each change about why, and enumerating the steps to reproduce the problem. Please take as example https://github.com/OCA/OpenUpgrade/pull/2633, where the exact initial data and the expected ones are shown.

Without that, everything is done blindly, and the SQL modifications are very convoluted for validating them reading them.

addons/product/migrations/13.0.1.2/pre-migration.py
MiquelRForgeFlow: Ok, you have a point here. I don't know why but I thought archived ptavs were avoided. Dropped that commit.

addons/product/migrations/13.0.1.2/openupgrade_analysis_work.txt
MiquelRForgeFlow: I initially moved that to pre-migration because the field attribute_line_id of ptav is required. Later, it proves useful in pre-migration, in correctly filling the missing ptavs (the query of missing ptavs use that field in order to make it compatible with the sql_constraint `unique(attribute_line_id, product_attribute_value_id)`)

addons/product/migrations/13.0.1.2/pre-migration.py
MiquelRForgeFlow: Ok, done.

addons/product/migrations/13.0.1.2/pre-migration.py
MiquelRForgeFlow: It's explained in that commit message. In mrp module, the bom lines also have attributes, and some are missing in a similar way the missing attributes on products case you detected. In order to avoid these missing attributes being missing, we need to make use of mrp_bom_line_product_attribute_value_rel, which plays a very similar role that product_attribute_value_product_product_rel has for products. We have tested it and works excellent.

MiquelRForgeFlow:
> I still miss an explanation of each change about why, and enumerating the steps to reproduce the problem

Well, I thought I added the enough explanations in the commit message, in the code or in the work.txt (depending the commit). Just tell me exactly which commits you need more info about and I will try to add more explanation on them.

MiquelRForgeFlow:
PR updated :)

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -134,38 +171,76 @@ def calculate_product_product_combination_indices(env):
...
+        UPDATE product_template_attribute_value ptav
+        SET attribute_line_id = ptal.id
+        FROM product_template_attribute_line ptal
+        JOIN product_template pt ON ptal.product_tmpl_id = pt.id
+        JOIN product_attribute_value pav ON ptal.attribute_id = pav.attribute_id
+        WHERE ptal.active IS DISTINCT FROM TRUE AND
+            ptav.product_tmpl_id = pt.id AND
+            ptav.product_attribute_value_id = pav.id
+            AND ptav.attribute_line_id IS NULL""",
+    )
```
kos94ok-3D: Now we can do it in one query:
<pre>
  openupgrade.logged_query(
      env.cr, """
      UPDATE product_template_attribute_value ptav
      SET attribute_line_id = ptal.id
      FROM product_template_attribute_line ptal
      JOIN product_attribute_value pav ON ptal.attribute_id = pav.attribute_id
      WHERE ptav.product_tmpl_id = ptal.product_tmpl_id
          AND ptav.product_attribute_value_id = pav.id
          AND ptav.attribute_line_id IS NULL""",
  )
</pre>

addons/mrp/migrations/13.0.2.0/post-migration.py
```diff
@@ -14,12 +15,14 @@ def fill_bom_product_template_attribute_value(env):
         (mrp_bom_line_id, product_template_attribute_value_id)
     SELECT mbl_pav_rel.mrp_bom_line_id, ptav.id
     FROM mrp_bom_line_product_attribute_value_rel mbl_pav_rel
+    JOIN product_attribute_value pav ON pav.id = mbl_pav_rel.product_attribute_value_id
     JOIN mrp_bom_line mbl ON mbl.id = mbl_pav_rel.mrp_bom_line_id
     JOIN mrp_bom mb ON mbl.bom_id = mb.id
     JOIN product_template pt ON mb.product_tmpl_id = pt.id
```
kos94ok-3D: useless join we need only `id` which we already have in `mb.product_tmpl_id`

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -63,20 +58,37 @@ def insert_missing_product_template_attribute_line(env):
...
+        # make use of mrp_bom_line_product_attribute_value_rel
+        openupgrade.logged_query(
+            env.cr, """
+            INSERT INTO product_template_attribute_line
+                (active, product_tmpl_id, attribute_id)
+            SELECT False, pt.id, pav.attribute_id
+            FROM mrp_bom_line_product_attribute_value_rel mblpavr
+            JOIN mrp_bom_line mbl ON mbl.id = mblpavr.mrp_bom_line_id
+            JOIN mrp_bom mb ON mbl.bom_id = mb.id
+            JOIN product_template pt ON pt.id = mb.product_tmpl_id
```
kos94ok-3D: useless join we need only `id` which we already have in `mb.product_tmpl_id`

addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -107,20 +119,45 @@ def insert_missing_product_template_attribute_value(env):
...
+        # make use of mrp_bom_line_product_attribute_value_rel
+        openupgrade.logged_query(
+            env.cr, """
+            INSERT INTO product_template_attribute_value
+                (ptav_active, product_attribute_value_id, product_tmpl_id, attribute_line_id)
+            SELECT False, mblpavr.product_attribute_value_id, pt.id, ptal.id
+            FROM mrp_bom_line_product_attribute_value_rel mblpavr
+            JOIN mrp_bom_line mbl ON mbl.id = mblpavr.mrp_bom_line_id
+            JOIN mrp_bom mb ON mbl.bom_id = mb.id
+            JOIN product_template pt ON pt.id = mb.product_tmpl_id
```
kos94ok-3D: useless join we need only `id` which we already have in `mb.product_tmpl_id`

kos94ok-3D (commented):
Work as expected.
I have some cosmetic suggests

addons/mrp/migrations/13.0.2.0/post-migration.py
MiquelRForgeFlow: that's https://github.com/OCA/OpenUpgrade/pull/2957 😆

kos94ok-3D:
@MiquelRForgeFlow You forgot about `group by` section

MiquelRForgeFlow:
@kos94ok-3D done

kos94ok-3D (approved):

kos94ok-3D:
HI, @MiquelRForgeFlow
Are we waiting for the review from @pedrobaeza?

MiquelRForgeFlow:
> Are we waiting for the review from @pedrobaeza?

Yeah, a long awaited review. He wants me to wait for his reviews in this repo for all the PRs.

pedrobaeza:
I still need commit messages explaining the whys of these commits:

97142fbe68e913dc245225828c533887c2326060
c263c885cb9ff813d7336046b4af3f2a7e535ee2
e54a1412c6c2710973a4783eb148cafbb8759e20
1bc8c9820c6dbec4e277cdc00114579264da409e
f6dbddc5400cc22fbf084e743cc7f42c9e5eeac0
a823cdec234b7b6902349f147b6f858f987cafb9

and steps to reproduce a problem.

Please see a good example like the one I did for a similar case: 3dabbccb7d96c9627fae0f1e4fbdf790c1589be9

MiquelRForgeFlow:
- https://github.com/OCA/OpenUpgrade/commit/97142fbe68e913dc245225828c533887c2326060 - it's just a simplification. It reduces the selects by one.
- https://github.com/OCA/OpenUpgrade/commit/e54a1412c6c2710973a4783eb148cafbb8759e20 is the core of this PR. All this refactor began with this commit, which fixes the correct number of archived ptavs that are being created. The key is adding this condition in the left join:
Selection_054
And to add this condition, then we need to have that `ptav.attribune_line_id` filled, which is done previously in https://github.com/OCA/OpenUpgrade/commit/c263c885cb9ff813d7336046b4af3f2a7e535ee2.
- https://github.com/OCA/OpenUpgrade/commit/1bc8c9820c6dbec4e277cdc00114579264da409e already has the explanation included in the docstring (of `fill_product_template_attribute_value__attribute_id_related` method).
- https://github.com/OCA/OpenUpgrade/commit/f6dbddc5400cc22fbf084e743cc7f42c9e5eeac0 just solves a similar case of losing data (and thus not having enough archived ptavs). The mrp_bom_line_product_attribute_value_rel table becomes deprecated and is replaced by mrp_bom_line_product_template_attribute_value_rel in mrp module. This extra archived ptavs are created here instead of doing it in mrp module because we can take advantage of the other queries to fill the collateral data. If not, in mrp we would have to copy and paste the same queries, like the one that fills attribute_id in ptavs, and thus duplicating a lot of code.
- https://github.com/OCA/OpenUpgrade/commit/a823cdec234b7b6902349f147b6f858f987cafb9 this one is a pretty obvious fix and I doubt it needs any explanation. Again, without this extra conditions, we are having wrong data.

kos94ok-3D:
Case1 Steps to reproduce:
1. Create a Product Template with
   Attribute lines: `Color (Red, Blue)`, `Material (Iron, Wood)`
 2. Create and Validate transfer with Variant `(Red, Iron)`
 3. Remove `Red` value from attribute line
  (on this step we lose here link between `product.attribute.value` and `product.template.attribute.value`)
 4. run migration
 5. Go to Product Template form
 6. Press on `Configure Variants` button 
 7. Set `Inactive` filter
 
Actual Result:
Missing attribute for inactive `product.template.attribute.value`

Case 2 Steps to reproduce:
1. Create a Product Template with
   Attribute lines: `Color (Red, Blue)`, `Material (Iron, Wood)`
 2. Create BOM for this Product in the BOM line set apply on `Color: Red`
 3. Remove `Red` value from attribute line
  (on this step `product.template.attribute.value` and `product.variant` will be removed)
 4. run migration
 5. Go to Product Template form
 6. Press on `Configure Variants` button 
 7. Set `Inactive` filter not found 
Actual Result:
Missing `Color: Red` record

8. Go to BOM form
Actual Result:
Missing `Color: Red` in BOM line

pedrobaeza (approved):
Pushed a rebased branch with all of these comments in commit messages and review thoroughly. There are few doubts about some extra linking, but let's see applied to real migrations.

PR #2675, opened 2021-04-23
[10.0][OU-IMP] Reuse the rescued expense sheets from Odoo 8.0

StefanRijnhart:
The sheets where already preserved in https://github.com/OCA/OpenUpgrade/pull/2327, but new sheets were still created for the related expenses.

--


StefanRijnhart:
@pedrobaeza let me know if I missed something. Seems like something like this is needed to make #2327 work.

MiquelRForgeFlow:
by error you created in mrp a file named `fl`

StefanRijnhart:
Ah thanks. Removed now.

hbrunn (approved):

PR #2679, opened 2021-04-27
[13.0][OU] product: Empty company field in product if the instance only has one company.

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @chienandalu


pedrobaeza:
Not sure about this. When you create a new product in v13, `company_id` is filled by default with the company? If so, we shouldn't do this, but query Odoo to apply the same default on pricelists.

chienandalu (approved):

pedrobaeza:
And please remember to document the why of the action

PR #2689, opened 2021-04-29
(WIP)[11.0][IMP] stock: Fill price unit of stock_moves coming from inventory adjustments

AaronHForgeFlow:
In v10 and v13 (probably v11 and v12 as well) when there is an inventory adjustment adding qty the price unit in the stock_move is null and the cost in the quant is the standard price. Here we align the unit_price in the stock_move to the value in the quant.


AaronHForgeFlow:
Invalid as long price_unit will be zero in standard Odoo13, not valid for OU,

PR #2691, opened 2021-04-30
[13.0][FIX]stock_account: prevent empty layers to be created

AaronHForgeFlow:
For some reason it is possible to have qty zero in incoming layers.
```

This prevents:
devel=# SELECT count(*) from stock_valuation_layer WHERE quantity = 0.0;
 count 
-------
     8
(1 row)
```

@MiquelRForgeFlow


MiquelRForgeFlow (approved):

kos94ok-3D (approved):
I didn`t test it but look correct.

omar7r (approved):

PR #2712, opened 2021-05-07
[13.0][FIX]stock_account: transit stock move should not create valuation layers (WIP)

AaronHForgeFlow:
Stock moves to transit locations do not create stock valuation layers in version 13.0. Should we exclude those stock moves?

Still checking. cc @ForgeFlow

### DIFF
--- a/addons/stock_account/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock_account/migrations/13.0.1.1/post-migration.py
@@ -123,11 +123,11 @@ def get_stock_moves(env, company_id, product_id):
         SELECT sm.id, sm.company_id, sm.product_id, sm.date, sm.product_qty, sm.reference,
             COALESCE(sm.price_unit, 0.0) AS price_unit, rel.id AS account_move_id,
             sm.create_uid, sm.create_date, sm.write_uid, sm.write_date,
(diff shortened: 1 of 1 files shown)


kos94ok-3D:
I don`t know where you found this information because the code shows that transit has to generate SVL.
image
image
image

AaronHForgeFlow:
Yes, I see, maybe there is some customization affecting that behavior. For now I am closing this unless I discover anything new.

kos94ok-3D:
@AaronHForgeFlow This code show that SVL has to create but not always. See this parts:
image
image
image

AaronHForgeFlow:
I can see that if the transit location has a company assigned then svl are created. I will double check if the same criteria is happening in the migration. 

The problem I encountered is, for some products, the quantity of the quants is different than the remaining quantity in the SVL, and the products with that problem are usually moved across warehouses.

AaronHForgeFlow:
@kos94ok-3D  In the code you mentioned above, the methods _get_out_move_lines and _get_in_move_lines will return and empty recordset when:

*  Movements from internal location to a transit location in the same company
*  Movements from a transit location of a company to an internal location of the same company

For those cases no SVL are created. ~~However, it seems I am having SVL created for those moves in the migration~~. Of course, removing the condition as I did in this PR is also not correct, but the query may need to be adapted to this corner case.

Edit: SVL for transit* to internal and from internal to transit* are not created during the migration.
*(in the same company)

PR #2713, opened 2021-05-07
[9.0][IMP] re-enable used currencies

hbrunn:


StefanRijnhart (approved):

pedrobaeza:
I think the criteria is very partial, as you can use currencies in invoices without being assigned in an account. Maybe it's better to not include the noupdate data?

hbrunn:
why not add the invoice currency to the union too? I think it's a good thing that Odoo deactivates most currencies there, we just don't want used ones to be deactivated

pedrobaeza:
OK then.

hbrunn:
added the union over invoices

pedrobaeza (commented):
This will update as much times as records for each table, isn't it? Shouldn't you need to use `DISTINCT(currency_id)`?

hbrunn:
no, union (without all) takes care of duplicates

pedrobaeza:
Good to know, thanks.

PR #2714, opened 2021-05-07
[10.0][IMP] recompute invoice_status for po's with products invoiced by delivered quantities

hbrunn:


StefanRijnhart (approved):

StefanRijnhart:
For the record, this updates the result of https://github.com/OCA/OpenUpgrade/blob/9.0/addons/purchase/migrations/9.0.1.2/post-migration.py#L249-L285 with the logic introduced in https://github.com/odoo/odoo/pull/16098.

PR #2715, opened 2021-05-10
[13.0][OU-FIX] delivery: invoice_policy should be 'estimated' instead

MiquelRForgeFlow:
The `_create_delivery_line` method of sale.order (in delivery module) in v12 was using the unit_price, so it was as if the v13 invoice_policy was 'estimated'.

v12:

```python
values = {
            'order_id': self.id,
            'name': so_description,
            'product_uom_qty': 1,
            'product_uom': carrier.product_id.uom_id.id,
            'product_id': carrier.product_id.id,
            'price_unit': price_unit,
            'tax_id': [(6, 0, taxes_ids)],
            'is_delivery': True,
        }
```

v13:
```python
        values = {
            'order_id': self.id,
            'name': so_description,
            'product_uom_qty': 1,
            'product_uom': carrier.product_id.uom_id.id,
            'product_id': carrier.product_id.id,
            'tax_id': [(6, 0, taxes_ids)],
            'is_delivery': True,
        }
        if carrier.invoice_policy == 'real':
            values['price_unit'] = 0
            values['name'] += _(' (Estimated Cost: %s )') % self._format_currency_amount(price_unit)
        else:
            values['price_unit'] = price_unit
```

--

### DIFF
--- a/addons/delivery/migrations/13.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/delivery/migrations/13.0.1.0/openupgrade_analysis_work.txt
@@ -8,16 +8,14 @@ delivery     / delivery.carrier         / return_label_on_delivery (boolean): NE
 # NOTHING TO DO: New feature for allowing to print the return label. Default False value is OK for preserving previous behavior (no return label).
 
 delivery     / delivery.carrier         / invoice_policy (selection)    : NEW required, selection_keys: ['estimated', 'real'], req_default: function, hasdefault
-# DONE: post-migration: Fill all delivery.carrier records to have `real` value in invoice_policy, as this is the most similar one to the v12 behavior
+delivery     / sale.order               / invoice_shipping_on_delivery (boolean): DEL
+# DONE: post-migration: Set delivery.carrier.invoice_policy to 'real' only if invoice_shipping_on_delivery was True. Records by default should have `estimated` value in invoice_policy, as this is the most similar one to the v12 behavior (just compare _create_delivery_line method)
 
 delivery     / delivery.carrier         / margin (integer)              : type is now 'float' ('integer')
 # NOTHING TO DO: ORM/PG does the switch automatically
 
 delivery     / sale.order               / delivery_price (float)        : DEL
 # NOTHING TO DO: There's no carrier selection now in sale.order, only via wizard, so no computed price is shown in header
-
-delivery     / sale.order               / invoice_shipping_on_delivery (boolean): DEL
-# NOTHING TO DO: The control for adding delivery line on picking is done with delivery.carrier.invoice_policy
 
 delivery     / sale.order               / recompute_delivery_price (boolean): NEW
 # DONE: post-migration: marked as True if it has carrier and still quotation, for having the button "Update shipping cost" as highlighted
--- a/addons/delivery/migrations/13.0.1.0/post-migration.py
+++ b/addons/delivery/migrations/13.0.1.0/post-migration.py
@@ -3,12 +3,23 @@
 from openupgradelib import openupgrade
 
 
-def fill_delivery_carrier_invoice_policy(env):
-    """Fill all delivery.carrier records to have `real` value in invoice_policy,
-    as this is the most similar one to the v12 behavior.
+def update_delivery_carrier_invoice_policy(env):
+    """Update delivery.carrier records to have `real` value in invoice_policy
+    only if all its sale orders have invoice_shipping_on_delivery = True.
     """
     openupgrade.logged_query(
-        env.cr, "UPDATE delivery_carrier SET invoice_policy = 'real'"
+        env.cr, """
+        WITH carriers_to_real AS (
+            SELECT dc.id
+            FROM sale_order so
+            JOIN delivery_carrier dc ON so.carrier_id = dc.id
+            GROUP by dc.id
+            HAVING bool_and(so.invoice_shipping_on_delivery)
+        )
+        UPDATE delivery_carrier dc
+        SET invoice_policy = 'real'
+        FROM carriers_to_real
+        WHERE carriers_to_real.id = dc.id"""
     )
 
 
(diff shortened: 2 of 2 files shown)


JordiMForgeFlow:
I would also add that in V12, when adding the delivery cost the carrier price was computed if the order had the invoice_shipping_on_delivery True.

```
@api.multi
    def _add_delivery_cost_to_so(self):
        self.ensure_one()
        sale_order = self.sale_id
        if sale_order.invoice_shipping_on_delivery:
            carrier_price = self.carrier_price * (1.0 + (float(self.carrier_id.margin) / 100.0))
            sale_order._create_delivery_line(self.carrier_id, carrier_price)
```
However, by default this field was False:
`invoice_shipping_on_delivery = fields.Boolean(string="Invoice Shipping on Delivery", copy=False)`

pedrobaeza:
No, that's not correct. What we want is that the price is updated when validating picking, which is the v12 behavior.

MiquelRForgeFlow:
Updated code. @JordiMForgeFlow this is what you were pointing?

pedrobaeza:
As said, it's correct the other way for preserving v12 behavior.

JordiMForgeFlow:
> Updated code. @JordiMForgeFlow this is what you were pointing?

No, I was just adding another argument. There is no field invoice_shipping_on_delivery in delivery.carrier, i was just pointing out the default behaviour.

JordiMForgeFlow:
@pedrobaeza as I pointed out in a previous comment, the price is updated in the SO if it is marked as invoice_shipping_on_delivery. I don't think that the default behaviour is to have it True.

pedrobaeza:
You are mixing things... `invoice_shipping_on_delivery` is on sale.order, and it has nothing to do with the `real` counterpart, but a way to check if the delivery line was already added or not in the SO.

pedrobaeza:
If I don't remember bad, we saw this on the training, so please check the videos.

pedrobaeza:
This is still an error. You should always put `real`, which mimics more the v12 behavior, and then change it for the carriers you don't want to have it. It's not valid to look at SOs level.

MiquelRForgeFlow:
V12:

```python
@api.multi
    def _add_delivery_cost_to_so(self):
        self.ensure_one()
        sale_order = self.sale_id
        if sale_order.invoice_shipping_on_delivery:
            carrier_price = self.carrier_price * (1.0 + (float(self.carrier_id.margin) / 100.0))
            sale_order._create_delivery_line(self.carrier_id, carrier_price)
```


V13:

```python
    def _add_delivery_cost_to_so(self):
        self.ensure_one()
        sale_order = self.sale_id
        if sale_order and self.carrier_id.invoice_policy == 'real' and self.carrier_price:
            delivery_lines = sale_order.order_line.filtered(lambda l: l.is_delivery and l.currency_id.is_zero(l.price_unit) and l.product_id == self.carrier_id.product_id)
            carrier_price = self.carrier_price * (1.0 + (float(self.carrier_id.margin) / 100.0))
            if not delivery_lines:
                sale_order._create_delivery_line(self.carrier_id, carrier_price)
            else:
                delivery_line = delivery_lines[0]
                delivery_line[0].write({
                    'price_unit': carrier_price,
                    # remove the estimated price from the description
                    'name': sale_order.carrier_id.with_context(lang=self.partner_id.lang).name,
                })
```

As you can see here, having sale_order.invoice_shipping_on_delivery == True is equivalent to have self.carrier_id.invoice_policy == 'real'.

JordiMForgeFlow:
Also the invoice_policy is only relevant for integrated shipping methods. If we force all the methods to real in the V13 behaviour there is another problem which is that the shipping cost is not added in the SO.

pedrobaeza:
The trick is how is that field set. If you want to do this change, please do it in your private fork. For normal installations, real must be used for all.

pedrobaeza:
That's not really true. It's not added the same in one way or another (unless you do it manually). The change is the price. If you want to be auto-added, use the module `delivery_auto_refresh` that we have in OCA.

JordiMForgeFlow:
In my opinion I think that we should only set to real the carriers with external integration, in fact the field is only relevant and displayed in these cases:
`<field name="invoice_policy" widget="radio" attrs="{'invisible': ['|', ('delivery_type', 'in', ('fixed', 'base_on_rule')), ('integration_level', '=', 'rate')]}"/>`

pedrobaeza:
> In my opinion I think that we should only set to real the carriers with external integration, in fact the field is only relevant and displayed in these cases

No, as you can change the used carrier in the picking (example: you put SEUR in the SO, but at the end, it's DHL). You want that the SO reflects the price for DHL, not for SEUR. That's the most conservative approach and in fact you shouldn't use any other except `real` in all your customers.

JordiMForgeFlow:
I am not discussing that this may be the most coherent approach, but I don't see that this behaviour is the default one in V12.
If I add the shipping costs in the SO, and then change the shipping method on the picking, when processing the picking the price is not updated in the SO.

pedrobaeza:
But that's a different debate. What we are discussing here is the method for delivery carriers on v13 from now on, and that one should be independent from what you have done in the past (adding or not the delivery line before confirming it).

MiquelRForgeFlow:
> What we are discussing here is the method for delivery carriers on v13 from now on, and that one should be independent from what you have done in the past

Sorry, I am lost now. What the migration should do is trying to mimic the older behavior for old records. Thus, I don't understand your point. As @JordiMForgeFlow said, the default behavior of v12 was not the 'real' thing, as the boolean invoice_shipping_on_delivery was False by default.

pedrobaeza:
Let's meet to explain you.

StefanRijnhart (approved):

pedrobaeza (changes requested):
We still don't have this clear

pedrobaeza:
We are also in a not comfortable with the `real` option now that we have seen more cases, and we are proposing to go to `estimated`. Please check #2913 

The problem is that none of the current ones are exactly the same as the options in v12, and you will find several drawbacks in one or the other.

PR #2729, opened 2021-05-15
update to v12 l10n-brazil module layout

rvalyi:
Hello, from v7 to v10 the OCA/l10n-brazil kept roughly the same module layout, addressing only the electromic invoice document. In v12 however, while aiming at full fiscal documents support, we extracted an independent l10n_br_fiscal module encapsulating most of the fiscal logic out from the former e-invoicing l10n_br_account_product and a few other modules got renamed or merged. So far this what we are using as the apriori.py file when migrating customers from v10 to v12. Pretty much production battle tested already.

While this merge can be done safely no matter what, it's worth mentionning I recently pushed our latest OpenUpgrade script  for migrating OCA/l10n-brazil from v10 to v12:
https://github.com/OCA/l10n-brazil/pull/1101
https://github.com/OCA/l10n-brazil/pull/1353
https://github.com/OCA/l10n-brazil/pull/1354

cc @renatonlima @marcelsavegnago @mileo 

--

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -33,6 +33,10 @@ renamed_modules = {
     'stock_pack_operation_auto_fill': 'stock_move_line_auto_fill',
     # OCA/web
     'web_advanced_filters': 'web_advanced_filter',
+    # OCA/l10n-brazil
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
seems huge refactoring !

Optimistic approval.

PR #2730, opened 2021-05-15
[13.0][FIX] base: Don't rename category if hierarchy exists

pedrobaeza:
Coming from previous versions, you can already have `Manufacturing/Manufacturing` in module category, and with current
renaming scheme, the parent category is merged in the child, but the parent_id field points then to itself.

We avoid the problem not doing anything in this case.

@Tecnativa TT28427

### DIFF
--- a/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
@@ -760,6 +760,15 @@ def rename_ir_module_category(env):
         old_row = env.cr.fetchone()
         if old_row:
             if new_row:
+                query = "SELECT parent_id FROM ir_module_category WHERE id = %s"
+                env.cr.execute(query, (new_row[0], ))
+                new_parent_row = env.cr.fetchone()
+                if new_parent_row and new_parent_row[0] == old_row[0]:
+                    # When we already have recursive categories with same name
+                    # (example: Manufacturing/Manufacturing), doing a merge let
+                    # the akward situation where the parent points to itself,
+                    # so we avoid it checking this condition
+                    continue
                 openupgrade_merge_records.merge_records(
                     env, "ir.module.category", [old_row[0]], new_row[0],
                     method="sql", model_table="ir_module_category")


odoo/addons/base/migrations/13.0.1.3/pre-migration.py
```diff
@@ -760,6 +760,15 @@ def rename_ir_module_category(env):
         old_row = env.cr.fetchone()
         if old_row:
             if new_row:
+                query = "SELECT name FROM ir_module_category WHERE id = %s"
+                old_name = env.cr.execute(query, (old_row[0], ))
+                new_name = env.cr.execute(query, (new_row[0], ))
+                if old_name == new_name:
```
kos94ok-3D: where is `fetchone()`?

pedrobaeza: Yeah, semi-typo one... fix on the way.

pedrobaeza: Fixed

MiquelRForgeFlow:
I don't quite understand, but if you assure this will not give problems, then ok :+1: for me.

pedrobaeza:
This is a real v12 ir_module_category state:

Selección_006

MiquelRForgeFlow:
This is a case of a child merging to a parent, which should not have problems. In your initial post, you spoke about a parent merging into its child, which is different, and yeah, may give problems. Thus, the condition should not be `old.name == new.name`, but `old.id == new.parent_id`.

pedrobaeza:
It's the same. Maybe the order is the contrary, but at the end, I end up with only one ir_module_category having `parent_id` pointing to itself.

MiquelRForgeFlow:
If you have a child merging to its parent you don't end having a category with a `parent_id` pointing to itself. Thus it's not the same. Thus, the fine condition is `old.id == new.parent_id`, which excludes only the problematic case, which is a parent merging into its child.

pedrobaeza:
I still don't get you. Can you write the exact code you are proposing?

MiquelRForgeFlow:
```python
if old_row:
            if new_row:
                query = "SELECT parent_id FROM ir_module_category WHERE id = %s"
                env.cr.execute(query, (new_row[0], ))
                new_paren_id = env.cr.fetchone()
                if new_paren_id and old_row[0] == new_paren_id[0]:
                    # When we already have recursive categories, doing a 
		    # merge of a parent into one of its childs let
                    # the akward situation where the parent points to itself,
                    # so we avoid it checking this condition
                    continue
                openupgrade_merge_records.merge_records(...
```

pedrobaeza:
Miquel, changed according your proposal, and trying now in the conflicting DB.

pedrobaeza:
OK, working in the problematic DB, so this one is good to go.

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
```diff
@@ -760,6 +760,15 @@ def rename_ir_module_category(env):
...
         if old_row:
             if new_row:
+                query = "SELECT parent_id FROM ir_module_category WHERE id = %s"
+                env.cr.execute(query, (new_row[0], ))
+                new_parent_row = env.cr.fetchone()
+                if new_parent_row and new_parent_row[0] == old_row[0]:
+                    # When we already have recursive categories with same name
+                    # (example: Manufacturing/Manufacturing), doing a merge let
+                    # the akward situation where the parent points to itself,
+                    # so we avoid it checking this condition
```
MiquelRForgeFlow, line 770:
```suggestion
                    # When we already have recursive categories, doing a 
                    # merge of a parent into one of its childs let
                    # the akward situation where the child points to itself,
                    # so we avoid it checking this condition
```
maybe you would like to rewrite it a bit.

MiquelRForgeFlow (approved):

odoo/addons/base/migrations/13.0.1.3/pre-migration.py
pedrobaeza: Not really, as it's important that `Manufacturing/Manufacturing` text appears as reference. The rest helps, but I think it's enough for not launching another CI. Your call BTW.

MiquelRForgeFlow: ok, then merge

pedrobaeza: Thanks for your comprehension.

PR #2733, opened 2021-05-18
[13.0][FIX] account: use old customer/supplier fields for ranks

MiquelRForgeFlow:
If in v12 some partners were flagged as customer or supplier but they were not in any move line, then they will have a customer_rank/supplier_rank = 0, which is not expected.

--


pedrobaeza (approved):

kos94ok-3D:
Maybe for the supplier is OK, but the customer True by default.
I think that majority of people don`t set False it.

pedrobaeza:
But that doesn't prevent that this is keeping the "behavior" of v12 in a more accurate way. Is it interesting to keep it in this way? I think it doesn't hurt this way and more important: you don't get a different count when you are accessing to "Customers" menu.

kos94ok-3D (approved):

PR #2736, opened 2021-05-20
[9.0][IMP] product: migrate simple purchase pricelists

hbrunn:

Current behavior before PR:

--

### DIFF
--- a/addons/product/migrations/9.0.1.2/post-migration.py
+++ b/addons/product/migrations/9.0.1.2/post-migration.py
@@ -280,6 +280,9 @@ def migrate_purchase_pricelists(env):
     """ Create product.supplierinfo records for "simple" pricelist items,
     those are items based on the cost price (not other lists or items)
     or items with price_discount == -1 (fixed prices)"""
+    env.cr.execute(
+        """ALTER TABLE product_supplierinfo
+        ADD COLUMN openupgrade_migrated_from_pricelist_item_id INTEGER""")
     openupgrade.logged_query(
         env.cr,
         """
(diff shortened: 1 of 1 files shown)


addons/product/migrations/9.0.1.2/post-migration.py
```diff
@@ -276,6 +277,79 @@ def update_product_supplierinfo(env):
             AND rc.currency_id != pp.currency_id """)
 
 
+def migrate_purchase_pricelists(env):
+    """ Create product.supplierinfo records for "simple" pricelist items,
+    those are items based on a field (not other pricelist or supplier items)"""
+    env.cr.execute("select id, field from product_price_type")
```
astirpe: Considering that it's about purchase pricelists, I think here we should filter field "standard_price" only:
```suggestion
    env.cr.execute("select id, field from product_price_type where field = 'standard_price'")
```

addons/product/migrations/9.0.1.2/post-migration.py
```diff
@@ -276,6 +277,79 @@ def update_product_supplierinfo(env):
...
+            )
+            insert into product_supplierinfo
+            (
+                name, min_qty, currency_id, date_start, date_end,
+                company_id, product_id, product_tmpl_id, price
+            )
+            select
+            partner_id, min_quantity, currency_id, p2i.date_start, p2i.date_end,
+            p2i.company_id, p2i.product_id, p2i.product_tmpl_id,
+            coalesce(pt1.%(field_name)s, pt2.%(field_name)s) *
```
astirpe: In case `%(field_name)s` is `standard_price` (and this should be the only case IMO) we should take in consideration that in V9 `standard_price` is a property of `product.product` (and not of `product.template`).

To retrieve the proper product prices this should work:

```sql
	select
	    split_part(res_id, ',', 2)::int product_id,
	    value_float amount,
	    company_id
	from ir_property
	where
	    fields_id in (
		    select res_id from ir_model_data
		    where name like 'field_product_product_standard_price'
	    ) and res_id is not null and value_float is not null
```

addons/product/migrations/9.0.1.2/post-migration.py
```diff
@@ -276,6 +277,79 @@ def update_product_supplierinfo(env):
...
+            (1 + price_discount) + price_surcharge
```
astirpe: ~Instead of `(1 + price_discount)` shouldn't be `(1 - 1 * (price_discount / 100))` ?~
 Sorry, ignore this comment. Here `price_discount` belongs to `product_pricelist_item` so `(1 + price_discount)` is correct.

astirpe (changes requested):

addons/product/migrations/9.0.1.2/post-migration.py
hbrunn: thanks for your input, the moving of the property to product.product happens indeed before this happens. Now this should be finr

StefanRijnhart:
@hbrunn given that we had something of a flood of vendor prices after the migration, there may be room for improvement.

pedrobaeza (approved):

PR #2737, opened 2021-05-21
[11.0][OU-FIX] mrp: fix wrong creation of stock move lines

JordiMForgeFlow:
In V10, when the stock moves in a manufacturing order are assigned, stock move lots are always created for them (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/mrp/models/stock_move.py#L140).

However, these stock move lots are only linked to a stock production lot if the quants are related to one (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/mrp/models/stock_move.py#L174).

Currently, the migration script is inserting lines for stock moves in manufacturing orders that don't have a stock production lot related, which does not make sense as the stock move lots are not related to a lot. We should prevent considering these stock moves when no lots are used by taking only the ones related to one.

--

### DIFF
--- a/addons/mrp/migrations/11.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/11.0.2.0/post-migration.py
@@ -100,7 +100,7 @@ def create_stock_move_lines_from_stock_move_lots(env):
         )
         SELECT %(select)s
         FROM %(from)s
-        WHERE sm.state NOT IN ('cancel', 'confirmed')
+        WHERE sm.state NOT IN ('cancel', 'confirmed', 'waiting')
             AND sm.id NOT IN (SELECT sq.reservation_id
                               FROM stock_quant sq
                               WHERE sq.reservation_id IS NOT NULL)
@@ -148,7 +148,7 @@ def create_stock_move_lines_from_stock_move_lots(env):
             sml.workorder_id"""
     from_ = """stock_move sm
         INNER JOIN mrp_production mp ON sm.production_id = mp.id
-        LEFT JOIN stock_move_lots sml ON sml.move_id = sm.id
+        INNER JOIN stock_move_lots sml ON sml.move_id = sm.id
         LEFT JOIN stock_production_lot spl ON sml.lot_id = spl.id"""
     openupgrade.logged_query(
         env.cr, """


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM

pedrobaeza:
AFAIK, products without lots still require stock.move.line records.

JordiMForgeFlow:
But is this not already done in the migration of stock (https://github.com/OCA/OpenUpgrade/blob/11.0/addons/stock/migrations/11.0.1.1/post-migration.py#L326) ?

pedrobaeza:
Then what I don't understand is why you need an extra insert here for this case.

JordiMForgeFlow:
Digging into https://github.com/OCA/OpenUpgrade/pull/1597, I think that the method is aimed at creating the stock move lines for production moves that are done, as they don't have linked move operations.

I thought that the problem I am facing was caused by the relation with lots but I think that it is caused by another issue in the script. I will propose another PR once I have clarified it more :)

JordiMForgeFlow:
With the current migration scripts, I am facing problems caused in the two queries of the method create_stock_move_lines_from_stock_move_lots:

In the first query, we are creating stock move lines for raw stock moves in MO. The problem is that, in the case where the raw move is linked to a previous operation, we are still creating the stock move line when we should not, as there are no quants reserved.
To fix it, we should prevent considering the raw moves that are waiting another move.

In the second query, we are creating stock move lines for finished moves in MO. Finished moves will not have stock move lots related until the MO is not produced (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/mrp/wizard/mrp_product_produce.py#L102). Therefore, the query should only consider the finished moves that have them, as they are the ones that will require having stock move lines created.

cc @pedrobaeza @LoisRForgeFlow

MiquelRForgeFlow (approved):
Good work!

pedrobaeza:
I can't say too much about this, as I didn't intervene in the `mrp` migration scripts too much, and creating here `stock.move.line` still sounds weird, but if you have it clear, let's merge it.

PR #2739, opened 2021-05-21
[13.0][FIX] account: Don't have duplicated invoice line links

pedrobaeza:
With previous code, same invoice line may be linked to several account move lines if there are several of them with same product,
quantity, amount and the rest of the distinct data that differentiates them, provoking later problems on the rest of things that depend on this (sales lines linked to the old invoice line for example).

With this iterative algorithm based on a divide-and-conquer strategy, we fill according that criteria, but excluding already linked invoice lines, and emptying duplicates from the second appearance onwards. The worst case will mean at least as many loops as the number of identical invoice lines we find in one invoice, but the time required for the SQLs is decreasing, so we can end the loop on an affordable balance between spent time and results.

Second attempt after reverting due to performance problems #2614. This one has been inspired on that one, but with several key changes for performance:

- Don't do a costly extra LEFT JOIN over invoice lines.
- Cache in an extra column if an invoice line has been already linked.
- Use such cached value for excluding invoice lines for next iteration.

Co-Authored-By: @MiquelRForgeFlow 

@Tecnativa TT29776

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -213,7 +213,9 @@ def migration_invoice_moves(env):
             env, ids1 + ids2, 'account.invoice', 'account.move')
     # Not Draft or Cancel Invoice Lines
     # 1st: update the ungrouped ones
-    query = """
+    openupgrade.logged_query(env.cr, "ALTER TABLE account_invoice_line ADD aml_matched BOOLEAN")
+    query = sql.SQL("""
+    WITH sub AS (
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm doing a real test for seeing if the performance degradation is acceptable.

pedrobaeza:
In a migration spending 3 hours and 40 minutes before this, with this PR takes 3 hours and 57 minutes, which means about 17 minutes more. The number of account.invoice.line is 2.291.178. I consider it as reasonable then.

MiquelRForgeFlow (approved):

PR #2740, opened 2021-05-22
[11.0][OU-FIX] mrp: fix wrong creation of stock move lines #2737

JordiMForgeFlow:
In the fix in #2737 I oversee that for finished moves the stock move lots are only created when the product has some form of tracking (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/mrp/wizard/mrp_product_produce.py#L105), which is weird as it does not perform this distinction with raw moves (https://github.com/OCA/OpenUpgrade/blob/10.0/addons/mrp/models/stock_move.py#L163).

With the previous fix, for products that do not have tracking the stock move lines for finished moves are not generated. However, in these cases, the stock move lines should only be created when the move has been done.

--

### DIFF
--- a/addons/mrp/migrations/11.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/11.0.2.0/post-migration.py
@@ -148,7 +148,7 @@ def create_stock_move_lines_from_stock_move_lots(env):
             sml.workorder_id"""
     from_ = """stock_move sm
         INNER JOIN mrp_production mp ON sm.production_id = mp.id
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow:
@MiquelRForgeFlow as you developed the migration scripts for MRP could you please help me on this one to see if we finally solve it? :)

JordiMForgeFlow:
cc @LoisRForgeFlow

MiquelRForgeFlow (approved):
Ok, this seems better.

PR #2670, opened 2021-04-21
[14.0][MIG] base_automation

RaulOForgeFlow:
---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2741, opened 2021-05-25
[14.0][MIG] account

RaulOForgeFlow:


StefanRijnhart:
In the pre-script, please rename account_move.move_type if it exists, to `None`. It will be picked up by https://github.com/OCA/account-financial-reporting/blob/14.0/account_tax_balance/migrations/14.0.1.0.0/pre-migration.py#L17-L21

MiquelRForgeFlow:
PR is ready. Meanwhile it has same kind of error as https://github.com/OCA/OpenUpgrade/pull/2670.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,112 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    fill_foreign_currency_id_and_currency_id_field(env)
+    fill_code_prefix_end_field(env)
+    fill_default_account_id_field(env)
+    fill_payment_id_and_statement_line_id_fields(env)
+    fill_active_field_account_reconcile_model(env)
+    fill_matching_order_field_default(env)
+    openupgrade.load_data(env.cr, "account", "14.0.1.0/noupdate_changes.xml")
```
gfcapalbo: will have to be 14.0.1.1 when ready.

MiquelRForgeFlow: It was done in the second commit.

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,648 @@
...
+new model sequence.mixin [abstract]
+# NOTHING TO DO: abstract and transient models
+
+---Fields in module 'account'---
+account      / account.account          / _order                        : _order is now 'is_off_balance, code, company_id' ('code')
+account      / account.account          / is_off_balance (boolean)      : NEW hasdefault, isfunction: function, stored
+# NOTHING TO DO: new field added to filter accounts.
+
+account      / account.account          / allowed_journal_ids (many2many): NEW relation: account.journal
+# NOTHING TO DO: New feature. In v13 you selected the allowed accounts form the journal form, now you can select the allowed journals from the account form
```
StefanRijnhart:
```suggestion
# DONE: renamed many2many table. In v13 you selected the allowed accounts from the journal form, now you can select the allowed journals from the account form
```

MiquelRForgeFlow: oh, not really, don't be confused :S

In `account.journal` you still have `account_control_ids = fields.Many2many('account.account', 'journal_account_control_rel', 'journal_id', 'account_id', string='Allowed accounts'`, which this field already existed in v13 (but the table changed and was renamed in pre-migration).

And then in `account.account` you have `allowed_journal_ids = fields.Many2many('account.journal', string="Allowed Journals"` which uses `account_account_account_journal_rel` table which is a true new feature if you analyze the odoo code.

account_account_account_journal_rel != journal_account_control_rel

MiquelRForgeFlow: I clarified the comment in the txt.

StefanRijnhart: Oh man. I can't believe they did this. Thanks for clarifying!

StefanRijnhart: Then again, I guess it makes sense because it can be the case that you want to restrict an account to only be used in a specific journal but at the same time do not want to restrict that journal for all other accounts.

StefanRijnhart:
The error on missing models when deleting selection fields is fixed in https://github.com/OCA/OpenUpgrade/pull/2795

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,383 @@
...
+        """,
+    )
+
+
+def fill_active_field_account_reconcile_model(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_reconcile_model
+        SET active = TRUE
```
StefanRijnhart: You don't have to set the value here manually, because the field has this exact same value as its default, and if a field has a default the ORM will set it on new fields on existing models. This is why the analysis mentions `hasdefault` as a hint that perhaps the column does not need to be populated manually.

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,648 @@
...
+account      / account.bank.statement.line / date (date)                   : not stored anymore
+account      / account.bank.statement.line / date (date)                   : now related
+account      / account.bank.statement.line / ref (char)                    : not stored anymore
+account      / account.bank.statement.line / ref (char)                    : now related
+account      / account.bank.statement.line / journal_id (many2one)         : not stored anymore
+account      / account.bank.statement.line / journal_id (many2one)         : now required
+# NOTHING TO DO: These fields now are in account.move
+
+account      / account.bank.statement.line / note (text)                   : DEL
+# NOTHING TO DO
```
StefanRijnhart: Do you think maybe the note contents can be appended to the related move's narration column? This field is displayed in statement line context: https://github.com/odoo/odoo/blob/14.0/addons/account/views/account_bank_statement_views.xml#L290

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,648 @@
...
+
+account      / account.group            / _order                        : _order is now 'code_prefix_start' ('code_prefix')
+account      / account.group            / code_prefix (char)            : DEL
+account      / account.group            / code_prefix_end (char)        : NEW
+account      / account.group            / code_prefix_start (char)      : NEW
+# DONE: pre-migration: renamed code_prefix to code_prefix_start
+# post-migration: assigned same start values to end field
+
+account      / account.group            / company_id (many2one)         : NEW relation: res.company, required, req_default: function, hasdefault
+# NOTHING TO DO
```
StefanRijnhart: Maybe change to TODO, because it will be the case that groups are shared between companies, for instance if this Spanish localization was installed in more than one company: https://github.com/odoo/odoo/blob/13.0/addons/l10n_es/data/account_group.xml

StefanRijnhart:
Thank you for this huge work!

A comment about the noupdate data: it contains an update of the body_html of `email_template_edi_invoice` and `mail_template_data_payment_receipt`. Can you delete their translations in the pre-migration so that these do not fall behind the update?

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
MiquelRForgeFlow: Not sure if we have to do the same with the communication field of account.payment

StefanRijnhart: Ah, account.payment now also inherits from account.move. In that case I would say, yes.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
MiquelRForgeFlow: done

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
MiquelRForgeFlow: done

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
MiquelRForgeFlow: done, also I added script to solve the issue

MiquelRForgeFlow:
@StefanRijnhart I deleted the template translations in post-migration because historically it has been done in post-migration.

StefanRijnhart:
Thanks for the updates! 

> I deleted the template translations in post-migration because historically it has been done in post-migration.

That surprises me, because I think by then you are too late for the new translations to load. Although it will mostly be the case that an --update all is executed sooner or later on the database in its production environment, the pre-migration script is the right place to do it.

pedrobaeza:
> Thanks for the updates!
> 
> > I deleted the template translations in post-migration because historically it has been done in post-migration.
> 
> That surprises me, because I think by then you are too late for the new translations to load. Although it will mostly be the case that an --update all is executed sooner or later on the database in its production environment, the pre-migration script is the right place to do it.

I think that unless you load noupdate changes also at pre, you won't get updated translations.

StefanRijnhart:
> I think that unless you load noupdate changes also at pre, you won't get updated translations.

I just did a small test, and if you delete the translations of a noupdate record in Odoo 14, the translations are reloaded during the next module update, and applied in the UI whether the src value in ir_translation matches the value in the model table or not. So based on this test it seems that deleting translations in the pre-migration and loading noupdate data in the post-migration will work fine.

pedrobaeza:
OK, then I think we can apply such technique from now on.

MiquelRForgeFlow:
> OK, then I think we can apply such technique from now on.

@StefanRijnhart but why doing now such a thing in pre-migration if doing it in post-migration is not wrong and tested a lot? I mean, what is the advantage?

StefanRijnhart:
> @StefanRijnhart but why doing now such a thing in pre-migration if doing it in post-migration is not wrong and tested a lot? I mean, what is the advantage?

Yes, valid point. I'll check if the translations are reloaded when deleted in the post-migration script.

StefanRijnhart:
@MiquelRForgeFlow you were right. The translations are loaded after the post-migration.

```
2021-07-08 17:04:07,672 1058908 INFO 14base odoo.modules.migration: module base: Running migration [14.0.1.5>] post-migrate 
> /home/odoo/odoo/addons/base/migrations/14.0.1.5/post-migrate.py(5)migrate()->None
2021-07-08 17:04:10,183 1058908 INFO 14base odoo.addons.base.models.ir_translation: module base: loading translation file (nl) for language nl_NL 
```

Sorry for the noise!

MiquelRForgeFlow:
Ready :green_apple: 

> Sorry for the noise!

Don't worry :)

StefanRijnhart:
Running the migration of commit d127ae95a on a database with 2 companies with l10n_es chart of account, I get
```
  File "/home/odoo/openupgrade/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py", line 373, in fix_group_company_and_create_groups_for_each_company_if_necessary
    new_group.parent_id = relation_dict[company.id][
  File "/home/odoo/odoo/odoo/fields.py", line 1127, in __set__
    records.write({self.name: write_value})
  File "/home/odoo/odoo/addons/account/models/account_account.py", line 596, in write
    res = super(AccountGroup, self).write(vals)
  File "/home/odoo/odoo/odoo/models.py", line 3681, in write
    self.flush([self._parent_name])
  File "/home/odoo/odoo/odoo/models.py", line 5475, in flush
    process(self.env[model_name], id_vals)
  File "/home/odoo/odoo/odoo/models.py", line 5430, in process
    recs._write(vals)
  File "/home/odoo/odoo/odoo/models.py", line 3763, in _write
    parent_records._parent_store_update()
  File "/home/odoo/odoo/odoo/models.py", line 4118, in _parent_store_update
    parent_ids = {int(label) for label in prefix.split('/')[:-1]}
  File "/home/odoo/odoo/odoo/models.py", line 4118, in <setcomp>
    parent_ids = {int(label) for label in prefix.split('/')[:-1]}
ValueError: invalid literal for int() with base 10: ''
```

Here is the database if it helps: [13es.psql.zip](https://github.com/OCA/OpenUpgrade/files/6786365/13es.psql.zip)

MiquelRForgeFlow:
@StefanRijnhart Solved your issue :)

StefanRijnhart:
> @StefanRijnhart Solved your issue :)

Haha, this is purely on behalf of the Spanish localization :stuck_out_tongue: 

Works perfectly now, thanks!

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,479 @@
...
+        UPDATE account_bank_statement bst
+        SET state = 'posted'
+        FROM account_bank_statement_line bstl
+        WHERE bst.state = 'confirm' AND bstl.statement_id = bst.id
+            AND bstl.is_reconciled IS DISTINCT FROM TRUE
+        """,
+    )
+
+
+def pass_bank_statement_line_note_to_invoice_narration(env):
```
StefanRijnhart: I believe the related `account.move`s are 'journal entry' move, not invoice moves (damn you account-pocalypse) so the method names here and the one down below are slightly misleading.

StefanRijnhart (approved):
Excellent work! The TODOs are clearly marked for those who come across the related issues in practice.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
MiquelRForgeFlow: methods renamed hehe

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,379 @@
...
+                ],
+            },
+        )
+
+
+def add_move_id_field_account_bank_statement_line(env):
+    if not openupgrade.column_exists(
+        env.cr, "account_bank_statement_line", "currency_id"
+    ):
+        openupgrade.logged_query(
```
gfcapalbo, line 154: getting an error "column statement_line exists in relation 'account_move', checking if it's data-specific or a recurring problem. will follow up in 30m

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        ALTER TABLE account_bank_statement_line
+        ADD COLUMN move_id integer
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE account_move
+        ADD COLUMN statement_line_id integer
```
gfcapalbo: getting an error "column statement_line_id exists in relation 'account_move', checking if it's data-specific or a recurring problem. will follow up in 30m
A check like done above may be in order.

StefanRijnhart: Is it this one? 

ou110/parts/odoo/addons/account/migrations/11.0.1.1/openupgrade_analysis.txt:account      / account.move             / statement_line_id (many2one)  : DEL relation: account.bank.statement.line

MiquelRForgeFlow: when older versions sneakily strike haha solved :)

StefanRijnhart:
@gfcapalbo this one alright for you now?

mdemaison:
Hi , i think there is a mistake here  in script/account/14.0.1.1 in pre-migration.py line 218
` openupgrade.logged_query(
        env.cr,
        """
        UPDATE account_bank_statement_line absl
        SET move_id = ap.move_id
        FROM account_move am
        WHERE absl.move_id IS NULL AND absl.account_number NOT IN ('', '/')
            AND absl.account_number = am.ref
        """,
    )
should be :
openupgrade.logged_query(
        env.cr,
        """
        UPDATE account_bank_statement_line absl
        SET move_id = ap.move_id
        FROM account_move am, account_payment ap
        WHERE absl.move_id IS NULL AND absl.account_number NOT IN ('', '/')
            AND absl.account_number = am.ref
        """,
    )
hope it help

`

MiquelRForgeFlow:
@mdemaison it seems you don't have last version. Make sure you fetch and pull last version of the branch.

StefanRijnhart:
@MiquelRForgeFlow what is the next step with all the reviews of all these modules? Your company are doing the work for the tender of the OCA's own database right? Does that tender entail that someone from your company will do a review of the PR's or are you leaving it up to the volunteers of the project? I imagine one review from someone from your company and one community review would be a good combination.

pedrobaeza:
@StefanRijnhart included in the tender is the review by my part, that I'll do during this month.

StefanRijnhart:
@pedrobaeza thanks for clarifying!

lmmar-ic:
Hi, In the migration it has given me the following error:

2021-10-06 15:38:17,545 9528 ERROR openupgrade odoo.schema: Table 'ir_model_data': unable to add constraint 'ir_model_data_name_nospaces' as CHECK(name NOT LIKE '% %')
2021-10-06 15:38:17,559 9528 ERROR openupgrade odoo.schema: Table 'account_move_line': unable to add constraint 'account_move_line_check_amount_currency_balance_sign' as CHECK(
                (
                    (currency_id != company_currency_id)
                    AND
                    (
                        (debit - credit <= 0 AND amount_currency <= 0)
                        OR
                        (debit - credit >= 0 AND amount_currency >= 0)
                    )
                )
                OR
                (
                    currency_id = company_currency_id
                    AND
                    ROUND(debit - credit - amount_currency, 2) = 0
                )
            )
2021-10-06 15:38:17,561 9528 ERROR openupgrade odoo.schema: Table 'account_payment': unable to set NOT NULL on column 'move_id'
2021-10-06 15:38:17,563 9528 ERROR openupgrade odoo.schema: Table 'account_bank_statement_line': unable to set NOT NULL on column 'move_id'

Despite the error, the migration continues and verifying the result, I have noticed the following:

- All invoices have the total amount of taxes and withholdings at 0, the subtotal and total of the invoices is correct.
- Regarding the reconciliation, there are two problems, the line in the invoice, which indicates when it was paid, indicates amount 0 and if we give it to publish a migrated bank statement in New status, it gives a permission error and does not allow publishing.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,590 @@
...
+                "statement_line_id": st_line.id,
+                "move_type": "entry",
+                "journal_id": st_line.statement_id.journal_id.id,
+                "company_id": st_line.statement_id.company_id.id,
+                "currency_id": st_line.statement_id.journal_id.currency_id.id
+                or st_line.statement_id.company_id.currency_id.id,
+            }
+        )
+        st_line.move_id = move
+        st_line._synchronize_to_moves(
```
jcdrubay: This line triggers:

`odoo.exceptions.UserError: You cannot add/modify entries prior to and inclusive of the lock date 01/01/2019.`

Maybe an earlier step could update the lock dates?

If using the module `account_journal_lock_date`, we might need to do  use context "bypass_journal_lock_date".

jcdrubay: I guess this should do the trick: 
`update res_company set period_lock_date=Null,fiscalyear_lock_date=Null;`

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,590 @@
...
+                "amount",
+                "payment_type",
+                "partner_type",
+                "payment_reference",
+                "is_internal_transfer",
+                "currency_id",
+                "partner_id",
+                "destination_account_id",
+                "partner_bank_id",
+                "journal_id",
```
jcdrubay: This one raises : `odoo.exceptions.UserError: You can't create a new payment without an outstanding payments/receipts account set on the Bank *** journal.`

jcdrubay: This seems to solve the issue when added at the beginning of this function:

```python
    env.cr.execute(
        """
        update account_journal
        set payment_debit_account_id=default_account_id
        where
            type='bank' and
            payment_debit_account_id is NULL;
        """
    )

    env.cr.execute(
        """
        update account_journal
        set payment_credit_account_id=default_account_id
        where
            type='bank' and
            payment_credit_account_id is NULL;
        """
    )

```

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,590 @@
...
+                "payment_id": payment.id,
+                "move_type": "entry",
+                "journal_id": journal.id,
+                "company_id": journal.company_id.id,
+                "currency_id": journal.currency_id.id
+                or journal.company_id.currency_id.id,
+            }
+        )
+        payment.move_id = move
+        payment._synchronize_to_moves(
```
jcdrubay: Another error from this function : `odoo.exceptions.UserError: The account Bank *** is deprecated.`

jcdrubay: As a workaround:

I guess we could do:

- List the deprecated accounts: `select id,name from account_account where deprecated='t';` 
- Set them as not deprecated: `update account_account set deprecated='f' where deprecated='t';`
- Run the _synchronize_to_moves.
- Deprecate the account which used to be deprecated: `update account_account set deprecated='t' where id in (?);`

pedrobaeza: All these things must be with SQL queries, not with ORM, or all these side effects will happen.

jcdrubay: You mean to re-implement `_synchronize_to_moves` account_payment.py, but bypass the ORM at the time to do the write?

pedrobaeza: Yes

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,590 @@
...
+        )
+        to_write = {
+            "line_ids": [
+                (0, 0, line_vals)
+                for line_vals in payment._prepare_move_line_default_vals(
+                    write_off_line_vals=False
+                )
+            ]
+        }
+        payment.move_id.write(to_write)
```
jcdrubay: This one triggers: 
```
odoo.exceptions.UserError: The journal entry Draft Entry (* 14680) reached an invalid state relative to its payment. To be consistent, the journal entry must always contains:
- one journal item involving the outstanding payment/receipts account.
- one journal item involving a receivable/payable account.
- optional journal items, all sharing the same account.
```

That Journal Entry is a manually created "Miscellaneous" Journal Entry reconciled with a Customer Invoice. That Journal Entry was created to reconcile invoices which will never get paid with a Journal Entry involving the accounts Receivable and Bad Debt. As a result, there is no related record of Account Payment.

Converting this single line calling the ORM to SQL won't be trivial.

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,381 @@
...
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line absl
+        SET move_id = am.id
+        FROM account_move am
+        WHERE (am.name NOT IN ('', '/') AND COALESCE(
+            NULLIF(absl.move_name, ''), absl.payment_ref) = am.name)
+            OR am.statement_line_id = absl.id
```
omar7r: You have to check the company_id too, entry name is unique but inside the company.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
MiquelRForgeFlow: It's impossible to trigger this error in this line. `st_line._synchronize_to_moves(...)` does a write in move for fields partner_id, currency_id and line_ids. But `move._check_fiscalyear_lock_date()` only is called is you write in move for field date.

MiquelRForgeFlow:
@jcdrubay I hope you can check again, to see if I solved all your problems.

jcdrubay:
I will test again tomorrow, thanks for the ping @MiquelRForgeFlow and thanks a lot for this contribution. :)

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
jcdrubay: I don't know how it is possible, but I have some old data which do not have the field date set:
This query returns 9 records from 2018 (when I was probably using an older version) `select id,name from account_move_line where date is NULL;`.
So the real source if the issue might be an old version of an OpenUpgrade.

Anyway, I will fix manually these few records: 
`update account_move_line set date=create_date where date is NULL;`

jcdrubay: Actually, it's even more strange, because in the source database in v13, I did not have account_move_line without dates. I ll keep digging...

jcdrubay:
@MiquelRForgeFlow After re-running, all worked well!

The only issues I faced were due to old data which had inconsistent currencies between account / journal / account move line. These issues were faced when running the function `fill_account_payment_with_no_move` so this minor change helped me a lot.


```diff
diff --git a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
index 7c0aa3b5759..81b3582d8ca 100644
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -541,21 +541,28 @@ def fill_account_payment_with_no_move(env):
             [("deprecated", "=", True), ("company_id", "=", payment.company_id.id)]
         )
         deprecated_accounts.deprecated = False
-        payment._synchronize_to_moves(
-            [
-                "date",
-                "amount",
-                "payment_type",
-                "partner_type",
-                "payment_reference",
-                "is_internal_transfer",
-                "currency_id",
-                "partner_id",
-                "destination_account_id",
-                "partner_bank_id",
-                "journal_id",
-            ]
-        )
+        import logging
+
+        try:
+            payment._synchronize_to_moves(
+                [
+                    "date",
+                    "amount",
+                    "payment_type",
+                    "partner_type",
+                    "payment_reference",
+                    "is_internal_transfer",
+                    "currency_id",
+                    "partner_id",
+                    "destination_account_id",
+                    "partner_bank_id",
+                    "journal_id",
+                ]
+            )
+        except Exception as e:
+            logging.error("Faield for payment with ID: %s", payment.id)
+            raise
+
         deprecated_accounts.deprecated = True
         to_write = {
             "line_ids": [
```

I guess that it is not something you want to include in the migration script, but I post it as it might help someone else.

In coming days, I will be running more functional tests and will inform if issues are being faced.

pedrobaeza:
That logging can be helpful indeed and it's acceptable to be included, while raising again the exception and not silently cover any error.

jcdrubay (approved):
It worked well when testing with one database. Thank you.

AntoniRomera (approved):

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
alexis-via: I confirm the error:
account: error in migration script /home/alexis/new_boite/dev/openupgrade4/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py: You cannot add/modify entries prior to and inclusive of the lock date 31/12/2020
In the backtrace, the first line for openupgrade_script is:
```
File "/home/alexis/new_boite/dev/openupgrade4/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py", line 544, in fill_account_payment_with_no_move
    payment._synchronize_to_moves(
```

MiquelRForgeFlow: Ok, I understand now, you are commenting the wrong line, so I got confused. `_synchronize_to_moves` is called two times, one for statement lines in `fill_statement_lines_with_no_move` and the other for payments in `fill_account_payment_with_no_move`. You are commenting in `fill_statement_lines_with_no_move` but you are having the error in `fill_account_payment_with_no_move`.

jcdrubay:
In version 13 and version 14, Odoo handles differently the values for the fields amount_currency and currency_id on the account.move.line.

When using company currency:
- Odoo 13 will keep both fields empty
- Odoo 14 will set both fields

After forcing the display of the 2 fields in the views ... on the left Odoo 13 and on the right Odoo 14

image

After running OpenUpgrade, the the field currency_id was set, but not the amount_currency.

 I guess that openupgrade should update the data to match Odoo 14 logic.

```sql 
update 
    account_move_line 
set 
    amount_currency=debit-credit
where 
    currency_id is not null and
    debit + credit > 0 and
    amount_currency = 0
 ```

**Comment by MiquelRForgeFlow** [comment]
@jcdrubay thanks, query added in post-migration

File: openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py (hunk 9 above)
Comment (MiquelRForgeFlow) [reply openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py]: I was able to reproduce the error of _check_fiscalyear_lock_date with a client database, so it should be a common issue. So I added query to avoid having moves without date.

File: openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py [hunk 14]
```diff
@@ -0,0 +1,646 @@
...
+            "account.sequence_payment_transfer",
+        ],
+    )
+
+
+def fill_account_move_amounts(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move_line
```
jcdrubay: It seems that this is not always a good solution.

Entries related to difference of exchange rates seems to have debit+credit > 0, but amount_currency should stay at 0.

Maybe this should be applicable only for move_type != entry.

I will investigate more.

jcdrubay: Also, in some cases, it could be very wrong if the currency_id is not the company currency.

MiquelRForgeFlow: I just added the `move_type != entry`.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,763 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move_line aml
+        SET amount_currency = aml.debit-aml.credit
+        FROM account_move am
+        WHERE aml.currency_id IS NOT NULL AND
+            aml.move_id = am.id AND am.move_type != 'entry' AND
+            aml.debit + aml.credit > 0 AND (
+                aml.amount_currency = 0 OR aml.amount_currency IS NULL)""",
```
jcdrubay, line 723: After testing, this did not work well as entries in Foreign currencies with amount_currency=0 where due to difference of exchange rate and therefore should stay at 0.

In my case, it seems that I only had to fix the account_move_lines which were in the company currency.

I will test with this:

```sql
UPDATE account_move_line aml
SET amount_currency = aml.debit-aml.credit
FROM res_company rco
WHERE
	aml.currency_id is not null AND
	(aml.amount_currency=0 or aml.amount_currency is NUll) AND
	aml.balance != 0  AND
	aml.currency_id = rco.currency_id
```

I will let you know about the results of my test later, but I wanted to let you know now that the current solution is not a good one.

jcdrubay:
@MiquelRForgeFlow FYI, I had to do the following changes to make the scrip run:

```diff
diff --git a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
index 8599c2a0afb..03f25747e4e 100644
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -434,7 +434,7 @@ def fill_statement_lines_with_no_move(env):
         for rc_id, rc_period_date, rc_fy_date in env.cr.fetchall():
             company_dates[rc_id] = [rc_period_date, rc_fy_date]
             temporal_date = (stl_dates_by_company[rc_id] - timedelta(days=1)).strftime(
-                "%y-%m-%d"
+                "%Y-%m-%d"
             )
             # use temporal_date to avoid _check_fiscalyear_lock_date
             env.cr.execute(
@@ -500,14 +500,14 @@ def fill_statement_lines_with_no_move(env):
             UPDATE res_company
             SET period_lock_date = {}, fiscalyear_lock_date = {}
             WHERE id = {}""".format(
-                "TO_DATE("
-                + company_dates[rc_id][0].strftime("%y-%m-%d")
-                + ", 'YYYY-MM-DD')"
+                "TO_DATE('"
+                + company_dates[rc_id][0].strftime("%Y-%m-%d")
+                + "', 'YYYY-MM-DD')"
                 if company_dates[rc_id][0]
                 else "NULL",
-                "TO_DATE("
-                + company_dates[rc_id][1].strftime("%y-%m-%d")
-                + ", 'YYYY-MM-DD')"
+                "TO_DATE('"
+                + company_dates[rc_id][1].strftime("%Y-%m-%d")
+                + "', 'YYYY-MM-DD')"
                 if company_dates[rc_id][1]
                 else "NULL",
                 rc_id,
@@ -606,7 +606,7 @@ def fill_account_payment_with_no_move(env):
         for rc_id, rc_period_date, rc_fy_date in env.cr.fetchall():
             company_dates[rc_id] = [rc_period_date, rc_fy_date]
             temporal_date = (p_dates_by_company[rc_id] - timedelta(days=1)).strftime(
-                "%y-%m-%d"
+                "%Y-%m-%d"
             )
             # use temporal_date to avoid _check_fiscalyear_lock_date
             env.cr.execute(
@@ -678,14 +678,14 @@ def fill_account_payment_with_no_move(env):
             UPDATE res_company
             SET period_lock_date = {}, fiscalyear_lock_date = {}
             WHERE id = {}""".format(
-                "TO_DATE("
-                + company_dates[rc_id][0].strftime("%y-%m-%d")
-                + ", 'YYYY-MM-DD')"
+                "TO_DATE('"
+                + company_dates[rc_id][0].strftime("%Y-%m-%d")
+                + "', 'YYYY-MM-DD')"
                 if company_dates[rc_id][0]
                 else "NULL",
-                "TO_DATE("
-                + company_dates[rc_id][1].strftime("%y-%m-%d")
-                + ", 'YYYY-MM-DD')"
+                "TO_DATE('"
+                + company_dates[rc_id][1].strftime("%Y-%m-%d")
+                + "', 'YYYY-MM-DD')"
                 if company_dates[rc_id][1]
                 else "NULL",
                 rc_id,
@@ -715,11 +715,12 @@ def fill_account_move_line_amounts(env):
         """
         UPDATE account_move_line aml
         SET amount_currency = aml.debit-aml.credit
-        FROM account_move am
-        WHERE aml.currency_id IS NOT NULL AND
-            aml.move_id = am.id AND am.move_type != 'entry' AND
-            aml.debit + aml.credit > 0 AND (
-                aml.amount_currency = 0 OR aml.amount_currency IS NULL)""",
+        FROM res_company rco
+        WHERE
+               aml.currency_id is not null AND
+               (aml.amount_currency=0 or aml.amount_currency is NUll) AND
+               aml.balance != 0  AND
+               aml.currency_id = rco.currency_id""",
     )
```

MiquelRForgeFlow:
Thanks @jcdrubay for your tips. With that last commit, now nobody will have the _check_fiscalyear_lock_date error :)

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
MiquelRForgeFlow: I made a second version of this query, could you check?

jcdrubay: I think that your new query works, but you don't need anymore the table `account_move am`

jcdrubay:
Everything has been running smoothly... I will check the data and let you know.

jcdrubay:
It might not be perfect, but I could not detect any issue with a small database doing only invoicing with a single service company. 

That database has been upgraded with OpenUpgrade since version 10 !!! 💪 

Thanks you @MiquelRForgeFlow for all your work on the upgrade from 13 to 14 :D

Let me know if I can help for the version 15 ;)

flotho (approved):
Successfully tested on 2 production DB.
Great job,
thanks

flotho (changes requested):
Hi,
The line content of the moves have been truncated.
we lost parts of the account.invoice.line.name

MiquelRForgeFlow:
> we lost parts of the account.invoice.line.name

that would be a v13 issue AFAIK 🤔

pedrobaeza:
And @flotho please always give more information and make sure the issue is due to OpenUpgrade itself, and to put the claim in the proper place.

flotho:
sorry, Ive tracked this there, I'll check this tommorow

flotho:
> > we lost parts of the account.invoice.line.name
> 
> that would be a v13 issue AFAIK thinking

impossible to reproduce from my side on v13, I'l try on v14

pedrobaeza:
General review done with not too much changes. Good work so far. Merging for now for not blocking the rest of the scripts, and further discoverings in real DBs will be put as new pull requests.

Thanks to all the contributors and reviewers:

/ocabot merge nobump

PR #2744, opened 2021-05-28
[14.0][ADD] github CI

MiquelRForgeFlow:
Let's see.

--


pedrobaeza:
@StefanRijnhart maybe you can provide a proper 13.0 DB, as it seems there's a problem with the current one and you are used to. Another question: why populating it in French?

StefanRijnhart:
OK, I'll have a look.

MiquelRForgeFlow:
Yeah, it's strange. For reference, I used https://github.com/OCA/OpenUpgrade/blob/13.0/scripts/create_test_db

StefanRijnhart:
OK, just remove dropping the public schema here: https://github.com/OCA/OpenUpgrade/pull/2744/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R43

Postgresql may or may not create the public schema in every new database depending on the version. Likewise (reversely), pg_dump will include the creation of the public schema depending on whether it is created already in a new database. Dropping the public schema before restoring the dump is a quick fix when problems arise here.

However, in this case the database versions are aligned well enough so that dropping the schema before the dump is not necessary and not helpful anymore as it gives this error:

```
pg_restore: error: could not execute query: ERROR:  schema "public" does not exist
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,74 @@
...
+  test:
+    runs-on: ubuntu-latest
+    env:
+      DB: "openupgrade"
+      DB_USERNAME: "odoo"
+      DB_PASSWORD: "odoo"
+      DB_HOST: "localhost"
+      DB_PORT: 5432
+      ODOO_REPO: "odoo/odoo"
+      ODOO: "./odoo-bin"
```
rvalyi: I think the current error "/home/runner/work/_temp/9122709a-260c-40af-971b-85055f6597eb.sh: line 6: ./odoo-bin: No such file or directory" is because after the checkout you are not inside the odoo repo, but just outside of it. A simple way to fix this woulde be to set `ODOO: "odoo/odoo-bin"` instead or alternatively tweak the run command line 74.

rvalyi (dismissed):

.github/workflows/test.yml
MiquelRForgeFlow: tried, but no

rvalyi: my bad sorry. So as you can see from the logs,  2 lines before the sed commands work. It means that at this point of the script we are in this directory: https://github.com/OCA/OpenUpgrade/tree/13.0/odoo/openupgrade and the odoo-bin scripts is 2 levels before:
So what about setting instead
`ODOO: "../../odoo-bin"` ?

MiquelRForgeFlow: tried, but no

StefanRijnhart: It's quite mysterious why it should be any different from https://github.com/OCA/OpenUpgrade/blob/13.0/.github/workflows/test.yml apart from the versions.

MiquelRForgeFlow: feel free to push here is you want to try things :S

rvalyi: May be before line 74 here `.github/workflows/test.yml` we could add these lines:
```
pwd
ls -la
```
to figure out better were we are and what is available to us at this point...

StefanRijnhart: Gotta cook, but will have a look later

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Test OpenUpgrade migration
+
+on:
+  push:
+    branches: ["14.0*"]
```
kos94ok-3D, line 8: Maybe the problem in this?
```suggestion
    branches: ["14.0"]
```

.github/workflows/test.yml
StefanRijnhart: OMG did we all forget that OpenUpgrade 14 is no longer a fork of Odoo.

StefanRijnhart: This works, for this job: https://github.com/OCA/OpenUpgrade/pull/2747/commits/be990059ba6a7d1f0883afe933262f5d5a3f837e. Do you want to continue the effort or shall I?

MiquelRForgeFlow: Cannot you push into this branch? We just need to fix the documentation.

StefanRijnhart: I cannot push to this branch it seems.

pedrobaeza:
Then what is missing? This seems mostly done.

StefanRijnhart:
```
 Warning, treated as error:
autodoc: failed to import module 'openupgrade' from module 'openupgradelib'; the following exception was raised:
No module named 'openerp'
Error: Process completed with exit code 2.
```

This probably means that `odoo/odoo` is not in the Python path when autodoc does its thing.

StefanRijnhart:
@MiquelRForgeFlow please pick https://github.com/StefanRijnhart/OpenUpgrade/commit/9a60efa584f294b3ae241d0a779a80d63d0beabb which fixes the documentation job by adding the `odoo` checkout dir to the Python path. It also streamlines the installation of openupgradelib.

pedrobaeza:
I have included Stefan's changes, so merging this. Thanks all for getting this working.

PR #2745, opened 2021-05-28
[13.0][IMP] product: Empty company_id

pedrobaeza:
On v13, there's no default company associated with the template nor the pricelist, on contrary than on v12. We need to empty the company_id field in case of having only one company for not having problems later when creating new pricelists and not being able to select old products due to the difference on the company_id field (on a pricelist without company, you can only select templates without company, and vice versa, but not a mix of both).

We need to empty the company of pricelists as well.

If there are more than one company in the DB, then everything is preserved as it is.

@Tecnativa TT29732


sergio-teruel (approved):

legalsylvain (approved):
makes sense. thanks for the explanation.

PR #2756, opened 2021-06-06
[14.0] CI: no flake8; yaml line length

StefanRijnhart:
* Remove flake8 github job as it is redundant because we also run the standard pre-commit job
* Honour YAML line length limit, improving readability
* Apply alphabetical order where appropriate
* Align test and documentation jobs


MiquelRForgeFlow (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -1,25 +1,28 @@
...
+# single version of Python. For more information see:
+# https://help.github.com/actions/language-and-framework-guides\
+# /using-python-with-github-actions
 
 name: Test OpenUpgrade migration
 
 on:
   push:
-    branches: ["14.0*"]
+    branches: ["14.0"]
```
pedrobaeza: Doing this change PRs will be executed?

StefanRijnhart: Mmm, I must admit I changed this because it was one of the suggestions on the previous PR. But I see that it was updated in 13.0 related to runbot in https://github.com/OCA/OpenUpgrade/commit/5801882f6e696d20011e11829dc97d8f441f7cc7.
I've put it back.

pedrobaeza: Oh, that was the reason. I remember it was there due to something, but I didn't remember the exact one.

PR #2757, opened 2021-06-07
[11.0][IMP] stock: fix company on quants for customers/supplier location

AaronHForgeFlow:
v11: https://github.com/OCA/OpenUpgrade/blob/11.0/addons/stock/models/stock_quant.py#L31 --> in practice this is null as the company is null for customer locations, at least it should.

v10:  https://github.com/OCA/OpenUpgrade/blob/10.0/addons/stock/models/stock_quant.py#L61 --> in practice this is the company that does the delivery in case of cusotmer locations

cc @MiquelRForgeFlow

### DIFF
--- a/addons/stock/migrations/11.0.1.1/pre-migration.py
+++ b/addons/stock/migrations/11.0.1.1/pre-migration.py
@@ -22,6 +22,25 @@ def delete_quants_for_consumable(env):
     )
 
 
+def remove_company_for_quants_in_vendor_customer(env):
+    """On v11, customer or vendor location's quants do not have company
+    (it is not assinged, it is related to the location) and customer
+    location should not have company. In v10 is was a _company_default_get.
+    We should remove the company on those in order to be aligned to the
+    new flow.
+    """
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE stock_quant sq
+        SET company_id = sl.company_id
+        FROM stock_location sl
+        WHERE sl.id = sq.location_id
+            AND sl.usage in ('customer', 'supplier')
+            AND sq.company_id != sl.company_id
+        """
+    )
+
+
 def fix_act_window(env):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not correct at all, as you are assuming the destination location doesn't have an assigned company. You should synchronize the location company with the quant company.

AaronHForgeFlow:
In case the location has a company that code does not change the company in the quant. Both are already synchronized when there is a company in the location.

The only case when both do not match is when the company is empty in the location

AaronHForgeFlow:
@pedrobaeza  does it look better now?

pedrobaeza:
No. I still insists you can unconditionally empty the company. You should just empty it if they customer location has empty company.

AaronHForgeFlow:
Ok, I included that condition, I think this is safe to be included in OU now.

addons/stock/migrations/11.0.1.1/pre-migration.py
```diff
@@ -22,6 +22,26 @@ def delete_quants_for_consumable(env):
...
+    """
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE stock_quant sq
+        SET company_id = sl.company_id
+        FROM stock_location sl
+        WHERE sl.id = sq.location_id
+            AND sl.usage in ('customer', 'supplier')
+            AND sq.company_id IS NOT NULL
+            AND sl.company_id IS NULL
```
pedrobaeza:
```suggestion
        UPDATE stock_quant sq
        SET company_id = sl.company_id
        FROM stock_location sl
        WHERE sl.id = sq.location_id
            AND sl.usage in ('customer', 'supplier')
            AND sq.company_id != sl.company_id
```

AaronHForgeFlow: I finally got the point :facepalm:

PR #2758, opened 2021-06-08
14.0 fix calendar migration

SilvioGregorini:


MiquelRForgeFlow:
Could you avoid in your second commit do too much visual diff?

SilvioGregorini:
> Could you avoid in your second commit do too much visual diff?

I made it that way so that it could be easier to recognize the column-value association, but if little diffs are preferred, I will change it as soon as I get back from lunch :smile:

SilvioGregorini:
@MiquelRForgeFlow done!

pedrobaeza:
Definitively something is wrong with the CI for not failing on previous SQLs.

MiquelRForgeFlow:
I see. Because doc file is not filled. Some modules like `calendar` and `product` are not marked in doc.

pedrobaeza:
Ouch!

MiquelRForgeFlow:
Now `calendar` it's being tested.

pedrobaeza (approved):

pedrobaeza:
Thanks for your fixes.

PR #2762, opened 2021-06-08
[11.0][OU-FIX] stock: quant merge group by company

AaronHForgeFlow:
I have a internal location without company... and several companies have quants inside that location... I am not crazy, I swear...

@MiquelRForgeFlow


pedrobaeza:
#2757 should fix that, or do the other query more general not only restricted to customer locations, but this one is not a valid one.

AaronHForgeFlow:
That  was not harmful but I understand the problem is in data consistency on the database so this is not necessary, I just had to try :)

PR #2763, opened 2021-06-08
[14.0][FIX] utm: use env instead of cr in delete xmlids

MiquelRForgeFlow:
hehehe


pedrobaeza:
And why again no error in CI ?

MiquelRForgeFlow:
Because `delete_records_safely_by_xml_id` method uses `raise_if_not_found=False` parameter, so any error in that method is logged with INFO instead of WARNING/ERROR.

PR #2764, opened 2021-06-08
[13.0][MIG] delivery_hs_code

kos94ok-3D:
Nothing to do


pedrobaeza (approved):
It's not "Nothing to do", but done on https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/openupgrade_records/lib/apriori.py#L31

PR #2765, opened 2021-06-09
[14.0][RFR] Github Pages: configuration changes

StefanRijnhart:
This should be everything that is needed to get an automatically updating set of docs on https://oca.github.io/OpenUpgrade. The version there now is served from my test branch https://github.com/oca/openupgrade/tree/imp/14.0%2Fgithub_pages.

Github pages only allows to serve documentation from the branch root or from a `docs` subdirectory. This is why I renamed `docs/source` to `docsource`.

To have the HTML files rebuild automatically I've tried some random commit action that I found lying on the Github marketplace (https://github.com/marketplace/actions/add-commit) and it seems to work. Here is an example commit: https://github.com/OCA/OpenUpgrade/commit/27b57a9208a51969b03411a65d64a531bbf245fd. I've added the configuration yml to this PR. It's basically the existing documentation job plus the following stanza:
```
      - name: Commit changes
        uses: EndBug/add-and-commit@v7
        with:
          add: "docs"
          cwd: "openupgrade"
          default_author: github_actions
          message: "[UPD] HTML documentation"
```
I've enabled it for pushes to 14.0 only (so it's not run for pull requests).

After merge, update https://github.com/OCA/OpenUpgrade/settings/pages to use this branch, and delete the temporary branch.


legalsylvain:
> To get this working we need to merge this change, and then get some gitbot to rebuild the documentation on every push to the main branch, and commit the built documentation to the branch.

Not sure to understand. AFAIK, gitbot should rebuild documentation on each push to **any** branches. if a V11 PR is merged, the coverage of the V11 serie will be updated, so the fetch_coverage should be launched on master branch. Not sure to understand how it can be handled.

Apart that point, :+1: thanks for that move !

StefanRijnhart:
It's probably not terrible if the older coverage in the docs lags behind until the next commit on the main branch.

legalsylvain:
> It's probably not terrible if the older coverage in the docs lags behind until the next commit on the main branch.

OK. let's say that it will always be better than it is now!

StefanRijnhart:
I've tried some random commit action that I found lying on the Github marketplace (https://github.com/marketplace/actions/add-commit) and it seems to work. Here is an example commit: https://github.com/OCA/OpenUpgrade/commit/27b57a9208a51969b03411a65d64a531bbf245fd. I've added the configuration yml to this PR. It's basically the existing documentation job plus
```
      - name: Commit changes
        uses: EndBug/add-and-commit@v7
        with:
          add: "docs"
          cwd: "openupgrade"
          default_author: github_actions
          message: "[UPD] HTML documentation"
```
I've enabled it for pushes to 14.0 only (so it's not run for pull requests).

pedrobaeza:
Do we need to have in this branch all modulesXXX-YYY.rst old files?

StefanRijnhart:
If we remove them, the docs will fail to build without fetching them from Github. But I guess we now have a clearly defined way of building the docs, and with the automatically updated online version there is less need for local builds. Deleted them in a separate commit.

pedrobaeza:
Is this finished then for merging?

GSLabIt:
did u have a look at copycat action?

StefanRijnhart:
@pedrobaeza as far as I am concerned it is ready for merging. We'll have to see how the add-and-commit action holds up in practice but so far it's looking good.

@GSLabIt I don't know the copycat action. What problem would it solve here?

GSLabIt:
> @pedrobaeza as far as I am concerned it is ready for merging. We'll have to see how the add-and-commit action holds up in practice but so far it's looking good.
> 
> @GSLabIt I don't know the copycat action. What problem would it solve here?

it easier to copy github pages, as far as I know.. have a look at:

https://github.com/marketplace/actions/copycat-action

pedrobaeza:
Then let's go with it although further improvements can come.

StefanRijnhart:
It worked! https://github.com/OCA/OpenUpgrade/commit/c03dfbdb7733d71dae94b93e5e410e19de34850d
Now switching the Github pages source branch to 14.0.

pedrobaeza:
I have adapted the README to be correct with latest things:

https://github.com/OCA/OpenUpgrade/blob/14.0/README.md

MiquelRForgeFlow:
@StefanRijnhart something is wrong with your first commit, now only base is tested, but not the other modules.

MiquelRForgeFlow:
Solved in https://github.com/OCA/OpenUpgrade/pull/2787.

PR #2770, opened 2021-06-15
[13.0][FIX] sale_mrp: fill bom_line_id in moves

MiquelRForgeFlow:
Oversight, we have to consider sale order lines, see https://github.com/OCA/OpenUpgrade/blob/13.0/addons/sale_mrp/models/sale_mrp.py#L21.


addons/sale_mrp/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2021 ForgeFlow <https://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def fill_stock_move_bom_line_id(env):
+    """v13 added a new functionality to this bom_line_id field.
+    A part from the usual value for raw moves in mrp productions,
```
pedrobaeza:
```suggestion
    Apart from the usual value for raw moves in mrp productions,
```

pedrobaeza (commented):
Is the same SQL?

JordiMForgeFlow (approved):
Approved and tested 👍🏼

MiquelRForgeFlow:
> Is the same SQL?

No, you can see it's not exactly the same. Now it has been corrected and tested. I know it looks a bit convoluted.

MiquelRForgeFlow:
now it's ready. we considered that bom has to be active, and also prioritize if the bom has company

PR #2771, opened 2021-06-15
[9.0][IMP] migrate bic numbers

astirpe:
When migrating bank accounts that are not linked to an existing bank, but do include a Bank Name and/or Bank Identifier Code (BIC):

- if a bank with a matching BIC exists, link the bank account to that bank;

- if no bank exists with a matching BIC, automatically create a new bank, with the new BIC;

- -  if the bank account details do not specify the bank name, use the BIC as bank name.

NOTE In case of inconsistencies to be fixed manually: some bank accounts have BIC different than the linked bank. To find them:

```SQL
	select rpb.bank_bic, rb.bic, rb.name, rb.id
	from res_partner_bank rpb
	  join res_bank rb on rpb.bank = rb.id
	where rpb.bank_bic <> rb.bic
```

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -370,6 +370,38 @@ def set_date_maturity(env):
     )
 
 
+def migrate_bic_numbers(env):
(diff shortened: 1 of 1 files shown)


astirpe:
Sorry wrong PR

PR #2772, opened 2021-06-16
[9.0][IMP] migrate bic numbers

astirpe:
When migrating bank accounts that are not linked to an existing bank, but do include a Bank Name and/or Bank Identifier Code (BIC):

- if a bank with a matching BIC exists, link the bank account to that bank;

- if no bank exists with a matching BIC, automatically create a new bank, with the new BIC; use the BIC as bank name.

NOTE In case of inconsistencies to be fixed manually: some bank accounts have BIC different than the linked bank. To find them:

```SQL
	select rpb.bank_bic, rb.bic, rb.name, rb.id
	from res_partner_bank rpb
	  join res_bank rb on rpb.bank = rb.id
	where rpb.bank_bic <> rb.bic
```

### DIFF
--- a/addons/account/migrations/9.0.1.1/pre-migration.py
+++ b/addons/account/migrations/9.0.1.1/pre-migration.py
@@ -370,6 +370,41 @@ def set_date_maturity(env):
     )
 
 
+def migrate_bic_numbers(env):
+    """ BIC numbers migrated in supplier bank details """
+    # if no bank exists with a matching BIC, automatically create a new bank, with the new BIC;
+    # if the bank account details do not specify the bank name, use the BIC as bank name.
+    openupgrade.logged_query(
+        env.cr, """
+        WITH banks AS (
+            SELECT ARRAY_AGG(bank_name order by bank_name is null) AS names, bank_bic
+            FROM res_partner_bank
+            WHERE bank_id IS NULL AND bank_bic IS NOT NULL AND bank_bic NOT IN (
+                SELECT bic
+                FROM res_bank
+                WHERE bic IS NOT NULL
+            )
+            GROUP BY bank_bic
+        )
+        INSERT INTO res_bank(name, bic)
+        SELECT COALESCE(names[1], bank_bic), bank_bic FROM banks
+        """
+    )
+    # if a bank object with a matching BIC exists, link the bank account to that bank
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE res_partner_bank rpb
+        SET bank_id = subquery.id
+        FROM (
+            SELECT rb.id, rb.bic
+            FROM res_bank rb
+            WHERE rb.bic IS NOT NULL
+        ) AS subquery
+        WHERE rpb.bank_id IS NULL AND rpb.bank_bic = subquery.bic
+        """
+    )
+
+
 def fast_create(env, settings):
     for setting in settings:
         (table_name, field_name, sql_type, sql_request) = setting
@@ -421,6 +456,7 @@ def migrate(env, version):
     merge_supplier_invoice_refs(env)
     openupgrade.rename_fields(env, field_renames)
     set_date_maturity(env)
+    migrate_bic_numbers(env)
 
     # Fast Create new fields
     fast_create(env, FAST_CREATIONS)


astirpe:
CC @StefanRijnhart

addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -370,6 +370,38 @@ def set_date_maturity(env):
...
+        env.cr, """
+        INSERT INTO res_bank(name, bic)
+        SELECT CASE WHEN bank_name IS NOT NULL THEN bank_name ELSE bank_bic END, bank_bic
+        FROM res_partner_bank
+        WHERE bank IS NULL AND bank_bic IS NOT NULL AND bank_bic NOT IN (
+            SELECT bic
+            FROM res_bank
+            WHERE bic IS NOT NULL
+        )
+        GROUP BY bank_name, bank_bic
```
StefanRijnhart: Even if the banks have a different name, I would advice against introducing multiple banks with the same BIC

StefanRijnhart:
```
2021-06-16 09:04:16,215 22953 INFO openupgrade openerp.sql_db: Programming error: column "bank" does not exist
LINE 5:         WHERE bank IS NULL AND bank_bic IS NOT NULL AND bank...
```
This is due to `bank` being renamed to `bank_id` in the pre-migration script.

StefanRijnhart:
If you want to use as many proper bank names as possible you can use something like
```
WITH banks AS (
    SELECT ARRAY_AGG(bank_name order by bank_name is null) AS names, bank_bic
    FROM res_partner_bank
	WHERE bank_id IS NULL AND bank_bic IS NOT NULL AND bank_bic NOT IN (
            SELECT bic
            FROM res_bank
            WHERE bic IS NOT NULL
	)
        GROUP BY bank_bic)
INSERT INTO res_bank(name, bic)
SELECT COALESCE(names[1], bank_bic), bank_bic FROM banks;
```

StefanRijnhart (approved):

StefanRijnhart:
Oh, please squash commits.

addons/account/migrations/9.0.1.1/pre-migration.py
```diff
@@ -370,6 +370,41 @@ def set_date_maturity(env):
...
 
 
+def migrate_bic_numbers(env):
+    """ BIC numbers migrated in supplier bank details """
+    # if no bank exists with a matching BIC, automatically create a new bank, with the new BIC;
+    # if the bank account details do not specify the bank name, use the BIC as bank name.
+    openupgrade.logged_query(
+        env.cr, """
+        WITH banks AS (
+            SELECT ARRAY_AGG(bank_name order by bank_name is null) AS names, bank_bic
```
MiquelRForgeFlow, line 380:
```suggestion
            SELECT ARRAY_AGG(bank_name order by bank_name) AS names, bank_bic
```

StefanRijnhart: Your suggestion would give precedence to NULL values instead of proper bank names.

StefanRijnhart: This is the SQL ordering + NULL value brainf*ck. By ordering on IS NULL, we get non-null values first because IS NULL will return FALSE for non-null values and FALSE comes before TRUE in the default ascending order.

MiquelRForgeFlow: Then it should be:

```suggestion
            SELECT ARRAY_AGG(bank_name ORDER BY bank_name NULLS LAST) AS names, bank_bic
```
See https://www.postgresql.org/docs/current/queries-order.html

StefanRijnhart: That would not be wrong, but we don't really care which bank name is selected as long as a non null value is selected when there is one available so the current version is fine really.

MiquelRForgeFlow: Well, the problem is in assure that the intended syntax works. The documentation says about "NULLS LAST" but says nothing about "IS NULL" in the ORDER BY clause. I am not saying your syntax doesn't work, just that I prefer using what the documentation says just to be safe.

StefanRijnhart: We are sorting here by a boolean expression. It is documented here: https://www.postgresql.org/docs/10/queries-order.html
```
The sort expression(s) can be any expression that would be valid in the query's select list.
```

MiquelRForgeFlow: Interesting, good to know. So the difference between `bank_name is null` and `bank_name NULLS LAST` is that the first is nondeterministic, which we don't know which random name would be first, and the second is deterministic, ordering them alphabetically, right?

StefanRijnhart: Correct. But given the bank names that people used to assign in the res.partner.bank form, the result is always going to be random ;-)

MiquelRForgeFlow (approved):

PR #2773, opened 2021-06-16
[11.0][FIX] sale_stock: fill sale_id of procurement groups without moves

MiquelRForgeFlow:
If some procurement groups don't have linked moves, then happens `sale_id` is not filled. We use old procurement order in those cases.


JordiMForgeFlow (approved):
Tested 👍🏼

AaronHForgeFlow (approved):

StefanRijnhart (approved):

PR #2779, opened 2021-06-21
[13.0][OU-IMP] account: Recompute totals for regular entries

pedrobaeza:
Regular journal entries inform as well the totals, although not being invoices, so we need to recompute them, using the equivalent SQL.

@Tecnativa TT30528


MiquelRForgeFlow:
LGTM :+1:  (for some reason, button to review doesn't appear to me)

MiquelRForgeFlow (approved):

PR #2784, opened 2021-06-23
[DOC][FIX]Bad directory reference

flotho:


flotho:
ping @StefanRijnhart

docsource/development.rst
```diff
@@ -52,7 +52,7 @@ Learn from existing migration scrips
 
 
 Since version 14, the migration scripts are located in:
-``OpenUpgrade/openupgrade-framework/openupgrade_scripts/scripts/<module-name>``
+``OpenUpgrade/openupgrade_scripts/scripts/<module-name>``
```
StefanRijnhart: It probably makes sense to drop the `OpenUpgrade` part here, analogous to the mention above.

flotho:
better @StefanRijnhart  ?

StefanRijnhart (approved):
Thanks!

docsource/development.rst
```diff
@@ -16,7 +16,7 @@ steps are not necessary):
 - end-migration.py
 
 Since version 14, these files can be found here:
-``openupgrade-framework/openupgrade_scripts/scripts/<module-name>``
+``openupgrade_scripts/scripts/<module-name>``
```
pedrobaeza, line 20: Now that you are here, it seems the initial intention was to put this into a new paragraph, but for proper RST syntax for that, you need to leave an extra empty line (the same for both changes).

pedrobaeza (approved):

PR #2797, opened 2021-07-08
[14.0][FIX] documentation: Prefix merged/renamed modules with `Done`

pedrobaeza:
So that the CI can detect them and check if there's something incorrect.

The paragraph about OCA modules has been removed as well, as that one is not maintained at all, and it's confusing then.

Let's see if this doesn't provoke a failed CI due to dependencies not yet "migrated".

@Tecnativa


pedrobaeza:
As expected, this fails as it installs all the dependencies chain. We will have to progressively switch the comments according the evolution of the migration scripts. I keep this opened meanwhile as reminder.

MiquelRForgeFlow:
I don't like the deletion of the OCA modules section, as it might be used someday again. It was used to link an obsolete odoo module converted into OCA module.

pedrobaeza:
@MiquelRForgeFlow that's not true. See the `account_bank_statement_import` one for example, that is documented inline in the Odoo modules table. It was used marginally for OCA modules that were renamed, but nobody cares/maintains/checks this.

StefanRijnhart:
Are all of these covered by migration scripts if necessary?

If the OCA section is unused, it can go as far as I am concerned.

StefanRijnhart:
Can you rebase to see what the failed Github workflow does this time? Or trigger a rerun in some other way.

MiquelRForgeFlow:
We cannot merge it as it depends of other modules like `account`.

StefanRijnhart:
So that's the point, the modules should not be set to done unless they and all their dependencies are covered by migration scripts.

MiquelRForgeFlow:
Due to conflicts, rebased in https://github.com/OCA/OpenUpgrade/pull/3523.

PR #2804, opened 2021-07-09
[14.0][Add] digest: Migration Script.

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2783.


openupgrade_scripts/scripts/digest/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2021 Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "digest", "14.0.1.1/noupdate_changes.xml")
```
pedrobaeza, line 8: Seeing the noupdate data, I don't think we should change the ones customized by the customer.

openupgrade_scripts/scripts/digest/14.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,27 @@
+---Models in module 'digest'---
+---Fields in module 'digest'---
+digest       / digest.digest            / periodicity (selection)       : selection_keys is now '['daily', 'monthly', 'quarterly', 'weekly']' ('['monthly', 'quarterly', 'weekly']')
```
pedrobaeza, line 3: Remember to be verbose in this analysis file:
```suggestion
digest       / digest.digest            / periodicity (selection)       : selection_keys is now '['daily', 'monthly', 'quarterly', 'weekly']' ('['monthly', 'quarterly', 'weekly']')
# NOTHING TO DO: New daily option available, but no need to change existing ones.
```

pedrobaeza (changes requested):

openupgrade_scripts/scripts/digest/14.0.1.1/post-migration.py
MiquelRForgeFlow: ok, I commented the code in the noupdate_changes.xml file

pedrobaeza (approved):

PR #2819, opened 2021-07-12
[14.0][MIG] survey

MiquelRForgeFlow:
Migration done.


openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
```diff
@@ -0,0 +1,234 @@
...
+        FROM survey_survey ss
+        WHERE sui.survey_id = ss.id AND ss.session_state IN ('ready', 'in_progress')""",
+    )
+    # nickname
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE survey_user_input sui
+        SET nickname = rp.name
+        FROM survey_survey ss, res_partner rp
```
StefanRijnhart: Looks like an unused join on survey_survey, or am I missing something?

openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
```diff
@@ -0,0 +1,234 @@
...
+        WHERE sui.survey_id = ss.id AND sui.partner_id = rp.id""",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE survey_user_input sui
+        SET nickname = sui.email
+        FROM survey_survey ss
+        WHERE sui.survey_id = ss.id AND sui.nickname IS NULL""",
```
StefanRijnhart: Same question about the join on survey survey here.

StefanRijnhart (approved):
Great work. I only had a small remark without technical impact.

openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
MiquelRForgeFlow: survey_survey is the FROM, what do you mean?

openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
MiquelRForgeFlow: unused? ss is used in the next line :S

MiquelRForgeFlow: Yeah, deleted the unused table.

openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
MiquelRForgeFlow: deleted unused table

openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
```diff
@@ -0,0 +1,242 @@
...
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, _column_copies)
+    openupgrade.rename_fields(env, _field_renames)
+    openupgrade.rename_models(env.cr, _model_renames)
+    openupgrade.rename_tables(env.cr, _table_renames)
+    openupgrade.rename_xmlids(env.cr, _xmlid_renames)
```
davidtranhp, line 247: Should the new field `session_code` be filled here?

This field has default `_get_default_session_code()` but it would generate the same codes for all rows and break the _sql_constraints `('session_code_unique', 'unique(session_code)', 'Session code should be unique'),`

Draft idea is something like this
```python
def _fill_survey_session_code(env):
    openupgrade.logged_query(
        env.cr,
        "ALTER TABLE survey_survey ADD COLUMN session_code character varying;"
        )
    env.cr.execute(
        env.cr,
        "SELECT id FROM survey_survey ORDER BY id ASC;"
        )
    rows = env.cr.dictfetchall()
    start_code = 1000
    for row in rows:
        openupgrade.logged_query(
            env.cr,
            """
            UPDATE survey_survey SET session_code = %s WHERE id=%s;
            """,
            (start_code, row['id'])
            )
        start_code += 1
```

davidtranhp (changes requested):

openupgrade_scripts/scripts/survey/14.0.3.2/pre-migration.py
MiquelRForgeFlow: filled

PR #2834, opened 2021-07-15
[13.0][OU-ADD] hr_maintenance: Nothing to do

pedrobaeza:
@Tecnativa TT27170


pedrobaeza:
We can fast-track as nothing to do.

PR #2843, opened 2021-07-19
[14.0][OU-ADD] sale: Migration scripts

MiquelRForgeFlow:
Migration done.


flotho (approved):
Successfully tested on 2 production DB migrated from 10.0

pedrobaeza (approved):
Adjusted noupdate data due to the renaming + optimize to not write some non modified record rules.

pedrobaeza:
> Successfully tested on 2 production DB migrated from 10.0

If you have such DBs already in production, you would need to adjust manually the record rules for accessing account.move and account.move.line due to the renaming of `type` to `move_type`.

PR #2847, opened 2021-07-20
[14.0][UPD] apriori + analysis

MiquelRForgeFlow:
New analysis due to apriori update.


openupgrade_scripts/apriori.py
```diff
@@ -6,6 +6,7 @@
 renamed_modules = {
     # odoo
     "account_facturx": "account_edi_facturx",
+    "sale_coupon": "coupon",
```
legalsylvain, line 9: Well. If I imagine the following use case, i'm not sure it's correct : 

- V13 : ``sale_coupon`` installed.

- openupgraded DB with that code in V14 : ``coupon`` : installed ; ``sale_coupon`` not installed. (it is not autoinstallable)

That is not the expected result. don't you think ?

legalsylvain (commented):
One blocking point. The rest LGTM for me.

thanks !

openupgrade_scripts/apriori.py
MiquelRForgeFlow: That's why in the pre-migration of `coupon`, I added script to install `sale_coupon`.

legalsylvain: Ow. Sorry, I didn't saw that part. 

:+1:

legalsylvain (approved):

pedrobaeza:
I don't think we should do such thing, as the same parts goes to `coupon` than to `sale_coupon`, but the decision to rename to `coupon` the previous `sale_coupon` makes harder other things.

I prefer that you leave it as is. ORM is going to handle the field/model module allocation. Which other thing is moving you to do this?

MiquelRForgeFlow:
No, the problem is that the models are renamed, and some migration scripts will be needed in `coupon`. If I do not rename the module, then the scripts in `coupon` will not be executed. It's the same case that happened in v13 with crm_reveal and crm_iap_lead, which in crm_iap_lead I had to install crm_iap_lead_website.

That's why we should do the rename as it is now.

docsource/modules130-140.rst
```diff
@@ -104,7 +104,7 @@ Module coverage 13.0 -> 14.0
 +--------------------------------------------+-------------------------------------------------+
 |contacts                                    | Nothing to do                                   |
 +--------------------------------------------+-------------------------------------------------+
-| |new| coupon                               |                                                 |
+| |new| coupon                               | Split from sale_coupon                          |
```
pedrobaeza: We should prefix it with `Done` or `Nothing to do` to be handled by CI.

openupgrade_scripts/scripts/coupon/14.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+    WHERE name = {} AND state='uninstalled'
+    """.format(
+        tuple(to_install),
+    )
+    openupgrade.logged_query(cr, sql)
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    install_new_modules(env.cr)
```
pedrobaeza, line 23: No model renames here?

openupgrade_scripts/scripts/sale_coupon/14.0.1.0/upgrade_analysis.txt
```diff
@@ -1,77 +1,4 @@
 ---Models in module 'sale_coupon'---
-obsolete model report.sale_coupon.report_coupon [abstract]
-obsolete model sale.coupon
-obsolete model sale.coupon.generate [transient]
-obsolete model sale.coupon.program
-obsolete model sale.coupon.reward
-obsolete model sale.coupon.rule
+model sale.coupon.apply.code (moved from sale_coupon) [transient]
```
pedrobaeza, line 2: Weird analysis file...

docsource/modules130-140.rst
MiquelRForgeFlow: But that's when the full script will be added. Not now.

openupgrade_scripts/scripts/coupon/14.0.1.0/pre-migration.py
MiquelRForgeFlow: I am not doing the migration of coupon now, just updating the analysis.

openupgrade_scripts/scripts/sale_coupon/14.0.1.0/upgrade_analysis.txt
MiquelRForgeFlow: yeah haha 😂

docsource/modules130-140.rst
MiquelRForgeFlow: Well, I removed that annotation to avoid confusions.

PR #2855, opened 2021-07-22
[14.0][MIG] event_crm

MiquelRForgeFlow:
Nothing to do.


StefanRijnhart (approved):

RaulOForgeFlow (approved):

StefanRijnhart:
I think these nothing-to-do PRs can be best merged manually without ocabot triggering another run from the github workflows, so I'll do just that, even if it feels a bit dirty.

pedrobaeza:
But the dependency chain has been respected?

StefanRijnhart:
I would think yes, because otherwise, CI would not have been green on the PR.

pedrobaeza:
I see that both `event` and `crm` has been merged, but CI is not necessarily saying the truth, as if there are no migration scripts, but there's nothing that provokes an error, CI will be green.

StefanRijnhart:
OK. I did check for the 'blocked by dependency' label before merging but I'll do an extra check when merging manually next time.

pedrobaeza:
I think @MiquelRForgeFlow is very conscientious with the labels, so it can be OK if it has been removed. Sorry for the noise.

StefanRijnhart:
That's true, @MiquelRForgeFlow is very good with those labels!

PR #2888, opened 2021-08-06
[13.0][FIX] account: Swap reference/name assignations depending of type

MiquelRForgeFlow:
In invoices '`in_*`', the field 'ref' always is used. Only in bills '`out_*`', we have to be picky about 'invoice_payment_ref' or 'ref'.

Further refining of https://github.com/OCA/OpenUpgrade/commit/a4234f892ecee97fa307e580caa56e6e027f5e6a.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -146,7 +146,11 @@ def migration_invoice_moves(env):
         write_date) = (
         ai.message_main_attachment_id, ai.access_token,
         COALESCE(ai.number, ai.move_name, am.name), COALESCE(ai.date, am.date),
-        ai.name, ai.comment, ai.type, ai.journal_id, ai.company_id,
+        CASE WHEN ai.type IN ('in_invoice', 'in_refund') THEN ai.reference ELSE ai.name END,
+        CASE WHEN ai.type = 'in_refund' THEN COALESCE(ai.comment, ai.name)
+            WHEN ai.type = 'out_refund' THEN COALESCE(ai.comment, ai.reference)
+            ELSE ai.comment END,
+        ai.type, ai.journal_id, ai.company_id,
         ai.currency_id, ai.partner_id, ai.commercial_partner_id,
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -146,7 +146,10 @@ def migration_invoice_moves(env):
         write_date) = (
         ai.message_main_attachment_id, ai.access_token,
         COALESCE(ai.number, ai.move_name, am.name), COALESCE(ai.date, am.date),
-        ai.name, ai.comment, ai.type, ai.journal_id, ai.company_id,
+        CASE WHEN ai.type IN ('in_invoice', 'in_refund') THEN ai.reference ELSE ai.name END,
+        CASE WHEN ai.type IN ('in_invoice', 'in_refund') THEN COALESCE(ai.comment, ai.name)
```
pedrobaeza: Why do you need to provide alternative "comments"?

MiquelRForgeFlow: It's not "needed" strictly, but I observed in my BD, in old invoices of type '`in_*`' that the ai.name was used as a comment (for example "Wrong currency" or "Wrong payment mode"). Thus, having this info in the invoice_payment_ref doesn't make much sense, and I thought putting that info in the narration (comment) field makes more sense.

pedrobaeza: I think that case is only in refunds

MiquelRForgeFlow: You are right, I checked it.

MiquelRForgeFlow: What do you propose?

pedrobaeza: Split the case more for each type (`in_refund`, `in_invoice`, `out_refund`, `out_invoice`)

MiquelRForgeFlow: done. can you check again?

pedrobaeza (approved):

PR #2890, opened 2021-08-09
[13.0][MIG] sale_order_variant_mgmt: Migration to v13 (merged on sale_product_matrix)

CarlosRoca13:
cc @Tecnativa TT28623

please @pedrobaeza @Yajo , review this

### DIFF
--- a/addons/sale_product_matrix/migrations/13.0.1.0/post-migration.py
+++ b/addons/sale_product_matrix/migrations/13.0.1.0/post-migration.py
@@ -1,0 +1,28 @@
+# Copyright 2021 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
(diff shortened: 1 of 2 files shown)


addons/sale_product_matrix/migrations/13.0.1.0/post-migration.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2021 Tecnativa - Carlos Roca
```
yajo, line 1: In general terms I think we use the full version as folder name. This file should be in the `13.0.1.0.0` folder.

yajo (approved):

addons/sale_product_matrix/migrations/13.0.1.0/post-migration.py
pedrobaeza: That doesn't work. It should follow the exact manifest version.

pedrobaeza (approved):

PR #2892, opened 2021-08-19
[REM] remove migration script from net_weight to weight

legalsylvain:
the previous migration script was based on bad interpretation of the new unique ``weight`` field.

see https://github.com/OCA/product-attribute/pull/894 for the whole discussion.

CC : @mbcosta, @pedrobaeza

### DIFF
--- a/addons/product/migrations/9.0.1.2/pre-migration.py
+++ b/addons/product/migrations/9.0.1.2/pre-migration.py
@@ -125,11 +125,3 @@ def migrate(env, version):
     cr.execute(
         "update product_pricelist set active=False where type='purchase'"
     )
-
-    # if weight_net != 0.0 in product_template, override weight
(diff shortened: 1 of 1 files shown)


mbcosta:
Hi @legalsylvain 
I think should be better keep the script and just change the field, instead weight_net  use weight.

legalsylvain:
hi @mbcosta not sure to understand. 
In v8 : ``weight`` and ``weight_net``
in V9 : ``weight``

due to a bad documentation, I thought that the ``weight`` field in V9 was the old ``weight_net`` in V8, but if it's not the case, there is nothing to do in my opinion. 
if not agree, could you explain which request to write ?

thanks !

mbcosta:
you right, good work

mbcosta (approved):

PR #2894, opened 2021-08-26
[11.0][OUFIX] Set stock_move.created_purchase_line_id correctly

StefanRijnhart:
The created purchase line needs to be set on the destination move of the procurement, and not on the move that was created for the incoming stock of the purchase order line: https://github.com/odoo/odoo/blob/10.0/addons/purchase/models/purchase.py#L719

Relatedly, also set the created_production_id on the destination move of the procurement.

### DIFF
--- a/addons/mrp/migrations/11.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/11.0.2.0/post-migration.py
@@ -4,6 +4,18 @@ from openupgradelib import openupgrade
 from psycopg2.extensions import AsIs
 
 
+def set_stock_move_created_production_id(cr):
+    """Migrate the created production order to the destination move"""
+    openupgrade.logged_query(
+        cr, """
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

addons/mrp/migrations/11.0.2.0/post-migration.py
```diff
@@ -4,6 +4,18 @@
 from psycopg2.extensions import AsIs
 
 
+def set_stock_move_created_production_id(cr):
+    """Migrate the created production order to the destination move"""
+    openupgrade.logged_query(
+        env.cr, """
```
pedrobaeza:
```suggestion
        cr, """
```

StefanRijnhart: Ah yes thanks!

PR #2895, opened 2021-08-26
[11.0][OUFIX] Set stock_move.created_production_id correctly

StefanRijnhart:
Huge irony here. Follow-up of https://github.com/OCA/OpenUpgrade/pull/2894


pedrobaeza (approved):
OK, I see...

PR #2897, opened 2021-08-31
[13.0][OU-ADD] sale_coupon: Migration scripts

pedrobaeza:
When coming from enterprise v12 module.

@Tecnativa TT27871


MiquelRForgeFlow (approved):
LGTM

(I thought you didn't use enterprise... hehe)

pedrobaeza:
> (I thought you didn't use enterprise... hehe)

As this has been open-sourced, we can make an exception. We have anyways enterprise customers, but limiting at maximum enterprise modules in their instances (not always possible, but it's a pain in such cases).

PR #2903, opened 2021-09-13
[13.0][FIX] mrp: Always define the correct company for the mrp.unbuild model.

victoralmau:
Always define the correct company for the `mrp.unbuild` model.

@Tecnativa TT31747

### DIFF
--- a/addons/mrp/migrations/13.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/13.0.2.0/post-migration.py
@@ -152,7 +152,7 @@ def fill_unbuild_company_id(cr):
         SET company_id = mb.company_id
         FROM mrp_bom mb
         WHERE mu.bom_id = mb.id AND mb.company_id IS NOT NULL
-            AND mu.company_id IS NULL"""
+            AND mu.company_id IS DISTINCT FROM mb.company_id"""
     )
     # from stock_move/res_users
     openupgrade.logged_query(


pedrobaeza:
Please include in the commit message and here the reason to do this: the default is applied via ORM, and remember to put the TT reference also for having it stored in the repository.

StefanRijnhart:
Is it on purpose that this file's line endings are swapped out?

victoralmau:
> Is it on purpose that this file's line endings are swapped out?

I have only changed an `AND` operator to make updating records more fine.
I think it is github that shows the change is very big.

pedrobaeza:
@victoralmau, it's your IDE that has the configuration of overriding the end of line configuration (CF+LF, LF only, etc).

victoralmau:
Changes done.

joao-p-marques:
@pedrobaeza @StefanRijnhart as a note, only a couple of files from this repo have `CRLF` endings, all the others have `LF`. Although this change must be a bug in the IDE, the line ending could probably be standardized...

```
➤ file -k addons/**/*.py | grep CRLF
addons/mrp/migrations/13.0.2.0/post-migration.py:                                                         Python script, ASCII text executable, with CRLF line terminators
addons/mrp/migrations/13.0.2.0/pre-migration.py:                                                          Python script, ASCII text executable, with CRLF line terminators
addons/test_mail/tests/test_message_management.py:                                                        Python script, ASCII text executable, with CRLF line terminators
addons/uom/tests/__init__.py:      
```

pedrobaeza:
> @pedrobaeza @StefanRijnhart as a note, only a couple of files from this repo have `CRLF` endings, all the others have `LF`. Although this change must be a bug in the IDE, the line ending could probably be standardized...

Being honest, I have never worried about the line ending: I let it to the IDE default when creating new files, or copying files and my IDE respecting the existing one, but it's true that the standard good one should be the Unix one for avoiding bad diffs or mis-views in terminals. I think now it's a bit late for changing this I'm afraid.

addons/mrp/migrations/13.0.2.0/post-migration.py
```diff
@@ -152,7 +152,7 @@ def fill_unbuild_company_id(cr):
         SET company_id = mb.company_id
         FROM mrp_bom mb
         WHERE mu.bom_id = mb.id AND mb.company_id IS NOT NULL
-            AND mu.company_id IS NULL"""
+            AND mu.company_id != mb.company_id"""
```
pedrobaeza: Better to use this operator for avoiding problems with NULL values:

```suggestion
            AND mu.company_id IS DISTINCT FROM mb.company_id"""

```

StefanRijnhart:
Thanks for restoring the original line endings! I'm not against correcting them but it should be a separate PR or a separate commit at least, so that the actual change in this PR can be reviewed properly as is the case now.

pedrobaeza (approved):

joao-p-marques:
> > @pedrobaeza @StefanRijnhart as a note, only a couple of files from this repo have `CRLF` endings, all the others have `LF`. Although this change must be a bug in the IDE, the line ending could probably be standardized...
> 
> Being honest, I have never worried about the line ending: I let it to the IDE default when creating new files, or copying files and my IDE respecting the existing one, but it's true that the standard good one should be the Unix one for avoiding bad diffs or mis-views in terminals. I think now it's a bit late for changing this I'm afraid.

Sure. The problem here was that, as there was no `editorconfig` in this repo, it was assuming another global configuration (that had `LF`) so it kept reformatting the file... Maybe, as @StefanRijnhart suggested, the best option would be to, next time a new change is needed in one of these files, push a separate commit just changing the line endings, and another one with the actual changes...

pedrobaeza:
Well, if we do it, better to do it at once, putting such file and reformatting what needed at a whole.

PR #2904, opened 2021-09-13
[FIX] Remove documentation source links

### LINKED ISSUE #2871: Page source link broken in docs
Just go to the bottom of https://oca.github.io/OpenUpgrade/index.html to the "Page source" link:

image

It's broken.

StefanRijnhart:
Fixes #2871

To make the links work, the copy of the sources that Spinx creates in `docs/_sources` would have to be committed into the branch by the Github workflow. Having this second copy of the sources would be confusing, I think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,6 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
-docs/_sources/
 docs/.doctrees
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

pedrobaeza:
I agree on not having the second commit that doesn't add too much value.

PR #2907, opened 2021-09-16
[FIX] account: Check parent in order to fill the data

etobella:
If the data was present on the parent account, we were unable to find it.

@MiquelRForgeFlow @pedrobaeza

### DIFF
--- a/addons/account/migrations/13.0.1.1/end-migration.py
+++ b/addons/account/migrations/13.0.1.1/end-migration.py
@@ -21,7 +21,11 @@ def add_default_account_data(env):
              "default_cash_difference_income_account_id"),
         ]:
             chart = company.chart_template_id
-            account_template = chart[chart_field]
+            check_chart = chart
+            account_template = False
+            while not account_template and check_chart:
+                account_template = check_chart[chart_field]
+                check_chart = check_chart.parent_id
             if not account_template:
                 continue
             tmpl_xml_id = account_template.get_external_id()[account_template.id]


MiquelRForgeFlow (approved):
I detected this bug some time ago, but forgot to make the fix. Thanks.

addons/account/migrations/13.0.1.1/end-migration.py
```diff
@@ -21,7 +21,13 @@ def add_default_account_data(env):
...
         ]:
             chart = company.chart_template_id
-            account_template = chart[chart_field]
+            check_chart = chart
+            account_template = False
+            while check_chart:
+                account_template = check_chart[chart_field]
+                if account_template:
+                    break
+                check_chart = check_chart.parent_id
```
pedrobaeza: Let's improve a bit the code/readability:
```suggestion
            account_template = False
            while not account_template and chart:
                account_template = chart[chart_field]
                chart = chart.parent_id
```

MiquelRForgeFlow: I think your proposal can produce an endless `while`, right? In case the root parent doesn't have a template...

pedrobaeza: No because the `and chart` part.

MiquelRForgeFlow: I think we should stick with the proposal of @etobella because it's strictly correct and super safe despite the readability.

pedrobaeza: If you point to any problem in my proposal, then I'll accept the current one, but I see no drawback.

MiquelRForgeFlow: ok, your proposal convinced me 👍

etobella: Changes applied. Cannot reuse chart, as it is used later, but your idea was kept :wink:

pedrobaeza:
I think you have committed wrong changes.

etobella:
Sorry!!! I think it is fixed now

pedrobaeza (approved):

PR #2908, opened 2021-09-16
[IMP] account: Migration of grouped invoice lines. Allow to import notes and sections

etobella:
I tested the following invoice:
imagen

Once migrated with the original code we obtained the following result
imagen

With the following journal items when it was not grouped:
imagen

With the following journal items when it was grouped:
imagen

With the new code:
imagen
The journal items had the expected values

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -292,7 +292,7 @@ def migration_invoice_moves(env):
         )
         if not env.cr.fetchone():
             break  # exit condition not having more duplicates
-    # 2st: exclude from invoice_tab the grouped ones, and create a new separated ones
+    # 2st: exclude from invoice_tab the grouped ones
     openupgrade.logged_query(
         env.cr, """
         UPDATE account_move_line aml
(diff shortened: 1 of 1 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -310,50 +325,31 @@ def migration_invoice_moves(env):
...
-            SELECT ail.id
-            FROM account_invoice_line ail
-                JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
-                JOIN account_move am ON ail.invoice_id = am.old_invoice_id
-                JOIN account_move_line aml ON am.id = aml.move_id
-            WHERE ail.company_id = aml.company_id AND ail.account_id = aml.account_id
-                AND ail.partner_id = aml.partner_id
-                AND ((ail.product_id IS NULL AND aml.product_id IS NULL) OR ail.product_id = aml.product_id)
-                AND ((ail.uom_id IS NULL AND aml.product_uom_id IS NULL) OR ail.uom_id = aml.product_uom_id)
-                AND ((ail.account_analytic_id IS NULL AND aml.analytic_account_id IS NULL)
```
etobella, line 328: That might give you problems if your system edited the following function
https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_invoice.py#L1234

It might be better to import all the data if possible

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -310,50 +325,31 @@ def migration_invoice_moves(env):
...
-                    OR ail.account_analytic_id = aml.analytic_account_id)
-                AND aml.tax_line_id IS NULL
-                AND aml.id IN %s""", (aml_ids2, ),
```
etobella, line 331: Using this might make the SQL big

addons/account/migrations/13.0.1.1/post-migration.py
legalsylvain: Do you think it could be possible to define at the beginning something like : 
hash_keys = ["account_id", "tax_ids", "tax_line_id", "product_id", "analytic_account_id", "date_maturity", "analytic_tag_ids"]

and then generate an sql query based on the values ? 
so user that has overloaded the function could more easily write the correct settings.

(not sure if what I say is relevant : never migrated from 12.0 to 13.0 for the time being).

etobella: I think it is not relevant. The point is to keep all the data as it was on the invoice. In my casa, I removed a key on that hook and migration removed all the lines except one

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -292,6 +292,21 @@ def migration_invoice_moves(env):
...
+        """
+    )
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_invoice_line ail
+        SET aml_matched = TRUE
+        FROM account_move_line aml
+        WHERE aml.old_invoice_line_id = ail.id
+        """
+    )
```
MiquelRForgeFlow:
```suggestion
```
I think this part is not needed, because aml_matched already have those values there

etobella: I thought the same, but when I tested, I found it was necessary as the not related lines were marked as aml_matched

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: Yeah, when I did that part of the code, the `aml_matched` column was not introduced yet. I was added later by pedro.

addons/account/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: I don't understand. the `aml_matched` is filled previously for the invoice lines matching move lines.

gdgellatly: @MiquelRForgeFlow I have serious doubts on this aml_matched code. I got 2 cases now where it just isn't working. It basically just displays the same problem as before except instead of first it is last match. @etobella is actually just covering the effect, not the cause (which IMHO is the total ignorance of line values in the algorithm), but I spent so long inside that while loop I just can't work out where it goes wrong. It looks perfect.

I honestly feel like we have adopted the wrong approach here altogether with invoice migration and that Odoo SA has got it more right by applying ever restrictive rules and making decision based on amounts, among other things. 

I used this really ugly code here (note it doesn't do grouped, because I didn't need) https://github.com/odoonz/OpenUpgrade/blob/13.0/addons/account/migrations/13.0.1.1/post-migration.py to kind of emulate and mismatches dropped from 300,000+ with current code to 57. Of the 57, 51 were due to an import/export bug allowing overwriting posted journals.

Now Odoo SA does make grouped invoice lines into comments and in this case I feel we do better. 

FYI, here is some enterprise log output from a currently running enterprise mig, I think all in it applies 250 of these conditions. They are actually stored as data in postgres table so can provide all.
```
2021-09-20 06:50:39,940 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: insert query 183: cond:{'same_account', 'same_trimmed_name', 'paid_invoice_posted_move', 'nearly_same_amount', 'same_product'} 
2021-09-20 06:50:43,403 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: fill line mapping with +8/-8 entries using conditions {'same_account', 'same_trimmed_name', 'paid_invoice_posted_move', 'nearly_same_amount', 'same_product'} 
2021-09-20 06:50:43,754 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: still 55318 to match 
2021-09-20 06:50:43,754 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: insert query 184: cond:{'same_chopped_name', 'same_account', 'paid_invoice_posted_move', 'nearly_same_amount', 'same_product'} 
2021-09-20 06:50:44,196 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: fill line mapping with +0/-0 entries using conditions {'same_chopped_name', 'same_account', 'paid_invoice_posted_move', 'nearly_same_amount', 'same_product'} 
2021-09-20 06:50:44,543 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: still 55318 to match 
2021-09-20 06:50:44,543 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: insert query 185: cond:{'same_name_first_line', 'same_account', 'paid_invoice_posted_move', 'nearly_same_amount', 'same_product'} 
2021-09-20 06:50:48,081 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: fill line mapping with +8/-8 entries using conditions {'same_name_first_line', 'same_account', 'paid_invoice_posted_move', 'nearly_same_amount', 'same_product'} 
2021-09-20 06:50:48,428 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: still 55318 to match 
2021-09-20 06:50:48,428 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: insert query 186: cond:{'same_account', 'paid_invoice_posted_move', 'nearly_same_amount', 'substring_name', 'same_product'} 
2021-09-20 06:50:51,878 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: fill line mapping with +8/-8 entries using conditions {'same_account', 'paid_invoice_posted_move', 'nearly_same_amount', 'substring_name', 'same_product'} 
2021-09-20 06:50:52,228 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: still 55318 to match 
2021-09-20 06:50:52,228 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: insert query 187: cond:{'same_account', 'same_not_null_product', 'paid_invoice_posted_move', 'nearly_same_amount'} 
2021-09-20 06:50:55,712 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: fill line mapping with +8/-8 entries using conditions {'same_account', 'same_not_null_product', 'paid_invoice_posted_move', 'nearly_same_amount'} 
2021-09-20 06:50:56,067 2012 INFO db_31437 odoo.addons.base.maintenance.migrations.account.saas-12.4.end-09-account-pocalypse: invoices: still 55318 to match 
```

MiquelRForgeFlow: @gdgellatly of course Odoo SA migration code might be more convoluted, but here @pedrobaeza advocates a simple readable non-repeatable code.

gdgellatly: @MiquelRForgeFlow I'm all for simple code. But as I say I got 2 cases where it just produces wrong results. And I still think ignoring the amounts is flawed. Obviously my hacky flattening is not how Odoo SA does it, it is just 1 function called with different arguments.

And it sounds like @etobella is experiencing something related.

pedrobaeza:
Please split in 2 PRs the section/notes part from the grouped line one.

etobella:
I cannot split it. I mean it is done in the same query, so I don't know how to do that. The fix in `# 4th. Adding all the missing lines
` imports all missing lines, the grouped and the sections.

MiquelRForgeFlow (approved):

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -292,6 +292,21 @@ def migration_invoice_moves(env):
...
     # 2st: exclude from invoice_tab the grouped ones, and create a new separated ones
```
MiquelRForgeFlow:
```suggestion
    # 2st: exclude from invoice_tab the grouped ones
```

as the creation is done in the 4rt step

MiquelRForgeFlow (commented):
With this, I think it can be merged.

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -292,7 +292,22 @@ def migration_invoice_moves(env):
         )
         if not env.cr.fetchone():
             break  # exit condition not having more duplicates
-    # 2st: exclude from invoice_tab the grouped ones, and create a new separated ones
+    # Checking the missing lines to import
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_invoice_line ail
+        SET aml_matched = FALSE
```
pedrobaeza: I don't see this assignation, as there shouldn't be any incorrect put as True. Take into account that this assignation can take a lot of minutes on a very populated table. I think the problem is that the default is NULL, not False/True. Then you should use `WHERE ail.aml_matched is not TRUE` as we have done in other places.

MiquelRForgeFlow: `WHERE ail.aml_matched IS DISTINCT FROM TRUE` is the correct one that excludes both null and False.

etobella: When we migrated, something didn't work if we didn't apply this change :disappointed: 

I can remove it if you prefer and set the IS DISTINCT FROM TRUE

MiquelRForgeFlow: Yeah, let's do this change, and merge. If then you still face the same problem, we can make another PR :)

pedrobaeza (approved):

PR #2911, opened 2021-09-17
[13.0][OU-FIX] payment: Disable test acquirers not published

pedrobaeza:
On v13, if you have an acquirer in test mode, it will be available in the website, but in such mode.

For respecting the website_published setting of v12, we have to disable the acquirer in this version.

@Tecnativa TT31959


MiquelRForgeFlow (approved):
LGTM

PR #2913, opened 2021-09-20
[13.0][FIX] delivery: invoice_policy

chienandalu:
Odoo changed the behavior of delivery charges from v12 to v13 delegating
the mode check into the new `invoice_policy` field in
`delivery.carrier`.

Initially, it was decided to keep the behavior in v12 that recomputes
the price when the picking is validated, but there's no accurate flow
translation between versions and some issues raise with this default:

- Core UI forbids users changing this value in the standard shipping methods.
  Only extra carriers can choose it and the default `estimated` value
  remains.
- The eCommerce behaves weirdly with "Real cost" methods, leading the
  customers to think that they get free delivery orders.

For that reasons, a sane default must be Estimated cost for the majority
of cases.

cc @Tecnativa TT31975

please review @pedrobaeza @victoralmau

### DIFF
--- a/addons/delivery/migrations/13.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/delivery/migrations/13.0.1.0/openupgrade_analysis_work.txt
@@ -8,7 +8,7 @@ delivery     / delivery.carrier         / return_label_on_delivery (boolean): NE
 # NOTHING TO DO: New feature for allowing to print the return label. Default False value is OK for preserving previous behavior (no return label).
 
 delivery     / delivery.carrier         / invoice_policy (selection)    : NEW required, selection_keys: ['estimated', 'real'], req_default: function, hasdefault
-# DONE: post-migration: Fill all delivery.carrier records to have `real` value in invoice_policy, as this is the most similar one to the v12 behavior
+# NOTHING TO DO: Default `estimated` value has is the less conflicting one according the new Odoo rules. Users can't choose 'real' on the core standard methods and the behavior change isn't that intuitive.
 
 delivery     / delivery.carrier         / margin (integer)              : type is now 'float' ('integer')
 # NOTHING TO DO: ORM/PG does the switch automatically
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
If it's the default value, shouldn't it be "Nothing to do" and do nothing?

pedrobaeza:
Also being discussed in #2715

chienandalu:
> If it's the default value, shouldn't it be "Nothing to do" and do nothing?

I wasn't sure if the default value was set in the existing records. If so, then it's just removing stuff

JordiMForgeFlow:
Yes, it is the default value: https://github.com/odoo/odoo/blob/13.0/addons/delivery/models/delivery_carrier.py#L51

MiquelRForgeFlow:
By default is "estimated". That's why I proposed only to put "real" only for the case "invoice_shipping_on_delivery = True". And that's why I prefer https://github.com/OCA/OpenUpgrade/pull/2715 than this PR.

chienandalu:
> By default is "estimated". That's why I proposed only to put "real" only for the case "invoice_shipping_on_delivery = True". And that's why I prefer #2715 than this PR.

But that value can vary in every sale order, as it will only depend on how the order is processed before the picking is validated. Keep also in mind the two main issues we've found for falling back to this default:

- Core UI forbids users changing this value in the standard shipping methods. Only extra carriers can choose it and the default `estimated` value  remains.
- The eCommerce behaves weirdly with "Real cost" methods, leading the customers to think that they get free delivery orders.

So our conclusion is that the sane option is to keep the default value as the less conflicting one and teach the users this change of behavior to let them act accordingly.

JordiMForgeFlow (approved):

MiquelRForgeFlow (approved):

pedrobaeza:
For clarification, the e-commerce weird behavior is a bug of Odoo, but for not having to "fight" with Odoo about it, let's go this way with the saner default. It consists of the initial payment cart render puts the proper price, but when selecting the desired delivery method, the sales line is added with 0, and then the ribbon goes to "Free" in the website.

PR #2919, opened 2021-09-23
[IMP] account: Do the transformation of messages on a single query, not one for each record

etobella:
@pedrobaeza @MiquelRForgeFlow 

Tested. From 42 minutes to 8 in my case.
Also added a commit in order to fix a M2M table. (they need to be recreated in order to have the right Foreign Keys)
The last commit creates a column in order to avoid odoo from filling it on creation. Then, we fill it in post-migration with a single query.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -1037,6 +1037,21 @@ def _recompute_move_entries_totals(env):
     )
 
 
+def fill_account_move_line_missing_fields(env):
+    openupgrade.logged_query(env.cr, """
+        UPDATE account_move_line aml
+        SET account_root_id = aa.root_id
+        FROM account_account aa
+        WHERE aa.id = aml.account_id
+    """)
+    openupgrade.logged_query(env.cr, """
+        UPDATE account_move_line aml
+        SET tax_group_id = at.tax_group_id
+        FROM account_tax at
+        WHERE at.id = aml.tax_line_id
+    """)
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     fill_account_reconcile_model_second_analytic_tag_rel_table(env)
@@ -1048,6 +1063,7 @@ def migrate(env, version):
     fill_account_journal_restrict_mode_hash_table(env)
     archive_account_tax_type_tax_use_adjustment(env)
     fill_account_journal_invoice_reference_type(env)
+    fill_account_move_line_missing_fields(env)
     migration_invoice_moves(env)
     if openupgrade.table_exists(env.cr, 'account_voucher'):
         migration_voucher_moves(env)
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -415,10 +415,11 @@ def migration_invoice_moves(env):
     )
     openupgrade.logged_query(
         env.cr, """
-        UPDATE account_invoice_payment_rel aipr
-        SET invoice_id = am.id
-        FROM account_move am
-        WHERE am.old_invoice_id = aipr.invoice_id
+        INSERT INTO account_invoice_payment_rel
+            (invoice_id, payment_id)
+        SELECT am.id, aipr.payment_id
+        FROM old_account_invoice_payment_rel as aipr
+            INNER JOIN account_move am ON am.old_invoice_id = aipr.invoice_id
         """,
     )
     # Relink analytic tags
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -173,10 +173,8 @@ def migration_invoice_moves(env):
         ai.create_uid, ai.create_date, ai.write_uid, ai.write_date)
         FROM account_invoice ai
         WHERE am.id = ai.move_id AND ai.state not in ('draft', 'cancel')
-        RETURNING ai.id, am.id
         """,
     )
-    ids1 = env.cr.fetchall()
     # Insert moves for draft or canceled invoices
     openupgrade.logged_query(
         env.cr, """
@@ -214,13 +212,10 @@ def migration_invoice_moves(env):
         vendor_display_name, cash_rounding_id, id, create_uid, create_date,
         write_uid, write_date
         FROM account_invoice ai
-        WHERE ai.state in ('draft', 'cancel')
-        RETURNING old_invoice_id, id""",
+        WHERE ai.state in ('draft', 'cancel')""",
     )
-    ids2 = env.cr.fetchall()
-    if ids1 or ids2:
-        _move_model_in_data(
-            env, ids1 + ids2, 'account.invoice', 'account.move')
+    _move_model_in_data(
+        env, 'account.invoice', 'account.move', 'old_invoice_id')
     # Not Draft or Cancel Invoice Lines
     # 1st: update the ungrouped ones
     openupgrade.logged_query(env.cr, "ALTER TABLE account_invoice_line ADD aml_matched BOOLEAN")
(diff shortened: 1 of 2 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -604,35 +605,45 @@ def migration_voucher_moves(env):
...
+    openupgrade.logged_query(env.cr, sql.SQL("""
+        UPDATE mail_followers mf
+            SET res_model = %(new_value1)s, res_id = am.id
+            FROM account_move am
+                JOIN mail_followers mf1
+                    ON am.{field} = mf1.res_id AND mf1.res_model = %(old_value1)s
+                LEFT JOIN mail_followers mf2
+                    ON am.id = mf2.res_id
+                        AND mf2.res_model = 'account.move'
+                        AND mf2.partner_id = mf1.partner_id
```
MiquelRForgeFlow:
```suggestion
                    ON (am.{field} = mf1.res_id AND mf1.res_model = %(old_value1)s)
                LEFT JOIN mail_followers mf2
                    ON (am.id = mf2.res_id
                        AND mf2.res_model = 'account.move'
                        AND mf2.partner_id = mf1.partner_id)
```
parenthesis not needed, but helps readability :)

MiquelRForgeFlow (approved):
LGTM

flotho:
@gaelTorrecillas @MAG-Florian-Chevalier, could you please try this PR ?
Regards

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -220,7 +220,7 @@ def migration_invoice_moves(env):
     ids2 = env.cr.fetchall()
     if ids1 or ids2:
```
pedrobaeza: I think you can remove the `RETURNING` part of both queries and don't store the IDs, as they are not used at all.

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -420,10 +420,11 @@ def migration_invoice_moves(env):
     )
     openupgrade.logged_query(
         env.cr, """
-        UPDATE account_invoice_payment_rel aipr
-        SET invoice_id = am.id
-        FROM account_move am
-        WHERE am.old_invoice_id = aipr.invoice_id
+        INSERT INTO account_invoice_payment_rel
```
pedrobaeza, line 418: If you do this (which is not incorrect at all), you should rename the old table in pre-migration for not twisting references in the same table.

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -1035,6 +1046,21 @@ def _recompute_move_entries_totals(env):
     )
 
 
+def fill_account_move_line_missing_fields(env):
```
pedrobaeza, line 1040: This is not needed. Odoo already performs the exact same query since https://github.com/odoo/odoo/pull/65232

addons/account/migrations/13.0.1.1/post-migration.py
etobella: I already did that https://github.com/OCA/OpenUpgrade/pull/2919/files#diff-056cf6d26f1211252b7063df64efb06222e4797b0973bf784c70b5d3e25d25ceR54

pedrobaeza: Oh, I didn't see it, sorry.

addons/account/migrations/13.0.1.1/post-migration.py
etobella: It is needed, because this fields are computed https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L864

addons/account/migrations/13.0.1.1/post-migration.py
etobella: Done!

addons/account/migrations/13.0.1.1/post-migration.py
pedrobaeza: They are related AFAIK: https://github.com/odoo/odoo/blob/917be4131491dada93f8cb9cd9cc1ae4ce731f2b/addons/account/models/account_move.py#L2644

etobella: They are related but the related field cannot be computed

addons/account/migrations/13.0.1.1/post-migration.py
etobella: Don't worry, too many lines to review, so that could happen to anyone :wink:

addons/account/migrations/13.0.1.1/post-migration.py
pedrobaeza: OK, then no problem to put this that way.

addons/account/migrations/13.0.1.1/post-migration.py
pedrobaeza: I think you haven't pushed this change

etobella: Sorry, i forgot to make the push

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -319,7 +314,10 @@ def migration_invoice_moves(env):
             SELECT ail.id
             FROM account_invoice_line ail
                 JOIN account_invoice ai ON ail.invoice_id = ai.id AND ai.state NOT IN ('draft', 'cancel')
-                JOIN account_move am ON ail.invoice_id = am.old_invoice_id
+                JO
+
+This is not needed. Odoo already performs the exact same query since odoo/odoo#65232
```
pedrobaeza: Copy/paste glitch

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -1035,6 +1041,21 @@ def _recompute_move_entries_totals(env):
     )
 
 
+def fill_account_move_line_missing_fields(env):
+    openupgrade.logged_query(env.cr, """
```
pedrobaeza, line 1041: Have you tested if performing the UPDATE of both fields in one query joining with `LEFT JOIN` the 3 tables works faster?

etobella: I am not sure if that will work, as one of the fields is not required. I am not sure what will happen in that case. On an update you cannot pass the value on LEFT JOIN / ON, you must use it on the WHERE, the query should be like
```sql
        UPDATE account_move_line aml
        SET 
        account_root_id = aa.root_id,
        tax_group_id = at.tax_group_id
        FROM account_tax at, account_account aa
        WHERE aa.id = aml.account_id AND at.id = aml.tax_line_id
```
but I cannot assure that it will work as expected

MiquelRForgeFlow: no, that query is incorrect.

addons/account/migrations/13.0.1.1/pre-migration.py
```diff
@@ -146,36 +152,24 @@ def create_account_move_new_columns(env):
...
         ADD COLUMN account_internal_type varchar""",
     )
     openupgrade.logged_query(
         env.cr, """
         UPDATE account_move_line aml
-        SET account_internal_type = aat.type
-        FROM account_account aa
-        JOIN account_account_type aat ON aa.user_type_id = aat.id
-        WHERE aml.account_id = aa.id""",
+        SET parent_state = am.state
```
MiquelRForgeFlow: account_internal_type?

etobella: Thanks, error joining it in a single script smile :smile:

MiquelRForgeFlow (approved):

pedrobaeza (approved):
Another bite we do on the time spent. Thanks!

PR #2922, opened 2021-09-30
[13.0][OU-IMP] account: remove fks from obsolete table

MiquelRForgeFlow:
In pre-migration, table `account_invoice_payment_rel` is renamed to `old_account_invoice_payment_rel`

Due to https://github.com/OCA/OpenUpgrade/commit/6c4b6cbcf3e0d9cbdaf9ff2a287c29476e40640e.

### DIFF
--- a/addons/account/migrations/13.0.1.1/pre-migration.py
+++ b/addons/account/migrations/13.0.1.1/pre-migration.py
@@ -309,6 +309,7 @@ def migrate(env, version):
             WHERE key = 'sale.use_sale_note'"""
         )
     openupgrade.rename_models(cr, _model_renames)
+    openupgrade.remove_tables_fks(env.cr, ["account_invoice_payment_rel"])
     openupgrade.rename_tables(cr, _table_renames)
     openupgrade.add_fields(env, _field_adds)
     type_change_account_fiscal_position_zips(env)


addons/account/migrations/13.0.1.1/pre-migration.py
```diff
@@ -309,6 +309,7 @@ def migrate(env, version):
             WHERE key = 'sale.use_sale_note'"""
         )
     openupgrade.rename_models(cr, _model_renames)
+    openupgrade.remove_tables_fks(env.cr, ["account_invoice_payment_rel"])
```
pedrobaeza, line 312: Shouldn't we should do this also with `account.invoice` and `account.invoice.line`?

MiquelRForgeFlow: It's already done in base.

etobella: :thinking: It will be probably be good to remove them in all three tables. Just a question, might be faster if FK exists for `account.invoice` and `acocunt.invoice.line`?

etobella (approved):

addons/account/migrations/13.0.1.1/pre-migration.py
pedrobaeza: Curious that it's done in base instead of here. No fk will be used until the load of this module. @etobella It's faster without FKs as no checks are performed.

MiquelRForgeFlow: It's done in base because 2 reasons:
- historically, we have removed the FK for all obsolete tables in base, just to have them together in a place, to easily check them
- when the migration of base was done, we put all this account related tables already, we know we needed to delete their FKs. Also, it includes others like account_invoice_tax, vouchers and many2manys.

pedrobaeza: It was done historically in base because previous your work on the migration logs, the only place obsolete models were logged was on base, but now obsolete models are also logged on each module. We can let it this version on base, but not this one here and the rest on base. Homogeneity.

MiquelRForgeFlow:
Ups, it's already there https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/base/migrations/13.0.1.3/pre-migration.py#L567.

PR #2924, opened 2021-10-04
[13.0][OU-FIX] website_slides: categories migration

chienandalu:
The fields `is_preview` and `is_published` are values expected be `True` as default when the category is created. It can be seen how in the `create()` and in the `write()` this is explicitly evaluated as long as in the controller, which is the main way to create categories now. Also, we must ensure that those categories are active. Al should be as the old specific model didn't have the field itself.

cc @Tecnativa TT32244

We should also decide what to do with public slides. In v12, they were public or not. In v13 with the eLearing orientation  they're visible only if:

- They are free previews (field `is_preview` set to `True`)
- The user is subscribed to the channel.

So to translate that from one version to the other we should either:

- Set `is_preview` on (then an ugly `Free preview` badge is shown.
- Autosubscribe public and portal users.

What's would be the better way?

### DIFF
--- a/addons/website_slides/migrations/13.0.2.0/post-migration.py
+++ b/addons/website_slides/migrations/13.0.2.0/post-migration.py
@@ -63,12 +63,14 @@ def convert_slide_categories(env):
     openupgrade.logged_query(
         env.cr, sql.SQL("""
         INSERT INTO slide_slide (
-            create_date, create_uid, write_date, write_uid, is_category,
-            name, channel_id, old_category_id, sequence, slide_type
+            active, create_date, create_uid, write_date, write_uid, is_category,
+            name, channel_id, old_category_id, sequence, slide_type, is_published,
+            is_preview
         )
         SELECT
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
For the tag for other times, you can put it as `[OU-FIX]`.

For the public thing, are you able to subscribe to groups or to have an auto-subscribe mechanism?

chienandalu:
> For the public thing, are you able to subscribe to groups or to have an auto-subscribe mechanism?

Yes, you can assign groups to the channel and users belonging to them will autosubscribe. Otherwise, only portal and internal users can subscribe manually (I should check this fact) to the public channels in order to view their slides.

pedrobaeza:
Then the migration script can subscribe automatically Portal and Public groups to this public channels for equivalent behavior, isn't it?

chienandalu:
Yes if we choose that way to go :) The other is to set those slides as `is_preview`, which has the lateral effect of showing that ugly label. So answering myself :sweat_smile: , yes the more seamless method will be autosubscribing those groups

PR #2926, opened 2021-10-04
[14.0][IMP] openupgrade_framework: avoid AttributeError in view validation

MiquelRForgeFlow:
If corresponding_field is None, we need to avoid the "AttributeError: 'NoneType' object has no attribute 'get'" error.

See https://github.com/odoo/odoo/pull/77739

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
@@ -47,7 +47,16 @@ def handle_view_error(
             self.write({"active": False})
 
 
+def _postprocess_view(self, node, model, validate=True, editable=True):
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Upstream fix would have been better. I don't understand the decisions these people make when they look at a trivial PR like that.

As it is now, It seems like your change has two redundant angles: it ensures the NameManager objects are called with validate=False, and it disarms its method `check_view_fields`. However, that method immediately returns if validate is not set: https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_ui_view.py#L1978. Can you remove one of the angles?

MiquelRForgeFlow:
> Can you remove one of the angles?

Done, I think now it encompasses all cases.

StefanRijnhart (approved):

PR #2927, opened 2021-10-04
[14.0][FIX] digest: delete not-null constraint

JordiBForgeFlow:
Otherwise when the related mail is removed odoo fails because there's a digest.digest record that still references the old template.

cc @MiquelRForgeFlow @Chanakya-OSI

### DIFF
--- a/openupgrade_scripts/scripts/digest/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/digest/14.0.1.1/pre-migration.py
@@ -1,0 +1,8 @@
+# Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/digest/14.0.1.1/post-migration.py
```diff
@@ -6,3 +6,10 @@
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "digest", "14.0.1.1/noupdate_changes.xml")
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE digest_digest
+        SET mail_template_id = NULL
+        """,
+    )
```
MiquelRForgeFlow: I think this is not needed, because in pre-migration you have already moved these values.

MiquelRForgeFlow (approved):
I have updated the PR. Just to delete the constraint instead of deleting/moving the values.

pedrobaeza (approved):
A new thing to have in mind: deleted required fields should be lifted from constrains

PR #2928, opened 2021-10-07
[13.0][FIX] account: generic journal is required

MiquelRForgeFlow:
During the load of the `account` module, the new `journal_id` field of `account.move` model is filled by a default journal of type 'general'. If your company doesn't have any journal of that type, then the following UserError of https://github.com/OCA/OpenUpgrade/blob/13.0/addons/account/models/account_move.py#L70 is blocking.

### DIFF
--- a/addons/account/migrations/13.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/account/migrations/13.0.1.1/openupgrade_analysis_work.txt
@@ -323,7 +323,7 @@ account      / account.move             / reverse_date (date)           : DEL
 
 account      / account.move             / currency_id (many2one)        : not a function anymore
 account      / account.move             / currency_id (many2one)        : now required, req_default: function
-# DONE: post-migration: filled empty cases
+# DONE: pre-migration: filled empty cases and add required constraint
 
 account      / account.move             / matched_percentage (float)    : DEL
 # NOTHING TO DO: obsolete fields, replaced by _get_cash_basis_matched_percentage() method
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -669,20 +669,6 @@ def fill_account_move_type(env):
     )
 
 
-def fill_account_move_currency_id(env):
-    openupgrade.logged_query(
-        env.cr, """
-        UPDATE account_move am
-        SET currency_id = COALESCE(am_rel.cur1, am_rel.cur2)
-        FROM (SELECT am1.id, aj.currency_id as cur1, rc.currency_id as cur2
-          FROM account_move am1
-          LEFT JOIN account_journal aj ON aj.id = am1.journal_id
-          LEFT JOIN res_company rc ON rc.id = aj.company_id) am_rel
-        WHERE am.id = am_rel.id AND am.currency_id IS NULL
-        """,
-    )
-
-
 def fill_res_partner_ranks(env):
     # customer_rank
     openupgrade.logged_query(
@@ -1070,7 +1056,6 @@ def migrate(env, version):
         migration_voucher_moves(env)
     fill_account_move_reversed_entry_id(env)
     fill_account_move_type(env)
-    fill_account_move_currency_id(env)
     fill_res_partner_ranks(env)
     try_to_fill_account_account_tag_country_id(env)
     create_account_tax_repartition_lines(env)
--- a/addons/account/migrations/13.0.1.1/pre-migration.py
+++ b/addons/account/migrations/13.0.1.1/pre-migration.py
@@ -222,6 +222,24 @@ def fill_account_move_commercial_partner_id(env):
     )
 
 
+def set_account_move_currency_id_required(env):
+    # with this, we avoid calling _get_default_currency during the load
+    openupgrade.logged_query(
+        env.cr,
+        """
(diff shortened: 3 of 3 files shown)


legalsylvain (changes requested):
Hi @MiquelRForgeFlow. 

I'm not sure to understand this PR. You say "_the **new** journal_id field of account.move_", but AFAIK, the fields exist in both V12 and V13 revision (and are required)

V12 : https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_move.py#L101
V13 : https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L119

pedrobaeza (changes requested):
What a weird case, but OK to handle it. The problem is that you can't use ORM with a model that not yet exists (`account.journal`), as this is pre-migration. You need to do it through SQL.

pedrobaeza:
Tests seems to not be working correctly, or an error will arise using `env["account.journal"]` on pre-migration.

MiquelRForgeFlow:
> I'm not sure to understand

@legalsylvain Sorry, after some debugging, I found that the default methods are always called in the models during load, independently if those models have records or not. So `self._get_default_journal()` always is called, where `self` is an empty record set.

> The problem is that you can't use ORM with a model that not yet exists

@pedrobaeza Solved, but I didn't want to create the journal/sequence in the pre-migration through SQL, so I did the following approach: create a hardcoded one in the pre-migration, and then in the post-migration I delete the hardcoded one and create the desired ones through ORM. I know it's a bit ugly.

legalsylvain:
thanks for your answer.

> @legalsylvain Sorry, after some debugging, I found that the default methods are always called in the models during load, independently if those models have records or not. So self._get_default_journal() always is called, where self is an empty record set.

You mean during the upgrade, or allways, when you launch odoo ? I see it very weird. I mean, why call _default function when the field exist in database ?

As an alternative solution, we could just patch the function ``_get_default_journal`` in OpenUpgrade here https://github.com/oca/openupgrade/blob/13.0/addons/account/models/account_move.py#L40 with :

```
    @api.model
    def _get_default_journal(self):
    if not self:
        return False
```

I don't see it as a problem of Openupgrade. What do you think ?

pedrobaeza:
@legalsylvain no changes should be done outside migration scripts. That changes are not easily tracked, and can provoke conflicts on upstream changes. Anyways, indeed no default is called if the column exists, so something else is going on, and you must check what it is.

legalsylvain:
@legalsylvain no changes should be done outside migration scripts. That changes are not easily tracked, and can provoke conflicts on upstream changes.

:100:
However, sometimes it occures. (https://github.com/OCA/OpenUpgrade/blob/13.0/addons/stock/__init__.py#L15). It just a balance between adding 66 lines of weird and custom code that could have undesired side effects, and adding 2 lines of harmless code.
Just a proposal, maybe not the good solution however. and indeed, i'd like to know why this function is called !

Note : in V14+ it will be very more easy to track here : https://github.com/OCA/OpenUpgrade/tree/14.0/openupgrade_framework/odoo_patch/odoo/addons

> Anyways, indeed no default is called if the column exists, so something else is going on, and you must check what it is.

just to be sure, the "you" means me ? or @MiquelRForgeFlow. i didn't planned to investigate that problem for the time being.

regards.

pedrobaeza:
I mean Miquel. Of course I don't want also to add a bunch of code for a very corner case, but if the problem is real and can be mitigated, let's do it in migration scripts, although it requires more code, as it's very isolated in a specific method. But changing core code for the corner case, it's a no.

MiquelRForgeFlow:
> you must check what it is.

Ok, I dug enough. currency_id field v12 is not required, but in v13 is required. Then, when odoo tries to add the not-null constrain https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L914, it also calls the default method https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L2463. So, `_get_default_currency` is called, which also calls the `_get_default_journal`.

Maybe we could try to add the not-null constrain in the pre-migration (and filling currency_id in case there is an empty case). This way, the default method will not be called in the load. What do you think?

pedrobaeza:
Isn't better to add in your 12.0 DB the missing miscellaneous journal, hehe? Out of jokes, doing what you propose isn't going to be avoid the problem, as it will be called anyways (or maybe I miss the part of the default method). It's very interesting though that you have dug for knowing this awkward init process.

Out of this, isn't currently code for filling missing currency_id values for the new required?

Also, did you investigate why the CI was green while doing something incorrect like creating the journal by ORM in pre?

MiquelRForgeFlow:
> as it will be called anyways

No, when odoo detects now that the column currency_id has the not null constrain, then it doesn't call the default method anymore. Before it was called because the column for the field already existed but didn't have the constrain.

> why the CI was green while doing something incorrect

I didn't check, I suppose the demo ddbb used in the CI already have the general journal, thus the code wasn't executed probably. And it wasn't incorrect, it just was ugly that approach.


@pedrobaeza please, review again, I think the code now looks good, and we could merge it :)

pedrobaeza (approved):

pedrobaeza:
@legalsylvain please update your review

legalsylvain (approved):
Thanks for this work !

/ocabot merge nobump

PR #2930, opened 2021-10-09
[14.0][website_partner][mig] add migration script.

JordiBForgeFlow:


legalsylvain (changes requested):
hi. could you :
- add a upgrade_analysis_work.txt, as usual ?
- in this file, add a comment to say why it has to be done by script ? (orm should add this new field without script.)
- mark the module as done ? 

Thanks ! 

ref of the analysis file : https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/website_partner/14.0.0.1/upgrade_analysis.txt
ref new field : https://github.com/odoo/odoo/blob/14.0/addons/website/models/mixins.py#L26

JordiBForgeFlow:
@legalsylvain I realized that in reality there's nothing to do in this module. I marked the module as done.

legalsylvain (approved):

pedrobaeza:
`website` migration scripts are not yet merged, so this PR shouldn't be merged. Please check this the next time, @legalsylvain

JordiBForgeFlow:
Oops, my bad.

pedrobaeza:
Ah, right, it was you who merged. That makes all the dependency chain to be installed in the CI and can provoke future errors. That's why we normally don't merge them until we are sure that the rest of the migrations are merged.

PR #2935, opened 2021-10-14
[14.0][FIX] apriori: project_stage_closed rename -> project_stage_closed merge

MiquelRForgeFlow:
'project' is a module of odoo.


pedrobaeza:
This is not correct. The module we are merging is from OCA/project, not Odoo, although the target is a core module.

MiquelRForgeFlow:
Hey, project_stage_closed should be **merged** into "project", not **renamed** into "project". If you have both project and project_stage_closed installed, and the proceed with the migration, when doing the rename you will get an error:

`DETAIL: Key (name)=(project) already exists.`

pedrobaeza:
OK, I understand

PR #2945, opened 2021-10-26
[8.0][FIX] migrate_stock_production_lot syntax errors

JordyRitzen:
The syntax errors give the following end result:
ProgrammingError: syntax error at or near "FROM"
LINE 4:             FROM stock_quant_move_rel q, stock_move sm

The comma is invalid and q is only known inside the select


pedrobaeza (approved):
Thanks for the contribution

PR #2947, opened 2021-10-27
[14.0][FIX] base: delete duplicated properties and disable filters

MiquelRForgeFlow:
.

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """ Call disable_invalid_filters in every edition of openupgrade """
+    openupgrade.disable_invalid_filters(env)
--- a/openupgrade_scripts/scripts/base/14.0.1.3/pre-migrate.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/pre-migrate.py
@@ -20,10 +20,9 @@ except ImportError:
         " The upgrade process will not work properly."
     )
 
-
 module_category_xmlid_renames = [
     # Module category renames were not detected by the analyze. These records
-    # are created on the fly when intializing a new database in
+    # are created on the fly when initializing a new database in
     # odoo/modules/db.py
     (
         "base.module_category_accounting_expenses",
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Why do you always pack in the same PR totally different things? The invalid filters thing is direct, but the other must be checked carefully, so I can't review just now.

MiquelRForgeFlow:
Because 1) my commodity, 2) I don't have rush to merge for the invalid filters thing 3) both things are from `base` module, so they are not "totally" different things.

pedrobaeza:
Put the deletion in a method for better isolation, and use window function for performing the deletion:

```
DELETE FROM tablename
WHERE id IN (
    SELECT id
    FROM (
        SELECT id, row_number() over (partition BY column1, column2, column3 ORDER BY id) AS rnum
        FROM tablename
    ) t
    WHERE t.rnum > 1);
```

You have to check in both cases what happens with null values.

MiquelRForgeFlow:
Your code suggestion has two issues:
1- the partition doesn't handle the null values cases, the COALESCE part is important.
2- It's not that simple as `WHERE id IN (`, because we want to save one record for each duplication, not deleting all duplicated records.

pedrobaeza:
About the second part, you are only deleting the duplicated, leaving the first one. That's the idea of the `rnum > 1`. About nulls, I think they are properly handled with the partition window.

MiquelRForgeFlow:
Done and tested :)

pedrobaeza:
My only doubt now is which value was taken in previous version. Maybe it was not the lowest ID? Please check it, and if not, you will need to change the `ORDER BY` part.

MiquelRForgeFlow:
In https://github.com/odoo/odoo/commit/e85faf398659a5beb0b1570a06af64dcf78dc1c8, it says it was nondeterministically:

```
Having two properties for the same field, company and res_id, leads to
inconsistencies through the code, as reading a company-dependent field
nondeterministically returns one of the available property value.
```

pedrobaeza:
Puf... but let's check by direct experimentation the most common one. On v13, please define 2 times a property of a customer, and check which one is retrieved on UI, and we will assume that one as "the normal one".

MiquelRForgeFlow:
After investigation, the result is that it doesn't matter. If duplicated properties, the query https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/ir_property.py#L234 return multiple lines, obviously, but the result is used to update a dictionary, see:
Selection_004
Which having single or duplicated doesn't have any effect to retrieve the referenced value.

I think we can merge as it is now.

Note: BTW, I detected an error when deleting duplicated properties, which I just made a PR to odoo to fix that.

pedrobaeza:
I think that it does matter then, as the final result in the dictionary is the last one, not the first, so you should change the order.

MiquelRForgeFlow:
Changed the order.

pedrobaeza (approved):

PR #2948, opened 2021-10-28
[14.0][FIX] product: assure m2m table has correct values

MiquelRForgeFlow:
If you come migrating from older versions, you may have problems with this table.


pedrobaeza:
You have to be more deterministic, and assure that this introduced code is really due to the migration path, and not a bad DB condition that must be repaired only for that specific DB. We can't fill OU with a bunch of patches with the condition of `What if...`

legalsylvain:
CC : @sebastienbeau : you could be interested by this one.

MiquelRForgeFlow:
Note: the query should use `IS NOT DISTINCT FROM` instead.

PR #2949, opened 2021-10-28
[12.0][FIX] survey: drop obsolete constraint

MiquelRForgeFlow:
This constraint was deleted in v12.


pedrobaeza (approved):
You can merge it when ready.

PR #2950, opened 2021-11-01
[13.0][FIX] product: test message

kos94ok-3D:
First commit Steps to reproduce:
1. Create a Product Template with
   Attribute lines: `Color (Red, Blue)`, `Material (Iron, Wood)`
 2. Create and Validate transfer with Variant `(Red, Iron)`
 3. Remove `Red` value from attribute line
  (on this step we lose here link between `product.attribute.value` and `product.template.attribute.value`)
 4. run migration
 5. Go to Product Template form
 6. Press on `Configure Variants` button 
 7. Set `Inactive` filter
 
Actual Result:
Missing attribute for inactive `product.template.attribute.value`

Second commit Steps to reproduce:
1. Create a Product Template with
   Attribute lines: `Color (Red, Blue)`, `Material (Iron, Wood)`
 2. Create BOM for this Product in the BOM line set apply on `Color: Red`
 3. Remove `Red` value from attribute line
  (on this step `product.template.attribute.value` and `product.variant` will be removed)
 4. run migration
 5. Go to Product Template form
 6. Press on `Configure Variants` button 
 7. Set `Inactive` filter not found 
Actual Result:
Missing `Color: Red` record

8. Go to BOM form
Actual Result:
Missing `Color: Red` in BOM line

### DIFF
--- a/addons/product/migrations/13.0.1.2/post-migration.py
+++ b/addons/product/migrations/13.0.1.2/post-migration.py
@@ -17,6 +17,19 @@ def fill_product_template_attribute_value_attribute_line_id(env):
             ptav.product_attribute_value_id = avtalr.product_attribute_value_id
         """,
     )
+    # set attribute_line_id for archived ptals
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE product_template_attribute_value ptav
+        SET attribute_line_id = ptal.id
+        FROM product_template_attribute_line ptal
+        JOIN product_attribute_value pav ON pav.attribute_id = ptal.attribute_id
+        JOIN product_template pt ON ptal.product_tmpl_id = pt.id
+        WHERE ptav.attribute_line_id IS NULL
+            AND ptav.product_tmpl_id = pt.id
+            AND ptav.product_attribute_value_id = pav.id
+        """,
+    )
 
 
 def fill_product_template_attribute_value__attribute_id_related(env):
--- a/addons/product/migrations/13.0.1.2/post-migration.py
+++ b/addons/product/migrations/13.0.1.2/post-migration.py
@@ -4,34 +4,6 @@
 from openupgradelib import openupgrade
 
 
-def fill_product_template_attribute_value_attribute_line_id(env):
-    openupgrade.logged_query(
-        env.cr, """
-        UPDATE product_template_attribute_value ptav
-        SET attribute_line_id = ptal.id
-        FROM product_template_attribute_line ptal
-        JOIN product_template pt ON ptal.product_tmpl_id = pt.id
-        JOIN product_attribute_value_product_template_attribute_line_rel
-            avtalr ON avtalr.product_template_attribute_line_id = ptal.id
-        WHERE ptal.active = TRUE AND ptav.product_tmpl_id = pt.id AND
-            ptav.product_attribute_value_id = avtalr.product_attribute_value_id
-        """,
-    )
-    # set attribute_line_id for archived ptals
-    openupgrade.logged_query(
-        env.cr, """
-        UPDATE product_template_attribute_value ptav
-        SET attribute_line_id = ptal.id
-        FROM product_template_attribute_line ptal
-        JOIN product_attribute_value pav ON pav.attribute_id = ptal.attribute_id
-        JOIN product_template pt ON ptal.product_tmpl_id = pt.id
-        WHERE ptav.attribute_line_id IS NULL
-            AND ptav.product_tmpl_id = pt.id
-            AND ptav.product_attribute_value_id = pav.id
-        """,
-    )
-
-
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Maybe this links with #2672 ? Can't this be done in the existing queries?

kos94ok-3D:
> Maybe this links with #2672 ? Can't this be done in the existing queries?

I tested and his PR did not cover my case

kos94ok-3D:
@MiquelRForgeFlow Could you confirm that my commit can replace your commits from https://github.com/OCA/OpenUpgrade/pull/2672:
https://github.com/OCA/OpenUpgrade/pull/2672/commits/ed0a6d48a835fb7498cb7c1eafffe3768680553e
https://github.com/OCA/OpenUpgrade/pull/2672/commits/398cea93e36983fdb0757b9db5d3ea0f13a40ddc
https://github.com/OCA/OpenUpgrade/pull/2672/commits/f58f94c5aeba7da742c7f46d5a0a1127c3e03931
https://github.com/OCA/OpenUpgrade/pull/2672/commits/97a28461e1f2fa1ff6c307cbcfc2d97ffe91f156
https://github.com/OCA/OpenUpgrade/pull/2672/commits/5f7c64316fe75d08c50f27e43028c6fcce99eafc

addons/product/migrations/13.0.1.2/post-migration.py
```diff
@@ -17,6 +17,19 @@ def fill_product_template_attribute_value_attribute_line_id(env):
...
         """,
     )
+    # TODO: add comment
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE product_template_attribute_value ptav
+        SET attribute_line_id = ptal.id
+        FROM product_template_attribute_line ptal
+        JOIN product_attribute_value pav ON pav.attribute_id = ptal.attribute_id
+        JOIN product_template pt ON ptal.product_tmpl_id = pt.id
```
MiquelRForgeFlow: Here you are finding the ptal through pav using attribute_id. But the correct link between pav and ptal is product_attribute_value_product_template_attribute_line_rel table. In https://github.com/OCA/OpenUpgrade/pull/2672, I fill that m2m table in https://github.com/OCA/OpenUpgrade/pull/2672/commits/ed0a6d48a835fb7498cb7c1eafffe3768680553e for archived ptals (the active ones should already have records there). Then, the filling of attribune_line_id (which your PR is about), I move from post-migration to pre-migration in https://github.com/OCA/OpenUpgrade/pull/2672/commits/398cea93e36983fdb0757b9db5d3ea0f13a40ddc and https://github.com/OCA/OpenUpgrade/pull/2672/commits/f58f94c5aeba7da742c7f46d5a0a1127c3e03931. If those two commits don't fix your issue, then please check what thing about these two commits are wrong.

MiquelRForgeFlow (commented):
I think it should be a `complement` instead of a `replace`, because my PR fixes other things unrelated to these PR.

addons/product/migrations/13.0.1.2/post-migration.py
kos94ok-3D: The problem here https://github.com/OCA/OpenUpgrade/commit/ed0a6d48a835fb7498cb7c1eafffe3768680553e:
изображение
`insert_ptal` not contain lines for my case.
Because that query needs only for case when we remove full line like:
изображение
изображение

But I have a case when not removing all values from the attribute line:
изображение

MiquelRForgeFlow: Interesting, I will check! :D

kos94ok-3D: I can attach DB created for all of these cases if you need

MiquelRForgeFlow: Ok, if it's not a problem for you, please attach the DB :D

kos94ok-3D: [test_product_v12.zip](https://github.com/OCA/OpenUpgrade/files/7469252/test_product_v12.zip)
login: `admin`
password: `admin`

kos94ok-3D:
@MiquelRForgeFlow
in the second commit, I simplified your query.
And I think it has to cover all cases.

MiquelRForgeFlow:
@kos94ok-3D thanks to you, I revised my old PR and thus I made https://github.com/OCA/OpenUpgrade/pull/2672/commits/83df73bb6490f8c060f510248ad6a47a310d56fc in https://github.com/OCA/OpenUpgrade/pull/2672. Hope you can test and if it works, you can close this PR. If not, just tell me what doesn't work.

kos94ok-3D:
Done in #2672.

PR #857, opened 2017-05-30
[10.0] openupgradelib requirement

NL66278:
Current behavior before PR: Pip installed version of OpenUpgrade server does not run, because of missing openupgradelib

Desired behavior after PR is merged: Pip installed OpenUpgrade server runs correctly

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #2951, opened 2021-11-02
[FIX] dependency to openerp-client-lib

legalsylvain:
Hi. 

I try to use ``openupgrade_records`` to make diff from 10.0 to 12.0 custom module. 
I can not install the module because the dependency is incorrect.

this patch align the dependency of the module to the requirement of the main project.

see : https://github.com/OCA/OpenUpgrade/blob/10.0/requirements.txt#L18

related PR : #857

thanks.

### DIFF
--- a/odoo/addons/openupgrade_records/__manifest__.py
+++ b/odoo/addons/openupgrade_records/__manifest__.py
@@ -18,6 +18,6 @@
     ],
     'installable': True,
     'external_dependencies': {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't that declaration the name of the import?

legalsylvain:
Yes, you're right.

I was completely wrong on this point.

sorry for the noice.

PR #2955, opened 2021-11-03
[14.0][IMP] project: pre-create computed stored fields

MiquelRForgeFlow:
Tested.

### DIFF
--- a/openupgrade_scripts/scripts/project/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/project/14.0.1.1/post-migration.py
@@ -16,6 +16,13 @@ def map_project_project_rating_status(env):
 
 def _fill_res_users_m2m_tables(env):
     # TODO: Take into account channels and task followers part of the old rule
+    # Remove temp table and re-create m2m table through ORM method
+    openupgrade.logged_query(env.cr, "DROP TABLE project_allowed_internal_users_rel")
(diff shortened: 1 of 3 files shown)


legalsylvain (approved):
Thanks. makes sense.

A question : 

>  ``DONE: pre-migration: filled by SQL, to avoid computations and inverses``

I understand "computations" but I don't undertand inverses. What do you mean ?

MiquelRForgeFlow:
Well, the fields have inverse methods... 😅 See https://github.com/odoo/odoo/blob/14.0/addons/project/models/project.py#L822

pedrobaeza (commented):
Two comments:

- I don't see a comment in the analysis file for the project_allowed_portal_users_rel thing.
- I think the rest of the m2m table setup (fk constraints and indexes) is not correctly done if you create it that way. Check how I did in purchase v13

MiquelRForgeFlow:
Done.

> I don't see a comment in the analysis file for the project_allowed_portal_users_rel thing.

What do you mean? Check https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/project/14.0.1.1/upgrade_analysis_work.txt#L35

MiquelRForgeFlow:
In v14, due to https://github.com/odoo/odoo/commit/6eb7d5c6a07f678e1d1adca7137652e0626ea17e, we cannot call `update_db` like in v13. But the new trick of calling `init_models` seems to work fine.

pedrobaeza (approved):

PR #2956, opened 2021-11-03
[FIX] 12.0 openupgrade_records : make possible to call oldier version of odoo. (10.0 for instance)

legalsylvain:
Hi. 

I'm working on a migration from 10.0 to 12.0. In that context, I'd like to analyze differencies between code regarding custom addons. I so wanted to use ``openupgrade_records`` to generate analysis file. the modules doesn't exist in V11. So I made an old instance with openupgrade 10.0 and a new instance with openupgrade 12.0 (and openupgrade_records installed on both). 
(see : https://oca.github.io/OpenUpgrade/analyse.html)

However, openupgrade_records V12 seems to only work with openupgrade_rercords V11. when calling V10, some fields / data are missing.

this patch make such analysis possible, skipping missing keys to avoid errors. I guess that the final analysis is maybe not the exact one, but it allows to have a good overview of the diff (and so the workload).

What do you think regarding that patch ? do you have similar needs ? In my opinion, the patches don't hurt the current algorigthm.

CC : original author : @StefanRijnhart 

regards.

### DIFF
--- a/odoo/addons/openupgrade_records/lib/compare.py
+++ b/odoo/addons/openupgrade_records/lib/compare.py
@@ -57,14 +57,14 @@ def compare_records(dict_old, dict_new, fields):
         elif field == 'model':
             if model_rename_map(dict_old['model']) != dict_new['model']:
                 return False
-        elif field == 'other_prefix':
+        elif field == 'other_prefix' and 'prefix' in dict_old:
             if dict_old['module'] != dict_old['prefix'] or \
                     dict_new['module'] != dict_new['prefix']:
                 return False
             if dict_old['model'] == 'ir.ui.view':
                 # basically, to avoid the assets_backend case
                 return False
-        elif dict_old[field] != dict_new[field]:
+        elif field in dict_old and dict_old[field] != dict_new[field]:
             return False
     return True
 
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
The downside is that if a field is missing due to a bug it will go unnoticed after this change.

legalsylvain:
> The downside is that if a field is missing due to a bug it will go unnoticed after this change.

totally agree. For that reason, I don't know what to do with that branch !

Pro : Other people could requires similar changes for similar custom analysis.
Cons : it can hide errors. (in fact, theoritical problem, because in the Openupgrade project, now that 12.0 is unmaintained by Odoo SA, the official branch will not have new changes, and so regenerating a new analysis for Odoo SA modules will maybe never occur). But I see the problem.

I was thinking to make a check regarding the target version (the old one) something like "if version == 11 // else" but it fear it generate a huge diff, and a not readable code.

In a general way, we could imagine to support that feature (analysis between more than 1 revision) for the migration of the module ``upgrade_analysis`` in (OCA / server-tools repo).

StefanRijnhart:
Yes, agreed. The proper way to do so would be to implement versioning of the interface and then implement backwards (or forwards) compatible support for other versions but as it is, even if you contribute the initial versions on all branches, the project does not have enough contributors to maintain it all. So I think you might want to keep this code to yourself.

legalsylvain:
Thanks for your answer, @StefanRijnhart. 

So : 
- closing this PR.
- I'll try to add tests of the framework function. (``compare`` function for exemple) when migrating to 15.0 (https://github.com/OCA/server-tools/pull/2196). maybe it will make the code safer, and will allow such if / then / else in the futur.

PR #2957, opened 2021-11-03
[13.0][IMP] mrp: removed useless join

kos94ok-3D:


MiquelRForgeFlow (approved):
Maybe we could fuse it with https://github.com/OCA/OpenUpgrade/pull/2672/commits/f44bded6753014e55ec8cf0aa909d71c8bc049c8 of https://github.com/OCA/OpenUpgrade/pull/2672 haha

PR #2958, opened 2021-11-05
[MIG] survey_description to survey

olgamarcocb:
The "description" section and question field could be edited on 14.0 but not on 13.0


etobella (approved):

pedrobaeza:
Is the field name the same on both versions?

etobella:
Fiels already exist on the survey module, but they are hidden. The survey_desxription module just show them(there is no field declaration)

etobella:
There was a mistake, one the fields is there but it is not shown.

@kluna1998 OCA/survey#60

Sorry, we didn't notice :sob:

PR #2959, opened 2021-11-08
[13.0][OU-REV] Revert "[IMP] stock_account: add account_move_id information"

pedrobaeza:
This reverts commit 83ae489a1dc2728d355d90f62d06578f7ccf23cc.

The problem is the performance.

@Tecnativa

### DIFF
--- a/addons/stock_account/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock_account/migrations/13.0.1.1/post-migration.py
@@ -33,7 +33,6 @@ def _prepare_common_svl_vals(move, product):
         "remaining_value": 0.0,
         "quantity": 0.0,
         "old_product_price_history_id": None,
-        "account_move_id": move["account_move_id"],
     }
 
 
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
Is there a way to keep the feature improving the performance? I really need this.

pedrobaeza:
I will tell you about the performance penalty if it's true or not, but if so, it's your duty to make this opt-in or look for an optimized way, as we provide this initial implementation avoiding the performance problems.

AaronHForgeFlow:
I will to your results then.

kos94ok-3D:
I think the problem is that here, we are not filtered by company_id and product_id
<pre>
WITH account_move_rel AS (
    SELECT id, create_date
    FROM (
        SELECT id, create_date, COUNT(*) OVER(PARTITION BY create_date) AS qty
        FROM account_move
        WHERE stock_move_id IS NULL
    ) foo
    WHERE qty = 1
)
</pre>

pedrobaeza:
Definitively the problem is in this change, decreasing 28 hours the migration on a DB with 49K products and 500K stock moves. Merging the reversal. No partition window should be done if you want to be performant enough, and definitively this alternative query should be only activated if stock accounting is done.

kos94ok-3D:
Just link: https://github.com/OCA/OpenUpgrade/pull/3015

PR #2960, opened 2021-11-08
[14.0][MIG] website_cookie_notice: merged into website

MiquelRForgeFlow:
Adapting to of https://github.com/OCA/website/pull/814#issuecomment-957527622.

### DIFF
--- a/openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
@@ -1,0 +1,36 @@
+# Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def website_cookie_notice_post_migration(env):
+    cookie_message = env.ref("website.cookie_message", raise_if_not_found=False)
+    if cookie_message:
+        websites = env["website"].search([])
+        websites.write({"cookies_bar": True})
+        openupgrade.logged_query(
+            env.cr,
+            """
+            WITH keys AS (
+                SELECT key
+                FROM ir_ui_view iuv
+                JOIN ir_model_data imd ON (imd.model = 'ir.ui.view'
+                    AND imd.res_id = iuv.id AND module = 'website')
+                WHERE imd.name IN ('cookie_message', 'cookie_message_container')
+            )
+            DELETE FROM ir_ui_view iuv
+            USING keys
+            WHERE iuv.key = keys.key""",
+        )
+        openupgrade.logged_query(
+            env.cr,
+            """
+            DELETE FROM ir_model_data imd
+            WHERE imd.model = 'ir.ui.view' AND module = 'website'
+                AND imd.name IN ('cookie_message', 'cookie_message_container')""",
+        )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    website_cookie_notice_post_migration(env)
--- a/openupgrade_scripts/scripts/website/14.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/website/14.0.1.0/pre-migration.py
@@ -1,0 +1,33 @@
+# Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def website_cookie_notice_pre_migration(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        WITH keys AS (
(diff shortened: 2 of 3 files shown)


openupgrade_scripts/scripts/website/14.0.1.0/post_migration.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
```
pedrobaeza: Incorrect filename. It should be with `-`, not `_`.

openupgrade_scripts/scripts/website/14.0.1.0/post_migration.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def website_cookie_notice_post_migration(env):
+    env.cr.execute(
```
pedrobaeza: Why not using `self.env.ref(..., False)`?

pedrobaeza: And you should remove such views from all the websites.

openupgrade_scripts/scripts/website/14.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def website_cookie_notice_pre_migration(env):
+    openupgrade.logged_query(
```
pedrobaeza, line 7: Why making this? Please always add comments for explaining the why.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/website/14.0.1.0/pre-migration.py
MiquelRForgeFlow: Just renaming xmlids.

pedrobaeza: Why do you need to make these renamings?

MiquelRForgeFlow: Well, they have the same functionality. We always do that, right?

MiquelRForgeFlow: Although the xmlids are noupdate=0, so in fact it doesn't matter, but is not wrong.

pedrobaeza: Yeah, but in this case, renaming is just not so obvious. Starting with, due to the website COW process (create on write), there can be multiple copies of the view with the same key that are really used on the website, so this doesn't make the complete change (you would probably need to remove that copies).

MiquelRForgeFlow:
@pedrobaeza I think it's ready.

openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+
+def website_cookie_notice_post_migration(env):
+    cookie_message = env.ref("website.cookie_message", raise_if_not_found=False)
+    if cookie_message:
+        websites = env["website"].search([])
+        websites.write({"cookies_bar": True})
+        openupgrade.logged_query(
+            env.cr,
+            """
+            WITH keys AS (
```
pedrobaeza, line 14: Why doing this again?

openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+                WHERE imd.name IN ('cookie_message', 'cookie_message_container')
+            )
+            DELETE FROM ir_ui_view iuv
+            USING keys
+            WHERE iuv.key = keys.key""",
+        )
+        openupgrade.logged_query(
+            env.cr,
+            """
+            DELETE FROM ir_model_data imd
```
pedrobaeza, line 28: I think deleting by ORM the view, the COW system deletes the rest.

pedrobaeza (commented):
This must depends on #2840

openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
MiquelRForgeFlow: What do you mean again? here we delete all views that have as key the views of 'cookie_message' and 'cookie_message_container'

MiquelRForgeFlow:
Continue in https://github.com/OCA/OpenUpgrade/pull/2840.

PR #2961, opened 2021-11-10
15.0 openupgrade scripts add scripts

legalsylvain:


StefanRijnhart:
Thanks! I did a run myself to verify and I arrive at exactly the same output. Do you see any model or module renames that we can add to apriory? Perhaps product.packaging -> stock.package.type?

legalsylvain:
Thanks for checking ! Just generated for the time being. not checked the content. Try to make CI green as a first step !

>  Do you see any model or module renames that we can add to apriory?

What is the strategy ? merge a first version of the analysis (this current PR), and then update apriori and regenerate analysis, or try to have complete apriori file ?

> Perhaps product.packaging -> stock.package.type?

I don't get it. (but I don't use package, so I'm not expert of that models...)

1. V14 : [product.packaging](https://github.com/odoo/odoo/blob/14.0/addons/product/models/product.py#L716) exists and``stock.package.type`` doesn't exist.
2. V15 : [product.packaging](https://github.com/odoo/odoo/blob/15.0/addons/product/models/product.py#L755) exists and  [stock.package.type](https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_package_type.py#L7) exists.

In a first sight : 
- does'nt seems exactly a replacement ?
- why the analysis doesn't contains ``new model stock.package.type`` ? It should, don't you think ?

StefanRijnhart:
Ah you are right, the model is forked off here https://github.com/odoo/odoo/commit/1cd0a138cf#diff-ab6db4749bc80b2980c72b5a87dd1593f45e8f00b5255a87b2e4479d6931d1ed but apparently the original model still lives on as well.

The strategy is to add data to apriori immediately if the renames are obvious and if that data cleans up a big chunk of unnecessary changes reported in the analysis. But again, i don't really see anything like that here.

StefanRijnhart (approved):

pedrobaeza:
Let's merge it for starting the flow!

legalsylvain:
> Let's merge it for starting the flow!


I think we should use ``ocabot merge nobump``, to avoid to merge PR doesn't pass CI. Don't you think ?

pedrobaeza:
Not possible right now till the initial setup is finished:

- v14 initial database.
- base migration script

ahmedm0ktar:
can i migrate odoo v10 to 15 ?

Issue #2965, opened 2021-11-13
[13.0] base: ir_model_data forced to noupdate on cases where they shouldn't


pedrobaeza:
@MiquelRForgeFlow you write this specific piece of code for changing NULL values to True on ir_model_data.

https://github.com/OCA/OpenUpgrade/blob/1c5faa0f4a12c6407324576c4c4a63a6f164a051/odoo/addons/base/migrations/13.0.1.3/pre-migration.py#L642

but this is changing the status of ir.model entries for not being updated, which is incorrect. In fact, why NULL is interpreted as True for you? The ORM interprets them as False. Can you please clarify which "true noupdate data" is incorrect?

@Tecnativa TT32835

MiquelRForgeFlow:
It's due to https://github.com/odoo/odoo/pull/32881 if I not remember wrong or another commit. But stefan made a a fix later with https://github.com/OCA/OpenUpgrade/pull/2539, so the correct fix of the query in `fill_ir_model_data_noupdate(env)` should be `WHERE noupdate IS NULL AND model != 'ir.model'`, I think. Should I made the PR?

pedrobaeza:
Yes, please, and remove later fix that is not working at all right now, and put comments about the whys in the code.

pedrobaeza:
Fixed in #2966

PR #2966, opened 2021-11-15
[13.0][FIX] base: correct adapt noupdate=null data

MiquelRForgeFlow:
From https://github.com/OCA/OpenUpgrade/issues/2965.


pedrobaeza (commented):
Can you please update the method docstring for pointing to the pull request/commit that makes this to be applied?

pedrobaeza (approved):

pedrobaeza (approved):

PR #2969, opened 2021-11-15
[11.0][OU-ADD] stock_landed_costs: Migration scripts

pedrobaeza:
@Tecnativa TT30777


pedrobaeza:
I think the error is due to the outdated .travis.yml definition:

https://app.travis-ci.com/github/OCA/OpenUpgrade/builds/241805488#L538

@MiquelRForgeFlow do you remember how to fix it?

MiquelRForgeFlow:
Selection_051

pedrobaeza:
I was about to tell you that I saw it. Pushed the fixed flake8 version.

addons/stock_landed_costs/migrations/11.0.1.1/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+        env.cr, """
+        UPDATE stock_move sm
+        SET landed_cost_value = sub.landed_cost,
+            price_unit = price_unit + sub.landed_cost
+        FROM (
+            SELECT SUM(additional_landed_cost) AS landed_cost, move_id
+            FROM stock_valuation_adjustment_lines
+            GROUP BY move_id
+        ) sub
+        WHERE sub.move_id = sm.id AND sub.landed_cost != 0
```
MiquelRForgeFlow, line 20:
```suggestion
        WHERE sub.move_id = sm.id AND sub.landed_cost != 0.0
```

MiquelRForgeFlow (approved):
LGTM

addons/stock_landed_costs/migrations/11.0.1.1/post-migration.py
pedrobaeza: Why? It performs the same AFAIK.

MiquelRForgeFlow: ok

PR #2973, opened 2021-11-19
[13.0][IMP] account: avoid loop, do all in one query

MiquelRForgeFlow:
With the use of `array_agg` and `unnest`, we now obtain a **great speed optimization**.

In a client database, I have several invoices with a LOT of invoice lines with the same parameters, thus looping many iterations and also every iteration was during more time that the previous one, until arriving to 1 hour per iteration which it was broken.

Now in that database with this query I have:
Selection_074 
which is super acceptable.


pedrobaeza (approved):
Tested on some migrations. Let's avoid one of the ugliest log feeder.

PR #2975, opened 2021-11-22
[13.0][MIG] muk_dms*: Rename to dms.

victoralmau:
Rename addons, from `muk_dms*` to `dms`.

Related to https://github.com/OCA/dms/pull/5

@Tecnativa TT33055

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -97,6 +97,26 @@ merged_modules = {
     'website_logo': 'website',
     'website_megamenu': 'website',
     'website_snippet_anchor': 'website',
+    # muk-it/muk_base - Agreed to move to OCA
+    'muk_security': 'dms',
+    # muk-it/muk_web - Agreed to move to OCA
+    'muk_web_preview': 'mail_preview_base',
+    'muk_web_preview_audio': 'mail_preview_audio',
+    'muk_web_preview_image': 'mail_preview_base',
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
Wrong. You should use merged_modules dict for that purpose.

MiquelRForgeFlow:
I don't know if this is ok, as `muk-it` is not OCA, so theoretically we shouldn't put this here, and you should do in it in your custom migration instead, but let's @pedrobaeza decide.

pedrobaeza:
Yes, in this case, there was an agreement to move it to OCA, so we can provide the migration path from previous open-sourced place to new OCA one.

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -97,6 +97,26 @@
...
     'website_megamenu': 'website',
     'website_snippet_anchor': 'website',
+    # muk-it/muk_base - Agreed to move to OCA
+    'muk_security': 'dms',
+    # muk-it/muk_web - Agreed to move to OCA
+    'muk_web_preview': 'mail_preview_base',
+    'muk_web_preview_audio': 'mail_preview_audio',
+    'muk_web_preview_image': 'mail_preview_base',
+    'muk_web_preview_video': 'mail_preview_base',
+    'muk_web_searchpanel': 'dms',
```
pedrobaeza: This one should be to `web`, as the searchpanel is on core.

pedrobaeza (approved):

PR #2979, opened 2021-11-26
[11.0][FIX] hr_recruitment: Delete the action defined in previous versions.

victoralmau:
Delete the action defined in previous versions.
In v11 Odoo has removed the action from this menu. 
**v11**
https://github.com/odoo/odoo/blob/0eccd3973e056b4ac6fcd6754a8b6af2818aac4f/addons/hr_recruitment/views/hr_recruitment_views.xml#L582-L586

**v10**
https://github.com/odoo/odoo/blob/28c3f51c4878fbcd79b2e819948465fcf2160ebc/addons/hr_recruitment/views/hr_recruitment_views.xml#L549-L552

@Tecnativa TT33144

### DIFF
--- a/addons/hr_recruitment/migrations/11.0.1.0/post-migration.py
+++ b/addons/hr_recruitment/migrations/11.0.1.0/post-migration.py
@@ -32,6 +32,8 @@ def fill_applicant_activities(env):
 @openupgrade.migrate()
 def migrate(env, version):
     fill_applicant_activities(env)
+    # Delete the action defined in previous versions
+    env.ref("hr_recruitment.menu_crm_case_categ0_act_job").action_id = False
     openupgrade.load_data(
         env.cr, 'hr_recruitment', 'migrations/11.0.1.0/noupdate_changes.xml',
     )


pedrobaeza:
On the provided link to the menu, I don't see any action, and it's also of v11, so in any case, it should go to OpenUpgrade 12 instead. Remember also to provide absolute links for avoiding the degradation of the provided info through time.

Maybe the problem is some renamed XML-IDs?

victoralmau:
> On the provided link to the menu, I don't see any action, and it's also of v11, so in any case, it should go to OpenUpgrade 12 instead. Remember also to provide absolute links for avoiding the degradation of the provided info through time.
> 
> Maybe the problem is some renamed XML-IDs?

Sorry, in v10 that action exists, and in v11, it no longer exists.

pedrobaeza:
The links seem exchanged for each version, and please provide such information in the commit message, not only in the PR message.

MiquelRForgeFlow (approved):
LGTM

Issue #2983, opened 2021-11-28
[12.0 -> 13.0] [product] Error while updating product module


flotho:
migration v12 to v13
experiencing same issue as this one :
https://github.com/OCA/OpenUpgrade/issues/2268

Excpet that the message  is indicating that comibination_indices are not computed : 

```**kwargs)
  File "/home/florent.thomas/DEV/Odoo_Server_And_Libs/Servers_13/OpenUpgrade/odoo/sql_db.py", line 250, in execute
    res = self._obj.execute(query, params)
psycopg2.errors.UniqueViolation: ERREUR:  la valeur d'une clé dupliquée rompt la contrainte unique « product_product_combination_unique »
DÉTAIL : La clé « (product_tmpl_id, combination_indices)=(13, ) » existe déjà.
```
AFAICS, the definition of the field has changed https://github.com/OCA/OpenUpgrade/blob/13.0/addons/product/models/product.py#L100 with an explicit name of a relation so that the computation here https://github.com/OCA/OpenUpgrade/blob/13.0/addons/product/models/product.py#L215 always returning an empty combination indices.

I think that the solution should be to rename the relation table in the pre script.
@pedrobaeza , do you agree with this analysis ?

Regards

flotho:
even worse, it is computed in the v12 : https://github.com/OCA/OpenUpgrade/blob/12.0/addons/product/models/product.py#L121

MiquelRForgeFlow:
> Excpet that the message is indicating that comibination_indices are not computed :

No, the message is saying that the are two products which have the same combination_indices, and that a constrain forbids it. Then this error can only happen if the query in calculate_product_product_combination_indices of pre-migration doesn't assure the uniqueness. I hope that with https://github.com/OCA/OpenUpgrade/pull/2987 the problem is solved.

MiquelRForgeFlow:
If with that fix you still have the error, then it happens that you have duplicated products with the same attributes, which you can be solve it with https://github.com/OCA/OpenUpgrade/pull/2513.

MiquelRForgeFlow:
@flotho after doing a deeper analysis of the issue, let me ask you a question. In your v12 db, do you have two variant products of the same template and both without attributes?

I mean, could you do in your DB in v12 the following query and see if it shows any record?

```sql
SELECT pp1.id, pp2.id
FROM product_product pp1
JOIN product_template pt ON pp1.product_tmpl_id = pt.id
[...]

PR #2984, opened 2021-11-28
[13.0][FIX]Product combination indices move logic from post to pre migration

flotho:


flotho:
FIX #2983

flotho:
ping @pedrobaeza

flotho:
and squashed !

pedrobaeza:
Is this superseded by #2987 ?

flotho:
Hi @pedrobaeza , not at all, it doesn't superseed.

What I've discovered is that OpenUpgrade correctly compute the indices and put it on the product_product table.
Yet the Odoo core python logic is based on a computation (https://github.com/OCA/OpenUpgrade/blob/13.0/addons/product/models/product.py#L215) that call a many2many relation ( https://github.com/OCA/OpenUpgrade/blob/13.0/addons/product/models/product.py#L100) that doesn't exist in prior version. As a consequence during the update of the module, if the table is absent it will be created as an empty table and odoo will recompute the indices and use the many2many fields but no attribute_value_ids will be bound so that the indice computed is empty.
That's why I propose to create the table if it doesn't exist and fill it with values in the premigration so that the python logic will compute appropriate datas.
Additionally we could test the fact that if the table exists and is filled then the python logic will do its job properly and setting up the combination indices by SQL will could be irrelevant. I didn't try this for the moment.

pedrobaeza:
Yeah, but we need clear steps to reproduce the problem for knowing the why of the change, and not based on a code reading.

flotho:
superseeded by #3011

PR #2988, opened 2021-11-29
[13.0][MIG] hr_payroll: Move to OCA payroll

victoralmau:
Move to OCA payroll

@Tecnativa TT32606


legalsylvain (approved):

odoo/addons/openupgrade_records/lib/apriori.py
```diff
@@ -120,6 +120,8 @@
     'muk_dms_mail': 'dms',
     'muk_dms_thumbnails': 'dms',
     'muk_dms_view': 'dms',
+    # OCA/payroll
+    'hr_payroll': 'payroll',
```
MiquelRForgeFlow, line 124:
```suggestion
    'hr_payroll': 'payroll',
    'hr_payroll_account': 'payroll_account',
```

MiquelRForgeFlow (changes requested):
Should it be in renamed_modules, right?

In addition, where are the migration scripts for these modules? They are not needed?

odoo/openupgrade/doc/source/modules120-130.rst
```diff
@@ -175,9 +175,9 @@ missing in the new release are marked with |del|.
 +----------------------------------------------+-------------------------------------------------+
 |hr_org_chart                                  | Nothing to do                                   |
 +----------------------------------------------+-------------------------------------------------+
-| |del| hr_payroll                             |                                                 |
+| |del| hr_payroll                             | Replaced by OCA module                          |
```
pedrobaeza: and the other the same

```suggestion
| |del| hr_payroll                             | Replaced by OCA module `payroll`                  |
```

pedrobaeza (approved):

MiquelRForgeFlow:
I think this should be put in `renamed_modules` dict instead, right?

pedrobaeza:
Well, if no conflict, the result is the same, and this way, we avoid possible errors if you already have a module with such name in the previous version, but no strong opinion on it anyways.

pedrobaeza:
@MiquelRForgeFlow, can you make us the favor of generating the analysis file?

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
> can you make us the favor of generating the analysis file?

Ok

PR #2992, opened 2021-12-01
[8.0][FIX] procurement,stock: flake8

MiquelRForgeFlow:
.

### DIFF
--- a/addons/procurement/migrations/8.0.1.0/post-migration.py
+++ b/addons/procurement/migrations/8.0.1.0/post-migration.py
@@ -19,8 +19,8 @@
 #
 ##############################################################################
 
-from openerp.openupgrade import openupgrade, openupgrade_80
-from openerp import pooler, SUPERUSER_ID
+from openerp.openupgrade import openupgrade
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why commenting things?

MiquelRForgeFlow:
Because if not, we have the flake8 F401 error of 'imported but unused'

pedrobaeza:
There's no correspondence of what you are saying with the commented code. They are even calling to external methods of openupgradelib or executing a query...

MiquelRForgeFlow:
> There's no correspondence of what you are saying with the commented code.

Let's see.

> Why commenting things?

The only commented things of the code are:

Selection_128
Here `pool` is commented because is not used.

Selection_129
Here I am not commenting anything, just moving the align.

And the rest of the code is just deleting trailing spaces and unused imports. Where is your issue?

pedrobaeza (approved):
Sorry, my fault. I was seeing the commenting code as the diff, but it's not part of it. It was there before...

PR #2880, opened 2021-08-03
[14.0][MIG] website_sale

Murtaza-OSI:

PR #2994, opened 2021-12-02
[14.0][MIG] website_sale

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2880.


docsource/modules130-140.rst
```diff
@@ -218,7 +218,7 @@ Module coverage 13.0 -> 14.0
 +--------------------------------------------+-------------------------------------------------+
 |l10n_ar                                     |                                                 |
 +--------------------------------------------+-------------------------------------------------+
-| |new| l10n_ar_website_sale                 |                                                 |
+| |new| l10n_ar_website_sale                 | Done                                            |
```
royle-vietnam: @MiquelRForgeFlow Why mark this module as complete? What about the module `website_sale`?

MiquelRForgeFlow: Good catch, I didn't see that.

openupgrade_scripts/scripts/website_sale/14.0.1.0/noupdate_changes.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="product_pricelist_comp_rule" model="ir.rule">
-    <field name="global"/>
+<!--    <field name="global"/>-->
```
pedrobaeza: Comment full record instead for not having an empty write

openupgrade_scripts/scripts/website_sale/14.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,89 @@
...
+
+website_sale / product.style            / name (char)                   : DEL required
+website_sale / product.ribbon           / html (char)                   : NEW required
+# DONE: pre-migration: renamed fields
+
+website_sale / product.style            / html_class (char)             : now required, req_default: function
+# DONE: pre-migration: set default '' in case is NULL
+
+website_sale / product.ribbon           / bg_color (char)               : NEW
+website_sale / product.ribbon           / text_color (char)             : NEW
```
pedrobaeza: Shouldn't this preserved the old colors? What happens if not filled?

openupgrade_scripts/scripts/website_sale/14.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,89 @@
...
+NEW product.ribbon: website_sale.sold_out_ribbon (noupdate)
+NEW website.snippet.filter: website_sale.dynamic_filter_demo_products
+# NOTHING TO DO
+
+DEL product.style: website_sale.image_promo (noupdate)
+NEW product.ribbon: website_sale.sale_ribbon (noupdate)
+# DONE: pre-migration: renamed xmlid (and update values)
+
+DEL product.style: website_sale.image_full (noupdate)
+# DONE: post-migration: deleted safely
```
pedrobaeza: Are you sure you can remove it? The existing m2o has deletion restriction or just `set null`?

openupgrade_scripts/scripts/website_sale/14.0.1.0/openupgrade_analysis_work.txt
MiquelRForgeFlow: These are new fields, new features, for more customization. The old colors were defined by class, so they should still be in the html_class.
Selection_501
https://github.com/odoo/odoo/commit/c058f19bfa33b510addfed314aa27d7e5c7fdb18

pedrobaeza (approved):

PR #2782, opened 2021-06-22
[14.0][ADD] delivery: Migration script.

Chanakya-OSI:

PR #2998, opened 2021-12-03
[14.0][MIG] delivery

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2782.


legalsylvain (commented):
Are you sure to change the policy of the product ?

Otherwise lgtm.

pedrobaeza:
> Are you sure to change the policy of the product ?

In fact, it's not changed at all, as previously was using the default value which is that one. The only possible alteration is if you change the default invoicing policy and then install `delivery` module, so it's good to rewrite it.

PR #2999, opened 2021-12-05
[FIX] base : prevent double membership to 'portal' and 'user' group that is forbidden in V12

legalsylvain:
In v12, a new constrains prevent user to be member for base.group_portal AND base.group_user.
function ``_check_one_user_type`` defined [here](https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_users.py#L396).
this patch fixes that problem in case of the database contains users that belong to both groups. 
the group_user is conserved. (as it is the group with the most of features).

CC : @fcayre

### DIFF
--- a/odoo/addons/base/migrations/12.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/12.0.1.3/pre-migration.py
@@ -170,6 +170,43 @@ def fill_ir_attachment_res_model_name(cr):
         )
 
 
+def fix_double_membership(cr):
+    # avoid error raised by new function '_check_one_user_type'
+
+    # assuming that group_public < group_portal < group_user
+    # this script keept the highest group, if a user belong to many
+    # groups
+    confs = [
+        ("group_public", "group_portal"),
+        ("group_public", "group_user"),
+        ("group_portal", "group_user"),
+    ]
+    for conf in confs:
+        group_to_remove = conf[0]
+        group_to_keep = conf[1]
+        openupgrade.logged_query(
+            cr, """
+                DELETE FROM res_groups_users_rel
+                WHERE
+                gid = (
+                    SELECT res_id
+                    FROM ir_model_data
+                    WHERE module = 'base' AND name = %s
+                )
+                AND uid IN (
+                    SELECT uid FROM res_groups_users_rel WHERE gid IN (
+                        SELECT res_id
+                        FROM ir_model_data
+                        WHERE module = 'base'
+                        AND name IN (%s, %s)
+                    )
+                    GROUP BY uid
+                    HAVING count(*) > 1
+                );
+            """, (group_to_remove, group_to_remove, group_to_keep)
+        )
+
+
 @openupgrade.migrate(use_env=True)
 def migrate(env, version):
     openupgrade.remove_tables_fks(env.cr, _obsolete_tables)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Although this can be considered a bad user configuration, I think it's good to auto-fix it. Please include public group as well, as it is the same selectable.

legalsylvain:
@thanks pedro for your quick review.
indeed, updated to handle the three groups assuming group_public < group_portal < group_user.
Does this seem right to you?

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
```diff
@@ -170,6 +170,43 @@ def fill_ir_attachment_res_model_name(cr):
...
+        group_to_remove = conf[0]
+        group_to_keep = conf[1]
+        openupgrade.logged_query(
+            cr, """
+                DELETE FROM res_groups_users_rel
+                WHERE
+                gid = (
+                    SELECT res_id
+                    FROM ir_model_data
+                    WHERE module = 'base' AND name = '%s'
```
pedrobaeza: `pyscopg2` places the quotes when needed, so remove them (in all places)

```suggestion
                    WHERE module = 'base' AND name = %s
```

legalsylvain: you're right. Fixed. thanks !

pedrobaeza (approved):
Code review

legalsylvain:
thanks. please wait before merging. I'll test the new algorithm in the next upgrade.

fcayre (approved):
Nice to have this in openupgrade instead of fixing this manually in the original data before migrating, thanks.

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
```diff
@@ -170,6 +170,43 @@ def fill_ir_attachment_res_model_name(cr):
...
+                gid = (
+                    SELECT res_id
+                    FROM ir_model_data
+                    WHERE module = 'base' AND name = %s
+                )
+                AND uid IN (
+                    SELECT uid FROM res_groups_users_rel WHERE gid IN (
+                        SELECT res_id
+                        FROM ir_model_data
+                        WHERE module = 'base'
```
MiquelRForgeFlow, line 200:
```suggestion
                        WHERE module = 'base' AND model = 'res.groups'
```

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
```diff
@@ -170,6 +170,43 @@ def fill_ir_attachment_res_model_name(cr):
...
+        group_to_remove = conf[0]
+        group_to_keep = conf[1]
+        openupgrade.logged_query(
+            cr, """
+                DELETE FROM res_groups_users_rel
+                WHERE
+                gid = (
+                    SELECT res_id
+                    FROM ir_model_data
+                    WHERE module = 'base' AND name = %s
```
MiquelRForgeFlow, line 194:
```suggestion
                    WHERE module = 'base' AND model = 'res.groups' AND name = %s
```

MiquelRForgeFlow (approved):

odoo/addons/base/migrations/12.0.1.3/pre-migration.py
legalsylvain: I don't get it. full xml id (module + name) is unique by design.

PR #3000, opened 2021-12-07
[10.0][FIX] base: Merge records through SQL

pedrobaeza:
On this phase of the upgrade, ORM operations are very limited, as most models are not yet loaded, so we need to perform the merge by SQL for avoiding for example missing m2o replacements if the target model is not yet loaded.

Specific problem where I have found the issue is on `product_instrastat` module that adds a country m2o in `account.invoice`, and this one is not changed and when removing the record, the fk constraint is triggered.

@Tecnativa TT30237


pedrobaeza:
Tested and working on the problematic DB.

MiquelRForgeFlow (approved):

pedrobaeza:
Not sure this worths. I'll check this afternoon.

pedrobaeza:
I haved mixed PRs. This one is needed and has been tested.

PR #3002, opened 2021-12-10
[11.0][IMP] stock: Merge quants by SQL

pedrobaeza:
Now that the merge records SQL method covers the values adjustment, it should be faster to use it for this.

@Tecnativa

### DIFF
--- a/addons/stock/migrations/11.0.1.1/end-migration.py
+++ b/addons/stock/migrations/11.0.1.1/end-migration.py
@@ -27,6 +27,7 @@ def merge_quants(env):
             with env.cr.savepoint():
                 openupgrade_merge_records.merge_records(
                     env, 'stock.quant', quants[1:].ids, quants[0].id, QUANT_MERGE_OPS,
+                    method="sql",
                 )
         except ValidationError as error:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Testing in a real migration.

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
But this merge of quants by ORM has never supposed a long process (as far as I saw in the migrations I have done).

pedrobaeza:
Not sure this worths. I'll check this afternoon, but the process can be very huge, specially with lots. I'm finding a lot of errors, so maybe the SQL merge is not completed, but if I have similar times, I won't switch.

pedrobaeza:
Definitively it doesn't worth. I got even higher times (9h07' against 8h55') than with ORM, and there's an operation that is not correct through this way, so closing.

MiquelRForgeFlow:
> and there's an operation that is not correct through this way

which one?

pedrobaeza:
There was an integrity error doing an update, but I didn't take note of it and I don't remember exactly, sorry.

PR #3005, opened 2021-12-11
[13.0][FIX][account]Don't use name from invoice as it doesn't correspond

### LINKED ISSUE #3004: [13.0]Cancel and draft account move bad name
## Module
Account module

## Describe the bug
As you can see, the cancel and draft move (ex invoice) have bad names : 
image

## To Reproduce
migrate from v12 with no validated invoices

**Expected behavior**
Use '/' as default names for those invoices

flotho:
## Module

Account module

## Describe the bug
 
As you can see, the cancel and draft move (ex invoice) have bad names : image

## To Reproduce

migrate from v12 with no validated invoices

**Expected behavior** 

Use '/' as default names for those invoices

FIX #3004

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -190,7 +190,7 @@ def migration_invoice_moves(env):
         invoice_partner_display_name, invoice_cash_rounding_id, old_invoice_id,
         create_uid, create_date, write_uid, write_date)
         SELECT message_main_attachment_id, access_token,
-        COALESCE(NULLIF(number, ''), NULLIF(move_name, ''), NULLIF(name, ''), '/'),
+        COALESCE(NULLIF(number, ''), NULLIF(move_name, ''), '/'),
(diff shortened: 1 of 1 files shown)


flotho:
thanks in advance for your review @MiquelRForgeFlow @pedrobaeza

pedrobaeza:
This can't be absolute, as depending on the type of invoice, each field has different meaning. I remember that `name` field may contain the number in some cases.

flotho:
> This can't be absolute, as depending on the type of invoice, each field has different meaning. I remember that `name` field may contain the number in some cases.

Damned, did you remember which case ?
AFAICS, even the purchase invoices don't use the name field for number : 
image

Yet as you said the name field could reference a number : 
image

I'm not sure at all of the use of this field for purchase orders, any advice for me ?

Thanks for the time spent

pedrobaeza (approved):
I have checked and it seems there's no valid case for feeding the move/invoice name, so OK to this change. The value in such field are other possible things, but not the number.

flotho:
(bow) thanks a lot

PR #3008, opened 2021-12-13
[14.0][FIX] hr: wrong call of delete_records_safely_by_xml_id

MiquelRForgeFlow:
Fast-tracking.

### DIFF
--- a/openupgrade_scripts/scripts/hr/14.0.1.1/post-migrate.py
+++ b/openupgrade_scripts/scripts/hr/14.0.1.1/post-migrate.py
@@ -39,5 +39,5 @@ def migrate(env, version):
         env.cr, "hr", ["mail_template_data_unknown_employee_email_address"]
     )
     openupgrade.delete_records_safely_by_xml_id(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why tests weren't acting?

MiquelRForgeFlow:
Because the delete_records have a try exception. But with https://github.com/OCA/openupgradelib/commit/4c9ecada0dedc0b34ef58c67a1ac37bb89c27d9c, which was added later than this PRs, it made able to detect it.

pedrobaeza:
Maybe we should restrict the `except` part on the openupgradelib?

MiquelRForgeFlow:
Yes, it would be good to restrict by KeyError

pedrobaeza:
In fact I was thinking on the exception when DB constraint is raised (not sure which want it is).

MiquelRForgeFlow:
Then KeyError,ValidationError,IntegrityError, right?

pedrobaeza:
Yeah, I think so

MiquelRForgeFlow:
Restriction done in https://github.com/OCA/openupgradelib/pull/266

PR #3011, opened 2021-12-13
[13.0][FIX] product: move product_variant_combination logic from post to pre-migration

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2984.

### DIFF
--- a/addons/product/migrations/13.0.1.2/pre-migration.py
+++ b/addons/product/migrations/13.0.1.2/pre-migration.py
@@ -192,6 +192,31 @@ def calculate_product_product_combination_indices(env):
     )
 
 
+def create_and_fill_product_variant_combination(env):
+    if not openupgrade.table_exists(env.cr, "product_variant_combination"):
+        openupgrade.logged_query(
+            env.cr, """
+            CREATE TABLE product_variant_combination (
+                product_product_id INTEGER, product_template_attribute_value_id INTEGER)""",
+        )
+        openupgrade.logged_query(
+            env.cr, """
+            INSERT INTO product_variant_combination
+                (product_product_id, product_template_attribute_value_id)
+            SELECT pavppr.product_product_id, ptav.id
+            FROM product_attribute_value_product_product_rel pavppr
+            JOIN product_attribute_value pav
+                ON pav.id = pavppr.product_attribute_value_id
+            JOIN product_product pp
+                ON pp.id = pavppr.product_product_id
+            JOIN product_template_attribute_value ptav
+                ON (ptav.product_attribute_value_id = pav.id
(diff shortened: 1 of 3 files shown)


addons/product/migrations/13.0.1.2/pre-migration.py
```diff
@@ -192,6 +192,31 @@ def calculate_product_product_combination_indices(env):
     )
 
 
+def create_and_fill_product_variant_combination(env):
+    if not openupgrade.table_exists(env.cr, "product_variant_combination"):
+        openupgrade.logged_query(
+            env.cr, """
+            CREATE TABLE product_variant_combination (
+                product_product_id INTEGER, product_template_attribute_value_id INTEGER)""",
```
flotho, line 200: If the table exists, will the ORM rebuild the constraints like there https://github.com/OCA/OpenUpgrade/pull/2984/files#diff-b5b5a877946538c32d839ca4c2c087592ad42245ee44aa37fee846755c3f5b68R199 ?

MiquelRForgeFlow: Yeah, odoo automatically update constraints

pedrobaeza: On v13 that's not true AFAIK.

MiquelRForgeFlow: Yes, it's true, see https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/ir_model.py#L1310. The reflection is done when odoo is initialized, in https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L2510.

pedrobaeza: The methods you are signalling don't say exactly that. Please show me where `_reflect_constraint` is called unconditionally. If I don't remember bad, it was only called when the table is not found.

MiquelRForgeFlow: What do you mean? `_reflect_constraint` is called unconditionally in `_reflect_model` in https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/base/models/ir_model.py#L1383, which is called in `_reflect` in https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L382, which is called in `_auto_init`.

pedrobaeza: Oh, sorry, I was mixing the FK constraints of the m2m. Indeed that ones are always populated.

StefanRijnhart (approved):

flotho:
Thanks a lot for this merge !

PR #3015, opened 2021-12-15
[13.0] stock_account: link account moves to svl

kos94ok-3D:


pedrobaeza (changes requested):
This should be optional and don't get any penalty in case you are not using stock accounting entries generation. You can do an initial query for detecting if any move has been generated this way in the company, and then activate the complicated query, but if not, the previous query must remain.

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -82,24 +83,58 @@ def _prepare_man_svl_vals(price_history_rec, previous_price, quantity, company,
...
+                    AND type = 'entry' AND company_id = %s
+            ) foo
+            WHERE qty = 1
+        )
+        """
+        account_move_id = "rel.id"
+        account_move_join = """
+            LEFT JOIN account_move_rel rel ON rel.create_date = pph.create_date
+        """
+        args = [company_id, product_id] + args
```
MiquelRForgeFlow:
```suggestion
        args = [company_id] + args
```

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -111,12 +146,42 @@ def get_stock_moves(env, company_id, product_id):
...
+        WITH account_move_rel AS (
+            SELECT id, create_date
+            FROM (
+                SELECT id, create_date, COUNT(*) OVER(PARTITION BY create_date) AS qty
+                FROM account_move
+                WHERE stock_move_id IS NULL
+                    AND type = 'entry' AND company_id = %s
+            ) foo
+            WHERE qty = 1
+        )
```
MiquelRForgeFlow:
```suggestion
        WITH account_move_rel AS (
            SELECT id, stock_move_id
            FROM (
                SELECT id, stock_move_id, COUNT(*) OVER(PARTITION BY stock_move_id) AS qty
                FROM account_move
                WHERE stock_move_id IS NOT NULL AND company_id = %s
            ) foo
            WHERE qty = 1
        )
```

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -33,6 +33,7 @@ def _prepare_common_svl_vals(move, product):
         "remaining_value": 0.0,
         "quantity": 0.0,
         "old_product_price_history_id": None,
+        "account_move_id": move["account_move_id"],
```
pedrobaeza: Is this always available or it will give `KeyError` if no real time accounting?

addons/stock_account/migrations/13.0.1.1/post-migration.py
```diff
@@ -111,12 +145,41 @@ def get_stock_moves(env, company_id, product_id):
                  WHEN sl.usage = 'customer' AND sld.usage = 'supplier' THEN 'dropship_return'
                  ELSE 'other'
             END AS move_type
-        FROM stock_move sm LEFT JOIN stock_location sl ON sl.id = sm.location_id
-            LEFT JOIN stock_location sld ON sld.id = sm.location_dest_id
-        WHERE sm.company_id = %s AND sm.product_id = %s
-            AND state = 'done'
+        FROM stock_move sm
```
pedrobaeza: Please reduce the diff for being easier to read. These changes are not needed.

addons/stock_account/migrations/13.0.1.1/post-migration.py
AaronHForgeFlow: I think so, the `def get_stock_moves(` method always return the account move line, even if it is null.

pedrobaeza: And being `NULL`, I suppose that this is translated to `None` by psycopg2. Isn't there any problem with that?

AaronHForgeFlow: I migrated an actual customer with this, and for sure there were stock moves in/out the company without journal entries created. And it did not fail. In any case we can do like this:
```suggestion
        "account_move_id": move_account_move_id,
```
where:
`move_account_move_id = move.get("account_move_id", False)`

Would that be bad for performance?

pedrobaeza: I'm just asking as I don't know it, so no problem at all. Anyways, your code is not impacting performance, except populating the extra variable, that you can put inline:

`"account_move_id": move.get("account_move_id", False),`

pedrobaeza (approved):

PR #3018, opened 2021-12-18
[14.0][OU-FIX] account: Populate properly per company account groups

pedrobaeza:
Previous approach doesn't serve, as current groups populated by localization modules have an XML-ID noupdate=0 that will be tried to be removed on regular update process, leaving main company accounts orphaned of groups. Even further, Odoo expects that the groups populated by the installation of the chart of accounts have a noupdate XML-ID with the scheme `<company_id>_<xmlid_of_the_template_record>`.

With these facts, the approach has been changed to:

* Generate per company account groups with standard population method, as if you were populating them from the CoA template.
* Only unfold and assign proper company for manually created groups.
* Launch the account group computation after of all this.

@Tecnativa TT29943

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -317,75 +317,62 @@ def fill_company_account_cash_basis_base_account_id(env):
     )
 
 
-def fix_group_company_and_create_groups_for_each_company_if_necessary(env):
-    def _get_all_ordered_children(groups):
-        children = groups
-        for child_group in groups:
-            group_childs = env["account.group"].search(
-                [("parent_id", "in", child_group.ids)]
-            )
-            if group_childs:
-                children |= _get_all_ordered_children(group_childs)
-        return children
+def populate_account_groups(env):
+    """Generate the generic account groups for each company. Later code will
+    do it for manually created groups.
+    """
+    companies = env["res.company"].with_context(active_test=False).search([])
+    for company in companies.filtered("chart_template_id"):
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -317,75 +317,62 @@ def fill_company_account_cash_basis_base_account_id(env):
...
-        )
-        companies = accounts.mapped("company_id")
-        if len(companies) == 1:
-            group_plus_childs.filtered(lambda g: g.company_id != companies).write(
-                {"company_id": companies.id}
+def populate_account_groups(env):
+    """Generate the generic account groups for each company. Later code will
+    do it for manually created groups.
+    """
+    companies = env["res.company"].with_context(active_test=False).search([])
```
royle-vietnam, line 324: @pedrobaeza There is no `active` field on the `res.company`, so context `active_test` is not needed

pedrobaeza: It doesn't hurt. You can have this module installed (if migrated to 14.0): https://odoo-community.org/shop/company-active-4952#attr=6890

Even more, such context is propagated downstream, and can serve for other searches inside the scope. And it's a good practice in OpenUpgrade to always pass this context for avoiding to let data unprocessed.

Anyways, that's not the main focus of this PR. What is your problem with this?

royle-vietnam: Thanks for your information, I have no problem

PR #3026, opened 2021-12-22
[14.0][FIX] project: insert distinct project/user combinations

hbrunn:
we need the distinct when a user and their parent company follow a project

### DIFF
--- a/openupgrade_scripts/scripts/project/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/project/14.0.1.1/post-migration.py
@@ -40,7 +40,7 @@ def _fill_res_users_m2m_tables(env):
         """
         INSERT INTO project_allowed_portal_users_rel
         (project_project_id, res_users_id)
-        SELECT pp.id, ru.id
+        SELECT DISTINCT pp.id, ru.id
         FROM project_project pp
         JOIN mail_followers mf ON mf.res_model = 'project.project' AND mf.res_id = pp.id
         JOIN res_partner rp ON mf.partner_id = rp.id


pedrobaeza:
I don't see what you mean, as all relations are 1:1:

- Project had a follower.
- Follower is a partner.
- User has a partner.
- So now project has the users of the previous follower partners.

hbrunn:
- partner has parent, and if both the partner and parent follow some project, the user will be selected twice

We don't need the first distinct indeed, but the second

pedrobaeza:
> partner has parent, and if both the partner and parent follow some project, the user will be selected twice

But there can't be 2 partners for the same user. If you are saying that no need to add "children" contacts, then use `DISTINCT commercial_partner_id`.

hbrunn:
please read the full statement and reconsider

```sql
        INSERT INTO project_allowed_portal_users_rel
        (project_project_id, res_users_id)
        SELECT pp.id, ru.id
        FROM project_project pp
        JOIN mail_followers mf ON mf.res_model = 'project.project' AND mf.res_id = pp.id
        JOIN res_partner rp ON mf.partner_id = rp.id
            OR mf.partner_id = rp.commercial_partner_id
        JOIN res_users ru ON ru.partner_id = rp.id AND ru.share
        WHERE pp.privacy_visibility = 'portal'
```

openupgrade_scripts/scripts/project/14.0.1.1/post-migration.py
```diff
@@ -28,7 +28,7 @@ def _fill_res_users_m2m_tables(env):
         """
         INSERT INTO project_allowed_internal_users_rel
         (project_project_id, res_users_id)
-        SELECT pp.id, ru.id
+        SELECT DISTINCT pp.id, ru.id
```
pedrobaeza: Here there's no commercial partner, and that's why I was confused, as I didn't see the second one. Is this still needed?

hbrunn: no, just removed it

pedrobaeza (approved):

PR #3029, opened 2021-12-27
[13.0][FIX] Ask for unversioned openupgradelib in requirements

joao-p-marques:

@Tecnativa

ping @pedrobaeza


pedrobaeza:
Now that we are versioning very quick, it makes totally sense, so merging.

legalsylvain (commented):
Thansk.

PR #3030, opened 2021-12-28
[14.0][FIX] Ask for unversioned openupgradelib in requirements

joao-p-marques:
FWP of https://github.com/OCA/OpenUpgrade/pull/3029

@Tecnativa

ping @pedrobaeza


pedrobaeza:
Fast-tracking this as stated in the first PR of the series. Reminder: as we now publish `openupgradelib` through a regular quick process, no delay between latest pull requests and the pypi version, so no need to do such thing. On contrary, not doing it, we are restricted for overriding such library with a custom one for testing pull requests.

PR #3031, opened 2021-12-28
[15.0][FIX] Ask for unversioned openupgradelib in requirements

joao-p-marques:
FWP of https://github.com/OCA/OpenUpgrade/pull/3029

@Tecnativa

ping @pedrobaeza


pedrobaeza:
Weird error...

joao-p-marques:
Seems like it has always failed for v15: https://github.com/OCA/OpenUpgrade/actions/workflows/test.yml?query=branch%3A15.0 

Maybe there is still some setup missing...

pedrobaeza:
Oh, right, there's still no CI as the base module + framework is done.

PR #3040, opened 2021-12-31
[MIG][14.0] account_edi: Migration script

tranngocson1996:
This PR
-------
Migration script


openupgrade_scripts/scripts/account_edi/14.0.1.0/upgrade_analysis.txt
```diff
@@ -20,8 +20,6 @@ NEW ir.model.access: account_edi.access_account_edi_document_group_invoice
 NEW ir.model.access: account_edi.access_account_edi_document_readonly
 NEW ir.model.access: account_edi.access_account_edi_format_group_invoice
 NEW ir.model.access: account_edi.access_account_edi_format_readonly
-NEW ir.model.constraint: account_edi.constraint_account_edi_document_unique_edi_document_by_move_by_format
```
legalsylvain: Why did you removed the lines ?

openupgrade_scripts/scripts/account_edi/14.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,5 @@
+---Models in module 'account_edi'---
+---Fields in module 'account_edi'---
```
legalsylvain, line 6: The analysis work file is a copy paste of the analysis file with comment you could add.

legalsylvain (changes requested):
Same remarks for the 3 others PRs.

openupgrade_scripts/scripts/account_edi/14.0.1.0/upgrade_analysis_work.txt
tranngocson1996: > The analysis work file is a copy paste of the analysis file with comment you could add.

@legalsylvain ok thank you bro. I understand.

openupgrade_scripts/scripts/account_edi/14.0.1.0/upgrade_analysis.txt
tranngocson1996: > Why did you removed the lines ?

@legalsylvain Because I didn't notice. I fixed it already. plz review :D

openupgrade_scripts/scripts/account_edi/14.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
...
+account_edi  / account.edi.document     / attachment_id (many2one)      : NEW relation: ir.attachment
+account_edi  / account.edi.document     / edi_format_id (many2one)      : NEW relation: account.edi.format, required
+account_edi  / account.edi.document     / error (html)                  : NEW
+account_edi  / account.edi.document     / move_id (many2one)            : NEW relation: account.move, required
+account_edi  / account.edi.document     / state (selection)             : NEW selection_keys: ['cancelled', 'sent', 'to_cancel', 'to_send']
+account_edi  / account.edi.format       / code (char)                   : NEW required
+account_edi  / account.edi.format       / name (char)                   : NEW
+account_edi  / account.journal          / edi_format_ids (many2many)    : NEW relation: account.edi.format, isfunction: function, stored
+account_edi  / account.move             / edi_document_ids (one2many)   : NEW relation: account.edi.document
+account_edi  / account.move             / edi_state (selection)         : NEW selection_keys: ['cancelled', 'sent', 'to_cancel', 'to_send'], isfunction: function, stored
```
legalsylvain, line 18: Hi. 

thanks for the changes. 

One remarks : 
- ``account_edi`` is a new auto_installable module. (depends on ``account`` module). So, it will be installed on all database that has account moves.
- the field ``edi_state`` (new, stored = True), will be computed. I'm afraid that for large databases the computation will be slow. in that context, adding a ``pre-migration.py`` file that adds the new field with the default ``False`` value could improve the migration duration. 

Don't you think ? 

https://github.com/odoo/odoo/blob/14.0/addons/account_edi/models/account_move.py#L43

kind regards, and thanks for your work.

otherwise, LGTM.

tranngocson1996: @legalsylvain Thank you for the explanation. I learned a lot from you. I will add `pre-migration.py` file

tranngocson1996:
@legalsylvain i fixed it. what do you think?

tranngocson1996:
@legalsylvain Have you checked it for me yet?
cc: @pedrobaeza

docsource/modules130-140.rst
```diff
@@ -20,7 +20,7 @@ Module coverage 13.0 -> 14.0
 +--------------------------------------------+-------------------------------------------------+
 |account_debit_note                          |                                                 |
 +--------------------------------------------+-------------------------------------------------+
-| |new| account_edi                          |                                                 |
+| |new| account_edi                          | Nothing to do                                   |
```
pedrobaeza, line 23: This should be `Done`, as there's a migration script

tranngocson1996: > This should be `Done`, as there's a migration script

@pedrobaeza i will fix it

tranngocson1996:
@pedrobaeza done bro. Check it for me again. Thank

pedrobaeza:
I'm seeing that this module is a new one, and it will be auto-installed, so no migration script will be executed, so we have to turn the approach: instead of this, you should pre-create and fill the column on `account` migration script, commenting that is done for this module, and this one will be `Nothing to do` again.

tranngocson1996:
> I'm seeing that this module is a new one, and it will be auto-installed, so no migration script will be executed, so we have to turn the approach: instead of this, you should pre-create and fill the column on `account` migration script, commenting that is done for this module, and this one will be `Nothing to do` again.

@pedrobaeza ok i understand. I will edit it right away.

pedrobaeza (approved):
Thanks

tranngocson1996:
@pedrobaeza thank you sir. I'm trying to migrate 13 - 14, please support me and forgive the mistakes

pedrobaeza:
@legalsylvain please check your review

legalsylvain (approved):
Thanks pedro ! Indeed if the module was not installed, migration script will not ve executed. Nice catch.

/ocabot merge nobump

tranngocson1996:
Thank @legalsylvain @pedrobaeza

pedrobaeza:
@legalsylvain remember that the ocabot command doesn't serve in the approval comment, but in OU I prefer to manually merge, as there's no advantage in triggering another build.

legalsylvain:
@pedrobaeza : ocabot merge doesn't work with OU ? It was working before. Or I'm wrong ?

pedrobaeza:
I'm not saying that. I quote my text: `there's no advantage in triggering another build.`

legalsylvain:
OK. sorry I missread.

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -330,6 +330,27 @@ def add_move_id_field_account_payment(env):
...
+            ADD COLUMN edi_state varchar
+            """,
+        )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move
+        SET edi_state = false
+        """,
+    )
```
royle-vietnam, line 351: @pedrobaeza no need for this code, because `edi_state` is set as 'false' in database
CC @tranngocson1996

tranngocson1996: PR fix : https://github.com/OCA/OpenUpgrade/pull/3119

pedrobaeza: If nothing is done explicitly, it's set to Null, not False.

royle-vietnam: @pedrobaeza null or '' is not difference, if `SET edi_state = false`, in database will store as `false`, but no key as `false`

pedrobaeza: Isn't this a boolean field?

tranngocson1996: Yes :( char field

royle-vietnam: @pedrobaeza yes
Screenshot from 2022-01-27 17-02-58

pedrobaeza: OK, then indeed this is incorrect. I thought it was a boolean one.

PR #3045, opened 2022-01-06
[MIG][14.0] association: nothing to do

tranngocson1996:
This PR
-------
Nothing to do (it didn't have a migration folder)


legalsylvain:
Migration folder should be present.
Don't you think ? Cc @pedrobaeza, @StefanRijnhart ?
 Could you try to regenerate it, with upgrade analysis ?'otherwise thanks for the work.

tranngocson1996:
> Migration folder should be present. Don't you think ? Cc @pedrobaeza, @StefanRijnhart ? Could you try to regenerate it, with upgrade analysis ?'otherwise thanks for the work.

Thank @legalsylvain. I tried to regenerate it, but it didn't have a migration folder.

pedrobaeza:
When there's no change, no `migrations` folder is generated AFAIK.

tranngocson1996:
> When there's no change, no `migrations` folder is generated AFAIK.

@pedrobaeza I think that right

legalsylvain:
@pedrobaeza : thanks for your answer.

I don't understand the difference between : 
* ``association`` module : no analysis file created.
* ``analytic`` module : an analysis file created that says that nothing changed. [Ref](https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/analytic/14.0.1.1/upgrade_analysis.txt).

any idea ?

pedrobaeza:
Well, you tell me that maintains the `upgrade_analysis` module. I think there's a decision branch that creates nothing depending on the detected things.

legalsylvain:
I don't see it. 
https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/models/upgrade_analysis.py#L245

maybe there is a silent error like this one : 
https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/models/upgrade_analysis.py#L101

pedrobaeza:
I can't say, as I have never dug enough on such module.

legalsylvain:
I just meld association manually.
nothing changed.

/ocabot merge nobump

PR #3052, opened 2022-01-07
[MIG][14.0] crm_iap_lead_enrich: migration script

tranngocson1996:
This PR
-------
Migration done


tranngocson1996:
@legalsylvain PR has split
cc: @pedrobaeza

openupgrade_scripts/scripts/crm_iap_lead_enrich/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,6 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env.ref("crm_iap_lead_enrich.mail_message_lead_enrich_with_data").unlink()
```
MiquelRForgeFlow:
```suggestion
    openupgrade.delete_records_safely_by_xml_id(
        env, ["crm_iap_lead_enrich.mail_message_lead_enrich_with_data"]
    )
```

tranngocson1996: @MiquelRForgeFlow Done, but why i still see many modules using `unlink`

MiquelRForgeFlow: which modules?

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/crm_iap_lead_enrich/14.0.1.1/post-migration.py
tranngocson1996: Looks like I misread v13. Sr

Issue #3055, opened 2022-01-07
[13.0 --> 14.0] account Invoice payment widget does not show payment amounts


remi-filament:
## Module
account

## Describe the bug
After migration, when opening a paid invoice, the payment widget at the bottom shows only date of payment but not the amount.

After a bit of investigation, it is because new fields debit_amount_currency and credit_amount_currency added in v14.0 on account.partial.reconcile model are empty (these are usually filled during reconciliation with the following functions from account_move : reconcile() > _prepare_reconciliation_partials() but the old lines are not updated).

The solution should probably be to calculate these 2 fields as done in [_prepare_reconciliation_partials() function.](https://github.com/odoo/odoo/blob/077e554703b6d8a71c81fbf6783228d04e77d96f/addons/account/models/account_move.py#L4366)

As we have only 1 currency and we are not dealing with multi currencies we worked around that issue with simple SQL command (but more elaborated solution should be found for multi_currency contexts) :
```sql
update account_partial_reconcile
set 
    debit_amount_currency = amount,
    credit_amount_currency = amount
where debit_amount_currency is null;
```

## To Reproduce
**Affected versions**: migration to version 14.0

1. Get a database in v13.0
2. Create invoice and associated payments
3. Migration to v14.0
4. Check paid invoices : the amounts are not there on payment widget at the bottom (between total price and due amount)

**Expected behavior**
Display properly all payments with both dates and amounts

remi-filament:
Split from #3050  as requested by @pedrobaeza

pedrobaeza:
This sounds like you can do a pull request adding such SQL in the migration scripts ;)

remi-filament:
Yes I could but I think this will not work for multi-currency environment, so I am not sure whether we should add this in script or if it should be made more generic before ?

pedrobaeza:
You can add a `WHERE` clause for only those that match the currency, and with that, you fill proper data for a good percentage (or 100%) of the reconcilations. Put a TODO or try to develop the other case.

remi-filament:
Good idea thanks, will do !

PR #3058, opened 2022-01-07
[14.0][FIX] account.partial.reconcile compute missing fields

remi-filament:
Compute debit_currency_amount and credit_currency_amount for account.partial.reconcile table.

See #3055

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -91,13 +91,13 @@ def fill_partial_reconcile_debit_and_credit_amounts(env):
         .filtered(lambda line: line.credit_currency_id != line.debit_currency_id)
     )
     for line in partial_reconcile_lines:
-        line.debit_amount_currency = line.credit_currency_id._convert(
+        line.debit_amount_currency = line.company_currency_id._convert(
             line.amount,
             line.debit_currency_id,
             line.company_id,
             line.credit_move_id.date,
         )
-        line.credit_amount_currency = line.debit_currency_id._convert(
+        line.credit_amount_currency = line.company_currency_id._convert(
             line.amount,
             line.credit_currency_id,
             line.company_id,
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -68,6 +68,19 @@ def fill_payment_id_and_statement_line_id_fields(env):
     )
 
 
+def fill_partial_reconcile_debit_and_credit_amounts(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_partial_reconcile
+        SET debit_amount_currency = amount, credit_amount_currency = amount
+        WHERE debit_amount_currency IS NULL AND credit_amount_currency is NULL
+            AND credit_currency_id = debit_currency_id
+       """,
+    )
+    # TODO: compute debit and credit amount when currencies are different
+
+
 def create_account_reconcile_model_lines(env):
     openupgrade.logged_query(
         env.cr,
@@ -728,6 +741,7 @@ def migrate(env, version):
     fill_code_prefix_end_field(env)
     fill_default_account_id_field(env)
     fill_payment_id_and_statement_line_id_fields(env)
+    fill_partial_reconcile_debit_and_credit_amounts(env)
     create_account_reconcile_model_lines(env)
     create_account_reconcile_model_template_lines(env)
     create_account_tax_report_lines(env)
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -68,6 +68,19 @@ def fill_payment_id_and_statement_line_id_fields(env):
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_partial_reconcile
+        SET debit_amount_currency = amount, credit_amount_currency = amount
+        WHERE debit_amount_currency IS NULL AND credit_amount_currency is NULL
+            AND credit_currency_id = debit_currency_id
+       """,
+    )
+    # TODO: compute debit and credit amount when currencies are different
```
legalsylvain: Maybe you could make a request to check if the different currency caqe exists.
If yes, log an error.
Not a blocking point.

legalsylvain (approved):
Thanks for taking time to share your bugfixes.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -68,6 +68,19 @@ def fill_payment_id_and_statement_line_id_fields(env):
...
 
 
+def fill_partial_reconcile_debit_and_credit_amounts(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_partial_reconcile
+        SET debit_amount_currency = amount, credit_amount_currency = amount
+        WHERE debit_amount_currency IS NULL AND credit_amount_currency is NULL
+            AND credit_currency_id = debit_currency_id
```
MiquelRForgeFlow, line 79: if both are null, why compare if they have same value? something is wrong here

pedrobaeza: Maybe you are mixing field names? `debit_amount_currency` is not the same as `debit_currency_id`.

MiquelRForgeFlow: oh my god, right! hahaha

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -68,6 +68,43 @@ def fill_payment_id_and_statement_line_id_fields(env):
...
+            line.debit_currency_id,
+            line.company_id,
+            line.credit_move_id.date,
+        )
+        line.credit_amount_currency = line.debit_currency_id._convert(
+            line.amount,
+            line.credit_currency_id,
+            line.company_id,
+            line.debit_move_id.date,
+        )
```
MiquelRForgeFlow:
```suggestion
        line.debit_amount_currency = line.company_currency_id._convert(
            line.amount,
            line.debit_currency_id,
            line.company_id,
            line.debit_move_id.date,
        )
        line.credit_amount_currency = line.company_currency_id._convert(
            line.amount,
            line.credit_currency_id,
            line.company_id,
            line.credit_move_id.date,
        )
```

remi-filament: Hi @MiquelRForgeFlow 

I can modify as you propose (I suppose this should be the same, except maybe for date ?), but I put it there on purpose since the lines that normally compute these fields are as follow (taken from [here](https://github.com/odoo/odoo/blob/077e554703b6d8a71c81fbf6783228d04e77d96f/addons/account/models/account_move.py#L4456)) :
```python
  min_debit_amount_residual_currency = credit_line.company_currency_id._convert(
      min_amount_residual,
      debit_line.currency_id,
      credit_line.company_id,
      credit_line.date,
  )
  min_credit_amount_residual_currency = debit_line.company_currency_id._convert(
      min_amount_residual,
      credit_line.currency_id,
      debit_line.company_id,
      debit_line.date,
  )
```
You can see there that there is on purpose a mix between debit and credit.
Let me know if you still prefer to go your way ?

MiquelRForgeFlow: > I suppose this should be the same, except maybe for date ?

Yeah :)

Then it should be:

```suggestion
        line.debit_amount_currency = line.company_currency_id._convert(
            line.amount,
            line.debit_currency_id,
            line.company_id,
            line.credit_move_id.date,
        )
        line.credit_amount_currency = line.company_currency_id._convert(
            line.amount,
            line.credit_currency_id,
            line.company_id,
            line.debit_move_id.date,
        )
```

MiquelRForgeFlow: AFAIK line.amount is always in line.company_currency_id, thus that's why I think it should should be this way.

remi-filament: Thanks, makes sense then !

MiquelRForgeFlow (approved):
LGTM

JulianaPoudou (approved):

PR #3059, opened 2022-01-07
[14.0][FIX] account auto link statement with move

### LINKED ISSUE #3057: [13.0 --> 14.0] account bank statement lines linked to incorrect account moves
## Module
account

## Describe the bug
An account move is now mandatory for every account_bank_statement_line (since v14.0). The migration script tries to find which existing account_move should be linked to account_bank_statement_line (in pre-migration part) and the create the missing ones (in post-migration part) but we had a few issues that led to have multiple account_bank_statement_line linked to the same account_move.
These issues seemed to be caused by the following :

- we had different account_move with same name
- we had various account_bank_statement_line with same account_number and various account_move with same ref

Not sure why, or where it came from (probably v9 or v10 according to the dates) but it caused quite a mess after migration :

- various account_bank_statement_lines got linked to the same account_move
- when one of these account_move was modified, all related account_bank_statement_lines where modified as well because of _synchronize_business_models() function from account_move called on write() that caused all these bank statement lines to get wrong amounts and partners...

As far as I could tell, we should update pre-migration script to replace the SQL command from function add_move_id_field_account_bank_statement_line() on lines 214 - 225 in 2 SQL commands to first link the account_bank_statement_line to account_move.statement_line_id, then do the rest of the command for account_bank_statement_line which do not have move_id.
I should propose a PR on this next week but here is the code for the time being :

```python
    # Force move ids when statement_line_ids already set on account_move
        openupgrade.logged_query(
            env.cr,
            """
            UPDATE account_bank_statement_line absl
            SET move_id = am.id
            FROM account_move am
            WHERE am.statement_line_id = absl.id
            """,
        )
    # Find move_id by using either move_name or payment_ref from account_bank_statement_line
        openupgrade.logged_query(
            env.cr,
            """
            UPDATE account_bank_statement_line absl
            SET move_id = am.id
            FROM account_move am, account_bank_statement bs
            WHERE absl.statement_id = bs.id 
            AND absl.move_id IS NULL
            AND am.name NOT IN ('', '/')
            AND COALESCE(NULLIF(absl.move_name, ''), absl.payment_ref) = am.name
            AND am.company_id = bs.company_id
            """,
        )
```

## To Reproduce
**Affected versions**: migration to v14.0

remi-filament:
Fix issues related to link between statement line and account moves.

See #3057

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
@@ -211,19 +211,31 @@ def add_move_id_field_account_bank_statement_line(env):
         WHERE aml.move_id = am.id AND aml.statement_line_id IS NOT NULL
         """,
     )
+    # 1. set move_id from move_ids where statement_line_id is defined
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line absl
+        SET move_id = am.id
+        FROM account_move am
+        WHERE am.statement_line_id = absl.id
+        """,
+    )
+    # 2. try to match on statement move_name or payment_ref (if move_name empty)
     openupgrade.logged_query(
         env.cr,
         """
         UPDATE account_bank_statement_line absl
         SET move_id = am.id
         FROM account_move am, account_bank_statement bs
-        WHERE absl.statement_id = bs.id AND (
-                (am.name NOT IN ('', '/') AND COALESCE(
-                    NULLIF(absl.move_name, ''), absl.payment_ref) = am.name
-                AND am.company_id = bs.company_id)
-            OR am.statement_line_id = absl.id)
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -211,19 +211,31 @@ def add_move_id_field_account_bank_statement_line(env):
         WHERE aml.move_id = am.id AND aml.statement_line_id IS NOT NULL
         """,
     )
+    # 1. set move_id from move_ids where statement_line_id is defined
```
MiquelRForgeFlow:
```suggestion
    # 1. set move_id from moves where statement_line_id is defined
```

MiquelRForgeFlow (approved):

PR #3062, opened 2022-01-10
[14.0][OU-ADD] event_sale: Migration scripts

tranngocson1996:
This PR
-------
Migration done


openupgrade_scripts/scripts/event_sale/14.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,11 @@
+from openupgradelib import openupgrade
```
MiquelRForgeFlow: delete this file

openupgrade_scripts/scripts/event_sale/14.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,73 @@
...
+NEW ir.ui.view: event_sale.event_type_ticket_view_form_from_type
+NEW ir.ui.view: event_sale.event_type_ticket_view_tree_from_type
+DEL ir.ui.view: event_sale.event_ticket_form_view
+DEL ir.ui.view: event_sale.event_type_view_form_inherit_sale
+DEL ir.ui.view: event_sale.view_event_registration_ticket_form
+DEL ir.ui.view: event_sale.view_event_registration_ticket_search
+# NOTHING TO DO: noupdate="0" records, no name changed
+
+product.product: event_sale.product_product_event (noupdate switched)
+# DONE: pre-migration: switched to noupdate = "1"
```
MiquelRForgeFlow:
```suggestion
# NOTHING TO DO: unknown error in analysis (this xmlid is noupdate=True both in v13 and v14)
```

MiquelRForgeFlow (commented):
Also in the noupdate_changes.xml file, I think you should comment the line of the default_code, because when one migrates, it's expected that the default_code is still applied.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #3083, opened 2022-01-11
[12.0][ADD] apriori: account_analytic_asset -> account_asset_management

MiquelRForgeFlow:
Wrong module name in https://github.com/OCA/OpenUpgrade/commit/d2643fff5e421caeac1e958529ce1a512fd4d57c.

See https://github.com/OCA/account-analytic/tree/11.0.

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -60,7 +60,8 @@ merged_modules = {
     # OCA/account-analytic
     # although model is defined in "analytic", logic is in "account"
     'account_analytic_distribution': 'account',
-    'account_asset_analytic': 'account_asset_management',
+    'account_analytic_asset': 'account_asset_management',
+    'account_asset_analytic': 'account_asset_management',  # (from <= v10)
     # OCA/account-financial-reporting
     'customer_activity_statement': 'partner_statement',
     'customer_outstanding_statement': 'partner_statement',


legalsylvain (approved):
thanks !

pedrobaeza:
It's not wrong... https://odoo-community.org/shop/product/account-asset-analytic-78

legalsylvain:
> It's not wrong... https://odoo-community.org/shop/product/account-asset-analytic-78

@pedrobaeza : your link is about the V8 / V10 version.

@MiquelRForgeFlow is talking about migration from 11.0 to 12.0.
In V11, it **seems** that the name is indeed : ``account_analytic_asset``.

don't you think ?

pedrobaeza:
As there's no migration path from v10 `account_asset_analytic` to v12 module, it's OK to put that one there. The one for Miquel should be an extra entry.

MiquelRForgeFlow:
I have compared both modules, and I think both are the same. Maybe I should add a module rename in v11. Is it ok for you, @pedrobaeza ?

pedrobaeza:
And the fields are the same? Isn't better simply to add both modules here?

MiquelRForgeFlow:
No, the fields are different, so they would need a custom script. Yeah, it's faster do the merge here (there will not be "continuation" between v10 and v11, but is not my problem).

pedrobaeza:
Then please add both here

pedrobaeza (approved):

pedrobaeza:
To be merged when finished.

PR #3084, opened 2022-01-12
[MIG][14.0] mass_mailing_sms: nothing to do

tranngocson1996:
This PR
-------
Nothing to do


tranngocson1996:
@legalsylvain Do you have any idea why the test failed? It appears to be strange.

PR #3086, opened 2022-01-12
[13.0][FIX] account restrict_mode_hash_table more restrictive

remi-filament:
Limit restrict_mode_hash_table to non cash/bank journals since this causes issues afterwards in v14 when bank journals are updated by Odoo during reconciliation (with the use of suspense account until reconciliation)

Also in v14, these fields are not displayed for bank / cash account journal types : https://github.com/odoo/odoo/blob/04396e0627fd36d28b047e0e122ffb55a3a12ebc/addons/account/views/account_journal_views.xml#L105

and you cannot remove this boolean if you have posted moves for this journal afterwards (the only way is to remove it directly in DB), so I feel it would be safer not to set this field at all in migration between v12 and v13

### DIFF
--- a/addons/account/migrations/13.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/account/migrations/13.0.1.1/openupgrade_analysis_work.txt
@@ -301,7 +301,7 @@ account      / account.journal          / post_at_bank_rec (boolean)    : DEL
 
 account      / account.journal          / update_posted (boolean)       : DEL
 account      / account.journal          / restrict_mode_hash_table (boolean): NEW
-# DONE: post-migration: filled restrict_mode_hash_table reusing update_posted
+# NOTHING TO DO: probably better to leave this field to be set directly after migration (see https://github.com/OCA/OpenUpgrade/pull/3086)
 
 account      / account.journal          / secure_sequence_id (many2one) : NEW relation: ir.sequence
 account      / account.journal          / website_message_ids (one2many): NEW relation: mail.message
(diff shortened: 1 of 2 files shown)


remi-filament:
@MiquelRForgeFlow @pedrobaeza what do you think about that one ?

MiquelRForgeFlow:
> Limit restrict_mode_hash_table to non cash/bank journals since this causes issues afterwards in v14 when bank journals are updated by Odoo during reconciliation (with the use of suspense account until reconciliation)

which issues?

remi-filament:
@MiquelRForgeFlow : when you try to reconcile from bank statement, you get an the following UserError : 
`You cannot edit the following fields due to restrict mode being activated on the journal: debit, credit, account_id, partner_id.`

MiquelRForgeFlow (approved):
The narrow field `update_posted` in v12 was to allow cancelling entries (and invoices) and maybe it was not used in the reconciliation. Thus fine by me.

pedrobaeza:
In fact, I would say that no journal should be marked with that, as that lock is too much and it's not equivalent to previous behavior. What do you think?

remi-filament:
I agree with you @pedrobaeza these could be set manually after migration.
We probably should not assimilate not allowing cancellation of invoices (before v13) with blocking all moves and move_lines with hash - taking into account that this is not easily revestible (on v13+)

pedrobaeza:
OK, then please totally remove such lines.

MiquelRForgeFlow:
and update analysis_work file

MiquelRForgeFlow:
Please squash

JulianaPoudou (approved):

PR #3088, opened 2022-01-12
[14.0][FIX] account recompute amount_residual_currency

### LINKED ISSUE #3087: [14.0] account amount_residual_currency not recomputed
## Module
account

## Describe the bug
The field amount_residual_currency is normally computed on account_move_line in [_compute_amount_residual() function](https://github.com/odoo/odoo/blob/c68e1d1c13d116bda6b91fcbbcd61a26739a9894/addons/account/models/account_move.py#L3850)

This compute depends on the currency_id field hat needs to be defined.

This currency_id existed since a while in Odoo, but is required only since v14. It is filled in [pre-migration script](https://github.com/OCA/OpenUpgrade/blob/8a5c24145c8ea11f08454273c46186ce94c1d9d6/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py#L389), but the amount_residual_currency is not recomputed for existing lines since this change is made in SQL and not via ORM.

This causes issues when reconciling bank statement with invoices : the invoices change to Paid payment_state but you cannot see the related payment. This is because the [_prepare_reconciliation_partials() function ](https://github.com/odoo/odoo/blob/c68e1d1c13d116bda6b91fcbbcd61a26739a9894/addons/account/models/account_move.py#L4366) called from reconcile() one is using this amount_residual_currency field to complete the reconciliation. Without this field computed, the lines are not properly reconciled (it will actually not create partial_reconcile entries since exiting at lines 4430 or 4437, and afterwards not create full_reconcile)

## To Reproduce
**Affected versions**: 14.0

1. Get a version prior to v14 where currency_id is not filled in account_move_line
2. Migrate to v14.0
3. Try to reconcile a bank statement line with open invoice
4. Invoice is paid but no payment is listed

**Expected behavior**
Payment should be listed in invoice

The solution is easy : we should recompute amount_residual_currency, by running "account.move.line"._compute_amount_residual()
I could propose a PR but I am not sure where to add this, should it go to pre-migration script, in fill_account_move_line_currency_id() and be run only for the lines that were modified, or should it rather go in post-migration script ?

remi-filament:
This should fix #3087 

Let me know if this should not be done this way !

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -709,6 +709,24 @@ def fill_account_move_line_amounts(env):
             aml.debit + aml.credit > 0 AND (
                 aml.amount_currency = 0 OR aml.amount_currency IS NULL)""",
     )
+    # Recompute residual amounts
+    # This may need to be optimized further
+    # see https://github.com/OCA/OpenUpgrade/pull/3088
+    lines = (
+        env["account.move.line"]
+        .search(
+            [
+                ("full_reconcile_id", "=", False),
+                ("account_internal_type", "in", ("payable", "receivable")),
+                ("parent_state", "=", "posted"),
+                ("balance", "!=", 0),
+            ]
+        )
+        .filtered(lambda line: line.currency_id == line.company_currency_id)
+    )
+    lines.with_context(
+        tracking_disable=True, skip_account_move_synchronization=True
+    )._compute_amount_residual()
 
 
 def fill_account_move_line_date(env):
--- a/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
@@ -215,7 +215,7 @@ account      / account.move.line        / analytic_tag_ids (many2many)  : now a 
 # NOTHING TO DO: filled by Odoo
 
 account      / account.move.line        / currency_id (many2one)        : now required
-# DONE: filled with company currency if empty
+# DONE: filled with company currency if empty and recalculate amount residual when set
 
 account      / account.move.line        / matching_number (char)        : NEW isfunction: function, stored
 # NOTHING TO DO


openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -400,8 +400,12 @@ def fill_account_move_line_currency_id(env):
         UPDATE account_move_line
         SET currency_id = company_currency_id
         WHERE currency_id IS NULL
+        RETURNING id
         """,
     )
+    ids = [x[0] for x in env.cr.fetchall()]
+    if ids:
+        env['account.move.line'].browse(ids)._compute_amount_residual()
```
legalsylvain: I thought that orm function was not available in the pre-migration script.
Or did I missed something ?

MiquelRForgeFlow: You are correct, all account models are not in the environment during pre-migration.

legalsylvain: thanks for your review.

hum don't know how to implement the @remi-filament needs.

sql request should be done in pre-migration script.
call to _compute_amount_residual should be done on post-migration script but only on lines impacted by the request of the pre-migration script.
we could add a temporary extra column to_recompute_amount_residual, but it seems a little bit tricky. Do you have another idea ?

MiquelRForgeFlow: Yeah, the temporary column it's the best trick. later in post you have to delete that column.

remi-filament:
Maybe we do not need to limit to these lines, since recompute should not change fields which are correctly set.
What do you think if we go with something like this in post-migration :
```python
    lines = env["account.move.line"].search([('full_reconcile_id','=',False),('account_internal_type','in',('payable','receivable')),('parent_state','=','posted'),('balance','!=',0)])
    lines._compute_amount_residual()
```

MiquelRForgeFlow:
maybe you could also add `.filtered(lambda line: line.currency_id == line.company_currency_id)`

legalsylvain:
> Maybe we do not need to limit to these lines, since recompute should not change fields which are correctly set.

Well the only problem i see is to increase migration time uselessly, recomputing good values. But it depends on domain and datas. Maybe you could implement a first try with what you propose and add a link to this issue to mention that improvment could be done.

remi-filament:
I get an error now, but I am not sure what is going wrong...

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -709,6 +709,22 @@ def fill_account_move_line_amounts(env):
...
+            [
+                ("full_reconcile_id", "=", False),
+                ("account_internal_type", "in", ("payable", "receivable")),
+                ("parent_state", "=", "posted"),
+                ("balance", "!=", 0),
+            ]
+        )
+        .filtered(lambda line: line.currency_id == line.company_currency_id)
+    )
+    lines._compute_amount_residual()
```
MiquelRForgeFlow:
```suggestion
    lines.with_context(tracking_disable=True, skip_account_move_synchronization=True)._compute_amount_residual()
```

remi-filament: Thanks, just implemented it !

MiquelRForgeFlow (approved):

JulianaPoudou (approved):

pedrobaeza:
Can you look for an SQL alternative for performance reasons?

remi-filament:
I do not really know how to do that... Any help would be appreciated !

MiquelRForgeFlow:
@pedrobaeza FYI I have been migrating with this patch, and I didn't have any performance issues.

pedrobaeza:
In one DB with 32k applicable lines, it has taken about 30 minutes, which is a lot.

StefanRijnhart:
I'm now looking at the account module in 14.0 for the first time, so maybe I'm missing something, but if the values are only missing in case currency_id is equal to the company currency, isn't it the case that `amount_residual_currency` is equal to the move line's `amount_residual` which is already present in the 13.0 database? If that is indeed the case, the operation can be executed with a simple and performant SQL statement.

~~Do we also need to set amount_currency to the balance of the move line?~~ Ah, this is done in the post-script: https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py#L757

pedrobaeza:
Fixed in #3234, and Stefan, indeed it can be done very simple.

PR #3089, opened 2022-01-12
[13.0][OU-FIX] product: migrate pricelist settings

MiquelRForgeFlow:
Due to https://github.com/odoo/odoo/commit/7d8df4bec192e5a2ff3341dcf43d59e06a57f9a9.

Not detected before because the changes reused the same groups :S

### DIFF
--- a/addons/product/migrations/13.0.1.2/post-migration.py
+++ b/addons/product/migrations/13.0.1.2/post-migration.py
@@ -65,6 +65,67 @@ def empty_template_pricelist_company(env):
         )
 
 
+def adapt_pricelist_settings(env):
+    """In v12, if pricelist setting was marked it implied group_sale_pricelist,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you explain a bit the changes?

MiquelRForgeFlow:
Yeah. In v12, pricelist setting marked implied group_sale_pricelist, and if the setting was 'percentage' it also had group_product_pricelist. And in v13, pricelist setting marked implies group_product_pricelist, and if the setting is 'advanced', then it has also group_sale_pricelist. And about the options, 'percentage' has become 'basic' and 'formula' has become 'advanced'. So this scripts does the adaptation.

pedrobaeza:
OK, thanks for the explanation. Please add it to the commit message and as docstring of the new method.

pedrobaeza (approved):

Issue #3061, opened 2022-01-09
[14.0] openupgrade_framework: New installed modules dependencies are not auto-resolved


pedrobaeza:
1. Have a 13.0 DB with `account_facturx` installed.
2. Migrate it to 14.0
3. The rename to `account_edi_facturx` is done, but the new dependency `account_edi` is not installed, and thus, the module remains not usable.

**Expected behavior**
Previous framework cycled the needed times over pending to install/to upgrade modules until no one is left or no possible candidate is found, and new dependencies that are not installed, were marked to be installed on the next cycle. This seems to not be replicated on the new framework.

**Current workaround**
Install manually `account_edi` after the migration is done.

legalsylvain:
Hi @pedrobaeza thanks for reporting !

> previous framework cycled the needed times over pending to install/to upgrade modules 

Are you talking about odoo or openupgrade framework ?
I don't remember to have removed such code when we isolated patches during the openupgrade refactor with @StefanRijnhart.
Maybe something changed in the core. In a non openupgrade framework, if a new dependency is done on an installed module, making an update install the missing dependency afaik.

Unfortunately, I have no time to investigate immediately on that topic. 
Stefan any idea ?

Regards.

pedrobaeza:
It was something specific of OpenUpgrade, touching on `loading.py` if I don't remember bad.

MiquelRForgeFlow:
One question. After doing the migration, if you do an --update=all with odoo or OCB, should that module get installed, right?

pedrobaeza:
Not sure, but that will be too late, as the migration scripts of the affected modules won't be run until them, and then there can be already data loss or lack of the rest of the framework.

StefanRijnhart:
I see the linked PR was closed as "undesirable to solve this in a module's migration script" (with which I agree). Is anybody already chasing a framework solution? Otherwise, I'll gladly have a look this week.

pedrobaeza:
I think there isn't anybody, but let's wait for any other answer.

MiquelRForgeFlow:
Nobody. Feel free @StefanRijnhart

StefanRijnhart:
[...]

PR #3094, opened 2022-01-14
[IMP][14.0] install unmet dependencies

royle-vietnam:
When changing the dependency of module which is not installed on the target version, this PR to install the missing modules to avoid error unmet_dependencies

ex.
on 13.0, module A that directly depend on the module B
on 14.0, module A that directly depend on the new module C, which is not installed on 14.0

Current behavior before PR: Unmet dependencies

Expect: To need to install the new module C before running migrations script


tranngocson1996 (approved):

pedrobaeza:
This shouldn't be resolved this way, but by the framework. See #3061

PR #3095, opened 2022-01-14
[14.0][FIX] account: make use of enumerate

MiquelRForgeFlow:
Oversight of https://github.com/OCA/OpenUpgrade/pull/3018.


pedrobaeza (approved):

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -344,7 +344,7 @@ def unfold_manual_account_groups(env):
         accounts = env["account.account"].search([("group_id", "=", group.id)])
         companies = accounts.mapped("company_id").sorted()
         relation_dict = {}
-        for i, company in companies:
+        for i, company in enumerate(companies):
```
legalsylvain, line 347: hi. naive question : what is the interest of this syntax ? 
thanks !

PR #3099, opened 2022-01-14
[14.0][FIX] base: New installed modules dependencies are not auto-resolved

MiquelRForgeFlow:
This should do the trick and solve https://github.com/OCA/OpenUpgrade/issues/3061.

Note: @pedrobaeza I know you prefer to have this incorporated in the framework. But this is a first step :)


royle-vietnam (commented):
This way is similar to PR https://github.com/OCA/OpenUpgrade/pull/3094/files, but the code is quite long

PR #3107, opened 2022-01-17
[MIG][14.0] website_event_track: migration script

tranngocson1996:
This PR
-------
Migration done


openupgrade_scripts/scripts/website_event_track/14.0.1.1/post-migration.py
```diff
@@ -0,0 +1,13 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        ["website_event_track.event_type_data_tracks"],
+    )
+
+    openupgrade.load_data(
+        env.cr, "website_event_track", "14.0.1.1/noupdate_changes.xml"
+    )
```
MiquelRForgeFlow: usually the load is done before the delete, but I don't know if it really matters 😆

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/website_event_track/14.0.1.1/post-migration.py
tranngocson1996: @MiquelRForgeFlow I think it's no problem.
When deleting that record then load the file no_update at v14. The external id is no longer there, so a new record will not be created.
But I still change it back to look like a normal person 😄

RaulOForgeFlow (approved):
LGTM

PR #3108, opened 2022-01-17
[MIG][14.0] website_event_questions: migration script

tranngocson1996:
This PR
-------
Migration done


openupgrade_scripts/scripts/website_event_questions/14.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,27 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_tables(
+        env.cr,
+        [("event_answer", "event_question_answer")],
+    )
```
MiquelRForgeFlow, line 11: also should call rename_models

openupgrade_scripts/scripts/website_event_questions/14.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,62 @@
...
+website_event_questions / event.question           / question_type (selection)     : NEW required, selection_keys: ['simple_choice', 'text_box'], req_default: function, hasdefault
+# NOTHING TO DO
+
+website_event_questions / event.registration       / answer_ids (many2many)        : DEL relation: event.answer
+website_event_questions / event.registration       / registration_answer_ids (one2many): NEW relation: event.registration.answer
+# NOTHING TO DO
+
+website_event_questions / event.registration.answer / event_answer_id (many2one)    : DEL relation: event.answer, required
+website_event_questions / event.registration.answer / event_registration_id (many2one): DEL relation: event.registration, required
+website_event_questions / event.registration.answer / question_id (many2one)        : NEW relation: event.question, required
```
MiquelRForgeFlow, line 23: This field should be filled in post-migration:

```sql
UPDATE event_registration_answer era
SET question_id = eqa.question_id
FROM event_question_answer eqa
WHERE era.value_answer_id = eqa.id
```

tranngocson1996: @MiquelRForgeFlow I missed this field. Done bro

openupgrade_scripts/scripts/website_event_questions/14.0.1.1/pre-migration.py
MiquelRForgeFlow: but don't delete the rename_tables. you have to put both! haha

tranngocson1996: > but don't delete the rename_tables. you have to put both! haha

Haha I don't understand why should I do that?
You can explain for me🙏

MiquelRForgeFlow: rename_models rename the models, and rename_tables rename the tables of those models. they are two different things and both are needed.

tranngocson1996: I thought the system would understand by itself and only need 1 run.  It will update both

MiquelRForgeFlow (approved):

RaulOForgeFlow (approved):

PR #3109, opened 2022-01-17
[MIG][14.0]sale_project: migration script

hieu1211:
This PR
======
Done


openupgrade_scripts/scripts/sale_project/14.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,36 @@
...
+sale_project / sale.order.line          / project_id (many2one)         : previously in module sale_timesheet
+sale_project / sale.order.line          / task_id (many2one)            : previously in module sale_timesheet
+# NOTHING TO DO
+
+---XML records in module 'sale_project'---
+NEW ir.actions.server: sale_project.project_task_server_action_batch_invoice
+NEW ir.model.access: sale_project.access_sale_order_line_project_manager [renamed from sale_timesheet module]
+NEW ir.model.access: sale_project.access_sale_order_project_manager [renamed from sale_timesheet module]
+NEW ir.model.constraint: sale_project.constraint_project_project_sale_order_required_if_sale_line [renamed from sale_timesheet module]
+NEW ir.rule: sale_project.sale_order_line_rule_project_manager [renamed from sale_timesheet module] (noupdate)
```
MiquelRForgeFlow, line 30: I would call rename_xmlids() in pre.

hieu1211: following i tested on localhost, xml_id was renamed without having to do anything

MiquelRForgeFlow: I doubt it. Maybe you are seeing the new xmlid, but the old still exists.

It's better to assure the module rename in pre-migration. you may use openupgrade.update_module_names() instead of openupgrade.rename_xmlids(), as you prefer.

hieu1211: I verified that there is just one record in the database, that it is not duplicated, and renamed. But for sure, I'll include a code renaming for them.

RaulOForgeFlow (approved):
LGTM :+1:

tranngocson1996 (approved):

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
I don't think the migration scripts of newly installed dependencies are run anymore in the refactored OpenUpgrade 14.0 framework. I'll have a look at this next week.

pedrobaeza:
AFAIK, this has been the case always: new modules don't run migration scripts, so this should be executed in `sale_timesheet` module, not here.

StefanRijnhart:
In core Odoo, yes. But we always had this as a customization in OpenUpgrade. I think I see the problem and I will provide a patch shortly.

pedrobaeza:
> In core Odoo, yes

I think you are talking before the migration machinery refactoring, where `version` was passed as None, but this changed a lot of years ago, and they didn't run anymore migration scripts since 4 years ago or so.

MiquelRForgeFlow (changes requested):
> so this should be executed in sale_timesheet module, not here.

StefanRijnhart:
Was I not being clear? We have always run migration scripts for newly installed modules in OpenUpgrade, since https://github.com/OCA/OpenUpgrade/pull/31. A port to the new framework was attempted, but it is not complete: https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_framework/odoo_patch/odoo/modules/migration.py. I will fix this in the coming days.

pedrobaeza:
I didn't know about such change, but it's good to be known. I think all that changes should be documented in the README of the `openupgrade_framework` module for not requiring to be remembered .

StefanRijnhart:
@pedrobaeza you're right, and in this particular case it is! https://github.com/OCA/OpenUpgrade/blame/14.0/openupgrade_framework/readme/DESCRIPTION.rst#L17-L19

StefanRijnhart:
Fix here now: https://github.com/OCA/OpenUpgrade/pull/3155

pedrobaeza:
> @pedrobaeza you're right, and in this particular case it is! https://github.com/OCA/OpenUpgrade/blame/14.0/openupgrade_framework/readme/DESCRIPTION.rst#L17-L19

Oh, I see. I would change the redaction to something more general, as the patch allows to run the scripts on any newly installed module, being dependency or a fresh one. Do you want me to do the change?

StefanRijnhart:
@pedrobaeza yes, go ahead if you can make it more clear!

PR #3111, opened 2022-01-18
[12.0][OU-FIX] repair: correct xmlid renames

MiquelRForgeFlow:
This happens because the module rename is already done in that step, and the "module" part of the xmlid is already changed.


pedrobaeza (approved):

legalsylvain (approved):
indeed. Thanks !

PR #3114, opened 2022-01-20
[9.0][OU-FIX] account: Adjust range sequences

pedrobaeza:
The Odoo models were previously adapted in the migration scripts, but the correspondences depending the sequence implementation weren't, resulting in errors on later versions.

This code adds such part, exchanging values from no_gap to standard depending on the implementation on parent and child range sequences.

@Tecnativa


StefanRijnhart (approved):
Thanks!

PR #3117, opened 2022-01-26
[12.0][FIX] mrp: only fill new data pickings pbm_type_id and sam_type_id

MiquelRForgeFlow:
We should touch the rest of the existing picking types, as clients may have different configurations for them.

### DIFF
--- a/addons/mrp/migrations/12.0.2.0/post-migration.py
+++ b/addons/mrp/migrations/12.0.2.0/post-migration.py
@@ -1,7 +1,6 @@
 # Copyright 2019 Eficent <http://www.eficent.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
-from odoo.addons.mrp import _create_warehouse_data
 
 
 def fill_mrp_workcenter_resource_calendar_id(cr):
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
2 things:

- Don't you need `manu_type_id` type?
- Instead of verbatim copy, what about monkey patching `_get_picking_type_update_values` before the call to `_create_or_update_sequences_and_picking_types` (and revert back after) to only return the types you want?

MiquelRForgeFlow:
> Don't you need manu_type_id type?

Why? manu_type_id already exists, and we don't want to change its values, the same way we avoid in_type_id, out_type_id, pick_type_id, ...

addons/mrp/migrations/12.0.2.0/post-migration.py
```diff
@@ -27,13 +27,28 @@ def fill_mrp_workcenter_productivity_loss_loss_id(cr):
 
 
 def fill_stock_warehouse_picking_types(env):
+    def _get_picking_type_update_values(self):
+        return {
```
pedrobaeza: Instead of harcoding this, just call "super" (the original method stored in the variable), and return the values for the desired keys.

MiquelRForgeFlow: and what I put inside of super? I mean `data = super(????)._get_picking_type_update_values`. We don't have a class here.

pedrobaeza: I put super between quotes on purpose. You have to do:
```suggestion
        res = _get_picking_type_update_values._original_method(self)
        return {"pbm_type_id": res["pbm_type_id"], "sam_type_id": res["sam_type_id"]}
```

pedrobaeza (approved):

PR #3118, opened 2022-01-26
[12.0][FIX] stock: avoid duplicated putaway strategies

MiquelRForgeFlow:
As the putaway strategies defined by their product template are transformed to ones defined by the corresponding product variants, we should avoid creating the ones for which their variant has been already created in the previous query.

ADDITIONAL NOTE: I don't see why you modify the sequence, because I think we should respect the sequence that was defined in the stock_putaway_product module. My vote is for putting "spps.sequence" instead of "spps.sequence +1". Also, because if you happen to already have odoo putaway strat (defined by categories) and also these from the oca module, depending of the sequence, it may happen that the category ones have more priority (less sequence), which would look bad. Thus, my point is that we should not touch the sequence, and leave for the stock manager to resequence the putaway strategies (if needed) after migration.


pedrobaeza (approved):
About the sequence, check the comment:
```
# We put sequence + 1 for giving more priority by default to product specific rules
```
This was for not caring about the duplicates. With your change, it may be not needed, but what you prefer.

PR #3119, opened 2022-01-27
[FIX][14.0] account: set value wrong

tranngocson1996:
This PR
-------
Fix set value wrong


pedrobaeza:
Check my comment in the other place. I think it's wrong.

pedrobaeza:
It's a char field, so it's incorrect. Merging.

PR #3120, opened 2022-01-28
[13.0] update upstream 9a8319b2b71

MiquelRForgeFlow:
Up to today, commit 9a8319b2b71.

No conflicts.


addons/maintenance/data/maintenance_data.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
-<data>
+<data noupdate="1">
```
MiquelRForgeFlow, line 3: It seems I will need to add some script.

pedrobaeza:
As they have changed in the middle of the lifecycle the noupdate of some data, there may be a lot of installations with such data with the incorrect noupdate flag, so I think we should include the resetting of the flag also on v14. It's harmless for those with the proper flag, and fix those which not.

Anyways, this is correct and I merge now.

PR #3122, opened 2022-01-31
[13.0][FIX] stock: assign correct location to quants/packages (multicompany)

MiquelRForgeFlow:
Complete the leftover of the migration script. After the logic introduced in https://github.com/OCA/OpenUpgrade/commit/bcb3c4ec83a70d01954e7323bdbaebcdea23b387, it's no longer a problem to change the location of quants.


MiquelRForgeFlow:
Tested!!!

PR #3124, opened 2022-02-04
account : prevent double membership to 'group_show_line_subtotals_tax_excluded' and 'group_show_line_subtotals_tax_included' that is forbidden in V12

legalsylvain:
Same as here : https://github.com/OCA/OpenUpgrade/pull/2999 for account module.

CC : @fcayre, @pedrobaeza, @MiquelRForgeFlow

### DIFF
--- a/addons/account/migrations/12.0.1.1/pre-migration.py
+++ b/addons/account/migrations/12.0.1.1/pre-migration.py
@@ -132,6 +132,35 @@ def drop_obsolete_constraint_wizard_multi_charts_accounts(env):
     openupgrade.remove_tables_fks(env.cr, ['wizard_multi_charts_accounts'])
 
 
+def fix_double_membership(cr):
+    # avoid error raised by new function '_check_one_user_type'
+
+    # we arbitrarily keep the tax included
+    group_to_remove = "group_show_line_subtotals_tax_excluded"
+    group_to_keep = "group_show_line_subtotals_tax_included"
+    openupgrade.logged_query(
+        cr, """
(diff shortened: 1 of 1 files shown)


addons/account/migrations/12.0.1.1/pre-migration.py
```diff
@@ -132,6 +132,35 @@ def drop_obsolete_constraint_wizard_multi_charts_accounts(env):
     openupgrade.remove_tables_fks(env.cr, ['wizard_multi_charts_accounts'])
 
 
+def fix_double_membership(cr):
+    # avoid error raised by new function '_check_one_user_type'
+
+    # we arbitrarily keep the tax included
```
pedrobaeza, line 138: Please keep tax excluded as B2B is the most usual one in my experience.

legalsylvain: My point was : B2B is the default, so if the user has b2c selected, it's intented. Dont't you think ?

pedrobaeza: Is that really that way? Shouldn't all the users be homogeneous with one or the other? Applying res.config.settings apply to all users.

StefanRijnhart: I think Pedro is right. `res.config.settings` wipes the users from the groups that are deactivated: https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_config.py#L586-L587

PR #3128, opened 2022-02-08
[12.0][FIX] stock: avoid duplicated putaway strategies (fixup)

MiquelRForgeFlow:
Fixup of https://github.com/OCA/OpenUpgrade/pull/3118. Properly detect the template ones.

### DIFF
--- a/addons/stock/migrations/12.0.1.1/post-migration.py
+++ b/addons/stock/migrations/12.0.1.1/post-migration.py
@@ -155,16 +155,19 @@ def merge_stock_putaway_product(cr):
             FROM stock_product_putaway_strategy
             WHERE product_product_id IS NOT NULL"""
         ).format(sql.Identifier(column_name)))
-        # second, we add the ones with product product template
+        # second, we add the ones with product template
+        # We put sequence + 1000 for giving more priority by default to product
+        # specific rules
         openupgrade.logged_query(cr, sql.SQL(
             """INSERT INTO stock_fixed_putaway_strat (product_id, putaway_id,
                 fixed_location_id, sequence,
                 create_uid, create_date, write_uid, write_date, {})
             SELECT pp.id, spps.putaway_id, spps.fixed_location_id,
-                spps.sequence, spps.create_uid, spps.create_date,
+                spps.sequence + 1000, spps.create_uid, spps.create_date,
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Don't we need again then to reintroduce sequence + 1 for giving less priority to template ones?

MiquelRForgeFlow:
No, I think the sequence should be preserved from stock_putaway_product module.

pedrobaeza:
If you look at the code, they were processed after independently from the sequence and only if no specific product variant is set:

https://github.com/OCA/stock-logistics-warehouse/blob/12.0/stock_putaway_product_form/models/putaway_strategy.py#L46

so we can't merge both or we will have different behavior. The way to do that is through sequence. I would say even more: let's put a sequence + 10000 for avoiding coincidences.

MiquelRForgeFlow:
stock_putaway_product_form module doesn't exist in v11.

pedrobaeza:
Ah, which one is then?

MiquelRForgeFlow:
stock_putaway_product in v11, has a sequence field, but I don't see it being used as you describe, but yes, there exist a preference of variant over template in https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_putaway_product/models/product_putaway.py#L33.

MiquelRForgeFlow:
I mean, in the queries, first are put the variant with the sequence of the variant, and later are put the templates with the sequence of the templates. As we avoid duplication, the sequence of the variant prevail. Then I don't see why we should modify the sequence.

pedrobaeza:
Take this example:

Variant strategies with sequences 10, 11 and 12
Template strategies with sequences  1, 2 and 3

The template strategy will prevail now over the variant one if you don't modify the sequence for the templates when mixing both, changing the previous behavior.

MiquelRForgeFlow:
Yeah, I know your point. But my point was that in stock_putaway_product, both variants and templates are mixed with the same sequence, and is responsibility of the user of stock_putaway_product module to check the sequence and choose priorities. Later, after migration and merge of that module into stock, my point was to respect that sequence and priority. But hey, that's not blocking for me, I will pump the sequence for templates.

pedrobaeza (approved):

PR #3131, opened 2022-02-10
[13.0][FIX] stock: assign correct location to quants/packages (multicompany) [fixup]

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/commit/674bb2a3d8b21793e519cd5e0d1ea8dbe872c243

The PR https://github.com/OCA/OpenUpgrade/pull/3122 was incomplete (work in progress) when merged :S

Explanation: Although the previous PR solved the multicompany error I had (quants having different company than their locations), later I saw that quants appeared in the wrong adjustment valuation location. Thus, I was working on fixing it when that PR got merged.

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -266,6 +266,22 @@ def map_stock_locations(env, main_company):
     env['res.company'].create_missing_production_location()
     env['res.company'].create_missing_scrap_location()
 
+    # company of quant is related to company of its location_id
+    # but the specific locations we are dealing don't have company!
+    # thus, we put temporarily a company from their product/lots
+    openupgrade.logged_query(env.cr, """
+        UPDATE stock_quant sq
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Next time, please mention that is WIP for avoiding that.

pedrobaeza:
Products may not have company as well. Do you have any other strategy?

MiquelRForgeFlow:
> Next time, please mention that is WIP for avoiding that.

It had the WIP tag 🥲

> Products may not have company as well. Do you have any other strategy?

Well, here I am using the lot, as the lot has required company. What about quants with products without company and without lots? well, nothing I can do, they will be left in the company of their location, which should be the one in the company of the lowest id. To handle that would be hard, as we would need to check stock move lines, see the difference that enter and exit a location, and do shitty comparisons... blggh, I don't want to think about it.

pedrobaeza:
I meant to put the pull request in draft mode.

Let's merge it then as best possible effort.

PR #2694, opened 2021-05-04
[13.0][OU-ADD] stock_production_lot_multi_company: Merged stock_production_lot_multi_company to stock

victoralmau:
Merged `stock_production_lot_multi_company` to `stock` 

@Tecnativa TT26948

PR #3132, opened 2022-02-10
[13.0][FIX] stock: handle module stock_production_lot_multi_company migration

MiquelRForgeFlow:
Oversight of https://github.com/OCA/OpenUpgrade/pull/2694.

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -353,6 +353,13 @@ def map_stock_locations(env, main_company):
     """)
 
 
+def stock_production_lot_multi_company_migration(env):
+    rule = env.ref("stock.stock_production_lot_rule", raise_if_not_found=False)
+    if 'user' in rule.domain_force:
+        rule.write({"name": "Stock Production Lot multi-company",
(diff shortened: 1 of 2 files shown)


addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -353,6 +353,13 @@ def map_stock_locations(env, main_company):
     """)
 
 
+def stock_production_lot_multi_company_migration(env):
+    rule = env.ref("stock.stock_production_lot_rule", raise_if_not_found=False)
+    if 'user' in rule.domain_force:
+        openupgrade.load_data(
+            env.cr, "stock", "migrations/13.0.1.1/noupdate_changes3.xml")
```
pedrobaeza: Isn't better to simply write here the domain with Python?

MiquelRForgeFlow: I thought so. But as the name also changes I thought doing by xml. I could also do a write with the two fields... but meh.

MiquelRForgeFlow: What do you prefer?

pedrobaeza (approved):
To be merged when finished

PR #3133, opened 2022-02-10
[13.0][OU-ADD] pos_restaurant: Nothing to do

pedrobaeza:
FYI, `pos_order_line_note` is no more needed, as Odoo has amended the inconsistency between UI and data model.

@Tecnativa TT31765


chienandalu (commented):
Should we put it in the apriori?

pedrobaeza:
Yeah, but that should be in another PR.

chienandalu (approved):

PR #3138, opened 2022-02-15
[14.0][FIX] calendar: to insert such daily, monthly, yearly events in the recurrence table

royle-vietnam:
Event with daily, monthly, yearly will not insert in the recurrence table, because of no day is selected

Steps:
1. Create a daily event, repeat 10 times
2. Run migration scripts

### DIFF
--- a/openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
@@ -40,7 +40,13 @@ def fill_calendar_recurrence_table(env):
             FROM calendar_event
             WHERE recurrency AND recurrence_id IS NULL
                 AND (recurrent_id IS NULL OR recurrent_id = 0)
-                AND (mo OR tu OR we OR th OR fr OR sa OR su)
+                AND (
+                    rrule_type != 'weekly'
+                    OR (
+                        rrule_type = 'weekly' AND
+                        (mo OR tu OR we OR th OR fr OR sa OR su)
+                    )
+                )
             RETURNING id,base_event_id
         )
         UPDATE calendar_event ce


pedrobaeza:
This is not a revert seeing the code, but a complement. Anyways, please specify which case is not covered.

royle-vietnam:
> This is not a revert seeing the code, but a complement. Anyways, please specify which case is not covered.

@pedrobaeza as at description, event with daily will not insert in the recurrence table, because of no day is selected

pedrobaeza:
That's not what the description says, and again, it's not a revert. Please change commit message to indicate this and more if you put steps to reproduce the problem.

royle-vietnam:
> That's not what the description says, and again, it's not a revert. Please change commit message to indicate this and more if you put steps to reproduce the problem.

Done, please review @pedrobaeza

openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
```diff
@@ -40,7 +40,10 @@ def fill_calendar_recurrence_table(env):
             FROM calendar_event
             WHERE recurrency AND recurrence_id IS NULL
                 AND (recurrent_id IS NULL OR recurrent_id = 0)
-                AND (mo OR tu OR we OR th OR fr OR sa OR su)
+                AND (
+                        rrule_type != 'weekly'
+                        OR (mo OR tu OR we OR th OR fr OR sa OR su)
+                    )
```
pedrobaeza:
```suggestion
                AND (
                    rrule_type != 'weekly'
                    OR (
                        rrule_type = 'weekly' AND
                        (mo OR tu OR we OR th OR fr OR sa OR su)
                    )
                )
```

royle-vietnam: if `rrule_type != 'weekly'` then `rrule_type = 'weekly'`, why you need to check `rrule_type = 'weekly' AND (mo OR tu OR we OR th OR fr OR sa OR su)`

pedrobaeza: Because you may have marked the event as weekly recurrent, but then no recurrent day is selected. That's possible on <=v13, but on v14, there's a constraint avoiding that, so we have to avoid to insert such fake recurrent events.

royle-vietnam: ok. I fixed it

PR #3140, opened 2022-02-15
[FIX][14.0] account: process account payment transfer

royle-vietnam:
When migrating from Odoo 13.0 to Odoo 14.0, we need:
1. mapping payment_type from 'transfer' to 'inbound'/'outbound' for account payment tranfer, because key `transfer` of `payment_type` is deleted.
So we need set `is_internal_transfer` as `true` on account payment transfer to indicate that it was an account payment transfer 

2. On Odoo 13.0, when validating a account payment transfer, so 2 journal entries is created, equivalent to 1 receipt payment, and 1 send payment on Odoo 14.0
So, on Odoo 14.0, we need create new counterpart payment with account payment transfer

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -591,6 +591,45 @@ def fill_account_journal_payment_credit_debit_account_id(env):
         )
 
 
+def create_new_counterpart_account_payment_transfer(env):
+    # Create new counterpart payment with account payment transfer
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO account_payment (move_id, is_internal_transfer, partner_type,
+            payment_type,
+            amount, currency_id,
+            destination_account_id, partner_id, journal_id,
+            create_uid, create_date, write_uid, write_date)
+        SELECT move.id, true, ap.partner_type,
+            CASE
+            WHEN journal.id = ap.destination_journal_id THEN 'inbound' ELSE 'outbound'
+            END,
+            ap.amount, ap.currency_id,
+            ap.destination_account_id, ap.partner_id, move.journal_id,
+            ap.create_uid, ap.create_date, ap.write_uid, ap.write_date
+        FROM account_payment ap
+        JOIN account_move move
+            ON (move.payment_id = ap.id AND move.id != ap.move_id)
+        JOIN account_journal journal ON journal.id = move.journal_id
+        WHERE ap.payment_type = 'transfer'
+        """,
+    )
+
+
+def map_account_payment_transfer(env):
+    # map payment_type from transfer to 'inbound'/'outbound'
+    # and set is_internal_transfer as true on account payment transfer
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_payment ap
+        SET is_internal_transfer = true, payment_type = 'outbound'
+        WHERE ap.payment_type = 'transfer'
+        """,
+    )
+
+
 def fill_account_payment_with_no_move(env):
     p_data = {}
     p_dates_by_company = {}
@@ -752,6 +791,8 @@ def migrate(env, version):
     fill_company_account_journal_suspense_account_id(env)
     fill_statement_lines_with_no_move(env)
     fill_account_journal_payment_credit_debit_account_id(env)
+    create_new_counterpart_account_payment_transfer(env)
+    map_account_payment_transfer(env)
     fill_account_payment_with_no_move(env)
     _delete_hooks(env)
     openupgrade.delete_record_translations(
--- a/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
@@ -295,7 +295,8 @@ account      / account.payment          / payment_method_id (many2one)  : now a 
 # NOTHING TO DO
 
 account      / account.payment          / payment_type (selection)      : selection_keys is now '['inbound', 'outbound']' ('['inbound', 'outbound', 'transfer']')
-# TODO: transfer records should be changed to 'inbound' or 'outbound', but if we do that, we should remove two account move lines
+# DONE post-migration: mapped payment_type from 'transfer' to 'inbound'/'outbound' for account payment tranfer
+# created new counterpart payment with account payment transfer
 # (see https://github.com/odoo/odoo/commit/a30cfee04eaf48b581d2e327c1b56e115a1c751f)
 # DONE pre-migration: filled partner_id with journal_id.company_id.partner_id to detect transfers
 


tranngocson1996 (approved):

pedrobaeza:
Please one change per PR, and as always, document why are you doing these changes.

royle-vietnam:
>  Please one change per PR, and as always, document why are you doing these changes.

This one change to process account payment transfer, because of currently not handling it when upgrading from 13.0 to 14.0

pedrobaeza:
I see several methods, and you have to explain more what are you doing, as it's not clear from the code or your initial message.

royle-vietnam:
> I see several methods, and you have to explain more what are you doing, as it's not clear from the code or your initial message.

I split the method to make the code more readable
Where do you not see clearly?

pedrobaeza:
I need to know why are you doing this for validating it.

royle-vietnam:
> I need to know why are you doing this for validating it.

I have added more information, please review

hieu1211 (approved):

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,69 @@ def fill_account_journal_payment_credit_debit_account_id(env):
...
+            amount, currency_id,
+            destination_account_id, partner_id, journal_id,
+            create_uid, create_date, write_uid, write_date)
+        SELECT move.id, true, ap.partner_type,
+            CASE
+            WHEN liquidity_line.amount_currency > 0 THEN 'inbound' ELSE 'outbound'
+            END,
+            ABS(liquidity_line.amount_currency), liquidity_line.currency_id,
+            ap.destination_account_id, ap.partner_id, move.journal_id,
+            move.create_uid, move.create_date, move.write_uid, move.write_date
```
pedrobaeza: Better to use the same dates as the original payment of type transfer:

```suggestion
            ap.create_uid, ap.create_date, ap.write_uid, ap.write_date
```

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,69 @@ def fill_account_journal_payment_credit_debit_account_id(env):
...
 
 
+def create_new_counterpar_account_payment_transfer(env):
+    # Create new counterpart payment with account payment transfer
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO account_payment (move_id, is_internal_transfer, partner_type,
+            payment_type,
+            amount, currency_id,
```
pedrobaeza, line 601: Missing `is_internal_transfer` flag (forced to True):

```suggestion
            amount, currency_id, is_internal_transfer,
```

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,69 @@ def fill_account_journal_payment_credit_debit_account_id(env):
...
+            JOIN account_journal journal ON journal.id = ap.journal_id
+            LEFT JOIN account_move_line line
+                ON line.move_id = ap.move_id
+                    AND line.account_id IN (
+                        journal.default_account_id,
+                        journal.payment_debit_account_id,
+                        journal.payment_credit_account_id
+                        )
+            WHERE ap.payment_type = 'transfer'
+        )
```
pedrobaeza: Unneeded. See my final explanation.

pedrobaeza (changes requested):
On v13, there were 2 journals `journal_id` and `destination_journal_id`. That and the sign of the amount must be they way to distinguish both payment types, not by matching by the defined accounts, that may vary through time.

Please change the approach according that.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
royle-vietnam: @pedrobaeza I added it at line 613

pedrobaeza: Ouch, sorry. I didn't see it. The other part of the type is the most important one.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,69 @@ def fill_account_journal_payment_credit_debit_account_id(env):
...
+        INSERT INTO account_payment (move_id, is_internal_transfer, partner_type,
+            payment_type,
+            amount, currency_id,
+            destination_account_id, partner_id, journal_id,
+            create_uid, create_date, write_uid, write_date)
+        SELECT move.id, true, ap.partner_type,
+            CASE
+            WHEN liquidity_line.amount_currency > 0 THEN 'inbound' ELSE 'outbound'
+            END,
+            ABS(liquidity_line.amount_currency), liquidity_line.currency_id,
```
royle-vietnam: @pedrobaeza Better to use the same amount and currency as the original payment of type transfer?
```suggestion
            ABS(ap.amount), ap.currency_id,
```

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
royle-vietnam: Thank you, I'll fix it

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
pedrobaeza: If it matches, it doesn't matter. What seems weird is the `abs`. Check it with the new approach.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
royle-vietnam: Still need this part, because we just set to True for the new payment, the old payment has not been set yet

pedrobaeza: Yeah, but I'm talking about the subquery and all of that. Simply those that have 2 journals are internal.

royle-vietnam:
@pedrobaeza I have processed all your comments, and it works fine, please review, thank you

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,45 @@ def fill_account_journal_payment_credit_debit_account_id(env):
...
+
+def map_account_payment_transfer(env):
+    # map payment_type from transfer to 'inbound'/'outbound'
+    # and set is_internal_transfer as true on account payment transfer
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_payment ap
+        SET is_internal_transfer = true, payment_type = 'outbound'
+        WHERE ap.payment_type = 'transfer'
```
MiquelRForgeFlow, line 628: I think here needs an additional condition, right? 🤔

royle-vietnam: No, simply applying for the send transfer
For receiving transfer, it is created above

MiquelRForgeFlow: Yeah, but the created above also will be changed with this query, that's what I mean.

MiquelRForgeFlow: I mean, your are putting all transfers (old and new ones) to outbound, and that's what doesn't look correct to me.

royle-vietnam: No, when created, it is already payment_type as outbound / inbound, not transfer

MiquelRForgeFlow: Oh, right, my bad.

MiquelRForgeFlow:
@royleviindoo please, update https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt#L298

royle-vietnam:
> @royleviindoo please, update https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt#L298

@MiquelRForgeFlow ok, I updated

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,45 @@ def fill_account_journal_payment_credit_debit_account_id(env):
...
+        JOIN account_move move
+            ON (move.payment_id = ap.id AND move.id != ap.move_id)
+        JOIN account_journal journal ON journal.id = move.journal_id
+        WHERE ap.payment_type = 'transfer'
+        """,
+    )
+
+
+def map_account_payment_transfer(env):
+    # map payment_type from transfer to 'inbound'/'outbound'
```
MiquelRForgeFlow, line 621:
```suggestion
    # map payment_type from transfer to 'outbound'
```
because in the query you are mapping them only to 'outbound'

MiquelRForgeFlow (approved):
LGTM

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
royle-vietnam: Thank you, I updated

MiquelRForgeFlow:
bad rebase, conflicts

royle-vietnam:
> bad rebase, conflicts

Done, I fixed

flotho:
hi @royleviindoo maybe ask @pedrobaeza to review the resolution your made for his changes requests.

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -605,6 +605,45 @@ def fill_account_journal_payment_credit_debit_account_id(env):
         )
 
 
+def create_new_counterpar_account_payment_transfer(env):
```
etobella: Typo: create_new_counterpart_account_payment_transfer

etobella (approved):
It make sense.

Just a small comment

LoisRForgeFlow (approved):

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
LoisRForgeFlow: @royleviindoo could you adjust this?

royle-vietnam: @LoisRForgeFlow ok, done

pedrobaeza (approved):

PR #3141, opened 2022-02-15
[OU-FIX][14.0] updating 'base' should update all modules

### LINKED ISSUE #3061: [14.0] openupgrade_framework: New installed modules dependencies are not auto-resolved

StefanRijnhart:
Port of https://github.com/odoo/odoo/pull/83543

Fixes #3061

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/__init__.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/__init__.py
@@ -1,2 +1,3 @@
 from . import ir_model
+from . import ir_module
 from . import ir_ui_view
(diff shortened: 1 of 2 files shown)


flotho (approved):
Code review, LGTM

StefanRijnhart:
Fixed in https://github.com/odoo/odoo/pull/85663.

PR #3143, opened 2022-02-16
[12.0][FIX] *: lift obsolete many2one constraints

MiquelRForgeFlow:
I only checked this three modules, as they are very important.

### DIFF
--- a/addons/account/migrations/12.0.1.1/pre-migration.py
+++ b/addons/account/migrations/12.0.1.1/pre-migration.py
@@ -120,7 +120,7 @@ def prefill_account_chart_template_transfer_account_prefix(env):
                 "SET transfer_account_code_prefix = 'OUB'")
 
 
-def drop_obsolete_constraint_wizard_multi_charts_accounts(env):
+def drop_obsolete_fk_constraints(env):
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Why not removing FKs also on obsolete models?

MiquelRForgeFlow:
The ones in `account` you mean? I was too lazy haha

pedrobaeza:
Yeah, but the same problem that you have faced with fields may arise due to this.

MiquelRForgeFlow:
Ready!

addons/account/migrations/12.0.1.1/pre-migration.py
```diff
@@ -130,6 +130,26 @@ def drop_obsolete_constraint_wizard_multi_charts_accounts(env):
     has been removed in V12.0
     """
     openupgrade.remove_tables_fks(env.cr, ['wizard_multi_charts_accounts'])
+    openupgrade.remove_tables_fks(env.cr, ['account_aged_trial_balance'])
```
pedrobaeza: Why not passing all tables in one call?

MiquelRForgeFlow: done

PR #3145, opened 2022-02-16
[12.0][REM] l10n_nl: Stop deletion of tag 3b

ddejong-therp:
The `l10n_nl` module removes an old tag by xmlid, because in 12.0, it is split into two seperate tags. However, in 13.0, it is back to one tag again. Removing the tag, also removes any link to the tags (in account.tax if I'm not mistaken). That is an issue, because then it becomes impossible to migrate it to 13.0 properly.

There is another issue that causes none of the `l10n_nl` tags to be properly migrate to 13.0, but I've fixed that as well, and will be part of another PR. However, as long as the old tag 3b is deleted, a proper migration of that specific tag is impossible, because any link to it is gone.

Not sure what the reasoning behind removing this old tag was, perhaps it was a way of forcing people to 'fix' their migration by reapplying one of the new 2 different 3b tags manually everywhere.

### DIFF
--- a/addons/l10n_nl/migrations/12.0.3.0/post-migration.py
+++ b/addons/l10n_nl/migrations/12.0.3.0/post-migration.py
@@ -5,8 +5,4 @@ from openupgradelib import openupgrade
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.delete_records_safely_by_xml_id(
-        env, [
-            'l10n_nl.tag_nl_13',
-        ],
-    )
+    openupgrade.set_xml_ids_noupdate_value(env, 'l10n_nl', ['tag_nl_13'], True)


addons/l10n_nl/migrations/12.0.3.0/post-migration.py
```diff
@@ -1,12 +0,0 @@
-# Copyright 2019 Onestein (<https://www.onestein.eu>)
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from openupgradelib import openupgrade
-
-
-@openupgrade.migrate()
-def migrate(env, version):
-    openupgrade.delete_records_safely_by_xml_id(
```
MiquelRForgeFlow, line 8: Then, instead of this, you have to remove the corresponding ir.model.data, to avoid the tag being deleted by odoo later.

pedrobaeza: Better to mark it as noupdate=1 for having traces.

ddejong-therp: Ah yes, that makes sense. May I ask what problem leaving the xmlid in ir.model.data might cause?

MiquelRForgeFlow: If it is noupdate=0, then odoo will remove that ir.model.data, and by doing that, will remove the tag. If the tag don't have ir.model.data or the ir.model.data has noupdate=1, then the tag will not be removed.

MiquelRForgeFlow: I think also it is better to leave the xmlid for having traces. Maybe in a migration script in later versions will need to make us of that xmlid.

ddejong-therp:
I ended up removing the `ir.model.data`.

MiquelRForgeFlow:
> I ended up removing the `ir.model.data`.

Why?

ddejong-therp:
Ah, yeah I forgot about your message that it might be useful to keep the xmlid. It's been a while since I got back to this issue.

ddejong-therp:
This should be better.

addons/l10n_nl/migrations/12.0.3.0/post-migration.py
```diff
@@ -5,8 +5,4 @@
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.delete_records_safely_by_xml_id(
-        env, [
-            'l10n_nl.tag_nl_13',
-        ],
-    )
+    openupgrade.set_xml_ids_noupdate_value('l10n_nl', ('tag_nl_13',), True)
```
MiquelRForgeFlow:
```suggestion
    openupgrade.set_xml_ids_noupdate_value(env, 'l10n_nl', ['tag_nl_13'], True)
```

ddejong-therp: Ah my bad. I thought I tested it, but apparently I did not.. But I changed it and tested it this time.

MiquelRForgeFlow (approved):

StefanRijnhart (approved):

PR #3147, opened 2022-02-21
[14.0][FIX] account: add hooks to avoid hardcoding lock dates

MiquelRForgeFlow:
I think this makes the code a bit cleaner.


pedrobaeza:
Yeah, indeed it's more readable and consistent. Please create also a method for registering hooks, and another for unregistering them, and call such methods from the rest of places.

pedrobaeza (approved):
Although some docstrings are missing on methods to know why they are there, I'm not going to block this more time

PR #3150, opened 2022-02-23
[14.0][FIX] account: fill partner_bank_id in statement line moves

MiquelRForgeFlow:
`absl.bank_account_id` -> `am.partner_bank_id`

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -546,6 +546,15 @@ def fill_statement_lines_with_no_move(env):
         st_line.move_id.with_context(skip_account_move_synchronization=True).write(
             to_write
         )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move am
+        SET partner_bank_id = absl.bank_account_id
+        FROM account_bank_statement_line absl
+        WHERE am.statement_line_id = absl.id AND am.partner_bank_id IS NULL
+            AND absl.bank_account_id IS NOT NULL""",
+    )
 
 
 def fill_account_journal_payment_credit_debit_account_id(env):
--- a/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
@@ -435,3 +435,9 @@ def migrate(env, version):
     openupgrade.delete_records_safely_by_xml_id(
         env, ["account.constraint_account_move_line_check_amount_currency_balance_sign"]
     )
+    openupgrade.remove_tables_fks(
+        env.cr, ["account_bank_statement_import_ir_attachment_rel"]
+    )
+    openupgrade.lift_constraints(
+        env.cr, "account_bank_statement_line", "partner_account_id"
+    )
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Shouldn't this be on `account_statement_import`?

MiquelRForgeFlow:
No, because independently of that module, you should want to have the info of `absl.bank_account_id` into `am.partner_bank_id`.

I mean, `absl.bank_account_id` is in account (v13) and `am.partner_bank_id` is in account (v14), nothing to do with `account_statement_import`.

AaronHForgeFlow (approved):
:+1:

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -270,6 +270,15 @@ def add_move_id_field_account_bank_statement_line(env):
         WHERE am.id = absl.move_id AND am.statement_line_id IS NULL
         """,
     )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move am
+        SET partner_bank_id = absl.bank_account_id
+        FROM account_bank_statement_line absl
+        WHERE am.statement_line_id = absl.id AND am.partner_bank_id IS NULL
```
pedrobaeza:
```suggestion
        WHERE am.statement_line_id = absl.id AND am.partner_bank_id IS NULL AND absl.bank_account_id IS NOT NULL
```

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -546,6 +546,16 @@ def fill_statement_lines_with_no_move(env):
...
             to_write
         )
+    if st_lines:
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE account_move am
+            SET partner_bank_id = COALESCE(absl.bank_account_id, am.partner_bank_id)
+            FROM account_bank_statement_line absl
+            WHERE am.statement_line_id IN %s""",
```
pedrobaeza: Why not include the value in https://github.com/OCA/OpenUpgrade/pull/3150/files#diff-c7c0a45514483f4ef6990fbaa0404370e1205cb5be272c0e87d4319199526d17R515 ?

MiquelRForgeFlow: because absl.bank_account_id doesn't exist in the environment. it is a deprecated field.

pedrobaeza: Then write just the needed ones:

```suggestion
            UPDATE account_move am
            SET partner_bank_id = absl.bank_account_id
            FROM account_bank_statement_line absl
            WHERE am.statement_line_id IN %s AND absl.bank_account_id IS NOT NULL AND am.partner_bank_id IS NULL""",
```

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -546,6 +546,17 @@ def fill_statement_lines_with_no_move(env):
...
+    if st_lines:
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE account_move am
+            SET partner_bank_id = absl.bank_account_id
+            FROM account_bank_statement_line absl
+            WHERE am.statement_line_id IN %s AND absl.bank_account_id IS NOT NULL
+                AND am.partner_bank_id IS NULL""",
+            (tuple(st_lines.ids),),
```
pedrobaeza: Why not doing the sentence in pre here and remove the one in pre?

MiquelRForgeFlow: ok

pedrobaeza (approved):

PR #3151, opened 2022-02-25
[14.0][OU-FIX] account: map old payment's state to move's state when filling account payment with no move

royle-vietnam:
Steps:
1. Payment is in cancel state and with no move
2. Run migration
3. Payment is in draft state

=> Expected: Payment is in cancel state


MiquelRForgeFlow (approved):
LGTM

pedrobaeza:
For other times, remember that this should be `[OU-FIX]` as commit code.

MiquelRForgeFlow:
> For other times, remember that this should be [OU-FIX] as commit code.

Here is v14, all commits will be OU, so no need to put OU, IMO.

pedrobaeza:
Well, not really. There are commits for the framework, analysis files, etc. I think it's good to keep the same nomenclature on that part.

PR #3155, opened 2022-02-27
[14.0][OU-FIX] Migration scripts were not run during install

StefanRijnhart:
Includes https://github.com/OCA/OpenUpgrade/pull/3109 by @hieu1211 to demonstrate the fix and set `no_version=True` in the decorator.


legalsylvain (approved):
Thanks !

pedrobaeza (approved):
It can be seen working in:

https://github.com/OCA/OpenUpgrade/runs/5350411157?check_suite_focus=true#step:13:4844

Issue #3056, opened 2022-01-07
[13.0 --> 14.0] account incorrect account moves created for bank statement lines


remi-filament:
## Module
account

## Describe the bug
Every account_move created for account_bank_statement_line which did not already have related account_move had 2 lines of suspense account (instead of only 1) + 1 line for bank account (as expected).
This caused all of these moves to be not balanced and therefore remained could not be validated directly without removal of 1 one the suspense account line.

This might be related to the [post-migration script fill_account_journal_payment_credit_debit_account_id() function](https://github.com/OCA/OpenUpgrade/blob/fb786b529fc460bb99334fac12ea79db63476191/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py#L394) but I am not sure...
It seems to me that suspense lines are probably created twice, since account_move_line _prepare_move_line_default_vals() function will be called twice : once in [st_line._synchronize_to_moves() line 458](https://github.com/OCA/OpenUpgrade/blob/fb786b529fc460bb99334fac12ea79db63476191/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py#L458) and once on [line 475](https://github.com/OCA/OpenUpgrade/blob/fb786b529fc460bb99334fac12ea79db63476191/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py#L475).

## To Reproduce
**Affected versions**: migration to 14.0

1. Get a v13.0 database
2. Create bank statement and leave some statement lines unreconciled (so that no related account_move would be created for these)
3. Migrate to v14.0
4. Check the account_move created during migration for these unreconciled bank statement lines

**Expected behavior**
Balanced account_moves with 1 line for bank account and 1 counterpart line with suspense account

remi-filament:
Split from #3050 as requested by @pedrobaeza

MiquelRForgeFlow:
Solved in https://github.com/OCA/OpenUpgrade/pull/3251.

StefanRijnhart:
Fix merged. Thanks!

PR #3159, opened 2022-03-01
[FIX][14.0] account: duplicate journal items when filling account payment / statement line with no move

royle-vietnam:
Steps:
1. Create a payment transfer that is in draft state
2. Run migration
3. The payment that has been linked with journal entry, but journal
entry contains 4 journal items

Expected: journal entry contains 2 journal items

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -521,17 +521,7 @@ def fill_statement_lines_with_no_move(env):
             _logger.error("Failed for statement line with id %s: %s", st_line.id, e)
             raise
         deprecated_accounts.deprecated = True
-        to_write = {
-            "line_ids": [
-                (0, 0, line_vals)
-                for line_vals in st_line._prepare_move_line_default_vals(
-                    counterpart_account_id=False
-                )
-            ]
-        }
-        st_line.move_id.with_context(skip_account_move_synchronization=True).write(
-            to_write
-        )
+
     openupgrade.logged_query(
         env.cr,
         """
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
> Create a payment transfer that is in draft state

What if the payment is not a transfer?

royle-vietnam:
> > Create a payment transfer that is in draft state
> 
> What if the payment is not a transfer?

Also duplicate journal items

MiquelRForgeFlow:
What if we add a (5, 0) in line_ids? This way the previous ones will be deleted, right? But I don't know what differ the previous ones from the new ones.

MiquelRForgeFlow (approved):
Yeah, ok.

pedrobaeza:
Please rewrite the commit message indicating that is not only with a payment transfer, but with any payment.

MiquelRForgeFlow:
Then, in `fill_statement_lines_with_no_move()` should do the same, right? the moves of bank statement lines will have duplicated journal items. Please fix that method too.

royle-vietnam:
> Please rewrite the commit message indicating that is not only with a payment transfer, but with any payment.

I updated

royle-vietnam:
> Then, in `fill_statement_lines_with_no_move()` should do the same, right? the moves of bank statement lines will have duplicated journal items. Please fix that method too.

I handle it in this PR or new PR?

MiquelRForgeFlow:
I think in the same PR.

royle-vietnam:
> I think in the same PR.

Ok, I will handle

royle-vietnam:
> I think in the same PR.

Done, please review

MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):

pedrobaeza:
Please squash in one commit.

royle-vietnam:
> Please squash in one commit.

Done

remi-filament:
We have tested this PR yesterday because we also had an issue with account_move created from bank_statement_lines (see #3056 )
However we previously had 3 lines in account_move (2 of bank account, 1 of suspense account), we now have only 1 of suspense account and none of bank account so the account move is still not valid !
I should try to investigate on how to fix this for bank statement, I keep you posted of my findings...

pedrobaeza:
OK. I trusted on the review of my colleagues, but I suppose this manual entry was added by a reason

remi-filament:
Hi @pedrobaeza @royleviindoo @MiquelRForgeFlow PR #3251 should fix the problem you had for payment for account bank statement lines as well !

PR #3160, opened 2022-03-02
[IMP][14.0] base: remove all modules that have been deleted on target version to avoid warning module: not installable, skipped

hoangtiendung070797:
Perform the various steps required to uninstall all modules completelyincluding the deletion of all database structures created by the module: tables, columns, constraints, etc. and remove all modules

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
@@ -3,6 +3,8 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from openupgradelib import openupgrade
 
+import odoo
+
 
 def fix_module_category_parent_id(env):
     # due to renames, we need to correct the parent_id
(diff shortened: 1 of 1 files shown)


royle-vietnam (approved):
LGTM

hieu1211 (approved):

pedrobaeza (changes requested):
I don't think the uninstallation should be done automatically, as it's a very hard and dangerous step, and also you can't recover from that step. For example, you forget to add a module to your addons path, and perform the migration, and get such module (and its dependants) uninstalled. That's too much IMO.

royle-vietnam:
> I don't think the uninstallation should be done automatically, as it's a very hard and dangerous step, and also you can't recover from that step. For example, you forget to add a module to your addons path, and perform the migration, and get such module (and its dependants) uninstalled. That's too much IMO.

I think deleting junk data will bring more benefits than forgetting to add a module to your addons path
Also, when we do it wrong, we will need to do it again with backup database another

pedrobaeza:
They are not junk, and you have to think in a cross version migration: you don't have module X in version 14, but your target is version 15, and you have such module in version 15. With this, you are uninstalling it in version 14 with no possibility of continuing. Besides, you have database_cleanup module for such cleaning (or later manual uninstallation).

legalsylvain (changes requested):
Hi. Uninstalling unavailable modules is not possible because in a migration of various versions, modules can be anavailable in intermediate steps.

The objective of openupgrade is just to migrate.
You should use database_cleanup at the end, when your migration is ok.

Edit : sorry duplicate with pedro.

hoangtiendung070797:
Thanks, i understand

PR #3162, opened 2022-03-02
[13.0][OU-FIX] base,hr,product,website_slides: Protect image conversion

pedrobaeza:
There can be dangling attachments that belong to non existing records. There are also another possibilities of the write operation fails.

Protecting this conversion, we avoid unneeded crashes.

NOTE: This method may be put as openupgradelib function.

@Tecnativa

### DIFF
--- a/addons/hr/migrations/13.0.1.1/post-migration.py
+++ b/addons/hr/migrations/13.0.1.1/post-migration.py
@@ -1,6 +1,10 @@
 # Copyright 2020 Payam Yasaie <https://www.tashilgostar.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from openupgradelib import openupgrade
+import logging
+
+
+_logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


addons/hr/migrations/13.0.1.1/post-migration.py
```diff
@@ -15,9 +19,16 @@ def convert_image_attachments(env):
             ('res_id', '!=', False),
         ])
         for attachment in attachments:
-            # for not having dangling attachments
-            attachment.res_field = "image_1920"
```
MiquelRForgeFlow, line 19: why delete this?

MiquelRForgeFlow (approved):

addons/hr/migrations/13.0.1.1/post-migration.py
pedrobaeza: Because it's not following the preservation principle and it's not done in the rest. There's no extra space used, being the same file with the same hash.

PR #3167, opened 2022-03-03
[FIX][14.0] calendar: an KeyError occurred while creating recurrent events

royle-vietnam:
Steps:
1. Create an meeting activity that link to recurrent event on project.task
2. Run migration
3. An KeyError occurred while creating recurrent events

Because In v14, now all occurrences of recurrent events are created as real records, not virtual ones, so we need to regenerate them for all the existing ones.
But while creating real records, Odoo will try to create an activity on the related record. But project.task is not in self.env because the module calendar doesn't depend on the module project, so models not installed (is not in self.env)

Solution: we do not create an activity on the real records. An activity that has been link to real record.

### DIFF
--- a/openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
@@ -70,9 +70,10 @@ def fill_calendar_recurrence_table(env):
 def create_recurrent_events(env):
     """In v14, now all occurrences of recurrent events are created as real records, not
     virtual ones, so we need to regenerate them for all the existing ones.
+    But we do not create an activity on the real records.
     """
     recs = env["calendar.recurrence"].search([("base_event_id", "!=", False)])
-    recs._apply_recurrence()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't the solution be to do this on end-migration?

royle-vietnam:
> Shouldn't the solution be to do this on end-migration?

No, because error occurs when we call method `_apply_recurrence`

pedrobaeza:
That error won't happen as end migration has all models loaded.

royle-vietnam:
> That error won't happen as end migration has all models loaded.

ok, I got your point. let me try

pedrobaeza:
The question is if the activity must be created or not. I don't know such functional case.

royle-vietnam:
> The question is if the activity must be created or not. I don't know such functional case.

On v14, when creating a recurrent event, only 1 activity is created, so I think it should not be created to look like v14
If we use end-migration, it will create a lot of activity, I don't think it's necessary

pedrobaeza:
OK then

pedrobaeza (approved):

PR #3170, opened 2022-03-04
[13.0][FIX] l10n_es: parent taxes should be of l10n_es module

MiquelRForgeFlow:
I just found a rare case where a client had custom parent taxes, so this PR is just to avoid having errors.

### DIFF
--- a/addons/l10n_es/migrations/13.0.4.0/post-migration.py
+++ b/addons/l10n_es/migrations/13.0.4.0/post-migration.py
@@ -43,7 +43,6 @@ def use_new_taxes_and_repartition_lines_on_move_lines(env):
         [('children_tax_ids', '!=', False), ('company_id', 'in', companies_ids)]
     ).filtered(lambda t: not t.invoice_repartition_line_ids
                and not t.refund_repartition_line_ids)
-    children_tax_ids = taxes_with_children.mapped('children_tax_ids').ids
     tax_ids = taxes_with_children.ids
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This doesn't serve if you have extension modules for `l10n_es`, like `l10n_es_irnr` (that ones don't have children taxes, but it's an example of possible wrong thing). If the company has a compatible plan (child of l10n_es_common), then this should be applied.

pedrobaeza:
That trick is still limiting. Better to check only companies with the Spanish CoA, and taxes with XML-ID.

MiquelRForgeFlow:
> companies with the Spanish CoA

That is handled already in above the affected code.

pedrobaeza:
OK, then the problem is that the taxes were manually added?

MiquelRForgeFlow:
Yes. I need to exclude the `__export__` ones.

pedrobaeza:
OK, then please reflect on the commit message these conditions for better later acknowledge.

MiquelRForgeFlow:
I mean, if they added manually, they should handle manually the link between other taxes/move lines. The code should avoid always handle manual data when there is "complex" code.

pedrobaeza:
Yeah, what I'm saying is that you should dump all these explanations in the commit message.

MiquelRForgeFlow:
Ready

pedrobaeza (approved):

PR #3173, opened 2022-03-09
[13.0][FIX] account: make us of merge_models openupgradelib method

MiquelRForgeFlow:
- Consider several models like mail.template, ir.filters, ...
- Consider model_id fields

Note: I am considering adding this method in openupgradelib. What do you thing?


legalsylvain:
> Note: I am considering adding this method in openupgradelib. What do you thing?

+1 to propose a generic function in openupgradelib.

MiquelRForgeFlow:
> +1 to propose a generic function in openupgradelib.

Done in https://github.com/OCA/openupgradelib/pull/283.

MiquelRForgeFlow:
Ready

PR #3177, opened 2022-03-14
[IMP][14.0]account: correct account type

hieu1211:
Because some account templates type have changed in version 14, so i think we should change the account type to match the account template type.
Ex: module l10n_vn, file data account.account.template.csv
in v13:
`"chart141","Tạm ứng","141","account.data_account_type_current_assets","l10n_vn.vn_template","True"
`
in v14:
`"chart141","Advances",141,"account.data_account_type_receivable","vn_template","True"
`
In Odoo, account.account data is generated from account.account.template data by method 'generate_account', they are not automatically updated following account.account.template, so we need update them using sql.

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/end-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/end-migration.py
@@ -1,0 +1,38 @@
+from openupgradelib import openupgrade
+
+
+def _make_correct_account_type(env):
+    query = """
+        UPDATE account_account as ac
+            SET user_type_id=aat.user_type_id
+        FROM account_account_template as aat
+            JOIN account_chart_template as act
+                        ON aat.chart_template_id = act.id
+            JOIN res_company as c
+                        ON c.chart_template_id = act.id
+        WHERE ac.code =
+                CASE
+                    WHEN
+                        act.code_digits < LENGTH(aat.code) THEN aat.code
+                    ELSE
+                        CONCAT(aat.code,
+                            REPEAT('0',act.code_digits - LENGTH(aat.code)))
+                END
+            AND ac.user_type_id != aat.user_type_id
+            AND ac.company_id = c.id;
+        UPDATE account_account as ac
+            SET internal_type=at.type,
+                internal_group=at.internal_group
+        FROM account_account_type as at
+        WHERE ac.user_type_id=at.id AND (ac.internal_type != at.type OR
+                                    ac.internal_group != at.internal_group);
+        """
+    openupgrade.logged_query(
+        env.cr,
+        query,
+    )
+
+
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you specify which have changed and why it should be handled this way?

hieu1211:
> Can you specify which have changed and why it should be handled this way?

i updated, thanks!

tranngocson1996:
@pedrobaeza when upgrade database 13 to 14, type of account not updated yet. So need to write migration code to update type follow by new COA. Handling in the account module will be true for all COA

openupgrade_scripts/scripts/account/14.0.1.1/end-migration.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def _make_correct_account_type(env):
+    query = """
+        UPDATE account_account as ac
+            SET user_type_id=aat.user_type_id
+        FROM account_account_template as aat
+            LEFT JOIN account_chart_template as act
+                        ON aat.chart_template_id = act.id
+            LEFT JOIN res_company as c
```
MiquelRForgeFlow:
```suggestion
            JOIN account_chart_template as act
                        ON aat.chart_template_id = act.id
            JOIN res_company as c
```

openupgrade_scripts/scripts/account/14.0.1.1/end-migration.py
```diff
@@ -0,0 +1,37 @@
...
+                        CONCAT(aat.code,
+                            REPEAT('0',act.code_digits - LENGTH(aat.code)))
+                END
+            AND ac.user_type_id != aat.user_type_id
+            AND ac.company_id = c.id;
+        UPDATE account_account as ac
+            SET internal_type=at.type,
+                internal_group=at.internal_group
+        FROM account_account_type as at
+        WHERE ac.user_type_id=at.id;
```
MiquelRForgeFlow:
```suggestion
        WHERE ac.user_type_id=at.id AND (ac.internal_type != at.type OR ac.internal_group != at.internal_group);
```

hieu1211: Thanks for reviewing!

MiquelRForgeFlow (approved):
LGTM

tranngocson1996 (approved):

pedrobaeza:
This may happen on each version, not only on v13, but this can't be merged here in OU. We have `account_chart_update` for this kind of changes, and this mustn't be something that it's done unadvertised to accountant users, as there can be strong reasons for changing an account type linked to an account template.

Due to these reasons, I'm closing it, but if you want to use it on your upgrades, just add this PR to your deployment.

PR #3178, opened 2022-03-15
[FIX][14.0]hr_timesheet: fix noupdate rule

hieu1211:
Because noupdate rules has been changed from False to True, so we need migration it in post instead of pre.

### DIFF
--- a/openupgrade_scripts/scripts/hr_timesheet/14.0.1.0/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/hr_timesheet/14.0.1.0/noupdate_changes.xml
@@ -37,9 +37,9 @@
                     ('project_id.allowed_internal_user_ids', 'in', user.ids),
                     ('task_id.allowed_user_ids', 'in', user.ids)
             ]</field>
-    <!-- <field name="perm_create"/>
+    <!-- <field name="perm_create"/> -->
     <field name="perm_read"/>
-    <field name="perm_unlink"/>
+    <!-- <field name="perm_unlink"/>
     <field name="perm_write"/> -->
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why? Is there any change in standard that must be reflected before changing to noupdate=True?

hieu1211:
> Why? Is there any change in standard that must be reflected before changing to noupdate=True?

Yes, rules were changed so we need update it before set noupdate=True

pedrobaeza:
Indeed they are changed from:

https://github.com/odoo/odoo/blob/0dd7c78b06726543825008a48d3dc004d2af4557/addons/hr_timesheet/security/hr_timesheet_security.xml#L23

to:

https://github.com/odoo/odoo/blob/64af64735f71a9dbe852c7c222e39e482bd09838/addons/hr_timesheet/security/hr_timesheet_security.xml#L33

But this should be covered by the noupdate_changes.xml load:

https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/hr_timesheet/14.0.1.0/noupdate_changes.xml#L31

Maybe we need to uncomment the `perm_read` field part.

pedrobaeza:
Please check the rest of the records.

hieu1211:
thanks, i updated code!

royle-vietnam (approved):

pedrobaeza:
Have you checked the rest?

royle-vietnam:
> Have you checked the rest?

Yes, I checked

hieu1211:
> Have you checked the rest?

Yes, its ok

PR #3180, opened 2022-03-16
[FIX][14.0] account: correct payment's currency

royle-vietnam:
Steps:
1. Create a customer invoice with USD
2. Register payment with EUR
3. A payment with EUR and move with USD are created
4. When migrated to add move_id field to account_payment, then payment's currency is not correct, because payment's currency is inherit from move

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
@@ -158,6 +158,7 @@ def copy_fields(env):
                 ("journal_id", None, None),
                 ("name", None, None),
                 ("payment_date", None, None),
+                ("currency_id", None, None),
             ],
         },
     )
@@ -273,16 +274,6 @@ def add_move_id_field_account_bank_statement_line(env):
 
 
 def add_move_id_field_account_payment(env):
-    openupgrade.logged_query(
-        env.cr,
-        """
-        UPDATE account_payment ap
-        SET currency_id = COALESCE(aj.currency_id, rc.currency_id)
-        FROM account_journal aj
-        JOIN res_company rc ON aj.company_id = rc.id
-        WHERE ap.journal_id = aj.id
-        """,
-    )
     if not openupgrade.column_exists(env.cr, "account_payment", "move_id"):
         openupgrade.logged_query(
             env.cr,
@@ -315,7 +306,7 @@ def add_move_id_field_account_payment(env):
         UPDATE account_payment ap
         SET move_id = am.id
         FROM account_move am
-        WHERE am.payment_id = ap.id""",
+        WHERE ap.move_id IS NULL AND am.payment_id = ap.id""",
     )
     # 2. try to match on payment move_name or payment_reference (if move_name empty)
     openupgrade.logged_query(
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -273,16 +274,6 @@ def add_move_id_field_account_bank_statement_line(env):
 
 
 def add_move_id_field_account_payment(env):
-    openupgrade.logged_query(
-        env.cr,
-        """
-        UPDATE account_payment ap
-        SET currency_id = COALESCE(aj.currency_id, rc.currency_id)
```
MiquelRForgeFlow, line 280: Don't. account.payment has a computed stored field: https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_payment.py#L396

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -352,6 +343,17 @@ def add_move_id_field_account_payment(env):
         WHERE am.id = ap.move_id
         """,
     )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move am
+        SET currency_id = ap.{}
+        FROM account_payment ap
+        WHERE am.id = ap.move_id
```
MiquelRForgeFlow:
```suggestion
        WHERE am.id = ap.move_id AND ap.{} is not null
```

royle-vietnam: Unnecessary, `currency_id` is required on `account.payment` on Odoo 13

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
royle-vietnam: No, `currency_id` is inherited from `account.move`, I updated it at line 346

MiquelRForgeFlow: It's not inherited: https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_payment.py#L92

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
MiquelRForgeFlow: Sorry, I meant

```suggestion
        WHERE am.id = ap.move_id AND am.currency_id != ap.{}
```

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
royle-vietnam: In essence, it inherits from `account.move`, so it only needs to be updated at line 346

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
royle-vietnam: Ok, done, thank you

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
MiquelRForgeFlow: In line 346, you are updating the `am.currency_id`, nothing to do with `ap.currency_id`, which this query handles. From a database point of view, `am.currency_id` and `ap.currency_id` are two different columns in two different tables. This query you are deleting assures the value of `ap.currency_id` is the same as the value of the compute I linked before.

MiquelRForgeFlow:
FYI: https://github.com/OCA/OpenUpgrade/pull/3226

pedrobaeza:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3226

PR #3182, opened 2022-03-17
[FIX][14.0] calendar: fix duplicated events

royle-vietnam:
Steps:
1. Create a recurrency event, on 03/16/2022 (wednesday),
- Weekly repeating on monday, wednesday, friday
2. Run migration script

Current: Create new events on 03/14/2022 (monday) and on 03/16/2022 (wednesday)
Expected: No event is created

### DIFF
--- a/openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
@@ -73,7 +73,44 @@ def create_recurrent_events(env):
     But we do not create an activity on the real records.
     """
     recs = env["calendar.recurrence"].search([("base_event_id", "!=", False)])
-    recs.with_context(default_activity_ids=[(6, 0, [])])._apply_recurrence()
+    env.cr.execute(
+        """
+        SELECT id, recurrent_id, recurrent_id_date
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/calendar/14.0.1.0/post-migration.py
```diff
@@ -73,7 +73,44 @@ def create_recurrent_events(env):
     But we do not create an activity on the real records.
     """
     recs = env["calendar.recurrence"].search([("base_event_id", "!=", False)])
-    recs.with_context(default_activity_ids=[(6, 0, [])])._apply_recurrence()
+    env.cr.execute(
+        """
+        SELECT id, recurrent_id, recurrent_id_date
+        FROM calendar_event
+        WHERE recurrent_id > 0
```
flotho, line 80: Hi, in SQL isn't ```IS NOT NULL ``` syntax better ?

flotho (approved):
code review, great work,
thanks

pedrobaeza (approved):
I finally had the time to check this, and I understand the motivation of this code. Maybe it seems a bit bloated, but I don't see either a better option to do it. Thanks for it.

PR #3152, opened 2022-02-25
[13.0][ADD] hr_expense: Added a function that migrates the tax tags (accoun…

ddejong-therp:
…t.account.tag) of move lines (account.move.line) that belong to an expense (hr.expense).

The account module does not migrate tags for these lines.

PR #3183, opened 2022-03-17
[13.0][FIX] account: Added extra queries to migrate tags from non-invoice move lines wherever possible.

ddejong-therp:
Sometimes the tags of a move line don't get migrated properly. Currently it only migrates them if the move line belongs to an invoice. However, almost all other cases are forgotten. I've wrote a query that fixes it for move lines that are the result of an hr.expense in https://github.com/OCA/OpenUpgrade/pull/3152 . However, that still doesn't account for all move lines.

I've added queries so that the tax tags of any move lines that have tax_line_id set, can now also migrate. I've left the previous queries in there, as they use the tax tags info of their corresponding invoice, which I guess is still nice to do. However, for those lines which do not have an corresponding invoice, they'll have their tags migrated now too.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -905,6 +905,32 @@ def move_tags_from_taxes_to_repartition_lines(env):
 
 
 def assign_tax_repartition_line_to_move_lines(env):
+    # Use tax_line_id to find repartition line by default
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_move_line aml
+        SET tax_repartition_line_id = atrl.id
+        FROM account_move_line aml2
+        JOIN account_move am ON aml2.move_id = am.id
+        JOIN account_tax_repartition_line atrl ON (
+            balance >= 0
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):
LGTM (not tested, but makes sense)

pedrobaeza:
I think this should be restricted to those lines that don't have the field filled, isn't it?

ddejong-therp:
> I think this should be restricted to those lines that don't have the field filled, isn't it?

You mean the tax_line_id field? The problem is that tax_repartition_line_id field does not get set for some move lines. I use tax_line_id to find an appropriate repartition line id.

ddejong-therp:
Or do you mean the lines which don't have tax_repartition_line_id set?

MiquelRForgeFlow:
I think he refers to type = 'entry'? :S

MiquelRForgeFlow:
> I think this should be restricted to those lines that don't have the field filled, isn't it?

Prior to this method, the field tax_repartition_line_id is empty, so don't need that restriction. About putting it in the following queries beside the first one, I think it's superfluous, as the JOINs already do the filter. Each query handle different move lines.

PR #3184, opened 2022-03-17
[13.0][ADD] l10n_nl: Added migration script for the Dutch localization module.

ddejong-therp:
I've added a migration script for the `l10n_nl` module. 

Without this migration script, each move line and repartition line will have too many tags (account.account.tag). The script removes the tags where they are not needed. Moreover, l10n_nl started using account.tax.report.line's in 13.0, which are responsible for creating the tags now (instead of them being defined in xml files). This script transforms all the tags to their new corresponding tags that were created by the report line.
It currently uses helper functions which I've PR'ed to openupgradelib: https://github.com/OCA/openupgradelib/pull/286 (so only merge this after that PR is merged as well)

I will be posting screenshots to demonstrate what I mean with lines having 'too many tags'.

### DIFF
--- a/addons/l10n_nl/migrations/13.0.3.0/post-migration.py
+++ b/addons/l10n_nl/migrations/13.0.3.0/post-migration.py
@@ -1,0 +1,150 @@
+from openupgradelib import openupgrade, openupgrade_130
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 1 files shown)


gjotten (approved):
I can confirm this working very well on our v10 database functionally, when migrating it to v14.

NL66278 (approved):
:+1: LGTM

PR #3185, opened 2022-03-18
[13.0][FIX] account: Have taxes that have children have their tags migrated too.

ddejong-therp:
The migration script of the `account` module does something where it creates `account.tax.repartition.line`'s for each `account.tax`. However, for some reason I could not deduce, it only does so for the `account.tax` records that don't have children. This effectively means that not **all** `account.tax` records will get repartition lines. Again, I don't know what the reasoning behind this was, but if someone has some input on this it is much appreciated.

However, as for now, I'm of the opinion that this is unnecessary. If not all tax records get repartition lines, than not all tags (`account.account.tag`) will be migrated either, as they will be connected to the repartition lines from 13.0 and onwards.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -750,14 +750,12 @@ def try_to_fill_account_account_tag_country_id(env):
 def create_account_tax_repartition_lines(env):
     # account_tax_repartition_line
     all_taxes = env['account.tax'].with_context(active_test=False).search([])
-    taxes_with_children = env['account.tax'].with_context(active_test=False).search(
-        [('children_tax_ids', '!=', False)])
     # taxes_children = taxes_with_children.mapped('children_tax_ids')
     complete_taxes = env['account.tax.repartition.line'].search(
         [('invoice_tax_id', '!=', False)]).mapped('invoice_tax_id') | env[
         'account.tax.repartition.line'].search(
         [('refund_tax_id', '!=', False)]).mapped('refund_tax_id')
-    tax_ids = (all_taxes - taxes_with_children - complete_taxes).ids
+    tax_ids = (all_taxes - complete_taxes).ids
     if tax_ids:
         openupgrade.logged_query(
             env.cr, """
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
> However, for some reason I could not deduce, it only does so for the `account.tax` records that don't have children. This effectively means that not all `account.tax` records will get repartition lines. Again, I don't know what the reasoning behind this was, but if someone has some input on this it is much appreciated.

The case for `account.tax` records that have children should be handled in the corresponding `l10n_*` module. Basically, because in v13, in many `l10n_*` modules, the child taxes have been fused with their parent taxes, and thus, we cannot do something too generic here in `account` like what you propose for parent taxes.

ddejong-therp:
Ahhh, well thank you for the input then.

ddejong-therp:
I guess, I will be making a PR for l10n_nl then soon. :)

ddejong-therp:
Ok, so it's has been a while since I have actually gotten back into this issue.

However, for `l10n_nl`, I would want the tax tags of `account.tax` with children to be migrated in the exact same way as all other `account.tax`. In order to do that in `l10n_nl`, I would have to use almost all queries/logic from the following functions in `post-migration.py` from the `account` module:
```
create_account_tax_repartition_lines
move_tags_from_taxes_to_repartition_lines
assign_tax_repartition_line_to_move_lines
assign_tax_repartition_line_to_move_lines
assign_account_tags_to_move_lines
```
Now, that is a lot of code to duplicate, which is kinda silly.

How about I just alter the `account` `post-migration.py` script to read from a `ir.config_parameter` to see which other `account.tax` records should have its tags migrated too (in addition to records without children). I think that will be the cleanest solution to be honest. Similarily as what is done here: https://github.com/OCA/OpenUpgrade/blob/4fa1430655c68300cb046f96e2943e6ff7ac2fe3/addons/project/migrations/11.0.1.1/post-migration.py#L38
I'm wondering if this is going to be a good idea though, before I invest the time to fix this.

StefanRijnhart:
@ddejong-therp you can see here that in l10n_nl, the child taxes are eliminated: https://github.com/odoo/odoo/commit/3936d655c4bfd845b6ead724cecdc8e7c62816eb#diff-de6184d4eb5e7e2450afafb9d22046721e4931216a636ec6b945ea27a1488591

PR #3189, opened 2022-03-23
[11.0][FIX] hr_timesheet: Migrate the users of the timesheets' old "Officer"…

ddejong-therp:
… group to the new "Manager" group.

In odoo 10 xmlid group_hr_timesheet_user is the timesheet "Officer" group, while in Odoo 11 the same xmlid is used to represent a regular user, The new xmlid group_timesheet_manager is from then on used as the timesheet "Manager" group.

However, this causes the users that were assigned to group group_hr_timesheet_user in 10.0, to remain assigned to the new group_hr_timesheet_user in 11.0. I believe the manager group in 11.0 is what replaces the old officer group. Nothing is currently being done to assign them to the new group_timesheet_manager group, so that is what this PR does.

### DIFF
--- a/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
+++ b/addons/hr_timesheet/migrations/11.0.1.0/post-migration.py
@@ -36,6 +36,23 @@ def migrate_project_issue_sheet(env):
     )
 
 
+def migrate_timesheet_managers(env):
+    """The group with xmlid group_hr_timesheet_user used to be "Officer"
+    group in 10.0. This is now replaced with the "Manager" group, while
+    the xmlid group_hr_timesheet_user is now used for the "User" group.
+    This moves all users of the "Officer" to the "Manager" group.
+    """
+    user_group_id = env.ref('hr_timesheet.group_hr_timesheet_user').id
+    manager_group_id = env.ref('hr_timesheet.group_timesheet_manager').id
+    root_user_id = env.ref('base.user_root').id
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE res_groups_users_rel SET gid = %s
+        WHERE gid = %s AND uid != %s""",
+        (manager_group_id, user_group_id, root_user_id)
+    )
+
+
 def recompute_tasks_from_issues_fields(env):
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
Don't you find safer migrate to v11 without this fix? I mean, after migration without this fix, all users in that group get right restrictions, and if later you want a few of them be HR manager, then do you put them to manager manually. With this fix, after migration, all those users happen to be HR manager, and then you are going to check whether several of them should be HR users instead of managers and put them to HR user group manually, but meanwhile, those users may connect and have accesses that maybe should not have. In fact, I view it as an error of v10, having an HR user group with HR manager rights, and later in v11 it gets fixed. Just my opinion.

StefanRijnhart:
I think there is a point to this change. Timesheet 'officers' in Odoo 10 had access to the [timesheet configuration](https://github.com/odoo/odoo/commit/a806f96fa267#diff-e02ba1b54ada5d33764fadf0f3cf20139379649c7b08c54dcc6a31055f0ff1afL3-R4), and were added to the group of HR officers (hr.group_hr_user) as per implied_ids. No special permissions were required in Odoo to access the timesheet menus.

In Odoo 11, `base_group_user` was not enough anymore for user access to timesheets, that's what group_hr_timesheet_user is for now, and the old function of that group shifted to group_timesheet_manager (which also inherits from group_hr_user).

So it seems to me that group_hr_timesheet_user needs to be renamed to group_timesheet_manager and all users for which there are timesheets in the system may need to be added to hr_timesheet_user.

Without such change, it seems that timesheets are hardly usable after the migration to Odoo 11 without a tedious manually moving around of users between the groups.

pedrobaeza:
So do you consider this PR correct, Stefan?

StefanRijnhart:
Not entirely! With this change, users that used to be able to manage timesheets in 10.0 will still be able to do so in 11.0, but no ordinary user will be able to write their timesheets like they used to.

pedrobaeza:
OK, @ddejong-therp can you please perform the commented changes?

victoralmau:
Superseed by: https://github.com/OCA/OpenUpgrade/pull/3726

PR #3191, opened 2022-03-25
[14.0][FIX] base: assure existing users can export

MiquelRForgeFlow:
To maintain same behavior as previous versions.

Also, one expect users be in this export group due to:
Selection_282


pedrobaeza (commented):
If I understand correctly, you are putting the export allowed group as inherited to internal use group, but with that, all internal users will always have the export group, which is not the desired behavior. Only legacy ones should have it, so you have to add to existing internal users manually the export group.

MiquelRForgeFlow:
Fixed, sorry, I took the wrong table hehe

MiquelRForgeFlow:
Ready

pedrobaeza (approved):

PR #3199, opened 2022-03-26
[12.0][IMP] mrp: Pre-compute mrp.production>product_uom_qty by SQL

pedrobaeza:
It can take a lot of time when you have a considerable number of production orders.

There's an assumption here when a conversion must be done to round to 4 decimals, as UoM expresses rounding through the rounding precision instead, but the selected fixed precision should be OK for most cases.

@Tecnativa


MiquelRForgeFlow:
> as UoM expresses rounding through the rounding precision instead

Why not use this precision in the script? Alternatively, we could left this field empty when the precision is different than 4 and let ORM compute it in those few cases.

pedrobaeza:
It doesn't work the same way. What we have in the UoM is for example 0.01, when we need `2` for passing to `round`.

MiquelRForgeFlow:
but transforming `0.01` to `2` is not too complicated, right?

pedrobaeza:
You can also put `0.05` as rounding...

MiquelRForgeFlow (approved):

PR #3209, opened 2022-03-28
[14.0][MIG] website_google_map

MiquelRForgeFlow:


legalsylvain (commented):
the migration_analysis file is absent. Could you generate it ?

StefanRijnhart (approved):
@legalsylvain is it necessary to generate the analysis for this module? There are no models or noupdate records in this module. Isn't that why the analysis was not generated in the first place?

legalsylvain:
Hi. 

@StefanRijnhart : 
- ``website_google_map`` exist in V13 and in V14.
- no folder has been generated during the analysis. See : https://github.com/OCA/OpenUpgrade/tree/14.0/openupgrade_scripts/scripts ``website_google_map`` is not present
- In the other side, for other module like ``barcode``, that is in the same situation (exist in both versions) + nothing has changed. An empty file is generated : https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/barcodes/14.0.2.0/upgrade_analysis.txt

My question is : why migration folder + file is not generated for ``website_google_map`` and generated for ``barcodes``. 
As the original developper of the module ``upgrade_analysis``, do you have an idea ? (otherwise, I'll take a look ...)

In fact, it's not related to that current PR, so OK to move forward ! Sorry for having blocking this PR.

legalsylvain:
merging trivial PR. Thanks @MiquelRForgeFlow.

StefanRijnhart:
@legalsylvain I think this is due to an implementation detail that may have changed over time. If there are no models and no change in the set of XML records, I am not surprised that there is no analysis but I agree it would probably better if there was.

StefanRijnhart:
And thanks for the merge!

PR #3210, opened 2022-03-28
[13.0][FIX] account: add same name condition for invoice line matching

MiquelRForgeFlow:
I have detected some mismatching in a client database. It seems the conditions are not enough. Let's hope that adding the `name` condition is enough.

Selection_286

I was also considering adding a comparison between `aml.balance` and `ail.price_subtotal`, but it's tricky, as the currency may not match and thus a conversion and round maybe needed.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -258,10 +258,23 @@ def migration_invoice_moves(env):
         env.cr,
         query.format(
             where=(sql.SQL(minimal_where + """
+            AND ail.name = aml.name
             AND ail.account_id = aml.account_id
             AND ai.commercial_partner_id = aml.partner_id
             AND ((ail.account_analytic_id IS NULL AND aml.analytic_account_id IS NULL)
                 OR ail.account_analytic_id = aml.analytic_account_id)"""))
+        ),
+    )
+    openupgrade.logged_query(
+        env.cr,
+        query.format(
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
Tested, and works like a charm.

pedrobaeza:
I don't get why restricting the condition works better.

MiquelRForgeFlow:
> I don't get why restricting the condition works better.

Well, I have client database that in a same move/invoice, the move lines/invoice lines have all the same conditions but their name is different (they come from different PO), and they also have different `aml.balance/ail.price_subtotal` because they have different taxes associated. Adding the name condition, the match is done correctly.

pedrobaeza:
I'm a bit reluctant to add more criteria to the stricter one, as there can be a lot of reasons to not match. Another valid concern is about the performance, as the operator `like` is not very performant. Have you measure differences? And finally, if several matches, the new SQL matches by order, so there are high chances that this order matches with the one in the invoice, and correct matching was done.

MiquelRForgeFlow:
The difference is not substantial. Now the strict query measures: 217839 rows affected after 0:56:25.622033 running

I had to do this as there were a lot of mismatching.

pedrobaeza:
And before using it?

pedrobaeza:
Is it the same criteria for all Odoo versions <=12? You can have account.move.line generated by for example Odoo 8.0?

pedrobaeza:
Discarding this as commented.

PR #3212, opened 2022-04-02
[FIX][14.0] account: fill amount residual currency on account move line

royle-vietnam:
Because currency_id on account_move_line is required on 14.0


pedrobaeza:
Better option here: #3234

PR #3096, opened 2022-01-14
[14.0][OU-ADD] sale_margin: Migration scripts

hieu1211:
This PR
======
migration done

PR #3217, opened 2022-04-06
[14.0][OU-ADD] sale_order_margin_percent: Merged sale_order_margin_percent to sale_margin. 

victoralmau:
Merged `sale_order_margin_percent` to `sale_margin`

@Tecnativa

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -63,6 +63,8 @@ merged_modules = {
     "pos_kitchen_printer": "pos_restaurant",
     "pos_reprint": "point_of_sale",
     "website_theme_install": "website",
+    # OCA/margin-analysis
+    "sale_order_margin_percent": "sale_margin",
     # OCA/partner-contact
     "base_vat_sanitized": "base_vat",
     "partner_bank_active": "base",


pedrobaeza:
@sergio-teruel I think the `read_group` part is missing on core for proper percentage average, isn't it? Should we keep the OCA module?

sergio-teruel:
@pedrobaeza This module has not add code to report, so is OK.

But @victoralmau please you must create a pre-migration script to pre-create the fields and compute data by sql queries because the percent fields in OCA module are computed and in Odoo core are store. This will do that in a database migration we have a great performance.

victoralmau:
> @pedrobaeza This module has not add code to report, so is OK.
> 
> But @victoralmau please you must create a pre-migration script to pre-create the fields and compute data by sql queries because the percent fields in OCA module are computed and in Odoo core are store. This will do that in a database migration we have a great performance.

Ok, I think that's a good suggestion. Checking `sale_margin` addon I see that this improvement is already implemented at https://github.com/OCA/OpenUpgrade/pull/3096.

sergio-teruel (approved):

sergio-teruel:
Thanks @victoralmau

PR #3223, opened 2022-04-21
[14.0][FIX] account: avoid duplicating account groups

MiquelRForgeFlow:
NOTE: If some group has been tweaked from v13 to v14, this script will treat both versions (v13 and v14) of this group as different groups. Then, the merge of them should be done in the corresponding `l10n_*` module.

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
@@ -429,6 +429,20 @@ def fill_account_payment_partner_id(env):
     )
 
 
+def delete_xmlid_existing_groups(env):
+    env.cr.execute(
+        """DELETE FROM ir_model_data imd
+        USING account_group ag
+        WHERE ag.id = imd.res_id AND imd.model = 'account.group'
+            AND imd.module != '__export__'
+        RETURNING imd.res_id"""
+    )
+    # end-migration script will:
+    # - populate account groups from the templates
+    # - unfold manual groups per company + proper company assignation
+    # - merge repeated groups
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.set_xml_ids_noupdate_value(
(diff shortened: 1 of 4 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
```diff
@@ -133,10 +133,10 @@ account      / account.group            / code_prefix (char)            : DEL
 account      / account.group            / code_prefix_end (char)        : NEW
 account      / account.group            / code_prefix_start (char)      : NEW
 # DONE: pre-migration: renamed code_prefix to code_prefix_start
-# post-migration: assigned same start values to end field
+# DONE: post-migration: assigned same start values to end field (why?)
```
StefanRijnhart: Maybe because in 13.0 there is no concept of an account group 'range', so this is the only valid value upon migration? And the upstream model takes much care to keep the values in sync if only one of both is set, in its onchange and write and create methods.

StefanRijnhart:
```suggestion
# DONE: post-migration: assigned same start values to end field

**Review by StefanRijnhart — APPROVED** [review]

File: openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py [hunk 2]
```diff
@@ -416,6 +416,20 @@ def fill_account_payment_partner_id(env):
     )
 
 
+def delete_xmlid_existing_groups(env):
+    env.cr.execute(
```
pedrobaeza, line 433: Why removing this? You are breaking possible import templates.

StefanRijnhart: I think it makes sense because of the auto generation logic in combination with company expansion that modules do not add xml records for this. Are there examples of such XML records?

pedrobaeza: OK, nevermind. I was thinking about importing accounts, but the group is now auto-computed, so now need to reference it from the import.

openupgrade_scripts/scripts/account/14.0.1.1/upgrade_analysis_work.txt
MiquelRForgeFlow: done

StefanRijnhart: Thanks!

PR #3232, opened 2022-05-03
[IMP] set web_diagram as merged into web to avoid to uninstall it manually

legalsylvain:
**rational**

- ``web_diagram`` exists in odoo V13 and is auto_installable
- ``web_diagram`` doesn't exist in V14
(https://github.com/OCA/OpenUpgrade/blob/14.0/docsource/modules130-140.rst)

So before that patch, we should uninstall manually that module when migrating from 13 to 14.

**question**
However, I don't know if ``web_diagram`` has been dropped, or has moved into EE. If yes, is this PR OK ? I don't think that the objective of ``openupgrade`` project is to migrate EE instance, but maybe I'm wrong...

comments / alternative welcome.

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
@@ -7,3 +7,10 @@ from openupgradelib import openupgrade
 def migrate(env, version):
     """Call disable_invalid_filters in every edition of openupgrade"""
     openupgrade.disable_invalid_filters(env)
+    # web_diagram has been remove in V14
+    # we merge into web, if no diagram are present, to avoid to
+    # have to uninstall the module manually
+    if not env["ir.ui.view"].search([("type", "=", "diagram")]):
+        openupgrade.update_module_names(
+            env.cr, [("web_diagram", "web")], merge_modules=True
+        )


pedrobaeza:
I prefer to not do such merge, as the module doesn't have a replacement as is. Most of the installations don't make use of the diagram view, but imagine you have a module that is using it, and doing this merge will put the module automatically as not absent, and its dependencies redirected to `web`. And maybe someone in OCA wants to migrate `web_diagram` module to upper versions, and with this, you are preventing it from a smooth migration path.

MiquelRForgeFlow:
FYI https://github.com/odoo/odoo/commit/362065654205df908e210ff45d40d2b502241d0e

They completely removed it, they didn't do a merge.

legalsylvain:
I understand your point of view. Does it mean that for each migration from 13 to 14, each people should manually uninstall this module ? 
Do you see any alternative ?
Idea : in the end-migration.py, if the module is not available in the addons_path, uninstall it. 

Note : the commit mentionned by @MiquelRForgeFlow says " [REM] web_diagram: remove unused module". So no chance that the module will be ported in 14 ?

pedrobaeza:
You can check if there's any remaining ir.ui.view with such type, and only merge it if there isn't any.

legalsylvain:
> You can check if there's any remaining ir.ui.view with such type, and only merge it if there isn't any.

:+1: Thanks !

legalsylvain:
Hi @pedrobaeza : not sure to see exactly what you mean. 
I propose to add before that line https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/base/14.0.1.3/pre-migrate.py#L127
a request to know if there are web diagram views. if there are any, pop the key ``web_diagram`` in the ``merged_modules`` dict. Otherwise, let it.

Is this what you have in mind?

pedrobaeza:
Not exactly: in `base` end-migration, search for any `ir.ui.view` with type='diagram'. If not found, then manually call there:

```python
openupgrade.update_module_names(cr, [("web_diagram", "web")], merge_modules=True)
```

legalsylvain:
Thanks @pedrobaeza, I made the changes and it fixed my problem. (no need to uninstall ``web_diagram`` manually).

openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
```diff
@@ -7,3 +7,11 @@
 def migrate(env, version):
     """Call disable_invalid_filters in every edition of openupgrade"""
     openupgrade.disable_invalid_filters(env)
+
```
pedrobaeza: Can you please avoid empty extra lines inside a method?

legalsylvain: > Can you please avoid empty extra lines inside a method?


Done. Is it an OCA rules ? If yes, could be great to have a pre-commit for that point.

pedrobaeza: It's stated as advisable in PEP8, and my thumb rule is to not introduce such empty lines where the previous code didn't have them.

zurawsm3:
What does it **IMP** tag mean?
important?

pedrobaeza:
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

PR #3234, opened 2022-05-05
[14.0][OU-FIX] account: Fill amount_residual_currency as well

pedrobaeza:
Currency is now mandatory for all the `account.move.line` records, so there's certain part of the reconcile logic that relies on
`amount_residual_currency` when `currency_id` is set (independently of which currency it is). Example:

https://github.com/odoo/odoo/blob/1569441f15adb0b6b2920483ba8a43ccc7845e61/addons/account/models/account_move.py#L4401

so we copy the residual amount as well as the currency on the filling.

@Tecnativa TT35896


MiquelRForgeFlow (approved):
LGTM

PR #3235, opened 2022-05-06
[13.0] update upstream 02d5481c02f

MiquelRForgeFlow:
I added commit to update `noupdate_changes.xml`.


legalsylvain (approved):
Hi @MiquelRForgeFlow. LGTM. Thanks ! 

Note : I'd like to realize the same operation for previous odoo version. Could you say what are the git command you make to generate that branch ?

thank !

MiquelRForgeFlow:
> Could you say what are the git command you make to generate that branch ?

@legalsylvain 

1. I do a fetch of odoo:
   -  `git fetch odoo 13.0`
It will show something like this:
Selection_369
And thus I take this code to create the branch.

2. Once in the new branch, I do:
   -  `git merge odoo/13.0 --edit --no-commit`
And then review everything is ok and then I do a manual commit.

3. (optional) Check for noupdate_changes, run new_analysis, etc.

That's all.

PR #3237, opened 2022-05-06
[IMP][13.0] openupgrade_records : add logger if errors

legalsylvain:
Add logger when using ``openupgrade_records``.

for the time being the error is written in files, but if it's not possible to write in the file (because write access is not allowed) the error is skip.
this trivial PR fixes the problem.


MiquelRForgeFlow:
Hope it doesn't add too much noise in the logs.

jcdrubay (approved):
LGTM

legalsylvain:
> Hope it doesn't add too much noise in the logs.

Well, no errors, no log ! So in normal situation, it should not change anything. (only if there are trouble when writing files, due to write access bad configuration).

MiquelRForgeFlow (approved):

StefanRijnhart (approved):

PR #3238, opened 2022-05-06
Update broken link to documentation

farazbdm:
Update broken link to documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,4 @@
 OpenUpgrade
 ===========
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please use https://oca.github.io/OpenUpgrade/ instead

farazbdm:
> Please use https://oca.github.io/OpenUpgrade/ instead

revised

pedrobaeza:
Thanks for caring about this.

PR #3241, opened 2022-05-10
[13.0][FIX] mrp, repair: map locations

MiquelRForgeFlow:
The deletion of the stock obsolete xmlids should be done in end-migration, this way some modules will be able to use those xmlids in their migrations scripts.

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -343,15 +343,6 @@ def map_stock_locations(env, main_company):
         WHERE sq.location_id = sl.id""")
     env["stock.quant.package"].search([])._compute_package_info()
 
-    # xmlids are deprecated in v13
-    openupgrade.logged_query(env.cr, """
-    DELETE FROM ir_model_data imd
-    WHERE imd.module = 'stock' AND imd.name IN (
-        'property_stock_inventory', 'property_stock_production',
-        'stock_location_scrapped', 'location_inventory', 'location_production',
-        'location_procurement')
-    """)
-
 
 def stock_production_lot_multi_company_migration(env):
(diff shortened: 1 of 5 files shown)


addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -344,13 +344,7 @@ def map_stock_locations(env, main_company):
     env["stock.quant.package"].search([])._compute_package_info()
 
     # xmlids are deprecated in v13
```
pedrobaeza: Delete these lines directly without making any comment. You have work file for indicating that.

pedrobaeza (changes requested):
As always, these changes require to explain very well the why and what. Please, don't be lazy

addons/stock/migrations/13.0.1.1/post-migration.py
MiquelRForgeFlow: done

pedrobaeza:
I still don't see any comment on the commits, and barely some on the code indicating the what, but not the big whys.

MiquelRForgeFlow:
> I still don't see any comment on the commits,

Sorry, I didn't see your fist comment. Now all commits have more info.

pedrobaeza:
What do you mean by "obsolete locations"? Please specify more, with even specific examples.

MiquelRForgeFlow:
> What do you mean by "obsolete locations"? Please specify more, with even specific examples.

The stock_location_scrapped, location_inventory, location_production... locations without company that get replaced by the new ones generated with company.

pedrobaeza:
Then put all that information in the commit message

MiquelRForgeFlow:
Ready

pedrobaeza (approved):
Redaction of the commits for being more concise and explanatory can be still improved, but good enough. There's also duplicated code, but it doesn't worth other refactoring.

PR #2967, opened 2021-11-15
15.0 mig openupgrade framework

legalsylvain:
Hi all. 

I just executed meld on odoo subfolder to check the differences, to guess if the openupgrade_framework was requiring changes. 
At first sight : very fiew changes. 

**some changes related to base framework** (for information)

-  [IMP] core: introduce api.ondelete decorator  (https://github.com/odoo/odoo/commit/1c8a95809852baefa851985507a07ac7f6465dc7)
-  [IMP] core: allow set [value] in selection ondelete. (https://github.com/odoo/odoo/commit/cc768f39ebd71b89da8a0a2dc680f6725e2a4359)
-  [IMP] base: remove get_object helper (https://github.com/odoo/odoo/commit/fd67339a3ef754553bc787965fe8cedd720ce618)

**Some changes related to OpenUpgrade framework**

- lot of changes in ``_check_xml``. Specially https://github.com/odoo/odoo/commit/46f1912c9b9ee9767d6f32b725c691509dc5e5b2. Nothing to do in my opinion 

- lot of changes in ``_post_process_view``. Here https://github.com/odoo/odoo/commit/331b54770e21a8dd4617d05f0bee2cc0383ba63b. Nothing to do in my opinion 

- ``handle_view_error`` has been splitted in two functions here https://github.com/odoo/odoo/commit/9ae8c1a7c984183d6d1bd131ffcd9d4269c19a1f. **TODO**

PR #3242, opened 2022-05-11
[15.0][MIG] mail: delete duplicates which trigger issue due to the new constaint mail_c…

jcdrubay:
…hannel_partner_partner_unique

Fixes: https://github.com/jcdrubay/OpenUpgrade/issues/2


openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,33 @@
+
```
MiquelRForgeFlow:
```suggestion
```

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,33 @@
...
+from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate(use_env=False)
+def migrate(cr, version):
+    # Cleanup potential duplicates
+    openupgrade.logged_query(
+        cr,
```
MiquelRForgeFlow:
```suggestion
@openupgrade.migrate()
def migrate(env, version):
    # Cleanup potential duplicates
    openupgrade.logged_query(
        env.cr,
```
This is @pedrobaeza preference and he has long convinced me to do this way as default.

pedrobaeza: You never know when you are going to need the environment, so better to have it already populated for not having a great diff later.

jcdrubay:
@legalsylvain I think that you can remove the label Blockedbydependency

legalsylvain:
Base is not done akaik ?

jcdrubay:
My bad, you are right. I have been dreaming someone had done it. :|

jcdrubay:
This commit was my dream becoming reality.
https://github.com/OCA/OpenUpgrade/pull/2967/commits/130d038040ce522208e14fb921f35f8b4a5aef15

Do I misunderstand something?

StefanRijnhart:
This PR looks a bit messy. Can you try to  rebase?

jcdrubay:
I think it is better to handle here: https://github.com/OCA/OpenUpgrade/pull/3273/files#r899917133

PR #3249, opened 2022-05-13
[13.0][OU] base: fix translation term states

chienandalu:
There is an unresolved bug in versions <=12 (probably coming from web_editor)
that causes wrong translation states. This causes crashes when those terms
get edited in v13. We also want to normalize states to the expected ones when
they are null.

cc @Tecnativa TT36678

please review @pedrobaeza


pedrobaeza (approved):

MiquelRForgeFlow (approved):
LGTM

PR #3257, opened 2022-05-26
[13.0][OU-FIX] pos_sale: map old pos.order to crm_team_id

chienandalu:
From now on, every pos.order will have stored its crm.team relation, so if it's
changed later, those sales don't go to the new team. Anyway we have to map it in
the old orders so we don't loose that info in the sales report.

cc @Tecnativa TT36891

please review @MiquelRForgeFlow @pedrobaeza @sergio-teruel

### DIFF
--- a/addons/pos_sale/migrations/13.0.1.0/openupgrade_analysis_work.txt
+++ b/addons/pos_sale/migrations/13.0.1.0/openupgrade_analysis_work.txt
@@ -7,7 +7,7 @@ pos_sale     / crm.team                 / team_type (False)             : DEL se
 # NOTHING TO DO: removed field, every team should be able to handle a POS or a website for example
 
 pos_sale     / pos.order                / crm_team_id (many2one)        : NEW relation: crm.team
-# NOTHING TO DO: new feature
+# DONE: post-migration: Map to config team relation. Before that relation was the one used to group in the sales report.
 
 pos_sale     / pos.order                / currency_rate (float)         : module is now 'point_of_sale' ('pos_sale')
 # NOTHING TO DO: Will be handled by ORM on module update
--- a/addons/pos_sale/migrations/13.0.1.0/post-migration.py
+++ b/addons/pos_sale/migrations/13.0.1.0/post-migration.py
@@ -3,6 +3,25 @@
 from openupgradelib import openupgrade
 
 
+def _map_crm_team_id(env):
+    """From now on, every pos.order will have stored its crm.team relation, so if it's
+    changed later, those sales don't go to the new team. Anyway we have to map it in
+    the old orders so we don't loose that info in the sales report."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE pos_order po
+        SET crm_team_id = pc.crm_team_id
+        FROM pos_config pc, pos_session ps
+        WHERE
+            ps.id = po.session_id
+            AND pc.id = ps.config_id
+            AND po.crm_team_id IS NULL
+            AND pc.crm_team_id IS NOT NULL
+        """,
+    )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(
@@ -12,3 +31,4 @@ def migrate(env, version):
             "point_of_sale.pos_config_main",
         ]
     )
+    _map_crm_team_id(env)


legalsylvain (approved):
Thanks for this patch.

question (non-blocking) : I fear that the write by ORM can be slow in huge database. did you faced issue ? did you envisage to write an SQL request ? 

rergards.

chienandalu:
It was a about 30 seconds on ~150.000 pos.order records from 6 configs. The write is done in bulk for all the records of a config, so surely the main issue could be on the filtered.

addons/pos_sale/migrations/13.0.1.0/openupgrade_analysis_work.txt
```diff
@@ -7,7 +7,7 @@ pos_sale     / crm.team                 / team_type (False)             : DEL se
 # NOTHING TO DO: removed field, every team should be able to handle a POS or a website for example
 
 pos_sale     / pos.order                / crm_team_id (many2one)        : NEW relation: crm.team
-# NOTHING TO DO: new feature
+# DONE: Map to config team relation. Before that relation was the one used to group in the sales report.
```
pedrobaeza:
```suggestion
# DONE: post-migration: Map to config team relation. Before that relation was the one used to group in the sales report.
```

addons/pos_sale/migrations/13.0.1.0/post-migration.py
```diff
@@ -3,6 +3,17 @@
...
 
+def _map_crm_team_id(env):
+    """From now on, every pos.order will have stored its crm.team relation, so if it's
+    changed later, those sales don't go to the new team. Anyway we have to map it in
+    the old orders so we don't loose that info in the sales report."""
+    po_sale = env["pos.order"].search([("crm_team_id", "=", False)])
+    for config in po_sale.config_id.filtered("crm_team_id"):
+        po_sale.filtered(
+            lambda x: x.config_id == config
+        ).crm_team_id = config.crm_team_id
```
pedrobaeza: Always use SQL when possible, as you don't know if a custom is overwriting `write` method for doing a heavy check and in a certain DB times go upper exponentially:
```suggestion
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE pos_order po
        SET crm_team_id = pc.crm_team_id
        FROM pos_config pc
        WHERE pc.id = po.config_id
            AND po.crm_team_id IS NULL
            AND pc.crm_tema_id IS NOT NULL
        """,
    )
```

pedrobaeza (changes requested):

addons/pos_sale/migrations/13.0.1.0/post-migration.py
chienandalu: Done :+1:

addons/pos_sale/migrations/13.0.1.0/openupgrade_analysis_work.txt
chienandalu: Done

chienandalu:
Now with the sql update it takes about 3 seconds for those ~150000 records :)

addons/pos_sale/migrations/13.0.1.0/post-migration.py
```diff
@@ -3,6 +3,25 @@
...
+    the old orders so we don't loose that info in the sales report."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE pos_order po
+        SET crm_team_id = pc.crm_team_id
+        FROM pos_config pc, pos_session ps
+        WHERE
+            ps.id = po.session_id
+            AND pc.id = ps.config_id
```
MiquelRForgeFlow, line 18:
```suggestion
        FROM pos_session ps
        JOIN pos_config pc ON pc.id = ps.config_id
        WHERE
            ps.id = po.session_id
```
better look.

MiquelRForgeFlow (approved):

addons/pos_sale/migrations/13.0.1.0/post-migration.py
chienandalu: Isn't the list of tables in the `FROM` do that very same implicitly? I mean, I tested it with real data and it's working :slightly_smiling_face: 

```
[ WITH [ RECURSIVE ] with_query [, ...] ]
UPDATE [ ONLY ] table_name [ * ] [ [ AS ] alias ]
    SET { column_name = { expression | DEFAULT } |
          ( column_name [, ...] ) = [ ROW ] ( { expression | DEFAULT } [, ...] ) |
          ( column_name [, ...] ) = ( sub-SELECT )
        } [, ...]
    [ FROM from_item [, ...] ]
    [ WHERE condition | WHERE CURRENT OF cursor_name ]
    [ RETURNING * | output_expression [ [ AS ] output_name ] [, ...] ]
```

MiquelRForgeFlow: yeah, yeah, your query is not wrong. just that IMO, with the JOIN looks better. aesthetics!

pedrobaeza: Let's keep it this time as is for not relaunching CI.

pedrobaeza (approved):

PR #3258, opened 2022-05-26
[13.0][OU-FIX] mass_mailing: prevent deleting mailing list

chienandalu:
In <13 Odoo create a mailing list by default with the xml_id
mass_mailing.mass_mail_list_1, but now in 13.0 it doesn't exist
anymore so if we don't get rid of the former ir.model.data we
could be missing our mailing list and its records.

cc @Tecnativa TT37037


pedrobaeza:
@chienandalu please remember that the commit message should be `[OU-FIX] ...` or `[OU-IMP] ...`, and the PR title `[13.0][OU-FIX] ...`

pedrobaeza (approved):

pedrobaeza:
No reflect in the analysis work file?

legalsylvain (approved):

sergio-teruel (approved):

PR #3259, opened 2022-05-27
[9.0][FIX] account: Create new sequence instead of renaming

pedrobaeza:
When converting old fiscal years sequences to new range style, there can be the case where the same sequence is used in several intervals, so renaming will lead to an error on the second renaming.

Thus, we create a new sequence instead of renaming.

@Tecnativa


MiquelRForgeFlow:
cannot you do the following?

first, check whether if exists new_seq_name. if exists, then create new one. if not exists, do the rename.

pedrobaeza:
That can't be, as for the second occurrence, we will lose the firstly renamed and we won't know which is the following number.

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
I didn't go deep with this, but I believe in you.

pedrobaeza:
Tested this weekend in a migration.

PR #3260, opened 2022-05-29
[13.0][FIX] account: drop obsolete register payment m2m transient table

zurawmaco:
invoice_ids field was moved to another transient model. Old table still exists in database


pedrobaeza:
OpenUpgrade doesn't drop any table. You don't know if it's going to be needed by any other module during the migration process.

Use database_cleanup when you want to clean data.

Closing the PR then.

PR #3264, opened 2022-06-14
Enable a custom configuration which can be committed to enable renami…

jcdrubay:
…ng within custom code

### DIFF
--- a/docsource/development.rst
+++ b/docsource/development.rst
@@ -48,6 +48,46 @@ the full power of OpenUpgrade was unleashed to successfully overcome the
 https://github.com/OCA/OpenUpgrade/pull/2275/files.
 
 
+Handling of the renaming and merging of custom modules with a Custom Configuration File
+---------------------------------------------------------------------------------------
+
+To handle custom code, you can create a custom configuration file
+(ex: ``openupgrade-15.conf``) which can be part of your custom repo.
+
+This file can be used to store the renaming and merge of custom modules.
+The expected content structure of the file is the following:
+
+.. code-block:: txt
+
+  [renamed_modules]
+  old_module_name_1=new_module_name_1
+  old_module_name_2=new_module_name_2
+  [merged_modules]
+  merged_module_1=new_module_merged_1
+
+Then, you must declare this custom configuration file into your Odoo
+configuration file (``odoo.cfg`` or ``odoo.conf``) like this:
+
+.. code-block:: txt
+
+    [openupgrade]
+    config_path=../config/openupgrade-15.conf
+
+The purpose of using a configurable path is to allow you adding the openupgrade
+custom configuration file into the repo where your custom modules are so you can commit
+it.
+
+When running the upgrade, the configurations ``renamed_modules`` and ``merged_modules``
+will be taken into account as one of the first step of the migration as part of the
+pre-migration script of the module ``base``:
+
+.. code-block:: python
+
+    openupgrade.update_module_names(cr, renamed_modules.items())
+    openupgrade.update_module_names(cr, merged_modules.items(), merge_modules=True)
+
+
+
 Learn from existing migration scrips
 ------------------------------------
 
@@ -113,6 +153,20 @@ Note that for the time being, this is not available in all versions
 It is available only for 12 and 13:
 https://github.com/OCA/OpenUpgrade/pulls?q=is%3Apr+%5BFIX%5D+reset+exception
 
+
+Re-upgrade an upgraded module
+.............................
+
+In some cases, you might want to try forcing the upgrade of a module which was already
+upgraded by resetting the version of the module to a past version.
+
+You might want to do that if you fix the migration script of a module after it was
+migrated.
+
+Example to update the known version of ``base`` to 14.0 so OpenUpgrade will try
+to re-upgrade it to version 15:
+
+``update ir_module_module set latest_version='14.0' where name='base';``
 
 Learning resources
 ------------------
--- a/docsource/development.rst
+++ b/docsource/development.rst
@@ -57,7 +57,7 @@ To handle custom code, you can create a custom configuration file
 This file can be used to store the renaming and merge of custom modules.
 The expected content structure of the file is the following:
 
-.. code-block:: txt
+.. code-block:: text
 
   [renamed_modules]
   old_module_name_1=new_module_name_1
@@ -68,7 +68,7 @@ The expected content structure of the file is the following:
 Then, you must declare this custom configuration file into your Odoo
 configuration file (``odoo.cfg`` or ``odoo.conf``) like this:
 
-.. code-block:: txt
+.. code-block:: text
 
     [openupgrade]
     config_path=../config/openupgrade-15.conf
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -1,9 +1,45 @@
-""" Encode any known changes to the database here
-to help the matching process
-"""
+import logging
+
+import configparser as ConfigParser
+
+import odoo
+
+def load_custom_config():
+    openupgrade_config_path = odoo.tools.config.misc.get("openupgrade").get("config_path")
+    res = {}
+    if openupgrade_config_path:
+        logging.info("load_custom_config from %s: ", openupgrade_config_path)
+        p = ConfigParser.RawConfigParser()
+        try:
+            p.read([openupgrade_config_path])
+            for sec in p.sections():
+                res.setdefault(sec, {})
+                for (name, value) in p.items(sec):
+                    if value=='True' or value=='true':
+                        value = True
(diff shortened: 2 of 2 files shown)


openupgrade_scripts/apriori.py
```diff
@@ -1,9 +1,45 @@
...
-to help the matching process
-"""
+import logging
+
+import configparser as ConfigParser
+
+import odoo
+
+def load_custom_config():
+    openupgrade_config_path = odoo.tools.config.misc.get("openupgrade").get("config_path")
```
legalsylvain: Hi. @jcdrubay. Thanks for your contribution.

I don't think that having 2 configuration files is a good thing.
- it makes the code to maintain much heavier
- it is not in the current design of Odoo (or OCA) to have several configuration files.

If in your custom configuration, you have multiple configuration files, you can write a custom python script
- use ``read()`` function to read multiple files
- use ``write()`` to write a single complete complete file.

https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read

legalsylvain (changes requested):

openupgrade_scripts/apriori.py
jcdrubay: I think that there is a misunderstanding. For the custom configuration, there would only be one file.

In total, there would be only 2 configuration files

- odoo.cfg: which is usually not committed and which is the standard configuration file of Odoo with (optionally) a single path `openupgrade_config_path`.
- openupgrade.cfg: only one file. This file should be committed together with the custom code being upgraded with several (optional) sections `renamed_modules` and `merged_modules`

jcdrubay: Note that the `openupgrade_config_path` is optional and that this file would only be used to declare necessary renaming/merging for custom code.
Inside the `odoo.cfg` file there would be no other addition than the configuration openupgrade_config_path. The listing of modules to rename and merge from non-custom code is being kept in the same place in the `apriori.py` file.

jcdrubay:
I don't think I should fix the last unsuccessful check within this PR.

@legalsylvain @pedrobaeza @StefanRijnhart Do you think that we are good to merge?

docsource/development.rst
```diff
@@ -114,6 +154,20 @@ It is available only for 12 and 13:
 https://github.com/OCA/OpenUpgrade/pulls?q=is%3Apr+%5BFIX%5D+reset+exception
 
 
+Re-upgrade an upgraded module
+.............................
+
+In some cases, you might want to try forcing the upgrade of a module which was already
```
jcdrubay, line 160: This paragraph is  not related to this PR. Let me know if you want me to create a dedicated PR.

StefanRijnhart:
This is a rather advanced use case I think. I can imagine that someone that requires this functionality might just as well apply their own changes to https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/apriori.py. That makes me think that the target audience for this is rather small, and we should not go through with this.

docsource/development.rst
StefanRijnhart: Here as well, if you want to try to reset your module versions you have to know what you are doing as it will not work unless with the simpelest of migration scripts. So telling people just to go ahead and do this is misleading.

jcdrubay: Have you ever been using this trick @StefanRijnhart ? 

This page is for the people who want to do development on OpenUpgrade so somehow they should know what they are doing.

Also, running migration scripts takes quite a lot of time, and I think that this trick can help during development to test faster to be able to move forward. But well, at this stage of learning about OpenUpgrade, what I think is closer to imagination rather than well thought and tested solutions.

While you have way way more experience than me in developing for OpenUpgrade... I probably should just listen to you.

jcdrubay:
You must be right on the very small audience.
I will close this PR.
It's been interesting to do, and I learned on the way.

jcdrubay:
@StefanRijnhart @legalsylvain Thank you for taking the time to review. :)

docsource/development.rst
StefanRijnhart: I have only ever used this trick when I wanted to iteratively test whether minor migration scripts in custom projects do not raise syntax errors. Such simple scripts are usually reentrant, unlike the complex scripts in OpenUpgrade with their pre/post dependencies and renamed fields.
If this is to ease the development of openupgrade scripts, the trick I use is to raise an error explicitely at the end of the post migration script which rolls back the transaction so that you can safely start again after inspecting the results (in a debugger right before the raising of the exception)

pedrobaeza:
FTR, when we need to have custom merged/renamed modules, we simply fork this repo, and create a branch with such extra commit, and use it as source instead the main one (well, more specifically, with Doodba we just add the merge operation that is processed through git-aggregator and we don't miss any update on the main repo).

jcdrubay:
It seems that git-aggregator is a must have when working with OpenUpgrade while the scripts of a given version are under development... 

I will try to discover more about it and try to submit a documentation page on how to integrate it in the development process. Hopefully, that contribution will help ;)

legalsylvain:
Hi. 

a little disappointed to see this PR closed as I required this feature in all the migration I realized. (for grap and custom customer).

- When you see that this need seems to be shared by the Tecnativa team as well as JCdrubay, GRAP, and other companies, maybe it's not such a small audience as Stefan presents it...
- On the other hand: not integrating a feature but proposing to keep a custom fork of a file doesn't seem to me the most user-friendly.

Not a big deal, though.

pedrobaeza:
We have required this only in 2 occasions, and it's because we were converging bounced customers with ugly custom/bought modules. The rest of the time, as everything we do is OCA, it's already included here. Anyways, it's a matter of which flow you have on your pipeline, so not sure if this can be presented in a generic documentation.

jcdrubay:
I am going to explore the following flow:

- try to keep OpenUpgrade as much as possible as a library
- create a module in a separate repo `custom_openupgrade_scripts` which depends on `openupgrade_scripts`
- use that new module to override the value of the variables `renamed_modules` and `merged_modules` with a custom apriory.py file which would override the standard one
- this would also allow me having a separate repository for the migration scripts of my custom modules.
- since I have modules shared across several projects, a custom repo would help me sharing the migration scripts across projects, and possibly repeat the same method to handle project specific code

In the end, I could end up having 3 modules to migrate a project:

- openupgrade_scripts: the one from oca/openupgrade: possibly using git-aggregator while OpenUpgrade is not yet stable for a specific version
- komit_openupgrade_scripts: for the migration scripts of all our generic modules shared among our projects
- project_xxx_openupgrade_scripts: for the migration scripts of project specific modules

jcdrubay:
It might not be easy to "play" with the upgrade_path :/

jcdrubay:
Would it be ok to add this to the apriori.py file of openupgrade_scripts?

This would allow adding hooks to pre-load external apriori.py files and then override the default one

```
# Loading apriori.py files from upgrade scripts from other directories of upgrade_path
import os
import sys

for upgrade_path in odoo.tools.config.get("upgrade_path").split(","):
    module_path = os.path.dirname(upgrade_path)
    current_file_path = os.path.abspath(__file__)
    print(module_path)
    if module_path in os.path.abspath(__file__):
        print("skip : %s, in %s" % (module_path, module_path))
    else:
        if os.path.exists(os.path.join(module_path, "apriori.py")):
            sys.path.insert(1, module_path)
            import apriori
```

Then in my apriori.py file I can do:


```
from odoo.addons.openupgrade_scripts.apriori import merged_modules, renamed_modules
renamed_modules.update({"my_old_module_name": "my_new_module_name")
```

and any other thing I could dream of in the future...

pedrobaeza:
I think the best for your plan is to simply manipulate the variable in your special addons:

```python
from odoo.addons.openupgrade_scripts import apriori

apriori.renamed_modules.append(...)
```

and add it to the server_wide_modules

jcdrubay:
I tried that but it seems that base/pre-migration.py is running before the aprioriy.py of my special module is loaded/executed (even if I set my special module in the server_wide_modules).

That's why I force the loading of my apriori.py file in the apriori.py of native.

Unless I did something wrong. :/ But I just tried again with a simple `raise` in my apriori.py file no exception is raised.

pedrobaeza:
OK, it was just a theoretical patch.

pedrobaeza:
Closing as there's no consensus.

jcdrubay:
Thanks for the triage @pedrobaeza . 
I am a bit disappointed, but I am much more happy to feel that this community weight everyones suggestions and doesn't apply changes without consensus.

pedrobaeza:
We can talk about your use cases in the OCA days.

jcdrubay:
Thanks, unfortunately I won't be there. But I can join online if possible.

pedrobaeza:
Not sure it will be possible with the craziness there. I will be also on Odoo XP.

PR #3273, opened 2022-06-14
[MIG][15.0] mail: migration script

sang250399:
This PR
======
Migration done


openupgrade_scripts/scripts/mail/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,266 @@
...
+
+mail         / mail.channel             / email_send (boolean)          : DEL
+mail         / mail.channel             / moderation (boolean)          : DEL
+mail         / mail.channel             / moderation_guidelines (boolean): DEL
+mail         / mail.channel             / moderation_guidelines_msg (text): DEL
+mail         / mail.channel             / moderation_ids (one2many)     : DEL relation: mail.moderation
+mail         / mail.channel             / moderation_notify (boolean)   : DEL
+mail         / mail.channel             / moderation_notify_msg (text)  : DEL
+mail         / mail.channel             / moderator_ids (many2many)     : DEL relation: res.users
+# NOTHING TO DO: new feature
```
StefanRijnhart, line 61: Is it the case that channel functionality was split off to the new module mail_group? Do we need to prepare for it by renaming the models and fields in this migration, and select 'mail_group' for installation if there is actual data in these tables (or apply some other strategy to ensure the module is installed)?

royle-vietnam: @StefanRijnhart yes, channel functionality was split off to the new module mail_group, but we don't have to do.
Simply keep the old mail.channel, and the old mail.channel no longer has the ability to send mail
`mail_group` is used for mass mailing, and mail.group will use for new data

StefanRijnhart: Ah I see it's a little bit more complex than I thought. Thanks!

openupgrade_scripts/scripts/mail/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,266 @@
...
+mail         / mail.message.reaction    / content (char)                : NEW required
+mail         / mail.message.reaction    / guest_id (many2one)           : NEW relation: mail.guest
+mail         / mail.message.reaction    / message_id (many2one)         : NEW relation: mail.message, required
+mail         / mail.message.reaction    / partner_id (many2one)         : NEW relation: res.partner
+# NOTHING TO DO: new feature
+
+mail         / mail.moderation          / channel_id (many2one)         : DEL relation: mail.channel, required
+mail         / mail.moderation          / email (char)                  : DEL required
+mail         / mail.moderation          / status (selection)            : DEL required, selection_keys: ['allow', 'ban']
+# DONE: transfer data to model mail_group_moderation
```
StefanRijnhart: Did you actually do anything for this, or is this to be done by the person to undertake the migration of mail_group?

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -0,0 +1,55 @@
...
+
+def _map_mail_notification_failure_type(env):
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("failure_type"),
+        "failure_type",
+        [
+            ("SMTP", "mail_smtp"),
+            ("RECIPIENT", "mail_email_invalid"),
+            ("UNKNOWN", "unknown"),
```
StefanRijnhart, line 36: I think you should also handle BOUNCE, otherwise there can be invalid entries left after the migration. I would suggest `mail_email_invalid` or `unknown`.

openupgrade_scripts/scripts/mail/15.0.1.5/upgrade_analysis_work.txt
sang250399: As discussed above, we don't need migration for this change. I forgot to edit the description, this should be nothing to do.

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
sang250399: thanks, i will add it

openupgrade_scripts/scripts/mail/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,266 @@
...
+mail         / mail.ice.server          / uri (char)                    : NEW required
+mail         / mail.ice.server          / username (char)               : NEW
+# NOTHING TO DO: new feature call audio and video
+
+mail         / mail.mail                / failure_type (selection)      : NEW selection_keys: ['mail_bl', 'mail_dup', 'mail_email_invalid', 'mail_email_missing', 'mail_optout', 'mail_smtp', 'unknown']
+# NOTHING DO DO
+
+mail         / mail.mail                / is_notification (boolean)     : NEW
+mail         / mail.mail                / notification (boolean)        : DEL
+# DONE: pre-migration: rename fields notification -> is_notification
```
StefanRijnhart, line 116: I don't see this one in the pre-migration. Was it left out by mistake?

sang250399: ohh, it's my mistake

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,66 @@
...
+                "suggested_next_type_ids",
+            ),
+            (
+                "mail.activity.type",
+                "mail_activity_type",
+                "default_description",
+                "default_note",
+            ),
+            ("mail.message", "mail_message", "no_auto_thread", "reply_to_force_new"),
+            ("mail.notification", "mail_notification", "mail_id", "mail_mail_id"),
```
StefanRijnhart, line 40: Slightly puzzled why this seems to work, if it does. mail_notification is the new table name, but this method is called before `_rename_tables` is.

StefanRijnhart: Thanks for the fix! Maybe it's just me, but I prefer to leave 'resolving' comments to the submitter so that I (or other reviewers) can follow up on my own reviews more easily.

StefanRijnhart:
Thanks for the quick fixes! This is really great work especially for you as a new contributor.

I see that due to a small error in our scripting the `mail` module is not upgraded in this PRs github action. I'm hoping to have it fixed in https://github.com/OCA/OpenUpgrade/pull/3278. Would you like to try and merge that branch in here?

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
sang250399: > 

yes, I think it's reasonable !

StefanRijnhart:
Thanks for the quick fixes! This is really great work especially for you as a new contributor.

I see that due to a small error in our scripting the `mail` module is not upgraded in this PRs github action. I'm hoping to have it fixed in https://github.com/OCA/OpenUpgrade/pull/3278. Would you like to try and merge that branch in here?

sang250399:
> Thanks for the quick fixes! This is really great work especially for you as a new contributor.
> 
> I see that due to a small error in our scripting the `mail` module is not upgraded in this PRs github action. I'm hoping to have it fixed in #3278. Would you like to try and merge that branch in here?

okay, i will try

StefanRijnhart (approved):
Thanks, scripts are now run!

```
Testing modules base,base,mail
...
2022-06-14 09:47:58,233 5347 INFO openupgrade odoo.modules.migration: module mail: Running migration [>15.0.1.5] pre-migration 
...
2022-06-14 09:48:00,643 5347 INFO openupgrade odoo.modules.migration: module mail: Running migration [15.0.1.5>] post-migration
```

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,67 @@
...
+
+def _delete_channel_follower_records(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DELETE FROM mail_followers
+        WHERE partner_id IS NULL;
+        """,
+    )
+
```
jcdrubay, line 60: Could you please add this ?
```
    # Cleanup potential duplicates to comply with the new constraint mail_channel_partner_partner_unique
    openupgrade.logged_query(
        env.cr,
        """
        DELETE FROM mail_channel_partner mcp
        WHERE
            EXISTS (
                SELECT 1 FROM mail_channel_partner mcp2
                WHERE
                    mcp2.partner_id=mcp.partner_id
                    AND mcp2.channel_id=mcp.channel_id
                HAVING count(*) > 1
            )
            AND NOT EXISTS (
                SELECT 1 FROM ir_model_data imd
                WHERE
                    imd.model ='mail.channel.partner' and res_id = mcp.id
            )
        """,
    )
    
```

StefanRijnhart: Assuming that the ones with an ir_model_data entry are created before the other ones, it could be considered to remove the ones with a higher id (using `where exists ... where mcp.id > mcp2.id`). Also, maybe not take `__export__` ir_model_data entries into account.

docsource/modules140-150.rst
```diff
@@ -412,7 +412,7 @@ Module coverage 14.0 -> 15.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | lunch                                           |                      |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| mail                                            |                      |                                                 |
+| mail                                            |Done                  |                                                 |
```
pedrobaeza: Incorrect syntax

```suggestion
| mail                                            | Done                 |                                                 |
```

pedrobaeza (changes requested):
My comment is pending, and also on the test logs, there is the following line:
```
2022-06-15 06:18:28,759 5443 INFO openupgrade odoo.modules.graph: module mail: Unmet dependencies: base_setup, bus, web_tour 
```
so this is not correct and can't be merged. Please check it.

royle-vietnam:
> My comment is pending, and also on the test logs, there is the following line:

Done

> 2022-06-15 06:18:28,759 5443 INFO openupgrade odoo.modules.graph: module mail: Unmet dependencies: base_setup, bus, web_tour 

Maybe the docsource file is not standard, `|Nothing to do` -> `| Nothing to do`, Can you do mass editing PR? @pedrobaeza

pedrobaeza:
Yes, that syntax is not correct and it's not correctly parsed. Please fix it in the rest of your PRs. I'll fix the current main one.

pedrobaeza:
Are you the one that have put `No DB layout changes` in the docs? That's not a valid text. It should be `Nothing to do`.

royle-vietnam:
> Yes, that syntax is not correct and it's not correctly parsed. Please fix it in the rest of your PRs. I'll fix the current main one.

Okey

royle-vietnam:
> Are you the one that have put `No DB layout changes` in the docs? That's not a valid text. It should be `Nothing to do`.

No, not me, looks like @legalsylvain

legalsylvain:
> Are you the one that have put No DB layout changes in the docs? That's not a valid text. It should be Nothing to do.

Hum.... @pedrobaeza :
https://github.com/OCA/server-tools/pull/2206#issuecomment-968757796

pedrobaeza:
I didn't remember that this text is parsed by the tests: https://github.com/OCA/OpenUpgrade/blob/edb3b9cc0d1203ea737fd9defeea493dc414b436/.github/workflows/test.yml#L103

so this has to be changed.

pedrobaeza:
Sorry for the noise. @legalsylvain it was correct, as such column is an extra one. Digging more about why `base_setup` is not being considered.

pedrobaeza (approved):
This is weird, as now I have re-run the tests and they are working, so let's continue. I'm worried about these situations being masked on the CI and only searching for the green status. I will continue checking the logs of other PRs to see if this happens again.

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
royle-vietnam: @jcdrubay @StefanRijnhart I have a fix at PR https://github.com/OCA/OpenUpgrade/pull/3400

PR #3278, opened 2022-06-14
[15.0][FIX] ci: adapt regular expression that select modules to new column in table

StefanRijnhart:
Currently, only 'base' is upgraded even when other modules are set to Done (see https://github.com/OCA/OpenUpgrade/pull/3273)

-- edit -- Now merged in #3273, so maybe that one can be merged sooner rather than later, and rather than this one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,7 +102,7 @@ jobs:
               openupgrade/docsource/modules140-150.rst \
               | grep "Done\|Partial\|Nothing" \
               | grep -v "theme_" \
-              | sed -rn 's/((^\| *\|del\| *)|^\|)([0-9a-z_]*) *\|.*$/\3/g p' \
+              | sed -rn 's/((^\| *\|del\| *)|^\| *)([0-9a-z_]*)[ \|].*/\3/g p' \
               | sed '/^\s*$/d' \
               | paste -d, -s)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you please explain a bit the changes in the regexp and why this has changed over other versions? You know they are very hard to be read

legalsylvain (commented):
I don't understand why we don't have the problem in V14.

StefanRijnhart:
Oh, sorry for not being clear on this. The expression for 14 fails on the table in 15 because it has an extra column:
```
Module coverage 13.0 -> 14.0
============================

.. include:: coverage_legend.rst

|Module                                      |Status                                           |
+============================================+=================================================+
|account                                     | Done                                            |
| |del| account_analytic_default             | Merged into account                             |
| |del| account_analytic_default_hr_expense  | Merged into hr_expense                          |
```

```
Module coverage 14.0 -> 15.0
============================

.. include:: coverage_legend.rst

| Module                                          | Status               + Extra Information                               |
+=================================================+======================+=================================================+
| account                                         |                      |                                                 |
| account_check_printing                          |                      |No DB layout changes.                            |
```

pedrobaeza (approved):
Thanks for the information. Not sure this extra column is a good idea, but anyways, the CI should work

PR #3283, opened 2022-06-15
[MIG][15.0] product: migration script

sang250399:
This PR
======
Migration done


openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,55 @@
...
+
+def convert_text_to_html(env):
+    openupgrade.convert_field_to_html(
+        env.cr, "product_template", "description", "description"
+    )
+
+
+def compute_detailed_type_and_type(env):
+    if not openupgrade.column_exists(env.cr, "product_template", "detailed_type"):
+        openupgrade.logged_query(
```
gdgellatly, line 34: Is there any reason we cannot use rename column?

StefanRijnhart: Oh, the old `type` column is still there as well. The new field `detailed_type` can branch out to things like 'event' which are then automatically mapped to value `service` for `type`.

StefanRijnhart: @sang250399 I do wonder what the use case is of a defensive strategy like this one. If the field already exists, what are the odds that the values are compatible with what is introduced in Odoo 15? Would it not be safer to create an unconditional copy and let OpenUpgrade halt on the SQL error if the column already exists?

openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,55 @@
...
+            ALTER TABLE product_template
+            ADD COLUMN detailed_type varchar
+            """,
+        )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE product_template
+        SET detailed_type = type
+        WHERE type = 'consu' OR type = 'service'
```
StefanRijnhart, line 46: Looks to me like you can copy all values unconditionally, because `product` is going to be added in the `stock` module, and its value is the same in both tables as well. Something might need to be done specifically for events, but it can be handled in `event_sale`.

openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
tranngocson1996: @StefanRijnhart I don't believe we can be certain whether the column is already present. As a result, this sql check simply prevents an error from occurring because the column already exists. If it already exists and contains data, replace it with a new value that corresponds to code 15. (the old data of that column may not be correct)
cc: @sang250399

openupgrade_scripts/scripts/product/15.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,65 @@
...
+# NOTHING TO DO
+
+product      / product.template         / priority (selection)          : NEW selection_keys: ['0', '1'], hasdefault
+# NOTHING TO DO: new fields
+
+product      / product.template.attribute.line / value_count (integer)         : NEW isfunction: function, stored
+# DONE: pre-migration: compute_value_count
+
+product      / product.template.attribute.value / color (integer)               : NEW hasdefault
+# NOTHING TO DO: new field
```
StefanRijnhart: Mmm, the default is random: https://github.com/odoo/odoo/blob/15.0/addons/product/models/product_attribute.py#L104-L105. Upon initialization of the field by the ORM, a value is going to be generated once and applied to all attribute lines.  That means that if you're going to do several test migrations, the color is going to be wildly different between two migrations. I feel like that's bound to bug at least someone somewhere. Would you consider setting a neutral value of `0` (grey) explicitely instead?

openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
StefanRijnhart: Yes alright. It's a corner case anyway and it's no different from Odoo introducing new columns that may have already been claimed by third parties in earlier editions.

openupgrade_scripts/scripts/product/15.0.1.2/upgrade_analysis_work.txt
sang250399: this is a field with random color value, but we can give it a default color, okay i will add this

openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
sang250399: We had conditional WHERE type = 'consu' OR type = 'service'
In the scope of this module, there are only 2 types: consu and service
The stock module adds a new type, when we write migration for the stock module, we will copy its types.

StefanRijnhart: Alright, thanks for clarifying!

openupgrade_scripts/scripts/product/15.0.1.2/upgrade_analysis_work.txt
StefanRijnhart: If you feel adventurous, you can assign a *different* random color to each attribute value (I'm sure there is a quick way to do so in SQL). I just want to prevent that *all* attribute values are assigned the same color (which is going to be a different color in each test migration, which may/should raise questions from QA testers)

sang250399: > If you feel adventurous, you can assign a _different_ random color to each attribute value (I'm sure there is a quick way to do so in SQL). I just want to prevent that _all_ attribute values are assigned the same color (which is going to be a different color in each test migration, which may/should raise questions from QA testers)

Okay, i think this is reasonable, i will randomly add color to the records

daonamutc1:
@StefanRijnhart May you review this PR again?

openupgrade_scripts/scripts/product/15.0.1.2/upgrade_analysis_work.txt
StefanRijnhart: Cool!

StefanRijnhart (approved):

MiquelRForgeFlow:
Fine-tuned in https://github.com/OCA/OpenUpgrade/pull/3492.

Issue #3285, opened 2022-06-15
Override of migrate_module not working as expected


jcdrubay:
## Module
openupgrade_framework

## Describe the bug

This function is not honoring it's docstring.

```
def migrate_module(self, pkg, stage):
    """In openupgrade, also run migration scripts upon installation.
    We want to always pass in pre and post migration files and use a new
    argument in the migrate decorator (explained in the docstring)
    to decide if we want to do something if a new module is installed
    during the migration.
    We trick Odoo into running the scripts by setting the update attribute if necessary.
    """
    has_update = hasattr(pkg, "update")
    if not has_update:
        pkg.update = True
    MigrationManager.migrate_module._original_method(self, pkg, stage)
    if not has_update:
        delattr(pkg, "update")
```

## To Reproduce

Version 15.0 (probably earlier version as 13.0 and 14.0 of Odoo has the same if statement)

1. Run upgrade for a module which was renamed, the pre-migration is not executed.

**Expected behavior**
Run the pre-migration script.

**Additional context**

In https://github.com/odoo/odoo/blob/15.0/odoo/modules/migration.py#L101
The if statement of original function `migrate_module` is positive because the state of the module is "to install".

```python
        if not (hasattr(pkg, 'update') or state == 'to upgrade') or state == 'to install':
            return
```

## Suggested new code:

```diff
--- a/openupgrade_framework/odoo_patch/odoo/modules/migration.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/migration.py
@@ -14,9 +14,15 @@ def migrate_module(self, pkg, stage):
     has_update = hasattr(pkg, "update")
     if not has_update:
         pkg.update = True
+
+    to_install = pkg.state == "to install"
+    pkg.state = "to ugprade"
     MigrationManager.migrate_module._original_method(self, pkg, stage)
     if not has_update:
         delattr(pkg, "update")
+    if to_install:
+        pkg.state = "to install"
+
```

pedrobaeza:
The question is that the rename is done after this code. What I see that we have lost is the recursive upgrade cycle that install/upgrade remaining modules that comes for this kind of things to do. I rose this problem in #3061, and it seems solved for the specific case I put, but it seems it's not enough for others where the dependency chain has several levels.

cc @StefanRijnhart

StefanRijnhart:
[...]

PR #3286, opened 2022-06-15
[15.0][OU-FIX] Migration scripts were not run during install

StefanRijnhart:
Port of https://github.com/OCA/OpenUpgrade/pull/3155

Fixes: #3285

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/modules/migration.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/migration.py
@@ -9,14 +9,15 @@ def migrate_module(self, pkg, stage):
     argument in the migrate decorator (explained in the docstring)
     to decide if we want to do something if a new module is installed
     during the migration.
-    We trick Odoo into running the scripts by setting the update attribute if necessary.
+    We trick Odoo into running the scripts by temporarily changing the module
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

jcdrubay:
Thanks @StefanRijnhart :)

pedrobaeza:
Hi, @StefanRijnhart I'm facing a problem with this patch in v16. I have added migration scripts in https://github.com/OCA/account-analytic/pull/646, but they are not executed in the module installation:

image

Maybe the framework has changed in this version?

StefanRijnhart:
@pedrobaeza Yes indeed, a new install is being termined in an earlier phase now, and used to avoid calling `migrate_module`: https://github.com/odoo/odoo/commit/0e2b2c9c8a9fa53e80df7e0f9ae1f39275315460#diff-250c832b5654555cb37bc4e064c7d69e0c970ce2b0f5eba5301d94b9e973c8d2R173

pedrobaeza:
Thanks for the research. I was about to propose the `pre_init_hook` as well, as you may not have `openupgrade_framework` available when installing this module later. Doing the changes.

PR #2478, opened 2020-12-09
[13.0][FIX] Dependencies of procurement_jit not met yet

MiquelRForgeFlow:
Avoid loading purchase_stock in purchase PR. Reverts https://github.com/OCA/OpenUpgrade/commit/b100c6e3928bcafddd947b4c1b752f359cf28a90.

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

PR #3298, opened 2022-06-17
[ADD] 13.0 - procurement_jit : nothing to do

legalsylvain:
Trivial PR.

Note : ``openupgrade_analysis.txt`` file is not present. (Meld used)


legalsylvain:
Note : in fact, it was already done here by @pedrobaeza https://github.com/OCA/OpenUpgrade/pull/2472/files
but was reverted here by @MiquelRForgeFlow https://github.com/OCA/OpenUpgrade/pull/2478/files

In fact, I don't understand the reason of the revert. maybe for CI reason. 
well, let me know if you think there is extra things to do...

pedrobaeza:
Yes, it was reverted as the `stock` module wasn't yet migrated, and this one depends on it. Being "Nothing to do", it's not very worrying, but it's good to have it explicitly said.

Issue #2190, opened 2020-02-05
Migration to version 14.0


MiquelRForgeFlow:
`(del)` means that the module has been removed in this version
`(add)` means that the module has been added in this version
`>` means that it has been identified that the module has been renamed to the name below
`<` means that it has been identified that the module has been rename from the name below
`->` means that the module has been merged in the module below
`<-` means that the module is the target where the module below has been merged 

- [x] account - By @RaulOForgeFlow / @MiquelRForgeFlow / @JordiBForgeFlow / @pedrobaeza  - #2741
- [x] (del) account_analytic_default -> account
- [x] (del) account_analytic_default_hr_expense -> hr_expense
- [x] (del) account_analytic_default_purchase -> purchase
- [x] (del) account_bank_statement_import - Adopted in OCA as `account_statement_import`
- [x] account_check_printing - By @tranngocson1996 - #3044 
- [ ] account_debit_note 
- [x] (new) account_edi - By @tranngocson1996 - #3040 
- [x] (new) account_edi_proxy_client - By @tranngocson1996 - #3042 
- [ ] account_edi_ubl 
- [x] account_facturx > account_edi_facturx - By @tranngocson1996 - #3046
- [ ] (new) account_fleet 
- [ ] account_lock 
- [x] account_payment - By @MiquelRForgeFlow - #2805
- [ ] account_qr_code_sepa 
- [ ] account_tax_python 
- [ ] account_test 
- [ ] (new) adyen_platforms 
- [x] analytic - By @RaulOForgeFlow - #2705 
- [x] association - By @tranngocson1996 - #3045 
- [x] attachment_indexation  - By @pedrobaeza - #3007
- [ ] auth_ldap 
- [x] auth_oauth - By @MiquelRForgeFlow - #2830 
- [x] auth_password_policy - By @tranngocson1996 - #3047 
- [ ] auth_password_policy_portal 
- [ ] auth_password_policy_signup 
- [x] auth_signup - By @MiquelRForgeFlow - #2791
- [ ] auth_totp 
- [ ] auth_totp_portal 
- [x] barcodes - By @MiquelRForgeFlow - #2810 
- [x] base - By @StefanRijnhart - #2482
- [x] base_address_city - By @RaulOForgeFlow - #2697 
- [x] base_address_extended - By @RaulOForgeFlow - #2698 
- [x] base_automation - By @RaulOForgeFlow / @pedrobaeza  - #2670 
- [ ] (del) base_gengo 
[...]

PR #3299, opened 2022-06-17
[ADD] 14.0 - delivery - migration done

legalsylvain:
Quite simple migration script.

- ``length`` has been renamed into ``packaging_length``
- I added comments in the ``noupdate_changes.xml`` to say why I don't load the data.

thanks for your review.


legalsylvain:
Hi. 

i don't understand why CI is failing.

At the end of the migration openupgrade try to make a delete on ``res.partner`` table...
https://github.com/OCA/OpenUpgrade/runs/6943798791?check_suite_focus=true#step:13:5198

```
ERROR: update or delete on table "res_partner" violates foreign key constraint "fleet_vehicle_assignation_log_driver_id_fkey" on table "fleet_vehicle_assignation_log"
DETAIL:  Key (id)=(31) is still referenced from table "fleet_vehicle_assignation_log".
```

In fact, in the last green build the delete are present at the end of the openupgrade log, but no error is raised... 
https://github.com/OCA/OpenUpgrade/runs/6932502486?check_suite_focus=true#step:13:5177

Thanks for your help.

legalsylvain:
Something is wrong in the CI. Don't you think ?
https://github.com/OCA/OpenUpgrade/pull/3300 is red, but I only added a text file...

legalsylvain:
And I rerun the test for https://github.com/OCA/OpenUpgrade/pull/3289 (that was green). And it is now red.

pedrobaeza:
This is because Odoo has added a new restriction that makes this to fail (as Odoo part is always got dynamically, which is good, as it serves to see things that can fail in real scenarios):

https://github.com/odoo/odoo/commit/91ceb950f90d42d07970aa4a43fe9a6f3e6ff629

We have to prevent such deletion searching which data is removing, and preventing it to be removed.

MiquelRForgeFlow:
FYI: https://github.com/OCA/OpenUpgrade/pull/2998

pedrobaeza:
Then we can close this one, but that doesn't prevent to fix current problem in CI, which is independent from the migration of this module, and comes from an update upstream.

Miquel, can you check it?

legalsylvain:
> FYI: https://github.com/OCA/OpenUpgrade/pull/2998

Damn'it, the related main issue was not up to date. (https://github.com/OCA/OpenUpgrade/issues/2190)

PR #3302, opened 2022-06-20
[MIG][15.0] stock: migration script

sang250399:
This PR
=======
Migration done


openupgrade_scripts/scripts/stock/15.0.1.1/post-migration.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _fill_product_template_detailed_type(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE product_template
+        SET detailed_type = type
+        WHERE type = 'product'
+        """,
```
gdgellatly, line 11: Checked, only 'product' key given in stock module - is correct

openupgrade_scripts/scripts/stock/15.0.1.1/post-migration.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+
+def _fill_stock_move_is_inventory(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE stock_move
+            SET is_inventory = True
+            Where inventory_id IS NOT NULL;
```
gdgellatly: Just minor, looks weird `Where`

openupgrade_scripts/scripts/stock/15.0.1.1/post-migration.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+            UPDATE stock_picking_type
+            SET print_label = True
+            WHERE code = 'outgoing';
+        """,
+    )
+
+
+def _create_default_return_type_for_all_warehouses(env):
+    all_warehouses = env["stock.warehouse"].with_context(active_test=True).search([])
```
gdgellatly: do we want `active_test=False` - seems strange to set to True?

openupgrade_scripts/scripts/stock/15.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,42 @@
...
+        """,
+    )
+
+
+def _fill_stock_quant_package_name_if_null(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE stock_quant_package
+            SET name = 'Unknown Pack'
```
gdgellatly, line 19: Is there a chance we get a lot of these? 

Maybe 
```suggestion
            SET name = 'Unknown Pack ' || id::text
```
I never use packages so maybe it is irrelevant, just a comment.

gdgellatly (commented):
I just static review, but I am not very good at this, it is more for other reviewers to confirm if I am stupid than for dev to take action.

openupgrade_scripts/scripts/stock/15.0.1.1/post-migration.py
leotranviindoo:
```suggestion
            WHERE inventory_id IS NOT NULL;
```

openupgrade_scripts/scripts/stock/15.0.1.1/post-migration.py
leotranviindoo: something went wrong
```suggestion
    all_warehouses = env["stock.warehouse"].with_context(active_test=False).search([])
```

openupgrade_scripts/scripts/stock/15.0.1.1/pre-migration.py
leotranviindoo: This field is required from 15.0, but at 14.0 it was set default value via `self.env['ir.sequence'].next_by_code('stock.quant.package')` and didn't insert by SQL before. So I think we don't need set value for it during migration from 14.0 to 15.0.

leotranviindoo: After reviewing, I think we still need to migrate here. Although `stock.quant.package` are usually created via transfers but it has some views and end user can edit it. It mean `name` of `stock.quant.package` can be blank.

leotranviindoo (approved):

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3510.

PR #3311, opened 2022-06-20
[MIG][15.0] website_sale_stock: migration script

sang250399:
This PR
======
Migration done


daonamutc1:
Check everything is done, waiting for depend.

openupgrade_scripts/scripts/website_sale_stock/15.0.1.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+            END
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE product_template
+            SET show_availability = CASE
+            WHEN inventory_availability = 'threshold' THEN true ELSE false
+            END
```
CarlosRoca13, line 21:
```suggestion
            UPDATE product_template
        SET allow_out_of_stock_order = CASE
            WHEN inventory_availability IN ('never', 'custom') THEN true ELSE false END,
        show_availability = CASE
            WHEN inventory_availability IN ('always', 'threshold') THEN true ELSE false END,
        available_threshold = CASE WHEN inventory_availability = 'always' THEN 999999999 END
```

You can do it in one query and coverage all options from older versions

CarlosRoca13 (changes requested):

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3725.

PR #3322, opened 2022-06-20
[14.0][FIX] fleet: remove obsolete table fks

MiquelRForgeFlow:
Let's greenify the repo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ jobs:
       PGHOST: "localhost"
       PGPASSWORD: "odoo"
       PGUSER: "odoo"
+      OPENUPGRADE_USE_DEMO: "yes"
     steps:
       - name: Set up Python
         uses: actions/setup-python@v2
--- a/openupgrade_framework/odoo_patch/odoo/modules/graph.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/graph.py
@@ -1,5 +1,7 @@
 # Copyright Odoo Community Association (OCA)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import os
+
 import odoo
 from odoo.modules.graph import Graph
 
@@ -7,6 +9,8 @@ from odoo.modules.graph import Graph
 def update_from_db(self, cr):
     """Prevent reloading of demo data from the new version on major upgrade"""
     Graph.update_from_db._original_method(self, cr)
+    if os.environ.get("OPENUPGRADE_USE_DEMO", "") == "yes":
+        return
     if (
         "base" in self
         and self["base"].dbdemo
--- a/openupgrade_framework/readme/DESCRIPTION.rst
+++ b/openupgrade_framework/readme/DESCRIPTION.rst
@@ -17,6 +17,9 @@ OpenUpgrade migration scripts:
 * Run migration scripts for modules that are installed as new dependencies
   of upgraded modules (when there are such scripts for those particular
   modules)
+* Production databases generated with demo data, will be transformed to
+  non-demo ones. If you want to avoid that, you have to pass through the
+  environment variable OPENUPGRADE_USE_DEMO, the value "yes".
 
 For detailed documentation see:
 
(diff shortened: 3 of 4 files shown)


pedrobaeza:
This is not the solution, as the check is not at FK level. We should prevent that some data are removed marking them as noupdate=1 for example. Maybe this is something only of the current DB we are using, so the other alternative is to fix such DB.

legalsylvain (commented):
Hi @MiquelRForgeFlow. Thanks a lot for trying to fix that annoying CI problem.

FYI, if you want to make test locally, the database used for this 13 -> 14 migration is available here : 

https://github.com/OCA/OpenUpgrade/releases/download/databases/13.0.psql

regards.

MiquelRForgeFlow:
PR is green, ready to merge! :)

.github/workflows/test.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       PGHOST: "localhost"
       PGPASSWORD: "odoo"
       PGUSER: "odoo"
+      OPENUPGRADE_USE_DEMO: "ok"
```
legalsylvain: Minor remark / non blocking : value is not very classic. 1 or True could be better. Don't you think ?

openupgrade_framework/odoo_patch/odoo/modules/graph.py
```diff
@@ -1,12 +1,16 @@
...
+
 import odoo
 from odoo.modules.graph import Graph
 
 
 def update_from_db(self, cr):
     """Prevent reloading of demo data from the new version on major upgrade"""
     Graph.update_from_db._original_method(self, cr)
+    if os.environ.get("OPENUPGRADE_USE_DEMO", "") == "ok":
+        return
```
legalsylvain, line 13: Not sure to understand. With that path, in witch case the following request (``UPDATE ir_module_module SET demo = false`` will be executed ? 
- in the CI with demo data - > no.
- in production without demo data -> no.

Maybe I missed something, and sorry if my question is not relevant.

legalsylvain (commented):
Thanks a lot for your work @MiquelRForgeFlow.
Some questions : 

- (minor / non blocking) : as you introduce a new environment variable, could you add a few words in the description of the module ``openupgrade_framework``

- @StefanRijnhart : could you take a look on this one ?

openupgrade_framework/odoo_patch/odoo/modules/graph.py
MiquelRForgeFlow: It will only executed in the case of production databases generated with demo data and then the demo data was customized. Thus, the script will transform that database to a not demo one.

.github/workflows/test.yml
MiquelRForgeFlow: Not sure if it can be True. That's an yml file, not python file.

MiquelRForgeFlow: It seems we can only pass strings.

openupgrade_framework/readme/DESCRIPTION.rst
```diff
@@ -17,6 +17,9 @@ OpenUpgrade migration scripts:
 * Run migration scripts for modules that are installed as new dependencies
   of upgraded modules (when there are such scripts for those particular
   modules)
+* Production databases generated with demo data, will be transformed to
+  non-demo ones. If you want to avoid that, you have pass in environment
+  the key OPENUPGRADE_USE_DEMO = "yes".
```
pedrobaeza:
```suggestion
  non-demo ones. If you want to avoid that, you have to pass through the
  environment variable OPENUPGRADE_USE_DEMO, the value "yes".
```

openupgrade_scripts/scripts/fleet/14.0.0.1/pre-migration.py
```diff
@@ -9,3 +9,6 @@ def migrate(env, version):
         env.cr,
         {"fleet_vehicle_log_contract": [("state", None, None), ("name", None, None)]},
     )
+    openupgrade.remove_tables_fks(
```
pedrobaeza, line 12: Why is this needed?

MiquelRForgeFlow: because... obsolete tables, always better to remove the fks

PR #3325, opened 2022-06-20
[13.0][FIX] stock: avoid triggering prefix change on sequence code filling

MiquelRForgeFlow:
We need to avoid using write method in order to avoid changing the prefix. Thus, queries are used instead.

Selection_451


pedrobaeza (approved):
I experimented this problem but don't know why. Thanks for the catch!

PR #3330, opened 2022-06-22
[IMP][15.0] openupgrade_framework: tracking number of rows per table before and after the migration

royle-vietnam:
Tracking number of rows per table before and after the migration to avoid losing time checking

Result in log:
Screenshot from 2022-06-22 10-42-12

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/modules/loading.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/loading.py
@@ -1,0 +1,109 @@
+import logging
+
+import odoo
+
+_logger = logging.getLogger(__name__)
+
+
+def load_modules(registry, force_demo=False, status=None, update_module=False):
+    with registry.cursor() as cr:
+        cr.execute("""DROP TABLE IF EXISTS openupgrade_track_row_count""")
+        # create a technical table
+        # to track number of rows per table before and after the migration
+        cr.execute(
+            """
+            CREATE TABLE IF NOT EXISTS openupgrade_track_row_count (
+                id serial NOT NULL,
+                table_name character varying(128) NOT NULL,
+                old_row_count integer DEFAULT 0,
+                new_row_count integer
+            );
+            """,
+        )
+        # update old rows count of all table before running the migration
+        cr.execute(
+            """
+            WITH tbl AS (
+              SELECT table_name
+              FROM information_schema.tables
+              WHERE table_name not like 'pg_%'
+                  AND table_name not like 'ir_%'
+                  AND table_name not like 'test_%'
+                  AND table_name not like '%_test'
+                  AND table_name != 'openupgrade_track_row_count'
+                  AND table_schema = 'public'
+            )
+            INSERT INTO openupgrade_track_row_count (table_name, old_row_count)
+            SELECT
+              table_name,
+              (xpath('/row/c/text()',
+                query_to_xml(format('select count(*) AS c from %I', table_name),
+                false,
+                true,
+                '')))[1]::text::int AS old_row_count
+            FROM tbl;
+            """,
+        )
+
+    load_modules._original_method(
+        registry,
+        force_demo=force_demo,
+        status=status,
+        update_module=update_module,
+    )
+
+    with registry.cursor() as cr:
+        # update new number of rows of all table after running the migration
+        cr.execute(
+            """
+            WITH tbl AS (
+              SELECT id, table_name
+              FROM openupgrade_track_row_count
+            )
+            UPDATE openupgrade_track_row_count otrc
+            SET new_row_count =
+              (xpath('/row/c/text()',
+                query_to_xml(format('select count(*) AS c from %I', otrc.table_name),
+                false,
+                true,
+                '')))[1]::text::int
+            FROM tbl
+            WHERE (
+                SELECT 1
+                FROM information_schema.tables
+                WHERE table_name=otrc.table_name
+                ) > 0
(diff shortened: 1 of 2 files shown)


openupgrade_framework/odoo_patch/odoo/modules/loading.py
```diff
@@ -0,0 +1,110 @@
+import logging
+
+import odoo
+
+_logger = logging.getLogger(__name__)
+
+
+def load_modules(registry, force_demo=False, status=None, update_module=False):
+    with registry.cursor() as cr:
```
StefanRijnhart, line 9: Before you execute this operation, can you check if an update is going to take place? Otherwise there can be quite some overhead on large databases.

openupgrade_framework/odoo_patch/odoo/modules/loading.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+
+            msg_all = msg("Table name", "Old row count", "New row count")
+            for r in result:
+                msg_all += msg(r["table_name"], r["old_row_count"], r["new_row_count"])
+            _logger.warning(
+                "The following tables have different number of rows:\n" "%s\n" % msg_all
+            )
+    with registry.cursor() as cr:
+        cr.execute("""DROP TABLE IF EXISTS openupgrade_track_row_count""")
```
StefanRijnhart: I wouldn't mind keeping the table in the database for future reference. Would you consider truncating the table if it exists at the top of this method instead?

legalsylvain (approved):
Nice addition ! Thanks. Agree with Stefan remarks. Otherwise LGTM. Thanks !

tranngocson1996 (approved):

openupgrade_framework/odoo_patch/odoo/modules/loading.py
royle-vietnam: @StefanRijnhart okey, I fixed

openupgrade_framework/odoo_patch/odoo/modules/loading.py
royle-vietnam: This is not affected on large databases and the cost is quite small. I think there is no need to worry about it

legalsylvain:
Thanks !

pedrobaeza:
What is the goal of this? I mean, seeing such numbers won't clarify anything about what's going on IMO.

pedrobaeza:
@legalsylvain why do you merge so quick?

legalsylvain:
> @legalsylvain why do you merge so quick?

because of 3 positives reviews without negative remarks.

> What is the goal of this? I mean, seeing such numbers won't clarify anything about what's going on IMO.

IMHO this feature can : 
- sometimes warn user regarding undesired changes
- sometimes help the user to see that the migration went well.

As it is a simple new log, it doesn't hurt in my opinion. 

One remarks thought, now that you question this feature, maybe the log level would be more appropriate if it was INFO, because it is not really a problem.

in any case, sorry for the too fast merge due to my enthusiasm to see new great features coming (especially from new contributors)

Let me know if you want some revert / fixes / etc. regarding that feature.

pedrobaeza:
I'm also worried about the performance penalty and possible side effects of this new patch in the upgrade pipeline. I think this should be something optional.

openupgrade_framework/odoo_patch/odoo/modules/loading.py
StefanRijnhart: That implies that you have not seen a large database :grin:

StefanRijnhart:
Merged while one of my comments was disputed and another was not yet followed up. Only one review and additional concerns from Pedro. I'm sorry but I'm going to revert this.

StefanRijnhart:
@legalsylvain thanks for merging the revert. @royleviindoo can you propose this interesting feature again with a new PR?

openupgrade_framework/odoo_patch/odoo/modules/loading.py
StefanRijnhart: For reference: https://wiki.postgresql.org/wiki/Slow_Counting

legalsylvain:
> @legalsylvain thanks for merging the revert

Sorry again for my hasty action.

> I think this should be something optional.

:+1: two ways : 
- by environment variables
- by a new key in the config.cfg file

Any preference ? 

Question : regarding the log level, is it OK for INFO or WARN ? (or don't care ?)

pedrobaeza:
Environment variable seems the common way, and better to be info

gdgellatly:
Maybe we have a VERBOSE style envvar. Keep it simple but allow some of these extra things just bundled together rather than make things more complex. For me, it should be INFO log level. But you guys know best.

openupgrade_framework/odoo_patch/odoo/modules/loading.py
gdgellatly: Actually I was reading about slow count 2 days ago. [There is a much faster way.](https://www.citusdata.com/blog/2016/10/12/count-performance/) but still slow and also that article is nearly as old as Stefan's ref.

StefanRijnhart: Am I reading this wrong or is there no faster way that is actually exact? Count estimates are not really useful in this case.

StefanRijnhart:
It's not so much the verbosity as well as the overhead. But unlike Pedro I would be fine running this on every upgrade, I just did not want to have it run on every non-upgrade registry initialization.

PR #3336, opened 2022-06-23
[MIG][15.0] hr_holidays: migration script

tranngocson1996:
Migration done


remi-filament (approved):
Tested OK on our dabatase migration.
Thank you @tranngocson1996

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+# NOTHING TO DO
+
+hr_holidays  / hr.leave.type            / allocation_validation_type (selection): selection_keys is now '['no', 'officer', 'set']' ('['both', 'hr', 'manager']')
+# DONE: pre-migration: map hr -> officer, both / manager -> set
+
+hr_holidays  / hr.leave.type            / employee_requests (selection) : NEW required, selection_keys: ['no', 'yes'], req_default: function, hasdefault
+# DONE: pre-migration: set employee_requests as 'no' if allocation_type = 'no' else 'yes'
+
+hr_holidays  / hr.leave.type            / requires_allocation (selection): NEW required, selection_keys: ['no', 'yes'], req_default: function, hasdefault
+# DONE: default='yes', but set to 'no' if raise exception in _check_allocation_id
```
remi-filament: I think here we should actually set requires_allocation = 'no' if allocation_type was 'no'

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+hr_holidays  / hr.leave.type            / _order                        : _order is now 'sequence' ('id')
+hr_holidays  / hr.leave.type            / accruals_ids (one2many)       : NEW relation: hr.leave.accrual.plan
+hr_holidays  / hr.leave.type            / allocation_type (selection)   : DEL selection_keys: ['fixed', 'fixed_allocation', 'no']
+# NOTHING TO DO
+
+hr_holidays  / hr.leave.type            / allocation_validation_type (selection): selection_keys is now '['no', 'officer', 'set']' ('['both', 'hr', 'manager']')
+# DONE: pre-migration: map hr -> officer, both / manager -> set
+
+hr_holidays  / hr.leave.type            / employee_requests (selection) : NEW required, selection_keys: ['no', 'yes'], req_default: function, hasdefault
+# DONE: pre-migration: set employee_requests as 'no' if allocation_type = 'no' else 'yes'
```
remi-filament: Here we should have something a bit more complex : 
allocation_validation_type = 'officer'  and employee_requests = 'yes' if allocation_type = 'fixed_allocation'
allocation_validation_type = 'set' and employee_requests = 'no'  if allocation_type = 'fixed'
My understanding being that 'set' means that holidays can only be set by manager, not requested by employee (as shown in https://github.com/odoo/odoo/blob/5ad9767784e6413e388535cbc7ea74e82275b647/addons/hr_holidays/tests/test_allocation_access_rights.py#L70-L77)

remi-filament (changes requested):
Double-checking holiday type configuration, I think there is a misunderstading on mapping allocation_type and allocation_validation_type (from v14) with requires_allocation, employee_requests and allocation_validation_type (from v15)

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
+from openupgradelib import openupgrade
+
+
+def _fast_fill_hr_leave_employee_company_id(env):
```
pedrobaeza: Odoo already fills with an SQL the related fields since https://github.com/odoo/odoo/pull/65232, so please remove all this method for not bloating the migration scripts.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+        (hr_leave_id integer, hr_employee_id integer)
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO hr_employee_hr_leave_rel (hr_leave_id, hr_employee_id)
+        SELECT hl.id, hl.employee_id
+        FROM hr_leave hl
+        WHERE hl.holiday_type = 'employee'
```
pedrobaeza: It's not required:

```suggestion
        WHERE hl.holiday_type = 'employee'
            AND hl.employee_id IS NOT NULL
```

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE hr_leave
+        ADD COLUMN IF NOT EXISTS multi_employee bool""",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE hr_leave
```
pedrobaeza: No need to do this. It's going to always be False, as we have pre-filled employee_ids and there's no possibility of having more than one employee. Please remove this method.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        UPDATE hr_leave hl
+        SET employee_company_id = empl.company_id
+        FROM hr_employee empl
+        WHERE empl.company_id IS NOT NULL AND hl.employee_id = empl.id""",
+    )
+
+
+def _map_hr_leave_state(env):
+    openupgrade.map_values(
```
pedrobaeza: Please copy the column here in pre-migration, and do the map on post-migration. This way, we preserve old values for any possible extra migration script (imagine there's an extension module that was putting an information if cancel - if we don't copy the column, we lose where we should see -).

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+hr_holidays  / hr.leave.accrual.plan    / time_off_type_id (many2one)   : NEW relation: hr.leave.type
+hr_holidays  / hr.leave.accrual.plan    / transition_mode (selection)   : NEW required, selection_keys: ['end_of_accrual', 'immediately'], req_default: function, hasdefault
+# NOTHING TO DO: new model
+
+hr_holidays  / hr.leave.allocation      / accrual_limit (integer)       : DEL
+hr_holidays  / hr.leave.allocation      / active (boolean)              : NEW hasdefault
+# NOTHING TO DO
+
+hr_holidays  / hr.leave.allocation      / accrual_plan_id (many2one)    : NEW relation: hr.leave.accrual.plan, isfunction: function, stored
+# NOTHING TO DO: handled by ORM
```
pedrobaeza, line 64: You should pre-create the column for avoiding to trigger the compute method being False for all the records.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+        ALTER TABLE hr_leave_allocation
+        ADD COLUMN IF NOT EXISTS approver_id integer""",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE hr_leave_allocation
+        SET approver_id = CASE
+            WHEN second_approver_id IS NOT NULL THEN second_approver_id
+            ELSE first_approver_id END
```
pedrobaeza:
```suggestion
        SET approver_id = COALESCE(second_approver_id, first_approver_id)
```

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE hr_leave_allocation
+        SET state = 'confirm'
+        WHERE state = 'validate1'""",
+    )
+
+
+def _convert_datetime_to_date_hr_leave_allocation_date_from(env):
+    openupgrade.logged_query(
```
pedrobaeza, line 38: You have to first rename the existing columns, then create a new ones, and then transfer data converted to date.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+        env.cr,
+        """
+        UPDATE hr_leave_allocation
+        SET date_to = date_to::date
+        WHERE date_to IS NOT NULL""",
+    )
+
+
+def _fast_fill_hr_leave_allocation_employee_company_id(env):
```
pedrobaeza: Another one that is not needed.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+        (hr_leave_allocation_id integer, hr_employee_id integer)""",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO hr_employee_hr_leave_allocation_rel (
+            hr_leave_allocation_id, hr_employee_id)
+        SELECT hla.id, hla.employee_id
+        FROM hr_leave_allocation hla
+        WHERE hla.holiday_type = 'employee'
```
pedrobaeza:
```suggestion
        WHERE hla.holiday_type = 'employee'
            AND hla.employee_id IS NOT NULL
```

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/pre-migration.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def _fast_fill_hr_leave_allocation_multi_employee(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE hr_leave_allocation
+        ADD COLUMN IF NOT EXISTS multi_employee boolean""",
+    )
+    openupgrade.logged_query(
```
pedrobaeza, line 167: Only pre-create the column. No need the other SQL as it's always False.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+# DONE: pre-migration: fast fill employee_company_id's value from employee_id.company_id on hr.leave.allocation model
+
+hr_holidays  / hr.leave.allocation      / employee_ids (many2many)      : NEW relation: hr.employee, isfunction: function, stored
+# DONE: pre-migration: create hr_employee_hr_leave_allocation_rel table and fill data
+
+hr_holidays  / hr.leave.allocation      / interval_number (integer)     : DEL
+hr_holidays  / hr.leave.allocation      / interval_unit (selection)     : DEL selection_keys: ['days', 'months', 'weeks', 'years']
+hr_holidays  / hr.leave.allocation      / lastcall (date)               : NEW hasdefault
+hr_holidays  / hr.leave.allocation      / number_per_interval (float)   : DEL
+# NOTHING TO DO: handled by ORM
```
pedrobaeza: This should be converted to an accrual plan instead, and link the allocation with it.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+
+hr_holidays  / hr.leave.allocation      / multi_employee (boolean)      : NEW isfunction: function, stored
+# DONE: pre-migration: create column and fill data
+
+hr_holidays  / hr.leave.allocation      / state (selection)             : selection_keys is now '['cancel', 'confirm', 'draft', 'refuse', 'validate']' ('['cancel', 'confirm', 'draft', 'refuse', 'validate', 'validate1']')
+# DONE: map validate1 -> confirm
```
pedrobaeza: This doesn't consist only on mapping the value, but also it should clear up the activities. A call to `activity_update` may serve.

openupgrade_scripts/scripts/hr_holidays/15.0.1.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,198 @@
...
+
+hr_holidays  / hr.leave.allocation      / taken_leave_ids (one2many)    : NEW relation: hr.leave
+hr_holidays  / hr.leave.allocation      / unit_per_interval (selection) : DEL selection_keys: ['days', 'hours']
```
pedrobaeza: This one goes also to the accrual plan.

pedrobaeza (changes requested):

sergio-teruel:
@tranngocson1996 Can you attend the comments? Thanks

MiquelRForgeFlow:
@pedrobaeza I am not able to recreate plans/levels, because not enough data. I see some vague map between hr.leave.allocation.unit_per_interval -> hr.leave.accrual.level.added_value_type and hr.leave.allocation.interval_unit -> hr.leave.accrual.level.frequency, but not enough old info to create those plans/levels with all those new fields.

pedrobaeza (approved):
Accrual plans are not created from previous existing allocations, as there's no 100% mapping between them, so that TODO is there, but being a very special case, let's continue with the merge.

PR #3342, opened 2022-06-23
[MIG][15.0] sale: migration script

hoangtiendung070797:
This PR
======

Migration done

### DIFF
--- a/openupgrade_scripts/scripts/sale/15.0.1.2/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/sale/15.0.1.2/upgrade_analysis_work.txt
@@ -1,9 +1,15 @@
 ---Models in module 'sale'---
 ---Fields in module 'sale'---
 sale         / product.packaging        / sales (boolean)               : NEW hasdefault
+# NOTHING TO DO: new feature
+
 sale         / sale.order               / note (text)                   : type is now 'html' ('text')
+# DONE: pre-migration: convert text to html
+
 sale         / sale.order.line          / product_packaging_id (many2one): NEW relation: product.packaging, hasdefault
 sale         / sale.order.line          / product_packaging_qty (float) : NEW
+# NOTHING TO DO: handled in sale_stock module (renamed column from product_packaging to product_packaging_id)
+
 ---XML records in module 'sale'---
 NEW ir.actions.act_window: sale.action_accrued_revenue_entry
 NEW ir.actions.act_window: sale.action_sale_order_form_view
@@ -23,3 +29,4 @@ DEL ir.ui.view: sale.assets_tests
 DEL ir.ui.view: sale.crm_team_salesteams_view_kanban
 DEL ir.ui.view: sale.qunit_suite
 DEL ir.ui.view: sale.report_assets_common_inherit_sale
+# NOTHING TO DO: noupdate="0" records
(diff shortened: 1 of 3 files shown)


openupgrade_scripts/scripts/sale/15.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,32 @@
...
+---Fields in module 'sale'---
+sale         / product.packaging        / sales (boolean)               : NEW hasdefault
+# NOTHING TO DO: new feature
+
+sale         / sale.order               / note (text)                   : type is now 'html' ('text')
+# DONE: pre-migration: convert text to html
+
+sale         / sale.order.line          / product_packaging_id (many2one): NEW relation: product.packaging, hasdefault
+sale         / sale.order.line          / product_packaging_qty (float) : NEW
+# NOTHING TO DO: handled in sale_stock module (renamed column from product_packaging to product_packaging_id)
```
legalsylvain, line 11: Hi. Not sure to understand that point : 

1)  regarding ``product_packaging_id`` (v14 named ``product_packaging`` in ``sale_stock`` ; v15 named ``product_packaging_id`` in ``sale``) 
This rename should be done in ``sale`` module in my opinion. Otherwise, ``sale`` upgrade will create an empty column that will never be populated.
the ``sale_stock`` migration will be done later.

2) AFAIK, ``product_packaging_qty`` is a computed field, computed with an [onchange](https://github.com/odoo/odoo/blob/15.0/addons/sale/models/sale_order_line.py#L435). So, it will not be populated by the ORM, and should be done in a script in sale/pre-migration.py. 

Don't you think ?

legalsylvain (commented):
One blocking point. Otherwise, LGTM. Thanks !

openupgrade_scripts/scripts/sale/15.0.1.2/upgrade_analysis_work.txt
hoangtiendung070797: @legalsylvain I'll have one PR for sale_stock to handle it; I write this this way to guarantee that the database does not install sale_stock; you'll see why when I push PR sale_stock up. I'll have a press release shortly, and I hope you'll pay attention to it.

legalsylvain: OK. there is no emergency. Please ping me on ``sale_stock`` PR ! 

Note : FYI, in openupgrade scripts, if you want to make a conditional action on a base module (``sale``) only if a dependency module is installed (``sale_stock``) you could use this function. 
https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2228

hoangtiendung070797: @legalsylvain I have a PR https://github.com/OCA/OpenUpgrade/pull/3345 for the sale stock module to handle. Please go through it again. This will guarantee migration

gdgellatly: > I'll have a press release shortly, and I hope you'll pay attention to it.

I was really impressed for a moment, putting out press releases for PR's, then confused then realised PR != press release, it is pull request. Love your work.

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3516.

PR #3345, opened 2022-06-23
[MIG][15.0] sale_stock: migration script

hoangtiendung070797:
This PR
======
migration done
depends: https://github.com/OCA/OpenUpgrade/pull/3342

### DIFF
--- a/openupgrade_scripts/scripts/sale_stock/15.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/sale_stock/15.0.1.0/end-migration.py
@@ -1,0 +1,10 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    all_sale_order_lines = env["sale.order.line"].search(
+        [("product_packaging_id", "!=", False)]
+    )
+    for line in all_sale_order_lines:
+        line._onchange_update_product_packaging_qty()
--- a/openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
@@ -1,0 +1,24 @@
+from openupgradelib import openupgrade
+
+
+def _rename_field_product_packaging_to_product_packaging_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS product_packaging_id integer
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line
+        SET product_packaging_id = product_packaging
+        WHERE product_packaging IS NOT NULL
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    _rename_field_product_packaging_to_product_packaging_id(env)
--- a/openupgrade_scripts/scripts/sale_stock/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/sale_stock/15.0.1.0/post-migration.py
@@ -1,0 +1,31 @@
+from openupgradelib import openupgrade
+
+
+def date_to_datetime_fields(env):
+    # convert column from date to datetime
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order ALTER COLUMN effective_date TYPE TIMESTAMP WITHOUT TIME ZONE
+        """,
+    )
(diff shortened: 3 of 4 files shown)


openupgrade_scripts/scripts/sale_stock/15.0.1.0/end-migration.py
```diff
@@ -0,0 +1,10 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    all_sale_order_lines = env["sale.order.line"].search(
+        [("product_packaging_id", "!=", False)]
+    )
+    for line in all_sale_order_lines:
+        line._onchange_update_product_packaging_qty()
```
legalsylvain, line 10: naive question : 
- why do you realize this action in an end-migration.py file (and not in a post-migration.py) ?

optimization : 
- in case of huge database, maybe write an SQL request instead of calling orm native function could be better. 
(note : not a blocking point, can be done later by other contributors. This implementation is valid)

openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
+from openupgradelib import openupgrade
+
+
+def _rename_field_product_packaging_to_product_packaging_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS product_packaging_id integer
```
legalsylvain, line 9: As ``sale`` as been updated at this step, the new field ``product_packaging_id`` has been created by the ORM, don't you think ?

openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line
+        SET product_packaging_id = product_packaging
+        WHERE product_packaging IS NOT NULL
+        """,
+    )
```
legalsylvain, line 19: I think that realize in a pre-migration.py script of ``sale`` module the following pseudo code could be simplier. Don't you think ? 

```
if is_installed("sale_stock"):
    rename("product_packaging", "product_packaging_id")
```

legalsylvain (commented):
following the disussion on the other sale PR, some comments inline. Thanks for your contribution.

openupgrade_scripts/scripts/sale_stock/15.0.1.0/end-migration.py
gdgellatly: Well it would be end migration to call all installed modules overrides. 

Plus that onchange calls a few other functions. I think it is really end-migration. It does filter out Falses, so on the bright side it might only be slow for huge databases with lots of sale lines using packages.

legalsylvain: > Well it would be end migration to call all installed modules overrides.

Oh Thanks. nice catch.

hoangtiendung070797: @gdgellatly Thanks you very much

openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
hoangtiendung070797: @legalsylvain Your idea is excellent, however as you stated in the Sale module, it will be migrated first. Because 'product packaging id' has already been generated in this scenario, renaming will result in an error.

openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
hoangtiendung070797: @legalsylvain Your idea is excellent, however as you stated in the Sale module, it will be migrated first. Because product_packaging_id has already been generated in this scenario, renaming will result in an error.

openupgrade_scripts/scripts/sale_stock/15.0.1.0/pre-migration.py
hoangtiendung070797: @legalsylvain This is only a guarantee. I don't think that will be an issue, you may reconsider.

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3531.

PR #3348, opened 2022-06-30
[15.0][IMP] base: override to not raise an exception

royle-vietnam:
- [x] Because we captured the exception in _raise_view_error and archived that view, so info is None, but it is called to info.get('select') in NameManager.check, which will raise an exception AttributeError, so we need to override to not raise an exception

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
@@ -5,7 +5,7 @@ import logging
 from odoo import api
 from odoo.tools import mute_logger
 
-from odoo.addons.base.models.ir_ui_view import View
+from odoo.addons.base.models.ir_ui_view import NameManager, View
 
 _logger = logging.getLogger(__name__)
 
@@ -15,6 +15,21 @@ def _check_xml(self):
     """Mute warnings about views which are common during migration"""
     with mute_logger("odoo.addons.base.models.ir_ui_view"):
         return View._check_xml._original_method(self)
+
+
+def _check(self, view):
+    """Because we captured the exception in _raise_view_error and archived that view,
+    so info is None, but it is called to info.get('select') in NameManager.check,
+    which will raise an exception AttributeError,
+    so we need to override to not raise an exception
+    """
+    try:
+        return NameManager.check._original_method(self, view)
+    except AttributeError as e:
+        if e.args[0] == "'NoneType' object has no attribute 'get'":
+            pass
+        else:
+            raise
 
 
 def _raise_view_error(


pedrobaeza:
Why not returning an empty dict instead in the previous override?

royle-vietnam:
> Why not returning an empty dict instead in the previous override?

@pedrobaeza Can't do that, because it doesn't assign the reverse value to the `info`, line 2277
https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/models/ir_ui_view.py#L2277-L2278

pedrobaeza (approved):

pedrobaeza:
@StefanRijnhart @legalsylvain now that I have checked the code for the view validation, do you think it's correct to deactivate the view? I don't think so, as migrating between several versions, one view may be invalid in the first migration, but becomes valid in the following one, but deactivating it, the view will remain inactive at the end.

royle-vietnam:
> @StefanRijnhart @legalsylvain now that I have checked the code for the view validation, do you think it's correct to deactivate the view? I don't think so, as migrating between several versions, one view may be invalid in the first migration, but becomes valid in the following one, but deactivating it, the view will remain inactive at the end.

I think it's still a good idea to deactivate it. Because we have a warning about this, users can re-active it if needed

pedrobaeza:
I'm afraid you are not understanding the whole problem: they are legitimate views, that gets incorrect in an specific moment, but return to view correct later. Why this happens? Maybe the view with such XML-ID is not present in version N+1, but again reappears in version N+2.

legalsylvain:
Indeed. Agree with Pedro. Use case : migration from 12 to 16 with a lot of custom modules. 
The custom modules will not be ported at 13/14 ans 15 revision but indeed will be available at 16 release.

Dont know how to solve the problem thought. 
@StefanRijnhart any idea ?

pedrobaeza:
Simply not deactivating the views as we did in the past.

phamgiang2510:
@pedrobaeza Could we do something like this:
- When migrating from version N to N+1:
  - in pre-migration of `base` module, use `copy_columns` on the `active` field of `ir_ui_view` table
  - then, inactive the views as current
- When migrating from version N+1 to N+2:
  - in pre-migration of `base` module, re-activate all views which have `active` is false but have `openupgrade_legacy_..._active` is true

cc @royleviindoo @legalsylvain @StefanRijnhart

pedrobaeza:
I don't think we need such convoluted scheme, as invalid views are finally removed due to noupdate=0. What we only need is to not crash.

gdgellatly:
As an FYI, this is what Odoo does in enterprise migration. Disables the views and it is really really frustrating. We don't want to do that if we can avoid.

Actual post migration code from 12-14 Enterprise migration
```sql
-- Fixup Deactivated views

DELETE FROM ir_ui_view where inherit_id=1820;

UPDATE ir_ui_view SET active = TRUE WHERE id IN (
    1909, 1987, 1935, 1897, 1914, 1975,
    1925, 1932, 1899, 2070, 1920, 2496, 2500, 2114, 2978, 2521, 1930,
    2192, 2107,1924,1991,1892,2116,1931,2958,2065,2104,2495,2138,
    3109, 1936, 1896, 1915,1848,1923,1926,2205,1885,1943,1703,2965,
    3196, 1933, 1877, 1871,2041, 2062,1913,2115,958
);
```

royle-vietnam:
> Simply not deactivating the views as we did in the past.

I have added code to re-validate views to active it when all views is loaded.
For example view `res_users_view_form_simple_modif` is archived because view `res_users_view_form_profile` has not been loaded

pedrobaeza:
But why not simply remove the `active=False` clause?

royle-vietnam:
> But why not simply remove the `active=False` clause?

I just want to check the archived views when loading the modules

pedrobaeza:
That's not the technique used in the past and we don't know if there are side effects, so I prefer the previous approach of v13 not deactivating at all.

royle-vietnam:
> That's not the technique used in the past and we don't know if there are side effects, so I prefer the previous approach of v13 not deactivating at all.

If not archived, the clients will get an error on the interface and they don't know how to handle it

pedrobaeza:
If your version is the target one, and the view has disappeared, then the view will be removed by the XML-ID cleaning, so that's not a problem. Your approach is good only for views with no XML-ID linked.

royle-vietnam:
> If your version is the target one, and the view has disappeared, then the view will be removed by the XML-ID cleaning, so that's not a problem. Your approach is good only for views with no XML-ID linked.

Do I need to revert that commit to merge?

pedrobaeza:
Please restrict this PR to the original goal, and open a new PR for the other thing, just removing the active=False instruction.

royle-vietnam:
> Please restrict this PR to the original goal, and open a new PR for the other thing, just removing the active=False instruction.

Done, please re-review, thanks

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
```diff
@@ -17,6 +17,21 @@ def _check_xml(self):
...
+def _check(self, view):
+    """Because we captured the exception in _raise_view_error and archived that view,
+    so info is None, but it is called to info.get('select') in NameManager.check,
+    which will raise an exception AttributeError,
+    so we need to override to not raise an exception
+    """
+    try:
+        return NameManager.check._original_method(self, view)
+    except AttributeError as e:
+        if not view.active or e.args[0] == "'NoneType' object has no attribute 'get'":
```
pedrobaeza: Why checking `not view.active`? It has nothing to do with the error.

royle-vietnam: Yes, I remove it

MiquelRForgeFlow:
conflicts

royle-vietnam:
> conflicts

Okey, done

pedrobaeza (approved):

pedrobaeza:
This should be done also in v14, isn't it?

royle-vietnam:
> This should be done also in v14, isn't it?

No, because Odoo v14 doesn't check view like v15

MiquelRForgeFlow:
Fixed in https://github.com/OCA/OpenUpgrade/pull/3560.

PR #3349, opened 2022-07-01
[MIG][15.0] hr_expense: add migration for 15.0

daonamutc1:
This PR
Migration done


remi-filament (changes requested):
Hi @daonamutc1 
Thank you for your work !

I found an issue testing this migration script :
The payment_state is not correctly computed here, since in v14 for account_move of type entry it would always end up being False (as per https://github.com/odoo/odoo/blob/20b8453f4cc69caefef8811485f8c8764f19efdb/addons/account/models/account_move.py#L1473).
In v15, the [_compute_amount function](https://github.com/odoo/odoo/blob/ac81d3b87852f25a753f104007fb4e05be5b522a/addons/account/models/account_move.py#L1439) is modified to call _payment_state_matters function which is inherited by hr_expense here : https://github.com/odoo/odoo/blob/ac81d3b87852f25a753f104007fb4e05be5b522a/addons/hr_expense/models/account_move_line.py#L9

We would probably need to recompute payment_state on account_move first before assigning it to hr_expense otherwise you always get False and therefore not_paid on hr_expense_sheets.

pedrobaeza:
Superseded by #3689 to fix the problem reported by @remi-filament

PR #3358, opened 2022-07-01
[MIG][15.0] sms: migration for 15.0

daonamutc1:
This PR
Migration done


gdgellatly (approved):
LGTM - interesting the use of phone_mobile_search, it isn't even a non-stored field really, just a search helper.

PR #3359, opened 2022-07-01
[MIG][15.0] pad: migration for 15.0

daonamutc1:
This PR
Migration done


openupgrade_scripts/scripts/pad/15.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,28 @@
...
+                {
+                    "key": "pad.pad_server",
+                    "value": pad_configs[0]["pad_server"],
+                },
+                {
+                    "key": "pad.pad_key",
+                    "value": pad_configs[0]["pad_key"],
+                },
+            ]
+        )
```
gdgellatly, line 32:
```
else:
    <make log entry instructing user - similar to docstring, suggest WARNING level>
```

daonamutc1: I was add log warning, can you review again?

openupgrade_scripts/scripts/pad/15.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    else:
+        _logger.warning(
+            "Pad: The system has many pad_servers, you need to setup it manually"
```
gdgellatly:
```suggestion
            "Pad: The system has multiple pad servers. Odoo 15.0 supports only 1. No pad settings have been changed. This may require manual configuration after migration"
```

gdgellatly (approved):
LGTM

PR #3363, opened 2022-07-01
[MIG][15.0] mass_mailing: migration for 15.0

daonamutc1:
This PR
Migration done


openupgrade_scripts/scripts/mass_mailing/15.0.2.5/pre-migration.py
```diff
@@ -0,0 +1,148 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE mailing_trace
+        SET trace_status = CASE state
+            WHEN 'outgoing' THEN 'outgoing'
+            WHEN 'exception' THEN 'error'
+            WHEN 'sent' THEN 'sent'
```
gdgellatly, line 67: Personally I think for `sent` and `outgoing` we document they haven't changed and exclude from query. A non representative sample from client databases had 60% of records in these 2 states, so purely from performance perspective to only update what is needed to update.

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3524.

PR #3381, opened 2022-07-05
[MIG][15.0] crm_iap_lead: migration script

daiduongnguyen-odoo:
**This PR**
Migration done


openupgrade_scripts/scripts/crm_iap_mine/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,116 @@
+from openupgradelib import openupgrade
+
+_xmlid_renames = [
+    (
+        "crm_iap_mine.crm_iap_lead_industry_30",
```
gdgellatly, line 5: Are we sure?

I would have though xml_ids in a rename would be `crm_iap_lead.x -> crm_iap_mine.x` . More for my knowledge, as maybe this happens earlier in process.

duong77476: @gdgellatly whoops my mistake, indeed it would be crm_iap_lead.x -> crm_iap_mine.x, i 'll fix right away

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/3565.

PR #3388, opened 2022-07-05
[14.0] account: keep account_journal fields sequence_id and refund_sequence_id

alexis-via:
Useful for those who want to use the OCA v14 module account_move_name_sequence https://github.com/OCA/account-financial-tools/tree/14.0/account_move_name_sequence

If this PR is accepted, I plan to update the module account_move_name_sequence so that the modules initializes the fields sequence_id and refund_sequence_id from openupgrade_legacy_14_0_sequence_id and openupgrade_legacy_14_0_refund_sequence_id

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
@@ -132,6 +132,11 @@ def copy_fields(env):
             "account_journal": [
                 ("default_credit_account_id", None),
                 ("default_debit_account_id", None),
+                # We keep the M2O ir.sequence fields for those who
+                # want to use the OCA v14 module 'account_move_name_sequence'
+                # from OCA/account-financial-tools
+                ("sequence_id", None),
+                ("refund_sequence_id", None),
             ],
         },
     )


MiquelRForgeFlow:
So, you want to rename here those fields and then you rename them later again in that module. Isn't it better to not do any rename than doing this complex double rename?

pedrobaeza:
I agree. Let the current names and just rename to the new module column name.

MiquelRForgeFlow:
Or just call:

```python
openupgrade.lift_constraints(env.cr, "account_journal", "sequence_id")
openupgrade.lift_constraints(env.cr, "account_journal", "refund_sequence_id")
```

To avoid the deletion of the columns in the migration.

pedrobaeza:
But the columns are not removed on migration...

MiquelRForgeFlow:
Oh, yeah. Then I don't understand this PR. Because the columns in that module are named exactly the same as the odoo old ones. EDIT: @pedrobaeza they are deleted when doing update=all with odoo?

alexis-via:
The columns are deleted during the migration. That's why I made this PR.

alexis-via:
If lift_constraints() avoids the deletion of the columns during the migration, then I will update my PR to use this method.

MiquelRForgeFlow:
Yeah, list_constrains delete all the references for these columns. They will become columns without any field. Thus, later when loading the account_move_name_sequence, the new fields will reuse the columns.

pedrobaeza:
But again, you don't have to do that, as columns are not dropped.

MiquelRForgeFlow:
> The columns are deleted during the migration. That's why I made this PR.

@alexis-via how did you migrated? can you check exactly how/when the columns were deleted?

alexis-via:
OK, I now understand why my sequences where re-generated. It was caused by an old version of account_move_name_sequence. I'll see what happens with my next migration and I'll propose another PR if needed.

PR #3398, opened 2022-07-06
[CI] explicit CI lines

legalsylvain:
Log more info in the CI to understand https://github.com/OCA/OpenUpgrade/pull/3273 problem.


legalsylvain:
That's seems OK @pedrobaeza. don't you think ?

https://github.com/OCA/OpenUpgrade/runs/7211788568?check_suite_focus=true#step:13:56

pedrobaeza:
Yes, they are. The problem has been in certain CIs, like the `mail` one, but once I restarted the build, it was OK, so it's not clear what has happened.

legalsylvain:
OK. Could we merge this one ?

pedrobaeza:
Yes, it doesn't hurt and clarifies a bit for future issues.

legalsylvain:
Thanks.

PR #3425, opened 2022-07-07
[MIG][15.0] mass_mailing_event : nothing to do

daiduongnguyen-odoo:
**This PR
Nothing to do**


pedrobaeza (changes requested):
- Please only one module per PR.
- Don't add anything in the third column. That's done from the upgrade analysis process.
- Leave an space between the column delimiter and your words.

PR #3435, opened 2022-07-07
[MIG][15.0] sale_project: nothing to do

daiduongnguyen-odoo:
**This PR
Nothing to do**


remi-filament:
Hi @duong77476 I have an issue with that one, actually on my migration the database constraint is not automatically deleted you need to use openupgradelib.openupgrade.delete_sql_constraint_safely function by adding the following in pre-migration.py script :

```python


@openupgrade.migrate()
def migrate(env, version):
    # Disappeared constraint
    openupgrade.delete_sql_constraint_safely(
        env, "sale_project", "project_project", "sale_order_required_if_sale_line"
    )
```

Let me know if you take care of it, otherwise I will create another PR for that !
Best Regards,

pedrobaeza:
Thanks for the pointer, Rémi. Superseded by #3686

PR #3438, opened 2022-07-07
[MIG] sale_timesheet_edit: nothing to do

daiduongnguyen-odoo:
**This PR
Nothing to do**


docsource/modules140-150.rst
```diff
@@ -640,11 +640,11 @@ Module coverage 14.0 -> 15.0
...
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| sale_timesheet_edit                       |                      |Merged into sale_timesheet.                      |
+| |del| sale_timesheet_edit                       |Nothing to do         |Merged into sale_timesheet.                      |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |new| sale_timesheet_margin                     |                      |                                                 |
+| |new| sale_timesheet_margin                     |Nothing to do         |No analysis file. No DB layout changes.          |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| sale_timesheet_purchase                   |                      |                                                 |
+| |del| sale_timesheet_purchase                   |Nothing to do         |                                                 |
```
pedrobaeza: This is not correct. It should be merged with something.

pedrobaeza (changes requested):

docsource/modules140-150.rst
daiduongnguyen-odoo: Hi, sorry i have been out for a longtime and can't quite remember this, do you remember which module it should be merged with?

P/s: probably i will comeback for upgrade 15 -> 16 in june :v:

pedrobaeza: No, sorry, but if you don't remember, just remove this change and let the module non handled. It's also a good practice to not handle several modules in the same PR.

daiduongnguyen-odoo: Thank you, i'll keep in mind

pedrobaeza (approved):

PR #3446, opened 2022-07-08
[15.0][IMP] base: update uninstallable modules state from installed to uninstallable  to avoid warning

royle-vietnam:
Example:
`module adyen_platforms: not installable, skipped`

### DIFF
--- a/openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py
@@ -5,7 +5,7 @@ import logging
 
 from openupgradelib import openupgrade
 
-from odoo import tools
+from odoo import modules, tools
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't get the exact use case, but I would say that this is not correct. If `adyen_platforms` disappears, there should be an specific instruction merging it with other module or you should manually uninstall it on target, but not to magically do it disappear.

royle-vietnam:
> I don't get the exact use case, but I would say that this is not correct. If `adyen_platforms` disappears, there should be an specific instruction merging it with other module or you should manually uninstall it on target, but not to magically do it disappear.

When I have module that is not available on target (e.g. not upgraded)

pedrobaeza:
I still don't understand.

royle-vietnam:
In a cross version migration, I have module X in version 13, but I don't have module X in version 14, but target is version 15, and I have such module in version 15

pedrobaeza:
For that use case, you don't have to do anything, simply let the warning in version 14. If you do this, you won't have in version 15 the module marked as installed and no migration will be done.

royle-vietnam:
Closed because it is not necessary

PR #3450, opened 2022-07-11
[IMP][15.0] openupgrade_framework: to not crash, simply not deactivating the views

royle-vietnam:


legalsylvain:
As said here, I don't think disabling views are the good solution. https://github.com/OCA/OpenUpgrade/pull/3348#issuecomment-1172085534

How do you enable them after ? 

thanks !

pedrobaeza:
This is indeed not deactivating views. Just let the warning.

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
```diff
@@ -35,15 +35,7 @@ def _raise_view_error(
                 from_traceback=from_traceback,
             )
         except ValueError:
-            _logger.warning(
```
pedrobaeza, line 38: Don't remove the warning

legalsylvain:
> This is indeed not deactivating views. Just let the warning.

sorry, I misread.

same remark as pedro. Otherwise, LGTM. Thanks.

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
royle-vietnam: Okey, done

jcdrubay (approved):

pedrobaeza (approved):

pedrobaeza:
This can be backported to v14

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
```diff
@@ -34,16 +34,16 @@ def _raise_view_error(
...
                 "Can't render custom view %s for model %s. "
-                "Assuming you are migrating between major versions of "
-                "Odoo, this view is now set to inactive. Please "
-                "review the view contents manually after the migration.",
+                "Assuming you are migrating between major versions of Odoo. "
+                "Please review the view contents manually after the migration.\n"
+                "Error: %s",
                 self.xml_id,
                 self.model,
+                e,
```
pedrobaeza, line 45: One last doubt. Why including `e` in the log? Isn't this too verbose?

royle-vietnam: So that we can review faster, otherwise you won't know why it failed

royle-vietnam: > This can be backported to v14

Yes, of course

daiduongnguyen-odoo (approved):

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
pedrobaeza: This is not answering my question. Isn't this too verbose? Can you put an example of the output with the exception included in the log message?

pedrobaeza:
The added text is only one line, so it can be acceptable.

PR #3451, opened 2022-07-11
[11.0][FIX] stock: stock.move.line generation for tracked products

alexis-via:
Add support for generation of stock.move.line for scapping operations
Remove lot_name in SQL request when not needed
Add comments in mrp module

Fixes a very old but very important bug #2000

### DIFF
--- a/addons/stock/migrations/11.0.1.1/post-migration.py
+++ b/addons/stock/migrations/11.0.1.1/post-migration.py
@@ -4,7 +4,6 @@
 from odoo.exceptions import UserError
 from odoo.tools import float_compare
 from openupgradelib import openupgrade
-from psycopg2.extensions import AsIs
 
 
 @openupgrade.logging()
@@ -102,10 +101,11 @@ def set_partially_available_state(env):
 
 
 @openupgrade.logging()
-def create_stock_move_line(env):
+def create_stock_move_line_done_untracked(env):
     """This method creates stock.move.line recreated from old
-    stock.pack.operation records. These records are created only for done
-    moves, as for those not done, it's handled later:
+    stock.pack.operation records for products without tracking by lot/serial.
+    These records are created only for done moves,
+    as for those not done, it's handled later:
 
     * For outgoing or internal transfers without reserved quantity, clicking on
       "Check availability" will work.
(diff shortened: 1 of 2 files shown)


addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -189,11 +176,164 @@ def create_stock_move_line(env):
...
+        INSERT INTO stock_move_line (
+            create_date,
+            create_uid,
+            write_date,
+            write_uid,
+            date,
+            location_dest_id,
+            location_id,
+            lot_id,
+            move_id,
```
fcayre, line 297: lot_name is missing here, I think it could easily be added.

fcayre (commented):
I just tested your PR on my 10.0 > 12.0 with a lot of tracked lots and it seems to work for me (see my single small comment regarding lot_name on scrap moves). I do not know how to check the result thoroughly however, any clue welcome.

addons/stock/migrations/11.0.1.1/post-migration.py
alexis-via: lot_name is not useful when lot_id is set. The field "lot_name" is only used to store the lot name for receptions before creating the stock.production.lot and setting lot_id.

alexis-via:
Just a small reminder: this PR is VERY important when you migrate a DB from v10 to upper version with products tracked by lot. If you don't use this PR, you dangerously screw up lot tracability in your Odoo database. I think merging this PR should be a priority.

addons/stock/migrations/11.0.1.1/post-migration.py
```diff
@@ -189,11 +176,164 @@ def create_stock_move_line(env):
...
+    'reserved_quand_id' because it is a dead field.
+    stock.pack.operation/stock.operation.lot as v10 grouped together several
+    stock.move lines with the same product in the same operation record, making
+    it very difficult/time consuming to split the stock.pack.operation.lot across
+    each move.
+    So this method has a known limitation: if, on a picking, there are several
+    stock.move with the same product (not so common...), the
+    stock.move.line created by the migration will all be attached to the same
+    stock.move. Therefore, in this particular case, the sum of the qty of the
+    stock.move.line will be superior to the qty of the stock.move.
```
StefanRijnhart, line 205: Thanks for mentioning this. I don't know if having more than one stock move on the same picking with the same product is that rare. In fact, it was the goal of Pedro's earlier PR to fix this: https://github.com/OCA/OpenUpgrade/commit/91bba0cf0f4aac58ff812f59f38d7ea28eda7634

Would it be possible to check for these cases after this query and correct them using ORM operations? It woul then only be a performance hit for those databases for which the operation is necessary.

anikeenko-viktor:
Can confirm it has worked for me

PR #3458, opened 2022-07-15
[13.0][OU-IMP] account: Empty move partner on regular entries

pedrobaeza:
On pre-v13, account.move>partner_id was a related stored field that pointed to the first partner found on lines, but now on v13, this field points only to the partner of the invoices, being empty when you fill a regular entry.

For consistency, and also for avoiding problems like duplicating a journal entry, change the parter on the lines, and get a different
partner on header vs lines, we should empty the partner in all the existing regular entries.

@Tecnativa TT38085

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -1043,6 +1043,22 @@ def fill_account_move_line_missing_fields(env):
     """)
 
 
+def _empty_move_partner_field(env):
+    """On pre-v13, account.move>partner_id was a related stored field that pointed to
+    the first partner found on lines, but now on v13, this field points only to the
+    partner of the invoices, being empty when you fill a regular entry.
+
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):
Not really sure, but seems ok. Go on.

MiquelRForgeFlow:
Shouldn't be better to do a copy_column in pre for this field?

pedrobaeza:
Isn't my explanation clear enough?

pedrobaeza:
> Shouldn't be better to do a copy_column in pre for this field?

Preserving such data is useless, as it's the first partner found on lines. We are not destructing information.

MiquelRForgeFlow:
No. My concern is for migration issues, maybe some other modules that depends on `account` need this partner_id info for some script or related field.

MiquelRForgeFlow:
> We are not destructing information.

Ah, ok :+1:

Issue #3287, opened 2022-06-15
Migration to version 15.0


legalsylvain:
- Full list of modules available here : https://github.com/OCA/OpenUpgrade/blob/15.0/docsource/modules140-150.rst 

# Pending / done work

- [x] account - By @MiquelRForgeFlow - #3505
- [x] account_check_printing - By @MiquelRForgeFlow - #3515
- [x] account_debit_note - By @pedrobaeza - #4202
- [x] account_edi - By @duong77476 - #3314
- [x] account_edi_facturx - By @duong77476 - #3373
- [x] account_edi_ubl - By @duong77476 - #3374
- [x] account_fleet - By @pedrobaeza - #4505
- [x] account_invoice_report_utm_campaign - By @nayatec - #4599
- [x] account_payment - By @duong77476 - #3375
- [x] account_qr_code_sepa - By @pedrobaeza - #4209
- [x] analytic - By @sang250399 - #3271 
- [x] association - By @pedrobaeza - #4043
- [x] attachment_indexation - By @legalsylvain - #3366
- [x] auth_ldap - By @pedrobaeza - #4399
- [x] auth_oauth - By @MiquelRForgeFlow - #3467
- [x] auth_password_policy - By @sergio-teruel - #3638
- [x] auth_password_policy_signup - By @sergio-teruel - #3637
- [x] auth_signup - By @MiquelRForgeFlow - #3463
- [x] auth_totp - By @MiquelRForgeFlow - #3480
- [x] auth_totp_portal - By @MiquelRForgeFlow - #3484
- [x] barcodes - By @sang250399 - #3301
- [x] base - By @MiquelRForgeFlow - #2967
- [x] base_address_city - By @pedrobaeza - #3635
- [x] base_address_extended - By @pedrobaeza - #4403
- [x] base_automation - By @legalsylvain - #3367
- [x] base_geolocalize - By @duong77476 - #3377
- [x] base_iban - By @legalsylvain - #3368
- [x] base_import - By @legalsylvain - #3369
- [x] base_setup - By @sang250399 - #3268
- [x] base_sparse_field - By @hhgabelgaard - #3601
- [x] base_vat - By @duong77476 - #3378
- [x] board - By @legalsylvain - #3370
- [x] bus - By @sang250399 - #3270
- [x] calendar - By @MiquelRForgeFlow - #3466
- [x] calendar_sms - By @legalsylvain - #3371
- [x] contacts - By @legalsylvain - #3372
[...]

PR #3284, opened 2022-06-15
[MIG][15.0] account: migration script

sang250399:
This PR
======
Migration done

PR #3460, opened 2022-07-19
[15.0][UPD] apriori + analysis + coverage

MiquelRForgeFlow:
Here it goes.


legalsylvain (commented):
hum. thanks for this work. 

1) do you know why there are such changes between few monthes ? 
2) Some changes are annoying. I don't know how to handle the ``account`` case (for exemple). Do you have an idea ?

thanks. 

regards.

MiquelRForgeFlow:
1. Hehe. Do you now how many commits odoo have added in those months? First year after new versions is always like this.
2. Don't worry. For `account`, some scripts will be added in 15.0.1.1 version and maybe some others in 15.0.1.2 folder. Depending of what has changed between them. To know what has changed between them, we will need to manually compare both analysis and detect the difference.

legalsylvain:
> Don't worry.

Hum, I'm not sure : I'm mean, 
- the current upgrade.txt file in the folder 15.0.1.1 is a diff of 14.0 and 15.0.1.1
- the current upgrade.txt file in the folder 15.0.1.2 is a diff of 14.0 and 15.0.1.2

AFAIU, the current upgrade.txt file in the folder 15.0.1.2 should be a diff of 15.0.1.1 and 15.0.1.2 (maybe minor).

What do we do ? 

Option A) We assume two revision.15.0.1.1 and 15.0.1.2 and we provides script for both "migrations". We have to change the upgrade.txt file 15.0.1.2 to be only the diff between 15.0.1.1 and 15.0.1.2. And we should take care that in the migration, the execution will be the following : 
- 15.0.1.1/pre-migration.py
- 15.0.1.2/pre-migration.py
- 15.0.1.1/post-migration.py
- 15.0.1.2/post-migration.py

So, in the pre-migration of the  15.0.1.2 the database is not totally a "15.0.1.1" (15.0.1.1/post-migration.py has not been executed). I faced similar issue when upgrading queue_job. (see : https://github.com/OCA/queue/tree/12.0/queue_job/migrations)

Option B) : we remove all the 15.0.1.1 folder. but it requires to rebase / adapt all the current (https://github.com/OCA/OpenUpgrade/pull/3284) for instance.

CC : @StefanRijnhart, @pedrobaeza : what do you think ?

pedrobaeza:
Everything should be done directly in `15.0.1.2` and directly the other folder must disappear.

jcdrubay:
Then does it make sense to keep the name `15.0.1.2`, why not only keeping `15.0`? or why not simply removing that folder from the path?

pedrobaeza:
The folder name must always match the latest version

legalsylvain:
> Everything should be done directly in 15.0.1.2 and directly the other folder must disappear.

Well, I'm in favour of this solution (for simplicity). But I'm not very confortable with the fact to say to 15.0 contributors, to add extra work to rebase and recheck diff, and fix some stuff. The work provided here https://github.com/OCA/OpenUpgrade/issues/3287 is important, and it lacks code review. (we should talk about that topic in the next coming soon Openupgrade Strategy meeting).

> Then does it make sense to keep the name 15.0.1.2, why not only keeping 15.0? or why not simply removing that folder from the path?

It doesn't work. And it doesn't solve the main problem. (1 folder or two folders)

> The folder name must always match the latest version

well, it is possible to have multiple folders, but it's a mess.

MiquelRForgeFlow:
I adapted the `account` issue:
Selection_491
Selection_492

pedrobaeza:
> Well, I'm in favour of this solution (for simplicity). But I'm not very confortable with the fact to say to 15.0 contributors, to add extra work to rebase and recheck diff, and fix some stuff. The work provided here https://github.com/OCA/OpenUpgrade/issues/3287 is important, and it lacks code review. (we should talk about that topic in the next coming soon Openupgrade Strategy meeting).

Complain to Odoo, not to us. The work is the same even if the PR is merged: the scripts must be passed to the other folder and recheck the diff. If not, migration may not work.

docsource/modules140-150.rst
```diff
@@ -670,6 +714,82 @@ Module coverage 14.0 -> 15.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | survey                                          |                      |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
+| test_access_rights                              |                      |No DB layout changes.                            |
```
pedrobaeza: Test modules shouldn't appear here

MiquelRForgeFlow: why?

pedrobaeza: Because they are noise and they are not installed in real databases. The script that populates this table should filter out them.

MiquelRForgeFlow:
Ready

openupgrade_scripts/scripts/bus/15.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2022 ForgeFlow S.L.  <https://www.forgeflow.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Disappeared constraint
+    openupgrade.logged_query(
```
pedrobaeza: Use the newly created openupgradelib feature

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -59,9 +59,31 @@ def _delete_channel_follower_records(env):
     )
 
 
+def delete_obsolete_constraints(env):
+    _contraints = [
```
pedrobaeza, line 65: Use the new openupgradelib method

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -39,6 +39,30 @@ def _map_mail_notification_failure_type(env):
...
+        env.cr,
+        """
+        SELECT mc.id
+        FROM mail_channel mc
+        WHERE mc.email_send""",
+    )
+    old_mail_groups = env["mail.channel"].browse([x[0] for x in env.cr.fetchall()])
+    if old_mail_groups:
+        try:
+            old_mail_groups.unlink()
```
pedrobaeza: Why doing this in ORM?

pedrobaeza: And shouldn't you remove one by one?

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -77,7 +79,199 @@ def delete_obsolete_constraints(env):
...
+    env.cr.execute("SELECT 1 FROM mail_moderation LIMIT 1")
+    has_moderation = env.cr.rowcount
+    env.cr.execute("SELECT 1 FROM mail_channel WHERE email_send LIMIT 1")
+    has_old_mail_groups = env.cr.rowcount
+    env.cr.execute(
+        """
+        SELECT 1 FROM ir_module_module
+        WHERE name = 'portal' AND state in ('installed', 'to upgrade', 'to install')""",
+    )
+    has_portal = env.cr.rowcount
```
pedrobaeza, line 93: Why not doing this if no portal?

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -77,7 +79,199 @@ def delete_obsolete_constraints(env):
...
+    openupgrade.rename_columns(
+        env.cr,
+        {
+            "mail_group_moderator_rel": [
+                ("mail_moderation_id", "mail_group_moderation_id"),
+            ]
+        },
+    )
+    # fill mail_group table
+    sql.create_model_table(
```
pedrobaeza, line 125: TIL!

openupgrade_scripts/scripts/mail_group/15.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,107 @@
...
+mail_group   / mail.group               / member_ids (one2many)         : NEW relation: mail.group.member
+mail_group   / mail.group               / moderation (boolean)          : NEW
+mail_group   / mail.group               / moderation_guidelines (boolean): NEW
+mail_group   / mail.group               / moderation_guidelines_msg (html): NEW
+mail_group   / mail.group               / moderation_notify (boolean)   : NEW
+mail_group   / mail.group               / moderation_notify_msg (html)  : NEW
+mail_group   / mail.group               / moderation_rule_ids (one2many): NEW relation: mail.group.moderation
+mail_group   / mail.group               / moderator_ids (many2many)     : NEW relation: res.users
+mail_group   / mail.group               / name (char)                   : NEW required
+# DONE (in mail): pre-migration: filled with mail.channel records that had email_send = True
```
pedrobaeza: I would put instead `# NOTHING TO DO: Done in mail pre-migration`, but isn't there lines for this in the `mail` work file?

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
MiquelRForgeFlow: Do you think it's better to delete the records by query?

openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
MiquelRForgeFlow: Because mail_group depends on portal module, and if portal is not installed, then we should avoid the installation of portal, right?

pedrobaeza: Ok

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
pedrobaeza: We will avoid side effects

MiquelRForgeFlow:
I tested the mail_group scripts in local, I fixed some things, and now it works like a charm :)

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
MiquelRForgeFlow: Yeah, I saw those effects, due to cascade deletions, and produced issues. Better by query, and thus no issues :)

pedrobaeza (approved):

PR #3461, opened 2022-07-20
[FIX][13.0] account: Recompute tax_exigible on account.move.line

phamgiang2510:
Current migration script does not fill values for `tax_exigible` column on the newly created `account.move.line`.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -575,6 +575,55 @@ def migration_voucher_moves(env):
     )
 
 
+def fill_account_move_line_tax_exigible(env):
+    # Compute the 'tax_exigible' field of newly created account move lines
+    # 1. Set tax_exigible = False for any lines (x) which have x.tax_line_id.tax_exigibility == 'on_payment'
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_move_line aml
+        SET tax_exigible = FALSE
+        FROM account_tax at
+        WHERE aml.tax_line_id = at.id
+            AND at.tax_exigibility = 'on_payment'
+            AND aml.tax_exigible IS NULL
+        """
+    )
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please do it at the same time that the lines are inserted.

phamgiang2510:
> Please do it at the same time that the lines are inserted.

It'll make the insert queries become too complicated and hard to maintain, so I think split it to a separate function is better.

pedrobaeza:
It's not better when you have millions of account.move.line and each SQL requires 20 minutes, so please include it in the main one. Properly written with `case` is still readable. Anyways, this case is only for draft invoices, isn't it?

legalsylvain:
Well. If it's harder to merge both requests or if @phamgiang2510 doesn't have time to do it, i'm in favor to accept this PR as is, and then to work on optimisation, if required.

Better to have a migration with 20 minutes more without bug than the reverse.

Dont you think ?

pedrobaeza:
I'm not. It's already very heavy to add more for not too much added value. We have migrated tons of instances to v13 as this hasn't been a problem for now, so it's not so critical to add to everyone this migration time.

phamgiang2510:
@pedrobaeza This NULL value of `tax_exigible` will cause error if we continue migrating to v14, because of this [part](https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L2171:L2174), I cannot make payments for old invoices (raising error in [here](https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_partial_reconcile.py#L165:L166)). But I found out that Odoo fixed the problem 2 days ago in this [commit](https://github.com/odoo/odoo/commit/184b78d774b21df4cd5211e2301a00f1d92e3947), so this PR may not needed anymore :grin:

MiquelRForgeFlow:
Then this PR can be closed if not needed anymore.

PR #3337, opened 2022-06-23
[MIG][15.0] auth_signup: migration script

sang250399:
This PR
======
Migration script

PR #3463, opened 2022-07-20
[15.0][MIG] auth_signup: migration script

MiquelRForgeFlow:
Replaces https://github.com/OCA/OpenUpgrade/pull/3337.


legalsylvain:
Why do you créate a new PR, once another one existe. (And is validated.) ?

MiquelRForgeFlow:
Because I want to edit the original PR but I cannot force push.

before:
Selection_497
after:
Selection_498

PR #3280, opened 2022-06-15
[MIG][15.0] hr: Migration script

sang250399:
This PR
======
Migration done

PR #3481, opened 2022-07-27
[15.0][MIG] hr: Migration script

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/3280.


legalsylvain (changes requested):
same remark as in https://github.com/OCA/OpenUpgrade/pull/3480

PR #3490, opened 2022-07-28
[15.0][FIX] sales_team: put more coherent create_uid, create_date

MiquelRForgeFlow:
As discussed in https://github.com/OCA/OpenUpgrade/pull/3489.


pedrobaeza (approved):
No strong opinion on this and not critical either, so let's approve it.

PR #3344, opened 2022-06-23
[MIG][15.0] project: migration script

hoangtiendung070797:
This PR
======
migration done

PR #3491, opened 2022-07-28
[15.0][MIG] project: migration script

MiquelRForgeFlow:
Continuation of https://github.com/OCA/OpenUpgrade/pull/3344.


openupgrade_scripts/scripts/project/15.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,201 @@
...
+project      / project.milestone        / is_reached (boolean)          : NEW hasdefault
+project      / project.milestone        / message_follower_ids (one2many): NEW relation: mail.followers
+project      / project.milestone        / message_ids (one2many)        : NEW relation: mail.message
+project      / project.milestone        / message_main_attachment_id (many2one): NEW relation: ir.attachment
+project      / project.milestone        / name (char)                   : NEW required
+project      / project.milestone        / project_id (many2one)         : NEW relation: project.project, required, req_default: function, hasdefault
+project      / project.milestone        / reached_date (date)           : NEW isfunction: function, stored
+project      / project.milestone        / website_message_ids (one2many): NEW relation: mail.message
+project      / project.project          / activity_ids (one2many)       : NEW relation: mail.activity
+project      / project.project          / allow_task_dependencies (boolean): NEW hasdefault
```
pedrobaeza: I think we should enable this one by default.

openupgrade_scripts/scripts/project/15.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,201 @@
...
+project      / project.project.stage    / mail_template_id (many2one)   : NEW relation: mail.template
+project      / project.project.stage    / name (char)                   : NEW required
+project      / project.project.stage    / sequence (integer)            : NEW hasdefault
+# NOTHING TO DO: new feature
+
+project      / project.task             / allowed_user_ids (many2many)  : DEL relation: res.users
+# DONE: post-migration: add followers to project task records
+
+project      / project.task             / analytic_account_id (many2one): previously in module sale_timesheet
+# NOTHING TO DO: Handle by ORM
```
pedrobaeza:
```suggestion
# NOTHING TO DO: Handled by ORM
```

openupgrade_scripts/scripts/project/15.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,201 @@
...
+
+project      / project.task             / analytic_tag_ids (many2many)  : NEW relation: account.analytic.tag
+project      / project.task             / depend_on_ids (many2many)     : NEW relation: project.task
```
pedrobaeza: This can be taken from `project_task_dependency` OCA module

openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_columns(env.cr, _column_renames)
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE project_project
+        ADD COLUMN IF NOT EXISTS last_update_status CHARACTER VARYING;
```
pedrobaeza: Please make 2 queries

openupgrade_scripts/scripts/project/15.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,201 @@
...
+project      / project.task             / dependent_ids (many2many)     : NEW relation: project.task
+project      / project.task             / personal_stage_type_ids (many2many): NEW relation: project.task.type
+# NOTHING TO DO: new feature
+
+project      / project.task             / display_project_id (many2one) : NEW relation: project.project
+# DONE: fill value for project task with parent_id is NULL
```
pedrobaeza:
```suggestion
# DONE: post-migration: fill value for project task with parent_id is NULL
```

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
```diff
@@ -0,0 +1,108 @@
...
+        env["project.task"],
+        "project_task",
+        "user_ids",
+        openupgrade.get_legacy_name("user_id"),
+    )
+
+
+def _create_missing_project_task_stage_personal(env):
+    tasks = env["project.task"].with_context(active_test=False).search([])
+    tasks._populate_missing_personal_stages()
```
pedrobaeza: This is doing nothing, as there aren't any `project.task.stage.personal` record.

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
```diff
@@ -0,0 +1,108 @@
...
+    # update for 'done' and 'cancelled' tasks
+    openupgrade.logged_query(
```
pedrobaeza: This is not correct. Personal stages are not the same as regular ones. They are stages with specified owner. We don't have to populate anything here, as there are no personal stages yet.

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
```diff
@@ -0,0 +1,108 @@
...
+        FROM project_task pt
+        JOIN project_task_type ptt ON pt.stage_id = ptt.id
+        JOIN project_task_type ptt2 ON (
+            ptt2.name = ptt.name AND ptt2.user_id IS NOT NULL)
+        WHERE rel.task_id = pt.id AND rel.user_id = ptt2.user_id
+        """,
+    )
+
+
+def _add_followes_allowed_internal_user_to_project_project(env):
```
pedrobaeza:
```suggestion
def _add_followers_to_project_for_allowed_internal_users(env):
```

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
```diff
@@ -0,0 +1,108 @@
...
+        JOIN res_users users ON users.id = prj_user_rel.res_users_id
+        JOIN project_project prj
+            ON prj.id = prj_user_rel.project_project_id
+            AND prj.privacy_visibility = 'followers'
+        ON CONFLICT DO NOTHING
+        """,
+    )
+
+
+def _add_followes_allowed_portal_user_to_project_project(env):
```
pedrobaeza:
```suggestion
def _add_followers_to_project_for_allowed_portal_users(env):
```

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
```diff
@@ -0,0 +1,108 @@
...
+        JOIN res_users users ON users.id = prj_user_rel.res_users_id
+        JOIN project_project prj
+            ON prj.id = prj_user_rel.project_project_id
+            AND prj.privacy_visibility = 'portal'
+        ON CONFLICT DO NOTHING
+        """,
+    )
+
+
+def _add_followes_allowed_user_to_project_task(env):
```
pedrobaeza:
```suggestion
def _add_followers_to_task_for_allowed_users(env):
```

pedrobaeza (changes requested):

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
MiquelRForgeFlow: Yes, there are records. The table of model project.task.stage.personal is project_task_user_rel, which is also the m2m table between project.task and res.users, which already exists in v14.
Selection_502

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
MiquelRForgeFlow: No, the way it works now is different. Each time a task is created, the method `_populate_missing_personal_stages()` is called:
Selection_503
Thus we must run this method for the already created tasks to work as they were created in v15. And this last query just ensures that the stage assigned to the done and cancelled tasks receive the corresponding stage instead of the one the method assigns.

openupgrade_scripts/scripts/project/15.0.1.2/upgrade_analysis_work.txt
MiquelRForgeFlow: Selection_504

should I just copy this?

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
pedrobaeza: But such table is useless now AFAIK. Which are their contents in v14?

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
pedrobaeza: But I insist: these stages are not created by default as such table is not populated.

openupgrade_scripts/scripts/project/15.0.1.2/upgrade_analysis_work.txt
pedrobaeza: Yes. Some people insisted in keep the module, but for most of us, we should:

- Merge the module in `project`.
- Include such code (if correct).

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
MiquelRForgeFlow: Sorry, I was wrong, confused by other thing. This table is new in v15, and is the transformation of the m2o field `user_id` to `user_ids` as m2m, which are the assignees to the task. Now each task can have more that one assignee.

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
MiquelRForgeFlow: As I said above, the table is not empty. They are the assignees of the task. And all the stages are created (when calling the _populate_missing_personal_stages):
Selection_505
.

openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        SET allow_task_dependencies = TRUE
+        WHERE id in (
+            SELECT DISTINCT project_id as id
+            FROM project_task
+            WHERE id in (
+                SELECT DISTINCT task_id as id
+                FROM task_dependencies_rel
+            )
+        )""",
+    )
```
pedrobaeza: This should be unconditionally set to True. Not having right now any dependency doesn't mean you don't want to set from now on such dependencies. Even more, when you had on v14 the previous module installed, dependencies were always shown without any flag.

```suggestion
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE project_project
        SET allow_task_dependencies = TRUE""",
    )
```

openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def fill_project_project_last_update_status(env):
+    openupgrade.logged_query(
```
pedrobaeza: Why not using here the openugradelib method for adding a field with such default?

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
pedrobaeza: Now I understand how this goes. Again, the lack of comments and specially the whys makes this more difficult.

There's an error here in Odoo core, which is that the stages are not created in the user language. Maybe you can do such PR in upstream.

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def _create_missing_project_task_stage_personal(env):
+    tasks = env["project.task"].with_context(active_test=False).search([])
+    tasks._populate_missing_personal_stages()
+    # update for 'done' and 'cancelled' tasks
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE project_task_user_rel rel
```
pedrobaeza: Here you must put on the equivalent "Done" and "Canceled" personal stages for each task that is done or cancel in the main stage, and also you need to map to the first `fold` personal stage if the task stage is another one folded.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
MiquelRForgeFlow: It's already ok.

Selection_508

Odoo by default have:
The non-personal tasks (without user_id) are New, In progress, Done and Cancelled.
The personal ones (with user_id) are Inbox, Today, This Week, This Month, Later, Done and Cancelled.

The name is what defines the stage, and the user_id defines if it's personal or not. So the highlighted red part assures to find the right one. In this case, by default, only Done and Cancelled ones.

About the folding, there is nothing to do about. The ones "folded" are Done and Cancelled, so it's already handled as said above.

openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
MiquelRForgeFlow: I didn't find a method in `openupgradelib` to do this.

openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
MiquelRForgeFlow: In fact, I don't know if it's an error or not. AFAIK, the non-personal stages normally are created when the module is installed, so the name of those stages are in the lang of the user who installed the module. I don't see what should I fix. Also, if later new non-personal custom stages are added, they will appear in the language of their creator.

It's different about the personal stages, they have and user_id, and the name will appear in the language of that user_id.

Now that I think about it, it's a problem for the query if the languages are different. Maybe I should transform the query in an ORM method and handle the comparison using the correct language.

openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
MiquelRForgeFlow: Done, using `add_fields` method.

pedrobaeza (approved):

PR #3492, opened 2022-07-28
[15.0][MIG] product: migration script

MiquelRForgeFlow:
Continues https://github.com/OCA/OpenUpgrade/pull/3283.


openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        ALTER TABLE product_template_attribute_value
+        ADD COLUMN color integer
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE product_template_attribute_value
+        SET color = FLOOR(random() * 11 + 1)::int
```
gdgellatly, line 106: Mimics upstream default

gdgellatly (approved):
LGTM

openupgrade_scripts/scripts/product/15.0.1.2/pre-migration.py
MiquelRForgeFlow: Yeah, this is equivalent to `randint(1, 11)`

gdgellatly: @MiquelRForgeFlow Of course, just for other reviewers I find it helpful if you see something unusual to write that you checked it and it is correct. I do wonder for some of these things if it isn't worth a permalink to link to the upstream default function in a docstring or comment.

gdgellatly:
@MiquelRForgeFlow When we do these replacement PR's I think we need some way to update the PR migration issue, manually or otherwise. Is there a command for this?

StefanRijnhart (approved):

PR #3106, opened 2022-01-17
[MIG][14.0]point_of_sale: migration script

hieu1211:
This PR
======
Done

PR #3499, opened 2022-08-01
[14.0][MIG] point_of_sale: migration script

MiquelRForgeFlow:
Finishing https://github.com/OCA/OpenUpgrade/pull/3106.


openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/noupdate_changes.xml
```diff
@@ -1,27 +1,27 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="product_product_consumable" model="product.product">
```
pedrobaeza: Please comment whole record tag for avoiding unneeded writings.

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
+---Models in module 'point_of_sale'---
+---Fields in module 'point_of_sale'---
+point_of_sale / pos.config               / iface_precompute_cash (boolean): DEL
+point_of_sale / pos.config               / manage_orders (boolean)       : NEW
```
pedrobaeza: I think we have to enable this by default on existing configs, for providing this new feature, specially if coming from `pos_order_mgmt`. We are working on the transition from this module for this version (it seems the return is not available, but the config says that it exists). cc @chienandalu

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,56 @@
...
+def fill_sesion_on_picking_by_pos_order(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_picking
+        SET pos_session_id = (SELECT session_id
+                                FROM pos_order
+                                WHERE pos_order.id = stock_picking.pos_order_id
+                                LIMIT 1)
+        WHERE pos_session_id IS NULL AND pos_order_id IS NOT NULL;
```
pedrobaeza: Simplify this expression:

```suggestion
        UPDATE stock_picking sp
        SET pos_session_id =  po.session_id
        FROM pos_order po
        WHERE po.id = sp.pos_order_id
            AND sp.pos_session_id IS NULL
```

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,56 @@
...
+def transfer_new_session_state(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE pos_session ps
+        SET state = CASE
+                        WHEN pc.cash_control=True AND
+                        COALESCE(ps.rescue,False)=False THEN 'opening_control'
+                        ELSE 'opened'
+                    END
```
pedrobaeza: Make it more readable:

```suggestion
        UPDATE pos_session ps
        SET state = CASE
           WHEN pc.cash_control AND NOT COALESCE(ps.rescue, False)
           THEN 'opening_control'
           ELSE 'opened'
       END
```

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,56 @@
+from openupgradelib import openupgrade
+
+
+def convert_field_m2o_to_o2m(env):
```
pedrobaeza: Better method name?

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,53 @@
...
+def fill_pos_config_default_picking_type_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE pos_config
+        SET picking_type_id = (SELECT pos_type_id
+                                FROM stock_warehouse
+                                WHERE stock_warehouse.company_id = pos_config.company_id
+                                LIMIT 1)
+        WHERE picking_type_id IS NULL;
```
pedrobaeza:
```suggestion
        UPDATE pos_config pc
        SET picking_type_id = sw.pos_type_id
        FROM stock_warehouse sw
        WHERE sw.company_id = pc.company_id
            AND pc.picking_type_id IS NULL
```

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/upgrade_analysis_work.txt
chienandalu: Yes, it says it exists but it doesn't. For point_of_sale, v14 is an advance of the functionallity that landed in v15. Indeed, this field doesn't exist in that version anymore, as they backed away from this approach (and in v16 the UI flow is similar although the whole module got a refactor.

In a summary I'd be setting it on by default as it's going to be in future versions.

openupgrade_scripts/scripts/point_of_sale/14.0.1.0.1/post-migration.py
MiquelRForgeFlow: In fact, when I made the PR I didn't checked if the queries could be simplified. Thanks

PR #3379, opened 2022-07-05
[MIG][15.0] crm: migration script

daiduongnguyen-odoo:
**This PR**
Migration done

PR #3502, opened 2022-08-01
[15.0][MIG] crm: migration script

MiquelRForgeFlow:
Replaces https://github.com/OCA/OpenUpgrade/pull/3379.


openupgrade_scripts/scripts/crm/15.0.1.6/post-migration.py
```diff
@@ -0,0 +1,17 @@
+from openupgradelib import openupgrade
+
+
+def try_delete_noupdate_records(env):
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "crm.mail_alias_lead_info",
```
pedrobaeza, line 8: Not sure about removing this one...

MiquelRForgeFlow: https://github.com/OCA/OpenUpgrade/pull/3504

PR #3506, opened 2022-08-05
[14.0][FIX] account: take payment_ref only if move_name is NULL (or '')

MiquelRForgeFlow:
As said in the point: `# 2. try to match on statement move_name or payment_ref (if move_name empty)`

That means we should avoid case where move_name is null, i.e, the statement line is not reconciled.

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
@@ -245,11 +245,13 @@ def add_move_id_field_account_bank_statement_line(env):
         UPDATE account_bank_statement_line absl
         SET move_id = ap.move_id
         FROM account_payment ap
+        JOIN account_move am ON ap.move_id = am.id
         JOIN account_journal aj ON ap.journal_id = aj.id,
             account_bank_statement bs
         WHERE absl.statement_id = bs.id AND aj.company_id = bs.company_id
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
```diff
@@ -330,7 +333,8 @@ def add_move_id_field_account_payment(env):
         WHERE ap.journal_id = aj.id
             AND ap.move_id IS NULL
             AND am.name NOT IN ('', '/')
-            AND COALESCE(NULLIF(ap.move_name, ''), ap.payment_reference) = am.name
+            AND COALESCE(NULLIF(COALESCE(ap.move_name), ''),
```
pedrobaeza: This coalesce is doing nothing

pedrobaeza (changes requested):
Why not adding conditions in the WHERE about not null or similar for not having to do such weird comparisons?

MiquelRForgeFlow:
> Why not adding conditions in the WHERE about not null or similar for not having to do such weird comparisons?

Because they would become more wordy. Although it's a little illegible, it gets synthetically simple with those nested coalesce/nullif IMO.

openupgrade_scripts/scripts/account/14.0.1.1/pre-migration.py
MiquelRForgeFlow: Corrected.

pedrobaeza (approved):

PR #3510, opened 2022-08-08
[15.0][MIG] stock: migration script

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/3302.


pedrobaeza (approved):
It would be better to not overwrite the text columns with the HTML content, for preserving the content, but let's not block it and finally merge this. Doing real tests soon to fine-tune this.

PR #3304, opened 2022-06-20
[MIG][15.0] payment: migration script

sang250399:
This PR
======
Migration done

PR #3515, opened 2022-08-09
[15.0][MIG] account_check_printing: migration script

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/3304.


openupgrade_scripts/scripts/account_check_printing/15.0.1.0/post-migration.py
```diff
@@ -0,0 +1,23 @@
+from openupgradelib import openupgrade
+
+
+def _create_account_payment_method_line(env):
```
pedrobaeza, line 4: Can you please express the logic behind adding these lines?

MiquelRForgeFlow: The key is in the line `WHERE apm.code = 'check_printing'`. The payment methods of this code need corresponding payment method lines.

pedrobaeza: That's still no reason. Please indicate the code where that lines are needed, and why only for checks.

MiquelRForgeFlow: In `account` module, we have created the ones for:
Selection_722
In `payment` module, we have created the ones for:
Selection_721
So it's logical we need to populate in this PR the ones for `code = 'check_printing'`.
About why we need to create payment method lines for the payment methods... well, it's just the new methodology, we need records in this new model in order to make all this the payment method logic be functional. I mean, the payment methods in this account_check_printing module uses this code https://github.com/odoo/odoo/blob/14.0/addons/account_check_printing/models/account_journal.py#L69.

pedrobaeza: OK, that's what the reference I was looking for. Please include it as comment in the code and then I'll merge it.

MiquelRForgeFlow: Done

pedrobaeza: I don't see the code link in the comment...

MiquelRForgeFlow: Added code link. The problem is that it was too long for flake8 tastes, but well, I put a shortened version.

pedrobaeza: It doesn't serve, as the line code is a changing one. Put the link in 2 lines or put the ignore directive.

MiquelRForgeFlow: What do you mean about "`(...)/account_check_printing/models/account_journal.py#L69`" doesn't serve? I don't understand you. It's super-pretty-obvious that `(...)` means "`https://github.com/odoo/odoo/tree/15.0/addons/`".

pedrobaeza: The line 69 today will be line XX on 2 months. You need an absolute link with the SHA hash in the link (the called permalink).

MiquelRForgeFlow: Ah, and how I obtain that permalink?

pedrobaeza: Selección_014

MiquelRForgeFlow: Thanks. Ready!

pedrobaeza (approved):

PR #3516, opened 2022-08-10
[15.0][MIG] sale: migration script

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/3342.


remi-filament (approved):
tested OK for our database migration, thank you @MiquelRForgeFlow

openupgrade_scripts/scripts/sale/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,34 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.convert_field_to_html(
+        env.cr, "sale_order", "note", "note", verbose=False
```
pedrobaeza: Copy the column on pre to preserve the old contents, and convert it in post

openupgrade_scripts/scripts/sale/15.0.1.2/end-migration.py
```diff
@@ -0,0 +1,11 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    all_sale_order_lines = env["sale.order.line"].search(
+        [("product_packaging_id", "!=", False)]
```
pedrobaeza: This should be restricted to quotations, as we don't want past orders to be modified on the migration. It's legit to do it for keeping a consistent state on pending ones:

```suggestion
        [("product_packaging_id", "!=", False), ('state', 'in', ['draft', 'sent'])]
```

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #3531, opened 2022-08-16
[15.0][MIG] sale_stock: migration script

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/3345.


openupgrade_scripts/scripts/sale_stock/15.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,16 @@
+---Models in module 'sale_stock'---
+---Fields in module 'sale_stock'---
+sale_stock   / sale.order               / effective_date (date)         : type is now 'datetime' ('date')
+# DONE: post-migration: set effective_date = min(pickings.date_done)
```
pedrobaeza: This is not a new field, so you should simply convert `effective_date` to datetime.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #3535, opened 2022-08-18
[14.0][OU-ADD] website_snippet_carousel_product: Merged to website_sale

chienandalu:
In v14 Odoo introduces dynamic carousels, which is an equivalent approach to this OCA module. This migration script aims to convert the former ones into these brand new snippets.

cc @Tecnativa TT38078

please review @CarlosRoca13 @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration-website_snippet_carousel_product.py
+++ b/openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration-website_snippet_carousel_product.py
@@ -1,0 +1,112 @@
+# Copyright 2022 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from lxml import etree
+from openupgradelib import openupgrade
+
+
+def prepare_dynamic_carousel(defaults_dict):
+    return """
+<section
+    data-snippet="s_dynamic_snippet"
+    class="s_dynamic_snippet_carousel s_dynamic pt24 pb24 o_colored_level d-none"
+    data-name="Dynamic Carousel"
+    data-number-of-elements="{products_per_slide}"
+    data-number-of-elements-small-devices="1"
+    data-carousel-interval="{interval}"
+    data-filter-id="{filter}"
+    data-number-of-records="{products_limit}"
+    data-template-key="website.dynamic_filter_template_image_title_footer"
+    {extra_attribs}
+>
+    <div class="container o_not_editable">
+        <div class="css_non_editable_mode_hidden" />
+        {previous_xml}
+        <div class="dynamic_snippet_template"/>
+        {next_xml}
+    </div>
+</section>
+    """.format(
+        **defaults_dict
+    )
+
+
+def update_carousels(env, views_with_carousel):
+    # For each view analyze and extract the snippet
+    default_snippet_filter = env.ref("website_sale.dynamic_filter_demo_products")
+    for view in views_with_carousel:
+        tree = etree.fromstring(view.arch_db)
+        carousels = tree.xpath("//section[hasclass('s_product_carousel')]")
+        for carousel in carousels:
+            try:
+                data = tree.xpath("//*[hasclass('js_product_carousel')]")[0]
+            except IndexError:
+                continue
+            # The snnipet could have relevant extra attributes like t-if
+            carousel.attrib.pop("class")
+            extra_attribs = " ".join({f"{a}='{v}'" for a, v in carousel.attrib.items()})
+            # If there are sibling elements, we'd like to append keep them in the final
+            # carousel
+            previous_xml = next_xml = ""
+            if data.getprevious() is not None:
+                previous_xml = etree.tostring(data.getprevious(), encoding="unicode")
+            if data.getnext() is not None:
+                next_xml = etree.tostring(data.getnext(), encoding="unicode")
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration-website_snippet_carousel_product.py
```diff
@@ -0,0 +1,93 @@
...
+    </div>
+</section>
+    """.format(
+        **defaults_dict
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if env["ir.module.module"].search(
```
MiquelRForgeFlow:
```suggestion
    if not env["ir.module.module"].search(
```
also, I think this will not work, as the module is merged to website_sale, so AFAIK it will not appear in ir.module.module

I think you have to do a different check to know that that module was installed. for example, check if website_sale.s_product_carousel xmlid exists.

MiquelRForgeFlow (commented):
I don't understand why you did this in another file (post-migration-website_snippet_carousel_product.py) instead of the existing post-migration.py. I don't know if both will be executed :S

pedrobaeza:
All files starting with `post- ` are executed, but I agree about the check on `ir.module.module`, as that doesn't serve.

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration-website_snippet_carousel_product.py
chienandalu: You're right. My initial approach was to do it in `pre-` but the I realized I needed some xlm_ids from 14.0. I'll try to guess it otherwise...

chienandalu:
> I don't understand why you did this in another file 

Mainly to separate this particular OCA module migration script from the core stuff. As @pedrobaeza says, all the `post-` files are executed.

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration-website_snippet_carousel_product.py
chienandalu: Finally. I simply check if there are OCA's carousel snippets in the website views and change them when they're found.

MiquelRForgeFlow (approved):
LGTM

I hope you fully tested the script :)

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration-website_snippet_carousel_product.py
```diff
@@ -0,0 +1,90 @@
...
+        <div class="dynamic_snippet_template"/>
+    </div>
+</section>
+    """.format(
+        **defaults_dict
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
```
MiquelRForgeFlow:
```suggestion
def migrate(env, version):
    if not env.ref("website_sale.s_product_carousel", raise_if_not_found=False):
        return
```

this way we avoid doing ir.ui.view search if that module is not installed.

chienandalu: Does the reference still exist in post-migration?

MiquelRForgeFlow: obviously, it's part of OU magics

chienandalu: Ok, I'll pinpoint it in my head :grin:

chienandalu: Suggestion added. Thanks @MiquelRForgeFlow :)

pedrobaeza (commented):
You need to use openupgradelib's `update_field_multilang` method, as the view is multi-language. You may use https://github.com/OCA/openupgradelib/blob/14ffca6e2f8d7033156ddbcbf9cbeffe9e56cfdd/openupgradelib/openupgrade_tools.py#L47 as well for helping you on the HTML transformations.

Check https://github.com/OCA/openupgradelib/blob/14ffca6e2f8d7033156ddbcbf9cbeffe9e56cfdd/openupgradelib/openupgrade_120.py as guide for some patterns.

chienandalu:
> You need to use openupgradelib's `update_field_multilang` method, as the view is multi-language. 

We're changing views content, which use xml_translation, so keeping the possible terms in the view code should keep the translation terms.

I've been improving the script thou to try to keep as many original variations made to the snnipet as possible.

>You may use https://github.com/OCA/openupgradelib/blob/14ffca6e2f8d7033156ddbcbf9cbeffe9e56cfdd/openupgradelib/openupgrade_tools.py#L47 as well for helping you on the HTML transformations.
> 
> Check https://github.com/OCA/openupgradelib/blob/14ffca6e2f8d7033156ddbcbf9cbeffe9e56cfdd/openupgradelib/openupgrade_120.py as guide for some patterns.

We need to get original data from the code and transform it into the new stuff, so a simple substitution is not enough.

pedrobaeza:
Let's keep it opened until a real test migration is done and checked. Please notify when done.

MiquelRForgeFlow:
@chienandalu have you done a real test?

PR #3536, opened 2022-08-22
[14.0][OU-FIX] mrp: Update context from action according to 14.0

victoralmau:
We need to update the context to the correct value because in v14 the context is not defined in the xml definition.

v13: https://github.com/odoo/odoo/blob/13.0/addons/mrp/views/mrp_production_views.xml#L421
v14: https://github.com/odoo/odoo/blob/14.0/addons/mrp/views/mrp_production_views.xml#L588-L593

@Tecnativa TT38623


pedrobaeza:
This is not strictly speaking noupdate change, but missing key that keeps previous value in noupdate=0, but I think we can keep it that way if you put as comment why it's there.

MiquelRForgeFlow (approved):

PR #3537, opened 2022-08-23
[13.0][OU-IMP] point_of_sale: stock_location_id

chienandalu:
In v13 the deafault location of the picking type is used to take the stock in the picking operations of the pos configs. In the migration, this can lead to pickings that were using a common operation type taking the stock from the wrong locations. We prevent it creating new picking types for that configs.

cc @Tecnativa TT23199

### DIFF
--- a/addons/point_of_sale/migrations/13.0.1.0.1/openupgrade_analysis_work.txt
+++ b/addons/point_of_sale/migrations/13.0.1.0.1/openupgrade_analysis_work.txt
@@ -53,8 +53,10 @@ point_of_sale / pos.config               / iface_available_categ_ids (many2many)
 point_of_sale / pos.config               / iface_payment_terminal (boolean): DEL
 point_of_sale / pos.config               / limit_categories (boolean)    : NEW
 point_of_sale / pos.config               / other_devices (boolean)       : NEW
-point_of_sale / pos.config               / stock_location_id (many2one)  : DEL relation: stock.location, required, req_default: function
 # NOTHING TO DO
+
+point_of_sale / pos.config               / stock_location_id (many2one)  : DEL relation: stock.location, required, req_default: function
+# DONE: post-migration: for pos configs with picking types which default location is different from the former location, create new picking types with those default locations
 
 point_of_sale / pos.order                / payment_ids (one2many)        : NEW relation: pos.payment
 point_of_sale / pos.order                / statement_ids (one2many)      : DEL relation: account.bank.statement.line
--- a/addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
+++ b/addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
@@ -86,6 +86,35 @@ def delete_pos_order_line_with_empty_order_id(env):
     )
 
 
+def create_pos_picking_types(env):
+    """Create picking types according to former pos.config locations
+    """
+    openupgrade.logged_query(
+        env.cr, """
+        SELECT pc.id, stock_location_id
+        FROM pos_config pc
+        JOIN stock_picking_type spt ON pc.picking_type_id = spt.id
+        WHERE stock_location_id != spt.default_location_src_id
+        """
+    )
+    # Unzip and zip again so we browse all the configs at once
+    try:
+        config_ids, location_ids = zip(*env.cr.fetchall())
+    except ValueError:
+        # No values to unpack so no results were fetched
+        return
+    for config, location_id in zip(
+        env["pos.config"].browse(config_ids),
+        location_ids
+    ):
+        config.picking_type_id = config.picking_type_id.copy({
+            "name": "[OU] {} - {}".format(
+                config.picking_type_id.name, config.name
+            ),
+            "default_location_src_id": location_id,
+        })
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     fill_pos_config_default_cashbox_id(env)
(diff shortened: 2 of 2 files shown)


addons/point_of_sale/migrations/13.0.1.0.1/pre-migration.py
```diff
@@ -21,6 +21,22 @@
...
+    """
+    configs = env["pos.config"].with_context(
+        active_test=False
+    ).search([]).filtered(
+        lambda x: x.stock_location_id
+        != x.picking_type_id.default_location_src_id
+    )
+    for config in configs:
+        config.picking_type_id = config.picking_type_id.copy({
+            "name": "{} - {}".format(config.picking_type_id.name, config.name),
```
legalsylvain: Maybe add "openupgrade" in the name to make explicit the création by openupgrade ?

legalsylvain (approved):
Make sense. Thanks.

addons/point_of_sale/migrations/13.0.1.0.1/pre-migration.py
chienandalu: Yes, it's a good idea to be able to tell which ones were created by OU quickly.

chienandalu:
Ready for review @MiquelRForgeFlow @pedrobaeza

addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
```diff
@@ -86,6 +86,37 @@ def delete_pos_order_line_with_empty_order_id(env):
...
+    # Unzip and zip again so we browse all the configs at once
+    try:
+        config_ids, location_ids = zip(*env.cr.fetchall())
+    except ValueError:
+        # No values to unpack so no results were fetched
+        return
+    for config, location_id in zip(
+        env["pos.config"].browse(config_ids),
+        location_ids
+    ):
```
MiquelRForgeFlow, line 109:
```suggestion
    data = env.cr.fetchall()
    if not data:
        return
    for config_id, location_id in data:
        config = env["pos.config"].browse(config_id)
```
simpler

chienandalu: Sure, but i want to avoid browsing in every loop step. 1 query vs n queries

addons/point_of_sale/migrations/13.0.1.0.1/pre-migration.py
```diff
@@ -14,6 +14,9 @@
         ('account_move', None),
         ('invoice_id', None),
     ],
+    'pos_config': [
+        ('stock_location_id', None),
```
pedrobaeza: No need to do this. This is only done when we fear the field is going to be overwritten. Anyways, we can let it this way, but it complicates the later SQL queries.

addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
```diff
@@ -86,6 +86,37 @@ def delete_pos_order_line_with_empty_order_id(env):
...
+        config.picking_type_id = config.picking_type_id.copy({
```
pedrobaeza, line 110: Only do this if the location is not the same as the one in the picking type.

pedrobaeza (changes requested):
Commit message tag is `[OU-IMP]` or `[OU-FIX]`, not only `[OU]`

addons/point_of_sale/migrations/13.0.1.0.1/pre-migration.py
chienandalu: That was my thinking, but it was failing in CI when it tried to get the legacy column name...

addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
chienandalu: I'm already filtering them in the sql

addons/point_of_sale/migrations/13.0.1.0.1/pre-migration.py
MiquelRForgeFlow: Then don't use the legacy column name and use the normal column name 🤔

MiquelRForgeFlow:
Please, document the change in the openupgrade_analysis_work.txt file.

addons/point_of_sale/migrations/13.0.1.0.1/pre-migration.py
chienandalu: Ok. I was assuming it was always renamed...

chienandalu:
Changes done

addons/point_of_sale/migrations/13.0.1.0.1/openupgrade_analysis_work.txt
```diff
@@ -54,7 +54,7 @@ point_of_sale / pos.config               / iface_payment_terminal (boolean): DEL
 point_of_sale / pos.config               / limit_categories (boolean)    : NEW
 point_of_sale / pos.config               / other_devices (boolean)       : NEW
```
pedrobaeza, line 55:
```suggestion
point_of_sale / pos.config               / other_devices (boolean)       : NEW
# NOTHING TO DO

```

chienandalu: Done

pedrobaeza (approved):

PR #3543, opened 2022-08-30
[15.0][FIX] hr: load noupdate data

MiquelRForgeFlow:
Oversight.


pedrobaeza:
Why do you need to put such department and address on the admin employee? Isn't it archived by default?

MiquelRForgeFlow:
Ok, I just commented it. If in the future, in a new analysis, something else changes in the noupdate data, the load_data will automatically load it.

PR #2789, opened 2021-07-06
[14.0][MIG] hr

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/2777 (I couldn't edit that PR).

Issue #3060, opened 2022-01-09
[14.0] openupgrade_framework: Crash on certain view validations


pedrobaeza:
Steps to reproduce the problem:

- Have an old view with a tree with `colors` attribute.
- Remove such view in v14 code.
- Perform the migration to v14.

As a result, the migration process is aborted with this error:

<details>

```
Traceback (most recent call last):
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 677, in _tag_root
    f(rec)
  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 580, in _tag_record
    record = model._load_records([data], self.mode == 'update')
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4209, in _load_records
    data['record']._load_records_write(data['values'])
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_ui_view.py", line 1841, in _load_records_write
    super(View, self)._load_records_write(values)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4146, in _load_records_write
    self.write(values)
  File "/opt/odoo/auto/addons/website/models/theme_models.py", line 267, in write
    res = super(IrUiView, other_views).write(vals)
  File "/opt/odoo/auto/addons/website/models/ir_ui_view.py", line 68, in write
    return super(View, self).write(vals)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_ui_view.py", line 500, in write
    res = super(View, self).write(self._compute_defaults(vals))
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3704, in write
    fields[0].determine_inverse(real_recs)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1187, in determine_inverse
    getattr(records, self.inverse)()
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_ui_view.py", line 300, in _inverse_arch
    view.write(data)
  File "/opt/odoo/auto/addons/website/models/theme_models.py", line 267, in write
    res = super(IrUiView, other_views).write(vals)
  File "/opt/odoo/auto/addons/website/models/ir_ui_view.py", line 68, in write
    return super(View, self).write(vals)
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/models/ir_ui_view.py", line 500, in write
    res = super(View, self).write(self._compute_defaults(vals))
[...]

PR #3545, opened 2022-09-01
[14.0][FIX] fix odoobot_state field not find in parent view

zurawsm3:
odoobot_state not used in new odoo 14 'res.users.preferences.form.inherit' view. 

PR solves problem in https://github.com/OCA/OpenUpgrade/pull/2789 comments


MiquelRForgeFlow:
I think you should add it in your custom migration instead, but it cannot be accepted in here. Not everybody is having problems with this view. Other people is having problems with other views. The problem is noted in https://github.com/OCA/OpenUpgrade/issues/3060.

PR #3477, opened 2022-07-24
[14.0][FIX] openupgrade_framework: Proper module dependency resolution

pedrobaeza:
@Tecnativa

PR #3547, opened 2022-09-02
[15.0][FIX] openupgrade_framework: Proper module dependency resolution

pedrobaeza:
Add extra dependencies directly to the graph for immediate installation and proper dependency resolution.

Forward-port of #3477

@Tecnativa


legalsylvain (approved):
Thanks for porting this patch !

rvalyi (approved):

PR #3554, opened 2022-09-03
[14.0][OU-FIX] website_event_questions_free_text: Proper SQL query

pedrobaeza:
@Tecnativa


pedrobaeza:
Tested on real migration and working, so merging.

PR #3555, opened 2022-09-07
[12.0][FIX] crm: Remove calendar.event obsolete record rule

pedrobaeza:
This record rule has disappeared in this version and it was no update, so it should be removed.

https://github.com/OCA/OpenUpgrade/blob/4fa1430655c68300cb046f96e2943e6ff7ac2fe3/addons/crm/security/crm_security.xml#L37

@Tecnativa


MiquelRForgeFlow (approved):
It's not in the analysis because this rule was deleted in v12 at **Mar 11, 2021**.

pedrobaeza:
Yeah, a very bad "stable" patch

PR #3557, opened 2022-09-09
[15.0][FIX] mail: assure unicity in mail.channel.partner

MiquelRForgeFlow:
This model comes with a new mandatory constraint 'mail_channel_partner_partner_unique'.

Constraint:
Selection_526

Error:
Selection_527


legalsylvain (approved):
lgtm. Thanks.

pedrobaeza:
Why a merge is needed? Isn't better to simply remove extra occurrences?

MiquelRForgeFlow:
> Isn't better to simply remove extra occurrences?

Not sure. I did merge in order to their many2one fields have a "first_not_null" filling. As this model is both a normal model and a m2m model table, it happens that many fields are null. So if any of this extra records have any of these fields filled, I want to save them.

pedrobaeza (approved):

PR #3558, opened 2022-09-12
[14.0][FIX] crm: assure deprecated table doesn't give problems

MiquelRForgeFlow:
...


pedrobaeza:
Shouldn't we do this by default with all obsolete tables?

MiquelRForgeFlow:
Not really. Because until the migration finishes, there is no way to detect which tables are deprecated, and this kind of rare errors usually are needed to be solved in pre-migration.

PR #3559, opened 2022-09-13
[13.0][OU-FIX] point_of_sale: missing configs to change picking_type

chienandalu:
cc @Tecnativa TT38570


legalsylvain (approved):
Thanks.

PR #3568, opened 2022-09-21
[14.0][FIX] hr_timesheet: Add missing group in the rule

pedrobaeza:
In 13.0, there's only one group linked to this noupdate rule:

https://github.com/odoo/odoo/blob/affc9abbcf596e025af0fd33654af3fca709b2ed/addons/hr_timesheet/security/hr_timesheet_security.xml#L54

But in 14.0, there are two:

https://github.com/odoo/odoo/blob/23a18bc17ebf076000e68fdea4ab9ac8cefe46ba/addons/hr_timesheet/security/hr_timesheet_security.xml#L63

so we need to link the missing one.

It seems the noupdate changes extractor is not analyzing the attribute `eval`.

@Tecnativa


pedrobaeza:
@MiquelRForgeFlow it seems there's a flaw in the noupdate changes script.

MiquelRForgeFlow:
@pedrobaeza It's not a flaw. That group was added in https://github.com/odoo/odoo/commit/70a0ad280e0c8a62f24d6943be082298fa98e443, 4 months ago.

MiquelRForgeFlow (approved):

pedrobaeza:
OK then.

PR #3572, opened 2022-09-26
[14.0][OU-IMP] account: Don't copy useless fields

pedrobaeza:
Both account.payment and account.bank.statement.line are now inherited by delegation over move_id, which means that all the fields in the parent object are related non stored by default, and thus, there are no columns in the DB.

Copying the old columns instead of renaming them doesn't have sense then, as it will confuse any DB reader.


MiquelRForgeFlow:
If the field in those models is not stored, I thought that the existing column didn't influence in any way. What problem has caused you?

pedrobaeza:
Incongruent information checking the database directly. It's an imp, not a fix.

MiquelRForgeFlow (approved):
Ok, fine :)

PR #3575, opened 2022-09-28
[14.0][FIX] project: complete project visibility when 'followers' option

MiquelRForgeFlow:
Selection_592

We forgot to consider two cases in which the user could see the project in case of 'followers' option was selected.

Thus, we add the channel part. We don't add the task follower part, because that would make the task follower be able to see all the tasks of the project, but instead we add the task follower in the task visibility.


pedrobaeza:
There can be overlap, so you should add `ON CONFLICT DO NOTHING`

PR #3577, opened 2022-09-29
[14.0][FIX] account: Only fill account.move>payment_id when account.payment>move_id is set

MiquelRForgeFlow:
Oversight of https://github.com/OCA/OpenUpgrade/pull/3570.


pedrobaeza:
We should remove the matching by name or payment_ref as well, isn't it?

MiquelRForgeFlow:
Why? Do you have any issue with those queries?

pedrobaeza:
They can match things that are not correct, even for draft payments.

MiquelRForgeFlow:
Ok, wait, let me do some testings.

pedrobaeza (approved):

PR #3585, opened 2022-10-10
[15][mail] Avoid crash when migrating a database from v8

florian-dacosta:
`mail_notification` table exists in version 8 and is re-introduced in version 15.
We need to rename it to avoid a crash during the mail module migration in this case.
@StefanRijnhart @sang250399


StefanRijnhart (approved):
Confirmed! https://github.com/OCA/OpenUpgrade/blob/9.0/openerp/addons/base/migrations/9.0.1.3/openupgrade_general_log.txt#L152
Thanks for the fix!

MiquelRForgeFlow (approved):

PR #3586, opened 2022-10-10
[11.0][FIX] project: Fix project_issue uninstalled bug.

ddejong-therp:
This may seem like an insignificant change at first. However, I came across this because I was trying to upgrade a database that once had module `project_issue` installed, but not anymore. The migration script of `project` check for the existence of table `project_issue` to decide whether it will do something with these issues: https://github.com/OCA/OpenUpgrade/blob/4fa1430655c68300cb046f96e2943e6ff7ac2fe3/addons/project/migrations/11.0.1.1/post-migration.py#L226

However, if the module is not installed, column `label_issues` wil not exist on table `project_project`, and that will cause an error.

So in short, if you had module `project_issue` installed once but not anymore, you would get an error. This fixes it.

### DIFF
--- a/addons/project/migrations/11.0.1.1/post-migration.py
+++ b/addons/project/migrations/11.0.1.1/post-migration.py
@@ -223,7 +223,7 @@ def remove_mail_subtypes(env):
 @openupgrade.migrate()
 def migrate(env, version):
     set_default_values(env)
-    if openupgrade.table_exists(env.cr, 'project_issue'):
+    if openupgrade.is_module_installed(env.cr, 'project_issue'):
         convert_issues(env)
     openupgrade.load_data(
         env.cr, 'project', 'migrations/11.0.1.1/noupdate_changes.xml'


StefanRijnhart (approved):
Nice one!

StefanRijnhart (changes requested):
Interestingly, in this edition of OpenUpgrade we still had unit tests, and your change breaks the unit test.

```
2022-10-10 14:58:11,123 7143 INFO openupgrade project: test_convert_issues (test_project.TestProject)
2022-10-10 14:58:11,291 7143 ERROR openupgrade project: FAIL
2022-10-10 14:58:11,292 7143 INFO openupgrade project: ======================================================================
2022-10-10 14:58:11,292 7143 ERROR openupgrade project: FAIL: test_convert_issues (test_project.TestProject)
2022-10-10 14:58:11,292 7143 ERROR openupgrade project: Traceback (most recent call last):
2022-10-10 14:58:11,292 7143 ERROR openupgrade project: `   File "/home/travis/build/OCA/OpenUpgrade/addons/project/migrations/11.0.1.1/tests/test_project.py", line 13, in test_convert_issues
2022-10-10 14:58:11,293 7143 ERROR openupgrade project: `     self.assertTrue(project)
2022-10-10 14:58:11,293 7143 ERROR openupgrade project: ` AssertionError: project.project() is not true
2022-10-10 14:58:11,293 7143 INFO openupgrade project: Ran 1 test in 0.169s
2022-10-10 14:58:11,294 7143 ERROR openupgrade project: FAILED
2022-10-10 14:58:11,294 7143 INFO openupgrade project:  (failures=1)
```

I'm afraid that the module `project_issue` is already merged in `project` at this point, so the module `project_issue` will not look like it's installed anymore.

Can you try and find a workaround for this?

pedrobaeza:
I think the real issue is that the table `project_issue` wasn't removed in module uninstallation.

MiquelRForgeFlow:
@ddejong-therp I think you should delete that table in your custom pre-migration.

PR #3590, opened 2022-10-17
[ADD] rename account_menu into account_usability

legalsylvain:
Ref : https://github.com/OCA/account-financial-tools/pull/1275/

CC : @AlvaroTForgeFlow, @ChrisOForgeFlow


MiquelRForgeFlow (approved):

pedrobaeza:
Account is not yet merged, but this doesn't affect. You can save CI builds in this case, merging directly.

PR #3591, opened 2022-10-18
[14.0][OU-FIX] account: Switch state to cancel on dangling payments

pedrobaeza:
Coming from older versions of Odoo, you can have a payment with state "validated" or "reconciled", but no linked move, as it was deleted through the regular journal entry list.

On that situation, we are going to recreate the moves the same, but putting them on cancel state for not altering the accounting.

@Tecnativa TT39954

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -588,7 +588,8 @@ def fill_account_payment_with_no_move(env):
         p_data[p_id] = {
             "journal_id": p_journal_id,
             "name": p_name,
-            "state": p_state,
+            # Switch to cancel when no linked move, but the payment was validated
+            "state": "cancelled" if p_state not in {"draft", "cancelled"} else p_state,
             "payment_date": p_payment_date,
         }
         if p_company in p_dates_by_company:


openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -588,7 +588,8 @@ def fill_account_payment_with_no_move(env):
         p_data[p_id] = {
             "journal_id": p_journal_id,
             "name": p_name,
-            "state": p_state,
+            # Switch to cancel when no linked move, but the payment was validated
+            "state": "cancelled" if p_state not in {"draft", "cancelled"} else p_state,
```
MiquelRForgeFlow, line 592:
```suggestion
            "state": "cancelled" if p_state not in ("draft", "cancelled") else p_state,
```

pedrobaeza: This is done on purpose. It's more efficient the inclusion with a set instead of a tuple.

MiquelRForgeFlow:
> you can have a payment with state "validated" or "reconciled", but no linked move

the states were "posted", "sent" and "reconciled" hehe

pedrobaeza:
Not on v13.

MiquelRForgeFlow:
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_payment.py#L55

pedrobaeza:
Ah, you mean the technical value? I was talking about the displayed name... I would put technical name as `posted` then. And I miss intentionally `sent` as well. It's just for having the explanation. But if you think that it would made the difference, I change it.

MiquelRForgeFlow (approved):
Do as you prefer, but you usually are very spiky about commit messages hehe

StefanRijnhart (approved):

PR #3593, opened 2022-10-19
[16.0] [ADD] apriori.py: rename knowledge module

marielejeune:
Depends on https://github.com/OCA/knowledge/pull/368


pedrobaeza:
Thanks for doing this. As we haven't yet merged the 16.0 OU foundation, this is red, and it will probably conflict when merged. We let it here till then.

MiquelRForgeFlow:
Conflicts. Added this in https://github.com/OCA/OpenUpgrade/pull/3600.

PR #3595, opened 2022-10-20
[FIX] base: Uninstall l10n_it_edi to avoid conflict with l10n_it_fatturapa

SirTakobi:
From version `14.0`, module `l10n_it_fatturapa` conflicts with `l10n_it_edi`: https://github.com/OCA/l10n-italy/blob/bb780459361ad45900ccfb74aea22ca99b958d75/l10n_it_fatturapa/__manifest__.py#L16.

We assume that if `l10n_it_fatturapa` is installed, then `l10n_it_edi` is not to be installed: this is the accepted solution when this same error happens while installing the modules manually, see https://github.com/OCA/l10n-italy/blame/a5bb1352e155dd1e5d81feea5e631ad76543559e/l10n_it_fatturapa/README.rst#L70-L75.

Without this fix, when upgrading from `13.0` to `14.0`, the following is raised during the migration of `base`:
> odoo.exceptions.UserError: I moduli "ITA - Fattura elettronica - Base" e "Italy - E-invoicing" non sono compatibili.

(It means that mentioned modules are incompatible)

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
@@ -3,6 +3,20 @@
 from openupgradelib import openupgrade
 
 
+def uninstall_conflicting_it_edi(cr):
+    it_edi_conflicting_modules = ("l10n_it_edi", "l10n_it_fatturapa")
+    if all(
+        openupgrade.is_module_installed(cr.cr, m) for m in it_edi_conflicting_modules
+    ):
+        it_edi_module = cr["ir.module.module"].search(
+            [
+                ("name", "=", "l10n_it_edi"),
+            ],
+            limit=1,
+        )
+        it_edi_module.button_uninstall()
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     """Call disable_invalid_filters in every edition of openupgrade"""
@@ -14,3 +28,4 @@ def migrate(env, version):
         openupgrade.update_module_names(
             env.cr, [("web_diagram", "web")], merge_modules=True
         )
+    uninstall_conflicting_it_edi(env)
--- a/openupgrade_scripts/scripts/base/14.0.1.3/pre-migrate.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/pre-migrate.py
@@ -87,6 +87,20 @@ def deduplicate_ir_properties(cr):
     )
 
 
+def uninstall_conflicting_it_edi(cr):
+    it_edi_conflicting_modules = ("l10n_it_edi", "l10n_it_fatturapa")
+    if all(openupgrade.is_module_installed(cr, m) for m in it_edi_conflicting_modules):
+        # Mark as 'to_remove' to avoid raising a conflict; it will be installed anyway,
+        # but we will uninstall it for good in end-migration.
+        openupgrade.logged_query(
+            cr,
+            """
+            UPDATE ir_module_module
+            SET state='to remove'
+            WHERE name = 'l10n_it_edi'""",
+        )
+
+
 @openupgrade.migrate(use_env=False)
 def migrate(cr, version):
     """
@@ -125,6 +139,7 @@ def migrate(cr, version):
     # Perform module renames and merges
     openupgrade.update_module_names(cr, renamed_modules.items())
     openupgrade.update_module_names(cr, merged_modules.items(), merge_modules=True)
+    uninstall_conflicting_it_edi(cr)
     # Migrate partners from Fil to Tagalog
     # See https://github.com/odoo/odoo/commit/194ed76c5cc9
     openupgrade.logged_query(


legalsylvain (commented):
hi. I don't understand why you don't simply write : 
```
if openupgrade.is_module_installed(cr, "l10n_it_fatturapa"):
        openupgrade.logged_query(
            cr,
            """
            UPDATE ir_module_module
            SET state='to remove'
            WHERE name = 'l10n_it_edi'""",
        )
```
(eventually with an extra where clause where state in ['to install', 'installed', to upgrade'] ...

I mean what you want to ensure is : if ``l10n_it_fatturapa`` is installed, then ``l10n_it_edi`` is not. Right ?

SirTakobi:
> hi. I don't understand why you don't simply write :
> 
> ```
> if openupgrade.is_module_installed(cr, "l10n_it_fatturapa"):
>         openupgrade.logged_query(
>             cr,
>             """
>             UPDATE ir_module_module
>             SET state='to remove'
>             WHERE name = 'l10n_it_edi'""",
>         )
> ```
> 
> (eventually with an extra where clause where state in ['to install', 'installed', to upgrade'] ...
> 
> I mean what you want to ensure is : if `l10n_it_fatturapa` is installed, then `l10n_it_edi` is not. Right ?

Thanks for having a look!
Almost right: I want to ensure that if `l10n_it_fatturapa` is installed and `l10n_it_edi` would auto_install, then `l10n_it_edi` is not installed.
That is why in this PR, the code only edits the state of `l10n_it_edi` when it would be installed automatically (i.e. its dependencies are installed).

Your proposal would work too, but in my opinion it would work in too many cases: it would edit the state of `l10n_it_edi` even when it will not be auto installed.

Maybe it is worth mentioning that `l10n_it_fatturapa` does not depend on `l10n_it`.

legalsylvain:
> Your proposal would work too, but in my opinion it would work in too many cases: it would edit the state of l10n_it_edi even when it will not be auto installed.

I don't think so. here, we are the base pre-migration step. so the update of the module to install / update / removed has been done. 2 cases : 
- the state of ``l10n_it_edi`` is uninstalled -> nothing to do. (certainly because dependencies are not all installed).
- the state of ``l10n_it_edi`` is 'to install', 'to upgrade' and so we have to set 'to remove'.

don't you think ? (I'm not sure what I say). ;-)

SirTakobi:
> don't you think ? (I'm not sure what I say). ;-)

As far as I can see in https://github.com/OCA/OpenUpgrade/pull/3595#pullrequestreview-1148810224, your proposal is to uninstall `l10n_it_edi` whenever `l10n_it_fatturapa` is installed, so reviewing the two cases:
_First case_
> * the state of `l10n_it_edi` is uninstalled -> nothing to do. (certainly because dependencies are not all installed).

Right, there is nothing to do here, but:
* your proposal would set `l10n_it_edi` as 'to_remove' anyway, if `l10n_it_fatturapa` is installed
* this PR's code does nothing.

So I think this PR's code doesn't have to be changed.

_Second case_
> * the state of `l10n_it_edi` is 'to install', 'to upgrade' and so we have to set 'to remove'.

This happens both with or without your proposal.

Maybe you are suggesting to check if both `l10n_it_edi` and `l10n_it_fatturapa` are installed? Since you say that
> the update of the module to install / update / removed has been done

legalsylvain:
yes that was the sense of the text "(eventually with an extra where clause where state in ['to install', 'installed', to upgrade'] ..."

but in facts, I wonder if it's a problem of Open Upgrade.

I mean, if in version 14, you make an update all of the database with ``l10n_it_fatturapa`` installed, it will try to install also ``l10n_it_edi`` if it is auto_install to True. Right ?

SirTakobi:
> yes that was the sense of the text "(eventually with an extra where clause where state in ['to install', 'installed', to upgrade'] ..."

Oh I thought that was for `l10n_it_fatturapa`! Now I have changed the code, can you please check if that's better?

> but in facts, I wonder if it's a problem of Open Upgrade.
> 
> I mean, if in version 14, you make an update all of the database with `l10n_it_fatturapa` installed, it will try to install also `l10n_it_edi` if it is auto_install to True. Right ?

The `l10n_it_edi` was installed anyway despite the pre-migration, so it is right that an update all raised the error again.
Now I have added an explicit uninstall in end-migration so it effectively uninstalled in the end, and an update all does not raise the error again.

`l10n_it_edi` is tricky because it not `auto_install` as I thought it was, but installs anyway since https://github.com/odoo/odoo/pull/74305.

legalsylvain (approved):
LGTM.

legalsylvain:
Note, you can use the module ``module_change_auto_install`` on your production (and even in openupgrade process) to switch off / on  auto_install for a list of modules.

See : https://github.com/OCA/server-tools/pull/2091

openupgrade_scripts/scripts/base/14.0.1.3/pre-migrate.py
```diff
@@ -123,6 +137,7 @@ def migrate(cr, version):
     """
     )
     # Perform module renames and merges
+    uninstall_conflicting_it_edi(cr)
     openupgrade.update_module_names(cr, renamed_modules.items())
     openupgrade.update_module_names(cr, merged_modules.items(), merge_modules=True)
     # Migrate partners from Fil to Tagalog
```
MiquelRForgeFlow, line 143: I think it should go here the uninstall_conflicting_it_edi(cr) call instead.

SirTakobi: Thanks for the review!
The modules are not renamed so I didn't think of that, but I'm talking about `14.0` modules so it makes sense to move after the renaming.
I have moved it, please check if it's better now :)

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
```diff
@@ -3,6 +3,20 @@
...
+    if all(
+        openupgrade.is_module_installed(cr.cr, m) for m in it_edi_conflicting_modules
+    ):
+        it_edi_module = cr["ir.module.module"].search(
+            [
+                ("name", "=", "l10n_it_edi"),
+            ],
+            limit=1,
+        )
+        it_edi_module.button_uninstall()
```
pedrobaeza, line 17: Why doing 2 times the same?

pedrobaeza:
Isn't better to switch the auto-install to False in OCB and use that source for the migrations?

openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
SirTakobi: Thanks for having a look!
> Mark as 'to_remove' to avoid raising a conflict; it will be installed anyway,
but we will uninstall it for good in end-migration.

I wrote it in a comment in the code in https://github.com/OCA/OpenUpgrade/pull/3595/files#diff-aec3724a38e5c01c0ea5b62012c57a9145be5ab7058f222be86607cb7c50301fR93-R94.

SirTakobi:
> Isn't better to switch the auto-install to False in OCB and use that source for the migrations?

As I wrote earlier (https://github.com/OCA/OpenUpgrade/pull/3595#issuecomment-1285500637), `l10n_it_edi` is not `auto_install`:
> `l10n_it_edi` is tricky because it not `auto_install` as I thought it was, but installs anyway since [odoo/odoo#74305](https://github.com/odoo/odoo/pull/74305).

openupgrade_scripts/scripts/base/14.0.1.3/end-migration.py
pedrobaeza: Yeah, but why it will be installed? You are saying in other comment that is not auto-install. And if it's installed by another module, uninstalling this one will uninstall the other module as well.

SirTakobi: Just summarizing a bit and adding a few more details:
In `13.0` DB I have:

| name | state |
|-|-|
 l10n_it_edi | installed
 l10n_it_fatturapa | to upgrade

When I upgrade to `14.0` without this PR, the error in the description is raised:
> odoo.exceptions.UserError: I moduli "ITA - Fattura elettronica - Base" e "Italy - E-invoicing" non sono compatibili.

If I include only the `pre-migrate` part of this PR, the migration concludes successfully BUT `l10n_it_edi` and `l10n_it_fatturapa` are both installed; that is why I also added the `end-migration` to uninstall the module.

---
Now back to your question:
> Yeah, but why it will be installed?

`l10n_it_edi` is added to the graph of modules to be loaded in https://github.com/odoo/odoo/blob/a46d20c89c24a4e60faef09cf8ad7721bd29d6cc/odoo/modules/loading.py#L341-L345:
```
        cr.execute("SELECT name from ir_module_module WHERE state IN %s" ,(tuple(states),))
        module_list = [name for (name,) in cr.fetchall() if name not in graph]
        if not module_list:
            break
        graph.add_modules(cr, module_list, force)
```
because it is called including the state `to_remove` in https://github.com/odoo/odoo/blob/a46d20c89c24a4e60faef09cf8ad7721bd29d6cc/odoo/modules/loading.py#L455-L457:
```
            processed_modules += load_marked_modules(cr, graph,
                ['installed', 'to upgrade', 'to remove'],
                force, status, report, loaded_modules, update_module, models_to_check)
```

Odoo thinks that `l10n_it_edi` has to be updated because the following condition is True:
```
        needs_update = (
            hasattr(package, "init")
            or hasattr(package, "update")
            or package.state in ("to install", "to upgrade")
        )
```
(from https://github.com/odoo/odoo/blob/a46d20c89c24a4e60faef09cf8ad7721bd29d6cc/odoo/modules/loading.py#L161-L165)
more precisely, `hasattr(package, "update")` is True because it is assigned in https://github.com/odoo/odoo/blob/cbab786eb2e153eb7eefbd994cb261fa96e56800/odoo/modules/graph.py#L85-L87:
```
                for kind in ('init', 'demo', 'update'):
                    if package in tools.config[kind] or 'all' in tools.config[kind] or kind in force:
                        setattr(node, kind, True)
```
More precisely, `package in tools.config[kind]` is True.

pedrobaeza: Thank you for the detailed explanation. Wow, such a difficult thing to handle...

PR #3600, opened 2022-10-26
[16.0][IMP] apriori: initial filling

MiquelRForgeFlow:
...

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -5,6 +5,18 @@ to help the matching process
 # Renamed modules is a mapping from old module name to new module name
 renamed_modules = {
     # odoo
+    "coupon": "loyalty",
+    "payment_test": "payment_demo",
+    "payment_transfer": "payment_custom",
+    "pos_coupon": "pos_loyalty",
+    "sale_coupon": "sale_loyalty",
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/apriori.py
```diff
@@ -5,21 +5,56 @@
...
+    "stock.location.route": "stock.route",
+    "stock.production.lot": "stock.lot",
     # OCA/...
 }
 
 # only used here for upgrade_analysis
-merged_models = {}
+merged_models = {
+    # odoo
+    "account.account.type": "account.account",
```
pedrobaeza: I don't think this can be considered a merge...

MiquelRForgeFlow: But it's a merge in a sense, right? Ok, I will remove it.

legalsylvain: indeed, in a sense, but here merged models is used by the module_analysis to understand if fields are new or not or moved on a model. writing that will generate potential bad analysis on ``account.account`` model I think.

legalsylvain (commented):
- PR ok for me. @pedrobaeza what do you think ? Could be merge to allow other PR that depends on the initial one.

- do you think that we should push in the same time, the update of the analysis. (analysis should change, due to renamings / mergings...)

PR #3343, opened 2022-06-23
[MIG][15.0] hr_recruitment: migration script

sang250399:
This PR
=========
Migration done

PR #3624, opened 2022-11-22
[15.0][OU-ADD] hr_recruitment_notification: Merged hr_recruitment_notification to hr_recruitment.

victoralmau:
Merged `hr_recruitment_notification` to `hr_recruitment`.

Depends on:

- [x] #3343

@Tecnativa TT40294

### DIFF
--- a/openupgrade_scripts/scripts/hr_recruitment/15.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_recruitment/15.0.1.1/pre-migration.py
@@ -9,3 +9,13 @@ def migrate(env, version):
     openupgrade.convert_field_to_html(
         env.cr, "hr_applicant", "description", "description"
     )
+    # Rename hr_recruitment_notification module data to the core one (if present)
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/hr_recruitment/15.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.is_module_installed("hr_recruitment_notification"):
```
pedrobaeza: This is not correct, as in this moment, `hr_recruitment_notification` module doesn't exist (it was merged).

Just do the renaming. If not originally installed, the SQL operation won't update anything.

pedrobaeza (approved):
@victoralmau take into account that once you put a module in apriori, the XML-IDs are renamed to the new module name, so you have to put such XML-IDs on any later reference. I have already changed them, so ready to be merged.

PR #3628, opened 2022-11-24
[14.0][OU] product_expiry: Fill expiration_date

etobella:
When reviewing product_expiry results I found that the query was not following the same logic from odoo function

https://github.com/odoo/odoo/blob/14.0/addons/product_expiry/models/stock_move_line.py#L18-L27

I might be wrong, but function uses use_create_lots for the second function, but the migration one used use_existing_lots

Also, I removed unnecessary filters that were not present on the original function :thinking:

### DIFF
--- a/openupgrade_scripts/scripts/product_expiry/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/product_expiry/14.0.1.0/post-migration.py
@@ -29,12 +29,9 @@ def migrate(env, version):
         """
         UPDATE stock_move_line sml
         SET expiration_date = spl.expiration_date
-        FROM stock_production_lot spl, product_product pp
-        JOIN product_template pt ON pp.product_tmpl_id = pt.id, stock_picking sp
-        JOIN stock_picking_type spt ON sp.picking_type_id = spt.id
-        WHERE spl.id = sml.lot_id AND sml.picking_id = sp.id
-            AND spt.use_existing_lots AND sml.product_id = pp.id
-            AND pt.use_expiration_date""",
+        FROM stock_production_lot spl
+        WHERE spl.id = sml.lot_id AND spl.expiration_date IS NOT NULL
+        """,
     )
     openupgrade.logged_query(
         # use sml.create_date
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
Selection_666

Where is this conditions reflected in the query?

etobella:
Thanks @MiquelRForgeFlow I didn't notice that :cry:

openupgrade_scripts/scripts/product_expiry/14.0.1.0/post-migration.py
```diff
@@ -29,12 +29,9 @@ def migrate(env, version):
...
         UPDATE stock_move_line sml
         SET expiration_date = spl.expiration_date
-        FROM stock_production_lot spl, product_product pp
-        JOIN product_template pt ON pp.product_tmpl_id = pt.id, stock_picking sp
-        JOIN stock_picking_type spt ON sp.picking_type_id = spt.id
-        WHERE spl.id = sml.lot_id AND sml.picking_id = sp.id
-            AND spt.use_existing_lots AND sml.product_id = pp.id
-            AND pt.use_expiration_date""",
+        FROM stock_production_lot spl
+        WHERE spl.id = sml.lot_id AND spl.expiration_date IS NOT NULL
```
MiquelRForgeFlow, line 33: why too short of conditions here? 🤔 can you please point to the corresponding odoo code?

etobella: https://github.com/odoo/odoo/blob/14.0/addons/product_expiry/models/stock_move_line.py#L20-L21

The first IF has not many conditions, I think it could be better to keep it simple :smile:

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/product_expiry/14.0.1.0/post-migration.py
```diff
@@ -46,7 +43,11 @@ def migrate(env, version):
         FROM product_product pp
         JOIN product_template pt ON pp.product_tmpl_id = pt.id, stock_picking sp
```
pedrobaeza, line 44: AFAIK, `stock_picking` is not being used.

MiquelRForgeFlow: Selection_674

pedrobaeza: That's the join condition (which by the way it should be put with ON instead), but still no use.

MiquelRForgeFlow: I don't see it. sml cannot be used in join condition, because sml is in the update clause. And clearly you need the picking to link the sml to the picking type.

pedrobaeza: I insist: it's not used for the update itself. Please recheck the query.

pedrobaeza: This is still pending.

etobella: the updated table cannot be used on the ON, for that reason, I had to do it on the where. The Postgres error was the following: `TIP: There is an entry for the "sml" table, but it cannot be referenced from this part of the query.`

The query tries to fill the following Python script:
https://github.com/odoo/odoo/blob/14.0/addons/product_expiry/models/stock_move_line.py#L22-L25

The `stock.picking` is used to get the `stock.picking.type`, as we should only use picking types with `use_create_lots` TRUE

PR #3630, opened 2022-11-28
[13.0] update upstream 7cca00acacf

MiquelRForgeFlow:
It's time.


pedrobaeza:
Still red

MiquelRForgeFlow:
FYI: It's thanks to https://github.com/OCA/OpenUpgrade/pull/3630/commits/b89b70cc196bb44c62b175f09be2961500ea79c9 that we are detecting this kind of errors.

pedrobaeza:
That's why the guideline to include in the XML-ID a reference to the model for avoiding collapsing with the same one belonging to other models.

MiquelRForgeFlow:
Ready

pedrobaeza (approved):

PR #3641, opened 2022-12-09
[15.0][IMP] apriori: stock_inventory_valuation_pivot to stock_account

pilarvargas-tecnativa:
stock_inventory_valuation_pivot is already merged in stock_acount (Odoo 15.0)

Blocked by:

- [x] #3527

TT36995 

ping @chienandalu

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -56,6 +56,8 @@ merged_modules = {
     "hr_attendance_user_list": "hr_attendance",
     # OCA/product-attribute
     "stock_account_product_cost_security": "product_cost_security",
+    # OCA/stock-logistics-reporting
+    "stock_inventory_valuation_pivot": "stock_account",
     # OCA/stock-logistics-warehouse
     "stock_orderpoint_manual_procurement": "stock",
     # OCA/stock-logistics-workflow


chienandalu (approved):
:+1:

legalsylvain (approved):
Thanks.
however I don't think this PR is blocked. merging it will not raise the test of the migration stock_account. CI is based on the coverage file. 
So for me it's a :+1: to merge now.
Or did i missed something ?

pedrobaeza:
The merge operation of the module will mark the target module (`stock_account`) as "installed", provoking that the CI acts on such module, and thus also making that the whole chain to be tested.

legalsylvain:
> The merge operation of the module will mark the target module (stock_account) as "installed", provoking that the CI acts on such module, and thus also making that the whole chain to be tested.

I don't think so.
- stock_inventory_valuation_pivot is an OCA 14.0 module (https://github.com/OCA/stock-logistics-reporting/tree/14.0/stock_inventory_valuation_pivot) it is so NOT installed on the OCA test database available [here](https://github.com/OCA/OpenUpgrade/releases/download/databases/14.0.psql) and dowloaded [here](https://github.com/OCA/OpenUpgrade/blob/15.0/.github/workflows/test.yml#L22)

- stock_account is an Odoo 14.0 module installed on the test database, but will be set as uninstallable before CI execution[here](https://github.com/OCA/OpenUpgrade/blob/15.0/.github/workflows/test.yml#L117) as far as the module is not set as "Done" / "Nothing to do" in the coverage file.

As a result, merging the current PR will not raise the update of stock_account as both modules are uninstalled or not available.

I think we can test it running ocabot merge command. If it merge, it means CI is OK. Don't you think ?

(Not a big deal thought, but If I'm wrong, I'd like to understand why to use Openupgrade correctly !).

pedrobaeza:
The conflicting part is that one:

https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1591

Right now won't be a problem for this PR, but a combination of merge modules may lead to the situation I'm commenting, so better to be self-ordered and only merge in the dependency chain order.

sergio-teruel (approved):

ernestotejeda (approved):

PR #3454, opened 2022-07-13
[MIG][15.0] website_form: nothing to do

daiduongnguyen-odoo:
**This PR
Nothing to do**

PR #3643, opened 2022-12-10
[15.0][OU-ADD] website_form_project: Nothing to do

sergio-teruel:
cc @Tecnativa TT35813

Depends on:

- [ ] https://github.com/OCA/OpenUpgrade/pull/3454


pedrobaeza:
It was already included in #3454, but the final PR with the proper documentation is #3692, so I close this one.

PR #3656, opened 2022-12-15
[OU-FIX] calendar: Fix possible invalid recursive settings

ypapouin:
When coming from a very old odoo/openerp version, some calendar recursive events might have invalid settings (no `end_type`, or `count/interval <= 0`.

Having one of these invalid values mislead the `_rrule_parse` return and implies an infinite loop in `_get_ranges`. So `create_recurrent_events` is stuck and will never end.

### DIFF
--- a/openupgrade_scripts/scripts/calendar/14.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/calendar/14.0.1.0/pre-migration.py
@@ -3,6 +3,39 @@
 from openupgradelib import openupgrade
 
 
+def fix_empty_end_type(env):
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
In v14, these fields are not stored anymore, so I don't understand your problem. Maybe this is interesting to do in previous versions, but not in v14 as far I understand 🤔

ypapouin:
They are copied to the new `calendar.recurrence` model in the post migration script: `fill_calendar_recurrence_table`.

MiquelRForgeFlow (approved):

PR #3658, opened 2022-12-16
[14.0][OU-FIX] hr_timesheet: Delete obsolete view

ypapouin:
In previous `hr_timesheet_analysis` module version, a pivot view was already defined for the `hr_timesheet.timesheet_action_all` action.

In 14.0, the `hr_timesheet` module introduced its own pivot view `timesheet_action_view_all_pivot` that collides with the deprecated one from the OCA module and raise a constraint exception:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "act_window_view_unique_mode_per_action"
DETAIL:  Key (act_window_id, view_mode)=(xxx, pivot) already exists.
```

### DIFF
--- a/openupgrade_scripts/scripts/hr_timesheet/14.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_timesheet/14.0.1.0/pre-migration.py
@@ -29,3 +29,8 @@ def migrate(env, version):
         True,
     )
     _fill_task_overtime(env)
+    # Remove obsolete view from "hr_timesheet_analysis" module that would
+    # collide with the new "timesheet_action_view_all_pivot" view
+    openupgrade.delete_records_safely_by_xml_id(
+        env, ["hr_timesheet_analysis.act_hr_timesheet_line_view_all_pivot"]
+    )


pedrobaeza:
Better to rename the view then, to auto-document the disappearing of such module. And shouldn't you put it on `apriori.py` as merged module?

ypapouin:
No the OCA module `hr_timesheet_analysis` still exists, this is only this `ir.actions.act_window.view` record that has been deleted since it is now built-in the `hr_timesheet` module.

I'm using unlink because without this collision, the view would have been deleted during the `hr_timesheet_analysis` update.

pedrobaeza:
I don't get it then. If if exists on 14.0, when you do the regular update, the view will be removed as the XML-ID doesn't exist anymore.

ypapouin:
But the `hr_timesheet` is migrated before `hr_timesheet_analysis` since the last one depends on the first one

MiquelRForgeFlow (approved):

pedrobaeza (approved):
OK, it's due to such constraint...

openupgrade_scripts/scripts/hr_timesheet/14.0.1.0/pre-migration.py
```diff
@@ -29,3 +29,9 @@ def migrate(env, version):
         True,
     )
     _fill_task_overtime(env)
+
```
pedrobaeza: Wait, please don't add empty lines inside a method.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #3678, opened 2023-01-10
[15.0][ADD] merge account_statement_import_base into account_statement_import

hbrunn:
the `_base` module has been dropped during the migration


legalsylvain (approved):

openupgrade_scripts/apriori.py
```diff
@@ -35,12 +35,16 @@
...
+    # OCA/bank-statement-import
+    "account_statement_import_base": "account_statement_import",
     # OCA/hr
     "hr_recruitment_notification": "hr_recruitment",
     # OCA/hr-attendance
     "hr_attendance_user_list": "hr_attendance",
     # OCA/product-attribute
     "stock_account_product_cost_security": "product_cost_security",
+    # OCA/server-ux
+    "mass_operation_abstract": "mass_editing",
```
MiquelRForgeFlow: not sure about this. please, could you give us some kind of proof?

hbrunn: we have the dependency in v13: https://github.com/OCA/server-ux/blob/13.0/mass_editing/__manifest__.py#L15 but it's gone in v14: https://github.com/OCA/server-ux/blob/14.0/mass_editing/__manifest__.py#L15, while the module still exists.
It doesn't exist in v15 any more, that's why I added the merge there. Or we treat that as an abandoned module, then it will have to be cleaned up with database_cleanup

MiquelRForgeFlow: I see.

Selection_717

My opinion then is not to merge and treat it as an abandoned module.

hbrunn: agreed, removed that part

MiquelRForgeFlow (changes requested):
It seems this must be pumped to v16.

hbrunn:
thanks, indeed

hbrunn:
...as you added that, should we do s merge at all? Won't we use it for every version but it just doesn't happen to be migrated?

MiquelRForgeFlow:
I would says, let's wait and see what happens finally in OCA/bank-statement-import v16, as there are several not yet merged PRs...

PR #3680, opened 2023-01-10
[16.0][MIG] openupgrade_framework, base v16

hbrunn:
this is preliminary because I didn't yet go through all monkey patches to verify we still need them, but this lets the base migration run without crashing

TODO:
- [x] verify we still need all monkey patches and no new ones
- [x] re-add the testing framework for migration scripts


openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
```diff
@@ -0,0 +1,45 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "base", "16.0.1.3/noupdate_changes.xml")
+
+    if openupgrade.table_exists(
+        env.cr, "ir_translation"
+    ) and not openupgrade.table_exists(env.cr, "_ir_translation"):
+        openupgrade.rename_tables(env.cr, [("ir_translation", "_ir_translation")])
+
+    openupgrade_160.migrate_translations_to_jsonb(
```
StefanRijnhart: Rather than on a module by module basis, can we do this here for all translatable fields in the database? I see the lib method uses the models from the environment, so that would have to be addressed.

StefanRijnhart:
Please copy any preexisting `company_registry` values on `res.company` records to `res.partner`.

openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
hbrunn: this would mean we have to convert columns to jsonb manually, and we'd also not be able to use https://github.com/OCA/OCB/blob/16.0/odoo/tools/translate.py#L1551 because this also expects the registry to be loaded for the field in question.

Or do you mean to move this to end-migration and do it in bulk there?

pedrobaeza: The second one as we both talked in OCA days :wink:

StefanRijnhart: Well, if it was the end-migration the process would not be applied to modules that are not loaded at that moment, such as when a setup is migrated from 15 to 17 and some modules are not provided for 16.0. So I would prefer here using our own SQL logic (also for performance).

StefanRijnhart: I mean, you can still do it in the end-migration so that you can benefit from the models that *are* in the environment to report any non-standard table names, but I would prefer to have models that are not loaded at that point covered as well, as much as possible.

pedrobaeza: OK, I see, but I'm not sure if we can deduce which SQL columns to convert.

StefanRijnhart: I think we can go by the value in the `translate` column in `ir_model_fields`.

StefanRijnhart: Oh, I must admit that an exception are the fields with a `callable` translate value. This is not available for fields that are not loaded in the environment (at the end stage).
Fortunately, this is not entirely ambiguous because those fields' database translations have type `model_terms` rather than `model`.

hbrunn:
I've added the testing part in https://github.com/OCA/OpenUpgrade/pull/3680/commits/3ea5741f512575782df1ca2558f5452cfe2f588f, but the [test for base I added](https://github.com/OCA/OpenUpgrade/pull/3680/commits/3ea5741f512575782df1ca2558f5452cfe2f588f#diff-08e429ab7266e360a22d3570870434ee1a74fa2c06f1bedacd5f455e67796454R7) will fail on CI because because the data isn't set anywhere.

In older times we had the tests load a bunch of yaml files via `--test-file`, but that's only python by now and then also unittests only.

What do you think of just executing all sql in `openupgrade_scripts/scripts/$module/$version/tests/*.sql` before `$module`'s pre-migration script is run?

In the case of the base migration, that would be something like

```sql
--- set company_registry in v15 to be sure we migrate it correctly to v16
update res_company c set company_registry='424242' from ir_model_data d where d.name='main_company' and c.id=d.res_id
```

pedrobaeza:
OK, seems a good option.

openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
hbrunn: I've implemented that in https://github.com/OCA/OpenUpgrade/pull/3680/commits/25830ead71b0183c91c7e486ad9295bcca95a6b2

hbrunn:
test data now happens in https://github.com/OCA/OpenUpgrade/pull/3680/commits/fdfb642b31a4f10e242afb277c347196a356525e

StefanRijnhart:
Thanks for reimplementing tests! On the one hand, this is a really lean and clean solution, but on the other hand, if we would be able to create data through one or more Odoo shell scripts, that might make it easier for contributors to create (more complex) tests. That does imply installing the previous version of Odoo first in the Github workflow first (like we had in the past) to run the shell scripts on the test database. How does everybody feel about that?

openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
StefanRijnhart: Brilliant, thanks! Shall I recreate a new version of the template database and install an additional language in it? We can then add a test :heart_eyes:
Do you plan a special treatment in the end-migration for fields that have a callable value for `translate` (once the field is loaded in the environment)?

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,92 @@
...
+        "WHERE r.view_id=v.id AND v.inherit_id IS NOT NULL AND v.mode != 'primary'"
+    )
+
+    # update all translatable fields
+    cr.execute(
+        "SELECT f.name, m.model FROM ir_model_fields f "
+        "JOIN ir_model m ON f.model_id=m.id WHERE f.translate"
+    )
+    for field, model in cr.fetchall():
+        table = model.replace(".", "_")
```
StefanRijnhart: I think we will end up here with a hard coded list of non-standard table names here to cover those models as well (if only to get rid of the warnings). Maybe you can already introduce it at this point and populate it with the known ir.actions* mapping.

hbrunn:
So we do something like

```shell
for f in $script_path/*/*/tests/data*.py; do
    odoo-bin(N-1) shell < $f
```

in the test script?

And do we want to keep SQL as an alternative or do everything this way?

StefanRijnhart:
Yes, that's it! And keeping the SQL method apart from that would be redundant I feel.

openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
hbrunn: I actually didn't think about translate=callable yet, I think that can be a later problem.

The DB with an additional language would be very welcome indeed.

StefanRijnhart:
@hbrunn new database is now uploaded with the French language installed.
```
15.0=> select count(*), lang from ir_translation group by lang;
 count | lang  
-------+-------
 56087 | fr_FR
(1 row)
```
I reran tests and they are still green! Would you like to add another test specifically to test one or two translations?

hbrunn:
I would, but probably won't this week. Thank you for this, I clicked 'rerun all jobs'

StefanRijnhart:
No problem. Would you also be so kind as to add an annotated analysis_work file with a TODO for the fields that have a callable `translate` attribute?

pedrobaeza:
I think #3706 can be also included here, as the v15 DB may be populated before this Odoo commit was present.

TadeuszKarpinski:
How can I start tests locally? Init 15.0 db with data, then migrate and run tests?

hbrunn:
@TadeuszKarpinski it's much too early to run any user tests, this PR is the bare necessities to start writing migration scripts

hbrunn:
@StefanRijnhart added a test for french and the _work file, @pedrobaeza added your commit

openupgrade_framework/README.rst
```diff
@@ -14,13 +14,16 @@ Openupgrade Framework
...
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2FOpenUpgrade-lightgray.png?logo=github
-    :target: https://github.com/OCA/OpenUpgrade/tree/15.0/openupgrade_framework
+    :target: https://github.com/OCA/OpenUpgrade/tree/16.0/openupgrade_framework
     :alt: OCA/OpenUpgrade
 .. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
-    :target: https://translation.odoo-community.org/projects/OpenUpgrade-15-0/OpenUpgrade-15-0-openupgrade_framework
+    :target: https://translation.odoo-community.org/projects/OpenUpgrade-16-0/OpenUpgrade-16-0-openupgrade_framework
     :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
```
pedrobaeza: There's no such runboat, and it will be misleading anyway thinking that users can try the migration.

docsource/modules150-160.rst
```diff
@@ -66,7 +66,7 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | barcodes_gs1_nomenclature                       |                      |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| base                                            |                      |                                                 |
+| base                                            |                      |Partial                                          |
```
pedrobaeza: The third column is for indicating if no DB layout changes. It should be filled in the second column, and `Partial` is not recognized for the modules to test according the "parser" integrated in the GH action. They are `Done` or `Nothing to do`.

You can put `Done. Partial` though.

openupgrade_scripts/scripts/base/16.0.1.3/post-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Hunki Enterprises BV (https://hunki-enterprises.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "base", "16.0.1.3/noupdate_changes.xml")
+
```
pedrobaeza: Please don't put empty lines inside a method.

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,114 @@
...
+
+from openupgradelib import openupgrade
+from psycopg2.extensions import AsIs
+
+from odoo import tools
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from odoo.addons.openupgrade_scripts.apriori import merged_modules, renamed_modules
```
pedrobaeza: Why not letting it failed directly? If not, people don't read...

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,114 @@
...
+    odoo/modules/registry.py will break on the 'base' module not yet having
+    been instantiated.
+    """
+    if "openupgrade_framework" not in tools.config["server_wide_modules"]:
+        _logger.error(
+            "openupgrade_framework is not preloaded. You are highly "
+            "recommended to run the Odoo with --load=openupgrade_framework "
+            "when migrating your database."
+        )
+
```
pedrobaeza: Please avoid empty lines inside a method. It's recommended in PEP8 and visually you already have the comment lines for separating blocks.

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
hbrunn: I just took that over from the previous version, remove then?

hbrunn:
coverage would be nice though, but that can be another PR

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
pedrobaeza: I think it's better to crash if you have this situation than having a weird error and a crash later when running the scripts.

pedrobaeza:
And please squash commits for merging this.

hbrunn:
rebased

pedrobaeza (approved):

StefanRijnhart:
Thanks for the updates, and great work overall!

PR #3683, opened 2023-01-13
[15.0][OU-ADD] project_timeline_task_dependency: Merged with project_timeline

CarlosRoca13:
As said in this comment https://github.com/OCA/project/pull/942#discussion_r935782545 this module is deprecated in v15 as the feature is added at project_timeline level

cc @Tecnativa TT40300

ping @chienandalu @pedrobaeza

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -49,6 +49,7 @@ merged_modules = {
     # OCA/project
     "project_mail_chatter": "project",
     "project_task_dependency": "project",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This should be `[OU-ADD]`, not `[MIG]`. You are not migrating the module, but adding a new OpenUpgrade specification. And also rephrase the part where it's saying `project_timesheet` instead of  `project_timeline`.

pedrobaeza (approved):

PR #3684, opened 2023-01-13
[13.0][OU-FIX] account: rename xml id of account.payment.term

HekkiMelody:
Seems to me that these two records are the same record, so trying to rename them instead of trying to delete one and creating a duplicate.

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -1100,7 +1100,6 @@ def migrate(env, version):
             "account.invoice_comp_rule",
             "account.voucher_comp_rule",
             "account.voucher_line_comp_rule",
-            "account.account_payment_term_net",
             "account.reconciliation_model_default_rule",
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Hi. 

The settings is the same, and the replacement is done in the same commit. :+1: for me.
https://github.com/odoo/odoo/commit/d4710d92ddbc37f92e8d9685d749a971c47b70b4#diff-b5bf1196ef534878acf18cb53a67973b6aee1b89afacbf0cd559cd7c7523cbafL54

CC : @alexis-via could you take a look on this one ?

pedrobaeza:
Yes, the change is OK. Please rename the commit:

- It shouldn't include the version. That's just for the PR title, not the commit message.
- Put as prefix `[OU-FIX]`, to distinguish that this is something of OpenUpgrade, and not a regular fix for the Odoo code.

HekkiMelody:
Done. Thank you for the pointers.

PR #3686, opened 2023-01-13
[15.0][OU-ADD] sale_project: Migration scripts

pedrobaeza:
@Tecnativa TT35781


remi-filament:
Hi @pedrobaeza , I see you modify sale_purchase here instead of sale_project, am I missing something or is it a mistake ?

pedrobaeza:
Yes, it has been a mistake. Moving files in a direct commit.

pedrobaeza:
Fixed in 7fe09768b1d6669a3b2e4527747aac9c98f25c68. Thanks to notify it.

PR #3689, opened 2023-01-13
[15.0][OU-ADD] hr_expense: Migration scripts

pedrobaeza:
@Tecnativa TT35728


pedrobaeza:
Added the load of noupdate_changes.xml file.

PR #3697, opened 2023-01-18
[14.0][OU-FIX] base: Assign parent module categories

pedrobaeza:
As these records are created as noupdate=1, we need to manually assign the parent categories that are new in v14.

@Tecnativa TT41258

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
@@ -44,6 +44,56 @@ def fix_module_category_parent_id(env):
     )
 
 
+def assign_module_category_parent(env):
+    """As these records are created as noupdate=1, we need to manually assign the
+    parent categories that are new in v14.
+    """
+    category_mapping = [
+        ("module_category_sales_sales", "module_category_sales"),
+        ("module_category_sales_point_of_sale", "module_category_sales"),
+        ("module_category_services_project", "module_category_services"),
+        ("module_category_services_timesheets", "module_category_services"),
+        ("module_category_accounting_accounting", "module_category_accounting"),
+        ("module_category_inventory_purchase", "module_category_inventory"),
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
Uff... 🤔 . Just remember that https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py#L7 also exists.

pedrobaeza:
Yes, and that doesn't serve for anything in this case, as they don't have parent. That SQL is for "incorrect parents", not "no parent at all".

legalsylvain (commented):
Hi. Sorry if my question is not relevant but I don't understand your PR.

AFAIK, for exemple for "base.module_category_sales_point_of_sale"

- both are created in the module "point_of_sale" (https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/security/point_of_sale_security.xml#L1 and https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/security/point_of_sale_security.xml#L1)
- Both are not defined as noupdate=1
- there is no parent_id value in the xml file in V14.

pedrobaeza:
@legalsylvain it's due to a hack that is done on module loading:

https://github.com/odoo/odoo/blob/27554204c3786ac4846972c87ce5aea6a923eb4e/odoo/modules/db.py#L132

At that level, when installing the module `point_of_sale`, the module category is already present in the DB, with noupdate=1. That extra XML serves on install for updating both fields, but the rest is already there (it's a special behavior on XML data for noupdate=1 on install mode).

MiquelRForgeFlow (approved):

legalsylvain (approved):
Thanks for the explanation @pedrobaeza.

pedrobaeza:
Actually, we should consider adding to noupdate_data.xml all the `ir.module.category` XML records we find across Odoo, as that information won't be updated (as we do update on migration process, not install).

PR #3699, opened 2023-01-21
[15.0][ADD] hr_expense: disable reconcile and fiscalyear check during migration

hbrunn:
this avoids the migration crashing when recomputing moves that are already reconciled or before the lock date

### DIFF
--- a/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
@@ -14,7 +14,18 @@ def _fill_payment_state(env):
     # v14 these ones were not computed being of type `entry`, which changes now
     # on v15 if the method `_payment_state_matters` returns True, which is the
     # case for the expense moves
+    # Disable the reconciliation check to be able to update reconciled moves
+    _check_reconciliation = env["account.move.line"].__class__._check_reconciliation
+    _check_fiscalyear_lock_date = env[
+        "account.move"
+    ].__class__._check_fiscalyear_lock_date
+    env["account.move.line"].__class__._check_reconciliation = lambda self: None
+    env["account.move"].__class__._check_fiscalyear_lock_date = lambda self: None
     env["hr.expense.sheet"].search([]).account_move_id._compute_amount()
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks. Lgtm.

Note : A new openupgrade decorator could make the code lighter. Something like : 


    @openupgrade.disable([
        env["account.move.line"].__class__._check_reconciliation,
        env["account.move"].__class__._check_fiscalyear_lock_date
    ])


Just an idea, not a blocking point.

hbrunn:
@legalsylvain I agree, but I'd probably make this a context manager instead of a decorator - will add with the next migration script that needs this

pedrobaeza:
I don't get why recomputing the amount touches reconciliation. If so, we should find another technique instead of just disabling checks IMO.

legalsylvain:
> I don't get why recomputing the amount touches reconciliation.

Indeed.
- _check_fiscalyear_lock_date is called when state changed or name / date changed once posted (that should not occures when calling _compute_amount)
ref : https://github.com/odoo/odoo/blob/46708fa43e0c878927ac458f15d6192bf0af9c91/addons/account/models/account_move.py#L2351
https://github.com/odoo/odoo/blob/46708fa43e0c878927ac458f15d6192bf0af9c91/addons/account/models/account_move.py#L2359

pedrobaeza:
But both triggers are on state or date change, not recomputing the amount...

hbrunn:
I'll debug through this next time I work on the migration and come back with the results

sholtode (approved):

hbrunn:
okay, so I've debugged. I've a reconciled move with a line ending up here:

https://github.com/OCA/OCB/blob/15.0/addons/account/models/account_move.py#L1511-L1513

so in

https://github.com/OCA/OCB/blob/15.0/addons/account/models/account_move.py#L1526

we have `total_currency == total == 0`

and this write then breaks in

https://github.com/OCA/OCB/blob/15.0/addons/account/models/account_move.py#L4777

So I think we need this.

pedrobaeza (approved):
Thanks for the explanation.

pedrobaeza:
Next time please use `[OU-FIX]`, or `[OU-IMP]`. This is not adding a new module migration script.

PR #3713, opened 2023-02-08
[13.0][FIX] account: Transfer invoice line description to matched journal items

pedrobaeza:
Previous versions of Odoo strip invoice line description to 64 characters when generating the corresponding `account.move.line`, so we need to transfer the invoice line description when matched with the corresponding journal item.

@Tecnativa TT41609


MiquelRForgeFlow (approved):

pedrobaeza:
Bonus: run this SQL for fixing your already migrated DBs:

```sql
UPDATE account_move_line aml
SET name = ail.name FROM account_invoice_line ail
WHERE aml.old_invoice_line_id = ail.id AND ail.name != aml.name
    AND substring(ail.name from 1 for 64) = aml.name;
```

PR #3716, opened 2023-02-10
[13.0][OU-IMP] l10n_es: Handle food taxes renaming

pedrobaeza:
We created the tax mapping for <=13.0 before Odoo reacts, assigning several XML-IDs. Later, Odoo created another PR adding such taxes with different XML-IDs.

We have homogenized such XML-IDs till 13.0, but on previous versions, on `l10n_es_extra_data` (already merged into `l10n_es` in apriori), old XML-IDs were preserved.

With this addition, we rename both tax templates and company taxes for avoiding an error on duplicated tax names + incorrect tax mappings.

@Tecnativa


MiquelRForgeFlow (approved):
LGTM

PR #3728, opened 2023-02-23
[15.0][OU-ADD] website_legal_page: Terms page added on account

CarlosRoca13:
cc @Tecnativa TT41844

ping @pedrobaeza @chienandalu

### DIFF
--- a/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
@@ -52,6 +52,42 @@ def set_res_company_account_setup_taxes_state_done(env):
         company.account_setup_taxes_state = "done"
 
 
+def _handle_website_legal_page(env):
+    """As the terms and conditions page has been added in the 'account' module,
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
```diff
@@ -52,6 +52,40 @@ def set_res_company_account_setup_taxes_state_done(env):
...
+                JOIN website w ON w.id = iuv.website_id
+            WHERE iuv.key = 'website_legal_page.legal_page' AND w.company_id = %s
+            """,
+            (company.id,),
+        )
+        row = env.cr.fetchone()
+        if row:
+            view = env["ir.ui.view"].browse(row[0])
+            view_temp = view.copy(
+                {"arch_db": view.arch.replace('t-call="website.layout"', "")}
```
chienandalu, line 80: Add a comment on the trick we're doing here so we can insert it into the field without issues.

pedrobaeza (approved):

CarlosRoca13:
Done @chienandalu

PR #3735, opened 2023-02-25
[11.0][IMP] Switch to GH actions

pedrobaeza:
Alternative attempt to greenify CI on this old branch.

No support to OU tests yet

@Tecnativa


pedrobaeza:
Tests are still being run, but without loading demo data, and that's why they fail. I'm giving up for now, as I got Travis build correct adjusting the configuration in #3734, but if someone wants to retake, feel free.

PR #3758, opened 2023-03-13
[15.0][FIX] sale_coupon: handle noupdate changes in template

MiquelRForgeFlow:
I generated manually this noupdate_changes.xml.


pedrobaeza (approved):
Good catch!

PR #3766, opened 2023-03-14
[9.0][FIX] hr: delete safely constraint

MiquelRForgeFlow:
Constraint removed in https://github.com/odoo/odoo/commit/b5b2718e76cbd0dd79870eafa4a2e2334bd5270e (odoo 9.0).

The field `no_of_hired_employee` always existed in `hr` module. But Odoo 15.0 added a compute for this field in https://github.com/odoo/odoo/commit/b4ba7f12ae367ef3ba99523ba8fcfb6d1d7321a1. So, for databases coming from 9.0 (and still have that constraint) it breaks during the compute.


pedrobaeza (approved):
Thanks!

Issue #3681, opened 2023-01-10
Migration to version 16.0


legalsylvain:
- Full list of modules available here : https://github.com/OCA/OpenUpgrade/blob/16.0/docsource/modules150-160.rst 

# Pending / done work

- [x] account - By @ngochung207 / @pedrobaeza - #4070
- [x] account_check_printing - By @CarlosRoca13 - #4909
- [x] account_debit_note - By @pedrobaeza - #4404
- [x] account_edi - By @remi-filament - #4300
- [x] account_edi_ubl - By @remi-filament - #4158
- [x] account_edi_ubl_cii - By @remi-filament - #4158
- [x] account_fleet - By @pedrobaeza - #4504
- [x] account_payment - By @remytms - #3920
- [x] account_qr_code_sepa - By @robinkeunen - #4337
- [x] account_sale_timesheet - By @MiquelRForgeFlow - #4721
- [x] analytic - By @duong77476-viindoo / @pedrobaeza - #4188
- [x] association - By @stefan-tecnativa - #4175
- [x] attachment_indexation - By @pedrobaeza - #4189
- [x] auth_ldap - By @marielejeune - #3776
- [x] auth_oauth - By @locdang8 - #4013
- [x] auth_password_policy - By @locdang8 - #3969
- [x] auth_password_policy_portal - By @locdang8 - #4000
- [x] auth_password_policy_signup - By @locdang8 - #4025
- [x] auth_signup - By @marielejeune - #3778
- [x] auth_totp - By @locdang8 - #4006
- [x] auth_totp_mail - By @pedrobaeza - #4365
- [x] auth_totp_portal - By @robinkeunen - #4338
- [x] barcode_gs1_nomenclature - By @locdang8 - #3934
- [x] barcodes - By @nguyenvietlam0640 - #3930
- [x] base - By @hbrunn - #3680
- [x] base_address_extended - By @marielejeune - #3769
- [x] base_automation - By @marielejeune - #3779
- [x] base_geolocalize - By @locdang8 - #4039
- [x] base_iban - By @remytms - #3870
- [x] base_import - By @marielejeune - #3772
- [x] base_setup - By @remytms - #3869
- [x] base_sparse_field - By @marielejeune - #3770
- [x] base_vat - By @remytms - #3871
- [x] board - By @remytms - #3900
- [x] bus - By @marielejeune - #3773
- [x] calendar - By @remytms - #3899
- [x] calendar_sms - By @SonCrits - #4034
[...]

PR #3769, opened 2023-03-14
[16.0][MIG] base_address_extended

marielejeune:
It seems that `base_address_city` was merged into `base_address_extended`. Is there something else to do than renaming the xmlids?
Edit: the modules will be merged with pre-migration script of `base`, as they are listed in `apriori.py`


marielejeune:
https://github.com/OCA/OpenUpgrade/issues/3681

pedrobaeza (changes requested):
You don't have to rename anything, as it's already handled in the call of `update_module_names` openupgradelib method in `base` module.

docsource/modules150-160.rst
```diff
@@ -70,7 +70,7 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | |del| base_address_city                         |                      |                                                 |
```
pedrobaeza, line 71:
```suggestion
| |del| base_address_city                         | Done                 | Merged into base_address_extended            |
```

As this has been already tackled in https://github.com/OCA/OpenUpgrade/blob/2d7f8de7289fdf36e0ac48f41ee52b6138190399/openupgrade_scripts/apriori.py#LL34C1-L34C1

openupgrade_scripts/scripts/base_address_extended/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,36 @@
...
+---Fields in module 'base_address_extended'---
+base_address_extended / res.city                 / __last_update (datetime)      : previously in module base_address_city
+base_address_extended / res.city                 / _order                        : previously in module base_address_city
+base_address_extended / res.city                 / country_id (many2one)         : previously in module base_address_city
+base_address_extended / res.city                 / display_name (char)           : previously in module base_address_city
+base_address_extended / res.city                 / name (char)                   : previously in module base_address_city
+base_address_extended / res.city                 / state_id (many2one)           : previously in module base_address_city
+base_address_extended / res.city                 / zipcode (char)                : previously in module base_address_city
+base_address_extended / res.country              / enforce_cities (boolean)      : previously in module base_address_city
+base_address_extended / res.country              / street_format (text)          : DEL required
```
pedrobaeza, line 14: Where is this now handled?

marielejeune: It seems that this behavior disappeared. See Fabien's commit description here:

https://github.com/odoo/odoo/commit/5b72f017662525362ff1e1bc8173a7ab395dbcf9

pedrobaeza: Reading the commit message (that you can annotate in this work file for reference), it seems there's now a `address_view_id` that reflects such format, so it's not disappearing at all. I don't think anybody has created custom countries, and I suppose they are provided correctly with the proper `address_view_id`, so it can be nothing to do, but please check for being sure, and reflect all of this in the analysis work file for not having to wonder this in the future.

marielejeune: All countries have an empty `address_view_id` by default.
So when Odoo deletes its field `street_format`, we are loosing the information they provided in previous versions.
Is there a way to prevent that? We won't create a custom view for all countries data?

hbrunn: I don't think `street_format` has anything to do with `address_view_id` - the former is about how we go from street_name, street_number etc to street, while the latter is about user input/display, and the whole address instead of just a street name anyways.

https://github.com/OCA/OCB/blob/16.0/addons/base_address_extended/models/res_partner.py#L22 indicates to me this is just dropped functionality as @marielejeune said. ~~I'd suggest to rename the column to the legacy name for some possible OCA module to re-add this functionality.~~

pedrobaeza: Don't rename it then. The rename should be done just in case of conflicting name. If not, it will require a pre-migration script to rename back the column instead of just installing the replacement module.

hbrunn (approved):

pedrobaeza (approved):

PR #3774, opened 2023-03-15
[16.0][OU-ADD] utm

marielejeune:


openupgrade_scripts/scripts/utm/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,15 @@
+---Models in module 'utm'---
+new model utm.source.mixin [abstract]
+# NOTHING TO DO
+
+---Fields in module 'utm'---
+utm          / utm.campaign             / title (char)                  : NEW required
+# NOTHING TO DO
```
pedrobaeza: You have to fill this field with a value. You can copy the name one as best-compromise solution.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/utm/16.0.1.1/upgrade_analysis_work.txt
marielejeune: Right, done

openupgrade_scripts/scripts/utm/16.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(env.cr, """ALTER TABLE utm_campaign ADD title jsonb;""")
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE utm_campaign
+        SET title = jsonb_build_object('en_US', to_jsonb(name));
```
pedrobaeza: This is losing possible translations. Isn't valid the method you provided in openupgradelib? https://github.com/OCA/openupgradelib/pull/306

marielejeune: You're right, I missed translations.
The method I provided is useless here because translations are already migrated in OpenUpgrade pre-migration script of base module.
Hence when starting migrating utm module, `name` column is a jsonb column containing all the translations. Hence it suffices to copy this column to populate the new `title` column.

pedrobaeza (approved):

PR #3778, opened 2023-03-15
[16.0][OU-ADD] auth_signup: Nothing to do

marielejeune:


remytms (changes requested):
I think that the `noupdate_changes.xml` contains changes in some `mail.template` that should be loaded.

You can check the documentation here: https://github.com/OCA/OpenUpgrade/pull/3864/files

It helped me a lot. :) Also there is know bug with the method `delete_record_translations()` as Pedro spotted it here: https://github.com/OCA/OpenUpgrade/pull/3899#discussion_r1215541893

remytms:
I think that this is not blocked by dependency anymore. `base_setup`, `mail` and `web` has been merged. :)

pedrobaeza (approved):

pedrobaeza:
Ouch, I didn't see the other comment about the noupdate_changes... Please add such migration script.

marielejeune:
Sorry for the delay. See https://github.com/OCA/OpenUpgrade/pull/3917

PR #3782, opened 2023-03-15
[16.0][OU-ADD] portal

marielejeune:


rvalyi (approved):

legalsylvain (commented):
Hi @marielejeune. Thanks for your contribution. 

I see that there is a noupdate_changes.xml file for that module, with a mail template. I guess that we should write a post-migration script to refresh the data. don't you think ? 

See according new documentation : https://oca.github.io/OpenUpgrade/use_cases/noupdate_xml_entry_changed.html#write-migration-script

legalsylvain (changes requested):
Hi @marielejeune. Thanks for your contribution. 

I see that there is a noupdate_changes.xml file for that module, with a mail template. I guess that we should write a post-migration script to refresh the data. don't you think ? 

See according new documentation : https://oca.github.io/OpenUpgrade/use_cases/noupdate_xml_entry_changed.html#write-migration-script

marielejeune:
> I see that there is a noupdate_changes.xml file for that module, with a mail template. I guess that we should write a post-migration script to refresh the data. don't you think ?

Right, done

legalsylvain (approved):
/ocabot merge nobump

pedrobaeza:
@legalsylvain this one is still missing to delete the record translations.

legalsylvain:
> @legalsylvain this one is still missing to delete the record translations.

Hum. That's annoying. I'd like to integrate the deletion of the translation in load_data, if version >= 16.0 this should avoid that noicy second line, that is mandatory in most case. What do you think ?

pedrobaeza:
I'm afraid that's not so easy to be done and will require to patch Odoo itself, but if got, it would be great of course. But meanwhile, we need to complete the scripts with current system.

legalsylvain:
> I'm afraid that's not so easy to be done and will require to patch Odoo itself, but if got, it would be great of course. 

Yes, I thought to this point, and it doesn't seems easy indeed.

> But meanwhile, we need to complete the scripts with current system.

@marielejeune could you provide a PR ?

marielejeune:
@legalsylvain Done here: https://github.com/OCA/OpenUpgrade/pull/4028

PR #3784, opened 2023-03-15
[16.0][OU-ADD] resource: Nothing to do

marielejeune:


pedrobaeza (changes requested):
Don't remove anything. Let that job to `database_cleanup` if you want, but one of the OpenUpgrade principles is to be the most conservative possible. This also favors if doing a multi-version migration (example, from 15 to 18), to reuse an existing table if needed (because the feature is reintroduced).

pedrobaeza (approved):

PR #3786, opened 2023-03-16
[15.0][OU-ADD] Use website_form_require_legal instead of website_crm_privacy_policy

CarlosRoca13:
The second one is deprecated on v15.0

cc @Tecnativa TT41895

ping @chienandalu @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py
@@ -1,0 +1,55 @@
+from lxml import etree
+from openupgradelib import openupgrade
+
+
+def deprecate_website_crm_privacy_policy(env):
+    module_website_crm_privacy_policy = env["ir.module.module"].search(
+        [("name", "=", "website_crm_privacy_policy")]
+    )
+    if module_website_crm_privacy_policy.state == "installed":
+        module_website_form_require_legal = env["ir.module.module"].search(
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py
```diff
@@ -0,0 +1,63 @@
+from lxml import etree
+from openupgradelib import openupgrade
+
+
+def deprecate_website_crm_privacy_policy(env):
+    if (
+        env["ir.module.module"]
+        .search(["name", "=", "website_crm_privacy_policy"])
```
pedrobaeza: Careful with the syntax:

```suggestion
        .search([("name", "=", "website_crm_privacy_policy")])
```

openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py
```diff
@@ -0,0 +1,63 @@
...
+        .state
+        == "installed"
+    ):
+        openupgrade.logged_query(
```
pedrobaeza: Don't do this with SQL, but ORM in this case, and using the recently searched record (+ search the other).

pedrobaeza (changes requested):

CarlosRoca13:
Changes done

pedrobaeza (approved):

PR #3790, opened 2023-03-17
[13.0][FIX] point_of_sale: multicompany

chienandalu:
Avoid _check_company errors when copying picking types across the companies pos configs

cc @Tecnativa TT38570

please review @pedrobaeza

### DIFF
--- a/addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
+++ b/addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
@@ -108,13 +108,22 @@ def create_pos_picking_types(env):
         env["pos.config"].browse(config_ids),
         location_ids
     ):
-        picking_type = config.picking_type_id.copy({
-            "name": "[OU] {} - {}".format(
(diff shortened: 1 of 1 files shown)


addons/point_of_sale/migrations/13.0.1.0.1/post-migration.py
```diff
@@ -108,13 +108,19 @@ def create_pos_picking_types(env):
...
-            "name": "[OU] {} - {}".format(
-                config.picking_type_id.name, config.name
-            ),
+        vals = {
+            "name": f"[OU] {config.picking_type_id.name} - {config.name}",
             "company_id": config.company_id.id,
             "default_location_src_id": location_id,
-        })
+        }
+        if config.picking_type_id.sequence_id:
```
pedrobaeza: Do we need to save some sequences?

```suggestion
        if config.picking_type_id.sequence_id and config.picking_type_id.sequence_id != config.company_id:
```

MiquelRForgeFlow (approved):

PR #3794, opened 2023-03-21
[15.0][OU-ADD] website_sale_attribute_filter_price: Price filter added to website_sale

CarlosRoca13:
cc @Tecnativa TT37011

ping @pedrobaeza @chienandalu

### DIFF
--- a/openupgrade_scripts/scripts/website_sale/15.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/website_sale/15.0.1.1/pre-migration.py
@@ -1,0 +1,22 @@
+from openupgradelib import openupgrade
+
+
+def enable_price_filter_view(env):
+    for website in env["website"].search([]):
+        view = env["ir.ui.view"].search(
+            [
+                ("key", "=", "website_sale_attribute_filter_price.pricefilter"),
+                ("website_id", "=", website.id),
+            ]
+        )
+        if view:
+            # It's important to set the context to create the new view related to the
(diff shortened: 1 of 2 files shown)


chienandalu (approved):
Thanks! :+1:

pedrobaeza (approved):

openupgrade_scripts/scripts/website_sale/15.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,22 @@
...
+
+def enable_price_filter_view(env):
+    for website in env["website"].search([]):
+        view = env["ir.ui.view"].search(
+            [
+                ("key", "=", "website_sale_attribute_filter_price.pricefilter"),
+                ("website_id", "=", website.id),
+            ]
+        )
+        if view:
```
pedrobaeza, line 12: I'm thinking now... shouldn't you remove the old view?

CarlosRoca13: In my tests, when the old module is uninstalled, the old view disappears. But if, to be on the safe side, you want to deactivate it, that's fine for me.

pedrobaeza: But here we are not uninstalling it, but merging it with `website_sale`, so yes, please do another PR unlinking them.

PR #3803, opened 2023-03-27
16.0 update pre commit 27 03 2023

legalsylvain:
Update pre-commit files.
As here, we don't use copier, because of CI specificities, I meld the file with the one present in OCA/web/16.0

Note : it fixes current error on V16.0 (bad repo replaced by https://github.com/OCA/pylint-odoo)

Thanks for your review.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ exclude: |
   (LICENSE.*|COPYING.*)
 default_language_version:
   python: python3
-  node: "14.18.0"
+  node: "16.17.0"
 repos:
   - repo: local
     hooks:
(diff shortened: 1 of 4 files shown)


hbrunn (approved):

.pre-commit-config.yaml
```diff
@@ -106,31 +112,27 @@ repos:
...
   - repo: https://github.com/PyCQA/flake8
     rev: 3.9.2
     hooks:
       - id: flake8
-        name: flake8 except __init__.py
-        exclude: /__init__\.py$
-        additional_dependencies: ["flake8-bugbear==20.1.4"]
-      - id: flake8
-        name: flake8 only __init__.py
-        args: ["--extend-ignore=F401"] # ignore unused imports in __init__.py
```
pedrobaeza, line 118: I think this one is important, specially for the now included migration tests.

legalsylvain: Hi. Thanks for your review. 

1) Not sure to understand. Do you mean you want to keep : 
```
      - id: flake8
        name: flake8 except __init__.py
        exclude: /__init__\.py$
        additional_dependencies: ["flake8-bugbear==20.1.4"]
```

or 
```
      - id: flake8
        name: flake8 only __init__.py
        args: ["--extend-ignore=F401"] # ignore unused imports in __init__.py
        files: /__init__\.py$
        additional_dependencies: ["flake8-bugbear==20.1.4"]
```
or both ?

2) In other OCA repo, that line (--extend-ignore=F401) is not present. Are you sure it is required ? 
https://github.com/OCA/server-tools/blob/16.0/.pre-commit-config.yaml#L124

pedrobaeza (approved):
OK, let's go with this and let's see if there's a problem

PR #3808, opened 2023-03-31
[14.0][FIX] stock: recompute_stock_picking_scheduled_date

chienandalu:
We were performing the compute for every picking as many times as stock moves it had.

cc @Tecnativa TT38570

ping @pedrobaeza @carlosdauden

### DIFF
--- a/openupgrade_scripts/scripts/stock/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/stock/14.0.1.1/post-migration.py
@@ -24,7 +24,7 @@ def merge_priorities(env):
 def recompute_stock_picking_scheduled_date(env):
     env.cr.execute(
         """
-        SELECT sp.id
+        SELECT DISTINCT sp.id
         FROM stock_picking sp
         JOIN stock_move sm ON sm.picking_id = sp.id
         WHERE sp.state NOT IN ('done', 'cancel')"""


openupgrade_scripts/scripts/stock/14.0.1.1/post-migration.py
```diff
@@ -24,12 +24,12 @@ def merge_priorities(env):
...
     env.cr.execute(
         """
-        SELECT sp.id
+        SELECT DISTINCT sp.id
         FROM stock_picking sp
         JOIN stock_move sm ON sm.picking_id = sp.id
         WHERE sp.state NOT IN ('done', 'cancel')"""
     )
-    picking_ids = [pick[0] for pick in env.cr.fetchall()]
+    picking_ids = [pick for pick, *_ in env.cr.fetchall()]
```
MiquelRForgeFlow: why?

chienandalu: Well, it's just a matter of style. The result is the same

carlosdauden: Original approach is "more efficent" because only unpack needed column.

In this case it doesn't matter because there is only one column, but if it's just a question of style it's better to get used to styles that offer the best performance. :wink:

carlosdauden (changes requested):

openupgrade_scripts/scripts/stock/14.0.1.1/post-migration.py
chienandalu: Well, I have to test it and you're completely right :)

```python
>>> timeit(lambda: [pick for pick, *_ in pickings], number=1)
0.7465282640041551
>>> timeit(lambda: [pick[0] for pick in pickings], number=1)
0.12576593900303124
```

I'd still prefer it when performance isn't an issue as a more expressive way of unpacking.

Anyway, I revert it ;)

pedrobaeza (approved):

PR #3817, opened 2023-04-06
[15.0][FIX+IMP] calendar: Remove constraint before + no verbose HTML conversion

pedrobaeza:
- calendar_recurrence_month_day should be removed before doing any operation with such table, or we will have constraint errors.
- Don't hammer logs displaying each UPDATE operation on the calendar event HTML conversion.

@Tecnativa


pedrobaeza:
Tested.

PR #3818, opened 2023-04-08
[16.0][ADD] Apriori : 'stock_picking_backorder_strategy' merged into 'stock'.

legalsylvain:
**Note** : When migrating 'stock' module

**Rename** ``stock_picking_type field`` 'backorder_strategy' -> 'create_backorder';

**Map values** : {'manual': 'ask', 'no_create': 'never', 'cancel': never', 'create': 'always'}

Ref : OCA/15.0 module : https://github.com/OCA/stock-logistics-workflow/blob/15.0/stock_picking_backorder_strategy/models/stock_picking_type.py#L10
Ref : odoo/16.0 code : https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_picking.py#L91

CC: @pedrobaeza, @MiquelRForgeFlow

### DIFF
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
@@ -1,6 +1,25 @@
 from openupgradelib import openupgrade
 
+def _handle_stock_picking_backorder_strategy(env):
+    # Handle the merge of OCA/stock-logistics-workflow/stock_picking_backorder_strategy
+    # feature in odoo/stock V16 module.
+    if openupgrade.column_exists(
+        env.cr, "stock_picking_type", openupgrade.get_legacy_name("backorder_strategy")
+    ):
+        openupgrade.map_values(
+            env.cr,
+            openupgrade.get_legacy_name("backorder_strategy"),
+            "create_backorder",
+            [
+                ("manual", "ask"),
+                ("create", "always"),
+                ("no_create", "never"),
+                ("cancel", "never"),
+            ],
+            table="stock_picking_type",
+        )
 
 @openupgrade.migrate()
 def migrate(env, version):
+    _handle_stock_picking_backorder_strategy(env)
     openupgrade.load_data(env.cr, "stock", "16.0.1.1/noupdate_changes.xml")
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/pre-migration.py
@@ -111,6 +111,20 @@ def _update_stock_quant_package_pack_date(env):
         """,
     )
 
+def _handle_stock_picking_backorder_strategy(env):
+    # Handle the merge of OCA/stock-logistics-workflow/stock_picking_backorder_strategy
+    # feature in odoo/stock V16 module.
+    if openupgrade.column_exists(env.cr, "stock_picking_type", "backorder_strategy"):
+        # Rename column
+        openupgrade.rename_columns(
+            env.cr,
+            {
+                "stock_picking_type": [
+                    ("backorder_strategy", None),
+                ],
+            },
+        )
+
 
 @openupgrade.migrate()
 def migrate(env, version):
@@ -122,3 +136,4 @@ def migrate(env, version):
     _update_stock_quant_package_pack_date(env)
     _update_sol_product_category_name(env)
     _compute_stock_location_replenish_location(env)
+    _handle_stock_picking_backorder_strategy(env)


pedrobaeza:
The values mapping seems forgotten.

legalsylvain:
> The values mapping seems forgotten.

In the PR ? It is intended. I can not work on the ``stock`` migration module, because of missing dependencies. (and lake of time ;-)). So this text is just reminder when people will work on the ``stock`` module in the futur.

pedrobaeza:
Well, we can include such part just now, letting `stock` itself for later. It won't be executed if we don't write `Done` or `Nothing to do` in the rst file.

legalsylvain:
> Well, we can include such part just now, letting stock itself for later. It won't be executed if we don't write Done or Nothing to do in the rst file.

I see. It's done.
(I was a bit reluctant to propose untested code by the CI. But let's be optimistic! and if there is a problem, we can fix it when we migrate stock).

openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (C) 2023-Today GRAP (http://www.grap.coop)
+# @author Sylvain LE GAL (https://twitter.com/legalsylvain)
```
pedrobaeza: Please don't create 2 lines for the same autorship. Only one with both credits.

pedrobaeza (commented):
Actually, the code will be the same untested although `stock` module is installed, as the column won't exist and the part inside the if won't be executed. The only way to test it is thanks to our beloved @hbrunn's latest contribution in this version, allowing to [execute code before migration](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/tests/data.py>), and then to [test it](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/tests/test_base_migration.py).

openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
legalsylvain: No Sorry.
1) that's matter of taste, not on OCA rule.
2) that is the pattern used in the 170 GRAP modules under OCA umbrella, so changing all the headers requires time and doesn't provide any value.

regards.

legalsylvain:
> The only way to test it is thanks to our beloved @hbrunn's latest contribution in this version, allowing to [execute code before migration](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/tests/data.py), and then to [test it](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/tests/test_base_migration.py).

That look nice !

openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
pedrobaeza: I'm not caring right now about the other 170 modules. I'm caring about this file in OpenUpgrade, and nobody uses a copyright double line, so please change it at least here.

legalsylvain: > I'm not caring right now about the other 170 modules. 

If I do this for this PR, you will feel legitimate to ask me on my other PRs, about the other 170 contributions, and it will start to represent a significant amount of work time and noise for me, and unnecessary time (for me & the OCA).

> and nobody uses a copyright double line, so please change it at least here.

Except :
Acsone : https://github.com/OCA/mis-builder/blob/16.0/mis_builder/models/accounting_none.py#L1
Akretion : https://github.com/OCA/l10n-france/blob/16.0/l10n_fr_account_tax_unece/__manifest__.py#L1
coopiteasy : https://github.com/OCA/cooperative/blob/14.0/cooperator/models/account_journal.py#L1
etc.

Pedro, I greatly appreciate the quality of your remarks in your many reviews and I learn a lot from reading you. But from time to time I think you confuse personal preferences with OCA rules.
This is also the case when you ask contributors to squash their commits (which are not fixups), or to remove empty lines in a function. These are developers' preferences. Everyone has their own, and it doesn't matter if several ways are present in the OCA.
I sincerely think that the level for contributing to the OCA is already very high (pre-commit, CI, keep history when migrating, don't push translation through github but through weblate, etc.)
I don't think that discussing this kind of things brings anything positive. it's a bit of a nitpick. And at some level, this may cause some contributors to become discouraged.

pedrobaeza: I'm saying here in OpenUpgrade. Where do you find double line here in OU? 

And it's not nitpicking. It's for not hightlighting your contribution over others. Why do you deserve 2 lines meanwhile other contributors only put one?

And yes, it's already stated in OCA template: https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L1

legalsylvain: > I'm saying here in OpenUpgrade. Where do you find double line here in OU?

Openupgrade is an OCA repo, as others. What you ask is not an OCA rule.

> It's for not hightlighting your contribution over others. Why do you deserve 2 lines meanwhile other contributors only put one?

I don't want to highlight anything. A line is for the company (copyright), another one is for the contributor (author). That's 2 differents things. [See my previous comment](https://github.com/OCA/OpenUpgrade/pull/3818#discussion_r1162011842). Other OCA contributors does the same, and there is no rules for that topics.

In fact, in contributors file, Tecnativa contributors "consumes" [two lines](https://github.com/OCA/reporting-engine/blob/16.0/base_comment_template/readme/CONTRIBUTORS.rst) and GRAP (and other OCA) contributors [one line](https://github.com/OCA/reporting-engine/blob/16.0/sql_request_abstract/readme/CONTRIBUTORS.rst).
 
It's a matter of taste. I don't think Tecnativa set two lines in the ``CONTRIBUTORS``  to be "highlighted". Do you ? 

> And yes, it's already stated in OCA template: https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L1

A template is a proposal. not a guideline.

> And it's not nitpicking

Yes, definitely. And this subject is quite boring. let's move on please.

pedrobaeza: Sorry, the same as you have set other red lines, for me this is one. It's not a matter of taste. It's a sign of respect about dedicating 2 lines for your copyright instead of 1.

legalsylvain:
Hi reviewers.

I think I have exhausted my arguments on the [subject of the header](https://github.com/OCA/OpenUpgrade/pull/3818#discussion_r1161487754) and I don't feel that pedro's request is valid.
So as it stands, I'm not going to make this change and I consider this PR as "ready to merge". 
If you have a point of view, feel free to express it.

andreampiovesana (approved):
LGTM

pedrobaeza (changes requested):
As stated, put only one copyright line, and more being in a file that will contain code from other contributors.

andreampiovesana:
@legalsylvain can you do it?

legalsylvain:
Done.

pedrobaeza (approved):

pedrobaeza:
@legalsylvain please check pre-commit

PR #3820, opened 2023-04-12
[14.0][OU-IMP] mass_mailing: compute medium_id

chienandalu:
`mailing_mailing.medium_id` existed in prior versions as a non computed field thus it isn't recomputed in the update. We need to do it as it's now a required field in the mailing view.

cc @Tecnativa TT29993

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/mass_mailing/14.0.2.2/post-migration.py
+++ b/openupgrade_scripts/scripts/mass_mailing/14.0.2.2/post-migration.py
@@ -1,0 +1,23 @@
+# Copyright 2023 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def set_mailing_medium(env):
+    """The field column exists before migration so the new compute behavior won't be
+    triggered. We need to load the values as it's required in the view"""
+    medium_email = env.ref("utm.utm_medium_email").id
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        UPDATE mailing_mailing
+        SET medium_id = {medium_email}
+        WHERE mailing_type = 'mail'
+        AND medium_id IS NULL
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    set_mailing_medium(env)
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):

MiquelRForgeFlow:
You could add comment in https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/mass_mailing/14.0.2.2/upgrade_analysis_work.txt#L24.

> required field in the mailing view.

I don't understand why odoo do this things

openupgrade_scripts/scripts/mass_mailing/14.0.2.2/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+def set_mailing_medium(env):
+    """The field column exists before migration so the new compute behavior won't be
+    triggered. We need to load the values as it's required in the view"""
+    medium_email = env.ref("utm.utm_medium_email").id
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        UPDATE mailing_mailing
+        SET medium_id = {medium_email}
+        WHERE mailing_type = 'mail'
```
pedrobaeza, line 15: Being required, we should fill all of them independently from the mailing_type, so we should provide an alternative. I would say to put to all the medium email, no matter the mailing_type.

pedrobaeza (commented):
And please change the work file

openupgrade_scripts/scripts/mass_mailing/14.0.2.2/post-migration.py
chienandalu: Well, the other medium_id is `sms`  and that one is provided by the module `mass_mailing_sms` (https://github.com/odoo/odoo/blob/7874c388db9fc476d38390fe5e27d8009031e620/addons/mass_mailing_sms/models/mailing_mailing.py#L55) and there's where another migration script should be made.

pedrobaeza: Uhm, I'm divided here about tolerating a NOT NULL constraint error until all the modules are updated (if they provide the correct migration scripts - you can have extra custom types or the sms one that you mention -), or putting a default and then overwrite it on successive modules.

The first approach allows to detect further problems, but it will only log an ERROR in the migration log, no matter if it's done correctly.

What do you think?

pedrobaeza: Ah, wait, is it only required on the view, but not on the model?

chienandalu: That's it. It's only required from the view

pedrobaeza: Ah,OK, then your approach is better.

PR #3823, opened 2023-04-18
[14.0][OU-FIX] account: change statement state after st_lines filling

pedrobaeza:
We are changing the state of the statement to "Processing" if not all the statement lines are reconciled, but the field used for such comparison is not yet filled, so we have to delay the query until the execution of the other method (fill_account_bank_statement_line_reconciliation).

TT42592


chienandalu (approved):

MiquelRForgeFlow:
Makes sense.

PR #3626, opened 2022-11-23
[14.0][OU-IMP] account: Compute stored function fields by SQL

etobella:
When managing files from account.move, account.move.line, account.payment, account.bank.statement or account.bank.statement.line we should always migrate them by SQL, not using the computed function (it could take hours or, even worse, fail for a Memory failure)

Candidate fields
- [x] account.move
  - [x] sequence_prefix
  - [x] sequence_number
  - [x] payment_state --> Already done on preovious migration
- [x] account.bank.statement
  - [x] sequence_prefix
  - [x] sequence_number
  - [x] is_valid_balance_start
  - [x] previous_statement_id
- [x] account.bank.statement.line
  - [x] amount_residual
  - [x] is_reconciled
- [x] account.move.line
  - [x] matching_number
- [x] account.payment
  - [x] is_internal_transfer
  - [x] destination_account_id
  - [x] is_matched
  - [x] is_reconciled
  - [x] partner_bank_id
- [x] account.partial.reconcile
  - [x] credit_currency_id
  - [x] debit_currency_id

PR #3832, opened 2023-04-25
[16.0] [OU-ADD] mail

marielejeune:
Significant changes:
* Model `mail.channel.partner` renamed to `mail.channel.member`
* Field `scheduled_date` on `mail.mail` is now datetime (not char anymore)


openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
```diff
@@ -0,0 +1,120 @@
...
+        SET state='mail_post'
+        WHERE state='email';
+        """,
+    )
+
+
+def ir_act_server_compute_mail_post_method(env):
+    """
+    Add mail_post_method column on ir_act_server and compute the values
+    """
```
marielejeune, line 67: Not sure if a migration script is useful or if we let Odoo compute and store the values when uploading the module?
Same question for the other new field of the model, `mail_post_autofollow`

legalsylvain: hi @marielejeune. 
Thanks for your contribution on the OpenUpgrade project ! 
Regarding your question : If I understand correctly, your SQL request make the same things than this function : 

https://github.com/odoo/odoo/blob/16.0/addons/mail/models/ir_actions_server.py#L99

The unique reason to "precompute" a field in the pre-migration script, by SQL, is to speed up the computation. 

So, this happens when these two conditions are met:
1) ORM computation is very slower than the SQL request
2) the table has a big amount of rows. This second point of course depends on your instance, but generally, it occures on stock.move(.line), sale.order(.line), mail.message, account.move(.line), etc...

In your case : 
- computation done by ORM is trivial
- there is never "a lot of" rows in ir_act_server table.

So both pre-compute are unnecessary and can be removed.

Note : even if both conditions are met, you can always base your migration on ORM if it's OK for your migration (small database for exemple), and let another contributor improving by an SQL request if needed.
 See for exemple that PR https://github.com/OCA/OpenUpgrade/pull/3626 that improves 13-14 account migration that has been done in a second step.

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,223 @@
...
+mail         / mail.link.preview        / og_mimetype (char)            : NEW
+mail         / mail.link.preview        / og_title (char)               : NEW
+mail         / mail.link.preview        / og_type (char)                : NEW
+mail         / mail.link.preview        / source_url (char)             : NEW required
+# NOTHING TO DO: new model
+
+mail         / mail.mail                / fetchmail_server_id (many2one): previously in module fetchmail
+# NOTHING TO DO
+mail         / mail.mail                / scheduled_date (char)         : type is now 'datetime' ('char')
+# DONE (pre- and post-migration): Keep old char value in DB (with legacy column name) and convert the value to datetime
```
legalsylvain: There is a function that change date to datetime. 
``use date_to_datetime_tz``  
https://oca.github.io/OpenUpgrade/API.html#openupgradelib.openupgrade.date_to_datetime_tz

Here, we could make char -> date (by SQL) then use ``date_to_datetime_tz`` ? 
I think code could be more simple. (and maybe faster, because the "for each" by ORM, in post migration can be slow on table with a lot of row in mail.mail.

Don't you think ?

legalsylvain (commented):
except my two remarks. LGTM. Thanks !

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
marielejeune: Good idea, I agree to (try to) doing it by SQL. The char version of `scheduled_date` contains a datetime, not a date. So I would rather use TO_TIMESTAMP to directly convert it? But then idk how  to deal with the timezone: the timezone might be specified in the char field, or it is assumed to be UTC. But TO_TIMESTAMP sql method doesn't allow to deal with timezones?

openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
marielejeune: Clear, thanks! I will remove these two methods

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
legalsylvain: >  the timezone might be specified in the char field, or it is assumed to be UTC

I don't know could you provides some samples of what the char field contains ? If it is UTC, I guess TO_TIMESTAMP could do the job.

hbrunn: judging from https://github.com/OCA/OCB/blob/15.0/addons/mail/models/mail_mail.py#L94-L97 and https://github.com/OCA/OCB/blob/15.0/addons/mail/models/mail_mail.py#L158-L159 we can safely assume this to be in the standard Odoo datetime format, and that postgres can autoconvert to timezone without time stamp - did you try doing nothing? I'd expect this to yield the correct result.

hbrunn (approved):

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
hbrunn: ...you can set a string scheduled date for v15 in CI in a file similar to `data.py` in https://github.com/OCA/OpenUpgrade/tree/16.0/openupgrade_scripts/scripts/base/16.0.1.3/tests, and then assert something about it in the test file

remytms:
@legalsylvain or @pedrobaeza  Can you reference this PR in the Migration to 16.0 issue with `/ocabot migration mail` ? :)

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,222 @@
...
+
+mail         / ir.mail_server           / mail_template_ids (one2many)  : NEW relation: mail.template
+# NOTHING TO DO: new field
+
+mail         / mail.channel             / _inherits                     : DEL _inherits: {'mail.alias': 'alias_id'}
+mail         / mail.channel             / alias_id (many2one)           : DEL relation: mail.alias, required
+# NOTHING TO DO: removed mail.alias.mixin from channel. See commit https://github.com/odoo/odoo/commit/ac00d19c8db572932a3863ab21a9a1f432dd93b5
+mail         / mail.channel             / channel_last_seen_partner_ids (one2many): DEL relation: mail.channel.partner
+mail         / mail.channel             / channel_member_ids (one2many) : NEW relation: mail.channel.member
+# DONE (pre-migration): field was renamed when calling rename_models
```
remytms: @marielejeune Does the values of the column `channel_last_seen_partner_ids` be copied to `channel_member_ids` ?

I'm not sure that renaming the model `mail.channel.partner` to `mail.channel.member` would be enough to keep values from the column `channel_last_seen_partner_ids` to `channel_member_ids`. Is it enough ?

remytms: @legalsylvain Any thought about that question about the migration of `mail` module and the rename of `channel_last_seen_partner_ids` to `channel_member_ids`. I can also make some tests to be sure, but if you already have an opinion about that, it will helps. :)

legalsylvain: Good question ! (and I started to write the use case "rename_model" yesterday...)

AFAIK, table should be renamed. 
https://github.com/OCA/openupgradelib/blob/f87f553aca1e8fb8447bc152473518c6b070ed1a/openupgradelib/openupgrade.py#LL872C1-L874C45

I'm not sure, thought how it has to be done.
(pseudo code)
A) rename_fields( channel_last_seen_partner_ids -> channel_member_ids)
B) rename_columns( channel_last_seen_partner_ids -> channel_member_ids)
C) rename_tables(channel_last_seen_partner_rel -> channel_member_rel)

@MiquelRForgeFlow, @pedrobaeza : any idea, as a regular contributor of OU scripts ?

MiquelRForgeFlow: A) yes, that's needed.
B) no, it's one2many.
C) no, it's not many2many.

legalsylvain: Thanks a lot for your reactivity !

remytms: Thanks for your replies. :) So how do we proceed. Should we merge this PR and then make a fix to add the rename of this field, or should I make a new PR with the work of @marielejeune with a commit fixing the rename of this field ?

MiquelRForgeFlow: My vote is just to wait for @marielejeune to do the suggested changes.

marielejeune: Hi, sorry I have a lot of work for the moment. My plan is to work on it during the Spanish OCA days next week :-)

marielejeune: Hi @MiquelRForgeFlow, I'm not sure why the field must be renamed? 
I made a test migrating from 15 to 16 and the `mail.channel.partner` records of `channel_last_seen_partner_ids` in v15 are still preserved (as `mail.channel.member` records) in `channel_member_ids` field. 
I'm in Valencia if you want to discuss about it.

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
marielejeune: @hbrunn you're right, doing nothing works. Going to add a test

marielejeune: Done, could you review it please?

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
MiquelRForgeFlow: Hi @marielejeune 
One thing is the data in the mail_channel_member, which as you say it's preserved, and another thing is the ir_model_fields data. By doing the field rename of channel_last_seen_partner_ids to channel_member_ids (using rename_fields), you are preserving the ir.model.fields data instead of forcing new one.

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
hbrunn: thanks, looks good. I did however just try this out with a db that has `scheduled_date==''` and that breaks, so maybe you could create demo data to this effect and add an update statement to set empty strings to null?

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
marielejeune: Done, could you check?

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
marielejeune: I added an update statement for empty strings and added a test case

hbrunn: great, thanks!

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
MiquelRForgeFlow: Good!

openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
```diff
@@ -0,0 +1,91 @@
...
+    "mail_message": [("add_sign", "email_add_signature")],
+}
+_columns_copies = {
+    "mail_channel_rtc_session": [
+        ("channel_partner_id", "channel_member_id", "integer")
+    ],
+}
+
+
+def mail_channel_member_ir_model_data(env):
```
MiquelRForgeFlow: please, use rename_xmlids method instead

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,222 @@
...
+DEL ir.model.constraint: mail.constraint_mail_channel_rtc_session_channel_partner_unique
+# NOTHING TO DO: noupdate="0" records
+
+NEW ir.rule: mail.ir_rule_mail_channel_member_group_system (noupdate)
+NEW ir.rule: mail.ir_rule_mail_channel_member_group_user (noupdate)
+# NOTHING TO DO: New noupdate="1" records
+
+DEL ir.rule: mail.ir_rule_mail_channel_partner_group_system (noupdate)
+DEL ir.rule: mail.ir_rule_mail_channel_partner_group_user (noupdate)
+# DONE (post-migration): try to delete
```
MiquelRForgeFlow: don't delete. you have to add in the rename_xmlids.

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,222 @@
...
+DEL ir.model.access: mail.access_mail_channel_partner_user
+DEL ir.model.access: mail.access_mail_resend_cancel
+NEW ir.model.constraint: mail.constraint_mail_activity_check_res_id_is_set
+NEW ir.model.constraint: mail.constraint_mail_channel_channel_type_not_null
+NEW ir.model.constraint: mail.constraint_mail_channel_group_public_id_check
+NEW ir.model.constraint: mail.constraint_mail_channel_member_partner_or_guest_exists
+NEW ir.model.constraint: mail.constraint_mail_channel_rtc_session_channel_member_unique
+NEW ir.model.constraint: mail.constraint_res_users_notification_type
+DEL ir.model.constraint: mail.constraint_mail_channel_partner_partner_or_guest_exists
+DEL ir.model.constraint: mail.constraint_mail_channel_rtc_session_channel_partner_unique
```
MiquelRForgeFlow, line 180: use delete_sql_constraint_safely method in pre-migration

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
marielejeune: These two ir.rules changed (name and domain_force for the second one). They are in noupdate="1" but not present in the noupdate_changes.xml file. Hence just renaming the xmlid is not sufficient to change them. What should I do? Adding them manually to noupdate_changes.xml?

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
marielejeune: Done

openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
marielejeune: Done

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
MiquelRForgeFlow: Yes, in the `noupdate_changes.xml` you should add:

```xml
  <record id="ir_rule_mail_channel_member_group_system" model="ir.rule">
    <field name="model_id" ref="model_mail_channel_member"/>
    <field name="name">mail.channel.member: admin can manipulate all entries</field>
  </record>
  <record id="ir_rule_mail_channel_member_group_user" model="ir.rule">
    <field name="domain_force">[
                '|',
                    '&amp;',
                        ('channel_id.channel_type', '!=', 'channel'),
                        ('channel_id.is_member', '=', True),
                    '&amp;',
                        ('channel_id.channel_type', '=', 'channel'),
                        '|',
                            ('channel_id.group_public_id', '=', False),
                            ('channel_id.group_public_id', 'in', [g.id for g in user.groups_id])]
            </field>
    <field name="model_id" ref="model_mail_channel_member"/>
    <field name="name">mail.channel.member: write its own entries</field>
  </record>
```

openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
```diff
@@ -0,0 +1,92 @@
...
+    openupgrade.rename_fields(env, _fields_renames)
+    openupgrade.rename_models(env.cr, _models_renames)
+    openupgrade.rename_tables(env.cr, _tables_renames)
+    openupgrade.rename_columns(env.cr, _columns_renames)
+    openupgrade.copy_columns(env.cr, _columns_copies)
+    openupgrade.rename_xmlids(env.cr, _xmlids_renames)
+    ir_act_server_rename_state_email(env)
+    mail_channel_channel_type_required(env)
+    scheduled_date_set_empty_strings_to_null(env)
+    delete_obsolete_constraints(env)
```
MiquelRForgeFlow: I would put this one above all else :)

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/mail/16.0.1.10/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,221 @@
+---Models in module 'mail'---
+obsolete model mail.channel.partner
+obsolete model mail.resend.cancel [transient]
+model fetchmail.server (moved from fetchmail)
```
cvinh, line 4: @marielejeune is there any specific reason why I don't see _models_renames for fetchmail model ? Thanks

cvinh: Ok I understood, the model has been 'moved' and not 'renamed'

PR #3834, opened 2023-04-27
[14.0][OU-ADD] website_blog: update cover properties field

pilarvargas-tecnativa:
In v13, blog posts without an image in the header were indicated with the parameters "backgroung-image" : "none" and "background-color" : "value" stored in the blog_post.cover_properties field.  
When migrating to v14 this style setting is indicated by "backgroung-image" : "none" and "backgroung_color_class" : "value" and "background-color" disappears.  
For this reason the old publications do not adopt the corresponding style on the website if they use colour and not an image in their header. This is solved by rewriting the data contained in the field blog_post.cover_properties.

cc @Tecnativa TT39016

@chienandalu @pedrobaeza please review

### DIFF
--- a/openupgrade_scripts/scripts/website_blog/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website_blog/14.0.1.0/post-migration.py
@@ -1,6 +1,18 @@
 from openupgradelib import openupgrade
 
+from odoo.tools.json import scriptsafe as json_safe
+
+
+def update_cover_properties_field(env):
+    blog_post = env["blog.post"].search([])
+    for post in blog_post:
+        cover_properties = json_safe.loads(post.cover_properties)
+        cover_properties["background_color_class"] = "o_cc3"
+        cover_properties.pop("background-color", None)
+        post.write({"cover_properties": json_safe.dumps(cover_properties)})
+
 
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "website_blog", "14.0.1.0/noupdate_changes.xml")
+    update_cover_properties_field(env)


chienandalu (commented):
Fast check I'd say the only major change is that:

- A mixin is used (no problem with this)
- Instead of a background-color, a color class is used, so there's no need to change the other keys IMO.

So the script should just consist in:

- Set the new `background_color_class` property
- Can old background colors be converted?
- Drop the old `background-color` one

pilarvargas-tecnativa:
> Fast check I'd say the only major change is that:
> 
> * A mixin is used (no problem with this)
> * Instead of a background-color, a color class is used, so there's no need to change the other keys IMO.
> 
> So the script should just consist in:
> 
> * Set the new `background_color_class` property
> * Can old background colors be converted?
> * Drop the old `background-color` one

The old background colour is not necessary to convert it because in v13 it was not customisable from the frontend, there was only a default colour that was defined in the class "oe_black" which is the equivalent to "o_cc o_cc3", as you say it would be enough to delete that key and add the new one. 
Thanks

chienandalu:
> The old background colour is not necessary to convert it because in v13 it was not customisable from the frontend, there was only a default colour that was defined in the class "oe_black" which is the equivalent to "o_cc o_cc3", as you say it would be enough to delete that key and add the new one

Ok if it isn't editable. Just point that o_cc3 isn't a color but a color combination. The default value should be just `o_cc3`: https://github.com/odoo/odoo/blob/1785047d58f28ba11e11e40ba134e89bb235fa8b/addons/website/models/mixins.py#L112

openupgrade_scripts/scripts/website_blog/14.0.1.0/post-migration.py
```diff
@@ -1,6 +1,20 @@
 from openupgradelib import openupgrade
 
+from odoo.tools.json import scriptsafe as json_safe
+
+
+def update_cover_properties_field(env):
+    records = env["blog.post"]
+    for record in records:
+        cover_properties = json_safe.loads(record.cover_properties)
+        if cover_properties.get("background-image") == "none":
```
chienandalu: There's no need to check this. Just set the new property and drop the old one

pilarvargas-tecnativa:
@chienandalu all done! thanks

openupgrade_scripts/scripts/website_blog/14.0.1.0/post-migration.py
```diff
@@ -1,6 +1,19 @@
...
+from odoo.tools.json import scriptsafe as json_safe
+
+
+def update_cover_properties_field(env):
+    records = env["blog.post"]
+    for record in records:
+        cover_properties = json_safe.loads(record.cover_properties)
+        cover_properties["background_color_class"] = "o_cc3"
+        if "background-color" in cover_properties:
+            del cover_properties["background-color"]
```
chienandalu:
```suggestion
        cover_properties.pop("background-color", None)
```

openupgrade_scripts/scripts/website_blog/14.0.1.0/post-migration.py
```diff
@@ -1,6 +1,18 @@
 from openupgradelib import openupgrade
 
+from odoo.tools.json import scriptsafe as json_safe
+
+
+def update_cover_properties_field(env):
+    records = env["blog.post"]
```
chienandalu: As we just tried. Fix:
```suggestion
    records = env["blog.post"].search([])
```

chienandalu (commented):
Change `records` and `record` for a more expressive name

chienandalu (approved):
Now it's perfect :ok_hand: :slightly_smiling_face:

MiquelRForgeFlow (approved):
LGTM

MiquelRForgeFlow:
Selection_1140

shouldn't be better to call this method?

chienandalu:
> shouldn't be better to call this method?

We don't want to squash existing valid data (e.g.: background images)

pedrobaeza (approved):
Tested on real migration

PR #3835, opened 2023-04-27
[13.0][OU-FIX] purchase: set date_approve in all the confirmed orders

chienandalu:
If there's no double approval, all the previous orders will be ignored.

cc @Tecnativa TT42662

please review @pedrobaeza


chienandalu:
This error is blocking CI: `setup command: use_2to3 is invalid`

I tried this on requirements.txt https://stackoverflow.com/a/69100830 but I guess setuptools come installed from somewhere else in the GH action

pedrobaeza (approved):

PR #3847, opened 2023-05-03
[16.0] [OU-FIX] base: convert translatable text columns to jsonb

marielejeune:
Previously we just converted varchar columns, but there exist some html fields that are translatable. Html fields are text columns in DB.

### DIFF
--- a/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
@@ -70,7 +70,7 @@ def migrate(cr, version):
             continue
         columns = tools.sql.table_columns(cr, table)
         if field in columns:
-            if columns[field]["udt_name"] == "varchar":
+            if columns[field]["udt_name"] in ["varchar", "text"]:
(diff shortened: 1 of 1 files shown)


marielejeune:
Hi @hbrunn, here is a minor fix for migrating HTML translatable fields. Could you have a look please?

hbrunn (approved):

pedrobaeza:
@marielejeune when you say HTML translatable fields, do you mean those with `translate=html_translate`? If so, AFAIK this requires extra work. Seeing code, it seems just `Text` fields, which are not HTML (or at least, they are not translated by fragments).

marielejeune:
> @marielejeune when you say HTML translatable fields, do you mean those with `translate=html_translate`? If so, AFAIK this requires extra work. Seeing code, it seems just `Text` fields, which are not HTML (or at least, they are not translated by fragments).

No I was just speaking of fields.Html fields, with translate=True attribute, as body_html in mail.template for example.
https://github.com/odoo/odoo/blob/1430644908d7d9a3119a1315389f5bf11888356a/addons/mail/models/mail_template.py#L59

pedrobaeza:
OK, let's go then.

PR #3848, opened 2023-05-03
[11.0][OU-IMP] sale: Auto-create icp from old deposit_product_id_setting field (v10) to new default_deposit_product_id field (v11)

victoralmau:
Auto-create icp from old `deposit_product_id_setting` field (v10) (https://github.com/odoo/odoo/blob/10.0/addons/sale/models/sale_config_settings.py#L117) to new `default_deposit_product_id` field (v11) (https://github.com/odoo/odoo/blob/11.0/addons/sale/models/res_config_settings.py#L141)

@Tecnativa TT42886

### DIFF
--- a/addons/sale/migrations/11.0.1.1/post-migration.py
+++ b/addons/sale/migrations/11.0.1.1/post-migration.py
@@ -131,3 +131,27 @@ def migrate(env, version):
         ],
     )
     activate_proforma(env)
+    # Auto-create icp from old deposit_product_id_setting field (v10)
+    # to new default_deposit_product_id field (v11)
+    openupgrade.logged_query(
+        env.cr,
+        """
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
make sense.

victoralmau:
Another possible approach (probably in `base`), would be to auto-create the `ir_config_parameter` records according to the existing records in `ir_values` with the `xx.config.settings` model:

The `ir_config_parameter` fields would be:
- _key_ ("xxx." + name field of `ir_values`).
- _value_ (value field of `ir_values`)

What do you think about this?

legalsylvain:
@victoralmau what is the interest to handle that in ``base`` module. AFAIK, it is a "sale" topic, so the current PR looks ok for me.

pedrobaeza (approved):

PR #3849, opened 2023-05-03
[DOC][FIX] Do not ignore output path. (Fix 'View page source ' broken links)

legalsylvain:
Rational : 

- go to the documentation https://oca.github.io/OpenUpgrade/
- click on "View page source" link (top right button)
- the link of the file in _sources is broken

Note : the original V14 branch doesn't contains that exclusion : see : 
https://github.com/OCA/OpenUpgrade/blame/14.0/.gitignore#L66

but the V15 initialization introduced it : 
https://github.com/OCA/OpenUpgrade/blame/15.0/.gitignore#L69

@pedrobaeza : an idea why this line has been introduced ? 

Side question : I have no idea why docs/_build/ is also ignored, but it doesn't seems to break anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,6 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
-docs/_sources/
 docs/.doctrees
 docsource/modules50-60.rst
 docsource/modules60-61.rst
--- a/docsource/conf.py
+++ b/docsource/conf.py
@@ -135,7 +135,10 @@ html_last_updated_fmt = "%b %d, %Y"
 # html_split_index = False
 
 # If true, the reST sources are included in the HTML build as _sources/<name>.
-# html_copy_source = True
+html_copy_source = False
+
+# Hide the Page source link in each documentation page's footer.
+html_show_sourcelink = False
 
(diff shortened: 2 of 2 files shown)


pedrobaeza (commented):
It was copied from 14.0, and originally that branch contained that line, but it was fixed by @StefanRijnhart in a21d79f348e7ad278fdd7f6cf15bec0be3351d56

You should better apply the full patch

legalsylvain:
> It was copied from 14.0, and originally that branch contained that line, but it was fixed by @StefanRijnhart in https://github.com/OCA/OpenUpgrade/commit/a21d79f348e7ad278fdd7f6cf15bec0be3351d56

Thanks for the link to the fix ! 
Thanks to the creation of a branch dedicated to documentation, we won't have to do this kind of duplicate PR in the future.

> You should better apply the full patch

done !

Tested locally. all is OK. Fast-tracking trivial backport fix :

/ocabot merge nobump

pedrobaeza:
There's no need of using ocabot. That's on OCA modules, where extra commits (or version bumping) are performed. Here on OpenUpgrade, you can just click on the merge button, and you save extra CI builds for no reason.

PR #3852, opened 2023-05-05
[15.0][IMP] mail: migrate mail templates from mako to qweb

marielejeune:
Mail templates are qweb templates since v15. Migrate mail templates regarding :
* Dynamic placeholders in char/text fields (`email_from`, `email_to`, `reply_to` and `subject`)
* Dynamic placeholders in html field `body_html`
* If / else conditions in `body_html`
* For loops in `body_html`
* Setting variables in `body_html`

NB: migrate all translations for translatable fields (`subject` and `body_html`)

Many thanks to @sbidoul for his help writing this migration script

### DIFF
--- a/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
+++ b/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
@@ -1,4 +1,10 @@
+import html
+import re
+
+from bs4 import BeautifulSoup
 from openupgradelib import openupgrade
+
+_PY_VAR_PATTERN = "[a-zA-Z_][a-zA-Z0-9_]*"
 
 
 def _map_activity_type_chaining_type_field(env):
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -64,6 +70,129 @@ def finish_migration_to_mail_group(env):
...
+                    )
+                )
+            )
+        )
+    )
+
+
+def _migrate_mail_templates(env):
+    """
+    Migrate all mail templates from jinja2 to qweb
```
sbidoul, line 180:
```suggestion
    Migrate all mail templates from mako to qweb
```

legalsylvain (commented):
nice addition ! 
Typically something that could be tested. @hbrunn. I remember something new regarding test in Openupgrade. Is it available in v15 ? could you mention how to write test ? 
I don't have a very clear vision of this subject. (and by the way, I should have one, as part of the project on documentation ;-))

sbidoul:
>  could you mention how to write test ?

Yeah, we'd be happy to add some unit tests, but not sure how.

pedrobaeza:
Not possible unless backporting 29d12801701bdfa71a0e409a6cf7f5e5ff50bd15 from v16

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -1,5 +1,11 @@
+import html
+import re
+
+from bs4 import BeautifulSoup
```
pedrobaeza, line 4: Is this library included in standard Odoo?

sbidoul: I don't think so. Is there a mechanism to declare third party library dependencies?

sbidoul:
About test, is it possible to `import` code in `post-migration.py`. In which case I'll add doctests.

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
pedrobaeza: That's a problem, as doing the upgrade, this will fail. Can you do the transformations without requiring this library?

legalsylvain:
> Not possible unless backporting https://github.com/OCA/OpenUpgrade/commit/29d12801701bdfa71a0e409a6cf7f5e5ff50bd15 from v16

Thanks I'll take a look.

> Yeah, we'd be happy to add some unit tests, but not sure how.

I'll document it ;-)

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
legalsylvain: > I don't think so. Is there a mechanism to declare third party library dependencies?

I'm working here #3854 to use classic generation of requirements.txt based on the dependency (still wip). If it's ok, we can just add BeautifulSoup as a dependency of openupgrade_scripts module in the manifest.
Then just change the doc : "pip install openupgradelib" by "pip install -r ./requirements.txt".
That should do the job. Don't you think ?

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -64,6 +70,129 @@ def finish_migration_to_mail_group(env):
...
+    def repl(match):
+        var_name, var_value = match.groups()
+        var_value = html.escape(var_value)
+        return f'<t t-set="{var_name}" t-value="{var_value}"/>'
+
+    return re.sub(pattern, repl, string, flags=re.MULTILINE)
+
+
+def _migrate_html_attributes(string):
+    soup = BeautifulSoup(string, multi_valued_attributes=None, features="html.parser")
```
hbrunn, line 146: maybe you can use https://lxml.de/lxmlhtml.html here to avoid the dependency on bs?

pedrobaeza: Yes, I think this can be done the same with standard Odoo XML manipulation library, which is lxml.

pedrobaeza:
Superseded by #4105

PR #3853, opened 2023-05-05
[REF] Remove docs from 16.0 branch

legalsylvain:
- remove all docs folder
- remove all docsource folder, except coverage file of the current version. (15 -> 16)

- simplify pre-commit stuff, removing exception regarding docs / docsource folders.
- minor changes in main README.md file

**Consequency :**
- the 16.0 branch is a more "classic" module folder, with less exception
- save CI execution with a reduction of about 20%. (from 13 minutes to 10 minutes)

Note : once validated, I'll do the same for recent branches with new modules design. (14 / 15)

Issue #3702, opened 2023-01-25
Documentation Branch


etobella:
@sbidoul @bguillot @pedrobaeza @legalsylvain 

We were reviewing the documentation and it is confusing to have a documentation folder for each version, 

Wouldn't have more sense to create a new repo or branch for OpenUpgrade documentation only?

pedrobaeza:
Not really, as the module coverage should be update with each PR integrating the migration scripts. It would be a lot of work to synchronize that with 2 repos.

pedrobaeza:
The rule is:

- Main explanations from latest branch.
- Module coverage taken from each branch.

legalsylvain:
@pedrobaeza, @StefanRijnhart : do you how the branch is selected to generate the website ? https://oca.github.io/OpenUpgrade/
(I don't have access to the "settings / page" part of this repo).
Thanks.

StefanRijnhart:
https://github.com/OCA/OpenUpgrade/settings/pages
![image](https://user-images.githubusercontent.com/1033124/214608477-566f6f80-7bcd-4de6-a53e-dc3deef64584.png)

etobella:
@pedrobaeza We were talking about how to do things, not the coverage of each module. 

There is a documentation of how to do a migration on https://github.com/OCA/OpenUpgrade/blob/15.0/docsource

And this information might not be ported from version to version. It mitght be cleaner to add this on a separate branch / repo. 

Obviously, the current version configuration must be on the expected branch

StefanRijnhart:
I imagine we can switch the documentation to 16.0 soon, once https://github.com/OCA/OpenUpgrade/pull/3680 is merged.

legalsylvain:
hi. 

I don't understand your points @pedrobaeza and @StefanRijnhart.

documentation and openupgrade scripts have different lifecycles.
- openupgrade script (and framework) are related to a version, so having a branch per version is perfect.
- documentation is for all the version, so having documentation in each branch doesn't makes sense.

So :
- Each time a new release is out and became the default one, it makes obsolete all the previous documentation PR.
A simple exemple is regarding [the use case "field renaming"](https://github.com/OCA/OpenUpgrade/pull/3493) PR.
- The PR has been done against the 15.0 branch (July 2022). 
- The branch 16.0 has been created in October 2022.
- The PR has been merged after the creation of the 16.0

[...]

PR #3855, opened 2023-05-06
[ADD][16.0] New documentation workflow to update documentation website if the coverage changed

legalsylvain:
on push on 16.0 AND if docsource/modules150-160.rst changed : 

- checkout documentation branch
- build documentation
- commit changes on documentation branch
-> as a result, documentation website is updated.

I had mentioned a scheduled task in my PoC a few months ago:  https://github.com/OCA/OpenUpgrade/issues/3702#issuecomment-1405805444 but the "paths" options looks better. (save CI AND builds are done only if the coverage file has changed)

This answers the questions of :
- @StefanRijnhart : https://github.com/OCA/OpenUpgrade/issues/3702#issuecomment-1404671537 
- @pedrobaeza : https://github.com/OCA/OpenUpgrade/pull/3853#issuecomment-1536709095
- ...

Note : Once merged, I'll do the same PR on recent branches (14.0 and 15.0)

CC : @hbrunn, @etobella, @bguillot

### DIFF
--- a/.github/workflows/documentation-commit.yml
+++ b/.github/workflows/documentation-commit.yml
@@ -1,0 +1,66 @@
+# On each push in 16.0 branch,
+# AND if the coverage file changed,
+# build documentation branch and commit the changes
+# so that the changes are visible on the website
+# https://oca.github.io/OpenUpgrade/
+
+name: Build and commit documentation
+
+on:
+  push:
+    paths: ["docsource/modules150-160.rst"]
+
+jobs:
+  documentation-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out OpenUpgrade Documentation
+        uses: actions/checkout@v2
+        with:
+          ref: "documentation"
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Check out Odoo
+        uses: actions/checkout@v2
+        with:
+          repository: odoo/odoo
+          ref: "16.0"
+          fetch-depth: 1
+          path: odoo
+      - name: Configuration
+        run: |
+          sudo apt update
+          sudo apt install \
+              expect \
+              expect-dev \
+              libevent-dev \
+              libldap2-dev \
+              libsasl2-dev \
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Seems OK reading the code. A real test should be done.

The only possible drawback right now is that docsource/modules150-160.rst is modified with incorrect RST syntax, and this is not detected until the push, having a red here.

Is there any possibility to have an RST syntax checker on PR as an extra CI like we had previously for the whole documentation?

legalsylvain:
> Seems OK reading the code. A real test should be done.

It looks working on this fork : https://github.com/grap/OpenUpgrade-1/, so I think it should be fine.

> The only possible drawback right now is that docsource/modules150-160.rst is modified with incorrect RST syntax, and this is not detected until the push, having a red here.

> Is there any possibility to have an RST syntax checker on PR as an extra CI like we had previously for the whole documentation?

- indeed, check correct syntax of the coverage module was done before by CI, but not in a good stage. it was a consequence of the fail of the documentation.yml when calling build_openupgrade_docs. (in https://github.com/OCA/OpenUpgrade/pull/3856 : travis raise an error ). IMO, That's not the correct way. Should be done by pre-commit. Otherwise, developper can not see that the syntax is incorrect when working locally.

- In all the other OCA repo, it is handled very early by pre-commit. See : https://github.com/OCA/pos/pull/995 : github raise an error on pre-commit.

In Openupgrade project, all the rst file are checked by pre-commit because there are in a module (ex : #3858), but this rst file is out of a module (ex : #3857), so the check ``rst-syntax-error`` present in the file : https://github.com/OCA/OpenUpgrade/blob/16.0/.pylintrc-mandatory#L51 looks not executed by ``pylint_odoo`` for this file.

Not a big deal, 
- because the changes of this file are always very simple,
- it is not blocking the CI of the "normal" branch.
- we have a warning in the github notification so we can fix it quiclky

As that is not the topic of the current PR propose to merge. OK for you ?

Let's think about how to deal with this issue in the simplest way possible during this time.

pedrobaeza (approved):

PR #3859, opened 2023-05-07
15.0 rem documentation

legalsylvain:
Similar to : 
- https://github.com/OCA/OpenUpgrade/pull/3853
-  ~and https://github.com/OCA/OpenUpgrade/pull/3855~

but for v15.


.github/workflows/documentation-commit.yml
```diff
@@ -1,44 +1,35 @@
...
-      DB_PASSWORD: "odoo"
-      DB_PORT: 5432
-      DB_USERNAME: "odoo"
-      DOWNLOADS: https://github.com/OCA/OpenUpgrade/releases/download/databases
-      ODOO: "./odoo/odoo-bin"
-      PGHOST: "localhost"
-      PGPASSWORD: "odoo"
-      PGUSER: "odoo"
+  documentation-commit:
+    runs-on: ubuntu-latest
```
pedrobaeza: Not sure we should use always the latest version instead of Ubuntu. It can break the CI suddenly.

.github/workflows/documentation-commit.yml
```diff
@@ -1,44 +1,35 @@
...
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: '3.10'
       - name: Check out Odoo
         uses: actions/checkout@v2
         with:
           repository: odoo/odoo
-          ref: "15.0"
+          ref: "16.0"
```
pedrobaeza: Incorrect branch

.github/workflows/documentation-commit.yml
legalsylvain: Don't know. Is there a rule, in OCA ? 

in "classic OCA repo", using copier template : 
- for test file, it is ubuntu-latest, [see](https://github.com/OCA/pos/blob/16.0/.github/workflows/test.yml#L14)
- for pre-commit file, it is ubuntu-22.04 [see](https://github.com/OCA/pos/blob/16.0/.github/workflows/pre-commit.yml#L14).

.github/workflows/documentation-commit.yml
legalsylvain: Don't think so. here we have to install odoo to have the possibillity to install openupgradelib to have the possibility to build docs (that includes openupgradelib docstring). 
odoo version doesn't matter here. so all this part is just a copy / past of https://github.com/OCA/OpenUpgrade/blob/documentation/.github/workflows/documentation-commit.yml#L19-L57.

I fear that if we set various version of python, we get various version of sphynx, and so not the same result, if a push is done against V16 (python3.10) or V14 (python3.7).

Note : this second commit (that add again documentation-commit.yml) file on v15 is a proposal.

- In old design, if a push was done on V16, documentation was updated.
- In new design, I propose to make the update if a coverage changed in one of the 3 latest versions (v14 / v15 / v16)

**Note** : having all openupgradelib documentation in openupgrade documentation makes the things very complex. If one day, we decide to publish a dedicated documentation for openupgradelib and remove that part https://oca.github.io/OpenUpgrade/API.html#module-openupgradelib.openupgrade, all the workflow related to the documentation will be very more simple.
(something like https://oca.github.io/openupgradelib)

.github/workflows/documentation-commit.yml
pedrobaeza: It was following this logic: https://github.com/OCA/oca-addons-repo-template/pull/181

.github/workflows/documentation-commit.yml
pedrobaeza: But it means that with each new branch, we have to update all the "supported" branches to match the same Odoo and libraries version... Isn't possible to "download" the stuff from the default branch instead?

legalsylvain: Indeed. It is far from ideal. 

For the time being, I removed this commit that readd a documentation-commit file for 14.0 and 15.0 branch. So there is no regression regarding previous design. (documentation is updated if push on 16.0 branch) 

In the near future, I would try to play with https://docs.github.com/en/actions/using-workflows/reusing-workflows, but it is not the priority.

In the meantime, I think  The resolution of this issue will make all these questions obsolete. (https://github.com/OCA/OpenUpgrade/issues/3862)

.github/workflows/documentation-commit.yml
legalsylvain: thanks for the link. 
topic obsolete, as I removed the whole file. But thanks for the information.

legalsylvain:
Can be merged.

PR #3869, opened 2023-05-15
[16.0][ADD] base_setup: Nothing to do

remytms:
Analysis file is not generated by Upgrade Analysis.

I spotted theses changes:

```
models/res_config_settings.py:
del module_google_drive
del module_google_spreadsheet
del module_pad
```
These changes are done in a transient model so there is nothing to do (thanks @legalsylvain see below).

Theses fields trigger installation of modules that does not exists in 16.0 anymore. So I think it's ok to just do nothing during migration for `base_setup`.


remytms:
This module is a dependency of https://github.com/OCA/OpenUpgrade/pull/3779

remytms:
I migrated a database with `base_setup` installed and with `google_drive` module installed. I get an error on the settings view. I will investigate to know if it's something that should be fixed in this module or by removing the old `google_drive` module.

remytms:
The issue comes from the fact that the `google_drive` module was still installed.

legalsylvain:
Hi @remytms. 
> I spotted theses changes:

AFAIK, the changes are done in a transient model. changes in transient model are ignored by upgrade_analysis module, as there is nothing to do.

> I migrated a database with base_setup installed and with google_drive module installed

I don't see the relation between ``base_setup`` and ``google_drive``.  ``base_setup`` depends on ``base`` + ``web`` in V15, and ``google_drive`` is not autoinstallable. in fact ``google_drive`` depends on ``base_setup`` and ``google_account``.

> The issue comes from the fact that the google_drive module was still installed.

I am not very comfortable with this use case. If I summary correctly. (please correct me if I'm wrong).

- ``google_drive`` module exists in odoo 15 CE.
- ``google_drive`` module doesn't exist in odoo 16 CE nor in EE version. 
- ``google_drive`` is deprecated by Odoo S.A. see : https://github.com/odoo/odoo/pull/94747
- ``google_drive`` has not been recovered by OCA, AFAIK.

In that context, 
- if Openupgrade does nothing, the database is in an incorrect state in the end of the openupgrade V16 process that is not very desirable. (requieres manual uninstallation).
- if OpenUpgrade uninstall the module, we will loose data present in ``google.drive.config`` model.  (And it is not in the logic of openupgrade to delete data)

@pedrobaeza, @StefanRijnhart : how can openupgrade handle this use case elegantly? 

I was wondering if writing ``_merged_modules = {"google_drive": "base"}`` in the 16.0 ``apriori.py`` file could be a solution.

pedrobaeza:
No, that's not possible. You are losing features, so no merge should be done. The module remains "to upgrade". You decide if uninstall it in a safe way, or migrate it to the new version, but nothing has to be done at OpenUgprade level.

remytms:
> AFAIK, the changes are done in a transient model. changes in transient model are ignored by upgrade_analysis module, as there is nothing to do.

Indeed. Thanks for spotting that.

> I don't see the relation between `base_setup` and `google_drive`. `base_setup` depends on `base` + `web` in V15, and `google_drive` is not autoinstallable. in fact `google_drive` depends on `base_setup` and `google_account`.

The relation is that, on a 15.0 database, if you check the `module_google_drive` box, you end up with a database with `google_drive` module installed. So I wanned to check if such a database migrate correctly.

legalsylvain:
@remytms , limitation of the bot for the time being... only maintainers can call it.

/ocabot migration base_setup

pedrobaeza:
> limitation of the bot for the time being...

It's a limitation put on purpose, for avoiding people hijacking other older migration PRs.

pedrobaeza:
Next time please use `[OU-ADD]` as tag.

PR #3870, opened 2023-05-15
[16.0][OU-ADD] base_iban: Nothing to do

remytms:
There is no analysis file generated and there is no DB changes. So nothing to do.

But it depends on `account` which as no migration script yet.


legalsylvain:
/ocabot migration base_iban

<dependency>Depends on : 

pedrobaeza:
Please change tag to  `[OU-ADD]`.

pedrobaeza (approved):
/ocabot merge nobump

PR #3871, opened 2023-05-15
[16.0][OU-ADD] base_vat: Nothing to do

remytms:
There is two views that where removed. But they are not needed anymore and I presume they will be deleted by the upgrade mechanism.

It depends on `account` which has no migration script yet.


legalsylvain:
/ocabot migration base_vat

<dependency>Depends on : 

pedrobaeza:
Please change tag to  `[OU-ADD]`.

alexis-via (approved):

PR #3872, opened 2023-05-15
[DOC] use case : module renaming

legalsylvain:


docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+As long as the renaming has not been identified, the ``upgrade_analysis.txt`` generated
+at the initialization of the OpenUpgrade version is partially wrong, because openupgrade
+considers that the xml_id of the module are new, while they can come from the module of
+the previous version.
+
+For the modules that concern Odoo, you must therefore update the version analysis
+with the upgrade_analysis tool present in the OCA/server-tools project.
+
+Once done, the analysis file contains correct information, and can be studied to
+write remaing pre / post migration scripts.
```
bguillot:
```suggestion
write renaming pre / post migration scripts.
```

bguillot (approved):
Small typo otherwise looks good

docsource/use_cases/module_renaming.rst
legalsylvain: rereading the sentence, I think my idea was to write "remaining". OK with that ?

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+
+**Problem**:
+
+- The v14 module remains in a ``to upgrade`` state.
+- The v15 module has not been installed.
+
+As a result, the models of the module have not been loaded, the related features
+are not available, and the database is in a unstable state.
+
+V12 table ``ir_module_module`` (With openupgrade)
```
remytms:
```suggestion
V15 table ``ir_module_module`` (With openupgrade)
```

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+"""""""""""""""""""""""""""""""""""""""""""""""""
+
+.. csv-table::
+   :header: "id", "create_date", "write_date"
```
remytms:
```suggestion
   :header: "id", "name", "state", "latest_version"
```

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    @openupgrade.migrate(use_env=False)
+    def migrate(cr, version):
+        openupgrade.update_module_names(cr, renamed_modules.items())
+
+See `Full v15 base pre migration file <https://github.com/OCA/OpenUpgrade/blob/97491cb7d9a8ed494a49cf1db9b7fc8852aac254/openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py#L73>`_.
+
+Note: The addition is done at the initialization of the OpenUpgrade Project,
+for each new version. You will not have to do this,
+if you are working on a version of OpenUpgrade under development.
```
remytms: I find that the note is not clear. What do you mean by a version of "OpenUpgrade under development" ?

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Coverage File
+"""""""""""""
+
+A coverage file is present in each version:
+
+* Until version 13, the file is present in this path:
+  ``odoo/openupgrade/doc/source/modulesXXX-YYY.py``.
```
remytms:
```suggestion
  ``odoo/openupgrade/doc/source/modulesXXX-YYY.rst``.
```

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+
+* from version 14.0 onwards, the file is present in this path:
+  ``docsource/modulesXXX-YYY.py``.
```
remytms:
```suggestion
  ``docsource/modulesXXX-YYY.rst``.
```

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Edit this file, and add the following text:
+
+.. code-block:: rst
+
+    +-----------------------------------+-------------------------------------------------+
+    | |new| crm_iap_enrich              | Renamed from crm_iap_lead_enrich                |
+    +-----------------------------------+-------------------------------------------------+
+    | |del| crm_iap_lead_enrich         | Renamed to crm_iap_enrich                       |
+    +-----------------------------------+-------------------------------------------------+
```
remytms: There is 3 columns in the file `modulesXXX-YYY.rst`. So I think it's better to propose an example with 3 columns, even if the last one is left blank. Or you can name the column so that there is no ambiguity in which column to put the rename information.

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+Generate / Update analysis file
+"""""""""""""""""""""""""""""""
+
+As long as the renaming has not been identified, the ``upgrade_analysis.txt`` generated
+at the initialization of the OpenUpgrade version is partially wrong, because openupgrade
+considers that the xml_id of the module are new, while they can come from the module of
+the previous version.
+
+For the modules that concern Odoo, you must therefore update the version analysis
+with the upgrade_analysis tool present in the OCA/server-tools project.
```
remytms:
```suggestion
For the Odoo modules that are concerned by the renaming, you must update the version
analysis with the upgrade_analysis tool present in the OCA/server-tools project.
```

Or any other formulation, because I did not understand the previous one. What is a "modules that concern Odoo" ? It may just be a mistake. ;)

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Once done, the analysis file contains correct information, and can be studied to
+write remaining pre / post migration scripts.
```
remytms:
```suggestion
Once the analysis is done with the updated ``apriory.py`` file, the analysis
file contains correct information, and can be studied to write remaining
pre / post migration scripts.
```

This way it seams clearer to me that the new `apriory.py` should be used to run the analysis. It can be obvious, but I find it good to mention it.

docsource/use_cases/module_renaming.rst
```diff
@@ -0,0 +1,169 @@
...
+  to refactor these custom modules by renaming them,
+  you can add the following lines at the end of this file:
+
+.. code-block:: python
+
+    renamed_modules.update({
+        "my_obsolete_module": "my_renamed_module",
+    })
+
+(no need in this case to propose a PR to the community)
```
remytms, line 172: Very good idea. I didn't use this trick and I constantly run with merge conflict when upgrading the `apriory.py` file. Thanks for the advice. :)

remytms (changes requested):
Some fixes needed. :)

docsource/use_cases/module_renaming.rst
legalsylvain: modules that concern Odoo = Odoo modules. 

suggestion applied.

docsource/use_cases/module_renaming.rst
legalsylvain: header and third column added. Thanks for your remark.

docsource/use_cases/module_renaming.rst
legalsylvain: I mean that when openupgrade maintainers are initializing the repo for a new version (16.0 for example), they populate the apriori.py file, so, after that, there is no need to add new entries for odoo modules. (theoretically!).
but thinking about that point, maybe it's not allways true. So I propose to remove this note. What do you think ?

remytms: I will then suggest this
```suggestion
Note: The addition is done at the initialization of the OpenUpgrade Project,
for each new version. So you will generally not have to do this, as it's only OpenUpgrade maintainers that initialize the OpenUpgrade Project.
```
Or something like this. :) I think that the comment is still useful. We just need to say that it may be not always the case.

etobella (approved):
LGTM

I think we can proceed with merge in this case.

/ocabot merge nobump

PR #3873, opened 2023-05-15
[DOC] Use case : field fast precomputation

legalsylvain:


bguillot (approved):

docsource/use_cases/field_fast_precomputation.rst
```diff
@@ -0,0 +1,149 @@
...
+    from openupgradelib import openupgrade
+
+    def compute_product_template_attribute_line_value_count(env):
+        openupgrade.logged_query(
+            env.cr,
+            """
+            ALTER TABLE product_template_attribute_line
+            ADD IF NOT EXISTS COLUMN value_count integer
+            """,
+        )
```
remytms: Isn't it easier to use the [`add_fields()`](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2739) function from `openupgradelib` instead of adding the column by a custom SQL request ?

remytms (approved):
Just a question, but other way, it good for me. :)

docsource/use_cases/field_fast_precomputation.rst
legalsylvain: hehe ! you're right !

legalsylvain: done.

PR #3874, opened 2023-05-15
[WIP][DOC] Use Case : add module splitted

legalsylvain:
Hi all. (@pedrobaeza, @MiquelRForgeFlow, @hbrunn)

nothing to review for the time being, but a question : 

In the list of the "TODO Use Cases", there is an item named "A module has been split". ([RFQ Detail](https://drive.google.com/file/d/1sRis_GoZiVUOfdx-IJA8HMMDgXvhYZzV/view)).

I found some examples : 
- ``purchase``, splitted in ``purchase_stock`` (between v11 and v12), to remove dependency to stock module
- ``sale``, splitted in ``sale`` and ``sale_management`` (between v10 and v11) for commercial reason.

However, I don't feel that there is anything special to achieve in this case. In all the examples I found, the "new" modules are auto_install=True, so there is no need to "force" the installation. 
There may be some xml id to rename, but this use case is already described in this page: https://oca.github.io/OpenUpgrade/use_cases/xml_id_renaming.html

What do you think ? Do you see other examples, where there is something specific to do ? 

thanks for your insights !

### DIFF
--- a/docsource/development.rst
+++ b/docsource/development.rst
@@ -55,6 +55,7 @@ Learn from typical Use cases
 
    use_cases/field_renaming
    use_cases/xml_id_renaming
+   use_cases/module_splitted
 
 
 Learn from existing migration scrips


pedrobaeza:
There was previously the need of calling `update_module_moved_fields` and `update_module_moved_models` for these cases, but now the XML-IDs are handled by the standard, so indeed there's a nothing to do here. Maybe this can be considered in `upgrade_analysis` module for generating a better log without all the DEL models and fields that have been moved to the split module.

legalsylvain:
Thanks. So closing.

PR #3881, opened 2023-05-22
[14.0][OU-ADD] apriori: l10n_nl_tax_invoice_basis merged into l10n_nl_tax_statement

astirpe:
Module `l10n_nl_tax_invoice_basis` was merged into `l10n_nl_tax_statement`, as mentioned in the description of this PR https://github.com/OCA/l10n-netherlands/pull/298

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -89,6 +89,8 @@ merged_modules = {
     # OCA/e-commerce
     "website_sale_product_style_badge": "website_sale",
     "website_snippet_carousel_product": "website_sale",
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):
Are you sure that no additional migration scripts are needed? (I mean, the `l10n_nl_date_invoice` field is lost...)

astirpe:
Field `l10n_nl_date_invoice` was a computed stored field used for speedup and ease the filtering on move lines, but in V14 was not needed anymore since we can easily make use of `move_id.invoice_date`. I don't think we need any migration script.

PR #3882, opened 2023-05-22
[13.0][OU-FIX] stock: mail.template terms

chienandalu:
The mail.template `mail_template_data_delivery_confirmation` was formerly in the module `delivery`. Along with the regular template refreshing we need to set the translation terms module to the `stock` module so we can fetch them all and delete them properly.

cc @Tecnativa TT43496

ping @pedrobaeza

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -421,6 +421,15 @@ def migrate(env, version):
     if openupgrade.table_exists(env.cr, 'delivery_carrier'):
         openupgrade.load_data(
             env.cr, "stock", "migrations/13.0.1.1/noupdate_changes2.xml")
+        # This mail.template came from the module delivery to stock, so the translated
+        # terms will have the old module set. We need to change it so we can delete
+        # the translations properly
+        template = env.ref("stock.mail_template_data_delivery_confirmation")
+        env["ir.translation"].search(
+            [
+                ("name", "ilike", "mail.template,"), ("res_id", "=", template.id)
+            ]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
That's supposed to be done in the following lines:

https://github.com/OCA/OpenUpgrade/pull/3882/files#diff-23ca17a5c4c2204907213a7d5651c694fb2c55deb5e58a45cab51270acd76ee4R433

chienandalu:
The problem is that the resulting sql expect the terms to be in the specified module, which is the issue here

pedrobaeza:
Then the problem is on the openupgradelib renaming method (`rename_xml_ids`?), that should handle this.

pedrobaeza:
cc @chienandalu

MiquelRForgeFlow (approved):
> Then the problem is on the openupgradelib renaming method (rename_xmlids?), that should handle this.

@pedrobaeza the `rename_xmlids` method only renames the xmlid, but doesn't rename their related translations. So, in pre-migration, the `delivery.mail_template_data_delivery_confirmation` is renamed to `stock.mail_template_data_delivery_confirmation`, but then this mail template translations are still in `delivery`. So, `delete_record_translations` will delete the translations in `stock`, but the translations in `delivery` will remain (but unused). So this PR deletes these unused deprecated translations by moving them to `stock` before calling `delete_record_translations`.

pedrobaeza:
I insist this is something that should be handled in `rename_xmlids` method, as it's a side effect of renaming the XML-ID.

MiquelRForgeFlow:
Will be handled with https://github.com/OCA/openupgradelib/pull/402.

PR #3883, opened 2023-05-22
[14.0][MIG] apriori: hr_period -> hr_payroll_period

MiquelRForgeFlow:
The migration was done in https://github.com/OCA/payroll/pull/89.

Just FYI: user @nimarosa took `hr_period` module from OCA/hr (v12), migrated it to v14 in OCA/payroll, renamed it, didn't conserve commit history and auto-merged without approvals.

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -50,6 +50,8 @@ renamed_modules = {
     "edi_xml": "edi_xml_oca",
     # OCA/hr-holidays
     "hr_leave_hour": "hr_leave_custom_hour_interval",
+    # OCA/hr -> OCA/payroll:
+    "hr_period": "hr_payroll_period",
     # OCA/l10n-spain
     "l10n_es_account_bank_statement_import_n43": "l10n_es_account_statement_import_n43",
     # OCA/server-tools


AaronHForgeFlow (approved):

nimarosa:
Hello @MiquelRForgeFlow as stated in this thread https://github.com/OCA/hr/pull/1124 we asked the creator of the PR to move it to payroll. As we don't received response I moved it myself. 

hr_period is more a payroll module not a strictly hr one. Because it affects payroll it seems to me that it should be in payroll. It was renamed to payroll_period to not conflict with existing installs of hr_period. 

I have trouble migrating commit history from one repo to another, so I just moved the module. At that time, when I tried to copy commit history I ended up with a lot of errors and incomplete commits. If you want you can try to submit a PR to OCA/payroll migrating the commit history. 

As stated the reason for that module to be in payroll is that is a payroll exclusive functionality which is not used in another hr modules. Also, payroll module have changed a lot so for me it makes sense to have it in that repository to continue adapting it to the current payroll module. 

It was merged without approvals because i just moved the module didn't modify it. It was an administrative task. Only fixed some views that changed in payroll for the module to work and install properly.

nimarosa (approved):

PR #3887, opened 2023-05-23
[13.0][FIX] pin CI ubuntu to ubuntu-22.04

hbrunn:


legalsylvain:
maybe better pin to ubuntu-20.04.

The change from 20.04 to 22.04 has been done quite recently https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

"This change will be rolled out over a period of 8 weeks beginning on October 1, 2022."

However, the last recent commit on 13.0 worked...
https://github.com/OCA/OpenUpgrade/actions?query=branch%3A13.0++

don't know.

PR #3889, opened 2023-05-24
[9.0] Fix `Many2one.convert_to_write` in presence of None

sbidoul:
I got a situation where `Many2one.convert_to_write` got called with a `None` value.

The stack trace looked like this:

```
  File "/home/work/venv-08to09/src/odoo/openerp/service/server.py", line 894, in preload_registries
  File "/home/work/venv-08to09/src/odoo/openerp/modules/registry.py", line 390, in new
  File "/home/work/venv-08to09/src/odoo/openerp/modules/loading.py", line 446, in load_modules
  File "/home/work/venv-08to09/src/odoo/openerp/modules/loading.py", line 331, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check, upg_registry=upg_registry
  File "/home/work/venv-08to09/src/odoo/openerp/modules/loading.py", line 178, in load_module_graph
    init_module_models(cr, package.name, models)
  File "/home/work/venv-08to09/src/odoo/openerp/modules/module.py", line 300, in init_module_models
    result = obj._auto_init(cr, {'module': module_name})
  File "/home/work/venv-08to09/src/odoo/openerp/api.py", line 250, in wrapper
  File "/home/work/venv-08to09/src/odoo/openerp/models.py", line 2640, in _auto_init
    self._set_default_value_on_column(cr, k, context=context)
  File "/home/work/venv-08to09/src/odoo/openerp/api.py", line 250, in wrapper
  File "/home/work/venv-08to09/src/odoo/openerp/models.py", line 2450, in _set_default_value_on_column
    default = default(self, cr, SUPERUSER_ID, context)
  File "/home/work/venv-08to09/src/odoo/openerp/api.py", line 250, in wrapper
  File "/home/work/venv-08to09/src/odoo/openerp/api.py", line 354, in old_api
    result = method(recs, *args, **kwargs)
  File "/home/work/venv-08to09/src/odoo/openerp/fields.py", line 83, in <lambda>
    return api.model(lambda model: field.convert_to_write(value(model)))
  File "/home/work/venv-08to09/src/odoo/openerp/fields.py", line 1726, in convert_to_write
    return value.id
AttributeError: 'NoneType' object has no attribute 'id'
```

And the None value came from https://github.com/OCA/OpenUpgrade/blob/30b6c228c83024a8e55da4f210ff3e7d9b66c5d2/addons/account/models/account_invoice.py#L1103-L1109

apparently because OpenUpgrade had no `journal_id` in context.

### DIFF
--- a/openerp/fields.py
+++ b/openerp/fields.py
@@ -1723,7 +1723,7 @@ class Many2one(_Relational):
             return value.id
 
     def convert_to_write(self, value):
-        return value.id
+        return value.id if value else False
 
     def convert_to_export(self, value, env):
         return value.name_get()[0][1] if value else ''


legalsylvain:
hi @sbidoul. I don't understand when the function ``_default_account`` is called in an Openupgrade context. I mean, 
``_default_account`` is called when odoo create new invoices, but that is not done in an openupgrade process. And the field exist in V8, so no need to initialize it.
On wich invoices the default function is called ? 

> apparently because OpenUpgrade had no journal_id in context.

Well, if there is no journal_id in the context, I guess ``if self._context.get('journal_id')`` will return False so lines 1108 and 1109 will not be executed.

At this step, I don't think that's this patch to the Odoo core is the correct way to fix your problem.
Not sure, but I guess something is wrong, somewhere else.

sbidoul:
Investigating further I now see `_set_default_value_on_column` got called in a context where it should not  have been called, because I had dropped NOT NULL constraints before running the migration.

https://github.com/OCA/OpenUpgrade/blob/30b6c228c83024a8e55da4f210ff3e7d9b66c5d2/openerp/models.py#L2637-L2640

legalsylvain:
> Investigating further I now see _set_default_value_on_column got called in a context where it should not have been called, because I had dropped NOT NULL constraints before running the migration.

That's weird. I don't understand why odoo call ``_set_default_value_on_column``. I mean, if there isn't null values, it should not have been called. But well... Maybe I missed something.

Anyway. What is the reason to have dropped the constraint ?

MiquelRForgeFlow (changes requested):
> because I had dropped NOT NULL constraints before running the migration

We don't like doing this kind of changes to odoo code just because you decided to do the migration with dropping not null constrains. Sorry.

PR #3899, opened 2023-05-31
[16.0][OU-ADD] calendar

remytms:
No DB changes.

Import changed noupdate mail templates and removed translations for them.


openupgrade_scripts/scripts/calendar/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+    "calendar_template_meeting_invitation",
+    "calendar_template_meeting_reminder",
+    "calendar_template_meeting_update",
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "calendar", "16.0.1.1/noupdate_changes.xml")
+    openupgrade.delete_record_translations(env.cr, "calendar", _translations_to_delete)
```
pedrobaeza, line 17: The method `delete_record_translations` is no longer valid for v16, as it deletes rows in `ir_translation` table:

https://github.com/OCA/openupgradelib/blob/f87f553aca1e8fb8447bc152473518c6b070ed1a/openupgradelib/openupgrade.py#L2610

which is no longer used in this version. It doesn't fail, because this table is brought from v15.

so `openupgradelib` must be updated to handle this new situation. Options:

- This method is still used, but putting special code for v16 that rewrites the field removing other languages.
- With investigation, if method `load_data` can be improved, including a new keyword parameter `overwrite_translation` or similar, that makes this job at the same time as the XML data is loaded.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/calendar/16.0.1.1/post-migration.py
remytms: Thank you for the review. I will take a look on how to improve `delete_record_translations()`.

MiquelRForgeFlow: Handled in https://github.com/OCA/openupgradelib/pull/328.

remytms:
This is not blocked by dependencies anymore, since `base` and `mail` has been merged. :)

MiquelRForgeFlow (approved):

PR #3904, opened 2023-06-01
[IMP] migration_details: adjust configuration

AungKoKoLin1997:
@qrtl

### DIFF
--- a/docsource/migration_details.rst
+++ b/docsource/migration_details.rst
@@ -54,10 +54,10 @@ backup of your live database before you start this process!
 
 Edit the configuration files and command line parameters to point to the
 database you are going to upgrade. The recommended command line parameters are the
-``--update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
+``--upgrade-path=<path_to_openupgrade>/openupgrade_scripts/scripts --update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
 
 For versions earlier than 14.0 that are running the OpenUpgrade fork rather
-than Odoo itself, you do not pass the `load` parameter.
+than Odoo itself, you do not pass the `load` parameter nor `upgrade-path` parameters.
 
 Configuration options
 .....................


docsource/migration_details.rst
```diff
@@ -54,10 +54,14 @@ backup of your live database before you start this process!
...
 database you are going to upgrade. The recommended command line parameters are the
-``--update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
+``--upgrade-path=/home/OpenUpgrade/openupgrade_scripts/scripts --update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
 
 For versions earlier than 14.0 that are running the OpenUpgrade fork rather
 than Odoo itself, you do not pass the `load` parameter.
+Firstly, modify the addons_path in the configuration to use OpenUpgrade addons. The recommended command-line parameters to use are 
+``--config=yourconf --update all --stop-after-init.``
+
+Once the configuration setup is complete, execute odoo-bin from OpenUpgrade, not from the standard Odoo package.
```
pedrobaeza: This doesn't seem correct

docsource/migration_details.rst
```diff
@@ -54,10 +54,14 @@ backup of your live database before you start this process!
 
 Edit the configuration files and command line parameters to point to the
 database you are going to upgrade. The recommended command line parameters are the
-``--update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
+``--upgrade-path=/home/OpenUpgrade/openupgrade_scripts/scripts --update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
```
pedrobaeza:
```suggestion
``--upgrade-path=<path_to_openupgrade>/openupgrade_scripts/scripts --update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
```

docsource/migration_details.rst
AungKoKoLin1997: I thought for the versions before odoo14, we need to use openupgrade addons and we should use openupgrade odoo-bin. My understanding was wrong?

pedrobaeza: Yes, but the previous sentence is already saying that mostly. Compact everything in one sentence for avoiding misunderstandings.

docsource/migration_details.rst
```diff
@@ -54,7 +54,7 @@ backup of your live database before you start this process!
 
 Edit the configuration files and command line parameters to point to the
 database you are going to upgrade. The recommended command line parameters are the
-``--update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
+``--upgrade-path=<path_to_openupgrade>/openupgrade_scripts/scripts --update all --stop-after-init --load=base,web,openupgrade_framework`` flags.
 
 For versions earlier than 14.0 that are running the OpenUpgrade fork rather
 than Odoo itself, you do not pass the `load` parameter.
```
pedrobaeza:
```suggestion
than Odoo itself, you do not pass the `load` nor `upgrade-path` parameters.
```

MiquelRForgeFlow (approved):
LGTM

PR #3906, opened 2023-06-03
[14.0][IMP] ir_attachment: binary_conversion_for_db_attachments

AungKoKoLin1997:
In versions of Odoo prior to 14, attachments stored in the database were kept as base64-encoded data. However, starting from Odoo 14, these attachments are maintained as binary data, and only encoded to base64 within the _compute_datas method for the datas field. Therefore, when migrating from version 13 to 14 with attachments stored in the database, these attachments may not display properly within their respective records.

@qrtl

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
@@ -1,7 +1,12 @@
 # Copyright 2020 Odoo Community Association (OCA)
 # Copyright 2020 Opener B.V. <stefan@opener.am>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import base64
+import logging
+
 from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
 
 
 def fix_module_category_parent_id(env):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you please point to the relevant pieces of code in Odoo v13 and v14 where it's shown this change?

AungKoKoLin1997:
> Can you please point to the relevant pieces of code in Odoo v13 and v14 where it's shown this change?

In odoo13, [this line](https://github.com/odoo/odoo/blob/e6d1a7531dc02601eca7d704134990896ed493d4/odoo/addons/base/models/ir_attachment.py#L534) will store the `db_datas` with base64-encoded data. In odoo14, [this line](https://github.com/odoo/odoo/blob/652fb88281f4d3659c3203716dc3b75bd4789089/odoo/addons/base/models/ir_attachment.py#L611-L614) use decoded data to store.

pedrobaeza:
Thanks for the pointer. Correct me if I'm wrong, this is only preprocessing the dictionary in the `create` method, but that's independent on how this actual serialize in the DB. And this also only seen to be a problem if you store attachments in the DB, which is not the default since a lot. Is that the case?

AungKoKoLin1997:
Yes. You are right.

pedrobaeza:
In which thing am I right?

AungKoKoLin1997:
I mean this is preprocessing  on `create` method. It will not depend on database and use the same column type `bytea` that can be solve bytes and base64-encoded data in v13 and v14.

pedrobaeza:
Yeah, but I think this is only to adapt data in ORM layer, but on DB storage, the result is the same. I let others to review this, as I don't know this part very deep.

MiquelRForgeFlow (approved):
Makes sense to me. The change was done in https://github.com/odoo/odoo/commit/172722c4d229063ba964095df5bcfadc25ba65c9 and obviously the OU analysis didn't catch it.

AungKoKoLin1997:
@OpenUpgrade-maintainers
Is it possible to review this PR?

PR #3908, opened 2023-06-06
[MIG] sale_order_margin_percent: Migration to 16.0

hugo-cordoba:
We were reviewing the migration of the 'sale_order_margin_percent' module, and we have seen that its functionality is already incorporated in the 'sale_margin' module. There is no need to migrate that module.

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -50,6 +50,8 @@ merged_modules = {
     "account_move_force_removal": "account",
     # OCA/account-invoice-reporting
     "account_invoice_report_due_list": "account",
+    # OCA/margin-analysis
+    "sale_order_margin_percent": "sale_margin",
     # OCA/purchase-workflow
     "product_form_purchase_link": "purchase",
 }


legalsylvain (changes requested):
Hi @HugoCordobaLeal. Thanks for your contribution. 
However, this merge has been already marked as done in OpenUpgrade V14.
https://github.com/OCA/OpenUpgrade/pull/3217/files
So no need to add it again in V16. (if i missed something, please add precisions).

kind regards.

hugo-cordoba:
Hi, I didn't see it in odoo 16 added, that's why I included it, why shouldn't it be included? @legalsylvain

legalsylvain:
> Hi, I didn't see it in odoo 16 added, that's why I included it, why shouldn't it be included? @legalsylvain

sorry I don't understand your question.

- sale_order_margin_percent exists from 10 to 13 release. (see : https://apps.odoo.com/apps/modules/13.0/sale_order_margin_percent/)
- when you migrate the database from 13 to 14, with Openupgrade 14 the rename is done
- as a result, there is nothing to do in Openupgrade 16.

regards.

hugo-cordoba:
Oh, thanks you very much, I didn't quite understand how it worked. Now everything is much clearer to me.

Thanks again. Best regards!!

PR #3911, opened 2023-06-07
[16.0][OU-ADD] hr

remytms:
Move elements from `hr_contract` to `hr`.
Replace M2M relation between hr.plan and hr.plan.activity.type to O2M and warn about data loss.
Import noupdate_changes.

Tested with some data in hr.plan and hr.plan.activity.type.


openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+            FROM %(source_relation_table)s
+            GROUP BY %(relation_comodel_field)s
+            HAVING COUNT(*) > 1
+        )
+        """
+        % {
+            "source_relation_table": source_relation_table,
+            "relation_comodel_field": relation_comodel_field,
+        },
+    )
```
remytms: Using the escaping method to give parameters to the query, leads to a error in the syntax of the query.

```python
openupgrade.logged_querry(
    cr,
    "SELECT * FROM %(source_relation_table)s;",
    {"source_relation_table": "hr_plan_hr_plan_activity_type_rel"},
)
```
Give an error with the following query:
```sql
SELECT * FROM 'hr_plan_hr_plan_activity_type';
```
```
ERROR:  syntax error at or near "'hr_plan_hr_plan_activity_type_rel'"
LINE 1: SELECT * FROM 'hr_plan_hr_plan_activity_type_rel';
```

So I formatted the query string with string format mechanism, which is in general a bad idea.

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+    for res in cr.fetchall():
+        _logger.error(
+            "hr.plan.activity.type(%s,) is linked to several hr.plan. "
+            "hr.plan.activity.type can only be linked to one hr.plan. "
+            "Fix these data before migrating to avoid data loss.",
+            res[0],
+        )
+
+
+def m2m_to_o2m(
```
remytms, line 47: I wonder if such a function is generic enough to go into `openupgradelib` ?

pedrobaeza: Well, it's not a usual case, but if it can be parametrized, then a good host is `openupgradelib`.

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,29 @@
...
+    (
+        "hr_contract.access_hr_contract_type_manager",
+        "hr.access_hr_contract_type_manager",
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.update_module_moved_models(
```
marielejeune: Do you know what's the behavior of the method if the script is run on a DB where `hr` module was installed, but not `hr_contract`?

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,59 @@
...
+hr           / hr.contract.type         / name (char)                   : previously in module hr_contract
+# DONE: moved from hr_contract to hr
+hr           / hr.contract.type         / sequence (integer)            : NEW
+# NOTHING TO DO
+hr           / hr.department            / master_department_id (many2one): NEW relation: hr.department, isfunction: function, stored
+hr           / hr.department            / parent_path (char)            : NEW
+hr           / hr.department            / plan_ids (one2many)           : NEW relation: hr.plan
+# NOTHING TO DO
+hr           / hr.employee              / mobile_phone (char)           : now a function
+hr           / hr.employee              / work_contact_id (many2one)    : NEW relation: res.partner
```
marielejeune: I think we should call `_inverse_work_ontact_details` to create the `work_contact_id` records.
See https://github.com/odoo/odoo/blob/5ac58ebf983c1c02019253c0430fde3502e13e7c/addons/hr/models/hr_employee_base.py#L169

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,59 @@
...
+hr           / hr.employee              / work_email (char)             : now a function
+# NOTHING TO DO
+hr           / hr.job                   / active (boolean)              : NEW hasdefault: default
+hr           / hr.job                   / contract_type_id (many2one)   : NEW relation: hr.contract.type
+hr           / hr.job                   / state (selection)             : DEL required, selection_keys: ['open', 'recruit']
+# NOTHING TO DO: hr.job does not work with sate anymore.
```
marielejeune: `sate` -> `state`

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
marielejeune: I think using `AsIs` from `psycopg2.extensions` can help you. If I am not mistaking:

from pyscopg2.extensions import AsIs

openupgrade.logged_query( cr, "SELECT * FROM %(source_relation_table)s;", {"source_relation_table": AsIs("hr_plan_hr_plan_activity_type_rel")}, )

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    m2m_to_o2m(
+        env,
+        "hr.plan",
+        "plan_activity_type_ids",
+        "hr_plan_hr_plan_activity_type_rel",
+        "hr_plan_id",
+        "hr_plan_activity_type_id",
+    )
+    drop_table(env.cr, "hr_plan_hr_plan_activity_type_rel")
```
marielejeune: Should we really drop the data?
It is always risky...

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: I will try, thanks for the advice. :)

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
remytms: Yes, in that case, the function has no effect, because the model `hr.contract.type` does not exist, neither the module `hr_contract` in the tables: `ir_model_data`, `ir_model_constraint` and `ir_model_relation`.

Therefore, the model `hr.contract.type` will be created during the update of the `hr` module, after the pre-migration script.

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: I'm not sure either. @pedrobaeza or @legalsylvain do you have an opinion about the removal of a relation table for a Many2many field that is now a Many2one ?

Obviously, data will be lost in some cases. These cases are reported in the log by the `warn_about_dataloss()` function. But indeed, no way to retrieve the data after the migration. In my opinion, cleaning data should be done before migrating.

pedrobaeza: One of the principles of OpenUpgrade is to preserve the previous information at maximum (unless it supposes a problem. Example: quant merging in v11 was a "destructive" operation, well more a transformative operation, but it was needed). In this case, if it's a problem to have the table with that name, rename it to `None` with the openupgradelib method (which put it an special name). If not, just leave it as is. `database_cleanup` will remove on your command later.

pedrobaeza: You can also lift constraints with the 2 corresponding openupgradelib methods for avoiding relational problems. Check for examples in previous versions.

remytms: @pedrobaeza Thank you. Then I will go that way. :)

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: @marielejeune Thank you for the advice about `AsIs`. It works well. :)

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
remytms: Here is the details of a run on a database with only `hr` installed:
```
2023-06-21 08:50:50,691 1 INFO odooce_hr OpenUpgrade: Moving model hr.contract.type from module 'hr_contract' to module 'hr' 
2023-06-21 08:50:50,691 1 DEBUG odooce_hr OpenUpgrade: 0 rows affected after 0:00:00.000329 running UPDATE ir_model_data SET module='hr' WHERE model = 'hr.contract.type' AND module = 'hr_contract' 
2023-06-21 08:50:50,692 1 DEBUG odooce_hr OpenUpgrade: 0 rows affected after 0:00:00.000269 running UPDATE ir_model_data SET module='hr' WHERE name='model_' || 'hr_contract_type' AND model = 'ir.model' AND module = 'hr_contract' RETURNING 
res_id 
2023-06-21 08:50:50,692 1 DEBUG odooce_hr OpenUpgrade: 0 rows affected after 0:00:00.000396 running UPDATE ir_model_data imd
        SET module = 'hr'
        FROM ir_model_fields imf
        WHERE imd.model = 'ir.model.fields'
            AND imd.name = 'field_' || 'hr_contract_type' || '__' || imf.name
            AND imf.model = 'hr.contract.type' AND imd.module = 'hr_contract' 
2023-06-21 08:50:50,692 1 INFO odooce_hr OpenUpgrade: Moving fields __last_update, _order, display_name, name in model hr.contract.type from module 'hr_contract' to module 'hr' 
2023-06-21 08:50:50,693 1 DEBUG odooce_hr OpenUpgrade: 0 rows affected after 0:00:00.000486 running 
        UPDATE ir_model_data imd
        SET module = 'hr'
        FROM ir_model_fields imf
        WHERE
            imf.model = 'hr.contract.type' AND
            imf.name IN ('__last_update', '_order', 'display_name', 'name') AND
            imd.module = 'hr_contract' AND
            imd.model = 'ir.model.fields' AND
            imd.res_id = imf.id AND
            imd.id NOT IN (
               SELECT id FROM ir_model_data WHERE module = 'hr'
            )
         
2023-06-21 08:50:50,693 1 INFO odooce_hr OpenUpgrade: XMLID hr_contract.access_hr_contract_type_manager not found when renaming to hr.access_hr_contract_type_manager 
```

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: Done.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Well spotted. Indeed, we need to compute the now `work_contact_id` and filling it with information that already exist in the database.

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,186 @@
...
+                    attachment_values[7],
+                    attachment_values[8],
+                    attachment_values[9],
+                    attachment_values[10],
+                    attachment_values[11],
+                    attachment_values[12],
+                    attachment_values[13],
+                    attachment_values[14],
+                ),
+            )
```
remytms: This part is new and not tested yet. @robinkeunen can you have a look at this. Is there a way to put all that into one SQL request or any better way to do it ?

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
nguyenvietlam0640: I think we shouldn't create work_contact_id for old data, because this field not requires to avoid creating so much junk contact. Indeed, employee may already had contact by create manually

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,187 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def warn_about_dataloss(cr, source_relation_table, relation_comodel_field):
+    """Warn user about data loss when migrating data from many2many to
+    many2one.
+
+    :param source_relation_table: The many2many relation table
+    of the model that will on the 'one' side of the relation
```
carmenbianca: there's a missing verb after 'will', I think.

(review en passant, i was actually looking for some example code, thank you rémy)

carmenbianca (changes requested):

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: @nguyenvietlam0640 But it seams to me that it will be created anyway when the ORM will call the `inverse_work_contact_details()` to fill field like `work_mail` and `mobile_phone`. Isn't it ?

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: fixed

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: Robin has no time to suggest better way of doing it. So any though are welcome. :)

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
nguyenvietlam0640: @remytms I just tested it. And the `inverse_work_contact_details()` method **wasn't called** by ORM when migrate till I had changed the `mobile_phone` or `work_email` of employee and it was called to create a new contact. In addition,  The way you created `work_contact_id` by SQL, the field `active`  of `res_partner` will auto set to `false`  which will set to `true` when handle by ORM.

remytms: @nguyenvietlam0640 You'r right. I tested also to be sure that previous data for `work_email` and `mobile_phone` was not lost when performing migration.

Thanks a lot for your tests and pointing it out that my function for creating contact was not  correct and not useful. :)

I will drop it.

remytms: @marielejeune Do you still think it's a good idea to run the `_inverse_work_contact_details()` during the migration process ?

marielejeune: Not sure anymore. Maybe let's wait for one more advice?

remytms: @OCA/openupgrade-maintainers Do you have an advice on this topic ?

To summary:

`hr.employee` are now linked to a `res.partner` to store work information of the employee. Migration works well without doing anything. Meaning that the `res.partner` are not created during migration, but each time you modify an employee work information (email, phone, etc), a compute function is triggered and create a new partner with these informations.

Should we leave the migration like this or should we create all the new `res.partner` during the migration process ?

@nguyenvietlam0640 is in favor of doing nothing because it will create a lot of "junk" contact, and maybe these contact already exists but are not linked to the employee in the database.

@marielejeune and myself were in favor of creating all the partner during the migration process, because at the end it will be created.

I think that the three options are:

1. Do nothing, partner will be created on the fly or outside of the migration script.
2. Run the compute function `_inverse_work_contact_detail()` on each `hr.employee` to create all the partner during the migration process.
3. Create all the partner via SQL requests instead of by running the compute function during the migration process.

Any advice to decide are welcome. :)

carmenbianca: Doing nothing seems like the worst option to me.

- It is not transparent to a user of the software that they trigger a partner creation by changing the phone or e-mail of an employee. That's _terrible_ UX.
  - If they want to do it, they won't intuitively know how.
  - If they don't want to do it, they may never find out why they're slowly getting these additional partner records.
    - And if they consider these records 'junk', then it's much preferable to be able to deal with all junk records _at once_ instead of slowly having to deal with them over time.
- The module kind of _assumes_ that a partner exists, even if it's not strictly mandatory. It's best to go along with those assumptions.
- In the future, the partner field may become required/mandatory. We'd end up having to deal with corner cases anyway, but data integrity/quality-wise, it seems better to me to bite the bullet earlier on as many records as possible.
- An Odoo migration is the best time for data clean-up. Data records will change; dealing with the new 'junk' partner records is just one more thing on the todo list during the migration process.

In any case, this side effect of migration should be clearly documented.

Writing an SQL query is probably preferable. I may be getting this wrong, but it is my reading that using `_inverse_work_contact_detail()` would practically work, but might break if the user has another module installed that implements that method.

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,32 @@
...
+        env.cr, "hr.contract.type", "hr_contract", "hr"
+    )
+    openupgrade.update_module_moved_fields(
+        env.cr, "hr.contract.type", _moved_fields, "hr_contract", "hr"
+    )
+    openupgrade.rename_xmlids(env.cr, _xmlid_renames)
+    # Backup Many2many relation between hr.plan and hr.plan.activity.type
+    openupgrade.remove_tables_fks(env.cr, ["hr_plan_hr_plan_activity_type_rel"])
+    openupgrade.rename_tables(env.cr, [("hr_plan_hr_plan_activity_type_rel", None)])
```
gdgellatly: I have tested this using a git-aggregator file and it fails here.

```
2023-08-31 10:43:42,391 1 DEBUG mig OpenUpgrade: -1 rows affected after 0:00:00.000516 running ALTER TABLE "hr_plan_hr_plan_activity_type_rel" DROP CONSTRAINT "hr_plan_hr_plan_activity_type_rel_hr_plan_activity_type_id_fkey" 
2023-08-31 10:43:42,391 1 DEBUG mig OpenUpgrade: -1 rows affected after 0:00:00.000317 running ALTER TABLE "hr_plan_hr_plan_activity_type_rel" DROP CONSTRAINT "hr_plan_hr_plan_activity_type_rel_hr_plan_id_fkey" 
2023-08-31 10:43:42,391 1 INFO mig OpenUpgrade: table hr_plan_hr_plan_activity_type_rel: renaming to openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel 
2023-08-31 10:43:42,392 1 ERROR mig odoo.sql_db: bad query: ALTER INDEX "hr_plan_hr_plan_activity_type_rel_hr_plan_id_idx" RENAME TO "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_plan_id_idx"
ERROR: relation "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl" already exists
 
2023-08-31 10:43:42,393 1 ERROR mig OpenUpgrade: hr: error in migration script /odoo/src/openupgrade/openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py: relation "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl" already exists
 
2023-08-31 10:43:42,393 1 ERROR mig OpenUpgrade: relation "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl" already exists
 
  File "/odoo/.local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 2271, in wrapped_function
    func(
  File "/odoo/src/openupgrade/openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py", line 32, in migrate
    openupgrade.rename_tables(env.cr, [("hr_plan_hr_plan_activity_type_rel", None)])
  File "/odoo/.local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 838, in rename_tables
    cr.execute(
  File "/odoo/src/odoo/odoo/sql_db.py", line 321, in execute
psycopg2.errors.DuplicateTable: relation "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl" already exists

2023-08-31 10:43:42,419 1 WARNING mig odoo.modules.loading: Transient module states were reset 
2023-08-31 10:43:42,440 1 ERROR mig odoo.modules.registry: Failed to load registry 
  File "/odoo/src/odoo/odoo/modules/registry.py", line 90, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/odoo/src/odoo/odoo/modules/loading.py", line 484, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/odoo/src/odoo/odoo/modules/loading.py", line 372, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/odoo/src/odoo/odoo/modules/loading.py", line 183, in load_module_graph
    migrations.migrate_module(package, 'pre')
  File "/odoo/src/openupgrade/openupgrade_framework/odoo_patch/odoo/modules/migration.py", line 18, in migrate_module
    MigrationManager.migrate_module._original_method(self, pkg, stage)
  File "/odoo/src/odoo/odoo/modules/migration.py", line 189, in migrate_module
  File "/odoo/.local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 2271, in wrapped_function
    func(
  File "/odoo/src/openupgrade/openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py", line 32, in migrate
    openupgrade.rename_tables(env.cr, [("hr_plan_hr_plan_activity_type_rel", None)])
  File "/odoo/.local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 838, in rename_tables
    cr.execute(
  File "/odoo/src/odoo/odoo/sql_db.py", line 321, in execute
psycopg2.errors.DuplicateTable: relation "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl" already exists

2023-08-31 10:43:42,440 1 CRITICAL mig odoo.service.server: Failed to initialize database `mig`. 
  File "/odoo/src/odoo/odoo/service/server.py", line 1299, in preload_registries
  File "<decorator-gen-14>", line 2, in new
  File "/odoo/src/odoo/odoo/tools/func.py", line 87, in locked
    return func(inst, *args, **kwargs)
  File "/odoo/src/odoo/odoo/modules/registry.py", line 90, in new
    odoo.modules.load_modules(registry, force_demo, status, update_module)
  File "/odoo/src/odoo/odoo/modules/loading.py", line 484, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/odoo/src/odoo/odoo/modules/loading.py", line 372, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/odoo/src/odoo/odoo/modules/loading.py", line 183, in load_module_graph
    migrations.migrate_module(package, 'pre')
  File "/odoo/src/openupgrade/openupgrade_framework/odoo_patch/odoo/modules/migration.py", line 18, in migrate_module
    MigrationManager.migrate_module._original_method(self, pkg, stage)
  File "/odoo/src/odoo/odoo/modules/migration.py", line 189, in migrate_module
  File "/odoo/.local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 2271, in wrapped_function
    func(
  File "/odoo/src/openupgrade/openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py", line 32, in migrate
    openupgrade.rename_tables(env.cr, [("hr_plan_hr_plan_activity_type_rel", None)])
  File "/odoo/.local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 838, in rename_tables
    cr.execute(
  File "/odoo/src/odoo/odoo/sql_db.py", line 321, in execute
psycopg2.errors.DuplicateTable: relation "openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl" already exists
```

gdgellatly: Hmmm, if I comment the line it says the table doesn't exist, investigating

carmenbianca: I suspect this is an issue local to your computer.

Heads-up that the referenced code is being moved to https://github.com/OCA/openupgradelib/pull/343

remi-filament: Hi @remytms @carmenbianca thanks for your work !
I face the same issue here...
Found out that replacing the None by a dummy name on line 32 :
`openupgrade.rename_tables(env.cr, [("hr_plan_hr_plan_activity_type_rel", 'test_hr_plan_rel')])`

(and then replacing the `openupgrade.get_legacy_name` occurences in post_migration.py with my dummy name)

It works properly.
I am wondering if there is not a limit with the number of characters a database table name can support ? (I am using Postgres 13 here)

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: I discussed with @legalsylvain and we test the default behaviour in 16.0 regarding employee and `work_contact_id`.

If we create a fresh new contact in 16.0, without setting `mobile_phone` and `work_email` (in fact, setting any field except the name of the employee). The `_inverse_work_contact_details()` is triggered and a new partner is created with empty value.

So it seams clear that in 16.0 an employee must be related to a partner for the field `work_contact_id`.

As the philosophy of OpenUpgrade is that, when migrating from previous version you should end with data that are the same as if you started your project in the target version, the partner related to employee must be created during migration process.

remi-filament: Hi @remytms, from my point of view the employees are most of the time configured as follow (till Odoo v15)  : 
1. employee is linked to a user (field user_id which then points to a partner_id which is most of the time configured with parent_id = Your company (Odoo company_id.partner_id))
2. employee is linked to a private address (field address_home_id)

Therefore most of the time your new field work_contact_id already exists and I would try to assign in the following order :

1. employee user_id.partner_id (partner linked to user linked to employee)
2. search for partner with same work_email, type="contact" and parent_id = company_id.partner_id

If none of the above match, you may want to create a new contact.

For information, on Odoo v16 runbot, I get the following when looking at employees :

id | name | work_contact_id | address_home_id | user_partner_id
-- | -- | -- | -- | --
#7 | Marc Demo | 7,YourCompany, Marc Demo | 42,Mark Demo | 7,YourCompany, Marc Demo
#1 | Mitchell Admin | 3,YourCompany, Mitchell Admin | 41,Mitchell Admin | 3,YourCompany, Mitchell Admin

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
remytms: @remi-filament is right. It has to be with the maximum number of character for the name of a field in PostgreSQL.

I had to search the PostgreSQL manuel, but it's [here](https://www.postgresql.org/docs/14/datatype-character.html#DATATYPE-CHARACTER-SPECIAL-TABLE). 63 ASCII character long for all the names of the table, fields, etc.

If name specified is to long, PostgreSQL cut it at 63 characters. When creating default name for indices, constrains, etc it use a [dedicated algorithm](https://doxygen.postgresql.org/indexcmds_8c.html#a04ac525d457328bbe04c0f5924f761f7), which consist of cropping the longest element of the name until the whole name fits into 63 characters.

In this case, we use `get_legacy_name()` to retrieve a new name for the table. `get_legacy_name()` add something like `openupgrade_legacy_16_0_...`. Also the `rename_table()` function also tries to rename the indices, constrains, etc.

The original indices name for this table are:
```
Indexes:
    "hr_plan_hr_plan_activity_type_rel_pkey" PRIMARY KEY, btree (hr_plan_id, hr_plan_activity_type_id)
    "hr_plan_hr_plan_activity_type_hr_plan_activity_type_id_hr_p_idx" btree (hr_plan_activity_type_id, hr_plan_id)
Foreign-key constraints:
    "hr_plan_hr_plan_activity_type_rel_hr_plan_activity_type_id_fkey" FOREIGN KEY (hr_plan_activity_type_id) REFERENCES hr_plan_activity_type(id) ON DELETE CASCADE
    "hr_plan_hr_plan_activity_type_rel_hr_plan_id_fkey" FOREIGN KEY (hr_plan_id) REFERENCES hr_plan(id) ON DELETE CASCADE
```

Which gives after the rename with `get_legacy_name()`:
- `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel` (57 char)
- `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_pkey` (62 char)
- `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_hr_plan_activity_type_id_hr_p_idx` (87 char) -> `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_hr_plan_a` (63 char).
- `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_plan_activity_type_id_fkey` (87 char) -> `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl` (63 char)
- `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_plan_id_fkey` (73 char) -> `openupgrade_legacy_16_0_hr_plan_hr_plan_activity_type_rel_hr_pl` (63 char)

We see that two constrains are renamed the same way, which is a problem.

I think that this case should be fixed directly on `openupgradelib`. I will suggest something.

In this script I will set the new name of the table instead of using `get_legacy_name()`.

marielejeune: Hi @remytms did you progressed on this fix?

robinkeunen: Rémy is out of office for 2 weeks :-)

mostafabarmshory:
any update?

remytms:
I'm working on it this week.

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
remytms: @marielejeune I've not proposed fixes on `openupgradelib` yet, but in this migration script I use a custom name for the table. So that it should not be an issue anymore.

MiquelRForgeFlow:
this is wrong rebased...

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,167 @@
...
+            # and mobile than the employee may lead to a loss of data.
+            # work_email and mobile_phone will be replaced by the one
+            # from the partner.
+            employee.work_contact_id = employee.user_id.partner_id
+            _logger.info(
+                "Set work_contact_id of hr.employee(%s) to "
+                "res.partner(%s) (the res.user partner).",
+                employee.id,
+                employee.user_id.partner_id.id,
+            )
```
remytms: @remi-filament I implemented your idea to link the partner linked to the user_id if there is one.

But I'm not sure if such a link with work_contact_id should be done if the `mobile` and `email` of the employee and the partner are not the same. Because this may lead to a loss of data.

And I think that the goal of an openupgrade script is to avoid losing data.

Should I check that the partner found on the user_id has the same `mobile` and `email` as the employee before assigning it to `work_contact_id` ?

remytms:
@mostafabarmshory Your review is welcome, on this new proposal for the migration script of `hr` module. :)

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remi-filament: @remytms thanks for that !
Yes it would make sense to check this out (and maybe have some specific action in case fields are empty on partner or employee side)

remytms: Indeed, I will address this.

remytms: @remi-filament Done.

mostafabarmshory:
> @mostafabarmshory Your review is welcome, on this new proposal for the migration script of `hr` module. :)

Dear @remytms 
Tx for your mentions. That's your kindness.

I followed your changes and compared them with the head revision. I do not know how to write a review, honestly. Could you show me how to do that, please?

Finally tx to your work.

remytms:
@mostafabarmshory If it's your first time doing a pull request review, I invite you to read [this documentation on how to make comments and approve a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request)

If you are used to making pull request review but don't see the button, then you may need to be part of OCA organisation on github I think. Here is [documentation](https://odoo-community.org/get-involved/contribute).

mostafabarmshory (commented):
tx to your work

mostafabarmshory (approved):
tx to your work. Sorry, My last review is marked as a comment. It is my fault. I add a new review.

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,43 @@
...
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.update_module_moved_models(
+        env.cr, "hr.contract.type", "hr_contract", "hr"
+    )
+    openupgrade.update_module_moved_fields(
```
pedrobaeza: This one is not needed anymore: check https://github.com/OCA/openupgradelib/blob/29113956ab16eac1af01b4a4fd4db74257a45763/openupgradelib/openupgrade.py#L3042

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+hr           / hr.contract.type         / _order                        : previously in module hr_contract
+hr           / hr.contract.type         / display_name (char)           : previously in module hr_contract
+hr           / hr.contract.type         / name (char)                   : previously in module hr_contract
+# DONE: pre-migration: moved from hr_contract to hr
+
+hr           / hr.contract.type         / sequence (integer)            : NEW
+# NOTHING TO DO
+
+hr           / hr.department            / master_department_id (many2one): NEW relation: hr.department, isfunction: function, stored
+hr           / hr.department            / parent_path (char)            : NEW
```
pedrobaeza, line 23: It's always good to say why this is nothing to do:

```suggestion
hr           / hr.department            / parent_path (char)            : NEW
# NOTHING TO DO: Computed by ORM in the module update.

```

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+hr           / hr.contract.type         / __last_update (datetime)      : previously in module hr_contract
+hr           / hr.contract.type         / _order                        : previously in module hr_contract
+hr           / hr.contract.type         / display_name (char)           : previously in module hr_contract
+hr           / hr.contract.type         / name (char)                   : previously in module hr_contract
+# DONE: pre-migration: moved from hr_contract to hr
+
+hr           / hr.contract.type         / sequence (integer)            : NEW
+# NOTHING TO DO
+
+hr           / hr.department            / master_department_id (many2one): NEW relation: hr.department, isfunction: function, stored
```
pedrobaeza: Looking at the logs, it seems this one is computed before computing the parent_path:
```
2024-01-12 15:59:27,310 22110 INFO openupgrade odoo.modules.registry: module hr: creating or updating database tables 
2024-01-12 15:59:27,336 22110 INFO openupgrade odoo.models: Prepare computation of hr.department.master_department_id 
2024-01-12 15:59:27,336 22110 INFO openupgrade odoo.models: Computing parent_path for table hr_department... 
```
so the values won't be correct. Have you checked them in a test DB?

If not correct, you can:

- Pre-create the column for avoiding the normal computation.
- Call manually the compute on post-migration.

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,174 @@
...
+                    partner.id,
+                )
+        else:
+            matching_partner = env["res.partner"].search(
+                [
+                    ("email", "=", employee.work_email),
+                    ("mobile", "=", employee.mobile_phone),
+                ]
+            )
+            if len(matching_partner) == 1:
```
pedrobaeza: I would link to the first one if len > 1, as at the end, it's the same result.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+hr           / hr.employee              / work_contact_id (many2one)    : NEW relation: res.partner
+hr           / hr.employee              / work_email (char)             : now a function
+# TODO: post-migration: create work_contact_id for each employee
+
+hr           / hr.job                   / active (boolean)              : NEW hasdefault: default
+hr           / hr.job                   / contract_type_id (many2one)   : NEW relation: hr.contract.type
+hr           / hr.job                   / state (selection)             : DEL required, selection_keys: ['open', 'recruit']
+# NOTHING TO DO: hr.job does not work with state anymore.
+
+hr           / hr.plan                  / company_id (many2one)         : NEW relation: res.company, hasdefault: default
```
pedrobaeza, line 35: For preserving the previous behavior, pre-create the column for leaving it blank and don't let the default to act.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+hr           / hr.plan                  / department_id (many2one)      : NEW relation: hr.department
+# NOTHING TO DO
+
+hr           / hr.plan                  / plan_activity_type_ids (many2many): table is now 'False' ('hr_plan_hr_plan_activity_type_rel')
+hr           / hr.plan                  / plan_activity_type_ids (many2many): type is now 'one2many' ('many2many')
+# DONE: pre-migration and post-migration: move data from many2many table to plan_id colomn in hr.plan.activity.type
+
+hr           / hr.plan.activity.type    / company_id (many2one)         : NEW relation: res.company, hasdefault: default
+# NOTHING TO DO: not a required field so no default to set
```
pedrobaeza: The same comment as in `hr.plan`.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+
+hr           / hr.plan.activity.type    / plan_id (many2one)            : NEW relation: hr.plan
+# DONE: see plan_activity_type_ids from hr.plan
```
pedrobaeza, line 49: You can reorder this line to put it joined with the one2many field.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+
+hr           / res.users                / create_employee (boolean)     : NEW hasdefault: default
+hr           / res.users                / create_employee_id (many2one) : NEW relation: hr.employee
+# NOTHING TO DO
```
pedrobaeza: Comment why they are nothing to do for easing later reviews.

```suggestion
# NOTHING TO DO: store=False fields
```

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,75 @@
...
+
+hr           / resource.resource        / employee_id (one2many)        : NEW relation: hr.employee
+# NOTHING TO DO
```
pedrobaeza:
```suggestion
# NOTHING TO DO: inverse field already existing in previous version.
```

pedrobaeza (changes requested):
Wow, hard work with some of the changes here...

Please also comment in the noupdate_changes files the records with empty `plan_activity_type_ids`, as they are not adding value, and maybe they will empty their list.

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
mostafabarmshory: @pedrobaeza 

Would you explain why? The model is moved and the code is looking good.

pedrobaeza: This is already done by the orm

mostafabarmshory:
Dear @remytms, 
@pedrobaeza has a request to change the migration code. what next?

remytms:
@mostafabarmshory I will do the required changes. Normally this week.

pedrobaeza:
@remytms do you want me to do them in a superseded PR (respecting your attribution)?

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: It is not exactly the same result. If there is only one match, we can be pretty sure that the partner found is related to the employee.

But if there is several matches, then we cannot be sure which one to choose. So in such a case, I chose to create a new partner to link to the employee and show a warning in the logs (I will improve the log to differentiate the case where there is no matches and the case where there is several matches).

At the end, for a case with several matches, there will be one matches more. I think that a merge between all the contact should be done by a user after checking the reason why there is so many matches.

Also cleaning the database by removing the duplicate can be done before migrating. And with the warning, the duplicate partner can be found easily.

I will ping you when my new log warning proposal will be ready. :)

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Ok, so that, with a blank field for company_id, the hr.plan will not belong to a specific company. If we do nothing it will be linked to the default company set in `env` which is not the same. Do I understand correctly ?

pedrobaeza: Correct, changing the behavior from previous version, where all the plans were available for all the companies. Users now will have the option to restrict them in this version, but the provided default should be backwards compatible.

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
pedrobaeza: I still consider that linking to the first one found is a reasonable trade-off, better than finishing with still one more contact created. Any needed merge can be done later.

remytms: Ok, I will go it your way. If there is a warning then information can be found anyway.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Thanks for catching this. It works with demo data because there is no nested department in demo data.

pedrobaeza:
Please squash all commits into one. No need to split them, as GitHub already shows the differences between each push.

openupgrade_scripts/scripts/hr/16.0.1.1/pre-migration.py
remytms: Fixed.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Indeed, changed.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Done.

openupgrade_scripts/scripts/hr/16.0.1.1/post-migration.py
remytms: Done.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Fixed with tests.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Done.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Done.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Indeed, fixed.

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis_work.txt
remytms: Done.

remytms:
@pedrobaeza If I have not forget something, all comments were fixed. And commits were squashed.

pedrobaeza (approved):

robinkeunen:
🎉

PR #3913, opened 2023-06-13
[16.0][OU-ADD] hr_holidays

remytms:
Compute new values for allocation_validation_type on hr.leave.type as it is now a stored computed field with only two choices: "no" and "officer".

Load noupdate_changes.

### DIFF
--- a/openupgrade_scripts/scripts/hr_holidays/16.0.1.5/post-migration.py
+++ b/openupgrade_scripts/scripts/hr_holidays/16.0.1.5/post-migration.py
@@ -1,0 +1,28 @@
+# Copyright 2023 Coop IT Easy (https://coopiteasy.be)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def set_allocation_validation_type(env):
+    """Operate like the _compute_allocation_validation_type() function.
+
+    It set "no" by default except if employee_request is set to "no"
+    where it set "officer".
+    """
+    openupgrade.logged_query(
+        env.cr, "UPDATE hr_leave_type SET allocation_validation_type = 'no'"
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE hr_leave_type
+        SET allocation_validation_type = 'officer'
+        WHERE employee_requests = 'no'
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
/ocabot migration hr_holidays

<dependency></dependency>

openupgrade_scripts/scripts/hr_holidays/16.0.1.5/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def force_recompute_allocation_validation_type(env):
+    hr_leave_type_ids = (
+        env["hr.leave.type"].with_context({"active_test": False}).search([])
+    )
+    hr_leave_type_ids._compute_allocation_validation_type()
```
pedrobaeza: This is not performant and can lead to unexpected changes, triggering other computes including this one in depends, so it's advisable to do the equivalent one in SQL.

remytms: Thank you for the advice. It's now done by SQL directly. :)

openupgrade_scripts/scripts/hr_holidays/16.0.1.5/post-migration.py
```diff
@@ -0,0 +1,28 @@
...
+from openupgradelib import openupgrade
+
+
+def set_allocation_validation_type(env):
+    """Operate like the _compute_allocation_validation_type() function.
+
+    It set "no" by default except if employee_request is set to "no"
+    where it set "officer".
+    """
+    openupgrade.logged_query(
```
pedrobaeza, line 13: You only need to reallocate those with `set` value to the corresponding one. Please don't write the rest that are OK.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/hr_holidays/16.0.1.5/post-migration.py
pedrobaeza: As further explanation, imagine there's an extra module adding a new validation type. You are overwriting the leave types with this new validation type in this SQL.

pedrobaeza:
Superseded by #4335 correcting the SQL query about `allocation_validation_type`, and also doing the annotation in the coverage file, so the migration scripts are also tested.

PR #3915, opened 2023-06-13
[ADD] documentation : Use case SQL Constraint deleted

legalsylvain:


etobella (approved):

remytms (approved):
LGTM. Is it the same procedure when the constrain is modified ? Meaning the constrain still exist with the same name, but does not do the same check. If so, it can be good to mention it.

legalsylvain:
> LGTM. Is it the same procedure when the constrain is modified ? Meaning the constrain still exist with the same name, but does not do the same check. If so, it can be good to mention it.

I suppose. is this a real case you've dealt with before, or is it theoretical? 
If you have example, could you specify where ? 
In the "Use cases" section, I'd like to refer to real (and simple, if possible) examples. don't you think ? 

Édit : here I guess : https://github.com/OCA/OpenUpgrade/pull/4004#discussion_r1250964173

bguillot (approved):

Issue #3839, opened 2023-04-30
[DOC] Follow-up of the redesign of the documentation


legalsylvain:
this issue lists the changes made to the documentation.
If you are interested by the documentation, please review issue / PRs that are marked as "To review".

CC @etobella, @bguillot, @hbrunn, @StefanRijnhart, @pedrobaeza

**Note :**
- the up to date documentation is available [here](https://oca.github.io/OpenUpgrade/)
- a backup of the OCA documentation before refactoring is available [here](https://legalsylvain.github.io/openupgrade-old-documentation/).
 
### Already reviewed and merged 

<details>
  <summary>(Click to expand)</summary>

- January 25 : discussion about the interest of having a branch dedicated to documentation #3702 and validation. (April 28)
- April 29 : Issue about the broken documentation following the switch to the 16.0 branch : #3836
- April 30 : Initialization of the 'documentation' branch. #3837
- april, 30 : [FIX] Missing API references, for specific methods for openupgrade 12.0, 13.0 and 16.0 #3838 
- May, 1st : [REF] Switch to sphinx_rtd_theme #3841
- May, 1st : [REF] Refactor files that describes the branch and add text that indicates how to develop the documentation locally #3840
- May, 3 : [FIX] Do not ignore output path. (Fix 'View page source ' broken links) #3849 
- May, 7 : [REF] Remove ``docs`` folder and ``docsource`` folders (except coverage file) and related CI. + Update readme file in the recent openupgrade branches. (14.0, 15.0, 16.0). (see @pedrobaeza [comment](https://github.com/OCA/OpenUpgrade/issues/3839#issuecomment-1529640264)) + Update documentation when coverage files changed in "main" branches, for the branches : 
    - 16.0 :  #3853 +  #3855
    - 15.0 : #3859
    - 14.0 : #3860
- May, 16 : [REF] Move openupgradelib API Section in openupgradelib documentation (and make openupgradelib documentation up to date, the last update was in 2015).
    - openupgrade PR : https://github.com/OCA/OpenUpgrade/pull/3863 
    - openupgradelib PR : https://github.com/OCA/openupgradelib/pull/318 + https://github.com/OCA/openupgradelib/pull/324
- June, 1: Preliminary work : clean the intro part. #3902
- June, 1 : [REF] Documentation remove obsolete files #3903 
- June, 21 : 
    - [FIX] [10.0] modules coverage : remove title #3921
    - [12.0] modules coverage : remove title #3922
[...]

PR #3918, opened 2023-06-19
[16.0][OU-ADD] payment

remytms:
Changes:
- `acquirer` is now `provider`
- some fields and xmlids moved from `payment` to `account_payment`

Details in the `upgrade_analysis_work` file.

Testing is partial, because there is no migration script for the `account` module. Therefore the upgrade fails. But at least, the migration script for `payment` does not fail.


openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,213 @@
...
+payment      / account.payment          / authorized_transaction_ids (many2many): module is now 'account_payment' ('payment')
+payment      / account.payment          / payment_token_id (many2one)   : module is now 'account_payment' ('payment')
+payment      / account.payment          / payment_transaction_id (many2one): module is now 'account_payment' ('payment')
+payment      / account.payment          / refunds_count (integer)       : module is now 'account_payment' ('payment')
+payment      / account.payment          / source_payment_id (many2one)  : module is now 'account_payment' ('payment')
+payment      / account.payment          / suitable_payment_token_ids (many2many): module is now 'account_payment' ('payment')
+payment      / account.payment          / transaction_ids (many2many)   : module is now 'account_payment' ('payment')
+payment      / account.payment          / use_electronic_payment_method (boolean): module is now 'account_payment' ('payment')
+payment      / account.payment.method.line / payment_acquirer_id (many2one): DEL relation: payment.acquirer
+# TODO: theses fields needs to be moved to account_payment which is a new module that needs to be installed
```
remytms: I'm not sure how to handle this. Should I:

1. move these models and fields to `account_payment` and triggering the installation of `account_payment`.
2. backup these data and make a script in `account_payment` to get these data back again.

I'm in favor of the option 1. But I'm not sure how to trigger `account_payment` installation.

Am I wrong ? How to perform that ? :)

legalsylvain: Hi @remytms. 
>  triggering the installation of ``account_payment``

It is not necessary I think, as ``account_payment`` is installed if account is installed :
- V15 payment depends on account and auto_installable. [see](https://github.com/odoo/odoo/blob/15.0/addons/payment/__manifest__.py#L9-L35).
- V15 ``account_payment`` depends on account and is not auto_installable. [see](https://github.com/odoo/odoo/blob/15.0/addons/account_payment/__manifest__.py#L9).

BUT : 
- V16 account_payment is installed if account is installed. (and so force the installation of payment...)[see](https://github.com/odoo/odoo/blob/16.0/addons/account_payment/__manifest__.py#L8-L9). 

-> You can test in V16, install a fresh database with ``-i account`` and you'll see that payment and account_payment are installed.
-> That's a new odoo way to split "features" but force to have the features installed together.

So move of ``account`` fields in other ``account_payment`` installed module doesn't require work IMO (if fields are not renamed / has no changes).

What do you think @pedrobaeza ? 

regards.

remytms: Hi @legalsylvain ,

So, yes, we can assume that starting from a install on 15.0, if `payment` is installed on this DB, `account` will be also. And we cannot assume that `account_payment` is installed.

I don't know when the installation of `account_payment` will be triggered during the upgrade mechanism. Also I think that `account` and `payment` can be updated in any order, as they do not depends each other.

If moving theses models and field in the `pre-migration.py` script of `payment`, what will happen if after that `account_payment` is installed (either automatically by the upgrade mechanism or manually) ? Will it drop columns and tables and create new ones, or will it understand that there is existing models and fields that already exists in the database and that does need to be created again ?

Regards

legalsylvain: > So, yes, we can assume that starting from a install on 15.0, if payment is installed on this DB, account will be also. And we cannot assume that account_payment is installed.

In V16, if account is installed, it will force the installation of account_payment, so the installation of payment.
But if payment is installed, account will not be installed.

> I don't know when the installation of account_payment will be triggered during the upgrade mechanism. Also I think that account and payment can be updated in any order, as they do not depends each other.

- 'account' depends on 'base_setup', 'product', 'analytic', 'portal', 'digest'
- 'payment' depends : 'portal'
- 'account_payment' depends on 'account', 'payment'

So, the update order will be : 
- account, ..., payment, ... account_payment
- OR payment ..., account, ... account_payment

> I don't know when the installation of account_payment 

Yes, I have to write a page in the openupgrade documentation on that topic ! 

> Will it drop columns and tables and create new ones, or will it understand that there is existing models 

columns are not dropped. if you create a table (or a colum) a  before the load of the module thats "declares" that model (/ field) the database will be reused and not recreated. That's the core mecanism of upgrade. That's why we prepopulate "computed fields" in pre-migration, before the load of the module, to make it faster.

See for exemple : https://github.com/OCA/OpenUpgrade/pull/3873
review welcome in all the PR present here BTW https://github.com/OCA/OpenUpgrade/issues/3839 ;-)

remytms: Ok, so I think I have enough information to decide.

The module `account_payment` will always loads after the update of `payment` module. So moving the fields in the pre-migration script of `payment` is ok. :)

remytms:
@legalsylvain Can you reference this PR with the OCABot ? :)

openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,417 @@
...
+DEL payment.acquirer: payment.payment_acquirer_ogone (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_paypal (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_payulatam (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_payumoney (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_sepa_direct_debit (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_sips (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_stripe (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_test (noupdate)
+DEL payment.acquirer: payment.payment_acquirer_transfer (noupdate)
+# NOTHING TO DO: replaced by payment.provider element (see below)
```
carmenbianca: Are we sure about this? The user could have put modifications in these records that will not be carried over to the newly created records.

remytms: I'm pretty sure there is no change by the user on theses data. It contains data like this:

```xml

    <record id="payment_acquirer_adyen" model="payment.acquirer">
        <field name="name">Adyen</field>
        <field name="display_as">Credit Card (powered by Adyen)</field>
        <field name="image_128" type="base64" file="payment_adyen/static/src/img/adyen_icon.png"/>
        <field name="module_id" ref="base.module_payment_adyen"/>
        <field name="description" type="html">
            <p>
                A payment gateway to accept online payments via credit cards, debit cards and bank
                transfers.
            </p>
            <ul class="list-inline">
                <li class="list-inline-item"><i class="fa fa-check"/>Online Payment</li>
                <li class="list-inline-item"><i class="fa fa-check"/>Payment Status Tracking</li>
            </ul>
        </field>
        <!-- https://www.adyen.com/payment-methods -->
        <field name="payment_icon_ids"
               eval="[(6, 0, [
                   ref('payment.payment_icon_cc_bancontact'),
                   ref('payment.payment_icon_cc_maestro'),
                   ref('payment.payment_icon_cc_mastercard'),
                   ref('payment.payment_icon_cc_visa'),
                   ref('payment.payment_icon_cc_discover'),
                   ref('payment.payment_icon_cc_diners_club_intl'),
                   ref('payment.payment_icon_cc_jcb'),
                   ref('payment.payment_icon_cc_unionpay'),
               ])]"/>
    </record>
```

to 

```xml
    <record id="payment_provider_adyen" model="payment.provider">
        <field name="name">Adyen</field>
        <field name="display_as">Credit Card (powered by Adyen)</field>
        <field name="image_128" type="base64" file="payment_adyen/static/description/icon.png"/>
        <field name="module_id" ref="base.module_payment_adyen"/>
        <!-- https://www.adyen.com/payment-methods -->
        <field name="payment_icon_ids"
               eval="[(6, 0, [
                   ref('payment.payment_icon_cc_bancontact'),
                   ref('payment.payment_icon_cc_maestro'),
                   ref('payment.payment_icon_cc_mastercard'),
                   ref('payment.payment_icon_cc_visa'),
                   ref('payment.payment_icon_cc_discover'),
                   ref('payment.payment_icon_cc_diners_club_intl'),
                   ref('payment.payment_icon_cc_jcb'),
                   ref('payment.payment_icon_cc_unionpay'),
               ])]"/>
    </record>
```

carmenbianca: Users can absolutely modify these records.

Screenshot from 2023-07-03 18-09-56

In any case, currently after migration, you just get duplicates of everything:

```sql
> select id, name from payment_provider where cast(name->'en_US' as text) ilike '%mollie%';
+----+---------------------+
| id | name                |
|----+---------------------|
| 5  | {"en_US": "Mollie"} |
| 23 | {"en_US": "Mollie"} |
+----+---------------------+
SELECT 2
```

openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,417 @@
...
+# DONE: pre-migration: moved to payment.provider
+
+payment      / payment.acquirer         / pending_msg (html)            : DEL
+# DONE: pre-migration: moved to payment.provider
+
+payment      / payment.acquirer         / pre_msg (html)                : DEL
+# DONE: pre-migration: moved to payment.provider
+
+payment      / payment.acquirer         / provider (selection)          : DEL required, selection_keys: ['none']
+# NOTHING TO DO: Not needed anymore.
```
carmenbianca, line 113: This column still exists in the database after migration. Is that normal?

I do need this column to exist in my post-migration script in #3965, but it can be deleted afterwards as far as I'm concerned.

pedrobaeza: > This column still exists in the database after migration. Is that normal?

Yes, OU doesn't remove any column by design for preserving the maximum amount of data. You can remove it afterwards passing `database_cleanup`.

openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
remytms: Hum, indeed. I did not watch other field of the model. I will look deeper at this.

Thanks for pointing this out. :)

remytms: I think that I fixed it. :)

Renamed xmlids, added payment_provider in noupdate_changes, and loaded it.

stefan-tecnativa (approved):
LGTM! As account has already been merged, this can go on as well? TT42266

@pedrobaeza

pedrobaeza:
Hi, @remytms, thanks for the migration.

I have redone it in #4193 (as I can't push here) for being totally correct. Please take notes for other scripts you do of the following comments:

- No need to comment each line on the work file. If several of them apply the same comment, just group them and put the comment at the end of the block. It's also OK to reorder the lines to make the blocks larger.
- The same applies if there are non consecutive lines and both refers to the same action (model removed and model added for example): reorder them to go together in the same block with the same comment.
- There's no need anymore to use update_module_moved_fields. It's covered natively in OpenUpgrade till version 14, and by Odoo since them: https://github.com/OCA/openupgradelib/blob/d0aee908031876a95dc36c7a9a2c847186155aa2/openupgradelib/openupgrade.py#L3039-L3040
- There were some XML-IDs renamings that are not real, as they don't exist in target (Alipay, Ogone, etc). I have remove them.
- And for the remaining noupdate old records, you have to try to remove them with the corresponding method.
- For new noupdate records, there's nothing to do, so noupdate changes are not needed. They are new!
- And finally, for the noupdate changes, only real changes should be put. If not, any customization done by the user is lost, so better to keep changes at minimum.

remytms:
Hi @pedrobaeza 

Thanks for your comments. I take good note of your advice. :)

PR #3919, opened 2023-06-20
[14.0][OU-IMP] hr_timesheet: Remove permission group_timesheet_manager from user base.default_user

victoralmau:
Remove permission `group_timesheet_manager` from user `base.default_user`

Although odoo adds it at https://github.com/odoo/odoo/blob/14.0/addons/hr_timesheet/security/hr_timesheet_security.xml#L30 I think this permission should not be added for several reasons:
- Until now the user did not have this permission and it is strictly not necessary for a "basic" user.
- Having this permission inherits `hr.group_hr_user`, which in turn inherits `base.group_private_addresses` (which is quite controversial).

@Tecnativa TT44013


pedrobaeza (commented):
Please comment it instead of removing it for knowing that it was there, but explicitly disabled.

But I agree with the change.

PR #3920, opened 2023-06-20
[16.0][OU-ADD] account_payment: Nothing to do

remytms:
All the needed work is done in the pre-migration script of the `payment` module.


remytms:
Related to this PR: https://github.com/OCA/OpenUpgrade/pull/3918

legalsylvain:
/ocabot migration account_payment

<dependency></dependency>

victoralmau (approved):

chienandalu (commented):
Should we take care of this new init hook?: https://github.com/odoo/odoo/blob/16.0/addons/account_payment/__init__.py#L16

pedrobaeza:
Such hook is creating automatically an `account.payment.method`:

https://github.com/odoo/odoo/blob/16.0/addons/account_payment/models/payment_provider.py#L77

I think that's needed for migration, as we should have already created them in the previous version, but a double confirmation is interesting.

Meisolsson:
Hi,

I've tested this. For the "standard" payment providers it's fine since they included the method data in XML. But if you for instance created a provider manually in version 14 that provider will not have a method created i version 15 och 16. By that we should probably do that migration in 14 -> 15 and this can stay as is.

Meisolsson:
Actually there is one problem here but it comes from an failure in 14 -> 15. Relevant issue here: #4257.

mostafabarmshory:
Dear all,
This is not marked as a fixed task in `Migration to version 16.0`. Is it required to be checked? Can you give a feedback?

remytms:
Since the issue raised by @chienandalu was fixed [in 15.0](https://github.com/OCA/OpenUpgrade/pull/4259) by @Meisolsson, I think that this PR can be validated. Do you agree ?

chienandalu (approved):

PR #3932, opened 2023-06-28
[16.0][OU-ADD] gamification: Load the noupdate_changes

duyquyen96:
Load the noupdate_changes file.


legalsylvain (approved):
Noupdate change review. LGTM

Réf https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/gamification/16.0.1.0/noupdate_changes.xml

SonCrits (approved):

pedrobaeza:
I see that there's no deletion of the translations, so the scripts are incomplete. Please do another PR with such thing. Also reviewing the noupdate changes, current openupgradelib method is deleting all the translations of the record, while if only a few fields change (like in this case `name`), it's too much, so we should consider adding an optional field list to the method for only deleting that ones.

PR #3933, opened 2023-06-28
[16.0] [OU-ADD] hr_attendance

nguyenvietlam0640:


legalsylvain (approved):
Hi @nguyenvietlam0640 . thanks for tour contribution. Coule you take a look on red CI ?

Otherwise, lgtm. Noupdate change reviewed. 
https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/hr_attendance/16.0.2.0/noupdate_changes.xml

legalsylvain:
/ocabot migration hr_attendance

<dependency></dependency>

victoralmau (approved):

mostafabarmshory (approved):

openupgrade_scripts/scripts/hr_attendance/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Models in module 'hr_attendance'---
+---Fields in module 'hr_attendance'---
+hr_attendance / res.company              / attendance_barcode_source (selection): NEW selection_keys: ['back', 'front', 'scanner'], hasdefault: default
```
pedrobaeza, line 3: Please document always why they are nothing to do, and investigate further if the default is adequate or not:

```suggestion
hr_attendance / res.company              / attendance_barcode_source (selection): NEW selection_keys: ['back', 'front', 'scanner'], hasdefault: default
# NOTHING TO DO: Value only used in enterprise

```

openupgrade_scripts/scripts/hr_attendance/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
...
+hr_attendance / res.company              / attendance_kiosk_delay (integer): NEW hasdefault: default
```
pedrobaeza, line 6: The new default is different from the default behavior in v15.

Here the stored value is multiplied by 1000 in v16:

https://github.com/odoo/odoo/blob/17151f5c369dce7ca5d38c5fd0db8dceab5cfab7/addons/hr_attendance/models/hr_employee.py#L165

and here it's used:

https://github.com/odoo/odoo/blob/17151f5c369dce7ca5d38c5fd0db8dceab5cfab7/addons/hr_attendance/static/src/js/greeting_message.js#L86-L87

and in v15, the default value was 5000:

https://github.com/odoo/odoo/blob/17151f5c369dce7ca5d38c5fd0db8dceab5cfab7/addons/hr_attendance/static/src/js/greeting_message.js#L86-L87

So, to preserve previous behavior, we should set in all existing companies the value 5. Anyway I think they have changed this value for giving more time to read the message, so it's an acceptable change of behavior, and more being able to be configured later, but the important thing is to document all of this here.

openupgrade_scripts/scripts/hr_attendance/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
...
+hr_attendance / res.company              / attendance_kiosk_mode (selection): NEW selection_keys: ['barcode', 'barcode_manual', 'manual'], hasdefault: default
+
+# NOTHING TO DO: new fields
```
pedrobaeza: Please always document the nothing to do thing:

```suggestion
hr_attendance / res.company              / attendance_kiosk_mode (selection): NEW selection_keys: ['barcode', 'barcode_manual', 'manual'], hasdefault: default
# NOTHING TO DO: With the default value `barcode_manual`, we get the same behavior as v15
```

pedrobaeza (changes requested):

pedrobaeza (approved):
Added extra commit with the explanations of the nothing to do in the work file.

PR #3946, opened 2023-06-29
[16.0][OU-ADD] survey: Nothing to do

duyquyen96:
Nothing to do


openupgrade_scripts/scripts/survey/16.0.3.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,32 @@
...
+survey       / survey.question.answer   / value_image_filename (char)   : NEW
+survey       / survey.survey            / _order                        : _order is now 'create_date DESC' ('id')
+survey       / survey.user_input        / activity_ids (one2many)       : NEW relation: mail.activity
+survey       / survey.user_input        / message_follower_ids (one2many): NEW relation: mail.followers
+survey       / survey.user_input        / message_ids (one2many)        : NEW relation: mail.message
+survey       / survey.user_input        / message_main_attachment_id (many2one): NEW relation: ir.attachment
+survey       / survey.user_input        / website_message_ids (one2many): NEW relation: mail.message
+---XML records in module 'survey'---
+# NOTHING TO DO
+DEL ir.actions.server: survey.survey_action_server_clean_test_answers (noupdate)
```
MiquelRForgeFlow, line 17: needs to be deleted in post-migration

openupgrade_scripts/scripts/survey/16.0.3.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,32 @@
+---Models in module 'survey'---
+---Fields in module 'survey'---
+survey       / survey.question          / allow_value_image (boolean)   : DEL
+survey       / survey.question          / background_image (binary)     : NEW attachment: True, hasdefault: compute
+survey       / survey.question          / column_nb (selection)         : DEL selection_keys: ['12', '2', '3', '4', '6']
+# NOTHING TO DO: column_nb unuse from 15.0
+survey       / survey.question          / question_placeholder (char)   : NEW hasdefault: compute
```
victoralmau, line 7: IMO I think you should create the column and set the correct value in the migration to avoid odoo to do the compute and optimize time.

victoralmau (changes requested):
I think it would be a good idea to rename the xmlid of some views that are "_same_", for example `question_result_number_or_date` (v15) and `question_result_number_or_date_or_datetime` (v16) to keep a good compatibility if someone had inherited from some view.

legalsylvain:
> I think it would be a good idea to rename the xmlid of some views that are "same", for example question_result_number_or_date (v15) and question_result_number_or_date_or_datetime (v16) to keep a good compatibility if someone had inherited from some view.

Could you elaborate ? I dont' understand in wich case it is problematic. Thanks !

pedrobaeza:
Example: manually created views inheriting that one. Of course that "shouldn't happen", but in case you have some rebel users doing that on their own, the original view can't be auto-deleted, as there's a children view that prevents it, and you end up with several views for the same model or at least dead views.

legalsylvain:
> Example: manually created views inheriting that one. Of course that "shouldn't happen", but in case you have some rebel users doing that on their own, the original view can't be auto-deleted, as there's a children view that prevents it, and you end up with several views for the same model or at least dead views.

I have reservations about asking for these changes in a review. 
If someone encounters this case, he can make the change in a PR, but if it's systematically requested, it risks adding work to contributors, whereas it's a case of use "that shouldn't happen", as you say.

If this request is ["suggestion - non blocking"](https://conventionalcomments.org/), then everything's OK.

pedrobaeza:
Superseded by #4334 

Each field addition or deletion should be analyzed with care, as there can be consequences or changes of behavior. Please check the ones I have done in the other PR as reference.

PR #3949, opened 2023-06-30
[16.0][OU-ADD] website

duyquyen96:


legalsylvain:
/ocabot migration website

<dependency></dependency>

pedrobaeza:
On this version, Odoo has upgraded Bootstrap from 4 to 5, so similar changes than in v12 would be needed:

- openupgradelib classes transformations: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_120.py#L55
- OpenUpgrade code for applying the transformations: https://github.com/OCA/OpenUpgrade/pull/1819 + https://github.com/OCA/OpenUpgrade/pull/1971

daiduongnguyen-odoo:
> On this version, Odoo has upgraded Bootstrap from 4 to 5, so similar changes than in v12 would be needed:
> 
> * openupgradelib classes transformations: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_120.py#L55
> * OpenUpgrade code for applying the transformations: [[MIG] website: Handle Bootstrap 3->4 and Less->Scss #1819](https://github.com/OCA/OpenUpgrade/pull/1819) + [[IMP] website: Translate metadata #1971](https://github.com/OCA/OpenUpgrade/pull/1971)

Working on it via https://github.com/OCA/openupgradelib/pull/338 , hopefully not miss anything :v

daiduongnguyen-odoo:
Dependency

- [ ] https://github.com/OCA/openupgradelib/pull/338

openupgrade_scripts/scripts/website/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,6 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(env, ["website.website_menu"])
```
daiduongnguyen-odoo: Hmmm maybe missing load noupdate

openupgrade_scripts/scripts/website/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,100 @@
...
+        SELECT w.id, rl.id
+        FROM website w
+        CROSS JOIN res_lang rl
+        WHERE w.id NOT IN (SELECT website_id FROM website_lang_rel)
+        AND rl.active = true;
+        """,
+    )
+
+
+def keep_the_first_domain_when_duplicate(env):
```
daiduongnguyen-odoo: Why need this, do not understand ?

duyquyen96:
We are working on a separate repository, and once the code is completed, we will create a pull request to your repository.

chrisandrewmann:
Ran an upgrade from 15.0 with website. Generally ok but came across a strange issue when trying to edit header or footer on WYSIWYG editor afterwards. Not sure if it was OpenUpgrade or an Odoo bug so reported it as an issue on there.
https://github.com/odoo/odoo/issues/132860

daiduongnguyen-odoo:
@chrisandrewmann I suggest you regenerate asset after migrating, if not working then i'll try to reproduce it at my local

chrisandrewmann:
Thanks @duong77476 for the quick reply. I've tried regenerating asset bundles a couple of times.
No joy. 
If it helps also i'm using the Default website theme with "Slogan" header and the "Contact" footer option.

Update 30/09/23: I also tested an upgrade to v16 using the same DB but via the official Odoo upgrade platform, this issue does not appear so must be something OpenUpgrade specific regards website

daiduongnguyen-odoo:
> Thanks @duong77476 for the quick reply. I've tried regenerating asset bundles a couple of times. No joy. If it helps also i'm using the Default website theme with "Slogan" header and the "Contact" footer option.

oki thanks, i'll check it out later

gdgellatly:
In testing a v13-v16 migration I received an error in the 15-16 step.

A prior xml id with the name Restricted Editor already existed with xml id `website.group_website_publisher` causing a duplicate key error

PNurm:
> Ran an upgrade from 15.0 with website. Generally ok but came across a strange issue when trying to edit header or footer on WYSIWYG editor afterwards. Not sure if it was OpenUpgrade or an Odoo bug so reported it as an issue on there. [odoo/odoo#132860](https://github.com/odoo/odoo/issues/132860)

This has happened to two different databases after migration to 14.0 > 15.0 >16.0. It seems to be caused by changes in 'website.layout' view. I've not yet pinpointed the exact change causing it as there are many, easier to just replace the arch completely with 16 version.

daiduongnguyen-odoo:
Thank you everyone,  i will resolve conflict today

daiduongnguyen-odoo:
Done

pedrobaeza:
I have pushed a rebased version improving a bit the documentation on the work file, and making a backup of the arch_db column before updating it, as OpenUpgrade should try to be as conservative as possible.

@chienandalu we will need later to add the same HTML fields conversion (BS4 > BS5 this time) that we have made in previous version, but there's still the problem with the malformed HTMLs.

PR #3950, opened 2023-06-30
[16.0] [OU-ADD] hr_skills: Nothing to do

nguyenvietlam0640:


legalsylvain:
/ocabot migration hr_skills

<dependency></dependency>

SirAionTech (changes requested):
Please check the conflicts and the commits: there should be only one commit.

PR #3955, opened 2023-06-30
[16.0] [OU-ADD] hr_holidays_attendance: Nothing to do

nguyenvietlam0640:


legalsylvain:
/ocabot migration hr_holidays_attendance

<dependency></dependency>

zamberjo:
Hi @nguyenvietlam0640, can you check the PR? there seems to be something wrong and there are more changes than there should be.

PR #3959, opened 2023-06-30
[14.0][OU-IMP] website: convert all HTML fields from version 13 to 14

pilarvargas-tecnativa:
Convert all HTML strings in all HTML fields from Odoo version 13 to the new definitions in version 14.

cc @Tecnativa TT41471

### DIFF
--- a/openupgrade_scripts/scripts/website/14.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/website/14.0.1.0/end-migration.py
@@ -1,0 +1,23 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+from openupgradelib import openupgrade, openupgrade_140
+
+
+def convert_field_html_string(env):
+    fields = env["ir.model.fields"].search(
+        [("ttype", "=", "html"), ("store", "=", True)]
+    )
+    for field in fields:
+        model = field.model_id.model
+        if env.get(model, False) is not False and env[model]._auto:
+            if openupgrade.table_exists(env.cr, env[model]._table):
+                if openupgrade.column_exists(env.cr, env[model]._table, field.name):
+                    openupgrade_140.convert_field_html_string_13to14(
+                        env,
+                        model,
+                        field.name,
+                    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    convert_field_html_string(env)


legalsylvain (commented):
why do you put this call here. (website / end-migration.py). AfAIU, it can be interesting also if website is not installed.

pedrobaeza:
If `website` not installed, then you don't drag and drop blocks (snippets) that contain the HTML that is changing between versions, so outside `website`, it's useless.

legalsylvain (approved):
> If website not installed, then you don't drag and drop blocks (snippets) that contain the HTML that is changing between versions, so outside website, it's useless.

thanks for the explanation.

openupgrade_scripts/scripts/website/14.0.1.0/end-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def convert_field_html_string(env):
+    fields = env["ir.model.fields"].search(
+        [("ttype", "=", "html"), ("store", "=", True)]
+    )
+    for field in fields:
+        model = field.model_id.model
+        if env.get(model):
+            if openupgrade.table_exists(env[model]._table):
```
pedrobaeza: Missing `env.cr` parameter, isn't it?

pilarvargas-tecnativa: Yes, sorry, that's it

openupgrade_scripts/scripts/website/14.0.1.0/end-migration.py
```diff
@@ -0,0 +1,23 @@
...
+from openupgradelib import openupgrade, openupgrade_140
+
+
+def convert_field_html_string(env):
+    fields = env["ir.model.fields"].search(
+        [("ttype", "=", "html"), ("store", "=", True)]
+    )
+    for field in fields:
+        model = field.model_id.model
+        if env.get(model):
```
pedrobaeza: Sorry, I told you this incorrectly. It should be:

```suggestion
        if env.get(model, False) is not False:
```

pilarvargas-tecnativa: done!

openupgrade_scripts/scripts/website/14.0.1.0/end-migration.py
```diff
@@ -0,0 +1,23 @@
...
+from openupgradelib import openupgrade, openupgrade_140
+
+
+def convert_field_html_string(env):
+    fields = env["ir.model.fields"].search(
+        [("ttype", "=", "html"), ("store", "=", True)]
+    )
+    for field in fields:
+        model = field.model_id.model
+        if env.get(model, False) is not False:
```
pedrobaeza:
```suggestion
        if env.get(model, False) is not False and env[model]._auto:
```

chienandalu: Oh, good one :)

pedrobaeza (approved):

PR #3962, opened 2023-06-30
[14.0][OU-FIX] website: BS processing fine-tuning

pedrobaeza:
- Exclude directly mail.message table.

TODO: The update is being performed although the content doesn't change, so we need to check why.

@Tecnativa


pedrobaeza:
The reupdate is something of openupgradelib, so we can merge this.

PR #3971, opened 2023-07-01
[16.0][OU-ADD] stock_account: nothing to do

hoangtiendung070797:
Nothing to do


legalsylvain:
/ocabot migration stock_account

<dependency></dependency>

pedrobaeza:
This one should write `cogs` in `display_type` for the journal items that are of this type.

robinkeunen (approved):

robinkeunen:
> This one should write `cogs` in `display_type` for the journal items that are of this type.

@pedrobaeza display_type value is computed in `_compute_display_type`, do you suggest the code needs to be fixed ? Or is there something to do in this PR ? If so, I'd need some more guidance :-)

pedrobaeza:
`_compute_display_type` is not added in this module, so even if the technique used for filling the value is to launch the compute method, you should relaunch it here for recomputing those that the compute override will impact. Another technique if the compute method is used, is to launch the compute on end migration. But that has several problems:

- ORM compute methods are slower than SQL queries.
- They can trigger other dependent computes, so slower and maybe values overwrites that you don't want.
- If the value computed in the method is needed for any other process, you are deferring its computation, so maybe you obtain errors/wrong data on other places.

Thus, in this case, you should try to do a SQL query that performs that filling in an optimal way and in the adequate moment (pre or post).

remytms:
Superseeded by [this PR](https://github.com/OCA/OpenUpgrade/pull/4293)

andreampiovesana:
close for https://github.com/OCA/OpenUpgrade/pull/4293 ?

PR #3972, opened 2023-07-01
[16.0][OU-ADD] event_booth_sale: Nothing to do

diamondnguyenkc:
Nothing to do


legalsylvain:
/ocabot migration event_booth_sale

<dependency>Depends on : 

- event_booth : #3974</dependency>

PR #3983, opened 2023-07-03
[16.0] [OU-ADD] im_livechat

locdang8:


legalsylvain:
/ocabot migration im_livechat

<dependency></dependency>

openupgrade_scripts/scripts/im_livechat/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,6 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "im_livechat", "16.0.1.0/noupdate_changes.xml")
```
MiquelRForgeFlow, line 10: I think you should delete translations of `livechat_email_template` template

PR #3984, opened 2023-07-03
[16.0] [OU-ADD] snailmail: nothing to do

locdang8:


legalsylvain:
/ocabot migration snailmail

<dependency>Depends on : 

- iap_mail : #3991</dependency>

MiquelRForgeFlow (approved):

PR #3942, opened 2023-06-29
[16.0][OU-ADD] iap_crm: Nothing to do

duyquyen96:
Nothing to do

Blocked by dependency

PR #3988, opened 2023-07-03
[16.0][OU-ADD] crm_iap_mine: Nothing to do

diamondnguyenkc:
Nothing to do


legalsylvain:
/ocabot migration crm_iap_mine

<dependency>Depends on : 

- iap_crm : #3942
- iap_mail : #3991</dependency>

MiquelRForgeFlow (approved):

PR #3998, opened 2023-07-03
[16.0][OU-ADD] google_drive: Nothing to do

daiduongnguyen-odoo:
Deprecated module, this module is no longer present as its'feature completely removed from odoo due to change in OOB flow check PR `https://github.com/odoo/odoo/pull/94703`


docsource/modules150-160.rst
```diff
@@ -156,7 +156,7 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | google_calendar                                 |                      |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| google_drive                              |                      |                                                 |
+| |del| google_drive                              |Nothing to do         |Deprecated module, features completely removed   |
```
daiduongnguyen-odoo: Can i write like this XD

pedrobaeza:
In this case, you can't say "Nothing to do", as there's no replacement. Maybe the other description. This should be manually removed IMO.

daiduongnguyen-odoo:
> In this case, you can't say "Nothing to do", as there's no replacement. Maybe the other description. This should be manually removed IMO.

OK , i see, maybe i'll go with `Deprecated Module` instead

docsource/modules150-160.rst
```diff
@@ -156,7 +156,7 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | google_calendar                                 |Done                  |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| google_drive                              |                      |                                                 |
+| |del| google_drive                              |Deprecated module     |Obsolete module, features completely removed     |
```
MiquelRForgeFlow:
```suggestion
| |del| google_drive                              | Deprecated module    | Obsolete module                                  |
```
I prefer it shorter.

daiduongnguyen-odoo: Same like https://github.com/OCA/OpenUpgrade/pull/4002

MiquelRForgeFlow (approved):

PR #3999, opened 2023-07-03
[16.0][OD-ADD] sale_product_matrix: Nothing to do

diamondnguyenkc:
Nothing to do


legalsylvain:
/ocabot migration sale_product_matrix

<dependency>Depends on : 

- product_matrix : #4035</dependency>

PR #4002, opened 2023-07-03
[16.0][OU-ADD] google_spreadsheet: Nothing to do

daiduongnguyen-odoo:
Deprecated module, this module is no longer present as its'feature completely removed from odoo due to change in OOB flow check PR `https://github.com/odoo/odoo/pull/94703`


docsource/modules150-160.rst
```diff
@@ -162,7 +162,7 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | google_recaptcha                                |                      |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| google_spreadsheet                        |                      |                                                 |
+| |del| google_spreadsheet                        |Deprecated module     |Obsolete module, features completely removed.    |
```
MiquelRForgeFlow:
```suggestion
| |del| google_spreadsheet                        | Nothing to do        | Obsolete module                                 |
```
shorter

pedrobaeza: This is not Nothing to do, as there's no replacement, so the text should specify that.

MiquelRForgeFlow: Selection_1488

This are the only valid options for the first column.

pedrobaeza: Not really. If you put any other value, it's not parsed, but anyway, I prefer empty, and put the explanation. `Feature removed. Search for alternatives` or similar in the next column.

MiquelRForgeFlow: Ok, that's fine for me. In https://github.com/OCA/OpenUpgrade/blob/14.0/docsource/modules130-140.rst, we can see that the first column for `web_diagram` was left empty, and as many other similar cases in other versions.

daiduongnguyen-odoo: Ok, thank you both, i'll go with empty and put the explanation then

daiduongnguyen-odoo: > Selection_1488
> 
> This are the only valid options for the first column.

May i ask, where you get this, i don't know it does exist?

MiquelRForgeFlow (approved):

PR #3969, opened 2023-07-01
[16.0] [OU-ADD] auth_password_policy: nothing to do

locdang8:

PR #4003, opened 2023-07-03
[16.0] [OU-FIX] auth_password_policy: change docsource

locdang8:


legalsylvain (commented):
why this change ?

locdang8:
> why this change ?

mm, i see  value in cell auth_oauth should be in cell auth_password_policy, auth_oauth has not been migrate

legalsylvain:
Indeed : 
fixes : https://github.com/OCA/OpenUpgrade/pull/3969/files

PR #4004, opened 2023-07-03
[16.0][OU-ADD] sale: migration

hoangtiendung070797:
Migration module


legalsylvain:
/ocabot migration sale

Could you take a look on the red CI ?

<dependency>Depends on : 

- account_payment : #3920</dependency>

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+sale         / account.bank.statement.line / partner_shipping_id (many2one): module is now 'account' ('sale')
+sale         / account.move             / partner_shipping_id (many2one): module is now 'account' ('sale')
+sale         / account.move.line        / is_downpayment (boolean)      : NEW
+sale         / account.payment          / partner_shipping_id (many2one): module is now 'account' ('sale')
+# NOTHING TO DO
+
+sale         / payment.acquirer         / so_reference_type (selection) : DEL selection_keys: ['partner', 'so_name']
+sale         / payment.provider         / so_reference_type (selection) : NEW selection_keys: ['partner', 'so_name'], hasdefault: default
+# NOTHING TO DO: renamed in pre-migration above
```
remytms: This comment is confusing for me. Because the field `so_reference_type` has not been renamed. However, the model the field belongs to has been renamed from `payment.acquirer` to `payment.provider` and this rename has not been done.

You may take a look at this function: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2934

And example in this PR #906 :)

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+---Fields in module 'sale'---
+sale         / account.analytic.applicability / business_domain (False)       : NEW selection_keys: ['bill', 'expense', 'general', 'invoice', 'purchase_order', 'sale_order'], mode: modify
+# NOTHING TO DO
+
+sale         / account.bank.statement.line / partner_shipping_id (many2one): module is now 'account' ('sale')
+sale         / account.move             / partner_shipping_id (many2one): module is now 'account' ('sale')
+sale         / account.move.line        / is_downpayment (boolean)      : NEW
+sale         / account.payment          / partner_shipping_id (many2one): module is now 'account' ('sale')
+# NOTHING TO DO
```
remytms, line 20: Indeed the change of module from `sale` to `account`should be managed in the migration script of `account` as `account` will be updated first.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+sale         / payment.acquirer         / so_reference_type (selection) : DEL selection_keys: ['partner', 'so_name']
+sale         / payment.provider         / so_reference_type (selection) : NEW selection_keys: ['partner', 'so_name'], hasdefault: default
+# NOTHING TO DO: renamed in pre-migration above
+
+sale         / sale.order               / show_update_fpos (boolean)    : NEW
+sale         / sale.order               / show_update_pricelist (boolean): not stored anymore
+# NOTHING TO DO
```
remytms: Have you checked that the ORM correctly drop the column `show_update_pricelist` from table `sale_order` ?

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+sale         / sale.order.line          / analytic_distribution_stored_char (char): NEW isfunction: function, stored
+# NOTHINGH TO DO: not used any more
```
remytms: The comment is confusing, because it's a new feature. But agree, there is nothing to do. The field will be computed and stored on the fly by the ORM when needed.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+sale         / sale.order.line          / analytic_tag_ids (many2many)  : DEL relation: account.analytic.tag
+# NOTHINGH TO DO: not used any more
```
remytms: We may want to put a warning here, as the model `account.analytic.tag` has been removed from module `analytic`. But there is discussion on the OCA, on how to manage this removal.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+sale         / sale.order.line          / product_type (selection)      : previously in module sale_stock
+# NOTHING TO DO: handle by ORM
```
remytms, line 37: I think that you should move it from module `sale_stock` to module `sale`. And it should be done in the migration script of `sale` because `sale` will be updated before `sale_stock`.

If not moved manually, I fear that when `sale_stock` will be updated, it will drop the column as it was part of this module previously and not anymore. But I may be wrong.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+DEL ir.model.access: sale.access_account_move_manager
+DEL ir.model.access: sale.access_account_tax_sale_manager
+DEL ir.model.access: sale.access_product_product_attribute_custom_value_sale_manager
+DEL ir.model.access: sale.access_report_all_channels_sales
+DEL ir.model.access: sale.access_sale_payment_acquirer_onboarding_wizard
+DEL ir.model.access: sale.access_sale_report_manager
+# NOTHING TO DO
+
+ir.model.constraint: sale.constraint_sale_order_date_order_conditional_required (changed definition: is now 'check((state in('sale','done') and date_order is not null) or state not in('sale','done'))' ('check((state in('sale','done') and date_order is not null) or state not in('sale','done') )'))
+# pre-migration: dropped constraint and let ORM add it again with the new definition
```
remytms:
```suggestion
# DONE: pre-migration: dropped constraint and let ORM add it again with the new definition
```

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+DEL ir.ui.view: sale.product_template_form_view_invoice_policy
+DEL ir.ui.view: sale.product_template_sale_form_view
+DEL ir.ui.view: sale.report_invoice_document_inherit_sale
+DEL ir.ui.view: sale.sale_onboarding_order_confirmation_form
+NEW mail.template: sale.mail_template_sale_cancellation (noupdate)
+# NOTHING TO DO
+
+DEL res.groups: sale.group_delivery_invoice_address [renamed to account module]
+# DONE: pre-migration: renamed xmlids
```
remytms: I think that this rename should be done in migration script of `account` module as `account` module will be updated before module `sale` (`sale` depends on `payment` and `payment` depends on `account`). Else it will be already created when updating module `account`.

remytms (commented):
Thanks for this job. Some remarks. I may be wrong on some of them.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
daiduongnguyen-odoo: @remytms Hi remy can you give me the link to the discussion please

hoangtiendung070797:
@remytms Thanks for your help. I will check it again

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
legalsylvain: This one ? https://github.com/OCA/account-analytic/issues/551

daiduongnguyen-odoo: @legalsylvain Thank you very much

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
hoangtiendung070797: @remytms  I think in this case we don't need to do anything because i renamed that model in pre-migration. What do you think?

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
hoangtiendung070797: @remytms Ok, i will change comment to "handle by ORM"

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
hoangtiendung070797: This is difficult to fact check because it is dependent, according to the mechanism, I believe it has been removed

openupgrade_scripts/scripts/sale/16.0.1.2/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+        env,
+        [
+            "sale.mail_notification_paynow_online",
+        ],
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "sale", "noupdate_changes.xml")
```
huguesdk:
```suggestion
    openupgrade.load_data(env.cr, "sale", "16.0.1.2/noupdate_changes.xml")
```

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.update_module_moved_fields(
+        env.cr, "sale.order.line", _moved_fields, "sale_stock", "sale"
+    )
+    openupgrade.rename_models(env.cr, _models_renames)
+    openupgrade.rename_xmlids(env.cr, _renames_xmlids)
```
huguesdk, line 122: an xmlid should also be created for the `ir.config_parameter` with key `"sale.default_invoice_email_template"`, because it exists in 15.0 but without xmlid, and an xmlid for it has been added in 16.0, which causes the migration to fail with this error:
```
bad query: INSERT INTO "ir_config_parameter" ("create_date", "create_uid", "key", "value", "write_date", "write_uid") VALUES ('2023-07-06 13:58:38.027885', 1, 'sale.default_invoice_email_template', '7', '2023-07-06 13:58:38.027885', 1) RETURNING "id"
ERROR: duplicate key value violates unique constraint "ir_config_parameter_key_uniq"
DETAIL: Key (key)=(sale.default_invoice_email_template) already exists.
```

here’s how to add the xmlid. maybe there is a cleaner way?
```python
    openupgrade.add_xmlid(
        env.cr,
        "sale",
        "default_invoice_email_template",
        "ir.config_parameter",
        env["ir.config_parameter"]
        .search([("key", "=", "sale.default_invoice_email_template")], limit=1)
        .id,
        noupdate=True,
    )
```

huguesdk (changes requested):
here are some changes needed to be able to run the migration. i did not look into all changes.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
remytms: After reading of other PR. It seams that we do not drop column in OpenUpgrade, in order to preserve data. Dropping this column can be done after the migration by running `database_cleanup`.

So you're right, there is nothing to do. :)

pedrobaeza: What you can do for avoiding misunderstandings if you look directly to DB schema, is to rename the column to `None` for getting the OU automatic name.

remytms: Thanks for sharing this good practice. :)

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
hoangtiendung070797: @huguesdk Thanks for your help. I didn't realize this mistake

huguesdk: with a database initialized with only `sale_stock`, this `ir.config_parameter` does not exist. sorry, my fix does not work in that case. instead, a test is needed, and it is better to move it to a function:

```python
def _create_xml_ids(env):
    config_parameter = env["ir.config_parameter"].search(
        [("key", "=", "sale.default_invoice_email_template")]
    )
    if config_parameter:
        openupgrade.add_xmlid(
            env.cr,
            "sale",
            "default_invoice_email_template",
            "ir.config_parameter",
            config_parameter.id,
            noupdate=True,
        )
```

huguesdk: the tests in the ci are currently failing because the parameter does not exist. adding the test will fix this. the error is:
```
psycopg2.errors.DatatypeMismatch: column "res_id" is of type integer but expression is of type boolean
LINE 1: ..._invoice_email_template','sale','ir.config_parameter',false)
                                                                 ^
```

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+sale         / sale.order.line          / product_type (selection)      : previously in module sale_stock
+# DONE: pre-migration: moved from sale_stock to sale
+
+sale         / sale.order.line          / qty_delivered (float)         : not a function anymore
+# NOTHING TO DO: No database change
+
+sale         / sale.order.line          / qty_delivered_manual (float)  : DEL
+# NOTHING TO DO: not used any more, this field is useless since the new compute method from 13.0
+
+---XML records in moduldatabasee 'sale'---
```
huguesdk:
```suggestion
---XML records in module 'sale'---
```

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
+---Models in module 'sale'---
+obsolete model report.all.channels.sales [sql_view]
```
huguesdk, line 2: this sql view must be deleted.
```sql
drop view if exists report_all_channels_sales
```

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,56 @@
+from openupgradelib import openupgrade
+
+_models_renames = [
+    (
+        "sale.payment.acquirer.onboarding.wizard",
+        "sale.payment.provider.onboarding.wizard",
+    ),
```
huguesdk, line 7: this is not enough, as `openupgrade.rename_models()` does not rename the tables. a call to `openupgrade.rename_tables()` is needed.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+sale         / payment.acquirer         / so_reference_type (selection) : DEL selection_keys: ['partner', 'so_name']
+sale         / payment.provider         / so_reference_type (selection) : NEW selection_keys: ['partner', 'so_name'], hasdefault: default
+# NOTHING TO DO
+
+sale         / sale.order               / show_update_fpos (boolean)    : NEW
+sale         / sale.order               / show_update_pricelist (boolean): not stored anymore
+# NOTHING TO DO: handle by ORM
+
+sale         / sale.order.line          / analytic_distribution_stored_char (char): NEW isfunction: function, stored
+# NOTHINGH TO DO: handle by ORM
```
huguesdk: this is a new computed field (`analytic_distribution`) coming from the `analytic.mixin` model. its compute function is `_compute_analytic_distribution()`. a fast precomputation could be useful.

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,56 @@
...
+        "default_invoice_email_template",
+        "ir.config_parameter",
+        env["ir.config_parameter"]
+        .search([("key", "=", "sale.default_invoice_email_template")], limit=1)
+        .id,
+        noupdate=True,
+    )
+    openupgrade.update_module_moved_fields(
+        env.cr, "sale.order.line", _moved_fields, "sale_stock", "sale"
+    )
```
huguesdk: this is not needed. [the docstring of `openupgrade.update_module_moved_fields()`](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L3011) says:

> No need to use this method: moving the XMLID is covered in
> Odoo and OpenUpgrade natively.

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+
+sale         / sale.order.line          / analytic_tag_ids (many2many)  : DEL relation: account.analytic.tag
+# NOTHINGH TO DO: not used any more
+
+sale         / sale.order.line          / product_type (selection)      : previously in module sale_stock
+# DONE: pre-migration: moved from sale_stock to sale
+
+sale         / sale.order.line          / qty_delivered (float)         : not a function anymore
+# NOTHING TO DO: No database change
```
huguesdk:
```suggestion
# NOTHING TO DO: only the inverse compute function was removed
```

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+# NOTHING TO DO: not used any more, this field is useless since the new compute method from 13.0
+
+---XML records in moduldatabasee 'sale'---
+NEW ir.actions.act_window: sale.action_open_sale_payment_provider_onboarding_wizard
+DEL ir.actions.act_window: sale.action_open_sale_onboarding_payment_acquirer_wizard
+DEL ir.actions.act_window: sale.action_sale_order_form_view
+# NOTHING TO DO: noupdate="0" records
+
+ir.actions.server: sale.model_sale_order_action_share (noupdate switched)
+# DONE: pre-migration: noupdate switched to 1
```
huguesdk:
```suggestion
# DONE: pre-migration: noupdate switched to 0
```

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,93 @@
...
+DEL ir.model.access: sale.access_report_all_channels_sales
+DEL ir.model.access: sale.access_sale_payment_acquirer_onboarding_wizard
+DEL ir.model.access: sale.access_sale_report_manager
+# NOTHING TO DO
+
+ir.model.constraint: sale.constraint_sale_order_date_order_conditional_required (changed definition: is now 'check((state in('sale','done') and date_order is not null) or state not in('sale','done'))' ('check((state in('sale','done') and date_order is not null) or state not in('sale','done') )'))
+# DONE: pre-migration: dropped constraint and let ORM add it again with the new definition
+
+NEW ir.rule: sale.sale_payment_provider_onboarding_wizard_rule (noupdate)
+DEL ir.rule: sale.sale_payment_acquirer_onboarding_wizard_rule (noupdate)
```
huguesdk, line 75: the model, the name and the xml id has changed. it is better to simply delete `sale.sale_payment_acquirer_onboarding_wizard_rule`. the new rule will be created automatically anyway.

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,56 @@
...
+        "sale.payment.acquirer.onboarding.wizard",
+        "sale.payment.provider.onboarding.wizard",
+    ),
+]
+
+_renames_xmlids = [
+    (
+        "sale.sale_payment_provider_onboarding_wizard_rule",
+        "sale.sale_payment_acquirer_onboarding_wizard_rule",
+    ),
```
huguesdk: this does not work, as the old and new name are inverted. but anyway, it would be wrong (as explained in my comment on `upgrade_analysis_work.txt`: the only thing to do is to delete the old rule (`sale.sale_payment_acquirer_onboarding_wizard_rule`).

huguesdk (changes requested):
thanks for this great work! here are some needed changes.

hoangtiendung070797:
@huguesdk Thanks bro. I have learned more things from you lately :+1: . I will check it again

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
hoangtiendung070797: @huguesdk  `analytic_distribution_stored_char` fields is not found in source code. Can you show me where it is located?

daiduongnguyen-odoo: This the reason why we need the new analysis file because sometime it cause us some confusion
BTW, for the field `analytic_distribution_stored_char` which is in my lates analysis file it is `analytic_distribution`, i have created a dynamic method at https://github.com/OCA/OpenUpgrade/pull/4070/files#diff-90f733d1f5eba9ea514de973136064df20f321c35c019951653200d2c0e582d3R501-R515.
Please make a test with this situation then let me know it is working ok then i will decide to have it in the `openupgrade_160`

legalsylvain:
Hi @hoangtiendung070797 looks like your PR requires a rebase. 
Thanks !

chienandalu (commented):
Are you finishing this @hoangtiendung070797 ?

openupgrade_scripts/scripts/sale/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,97 @@
...
+---XML records in module 'sale'---
+NEW ir.actions.act_window: sale.action_open_sale_payment_provider_onboarding_wizard
+DEL ir.actions.act_window: sale.action_open_sale_onboarding_payment_acquirer_wizard
+DEL ir.actions.act_window: sale.action_sale_order_form_view
+# NOTHING TO DO: noupdate="0" records
+
+ir.actions.server: sale.model_sale_order_action_share (noupdate switched)
+# DONE: pre-migration: noupdate switched to 0
+
+NEW ir.config_parameter: sale.default_invoice_email_template (noupdate)
```
remi-filament, line 54: Hi @hoangtiendung070797 thanks for your work on this PR !
Actually this one is not new, it was already existing in v15.0 here : https://github.com/odoo/odoo/blob/7d90bce4c90cb77f2268332a5f1225cd884ac674/addons/sale/models/res_config_settings.py#L59-L65 (except it was not loaded from data but directly in Python model).

Obviously when migrating I get the following errors : 
```
2023-10-12 12:15:55,539 65829 INFO lf_20231011_v16mig odoo.modules.loading: loading sale/data/sale_data.xml 
2023-10-12 12:15:55,556 65829 ERROR lf_20231011_v16mig odoo.sql_db: bad query: INSERT INTO "ir_config_parameter" ("create_date", "create_uid", "key", "value", "write_date", "write_uid") VALUES ('2023-10-12 12:15:52.952317', 1, 'sale.default_invoice_email_template', '11', '2023-10-12 12:15:52.952317', 1) RETURNING "id"
ERROR: duplicate key value violates unique constraint "ir_config_parameter_key_uniq"
DETAIL:  Key (key)=(sale.default_invoice_email_template) already exists.
```

You should probably add a XML id on that parameter in pre_migration using openupgrade.add_xmlid() function (https://github.com/OCA/openupgradelib/blob/d0aee908031876a95dc36c7a9a2c847186155aa2/openupgradelib/openupgrade.py#L1288)

marielejeune:
Hi @hoangtiendung070797 , do you plan to rebase and finish this PR? We'll need it for our migration script. Thanks.

daiduongnguyen-odoo:
> Hi @hoangtiendung070797 , do you plan to rebase and finish this PR? We'll need it for our migration script. Thanks.

Hi , im in charge now, i will create a new PR

PR #4012, opened 2023-07-03
[15.0][OU-FIX] payment_transfer: don't squash pending_msg

chienandalu:
If we reload this aquirer default pending_msg we'll have all the company's journal bank accounts listed: https://github.com/OCA/OCB/blob/bf664b5bb58cf10e105ed7394b06c9a62a1af502/addons/payment_transfer/models/payment_acquirer.py#L55

It shouldn't be an issue to keep the old description.

cc @Tecnativa TT44190

FYI @pedrobaeza @MiquelRForgeFlow @legalsylvain


legalsylvain (approved):
make sense.

PR #4014, opened 2023-07-03
[16.0][OU-ADD] sale_management

huguesdk:
the renaming of the `sale_management.menu_product_attribute_action` menu item to `sale.menu_product_attribute_action` has to be done in the `sale` module because `sale_management` depends on it, so it is loaded before this module and creates the menu item if it doesn’t exist.


legalsylvain:
/ocabot migration sale_management

<dependency>Depends on : 

- sale : #4240</dependency>

remytms (approved):
LGTM

pedrobaeza (approved):

PR #4015, opened 2023-07-03
[16.0][OU-ADD] sale_stock

huguesdk:
i’m not sure whether this is enough or whether [`sale.order.delivery_status`](https://github.com/OCA/OCB/blob/16.0/addons/sale_stock/models/sale_order.py#L76) needs a fast precomputation. if it does, has anyone an idea on how to compute the aggregate value in an elegant way?

edit: added precomputation.


legalsylvain:
/ocabot migration sale_stock

<dependency>Depends on : 

- stock_account : #3971</dependency>

hoangtiendung070797:
@huguesdk I think it's necessary to precompute because the number of records in sale.order will be large, and its logic is also quite simple. I have implemented it here for you to consider, and one of the two PRs might be closed.

openupgrade_scripts/scripts/sale_stock/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,15 @@
+---Models in module 'sale_stock'---
+---Fields in module 'sale_stock'---
+sale_stock   / sale.order               / delivery_status (selection)   : NEW selection_keys: ['full', 'partial', 'pending'], isfunction: function, stored
```
hoangtiendung070797, line 3: @huguesdk I think it's necessary to precompute because the number of records in sale.order will be large, and its logic is also quite simple. I have implemented it here https://github.com/OCA/OpenUpgrade/pull/4057 for you to consider, and one of the two PRs might be closed.

huguesdk: thanks @hoangtiendung070797 for answering my question and providing a solution. i added the precomputation using a different query which should be more efficient. i also used `openupgrade.add_fields()` instead of altering the table directly, as is done in [the documentation](https://oca.github.io/OpenUpgrade/use_cases/field_fast_precomputation.html). what do you think?

happy 26th birthday! (guessed from your username) 😉

hoangtiendung070797: @huguesdk Yes, today is my 26th birthday. Thank you very much :smile:

hoangtiendung070797: @huguesdk your query is very short, i will try its effectiveness and close my PR. Thank you

openupgrade_scripts/scripts/sale_stock/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,57 @@
...
+            (
+                "delivery_status",
+                "sale.order",
+                False,
+                "selection",
+                False,
+                "sale_stock",
+            )
+        ],
+    )
```
hoangtiendung070797, line 20:
```suggestion
    openupgrade.add_fields(
        env,
        [
            (
                "delivery_status",
                "sale.order",
                 "selection",
                "character varying",
                "sale_stock",
                False,
            )
        ],
    )
```

@huguesdk  I think the order should be: field name, model name, SQL table name, field type, SQL field type, module name, (optional) initialization value
Maybe i'm wrong, can you check it again

huguesdk: [here](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2816) is the definition of `openupgrade.add_fields()`. the way the arguments work are not that intuitive. the only mandatory elements are the field name (0), the model name (1), the field type (3) and the module name (5). if the others are false, they will be determined automatically. i think the code is correct as it is. with your suggestion, it would try to add a column of type `sale_stock` to the `selection` table, for the `False` module. 🤔

hoangtiendung070797: @huguesdk ok bro :+1:

remytms (approved):
LGTM

remytms:
@hoangtiendung070797 Can you make a new review on this PR ? I think that all your comments was fixed. :)

pedrobaeza (approved):

PR #4018, opened 2023-07-04
[16.0][OU-ADD] google_gmail: Migration done

daiduongnguyen-odoo:


pedrobaeza:
The column `use_google_gmail_service` in `fetchmail.server` may not exists if `fetchmail_gmail` was not installed, so you should check the existence. Do you agree to switch to #4481 that contains this fix and the work file is rationalized by each change?

daiduongnguyen-odoo:
> The column `use_google_gmail_service` in `fetchmail.server` may not exists if `fetchmail_gmail` was not installed, so you should check the existence. Do you agree to switch to #4481 that contains this fix and the work file is rationalized by each change?

Hello @pedrobaeza , of course why not , please proceed what you have to do

pedrobaeza:
If you prefer, you can perform the `if` here and merge this one. I don't want to nullify your work :)

daiduongnguyen-odoo:
> If you prefer, you can perform the `if` here and merge this one. I don't want to nullify your work :)

Done, thank you sir

pedrobaeza (approved):
No need to put `= true` in queries, but it's just a detail.

PR #4028, opened 2023-07-04
[16.0][OU-FIX] portal: delete translations after loading data changes

marielejeune:


legalsylvain (approved):
Thanks !

pedrobaeza:
We are incorporating in https://github.com/OCA/openupgradelib/pull/328 a new parameter for deleting only certain fields translations, which makes sense for all these templates for not removing customizations on non needed fields.

This PR is green because `ir_translation` table exists from v15, but actually there's no support for v16 yet.

PR #4031, opened 2023-07-04
[16.0][OU-ADD] sale_project_stock: Nothing to do

diamondnguyenkc:
Nothing to do


legalsylvain:
/ocabot migration sale_project_stock

<dependency></dependency>

PR #4032, opened 2023-07-04
[16.0][OU-ADD] sale_expense_margin: Nothing to do

diamondnguyenkc:
Nothing to do


legalsylvain:
/ocabot migration sale_expense_margin

<dependency></dependency>

PR #4038, opened 2023-07-04
[15.0][OU-ADD] website_event_questions: Migration scripts

pedrobaeza:
@Tecnativa TT42044


openupgrade_scripts/scripts/website_event_questions/15.0.1.2/noupdate_changes.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="ir_rule_event_question_answer_event_user" model="ir.rule">
-    <field name="groups" eval="[(4, ref('event.group_event_registration_desk'))]"/>
+    <field name="groups" eval="[(3, ref('event.group_event_user'), (4, ref('event.group_event_registration_desk'))]"/>
```
legalsylvain: If I understand correctly : 
V14 :         ``<field name="groups" eval="[(4, ref('event.group_event_user'))]"/>``
V15 :         ``<field name="groups" eval="[(4, ref('event.group_event_registration_desk'))]"/>``

So yes, your change is correct. However, why the upgrade_analysis module generate that line.
It should generate 

        ``<field name="groups" eval="[(6, 0, [ref('event.group_event_registration_desk')])]"/>``

don't you think ? 
It will avoid openupgrade developper to make such changes as you had to do

pedrobaeza: That change is not conservative, as it replaces everything and don't preserve other possible manual groups.

It can be however improved for generating the `(3,` syntax. The only problem is that it should infer that the field is a m2m.

legalsylvain: :+1:

PR #4041, opened 2023-07-05
[16.0][OU-ADD] website_google_map: Nothing to do

daiduongnguyen-odoo:
No analyst file, so i think nothing to do


legalsylvain:
/ocabot migration website_google_map

<dependency></dependency>

chienandalu (approved):

PR #4042, opened 2023-07-05
[16.0][IMP] update analyst file

daiduongnguyen-odoo:
Regenerate analyst because some of them are oudated (eg: analytic)


daiduongnguyen-odoo:
@legalsylvain I regenerate analyst files here, if you have time please take a look.
One problem i have encountered when running analyst is if you install some of deprecated module like `payment_alipay` (this module is marked as deprecated in odoo 16) and then you hit button `generate record`, it will get errors because in that module have this code `https://github.com/odoo/odoo/blob/914870f2d4b9dbdf39cdde98fed3bb21ca0074cf/addons/payment_alipay/__init__.py#L12`

openupgrade_scripts/scripts/hr/16.0.1.1/upgrade_analysis.txt
```diff
@@ -9,6 +9,7 @@ hr           / hr.contract.type         / sequence (integer)            : NEW
 hr           / hr.department            / master_department_id (many2one): NEW relation: hr.department, isfunction: function, stored
 hr           / hr.department            / parent_path (char)            : NEW
 hr           / hr.department            / plan_ids (one2many)           : NEW relation: hr.plan
+hr           / hr.employee              / currency_id (many2one)        : previously in module hr_timesheet
```
daiduongnguyen-odoo, line 12: This one correct, manual checked

openupgrade_scripts/scripts/account/16.0.1.2/upgrade_analysis.txt
```diff
@@ -101,7 +101,7 @@ account      / account.chart.template   / complete_tax_set (boolean)    : DEL
 account      / account.chart.template   / use_storno_accounting (boolean): NEW hasdefault: default
 account      / account.journal          / default_account_type (many2one): relation is now 'False' ('account.account.type') [nothing to do]
 account      / account.journal          / default_account_type (many2one): type is now 'char' ('many2one')
-account      / account.journal          / payment_sequence (boolean)    : NEW hasdefault: default
+account      / account.journal          / payment_sequence (boolean)    : NEW hasdefault: compute
```
daiduongnguyen-odoo, line 104: This one correct too, manual checked

daiduongnguyen-odoo:
The red CI i have a PR fix here https://github.com/OCA/openupgradelib/pull/335

openupgrade_scripts/scripts/base/16.0.1.3/upgrade_general_log.txt
```diff
@@ -119,214 +122,33 @@ new model purchase.order.group [module purchase_requisition]
...
-fetchmail_outlook / fetchmail.server         / microsoft_outlook_access_token_expiration (integer): module is now 'microsoft_outlook' ('fetchmail_outlook')
-fetchmail_outlook / fetchmail.server         / microsoft_outlook_refresh_token (char): module is now 'microsoft_outlook' ('fetchmail_outlook')
-fetchmail_outlook / fetchmail.server         / microsoft_outlook_uri (char)  : module is now 'microsoft_outlook' ('fetchmail_outlook')
-fetchmail_outlook / fetchmail.server         / use_microsoft_outlook_service (boolean): DEL
----XML records in module 'fetchmail_outlook'---
-DEL ir.ui.view: fetchmail_outlook.fetchmail_server_view_form
-ERROR: module not in list of installed modules:
 ---Models in module 'gift_card'---
-obsolete model gift.card
+obsolete model gift.card (renamed to loyalty.card in module loyalty)
```
daiduongnguyen-odoo, line 760: This one too XD

openupgrade_scripts/scripts/stock/16.0.1.1/upgrade_analysis.txt
```diff
@@ -1,11 +1,11 @@
 ---Models in module 'stock'---
-obsolete model stock.location.route
-obsolete model stock.production.lot
+obsolete model stock.location.route (renamed to stock.route)
+obsolete model stock.production.lot (renamed to stock.lot)
 new model lot.label.layout [transient]
 new model picking.label.type [transient]
-new model stock.lot
+new model stock.lot (renamed from stock.production.lot)
```
daiduongnguyen-odoo, line 6: It seem like the tool are detecting just the model is renamed and do not need to have extra `del` and `new` line below so i assumed this one is correct right?

pedrobaeza: Yes, it's correct, and it's done due to https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/apriori.py#L74

daiduongnguyen-odoo: Oh i understood, so before this PR we haven't had that so it has extra line below

openupgrade_scripts/scripts/event_booth_sale/16.0.1.1/upgrade_analysis.txt
```diff
@@ -4,6 +4,9 @@ event_booth_sale / event.booth              / price (float)                 : is
 event_booth_sale / event.type.booth         / price (float)                 : is now stored
 event_booth_sale / product.template         / detailed_type (False)         : selection_keys is now '['consu', 'course', 'event', 'event_booth', 'product', 'service']' ('['consu', 'event', 'event_booth', 'gift', 'product', 'service']')
 ---XML records in module 'event_booth_sale'---
+event.booth.category: event_booth.event_booth_category_premium (noupdate) (noupdate switched)
+event.booth.category: event_booth.event_booth_category_standard (noupdate) (noupdate switched)
+event.booth.category: event_booth.event_booth_category_vip (noupdate) (noupdate switched)
```
legalsylvain: I see no changes between : 
https://github.com/odoo/odoo/blob/15.0/addons/event_booth_sale/data/event_booth_category_data.xml
https://github.com/odoo/odoo/blob/16.0/addons/event_booth_sale/data/event_booth_category_data.xml
(last changes done sept 2021, in both cases)

Are you sure you use head version for version 15 and version 16 ?

legalsylvain (changes requested):
Hi @duong77476 Thanks a lot for taking the time to update the files.

1) what a huge diff ! What a stable release ;-) !

2) a blocking point inline. Could you take a look ?

openupgrade_scripts/scripts/event_booth_sale/16.0.1.1/upgrade_analysis.txt
daiduongnguyen-odoo: You are correct, i don't know why it has that, I use the head version only lack a few commit but i don't thing that is a problem :(

legalsylvain: Hum. That's annoying. did you used a fresh DB also ?

daiduongnguyen-odoo: Yes, i did, brand new DB. I just found a commit that they have changed to `noupdate` in orignal data, see `https://github.com/odoo/odoo/commit/67b149fc3a8dffea37de59c9579d552c559b11f7` but it shouldn't be like this right, maybe i'll delete it anyway :(

legalsylvain: delete what ? commit ? 

And Something I don't understand : the two links I mentionned said that the last commit was in 2021. But your commit https://github.com/odoo/odoo/commit/67b149fc3a8dffea37de59c9579d552c559b11f7 has been done in 2022. totally lost.

daiduongnguyen-odoo: No i mean delete those 3 lines
`event.booth.category: event_booth.event_booth_category_premium (noupdate) (noupdate switched)
event.booth.category: event_booth.event_booth_category_standard (noupdate) (noupdate switched)
event.booth.category: event_booth.event_booth_category_vip (noupdate) (noupdate switched)`
Or let's me rerun again. Actually i just want to update some of major module like `account` but why not run all modules right :v

legalsylvain: > No i mean delete those 3 lines

OK. Well, I'd like to understand why there 3 false positive are here. Otherwise, we have a risk to add other false positive. we really have to make sure that these analysis files are correct. that's what the developers work on.

daiduongnguyen-odoo: Ok i understood, let's rerun again and see what happen, thank for spending your time to have a look .

daiduongnguyen-odoo: Just make not sure i did not skip any step, i follow step by step in this link https://oca.github.io/OpenUpgrade/analyse.html , what possibly i can miss here :((

daiduongnguyen-odoo:
@legalsylvain Done, i think there is something wrong this code in `server-tools`
`https://github.com/OCA/server-tools/blob/182257bf8670965791abb6c69aeb5ad942f43fb3/upgrade_analysis/wizards/upgrade_generate_record_wizard.py#L94`

As when i install all Odoo SA modules, then `generate record` , many xml id are obviously noupdate but they don't (Note this if you just install some modules not all because i can not identify the root cause then it works fine). So to deal with that i have to run the `query` in `https://github.com/OCA/server-tools/blob/182257bf8670965791abb6c69aeb5ad942f43fb3/upgrade_analysis/wizards/upgrade_generate_record_wizard.py#L94` by hand then start analyzing
Please take a look , thanks in advance

legalsylvain:
Unfortunately, I'm not expert of that module @duong77476. @StefanRijnhart Do you have some time to take a look on that topic ?

legalsylvain:
Propose to merge this PR to use the latest analysis files.
Any pov, @remytms, @pedrobaeza, @MiquelRForgeFlow ?

pedrobaeza:
I let @MiquelRForgeFlow to check, that is the expert on this.

MiquelRForgeFlow:
My issue is: many PRs have already been done, and once this is merged, I don't know if the people who made those PRs will update the analysis work file correspondingly.

daiduongnguyen-odoo:
> My issue is: many PRs have already been done, and once this is merged, I don't know if the people who made those PRs will update the analysis work file correspondingly.

Yes, i do aware of that issue, i just want to work latest update file especially with my recent PR related to `analytic` and `account` , in those PR i use this analysis file to ensure the latest change will be handled.

legalsylvain:
> My issue is: many PRs have already been done, and once this is merged, I don't know if the people who made those PRs will update the analysis work file correspondingly.

this issue is valid. however, if we wait, the problem will be bigger. 
how to get ahead ?
If we wanted to get things right, we'd have to review all PRs marked as done / nothing to do and which have an analysis file that has changed, to update the work file.
That sounds like a lot of work.

MiquelRForgeFlow:
> how to get ahead ?

I will review and check the changes when I have a bit of free time.

> That sounds like a lot of work.

Unfortunately, yes.

openupgrade_scripts/scripts/event_booth_sale/16.0.1.1/upgrade_analysis.txt
pedrobaeza: The problem here is that this data is only adding more values to the existing record that was defined in the module `event_booth`:

https://github.com/odoo/odoo/blob/aa90123b666784aea039ebb3c1febf0be81f8362/addons/event_booth/data/event_booth_category_data.xml#L4C77-L4C77

and in that definition, the record is noupdate.

This is then a problem of the `upgrade_analysis` module. I'm going to check if I can fix it, although the expert here is @MiquelRForgeFlow

pedrobaeza: It seems this query has not been executed in 15.0:

https://github.com/OCA/server-tools/blob/cd2f828fed05a65aae0dc18eef26bc6f2eb31d39/upgrade_analysis/wizards/upgrade_generate_record_wizard.py#L95

I got my session kicked out though when finished (not sure if this is normal or not), and maybe this side effect it's because one of the flushes overwrites these values. @StefanRijnhart @hbrunn you may know what is happening here.

I'm going to execute manually the code and perform the analysis in that condition.

pedrobaeza:
On #4180 I have solved the problems found here, so superseding this one. Thanks for starting the update.

Issue #3979, opened 2023-07-02
[RFC] new tools to mark PR as blocked or OK automatically


legalsylvain:
Hi @OCA/openupgrade-maintainers 

Since I've been working on Openupgrade, I'm regularly annoyed to see PRs reviewed (or very simple), but not merged.
the problem is that to merge PRs in OpenUpgrade, you have to follow the dependency ordering. (to merge "product", you need to have merged "uom").

so I spent a few hours making a script that automatically scans all OpenUpgrade open PRs, marking them as "Blocked by dependency" or "Dependency OK". (new green label).

As a result,
- I merge #3964 (done by @huguesdk some days ago)
- I run the script
- Automatically the PR #3782 (done by @marielejeune, 3 monthes ago) is marked as "Dependency OK"

![image](https://github.com/OCA/OpenUpgrade/assets/3407482/4f6d9c8f-590a-4a0f-ab8b-f937e696531d)

So once portal is merged, the following modules can be marked as Dependency OK (if all the other dependencies are OK): 
mail_group, website, test_mail_full, digest, mass_mailing_sms, auth_password_policy_portal, account, event, project, website_crm_partner_assign, payment, website_payment, portal_rating, auth_totp_portal
etc ...

- you can click here https://github.com/OCA/OpenUpgrade/pulls?q=is%3Apr+is%3Aopen+label%3A%22Dependency+OK%22 to review / merge PRs that are ready in a "dependency meaning"

**note:**

For the moment, the tools is available here : https://github.com/legalsylvain/openupgrade-extra-tools as a simple script. I'll execute it regularly. You can do it also, (as soon as you have maintainer accreditation).

CC : @vdewulf, @remytms

huguesdk:
great! simple and effective. thanks @legalsylvain!

remytms:
great tool ! :rocket:

legalsylvain:
Hi @remytms and @huguesdk. Thanks for your feedbacks ! 

FYI, I just updated the script to add another feature : 
- For each opened PRs that are "blocked", enumerate the list of the PRs that we have to review first.
For exemple : in the sale_stock PR (https://github.com/OCA/OpenUpgrade/pull/4015#issuecomment-1618878824), 
add a link to sale and stock_account PR.

![image](https://github.com/OCA/OpenUpgrade/assets/3407482/728a50b8-7145-4887-8823-5ce550573bd7)

Note : the text will be updated, each time I run the script and if the related PRs has been merged.

[...]

PR #4044, opened 2023-07-05
[15.0][OU-ADD] mass_mailing_event_track: Nothing to do

pedrobaeza:
Even no analysis file.

@Tecnativa TT41016


pedrobaeza:
It's not blocked by dependency, @legalsylvain. It's simply a bad interaction, as we do a trick in the CI installing all modules in the previous version (and thus, having the records), and then marking them as `uninstalled` by SQL. I'm going the straight way not doing the conflicting XML-IDs renaming. It was just completeness effort.

legalsylvain:
I didn't put "Blocked by Dependency". the script I ran did it ! 
(https://github.com/OCA/OpenUpgrade/issues/3979, feedback welcome BTW)

It is based on the parsing of the file https://github.com/OCA/OpenUpgrade/blob/15.0/docsource/modules140-150.rst. For instance, ``website_event_track`` is marked as TODO.

After investigation, it is because you forgot to update the coverage file in this PR. https://github.com/OCA/OpenUpgrade/pull/3751

This trivial PR should fix the problem : https://github.com/OCA/OpenUpgrade/pull/4049

pedrobaeza:
OK, thank you for checking.

PR #4050, opened 2023-07-05
[15.0][OU-FIX] mass_mailing: editable legacy mailings

chienandalu:
Lecacy mailing bodies don't work in v15 as the wrapper (or the absence of it) doesn't hook properly with the web editor. With this new script we rewrap them so they're usable in the new versions.

cc @Tecnativa TT42980 TT26694

please review @pedrobaeza


pedrobaeza (approved):
Tested and working

pedrobaeza:
Merging. For the next time, take into account that this is `[OU-IMP]` or `[OU-FIX]`, not `[OU-ADD]`, as the migration script already exists.

PR #4053, opened 2023-07-06
[16.0] [OU-ADD] mass_mailing

locdang8:


openupgrade_scripts/scripts/mass_mailing/16.0.2.5/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,113 @@
...
+mass_mailing / mailing.mailing          / _order                        : _order is now 'calendar_date DESC' ('sent_date DESC')
+mass_mailing / mailing.mailing          / copyvalue (char)              : DEL
+mass_mailing / mailing.mailing          / favorite (boolean)            : NEW
+mass_mailing / mailing.mailing          / favorite_date (datetime)      : NEW isfunction: function, stored
+mass_mailing / mailing.mailing          / mailing_filter_id (many2one)  : NEW relation: mailing.filter, hasdefault: compute
+mass_mailing / mailing.mailing          / model_object_field (many2one) : DEL relation: ir.model.fields
+mass_mailing / mailing.mailing          / null_value (char)             : DEL
+mass_mailing / mailing.mailing          / sub_model_object_field (many2one): DEL relation: ir.model.fields
+mass_mailing / mailing.mailing          / sub_object (many2one)         : DEL relation: ir.model
+# NOTHING TO DO
```
MiquelRForgeFlow, line 23: I think you should create mailing.filter records for each sub_object in mailing.mailing and then fill mailing_filter_id.

locdang8: but sub_object is not store, I don't know how to create mail_filter from it

cvinh: > but sub_object is not store, I don't know how to create mail_filter from it

I think you can extract sub_object data from ir.model and ir.model.field

pedrobaeza: `mailing.filter` are like search favorites, but inside the mailing, so no need to perform anything in this table. It's a new feature.

pedrobaeza (approved):
I have pushed an extra commit (apart from rebasing) for the missing noupdate changes load. Apart from that, it seems OK.

PR #4055, opened 2023-07-06
[16.0] [OU-FIX] calendar: misstake in _translations_to_delete list

NgoHoangHung:


daiduongnguyen-odoo (approved):

legalsylvain (approved):
Make sense.
ref noupdate file : https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/calendar/16.0.1.1/noupdate_changes.xml#L3

PR #4058, opened 2023-07-06
[14.0][OU-IMP] website_event: remove obsolete website views

chienandalu:
The `is_online` field is gone, so we don't want those views copies in our websites. Let's reset them.

cc @Tecnativa TT44287

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/website_event/14.0.1.1/post-migrate.py
+++ b/openupgrade_scripts/scripts/website_event/14.0.1.1/post-migrate.py
@@ -2,7 +2,20 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from openupgradelib import openupgrade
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This solution is not ideal due to several reasons:

- There might be children views and the removal will fail. Luckily (or not, depending on the PoV), there's no `oe_structure` section for inserting custom content that creates children view, but you can have as well extension modules that inherits this view, and that has been copied for the specific website.
- This is fixing the problem in a too heavy way, as it's losing also any customization done on the website.

If we can't remove just that field here in OU, I would say that this fix should be decided DB by DB removing manually the views. Another more elaborate approach may be to compare the master view with this one, and if they are the same (at XML level, which not sure if it's feasible), then remove them, as we are sure that no customization has been done, and the removal should be of all the hierarchy.

chienandalu:
Ok, let's close it then

PR #4059, opened 2023-07-06
[15.0][OU-FIX] sale_coupon: mail template translations

chienandalu:
We must remove the translations before the new xml_id is loaded so no orphan terms are left behind.

cc @Tecnativa TT44289

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/coupon/15.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/coupon/15.0.1.0/pre-migration.py
@@ -5,6 +5,15 @@ from openupgradelib import openupgrade
 
 @openupgrade.migrate()
 def migrate(env, version):
+    # Delete old template translations before the new template is renamed to ensure that
+    # we remove the terms properly
+    openupgrade.delete_record_translations(
+        env.cr,
+        "coupon",
+        [
+            "mail_template_sale_coupon",
+        ],
+    )
     openupgrade.rename_xmlids(
         env.cr,
         [("coupon.mail_template_sale_coupon", "sale_coupon.mail_template_sale_coupon")],
--- a/openupgrade_scripts/scripts/sale_coupon/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/sale_coupon/15.0.1.0/post-migration.py
@@ -6,10 +6,3 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "sale_coupon", "15.0.1.0/noupdate_changes.xml")
-    openupgrade.delete_record_translations(
-        env.cr,
-        "sale_coupon",
-        [
-            "mail_template_sale_coupon",
-        ],
-    )


MiquelRForgeFlow (changes requested):
No. In coupon module, an xmlid rename has been done instead of deleting the old xmlid. Thus, what you are doing is not working.

chienandalu:
Thanks @MiquelRForgeFlow  Then I'm deleting them before that

openupgrade_scripts/scripts/sale_coupon/15.0.1.0/post-migration.py
```diff
@@ -6,10 +6,3 @@
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "sale_coupon", "15.0.1.0/noupdate_changes.xml")
```
MiquelRForgeFlow, line 8: Comment this line. The file is not needed because only modifies a template, but the template is not noupdate anymore.

openupgrade_scripts/scripts/coupon/15.0.1.0/pre-migration.py
```diff
@@ -5,6 +5,15 @@
 
 @openupgrade.migrate()
 def migrate(env, version):
+    # Delete old template translations before the new template is renamed to ensure that
+    # we remove the terms properly
+    openupgrade.delete_record_translations(
```
MiquelRForgeFlow, line 10: No. Instead, call the `set_xml_ids_noupdate_value()` method with value = False.

MiquelRForgeFlow (changes requested):
After looking into it, I realized what is the problem, which is that this template changed its noupdate condition in addition from changing module. That was an oversight from my part.

openupgrade_scripts/scripts/coupon/15.0.1.0/pre-migration.py
MiquelRForgeFlow: In fact, to be more purist, you should do it in pre of sale_coupon.

chienandalu: That could be an option, but correct if I'm wrong:

If we set `set_xml_ids_noupdate_value(env, "coupon", ["mail_template_sale_coupon"])`, as the template isn't anymore in the module, it would be unlinked in the `coupon` module update and thus recreated as a new by `sale_coupon` without the annoyance of loading anything from a record.

MiquelRForgeFlow: It wouldn't be unlinked in `coupon` update due to the xmlid rename that moves the record to `sale_coupon`. Besides, here in OU we never take the approach of deleting a noupdate template in pre (and wait for its recreation in the module update) for a reason. Instead we always update the template in post with the load_data and then delete its translations.

pedrobaeza: @chienandalu is from today on holidays during 2 weeks, so if you can supersede it, I will appreciate it.

MiquelRForgeFlow: After checking correctly the code in both versions, I realized that the `upgrade_analysis_work.txt` was not up to date with the `upgrade_analysis.txt` due to old analysis. So, there was no change in the noupdate situation. Just that the template moved from one module to the other. Also, there were no changes inside the template.

Superseded in https://github.com/OCA/OpenUpgrade/pull/4083.

PR #4060, opened 2023-07-06
[16.0][FIX] crm: check falsy value in query

MiquelRForgeFlow:
...


legalsylvain (approved):
Indeed ! 
Thanks for catching it.

PR #4067, opened 2023-07-07
[16.0][OU-ADD] purchase_stock: migration

hoangtiendung070797:


legalsylvain:
/ocabot migration purchase_stock

<dependency>Depends on : 

- stock_account : #3971
- purchase : #4303</dependency>

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,73 @@
...
+        SET dest_address_id = FALSE
+        FROM stock_picking_type AS spt
+        JOIN stock_location AS sl
+        ON spt.default_location_dest_id = sl.id
+        WHERE po.picking_type_id = spt.id AND sl.usage != 'customer'
+        """,
+    )
+
+
+def migrate(env, version):
```
huguesdk, line 51:
```suggestion
@openupgrade.migrate()
def migrate(env, version):
```
otherwise `env` is a cursor, not an environment.

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,73 @@
...
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "purchase_stock.route_warehouse0_buy",
+        ],
+    )
```
huguesdk: this must not be deleted. it is seen as deleted and new by the analysis because the model name has changed from `stock.location.route` to `stock.route` (handled in the `stock` migration scripts). nothing needs to be done for this.

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def _compute_purchase_order_receipt_status(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE purchase_order
+        ADD COLUMN IF NOT EXISTS receipt_status character varying
+        """,
+    )
```
huguesdk: why alter the table directly instead of using `openupgrade.add_fields()`?

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,73 @@
...
+                    WHEN COUNT(sp.state)
+                        FILTER(WHERE sp.state = 'done') > 0
+                        THEN 'partial'
+                    ELSE 'pending'
+                END as receipt_status
+            FROM stock_picking as sp
+            JOIN purchase_order_stock_picking_rel as rel
+            ON rel.stock_picking_id = sp.id
+            GROUP BY 1
+            ORDER BY 1
```
huguesdk: interesting query. it looks somehow familiar. 😉 note that i changed the indentation in #4015 because `filter` refers to the `count` before it, so it makes more sense to keep them together.

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,73 @@
...
+        env.cr,
+        """
+        UPDATE purchase_order AS po
+        SET dest_address_id = FALSE
+        FROM stock_picking_type AS spt
+        JOIN stock_location AS sl
+        ON spt.default_location_dest_id = sl.id
+        WHERE po.picking_type_id = spt.id AND sl.usage != 'customer'
+        """,
+    )
```
huguesdk: [the compute method](https://github.com/OCA/OCB/blob/16.0/addons/purchase_stock/models/purchase.py#L77) sets the field to `False` (thus `null`) when these conditions are true. since its a new column, all values are already `null`. i ran the migration without calling this function, and the compute function is not called (i actually don’t know exactly why; i first thought that it was because the field was writable, but even when removing `readonly=False` from the field, it is not called). anyway, nothing needs to be done (even not add the field).

huguesdk (changes requested):

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
hoangtiendung070797: @huguesdk I'm not sure so I deleted it, thank you :smile:

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
hoangtiendung070797: @huguesdk  I referenced it from your PR :smile:.

openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py
hoangtiendung070797: @huguesdk  it's my coding habit, i think it's faster. i will try openupgrade.add_fields() :smile:

hoangtiendung070797: Thank you bro

huguesdk (approved):
lgtm. 👍

pedrobaeza (approved):
/ocabot merge nobump

PR #4079, opened 2023-07-10
[16.0][OU-ADD]  website_hr_recruitment: Nothing to do 

nguyenvietlam0640:


legalsylvain:
/ocabot migration website_hr_recruitment

<dependency>Depends on : 

- hr_recruitment : #4296</dependency>

pedrobaeza:
I have pushed an extra commit doing the needed pre-creation of the `job_details` column for avoiding a non-desired default value that will be shown on existing jobs.

PR #4084, opened 2023-07-10
[16.0] [OU-ADD] website_sale_stock: Nothing to do

NgoHoangHung:


legalsylvain:
/ocabot migration website_sale_stock

<dependency>Depends on : 

- sale_stock : #4015
- website_sale : TODO</dependency>

PR #4086, opened 2023-07-11
[16.0] [OU-ADD] website_event_track

nguyenvietlam0640:


legalsylvain:
/ocabot migration website_event_track

<dependency></dependency>

openupgrade_scripts/scripts/website_event_track/16.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+        """,
+    )
+
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE event_track_location
+        SET sequence = 10
+        """,
+    )
```
SirAionTech: The new column should have already been created by the module and have the default value of `10`, could you please check if this is actually needed?

SirAionTech (commented):
Thanks for the PR!

openupgrade_scripts/scripts/website_event_track/16.0.1.3/pre-migration.py
pedrobaeza: Indeed it is that way.

pedrobaeza:
I have pushed a rebased branch and fixing some things:

- No need to pre-create `sequence` column. It's done by the ORM.
- Translations of the fields changed in noupdate_changes.xml should be deleted.
- Some formatting on the work file.

Issue #4090, opened 2023-07-12
[13 -> 14] point_of_sale: full_product_name not initialized on pos.order.line


alexis-via:
In Odoo v14, a new field "full_product_name" has been added on pos.order.line. In the form view and tree view of pos.order.line, the product_id field is not displayed any more ; it has been replaced by full_product_name.
So, after migrating from v13 to v14 with openupgrade, you get:

![pos_order_full_product_name](https://github.com/OCA/OpenUpgrade/assets/1157917/fc12726c-2a4d-4393-8e45-70048ff21858)

Here is the commit that introduce this new field "full_product_name" https://github.com/odoo/odoo/commit/0c3fe8736d77763b4fb31fd76be1f41f1effd5c0
By default, Odoo writes in it the name of the product and it's variant caracteristics (and probably also the configuration params in case it's a configurable product).
This was spotted in the PR to add support for point_of_sale for v14 (https://github.com/OCA/OpenUpgrade/pull/3106) but was not fixed, cf comment by cvinh.

How could we support this:
- with good performance : as we are on POS order line, we can have millions of records
- lang: I guess it is written in the lang of the user of the POS => let's forget this, it is a problem even in prod use also.

We can't do it in direct SQL, because have don't have a char field with the product name with its caracteristics.
I'm sure devs with more experience than me on openupgrade will have an idea how to implement that with good perfs.
My only idea at the moment is:
- loop on all product.product with active_test=False
- for each product:
  . get the product name + caracteristics
  . search pos.order.line that have this product
  . UPDATE pos_order_line set full_product_name=xxx WHERE product_id=z
With this solution, you process product 1 by 1... it's better than processing pos.order.line 1 by 1... but it is still a bit slow.
But if nobody has a better idea, I will implement that.

PR #4091, opened 2023-07-12
[14.0][IMP] point_of_sale: full_product_name on pos.order.line

alexis-via:
Proposal to fix bug https://github.com/OCA/OpenUpgrade/issues/4090
If someone has a better idea to make it faster, feel free to improve it.


pedrobaeza:
That can be awfully slow. In this case, I would go for a not so bad solution that is to just put the template name, or mimic a bit the `display_name` in SQL, including the reference (if any) between brackets.

pedrobaeza:
Hehe, Miquel has read my mind.

PR #4092, opened 2023-07-12
[14.0][FIX] point_of_sale: full_product_name on pos.order.line

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4091.


pedrobaeza (approved):
This is not covering the variant with attributes case, but I think it's enough for past orders.

legalsylvain:
Fix : https://github.com/OCA/OpenUpgrade/issues/4090

alexis-via (approved):
Code review and test (with and without default_code). Everything ok.

PR #4093, opened 2023-07-13
[15.0][OU-ADD] website_sale: replace content in product custom_text

pilarvargas-tecnativa:
In v15 the html structure of the website_sale.product template has changed, so if you had customised the content of the "Terms and conditions" this is lost, so that this does not happen we copy the content of the website_sale.product template associated to a website, we create a copy of website_sale.product_custom_text which is the new template that adds this content, we associate the website id and the old content.

Example v13 (website_sale.product):
image

V15:
(website_sale.product)
image
(website_sale.product_custom_text)
image

cc @Tecnativa TT44424

@pedrobaeza can you review please?

### DIFF
--- a/openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
@@ -1,4 +1,38 @@
+import re
+
 from openupgradelib import openupgrade
+
+
+def replace_content_in_product_custom_text(env):
+    """Replace Terms and Conditions content in the new v15 template so as not to lose
+    content from previous versions if it has been customised."""
+    product_views = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<p class="text-muted"(.*?)<\/p>'
+        product_matches = re.findall(product_pattern, product_arch, re.DOTALL)
+
+        if product_matches:
+            product_custom_text_view = env["ir.ui.view"].search(
+                [("key", "=", "website_sale.product_custom_text")]
+            )
+            if product_custom_text_view:
+                product_custom_text_arch = product_custom_text_view.arch_db
+                product_custom_text_pattern = (
+                    r'<p class="text-muted h6 mt-3">(.*?)<\/p>'
+                )
+                product_custom_text_matches = re.findall(
+                    product_custom_text_pattern, product_custom_text_arch, re.DOTALL
+                )
+                new_arch = product_custom_text_arch.replace(
+                    product_custom_text_matches, product_matches
+                )
+                new_product_custom_text_view = product_custom_text_view.copy()
+                new_product_custom_text_view.write(
+                    {"website_id": view.website_id, "arch_db": new_arch}
+                )
 
 
 def set_visibility_product_attribute(env):
@@ -53,3 +87,4 @@ def migrate(env, version):
     )
     set_visibility_product_attribute(env)
     enable_price_filter_view(env)
+    replace_content_in_product_custom_text(env)


openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
```diff
@@ -1,6 +1,40 @@
...
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<p class="text-muted"(.*?)<\/p>'
+        product_matches = re.findall(product_pattern, product_arch, re.DOTALL)
+
+        if product_matches:
+            product_custom_text_view = env["ir.ui.view"].search(
+                [("key", "=", "website_sale.product_custom_text")]
```
pedrobaeza, line 19:
```suggestion
                [("key", "=", "website_sale.product_custom_text"), ("website_id", "=", False)]
```

and do this search only once at the beginning and don't perform the rest if it doesn't exist.

openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
```diff
@@ -1,6 +1,40 @@
...
+    """Replace Terms and Conditions content in the new v15 template so as not to lose
+    content from previous versions if it has been customised."""
+    product_views = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<p class="text-muted"(.*?)<\/p>'
+        product_matches = re.findall(product_pattern, product_arch, re.DOTALL)
+
```
pedrobaeza, line 16: Don't put empty lines inside a method.

openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
```diff
@@ -1,6 +1,40 @@
...
+                )
+                product_custom_text_matches = re.findall(
+                    product_custom_text_pattern, product_custom_text_arch, re.DOTALL
+                )
+                new_arch = product_custom_text_arch.replace(
+                    product_custom_text_matches, product_matches
+                )
+                new_product_custom_text_view = product_custom_text_view.copy()
+                new_product_custom_text_view.write(
+                    {"website_id": view.website_id, "arch_db": new_arch}
```
pedrobaeza, line 34: Pass these values directly in the `copy` method and save a write.

openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
```diff
@@ -1,6 +1,40 @@
+import re
+
 from openupgradelib import openupgrade
 
 
+def replace_content_in_product_custom_text(env):
```
pedrobaeza, line 6: Put a more descriptive method name (and you can also put docstring):

```suggestion
def _extract_custom_product_page_term_conditions(env):
```

openupgrade_scripts/scripts/website_sale/15.0.1.1/post-migration.py
```diff
@@ -1,6 +1,40 @@
...
+                product_custom_text_pattern = (
+                    r'<p class="text-muted h6 mt-3">(.*?)<\/p>'
+                )
+                product_custom_text_matches = re.findall(
+                    product_custom_text_pattern, product_custom_text_arch, re.DOTALL
+                )
+                new_arch = product_custom_text_arch.replace(
+                    product_custom_text_matches, product_matches
+                )
+                new_product_custom_text_view = product_custom_text_view.copy()
```
pedrobaeza, line 32: Shouldn't this only be done in case the terms are not the standard ones?

pilarvargas-tecnativa: With the change made this will only run in case the terms and conditions are not standard.
This migration script really needs to be done in v14 which is where the change is made to the templates.
You can continue to review the script in the new PR https://github.com/OCA/OpenUpgrade/pull/4096

pilarvargas-tecnativa:
PR closed because the script should be done in v14. https://github.com/OCA/OpenUpgrade/pull/4096

PR #4095, opened 2023-07-14
[15.0][OU-IMP] web: Auto-define company fields used in reports (report_footer + company_details)

victoralmau:
Auto-define company fields used in reports (`report_footer` + `company_details`)

@Tecnativa TT44012

### DIFF
--- a/openupgrade_scripts/scripts/web/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/web/15.0.1.0/post-migration.py
@@ -1,0 +1,33 @@
+# Copyright 2023-2024 Tecnativa - Víctor Martínez
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """It is important to set the appropriate report_footer and company_details values.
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't this be taken from the previous data instead of the defaults?

openupgrade_scripts/scripts/web/15.0.1.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2023-2024 Tecnativa - Víctor Martínez
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """It is important to set the appropriate report_footer and company_details values.
+    The report_footer field already existed before, but it only represented the 'extra'
+    text that was added after the company data (https://github.com/odoo/odoo/blob/14.0/addons/web/views/report_templates.xml#L367).
```
pedrobaeza: Please put permalinks, as this may change in the future.

pedrobaeza (approved):

PR #4096, opened 2023-07-19
[14.0][OU-ADD] website_sale: replace content in product custom_text

pilarvargas-tecnativa:
In v14 the html structure of the website_sale.product template has changed, so if you had customised the content of the "Terms and conditions" this is lost, so that this does not happen we copy the content of the website_sale.product template associated to a website, we create a copy of website_sale.product_custom_text which is the new template that adds this content, we associate the website id and the old content.

Example v13 (website_sale.product):
image

V14:
image

cc @Tecnativa TT44424

@pedrobaeza can you review please?

Note: This migration script had been done in a PR in v15 initially but the change in templates actually arises in v14 so the migration script is moved to this version. Initial PR https://github.com/OCA/OpenUpgrade/pull/4093

### DIFF
--- a/openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
@@ -1,8 +1,34 @@
 # Copyright (C) 2021 Open Source Integrators <https://www.opensourceintegrators.com/>
 # Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import re
 
 from openupgradelib import openupgrade
+
+
+def extract_custom_product_page_term_conditions(env):
+    """Replace Terms and Conditions content in the new v14 template so as not to lose
+    content from previous versions if it has been customised."""
+    product_custom_text_view = env.ref("website_sale.product_custom_text")
+    product_custom_text_arch = product_custom_text_view.arch_db
+    product_custom_text_pattern = r'<p class="text-muted">(.*?)<\/p>'
+    product_custom_text_content, *_ = re.findall(
+        product_custom_text_pattern, product_custom_text_arch, re.DOTALL
+    )
+    product_views = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<hr\s*/>\s*<p\s+class="text-muted">(.*?)</p>'
+        product_matches = re.findall(product_pattern, product_arch, re.DOTALL)
+        if product_matches:
+            new_arch = product_custom_text_arch.replace(
+                product_custom_text_content, product_matches[0]
+            )
+            product_custom_text_view.copy(
+                {"website_id": view.website_id.id, "arch_db": new_arch}
+            )
 
 
 @openupgrade.migrate()
@@ -57,3 +83,4 @@ def migrate(env, version):
         WHERE module = 'website_sale' and name = 'image_full'
         """,
     )
+    extract_custom_product_page_term_conditions(env)


MiquelRForgeFlow:
please, do `precommit run -a`

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,51 @@
...
+            standard_terms = """<hr />
+                <p class="text-muted">
+                30-day money-back guarantee<br />
+                Free Shipping in U.S.<br />
+                Buy now, get in 2 days
+                </p>"""
+            if all(
+                match.strip() == standard_terms.strip() for match in product_matches
+            ):
+                product_custom_text_view = env["ir.ui.view"].search(
```
pedrobaeza: You have missed this comment about putting this once and do anything if not found.

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,51 @@
...
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<hr />\n<p class="text-muted">(.*?)<\/p>'
+        product_matches = re.findall(product_pattern, product_arch, re.DOTALL)
+        if product_matches:
+            standard_terms = """<hr />
+                <p class="text-muted">
+                30-day money-back guarantee<br />
```
pedrobaeza: I'm afraid this doesn't serve, as this text may be translated by Odoo itself (if you load the website in the corresponding language). Let's write it always as a minor trade-off.

pilarvargas-tecnativa: I have removed it. In case the terms and conditions have not been modified a copy will be created as if they had been made, this is also not critical since if they are modified again in the future by a user no further copies of the view will be created, the changes will be reflected in the view with the associated website.
I can't think of anything universal to check since they are inserted directly into a view and simply create a copy with associated website when editing it but this is already covered in the search.

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
pilarvargas-tecnativa: Changes made, you can check. Thanks

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,40 @@
...
+    )
+    product_views = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<hr />\n<p class="text-muted">(.*?)<\/p>'
+        product_matches = re.findall(product_pattern, product_arch, re.DOTALL)
+        if product_matches:
+            new_arch = product_custom_text_arch.replace(
```
pedrobaeza, line 26: Is `replace` working with the result of `re.findall`?

pilarvargas-tecnativa: I tested in shell locally and it worked

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,40 @@
...
+import re
 
 from openupgradelib import openupgrade
 
 
+def extract_custom_product_page_term_conditions(env):
+    """Replace Terms and Conditions content in the new v14 template so as not to lose
+    content from previous versions if it has been customised."""
+    product_custom_text_view = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product_custom_text"), ("website_id", "=", False)]
```
chienandalu: If you want the main view, grab it by xml_id ref

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
chienandalu: Are you sure? Your `re.findall' will return a list of strings and that's not a type that `str.replace` will swallow

chienandalu (changes requested):

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
pilarvargas-tecnativa: Indeed not, my mistake. As there are two lists, accessing the first element of each one, the replace is done correctly.

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
pilarvargas-tecnativa: thanks!

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,38 @@
...
 
 from openupgradelib import openupgrade
 
 
+def extract_custom_product_page_term_conditions(env):
+    """Replace Terms and Conditions content in the new v14 template so as not to lose
+    content from previous versions if it has been customised."""
+    product_custom_text_view = env.ref("website_sale.product_custom_text")
+    if not product_custom_text_view:
+        return
```
chienandalu: The template is loaded by the module so you don't need to make this check
```suggestion
```

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,38 @@
...
+    product_custom_text_arch = product_custom_text_view.arch_db
+    product_custom_text_pattern = r'<p class="text-muted(.*?)<\/p>'
+    product_custom_text_matches = re.findall(
```
chienandalu: So you don't have to access by index (which isn't a very clean practice. Also rename the variable to something like `product_custom_text_content` that explains better what the code is doing
```suggestion
    product_custom_text_matches, *_ = re.findall(
```

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,38 @@
...
+        product_custom_text_pattern, product_custom_text_arch, re.DOTALL
+    )
+    product_views = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<hr />\n<p class="text-muted">(.*?)<\/p>'
```
chienandalu: Are you sure this is going to do a match? Blank spaces count as characters as well :)

pilarvargas-tecnativa: Changed and tested in the old template that performs the search correctly with this pattern, thank you.

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
pilarvargas-tecnativa: Right! much clearer, thank you!

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,10 +1,36 @@
...
+    product_custom_text_content, *_ = re.findall(
+        product_custom_text_pattern, product_custom_text_arch, re.DOTALL
+    )
+    product_views = env["ir.ui.view"].search(
+        [("key", "=", "website_sale.product"), ("website_id", "!=", False)]
+    )
+    for view in product_views:
+        product_arch = view.arch_db
+        product_pattern = r'<hr\s*/>\s*<p\s+class="text-muted">(.*?)</p>'
+        product_matches, *_ = re.findall(product_pattern, product_arch, re.DOTALL)
```
chienandalu: You can't do it here. If there's no match you'll raise unpack error.

chienandalu (changes requested):

openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
pilarvargas-tecnativa: thanks :)

chienandalu (approved):
Thanks :)

PR #4097, opened 2023-07-20
[15.0][OU-IMP] website: extract footer copyright company name

pilarvargas-tecnativa:
In versions prior to v15 the company name included in the copyright was done in a copy of the website.layout template. Now in v15 it is done in a new template that adds the content website.footer_copyright_company_name. This migration script is intended to extract the content of the old copyright and bring it into the new template.

cc @tecnativa TT44254

@pedrobaeza please review :)

### DIFF
--- a/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
@@ -1,4 +1,48 @@
+import re
+
 from openupgradelib import openupgrade
+
+
+def extract_footer_copyright_company_name(env):
+    """Replace Copyright content in the new v15 template so as not to lose
+    content from previous versions if it has been customised."""
+    main_copyright_view = env.ref("website.footer_copyright_company_name")
+    if not main_copyright_view:
+        return
+    main_copyright_arch = main_copyright_view.arch_db
+    main_copyright_pattern = r'<span class="o_footer_copyright_name mr-2">(.*?)<\/span>'
+    main_copyright_matches = re.findall(
+        main_copyright_pattern,
+        main_copyright_arch,
+        re.DOTALL,
+    )
+    website_layout_views = env["ir.ui.view"].search(
+        [("key", "=", "website.layout"), ("website_id", "!=", False)]
+    )
+    for view in website_layout_views:
+        website_layout_arch = view.arch_db
+        website_layout_pattern = (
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
```diff
@@ -1,6 +1,49 @@
+import re
+
 from openupgradelib import openupgrade
 
 
+def extract_footer_copyright_company_name(env):
+    """Replace Copyright content in the new v15 template so as not to lose
+    content from previous versions if it has been customised."""
+    footer_copyright_company_name_view = env["ir.ui.view"].search(
```
pedrobaeza: Why prepend all variables with `footer_copyright_company_name...` when the method is already having that name and the scope of the variables is local? Simplify them in this case: `main_footer_view`, `main_footer_arch`, etc...

openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
```diff
@@ -1,6 +1,49 @@
...
+    )
+    website_layout_views = env["ir.ui.view"].search(
+        [("key", "=", "website.layout"), ("website_id", "!=", False)]
+    )
+    for view in website_layout_views:
+        website_layout_arch = view.arch_db
+        website_layout_pattern = (
+            r'<span class="o_footer_copyright_name mr-2">(.*?)<\/span>'
+        )
+        website_layout_matches = re.findall(
```
pedrobaeza, line 27: Footer section was embedded in main layout view with no inheritance?

pilarvargas-tecnativa: Before v15 yes
https://github.com/odoo/odoo/blob/053a6eb1e38cce57d7abe202c50bdc2f6798be81/addons/website/views/website_templates.xml#L178

openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
pilarvargas-tecnativa: thanks!

pedrobaeza (approved):
Tested in a real migration

pedrobaeza (changes requested):
Wait, not really. Reviewing.

pedrobaeza (approved):
Now it's adjusted. The problem was that the view may exist.

PR #4098, opened 2023-07-21
[11.0][OU-IMP] account: Update attachment_use …

stefan-tecnativa:
…value

cc @Tecnativa TT44533

@pedrobaeza please review!

### DIFF
--- a/addons/account/migrations/11.0.1.1/post-migration.py
+++ b/addons/account/migrations/11.0.1.1/post-migration.py
@@ -44,6 +44,10 @@ def migrate_account_tax_cash_basis(env):
             WHERE company_id = rc.id AND tax_exigibility = 'on_payment')""")
 
 
+def migrate_account_report_invoice_with_payments(env):
+    env.ref("account.account_invoices").attachment_use = False
+
+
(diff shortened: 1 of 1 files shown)


addons/account/migrations/11.0.1.1/post-migration.py
```diff
@@ -44,6 +44,10 @@ def migrate_account_tax_cash_basis(env):
             WHERE company_id = rc.id AND tax_exigibility = 'on_payment')""")
 
 
+def migrate_account_report_invoice_with_payments(env):
+    env.ref("account.report_invoice_with_payments").attachment_use = False
```
pedrobaeza: I think this is not the report that is changing. Please check the correct XML-ID.

pedrobaeza:
The module is `account` (for the commit message), not `account.report...`

addons/account/migrations/11.0.1.1/post-migration.py
stefan-tecnativa: I was using the report name, not the id. I think now is correct.

pedrobaeza (approved):

PR #4099, opened 2023-07-27
[15.0][OU-FIX] website: Proper copyright footer writing

pedrobaeza:
- Copying the view is not setting things properly due to CoW processing. Done directly through write with website_id in context.
- There may not be any specific website.layout, so we need to fallback on the company name.

@Tecnativa TT44601


chienandalu (approved):

pilarvargas-tecnativa (approved):
thanks :)

PR #4100, opened 2023-07-31
[14.0][OU-IMP] website: extract footer copyright company name

pilarvargas-tecnativa:
In v13 the company name included in the copyright was done in a copy of the web.frontend_layout template. Now in v14 it is done in website.layout. This migration script aims to extract the content of the old copyright, in case it does not exist, the company name will be set.

cc Tecnativa TT44601

@pedrobaeza @chienandalu  please review :)

### DIFF
--- a/openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
@@ -1,9 +1,44 @@
 # Copyright (C) 2021 Open Source Integrators <https://www.opensourceintegrators.com/>
 # Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
+# Copyright 2023 Tecnativa - Pilar Vargas
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 import re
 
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/website_sale/14.0.1.0/post-migration.py
```diff
@@ -1,11 +1,55 @@
...
+        website_layout_pattern = (
+            r'<span class="o_footer_copyright_name mr-2">(.*?)<\/span>'
+        )
+        website_layout_matches, *_ = re.findall(
+            website_layout_pattern, website_layout_view.arch_db, re.DOTALL
+        )
+        new_arch = website_layout_view.arch_db.replace(
+            website_layout_matches, copyright_content
+        )
+        layout_copy_view = env["ir.ui.view"].search(
```
pedrobaeza: No need to search. It's done automatically when you write on the main one with the website_id context key.

pilarvargas-tecnativa: thanks! :)

pedrobaeza (approved):

PR #4105, opened 2023-08-07
[15.0][IMP] mail: migrate mail templates from mako to qweb

chienandalu:
Superseds #3852 with the following fixes:

- Use lxml for attributes extraction.
- Refine regex
- Extract inner methods so we can use the script from the shell without
  scope issues.
- Add missing template fields

cc @Tecnativa TT44698

### DIFF
--- a/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
+++ b/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
@@ -1,4 +1,10 @@
+import html
+import re
+
+from bs4 import BeautifulSoup
 from openupgradelib import openupgrade
+
+_PY_VAR_PATTERN = "[a-zA-Z_][a-zA-Z0-9_]*"
 
 
 def _map_activity_type_chaining_type_field(env):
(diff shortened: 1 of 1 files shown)


chienandalu:
Ready for review

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -64,6 +81,155 @@ def finish_migration_to_mail_group(env):
...
+    string = _migrate_set(string)
+    return string
+
+
+def _migrate_mail_templates(env):
+    """Migrate all mail templates from mako to qweb"""
+    templates = env["mail.template"].with_context(active_test=False).search([])
+    for template in templates:
+        for field in MAIL_TEMPLATE_MAKO_CHAR_FIELDS:
+            template[field] = _migrate_placeholder_char(template[field])
```
pedrobaeza, line 222: You need to do the same for char fields for all the languages.

chienandalu: The only translated field missing was `report_name`. Now it's addressed bellow in the installed langages context loop

pedrobaeza (approved):

PR #4107, opened 2023-08-18
[16.0][FIX] openupgrade_framework : end-migration scripts are forced …

### LINKED ISSUE #4106: [16.0] openupgrade_framework : patched method migrate_module should change load_state for end-migration scripts to be run at install
## Module
openupgrade_framework

## Describe the bug
In order to run migration scripts at install also, **migrate_module** method is patched to modify the module **state** field.
But in the original method, when dealing with end-migration scripts, state is initialized using **load_state** field

https://github.com/odoo/odoo/blob/9034602a028fb22fbd139cf41a2f5c71d9d8d032/odoo/modules/migration.py#L106-L110

So we need to modify the load_state field also for end-migration to run at install

## To Reproduce
**Affected versions**: 16.0

1. Create a end-migration script in module that is not yet installed but will be (with new dependency in another module)
2. Launch the migration

**Expected behavior**
The end-migration script should be run

metaminux:
…to run at install too

  Fixes OCA/OpenUpgrade#4106

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/modules/migration.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/migration.py
@@ -15,9 +15,14 @@ def migrate_module(self, pkg, stage):
     to_install = pkg.state == "to install"
     if to_install:
         pkg.state = "to upgrade"
+    load_state_to_install = getattr(pkg, "load_state", False)
+    if load_state_to_install == "to install":
+        pkg.load_state = "to upgrade"
     MigrationManager.migrate_module._original_method(self, pkg, stage)
     if to_install:
         pkg.state = "to install"
+    if load_state_to_install == "to install":
+        pkg.load_state = "to install"
 
 
 migrate_module._original_method = MigrationManager.migrate_module
--- a/openupgrade_scripts/scripts/base/16.0.1.3/tests/end-migration_test.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/tests/end-migration_test.py
@@ -1,0 +1,14 @@
+# Copyright 2023 Odoo Community Association (OCA)
+# Copyright 2023 Guillaume Masson <guillaume.masson@meta-it.fr>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import logging
+
+from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate(no_version=True)
+def migrate(env, version):
+    params = env["ir.config_parameter"].sudo()
+    params.set_param("openupgrade.test_end_migration", "Executed")
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,6 +95,12 @@ jobs:
           for snippet in openupgrade/openupgrade_scripts/scripts/*/*/tests/data*.py; do
             odoo-old/odoo-bin shell -d $DB < $snippet
           done
+      - name: Test end-migration script for newly installed module (i.e. "spreadsheet" for v16)
+        run: |
+          END_SCRIPT_TEST=openupgrade/openupgrade_scripts/scripts/base/16.0.1.3/tests/end-migration_test.py
+          NEW_MODULE_SCRIPT_DIR=openupgrade/openupgrade_scripts/scripts/spreadsheet/16.0.1.0/
+          mkdir -p $NEW_MODULE_SCRIPT_DIR
+          cp $END_SCRIPT_TEST $NEW_MODULE_SCRIPT_DIR
       - name: OpenUpgrade test
         run: |
           # select modules and perform the upgrade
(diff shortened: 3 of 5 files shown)


hbrunn (approved):
thanks, the CI errors are unrelated.

do we have any core modules where this is the case in v16? If so, a [test](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/tests/test_base_migration.py) for that would be great.

metaminux:
@hbrunn There is no end-migration script for v16 in openupgrade_scripts, but there are 2 files in native Odoo CE :

- https://github.com/odoo/odoo/blob/74b555766ae6e63efe70f07b3ebcbaa94050f183/addons/l10n_lu/migrations/2.1/end-migrate_update_taxes.py
- https://github.com/odoo/odoo/blob/74b555766ae6e63efe70f07b3ebcbaa94050f183/addons/l10n_ch/migrations/11.1/end-migrate_update_taxes.py#L4

Can we use those files to run a test ?

hbrunn:
never tried it, but I *think* it could just work to drop a folder with the module name and correct version into https://github.com/OCA/OpenUpgrade/tree/16.0/openupgrade_scripts/scripts, cf https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_framework/odoo_patch/odoo/tests/loader.py#L30

Edit: ah wait, those are localization modules and those aren't installed in the test database we use. And it must be a newly pulled module anyways. How about adding an end migration script for ie spreadsheet (that's new in v16 and pulled during the migration), check for environment variable [OPENUPGRADE_TESTS](https://github.com/OCA/OpenUpgrade/blob/f29f99fe04fdbe4a014be71f5fc0095e3eadff5c/.github/workflows/test.yml#L133) and if that's set, ie set some config parameter as marker that you can check for in the test?

metaminux:
Hello @hbrunn 

Instead of creating an end-migration script in the "production" tree of openupgrade_scripts, I'd like to copy such script during the tests using [data.py](https://github.com/OCA/OpenUpgrade/blob/a38ad4e4a5b98e439188caa98728a77026f5181c/openupgrade_scripts/scripts/base/16.0.1.3/tests/data.py) file... but I'm struggling to get a right path from the Odoo shell.

It seems I can't get the value of GITHUB_WORKSPACE... any advice ?

hbrunn:
just `$GITHUB_WORKSPACE` is empty? Not sure however what I think of this approach, anyone else an opinion?

Just to be clear, I don't *insist* on the test, just thought it was nice

pedrobaeza:
No idea, sorry.

hbrunn:
so... do we go ahead and merge this?

pedrobaeza:
OK, please rebase to check the CI.

metaminux:
H everyone,

CI is not OK after rebase, I'll convert this PR to draft for the moment because it seems to me that some others changes in the way odoo is loading modules are preventing post-* and end-* migration scripts to execute for newly installed ones...

metaminux:
Hi @hbrunn Hi @pedrobaeza 

Sorry for the delay... I just rebased and had to fix some typo for pre-commit to run OK.

I think we can go ahead...
Tests are using the end migration script for spreadsheet which is newly installed. See https://github.com/OCA/OpenUpgrade/actions/runs/12905959373/job/35986260407?pr=4107#step:16:18575

MiquelRForgeFlow:
Could you rebase correctly? I see merge branch commits there. I think you should have here only two commits (the pre-commit commit can be fused with the commit that adds tests)

metaminux:
@MiquelRForgeFlow Rebase correctly done, thanks

end migration script for newly installed spreadsheet module runs : https://github.com/OCA/OpenUpgrade/actions/runs/12910128466/job/35999439017?pr=4107#step:16:18582

openupgrade_framework/odoo_patch/odoo/modules/migration.py
```diff
@@ -15,9 +15,14 @@ def migrate_module(self, pkg, stage):
     to_install = pkg.state == "to install"
     if to_install:
         pkg.state = "to upgrade"
+    load_state_to_install = getattr(pkg, "load_state", False)
```
MiquelRForgeFlow, line 18: I think you should adapt the docstring of this method...

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/base/16.0.1.3/tests/end-migration_test.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Odoo Community Association (OCA)
```
pedrobaeza, line 1: This is not testing really the feature, as this module is not installed during the upgrade.

metaminux: Hi @pedrobaeza 

Look at .github/workflows/test.yml, this file is copied to **spreadsheet** directory which is installed during the upgrade.

Do you think I should create a directory containing only tests ?

pedrobaeza: OK, I see, it seems it works for this version, but for the next ones, `spreadsheet` will be already installed.

pedrobaeza (approved):

PR #4108, opened 2023-08-24
[15.0][OU-FIX] website: unlink view when extract footer copyright company name

pilarvargas-tecnativa:
cc @Tecnativa TT44794

The duplicate view of website.layout with associated website needs to be removed as in v15 this view should not be duplicated and if not removed it causes inheritance problems which are aggravated when there is more than one website.

### DIFF
--- a/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
@@ -3,6 +3,14 @@ import re
 from openupgradelib import openupgrade
 
 
+def _delete_views_recursively(env, view):
+    views = env["ir.ui.view"].with_context(active_test=False).browse(view.id).exists()
+    views |= views.mapped("inherit_children_ids")
+    while views.mapped("inherit_children_ids") - views:
+        views |= views.mapped("inherit_children_ids")
+    views.unlink()
+
+
 def extract_footer_copyright_company_name(env):
     """Replace Copyright content in the new v15 template so as not to lose
     content from previous versions if it has been customised, or directly put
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
```diff
@@ -36,6 +36,7 @@ def extract_footer_copyright_company_name(env):
             or f"Copyright © {website.company_id.name}",
         )
         main_copyright_view.with_context(website_id=website.id).arch_db = new_arch
+        view.unlink()
```
chienandalu: Probably you won't be able to do this due to inheritance chain

pilarvargas-tecnativa: Right, with this additional code the inheritances will also be eliminated, thank you!

openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
```diff
@@ -36,6 +44,7 @@ def extract_footer_copyright_company_name(env):
             or f"Copyright © {website.company_id.name}",
         )
         main_copyright_view.with_context(website_id=website.id).arch_db = new_arch
+        _delete_views_recursively(view)
```
chienandalu, line 47: This is pretty dangerous, you don't have control at this point of the script about what you're going to delete.

Wouldn't the way to go just changing the inheritance to the new views and the deleting the old ones safely?

pedrobaeza:
@pilarvargas-tecnativa what happened to this? Indeed remove all of that is risky. Is there any other solution?

pilarvargas-tecnativa:
> @pilarvargas-tecnativa what happened to this? Indeed remove all of that is risky. Is there any other solution?

Well, after a number of migrations, I think this PR could be cancelled, and when a view with inheritance arises, search for the exact view, with key and inheritance. It is too risky to merge for testing purposes.

PR #4111, opened 2023-08-25
[14.0][OU-FIX] website_sale: move script to website and disable it

pilarvargas-tecnativa:
This script should go in website, not website_sale. However, this script is causing errors in migrations to later versions and its call will be commented in case someone considers that its execution is relevant and can decomment it. The reason is that in v14 the main view website.layout was duplicated with the associated web when modifying the name of the company, something that this script reproduces and in the behaviour of later versions this view should not be duplicated. It is better not to automate tasks on main views to avoid errors in migrations to later versions.

cc @Tecnativa TT44794

### DIFF
--- a/openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
@@ -1,8 +1,44 @@
 # Copyright 2021 ForgeFlow S.L.  <https://www.forgeflow.com>
 # Copyright 2023 Tecnativa - Pilar Vargas
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import re
+
 from openupgradelib import openupgrade
 from openupgradelib.openupgrade_140 import convert_html_string_13to14
+
+
+def extract_footer_copyright_company_name(env):
+    """Replace Copyright content in the footer so as not to lose
+    content from previous versions if the copyright has been customised."""
+    for website in env["website"].search([]):
+        # Search for old copyright and if it does not exist set the default company name
+        web_frontend_layout_view = env["ir.ui.view"].search(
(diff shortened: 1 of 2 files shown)


pedrobaeza:
It's reasonable, but let's do one thing: don't remove the code, and just comment the line calling it, putting the explanation as comment, and saying that anyone can un-comment it if they consider it relevant.

pilarvargas-tecnativa:
> It's reasonable, but let's do one thing: don't remove the code, and just comment the line calling it, putting the explanation as comment, and saying that anyone can un-comment it if they consider it relevant.

Thank you, this has to be run on the website so what you recommend has been done in the right place.

pedrobaeza (approved):

openupgrade_scripts/scripts/website/14.0.1.0/post-migration.py
```diff
@@ -44,3 +80,10 @@ def migrate(env, version):
     openupgrade.delete_records_safely_by_xml_id(env, ["website.aboutus_page"])
     website_cookie_notice_post_migration(env)
     convert_html_string(env)
+    """ This call is commented out as this script causes conflicts with inheritance
+    in migrations to later versions. In v14 a duplicate of website.layout with
+    associated website is made when the company name is modified in the footer,
+    this script replicates that behaviour in the migration but in later versions
+    it can be a problem of inheritance. However, it can be uncommented and used if
+    deemed appropriate. """
```
pedrobaeza: This can't be a text, but several commented lines:

```suggestion
    # This call is commented out as this script causes conflicts with inheritance
    # in migrations to later versions. In v14 a duplicate of website.layout with
    # associated website is made when the company name is modified in the footer,
    # this script replicates that behaviour in the migration but in later versions
    # it can be a problem of inheritance. However, it can be uncommented and used if
    # deemed appropriate. """
```

PR #4113, opened 2023-08-31
[16.0][IMP] apriori: website_sale_require_login to website_sale

pilarvargas-tecnativa:
website_sale_require_login makes it mandatory to register/log in to make a purchase, this option is already provided for in website_sale v16 with the "Mandatory" option in the website settings under "Sign in/up at checkout" which does not allow guest checkout.

cc @Tecnativa TT44386

### DIFF
--- a/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
@@ -13,6 +13,32 @@ from odoo.addons.openupgrade_scripts.apriori import merged_modules, renamed_modu
 _logger = logging.getLogger(__name__)
 
 
+def login_or_registration_required_at_checkout(cr):
+    """website_sale_require_login merged into website_sale. Check if the
+    website_sale_require_login module was installed in v15 to set the
+    website.account_on_checkout field as mandatory so that the functionality
+    remains the same, login/registration required for checkout."""
+    # Check if the module is installed and its status is "installed".
+    module_name = "website_sale_require_login"
+    if openupgrade.is_module_installed(cr, module_name):
+        # Add the field 'account_on_checkout' to the 'website' table if it doesn't exist yet.
+        openupgrade.logged_query(
+            cr,
+            """
+            ALTER TABLE website
+            ADD COLUMN IF NOT EXISTS account_on_checkout
+            """,
+        )
+        # Set the value 'mandatory' in the field for all records in the table 'website'.
+        openupgrade.logged_query(
+            cr,
+            """
+            UPDATE website
+            SET account_on_checkout = 'mandatory'
+            """,
+        )
+
+
 def update_translatable_fields(cr):
     # map of nonstandard table names
     model2table = {
@@ -102,6 +128,7 @@ def migrate(cr, version):
             "recommended to run the Odoo with --load=openupgrade_framework "
             "when migrating your database."
         )
+    login_or_registration_required_at_checkout(cr)
     openupgrade.update_module_names(cr, renamed_modules.items())
     openupgrade.update_module_names(cr, merged_modules.items(), merge_modules=True)
     # restricting inherited views to groups isn't allowed any more
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -55,6 +55,8 @@ merged_modules = {
     "account_move_force_removal": "account",
     # OCA/account-invoice-reporting
     "account_invoice_report_due_list": "account",
+    # OCA/e-commerce
+    "website_sale_require_login": "website_sale",
     # OCA/purchase-workflow
     "product_form_purchase_link": "purchase",
 }


openupgrade_scripts/scripts/crm/16.0.1.8/upgrade_analysis_work.txt
chienandalu, line 1: What are these files doing here? :thinking:

pilarvargas-tecnativa: Pre-commit.. Maybe I should run pre-commit and save the changes in a separate commit?

chienandalu: You shouldn't run pre-commit globally, it will run automatically after your changes.

So undo these changes git reset origin 16.0, redo the change in apriori and then just commit

chienandalu (changes requested):
Beyond this, check how this new option is handled in the website_sale migration script and add a case son when this module was installed set it as mandatory

openupgrade_scripts/scripts/crm/16.0.1.8/upgrade_analysis_work.txt
pilarvargas-tecnativa: Still failing pre-commit

chienandalu:
Looks like there was a recent template upgrade so it looks like some code doesn't comply the new rules. @pedrobaeza do we do a new PR to fix them?

openupgrade_scripts/scripts/website_sale/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,17 @@
+from openupgradelib import openupgrade
+
+
+def login_or_registration_required_at_checkout(env):
+    """ In case the website_sale_require_login module was installed in v15,
+    the value of the account_on_checkout field is set to "mandatory" to make
+    login/registration mandatory for checkout. """
+    if env["ir.module.module"].search(
+        [("name", "=", "website_sale_require_login")]
+    ).state == "installed":
```
legalsylvain: Is it working ? 
Afaik, the renaming of the module (website_sale_require_login -> website_sale) is done in the pre-migration script of the base module, here : https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py#L106

So, in the post migration step of the website_sale module, the website_sale_require_login is not installed, as the entry has been dropped by https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1776

Did I missed something ?

pilarvargas-tecnativa: I was making changes for the pre-migration to run and just pushed them :) that wasn't going to work, thanks!

pilarvargas-tecnativa: >> So, in the post migration step of the website_sale module, the website_sale_require_login is not installed, as the entry has been dropped by https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1776

I had doubts if the record stayed in the database in post, @pedrobaeza has been telling me about it and this clears it up for me, thanks again!

pilarvargas-tecnativa:
> Beyond this, check how this new option is handled in the website_sale migration script and add a case son when this module was installed set it as mandatory

Done!

openupgrade_scripts/scripts/website_sale/16.0.1.1/post-migration.py
legalsylvain: the pre-migration of website_sale, is executed after the pre-migration of the base module. So, same problem, I think that it will return always false.

Note : an helper function exist https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2521

chienandalu: Then maybe we could take it out from the apriori and flag it to uninstall in the migration script. What do you think?

pilarvargas-tecnativa: So I understand that in the case of a merge module there is no way to know if it was installed or not, I will continue investigating, thank you very much.

pilarvargas-tecnativa: @chienandalu That's what I was thinking. I think it's the only option to be able to have the data and do the check. Thanks!

pedrobaeza: Do these operations in `base` pre-migration before the call of `update_module_names`

pilarvargas-tecnativa: Finally, the merge is not indicated in apriori, the appropriate operations are performed in pre-migration and the "update_module_names" method is called to perform the module merge. Correct me if it is not correct please.

pilarvargas-tecnativa: @pedrobaeza  I think we were writing at the same time, wouldn't it be correct to call the "update_module_names" method from the script itself? is it better to do it as you say?

pedrobaeza: That can be another option, yes. As you prefer.

legalsylvain: Hi @pilarvargas-tecnativa, thanks for the changes.

The implementation you propose is perfectly valid.

However, I'm a bit disappointed if the apriori file doesn't contain all the module merges.
I find it interesting to have a single file with all the changes from one version to another. It's very readable by any people, and it can also be parsed easily by a script.

Do you think we could implement Pedro's solution instead (do it in the "base" module)?

pilarvargas-tecnativa: @legalsylvain you are right. I have made the changes and will keep this in mind for the next time I am faced with something similar. Thank you.

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -13,6 +13,32 @@
...
 
 
+def login_or_registration_required_at_checkout(cr):
+    """ website_sale_require_login merged into website_sale. Check if the
+    website_sale_require_login module was installed in v15 to set the
+    website.account_on_checkout field as mandatory so that the functionality
+    remains the same, login/registration required for checkout. """
+    # Check if the module is installed and its status is "installed".
+    module_name = "website_sale_require_login"
+    if openupgrade.is_module_installed(cr, module_name):
```
chienandalu, line 23:
```suggestion
    if openupgrade.is_module_installed(cr, "website_sale_require_login"):
```

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -13,6 +13,32 @@
 _logger = logging.getLogger(__name__)
 
 
+def login_or_registration_required_at_checkout(cr):
+    """ website_sale_require_login merged into website_sale. Check if the
+    website_sale_require_login module was installed in v15 to set the
+    website.account_on_checkout field as mandatory so that the functionality
+    remains the same, login/registration required for checkout. """
```
chienandalu:
```suggestion
    """ The website_sale_require_login module is merged into website_sale. Check if the
    it was installed in v15 to set the website.account_on_checkout field as mandatory so that
    the functionality remains the same, login/registration required for checkout. """
```

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -13,6 +13,32 @@
...
+    # Check if the module is installed and its status is "installed".
+    module_name = "website_sale_require_login"
+    if openupgrade.is_module_installed(cr, module_name):
+        # Add the field 'account_on_checkout' to the 'website' table if it doesn't exist yet.
+        openupgrade.logged_query(
+            cr,
+            """
+            ALTER TABLE website
+            ADD COLUMN IF NOT EXISTS account_on_checkout
```
chienandalu, line 29: You have to declare the type
```suggestion
            ADD COLUMN IF NOT EXISTS account_on_checkout VARCHAR
```

legalsylvain (approved):
Thanks !
Lgtm.
Note something looks wrong with pre-commit test. Maybe rebase your branch coule solve the problem.

pedrobaeza (approved):

pedrobaeza:
Please do a separate PR fixing the other pre-commit things.

legalsylvain:
pre-commit fixed here : https://github.com/OCA/OpenUpgrade/pull/4119

/ocabot rebase

pedrobaeza:
Thanks for fixing the pre-commit, Sylvain:

/ocabot merge nobump

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
chienandalu: This wasn't fixed...

pedrobaeza: Sorry, my fault for not seeing it. @pilarvargas-tecnativa can you do another PR, please?

pilarvargas-tecnativa: Yes, I'm doing it right now

PR #4114, opened 2023-08-31
[15.0][OU-FIX] mass_mailing: don't change cron interval

chienandalu:
It isn't necessary to update the cron interval

cc @Tecnativa TT44876

cc @legalsylvain 

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/mass_mailing/15.0.2.5/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/mass_mailing/15.0.2.5/noupdate_changes.xml
@@ -1,8 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="ir_cron_mass_mailing_queue" model="ir.cron">
-    <field name="interval_number">1</field>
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):

openupgrade_scripts/scripts/mass_mailing/15.0.2.5/noupdate_changes.xml
```diff
@@ -1,8 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
   <record id="ir_cron_mass_mailing_queue" model="ir.cron">
-    <field name="interval_number">1</field>
```
pedrobaeza, line 4: Comment lines instead of removing them. That way, we know that is not something forgotten, and we can document why we comment them.

chienandalu: Done :)

chienandalu:
Ready to merge

PR #4127, opened 2023-09-12
[16.0] Fix openupgradelib dependencies

sbidoul:
~~This fixes the openupgradelib dependencies to be PEP 508 compliant.~~

~~However~~ git URLs are rejected by PyPI, so I would really prefer that we simply reference `openupgradelib`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 # generated from manifests external_dependencies
-git+https://github.com/OCA/openupgradelib.git@master
+openupgradelib @ git+https://github.com/OCA/openupgradelib.git@master
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,1 @@
-# generated from manifests external_dependencies
-openupgradelib
+openupgradelib @ git+https://github.com/OCA/openupgradelib@master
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,2 +1,2 @@
 # generated from manifests external_dependencies
-git+https://github.com/OCA/openupgradelib.git@master
+openupgradelib
```
pedrobaeza: We need the git one for having always the latest without having to do a release.

pedrobaeza (dismissed):

requirements.txt
sbidoul: That should be in `test-requirements.txt`.

That said, this PR is green so the latest release of openupgradelib is good enough.

pedrobaeza:
Of course right now the CI is green, but that can't be the situation in the future, where one pull request requires something not published. Putting test-requirements is enough for the CI, but not for people using requirements.txt for doing their migrations. I'm not one of them, so I won't block this, but just to let you know.

sbidoul:
I still think we should do more frequent release of openupgradelib.

But if you think it's important to have the git url in requirements.txt, then we can maintain it manually and remove its generation from the pre-commit config in this repo.

pedrobaeza:
Let's see what others think

sbidoul:
In the last commit, I have preserved the git URL in requirements.txt.

pedrobaeza (approved):

hbrunn (approved):

PR #4128, opened 2023-09-12
[16.0][FIX] base: case translation when no model data

MiquelRForgeFlow:
Apply https://github.com/odoo/odoo/commit/880538db4ac4833322e57adb3b4d8239eb382547.

### DIFF
--- a/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
@@ -4,7 +4,6 @@
 import logging
 
 from openupgradelib import openupgrade
-from psycopg2.extensions import AsIs
 
 from odoo import tools
 
@@ -85,7 +84,7 @@ def update_translatable_fields(cr):
         translation_name = "%s,%s" % (model, field)
         openupgrade.logged_query(
             cr,
-            """
+            f"""
             WITH t AS (
                 SELECT it.res_id as res_id, jsonb_object_agg(it.lang, it.value) AS value,
                     bool_or(imd.noupdate) AS noupdate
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -94,17 +82,15 @@ def update_translatable_fields(cr):
                 WHERE it.type = 'model' AND it.name = %(name)s AND it.state = 'translated'
                 GROUP BY it.res_id
             )
-            UPDATE "%(table)s" m
-            SET "%(field_name)s" = CASE WHEN t.noupdate THEN m.%(field_name)s || t.value
-                                    ELSE t.value || m.%(field_name)s END
+            UPDATE {table} m
+            SET "{field}" = CASE WHEN t.noupdate IS FALSE THEN t.value || m."{field}"
```
pedrobaeza, line 86: This switch in the condition is not correct, as you can have False or NULL, and both are interpreted as to update

pedrobaeza:
Can you please explain what are you changing?

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
MiquelRForgeFlow: No, you are wrong. Here what we need to do is put EXACTLY the same query that odoo have in their code. I only updated the query because odoo updated their query. And to why they did that change, is explained in their PR, see:

Selection_1761

pedrobaeza: We think the same both Raphael and me, hehe. I still think that there can be a possibility of having and existing `ir_model_data` record with the column set to NULL, and this will conflict with the NULL from `LEFT JOIN`, but OK if it's not a big problem in their samples.

PR #4131, opened 2023-09-20
[15.0][MIG] account_menu -> account_usability

Saran440:
rename module account_menu (v14) to account_usability (v15)

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -11,6 +11,8 @@ renamed_modules = {
     "mail_client_extension": "mail_plugin",
     "payment_ingenico": "payment_ogone",
     "website_mail_channel": "website_mail_group",
+    # OCA/account-financial-tools
+    "account_menu": "account_usability",
     # OCA/account-fiscal-rule
     "l10n_eu_oss": "l10n_eu_oss_oca",
     # OCA/e-commerce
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):

legalsylvain (changes requested):
Could you remove the line 56 in the same file. It is writen that account_menu is merged into account_usability, that is wrong.
Thanks !

Saran440:
@legalsylvain Fixed it. Thank you.

If i have module `A` and `B` in V14 and V15 it change name to module `C`. I should use `merged_modules` for this case?
I'm not sure what is difference `merged_modules` and `rename_modules`

legalsylvain:
Merge a + b -> b
Rename a -> b

More info in documentation use case : https://oca.github.io/OpenUpgrade/development.html#learn-from-typical-use-cases

legalsylvain (approved):

pedrobaeza:
Why changing this? Does a module `account_usability` exists in v14? If so, it should remain in `merged_modules`. There's is very few differences between using `update_modules` with merge_modules=True or False, but one on them is to avoid errors if the target name already exists.

legalsylvain:
> Why changing this? Does a module account_usability exists in v14? If so, it should remain in merged_modules. There's is very few differences between using update_modules with merge_modules=True or False, but one on them is to avoid errors if the target name already exists.

well. it is a rename, not a merge. I asked for a fix, because I made a mistake in the initial commit : https://github.com/OCA/OpenUpgrade/pull/3590
 
 account_usability doesn't exist in OCA before V15. https://odoo-community.org/shop/account-missing-menus-saxon-accounting-10561?search=account_usability#attr=22622

regards.

pedrobaeza:
But isn't present in Akretion branches? If so, it would help to keep it as is. It doesn't have any drawbacks, and for those having in the list the old one, it wouldn't crash.

legalsylvain:
> But isn't present in Akretion branches? If so, it would help to keep it as is. It doesn't have any drawbacks, and for those having in the list the old one, it wouldn't crash.

Indeed !

PR #4134, opened 2023-09-23
[OU][16.0] web_drop_target has been merged into web.

legalsylvain:
Ref : https://github.com/OCA/dms/issues/213#issuecomment-1719768162 by @vancouver29


pedrobaeza (approved):

pedrobaeza:
For reference, this should be tagged as `[OU-ADD]`.

PR #4135, opened 2023-09-25
[16.0][OU-ADD] stock_dropshipping: Nothing to do

victoralmau:
Nothing to do

Depends on:
- [x] https://github.com/OCA/OpenUpgrade/pull/4291

@Tecnativa TT44220


legalsylvain:
/ocabot migration stock_dropshipping

<dependency></dependency>

openupgrade_scripts/scripts/stock_dropshipping/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,6 @@
+---Models in module 'stock_dropshipping'---
+---Fields in module 'stock_dropshipping'---
+---XML records in module 'stock_dropshipping'---
+DEL stock.location.route: stock_dropshipping.route_drop_shipping (noupdate)
+NEW stock.route: stock_dropshipping.route_drop_shipping (noupdate)
+# Nothing to do.
```
pedrobaeza, line 6: For other times, please indicate always why nothing to do. In this case, because the model has been renamed in `stock`.

PR #4141, opened 2023-10-02
[14.0][OU-FIX] stock_account: Since v14, Billing group is no longer inherited from Stock Manager group

victoralmau:
Since v14, Billing group is no longer inherited from Stock Manager group

@Tecnativa TT45339

### DIFF
--- a/openupgrade_scripts/scripts/stock_account/14.0.1.1/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/stock_account/14.0.1.1/noupdate_changes.xml
@@ -3,4 +3,7 @@
   <!-- <record id="stock_valuation_layer_company_rule" model="ir.rule">
     <field name="global"/>
   </record> -->
+  <record id="stock.group_stock_manager" model="res.groups">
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/stock_account/14.0.1.1/noupdate_changes.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
-  <!-- <record id="stock_valuation_layer_company_rule" model="ir.rule">
-    <field name="global"/>
-  </record> -->
```
MiquelRForgeFlow: please don't delete this. we autogenerate this kind of data by script, and we don't want to think again if this needs to be commented or not when it reappears if we run the script again.

MiquelRForgeFlow (approved):
Good catch! The removal was done in https://github.com/odoo/odoo/commit/c87275a69f8f926b1f341d9237bb578b894335c4.

pedrobaeza (approved):

PR #4143, opened 2023-10-06
[15.0][Fix] mail: _migrate_placeholder_html should remove '| safe'

cuongnmtm:
This will fix the issue when Odoo renders this wrong syntax

Content in email template body v14:

`${object.company_id.email | safe}`

Content in email template body v15 (after migrate data):

`<t t-out="object.company_id.email | safe">`

Error:

```
odoo.addons.base.models.ir_qweb.QWebException: Error while render the template
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

### DIFF
--- a/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
+++ b/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
@@ -83,7 +83,7 @@ def finish_migration_to_mail_group(env):
 def _migrate_placeholder_char(string):
     """
     Replace dynamic placeholders in char/text fields:
-    Example: "Dear ${object.name}," -> "Dear {{object.name}},"
+    Example: "Dear ${object.name | safe}," -> "Dear {{object.name}},"
     """
     if not string:
         return string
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
@cuongnmtm could you rebase?

openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -103,8 +103,9 @@ def repl_placeholder(match):
 def _migrate_placeholder_html(string):
     """
     Replace dynamic placeholders in HTML fields:
-    Example: 'Your name is ${object.name}' -> 'Your name is <t t-out="object.name"></t>'
+    Example: 'Your name is ${object.name | safe}' -> 'Your name is <t t-out="object.name"></t>'
     """
+    string = re.sub(r"\s?\|\s?safe\s?", "", string)
```
hbrunn, line 108: there can be much more [filters](https://jinja.palletsprojects.com/en/stable/templates/#filters) - could you extend the regex to just replace anything after the first pipe?

hbrunn (changes requested):

MiquelRForgeFlow:
Done in https://github.com/OCA/OpenUpgrade/pull/4438.

PR #4151, opened 2023-10-09
[FIX] do not delete xml_id for routing sequence in mrp module as it will be renamed installing mrp_routing

sergiocorato:
Ref https://github.com/OCA/manufacture/pull/1102

### DIFF
--- a/openupgrade_scripts/scripts/mrp/14.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/mrp/14.0.2.0/post-migration.py
@@ -89,6 +89,6 @@ def migrate(env, version):
     map_mrp_production_state_planned(env)
     generate_workorders_for_draft_orders(env)
     openupgrade.load_data(env.cr, "mrp", "14.0.2.0/noupdate_changes.xml")
-    openupgrade.delete_records_safely_by_xml_id(
-        env, ["mrp.mrp_routing_rule", "mrp.sequence_mrp_route"]
-    )
+    openupgrade.delete_records_safely_by_xml_id(env, ["mrp.mrp_routing_rule"])
+    if not env["ir.module.module"].search([("name", "=", "mrp_routing")]):
+        openupgrade.delete_records_safely_by_xml_id(env, ["mrp.sequence_mrp_route"])


pedrobaeza:
Please put module name in the commit message, and we are tagging them as `[OU-FIX]` for differentiating them from regular fixes in for example the OpenUpgrade framework. Can you give some links to support the non removal?

openupgrade_scripts/scripts/mrp/14.0.2.0/post-migration.py
```diff
@@ -89,6 +89,4 @@ def migrate(env, version):
     map_mrp_production_state_planned(env)
     generate_workorders_for_draft_orders(env)
     openupgrade.load_data(env.cr, "mrp", "14.0.2.0/noupdate_changes.xml")
-    openupgrade.delete_records_safely_by_xml_id(
-        env, ["mrp.mrp_routing_rule", "mrp.sequence_mrp_route"]
-    )
+    openupgrade.delete_records_safely_by_xml_id(env, ["mrp.mrp_routing_rule"])
```
MiquelRForgeFlow, line 92:
```suggestion
    openupgrade.delete_records_safely_by_xml_id(env, ["mrp.mrp_routing_rule"])
    # not deleting mrp.sequence_mrp_route, as it is used in OCA/manufacture mrp_routing module.
```

MiquelRForgeFlow, line 92:
```suggestion
    openupgrade.delete_records_safely_by_xml_id(env, ["mrp.mrp_routing_rule"])
    if not openupgrade.is_module_installed(cr.cr, "mrp_routing"):
        openupgrade.delete_records_safely_by_xml_id(env, ["mrp.sequence_mrp_route"])
```

MiquelRForgeFlow (changes requested):

openupgrade_scripts/scripts/mrp/14.0.2.0/post-migration.py
sergiocorato: Thanks @MiquelRForgeFlow, but `mrp_routing` is a brand new module in v. 14.0, so this check will always result true and delete the sequence.

pedrobaeza:
I don't think this should be changed, as only installing the other module, you will need this record, so this shouldn't be unconditionally not deleted.

openupgrade_scripts/scripts/mrp/14.0.2.0/post-migration.py
MiquelRForgeFlow: But when installing `mrp_routing`, the migration script of `mrp_routing` will not execute 🤔

sergiocorato: Mmm, good point, I'll correct that.

sergiocorato:
> I don't think this should be changed, as only installing the other module, you will need this record, so this shouldn't be unconditionally not deleted.

I agree, this will only be useful in those migrations where this module will be installed.

pedrobaeza:
But is there a reason to keep this one? A new one with the exact same parameters is not enough?

sergiocorato:
A new one will not preserve current sequence and possible customization.
Anyway the pre_init_hook I added here https://github.com/OCA/manufacture/pull/1102 doesn't preserve them too, I don't know why.

openupgrade_scripts/scripts/mrp/14.0.2.0/post-migration.py
pedrobaeza: Migration scripts for newly installed modules happen in the context of OpenUpgrade (if you load `openupgrade_framework`. Please put here the condition to find the module in the module list, which is likely that you are going to install the module.

sergiocorato: So `pre-migration.py` script it's called even in a newly installed module?

I don't know what the phrase `put here the condition to find the module in the module list` means, does it refer to the `openupgrade.is_module_installed` method?

pedrobaeza: Yes and I mean:

`if env["ir.module.module"].search([("name", "=", "mrp_routing")]):`

sergiocorato: Wow, I'll try, thanks.

sergiocorato:
I haven't found a way to make it working, so I'll do it manually.

PR #4155, opened 2023-10-10
[16.0][OU-ADD] purchase_mrp: Nothing to do

victoralmau:
Nothing to do

Depends on:
- [x] #4381

@Tecnativa TT43624


legalsylvain:
/ocabot migration purchase_mrp

<dependency></dependency>

PR #3912, opened 2023-06-13
[16.0][OU-ADD] hr_contract

remytms:
Most of the changes are managed by the migration script of hr module. Load the noupdate_changes file.

PR #4156, opened 2023-10-11
[16.0][OU-ADD] hr_expense: Migration to 16.0

victoralmau:
Migration to 16.0

Depends on:
- [x] https://github.com/OCA/OpenUpgrade/pull/3911

@Tecnativa TT43223


legalsylvain:
/ocabot migration hr_expense

<dependency>Depends on : 

- hr_contract : #3912</dependency>

openupgrade_scripts/scripts/hr_expense/16.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,122 @@
...
+                        ON account.id = he.analytic_account_id
+                    WHERE he.analytic_account_id IS NOT NULL
+                    UNION ALL
+                    SELECT
+                        he.id AS expense_id,
+                        dist.account_id AS analytic_account_id,
+                        dist.percentage AS percentage
+                    FROM hr_expense he
+                    JOIN account_analytic_tag_hr_expense_rel tag_rel
+                        ON tag_rel.expense_id = he.id
```
marielejeune: Got an error on this line: 

`psycopg2.errors.UndefinedColumn: column tag_rel.expense_id does not exist`

I guess `tag_rel.expense_id` should be `tag_rel.hr_expense_id`?

marielejeune (commented):
Hi! Thanks for the migration script. Here is an error that I got.

openupgrade_scripts/scripts/hr_expense/16.0.2.0/post-migration.py
```diff
@@ -0,0 +1,78 @@
...
+                """UPDATE ir_module_module
+                SET state = 'to install'
+                WHERE name = 'hr_expense_analytic_tag'""",
+            )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr, "hr_expense", "migrations/16.0.2.0/noupdate_changes.xml"
```
marielejeune:
```suggestion
        env.cr, "hr_expense", "16.0.2.0/noupdate_changes.xml"
```

openupgrade_scripts/scripts/hr_expense/16.0.2.0/post-migration.py
```diff
@@ -0,0 +1,78 @@
...
+        env,
+        [
+            "hr_expense.product_product_fixed_cost",
+        ],
+    )
+
+
+def _hr_expense_process(env):
+    """Set amount_tax_company value only when it is necessary with ORM."""
+    env.cr.exeute(
```
marielejeune:
```suggestion
    env.cr.execute(
```

openupgrade_scripts/scripts/hr_expense/16.0.2.0/post-migration.py
```diff
@@ -0,0 +1,78 @@
...
+from openupgradelib import openupgrade
+
+
+def try_delete_noupdate_records(env):
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "hr_expense.product_product_fixed_cost",
+        ],
+    )
```
marielejeune: What's the purpose of deleting this record from the DB? And what should I do if this product was used (for example in account move lines)? 

Up to now I get the following (non blocking) error: 
```
2023-11-29 15:03:43,202 48396 ERROR ou-odna-16-mig odoo.sql_db: bad query: DELETE FROM "product_product" WHERE id IN (157)
ERROR: update or delete on table "product_product" violates foreign key constraint "account_move_line_product_id_fkey" on table "account_move_line"
DETAIL:  Key (id)=(157) is still referenced from table "account_move_line".
```

marielejeune (changes requested):
Some more comments

openupgrade_scripts/scripts/hr_expense/16.0.2.0/post-migration.py
victoralmau: Sorry, you're right, this is not necessary. Checking again, I have added a `rename_xmlids` of a same product in v15 and v16.

openupgrade_scripts/scripts/hr_expense/16.0.2.0/post-migration.py
```diff
@@ -0,0 +1,66 @@
...
+        SELECT id
+        FROM hr_expense
+        WHERE total_amount_company != total_amount AND amount_tax > 0
+        AND amount_tax_company != 0
+        """
+    )
+    expense_model = env["hr.expense"]
+    for expense_id in env.cr.fetchall():
+        expense = expense_model.browse(expense_id)
+        expense._compute_amount_tax()
```
remytms: In general, @pedrobaeza recommends to not call compute function, but rather implement it in SQL. The risk by calling a compute method is that it generate other compute that can give undesired side effects. Am I right ?

pedrobaeza: Yes, and for performance reasons, but sometimes there's no equivalent SQL or the number of records to compute is very low that it doesn't worth. Didn't check if this is the case.

remi-filament (changes requested):
Hi @victoralmau , thank you for this PR.
I have tested your module when migrating our instance, and I face an issue when trying to reconcile payment on account move (state and payment_state are not updated to paid on hr.expense.sheet).
After some digging, I realized that from v16.0 onwards, account_move related to hr_expense_sheet are created with move_type = "in_invoice" (when there was no move_type before so it ended up with default "entry") as per https://github.com/odoo/odoo/blob/87f4667d81f7d8a40dcf225f3daf1a9e2795680d/addons/hr_expense/models/hr_expense.py#L1297

Then in _compute_payment_state from account_move it will check if account.move is of type invoice (with is_invoice function here : https://github.com/odoo/odoo/blob/87f4667d81f7d8a40dcf225f3daf1a9e2795680d/addons/account/models/account_move.py#L926 and if move_type is "entry" it would ends up with payment_state = "not_paid".

Therefore, I think that this migration script needs to update move_type for all moves with an expense_sheet_id to "in_invoice"

pedrobaeza (approved):
All the comments attended including @remi-filament's one, plus the final fixes + improvements I have done reviewing the PR

PR #4158, opened 2023-10-12
[16.0][OU-ADD] account_edi_ubl_cii

remi-filament:
Rename XML-ids from merged modules


legalsylvain:
/ocabot migration account_edi_ubl_cii

<dependency></dependency>

MuhammadFayyazq125:
I have migrated odoo v15 to v16. I am unable to view my account.move.line records in account.move as in Progress I can see that the data is exist in v16 migrated database but cannot view in views. what could be the reason any one?

docsource/modules150-160.rst
```diff
@@ -14,15 +14,15 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | account_edi                                     | Nothing to do        |No DB layout changes.                            |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| account_edi_facturx                       |                      |                                                 |
+| |del| account_edi_facturx                       |                      |Merged into account_edi_ubl_cii                  |
```
pedrobaeza:
```suggestion
| |del| account_edi_facturx                       | Done                 | Merged into account_edi_ubl_cii                 |
```

openupgrade_scripts/scripts/account_edi_ubl_cii/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,25 @@
+from openupgradelib import openupgrade
+
+_xmlids_renames = [
+    (
+        "l10n_be_edi.edi_efff_1",
```
pedrobaeza: This is not needed, as the merging operation in the apriori.py already renames them. You only need it if the part since the dot is different, and anyway, the source XML-ID would be `account_edi_ubl_cii.edi_efff_1`, not with `l10n_be_edi`.

pedrobaeza (changes requested):
Finally this is nothing to do.

openupgrade_scripts/scripts/account_edi_ubl_cii/16.0.1.0/pre-migration.py
remi-filament: Thanks for your review @pedrobaeza indeed apriori.py should do the job, so there is nothing to do on this module.
I have removed pre-migration and updated docsource accordingly.

docsource/modules150-160.rst
```diff
@@ -14,15 +14,15 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | account_edi                                     | Nothing to do        |No DB layout changes.                            |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| account_edi_facturx                       |                      |                                                 |
+| |del| account_edi_facturx                       | Nothing to do        |Merged into account_edi_ubl_cii                  |
```
pedrobaeza: It's not nothing to do. Something has been done, which is put the entry in the merged dictionary.

```suggestion
| |del| account_edi_facturx                       | Done                 | Merged into account_edi_ubl_cii                 |
```

remi-filament: OK I added Done / Merged into account_edi_ubl_cii on all modules merged in account_edi_ubl_cii from apriori then

pedrobaeza (approved):

PR #4167, opened 2023-10-17
[15.0] [OU-ADD] digest

cuongnmtm:
Load noupdate_changes


rvalyi (approved):

openupgrade_scripts/scripts/digest/15.0.1.1/post-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright Odoo Community Association (OCA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "digest", "15.0.1.1/noupdate_changes.xml")
```
chienandalu, line 15: You should remove translations delete translations as well as in https://github.com/OCA/OpenUpgrade/blob/877ae6638a01bcdca7c9281f3e3bb9ca94fd5af5/openupgrade_scripts/scripts/digest/16.0.1.1/post-migration.py#L20

chienandalu (changes requested):
This strikes back in v16 when upgrading mass_mailing, which needs the new hooks provided by these templates

openupgrade_scripts/scripts/digest/15.0.1.1/post-migration.py
cuongnmtm: Thanks @chienandalu

chienandalu (approved):
Thanks :)

chienandalu (approved):

PR #4177, opened 2023-10-18
[16.0][OU-ADD] website_profile: Migration to 16.0

victoralmau:
Migration to 16.0

Depends on: 
- [x] https://github.com/OCA/OpenUpgrade/pull/3949

@Tecnativa TT45247


legalsylvain:
/ocabot migration website_profile

<dependency></dependency>

chienandalu (commented):
Depends on

- [x] https://github.com/OCA/OpenUpgrade/pull/3949

pedrobaeza:
I have pushed a proper version. @victoralmau, some comments for future scripts:

- noupdate=0 are removed by ORM. You don't have to explicitly remove in the scripts.
- The syntax for the delete translations method is to put the XML-ID, and optionally the fields, but not only the fields.
- The path of the file was not correct.

PR #4178, opened 2023-10-20
[16.0][OU-ADD] payment_stripe: Migration to 16.0

victoralmau:
Migration to 16.0

Depends on:
- [x] https://github.com/OCA/OpenUpgrade/pull/3918

@Tecnativa TT45236


legalsylvain:
/ocabot migration payment_stripe

<dependency></dependency>

chienandalu (commented):
I'd comment in the `/noupdate_changes.xml` file these fields, which don't change from version to version:

```xml
    <field name="allow_tokenization">True</field>
    <field name="code">stripe</field>
```

victoralmau:
> I'd comment in the `/noupdate_changes.xml` file these fields, which don't change from version to version:
> 
> ```
>     <field name="allow_tokenization">True</field>
>     <field name="code">stripe</field>
> ```

In another PR it was discussed about commenting on the `noupdate_changes.xml` file but it was concluded not to do so because the content of the file is generated "automatically" and I would suggest again to uncomment it.

Although I agree that it is not necessary to update those fields, I think it does not bother either, and it is useful to use the `noupdate_changes.xml` file (if it exists) in the migration in case new changes are added there in the future.

pedrobaeza:
The problem is to remove the lines, not to comment them.

chienandalu (approved):

pedrobaeza (approved):

openupgrade_scripts/scripts/payment_stripe/16.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_xmlids_renames = [
+    (
+        "payment_stripe.action_payment_acquirer_onboarding",
```
pedrobaeza, line 7: Please check in this module and the other if the declarations of both records have different fields and one leaves a dangling value.

victoralmau: The `action_payment_provider_onboarding` action of v16 (https://github.com/odoo/odoo/blob/16.0/addons/payment_stripe/views/payment_provider_views.xml) has the same data as `action_payment_acquirer_onboarding` of v15 (https://github.com/odoo/odoo/blob/15.0/addons/payment_stripe/views/payment_views.xml), only `name` and `res_model` changes.

The `payment_paypal` module does not have an onboarding action.

pedrobaeza: OK, I meant the same fields set. You know that if for example the old action defines a domain and the new one not, the old domain prevails when migrating.

victoralmau: Yes, exactly, when what you indicate happens, it is precisely what must be taken into account and left correct.

PR #4180, opened 2023-10-22
[16.0][OU] New analysis

pedrobaeza:
It solves the problems found in #4042 

With latest changes in Odoo care taken into account.

@Tecnativa


pedrobaeza:
I have pushed updated work files for the already merged modules, and nothing critical. Only a new noupdate_changes. I'll merge when green to continue the work of the migration scripts.

legalsylvain:
diff review. (except first commit). LGTM. Thanks, @pedrobaeza.

pedrobaeza:
The summary is that is not too many changes (only a bit of noise), although the `upgrade_analysis` process is a bit tough (I have always delegated this task on @MiquelRForgeFlow). I required whole Sunday to get it on working, and with some glitches.

PR #4185, opened 2023-10-23
[16.0][OU-ADD] website_payment: Migration to 16.0

victoralmau:
Migration to 16.0

@Tecnativa TT43633


legalsylvain:
/ocabot migration website_payment

<dependency></dependency>

chienandalu (approved):

openupgrade_scripts/scripts/website_payment/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "website_payment.mail_template_donation",
+        ],
+    )
+    openupgrade.load_data(
+        env.cr, "website_payment", "migrations/16.0.1.0/noupdate_changes.xml"
```
remytms: I think that the correct location is `"16.0.1.0/noupdate_changes.xml"` and not under the `migrations` directory.

openupgrade_scripts/scripts/website_payment/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "website_payment.mail_template_donation",
+        ],
+    )
```
remytms: I think that simply loading `noupdate_changes.xml` is enough. No need to delete this record before.

Based on: http://oca.github.io/OpenUpgrade/use_cases/noupdate_xml_entry_changed.html

remytms (changes requested):
I may be wrong, but I think that this PR needs some changes.

openupgrade_scripts/scripts/website_payment/16.0.1.0/post-migration.py
pedrobaeza: Mostly: you need to reset the translation for the field name.

pedrobaeza (changes requested):
@victoralmau please rebase an attend comments.

And other thing: this is not "Migration to 16.0", but "Migration scripts"

victoralmau:
Changes done.

openupgrade_scripts/scripts/website_payment/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
...
+
+_translations_to_delete = [
+    "mail_template_donation",
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "website_payment", "16.0.1.0/noupdate_changes.xml")
+    openupgrade.delete_record_translations(env.cr, "load_data", _translations_to_delete)
```
pedrobaeza: Please limit the deletion to the name field.

pedrobaeza (approved):

pedrobaeza:
For getting green CI:

/ocabot merge nobump

PR #4187, opened 2023-10-23
[16.0][OU-ADD] loyalty

pilarvargas-tecnativa:
cc @Tecnativa TT44332

@chienandalu @ernesto-garcia-tecnativa please review

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
@@ -1,0 +1,57 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+from odoo.tools.translate import _
+
+_deleted_xml_records = [
+    "loyalty.sale_coupon_generate_rule",
+]
+
+
+def convert_loyalty_program_rewards(env):
+    openupgrade.m2o_to_x2m(
+        env.cr, env["loyalty.program"], "loyalty_program", "reward_ids", "reward_id"
+    )
+
+
+def convert_loyalty_program_rules(env):
+    openupgrade.m2o_to_x2m(
+        env.cr, env["loyalty.program"], "loyalty_program", "rule_ids", "rule_id"
+    )
+
+
+def compute_portal_point_name(env):
+    """This is a computed field, but the _program_type_default_values method of the
+    loyalty module sets the following values in portal_point_name depending on the
+    program_type field. This is done in post so that the language context can be used."""
+    portal_point_names = {
+        "coupons": _("Coupon point(s)"),
+        "promotion": _("Promo point(s)"),
+        "gift_card": _("Gift Card"),
+        "loyalty": _("Loyalty point(s)"),
+        "ewallet": _("eWallet"),
+        "promo_code": _("Discount point(s)"),
+        "buy_x_get_y": _("Credit(s)"),
+        "next_order_coupons": _("Coupon point(s)"),
+    }
+    loyalty_programs = env["loyalty.program"].search([])
+    for program in loyalty_programs:
+        if program.program_type in portal_point_names:
+            translated_name = portal_point_names[program.program_type]
+            # By default when the module is installed it contains the terms in the
+            # language code "en_US".
+            program.with_context(lang="en_US").write(
+                {"portal_point_name": translated_name}
+            )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    convert_loyalty_program_rewards(env)
+    convert_loyalty_program_rules(env)
+    compute_portal_point_name(env)
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        _deleted_xml_records,
+    )
--- a/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
@@ -1,0 +1,796 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_field_renames = [
+    ("coupon.program", "coupon_program", "promo_applicability", "applies_on"),
+    ("coupon.program", "coupon_program", "promo_code_usage", "trigger"),
+    ("coupon.program", "coupon_program", "maximum_use_number", "max_usage"),
+    ("coupon.reward", "coupon_reward", "reward_description", "description"),
+    ("coupon.reward", "coupon_reward", "discount_percentage", "discount"),
+    ("coupon.reward", "coupon_reward", "discount_apply_on", "discount_applicability"),
+    ("coupon.reward", "coupon_reward", "discount_type", "discount_mode"),
+    (
+        "coupon.reward",
+        "coupon_reward",
+        "discount_specific_product_ids",
+        "discount_product_ids",
+    ),
+    ("coupon.reward", "coupon_reward", "reward_product_quantity", "reward_product_qty"),
+    ("coupon.rule", "coupon_rule", "rule_minimum_amount", "minimum_amount"),
+    ("coupon.rule", "coupon_rule", "rule_min_quantity", "minimum_qty"),
+    ("coupon.rule", "coupon_rule", "rule_products_domain", "product_domain"),
+    (
+        "coupon.rule",
+        "coupon_rule",
+        "rule_minimum_amount_tax_inclusion",
+        "minimum_amount_tax_mode",
+    ),
+]
+_models_renames = [
+    ("coupon.program", "loyalty.program"),
+    ("coupon.reward", "loyalty.reward"),
+    ("coupon.coupon", "loyalty.card"),
+    ("coupon.rule", "loyalty.rule"),
+]
+_tables_renames = [
+    ("coupon_program", "loyalty_program"),
+    ("coupon_reward", "loyalty_reward"),
+    ("coupon_coupon", "loyalty_card"),
+    ("coupon_rule", "loyalty_rule"),
+]
+_xmlids_renames = [
+    (
+        "loyalty.coupon_action",
+        "loyalty.loyalty_card_action",
+    ),
+    (
+        "loyalty.coupon_generate",
+        "loyalty.loyalty_generate_wizard_action",
+    ),
+    (
+        "loyalty.report_coupon_code",
+        "loyalty.report_loyalty_card",
+    ),
+    (
+        "loyalty.coupon_view_form",
+        "loyalty.loyalty_card_view_form",
+    ),
+    (
+        "loyalty.coupon_view_tree",
+        "loyalty.loyalty_card_view_tree",
+    ),
+    (
+        "loyalty.coupon_generate_view_form",
+        "loyalty.loyalty_generate_wizard_view_form",
+    ),
+    (
+        "loyalty.coupon_program_view_form_common",
+        "loyalty.loyalty_program_view_form",
+    ),
+    (
+        "loyalty.coupon_program_view_search",
+        "loyalty.loyalty_program_view_search",
+    ),
+    (
+        "loyalty.coupon_program_view_tree",
+        "loyalty.loyalty_program_view_tree",
+    ),
+    (
+        "sale_loyalty.mail_template_sale_coupon",
+        "loyalty.mail_template_loyalty_card",
+    ),
+    (
+        "gift_card.gift_card_product_50",
+        "loyalty.gift_card_product_50",
+    ),
+]
+
+_noupdate_xmlids = [
+    "mail_template_loyalty_card",
+    "gift_card_product_50",
+]
+
+_columns_copies = {
+    "loyalty_rule": [
+        ("minimum_amount_tax_mode", None, None),
+    ],
+    "loyalty_reward": [
+        ("discount_applicability", None, None),
+        ("discount_mode", None, None),
+    ],
+    "loyalty_program": [
+        ("applies_on", None, None),
+        ("trigger", None, None),
+        ("program_type", None, None),
+    ],
+}
+
+
+def update_loyalty_program_data(env):
+    # Fill date_to field: This data is updated with the values of the field 'rule_date_to
+    # field of the 'loyalty_rule' table. Sets the 'date_to' field of the 'loyalty_program'
+    # table to non-null values of to the non-null values of 'rule_date_to' where the old
+    # 'loyalty_program.rule_id' field is equal to loyalty_rule.id' and
+    # 'loyalty_rule.rule_date_to' is not null.
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_program
+        ADD COLUMN IF NOT EXISTS date_to DATE
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_program
+        SET date_to = loyalty_rule.rule_date_to
+        FROM loyalty_rule
+        WHERE loyalty_program.rule_id = loyalty_rule.id
+        AND loyalty_rule.rule_date_to IS NOT NULL
+        """,
+    )
+    # Sets the value of the limit_usage column to True for those records where the max_usage
+    # column is not null.
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_program
+        ADD COLUMN IF NOT EXISTS limit_usage BOOLEAN
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_program
+        SET limit_usage = TRUE
+        WHERE max_usage IS NOT NULL
+        """,
+    )
+    # Determine the visibility on the default portal of specific loyalty programs, based
+    # on their type (program_type)
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_program
+        ADD COLUMN IF NOT EXISTS portal_visible BOOLEAN
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_program
+        SET portal_visible =
+            CASE
+                WHEN program_type
+                IN ('gift_card', 'ewallet', 'loyalty', 'next_order_coupons') THEN true
+                ELSE false
+            END
+        """,
+    )
+    # Update standard selection values
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("applies_on"),
+        "applies_on",
+        [("on_current_order", "current"), ("on_next_order", "future")],
+        table="loyalty_program",
+    )
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("trigger"),
+        "trigger",
+        [("no_code_needed", "auto"), ("code_needed", "with_code")],
+        table="loyalty_program",
+    )
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("program_type"),
+        "program_type",
+        [("coupon_program", "coupons"), ("promotion_program", "promotion")],
+        table="loyalty_program",
+    )
+
+
+def update_loyalty_reward_data(env):
+    # Determine which program_id and company_id each reward belongs to
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_reward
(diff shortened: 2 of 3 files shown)


openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def convert_loyalty_program_rules(env):
+    openupgrade.m2o_to_x2m(
+        env.cr, env["loyalty.program"], "loyalty_program", "rule_ids", "rule_id"
+    )
+
+
+def compute_portal_point_name(env):
+    programs = env["loyalty.program"].search([])
```
chienandalu: This migration is going to be quite big. Document well the why's behing these methods

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,107 @@
...
+def set_portal_visible_default_values(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_program
+        SET portal_visible =
+            CASE
+                WHEN program_type IN ('gift_card', 'ewallet', 'loyalty', 'next_order_coupons') THEN true
+                ELSE false
+            END
```
chienandalu: Why this? The default value is `False`

This would be a simpler implementation, btw:
```suggestion
        SET portal_visible = program_type IN ('gift_card', 'ewallet', 'loyalty', 'next_order_coupons')
```

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pedrobaeza: I think you can do this with SQL.

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    portal_point_names = {
+        'coupons': _('Coupon point(s)'),
```
pedrobaeza: And here translations are useless, as there's no language context. Shouldn't this be put in all languages?

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,107 @@
...
+        """,
+    )
+
+
+def fill_date_to_field(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_program
+        ADD COLUMN IF NOT EXISTS date_to VARCHAR
```
FernandoRomera: date_to DATE

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: In this method the default values are set depending on the type of promotion, in this case I have taken the values of portal_visible to adapt whatever comes.
https://github.com/odoo/odoo/blob/16.0/addons/loyalty/models/loyalty_program.py#L188

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pilarvargas-tecnativa: Like the previous field I discussed, here set the default values for portal_point_names including translations
https://github.com/odoo/odoo/blob/16.0/addons/loyalty/models/loyalty_program.py#L188

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pilarvargas-tecnativa: I'm going to tune as much as possible to do as much as I can in SQL.

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pedrobaeza: This is not what I'm asking for. Anyway, instead of having to ask, you can put that explanations on comments in the code.

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: thanks :)

openupgrade_scripts/scripts/loyalty/16.0.1.0/noupdate_changes.xml
```diff
@@ -0,0 +1,271 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+    <record id="loyalty_card_action" model="ir.actions.act_window">
```
chienandalu: These records are noupdate=0, there's no need for this

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,630 @@
...
+        AND loyalty_program.rule_id = loyalty_rule.id
+        """,
+    )
+
+
+def fill_program_id_field(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_reward
```
chienandalu: I think it would be better to group these changes by table

chienandalu (changes requested):
Starting to look quite good :)

- Check CIs
- Comment code as much as you can

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pilarvargas-tecnativa: Finally, I have done it with a single language context, the one that comes by default when installing the module.

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: done :)

chienandalu: I still see a bunch of methods that could be group by table. I mean, a method for each table/model transformation

pilarvargas-tecnativa: done!

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,62 @@
...
+        "gift_card": _("Gift Card"),
+        "loyalty": _("Loyalty point(s)"),
+        "ewallet": _("eWallet"),
+        "promo_code": _("Discount point(s)"),
+        "buy_x_get_y": _("Credit(s)"),
+        "next_order_coupons": _("Coupon point(s)"),
+    }
+    openupgrade.logged_query(
+        env.cr,
+        """ SELECT id FROM loyalty_program """,
```
chienandalu: To do this, you can simply use the ORM and the result is going to be far more readable :slightly_smiling_face:

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    loyalty_program_ids = env.cr.fetchall()
+    for program_id in loyalty_program_ids:
+        program = env["loyalty.program"].browse(program_id)
```
chienandalu: Indeed, here you are fetching them one by one, which really underperformant. Anyway. Use a regular `search` as said before

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,815 @@
...
+            CASE
+                WHEN program_type = 'coupon_program' THEN 'coupons'
+                WHEN program_type = 'promotion_program' THEN 'promotion'
+            END
+        """,
+    )
+
+
+def update_loyalty_reward_data(env):
+    # Fill program_id
```
chienandalu: This kind of comments aren't very useful :sweat_smile:

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,815 @@
...
+    # coupons with a validity time, it is calculated by taking the creation date of
+    # the program to which the duration defined in the field is added to obtain the
+    # date.
+    # - rule_date_to (coupon.rule): in the case of promotional programs, the date of
+    # rule_date_to is taken.
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_program
+        ADD COLUMN IF NOT EXISTS date_to DATE
```
chienandalu, line 120: Why wouldn't it exist?

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,815 @@
...
+    (
+        "gift_card.gift_card_product_50",
+        "loyalty.gift_card_product_50",
+    ),
+]
+
+
+def update_loyalty_program_data(env):
+    # Fill date_to fill: This data takes the values of the following fields:
+    # - validity_duration (coupon.program): in case of coupon programs that generate
```
chienandalu: Wasn't this already set in v15 :thinking:

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,815 @@
...
+        ALTER TABLE loyalty_program
+        ADD COLUMN IF NOT EXISTS limit_usage BOOLEAN
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_program
+        SET limit_usage = TRUE
+        WHERE max_usage IS NOT NULL
```
chienandalu, line 147: Did you rename `maximum_use_number`?

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,815 @@
...
+            ELSE 1
+        END
+        """,
+    )
+
+
+# Field incorporated in new module loyalty_initial_date_validity
+def fill_date_from_field(env):
+    """Field incorporated in new module loyalty_initial_date_validity. This field takes
+    the data from the old rule_date_from field of the coupon.rule template."""
```
chienandalu, line 493: This is a good comment :+1:

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: Because in v15 this field did not exist in coupon.program(loyalty.program), the validity was calculated with the validity_duration field. When making scripts in pre it is better to make sure if it exists or not before creating it, isn't it?

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: In v15 the coupon view was set to a validity in days (validity_duration) and in the case of the promotions view a validity date was set in the rules (rule_date_to). In v16 there is no distinction for coupons or promotions, everything goes into loyalty programs and is set in the new date_to field.

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: Yes, here: https://github.com/OCA/OpenUpgrade/blob/824798f4dc539f115a686872e3342dee89a4c7f4/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py#L8

But with misspelling, I correct it :S

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
chienandalu: I'd say you're mixing concepts here. In one hand we already had `rule_date_to`, which is the one to dump into this new column (it's true that it was in another relation).

On the other, we've got the **coupon** experation, which is the one that `validity_duration` was setting. But that mechanism doesn't exist anymore.

I'd say that:

- there's nothing to do with `validity_duration`
- you should move `rule_date_to` values to `date_to` in `loyalty_program`

pilarvargas-tecnativa: But then, in the case of a coupon promotion, how do we establish the validity of the coupon?
For example:
In v15 we have program_type="coupon_program" which is the program that generates coupons and in this we set the validity_duration field.
image
In this view the field rule_date_to cannot be set because in v15 there are two different views for the two types of promotions and in the case of "coupon_program" this field is not in the view.

In v16 we have the same view for all program_types, in this case the counterpart would be program_type="coupons" and we need that validity but in date. 
image
In the case of coupon programs that come with a past validity date, when migrating them, they would be valid again if a validity date is not calculated from the duration.

chienandalu: Existing coupons already have their `expiration_date` set. That's shouldn't change. And as far as I can see, those dates aren't triggered by the dates on the program.

pilarvargas-tecnativa: It is true that the coupons have the expiry date set in these cases, what had led me to error is that in v16 it is not set when generating the coupons but in the data coming from v15 it will be. Thanks!

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,806 @@
...
+            END
+        FROM gift_card gc
+        WHERE lc.program_id = gc.program_id
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env.cr, _field_renames)
```
cormaza: Please change to pass only env, on current version of lib raise error https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L634
```suggestion
    openupgrade.rename_fields(env, _field_renames)
```

cormaza (changes requested):
Please, check api call

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: thanks :)

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,806 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _field_renames)
+    openupgrade.rename_models(env.cr, _models_renames)
+    openupgrade.rename_tables(env.cr, _tables_renames)
+    openupgrade.rename_xmlids(env.cr, _xmlids_renames)
+    # Renamed in rename_xmlids method.
+    # In v15 are noupdate=1 and in v16 are noupdate=0
+    openupgrade.set_xml_ids_noupdate_value(
```
xtanuiha: The error message when testing the upgrade script is as follows:
```
odoo.exceptions.UserError: XML IDs mail_template_loyalty_card must be a tuple or list!
```

These lines of code can be changed to:
```python
    openupgrade.set_xml_ids_noupdate_value(
        env, "loyalty", ["mail_template_loyalty_card"], False
    )
    openupgrade.set_xml_ids_noupdate_value(
        env, "loyalty", ["gift_card_product_50"], False
    )
```

pilarvargas-tecnativa: thanks

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,61 @@
...
+        env.cr, env["loyalty.program"], "loyalty_program", "rule_ids", "rule_id"
+    )
+
+
+def compute_portal_point_name(env):
+    """This is a computed field, but the _program_type_default_values method of the
+    loyalty module sets the following values in portal_point_name depending on the
+    program_type field. This is done in post so that the language context can be used."""
+    portal_point_names = {
+        "coupons": _("Coupon point(s)"),
```
pedrobaeza, line 29: AFAIK, these translation are never going to apply, as they are not included in the standard PO/POT files (the reference is by file name).

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,61 @@
...
+        "ewallet": _("eWallet"),
+        "promo_code": _("Discount point(s)"),
+        "buy_x_get_y": _("Credit(s)"),
+        "next_order_coupons": _("Coupon point(s)"),
+    }
+    openupgrade.logged_query(
+        env.cr,
+        """ SELECT id FROM loyalty_program """,
+    )
+    loyalty_programs = env["loyalty.program"].browse([])
```
pedrobaeza: This browse is not doing anything, and why not doing a regular search?

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,806 @@
...
+            CASE
+                WHEN applies_on = 'on_current_order' THEN 'current'
+                WHEN applies_on = 'on_next_order' THEN 'future'
+            END
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_program
```
pedrobaeza: Use `map_values` from openupgradelib instead of these queries

openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,806 @@
...
+        """
+        ALTER TABLE loyalty_reward
+        ADD COLUMN IF NOT EXISTS clear_wallet BOOLEAN
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_reward
+        SET clear_wallet = false
```
pedrobaeza: Instead of this, user `add_fields` from openupgradelib with a default value. (The same for others)

pedrobaeza (changes requested):

pedrobaeza:
Error trying the scripts:

```
  File "/opt/odoo/custom/src/openupgrade/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py", line 800, in migrate
    update_loyalty_reward_data(env)
  File "/opt/odoo/custom/src/openupgrade/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py", line 314, in update_loyalty_reward_data
    openupgrade.logged_query(
  File "/usr/local/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 1703, in logged_query
    cr.execute(query, args)
  File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 321, in execute
psycopg2.errors.UndefinedColumn: column "minimum_amount_tax_mode" does not exist
LINE 5:                 WHEN minimum_amount_tax_mode = 'tax_included...
```

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pilarvargas-tecnativa: https://github.com/odoo/odoo/blob/16.0/addons/loyalty/i18n/es.po#L746
But they are in the standard translation files and that is how the loyalty programs are named:
https://github.com/odoo/odoo/blob/16.0/addons/loyalty/models/loyalty_program.py#L197

pedrobaeza: But as said, translations work referring the file:

image

so this is not going to work.

openupgrade_scripts/scripts/loyalty/16.0.1.0/post-migration.py
pedrobaeza: Remove this SELECT

pilarvargas-tecnativa:
@pedrobaeza All changes done

pedrobaeza (approved):
Let's merge for now after first real tests. Later fixup/improvements in subsequent PRs.

PR #4191, opened 2023-10-25
[15.0][OU-ADD] pos_order_return: Merged in point_of_sale

stefan-tecnativa:
TT45179 @Tecnativa

The OCA module pos_order_return allows the user to partially return a sale_order in the frontend, and also links this refunds with their original orders. From Odoo v15 we can do this in core from the frontend and those links are paired to the ones we were providing. One thing that adds this module that isn't in core is a boolean indicating if you want to allow negative quantities on the POS for products. This feature can be discussed if it is worth it to be conserved in a custom module or not as it was a little out of context anyway.

@chienandalu please review!

### DIFF
--- a/openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
@@ -3,8 +3,34 @@
 from openupgradelib import openupgrade
 
 
+def merge_pos_order_return_module(env):
+    """
+    The pos_order_return module is now merged into point_of_sale. We can hook into
+    the new core functionality just renaming the proper fields.
+    """
+    if openupgrade.column_exists(env.cr, "pos_order", "returned_order_id"):
+        openupgrade.rename_fields(
+            env,
+            [
+                (
+                    "pos.order.line",
+                    "pos_order_line",
+                    "returned_line_id",
+                    "refunded_orderline_id",
+                ),
+                (
+                    "pos.order.line",
+                    "pos_order_line",
+                    "refund_line_ids",
+                    "refund_orderline_ids",
+                ),
+            ],
+        )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
+    merge_pos_order_return_module(env)
     openupgrade.rename_fields(
         env,
         [
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -68,6 +68,8 @@ merged_modules = {
     "website_sale_stock_available_display": "website_sale_stock",
     # OCA/hr-attendance
     "hr_attendance_user_list": "hr_attendance",
+    # OCA/pos
+    "pos_order_return": "point_of_sale",
     # OCA/product-attribute
     "stock_account_product_cost_security": "product_cost_security",
     # OCA/stock-logistics-reporting


openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py
```diff
@@ -47,6 +47,28 @@
     ("l10n_ec.state_ec_24", "base.state_ec_24"),
 ]
 
+
+def fill_point_of_sale_refunded_orderline_id_column(cr):
+    # fill pos_order_line field
```
chienandalu: Check if the module is installed. e. g.: https://github.com/oca/openupgrade/blob/15.0/openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py#L6-L9

Or checking if the origin columns exist.

openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py
```diff
@@ -47,6 +47,28 @@
...
+    openupgrade.logged_query(
+        cr,
+        """
+        ALTER TABLE pos_order_line
+        ADD COLUMN IF NOT EXISTS refunded_orderline_id INTEGER
```
chienandalu: I'd say it would be enough renaming the fields: `returned_line_id` and `refund_line_ids` to their new pairs in `point_of_sale`

openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py
```diff
@@ -66,7 +88,7 @@ def migrate(cr, version):
             "recommended to run the Odoo with --load=openupgrade_framework "
             "when migrating your database."
         )
-
+    fill_point_of_sale_refunded_orderline_id_column(cr)
```
chienandalu: Better call this method something like `merge_pos_order_return_module`

chienandalu (changes requested):
On the commit message (I paste it here as I can't comment inline of it :))

> This module

The OCA module `pos_order_return`

> allows

Write in past tense as we're leaving it behind

> the user to return a sale_order fully or partially in the frontend.

This isn't true. What it did was:

- to link refunds with their original orders
- to allow partial refunds (Odoo only allowed full ones)

> Now this feature is accomplished in core...

From Odoo v16 we can do this in core from the frontend and those links are paired to the ones we were providing.

> One think

thing

> that adds this module that isn't in core is a boolean indicating if you want to allow negative quantities on the POS for products. This feature can be discussed if it is worth it to be conserved in a custom module or not.

Talk about the logic:

This module also  added the restriction for negative quantities in the point of sale. It can be considered for an extra module as it was a little bit out of context anyway.

openupgrade_scripts/scripts/base/15.0.1.3/pre-migration.py
stefan-tecnativa: Understood

chienandalu (commented):
Why two commits?

Also, fit the commit message in 80 column width lines

stefan-tecnativa:
> Why two commits?
> 
> Also, fit the commit message in 80 column width lines

Changed!

chienandalu (approved):
ping @legalsylvain if you want to take a look

legalsylvain (approved):
Thanks !

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
```diff
@@ -3,8 +3,33 @@
 from openupgradelib import openupgrade
 
 
+def merge_pos_order_return_module(env):
+    module_pos_order_return = env["ir.module.module"].search(
+        [("name", "=", "pos_order_return")]
```
pedrobaeza: This is not valid, as at this stage the module is already merged in `point_of_sale`. You should inspect traces at DB structure level (if a column or table exists).

pedrobaeza (changes requested):

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
chienandalu: Is this code then wrong?: https://github.com/oca/openupgrade/blob/15.0/openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py#L6-L9

pedrobaeza: Not in this case, as you make the changes here (one marked as to remove, and the other to install), not on apriori.py: https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/apriori.py

chienandalu: Aah, oks :)

So @stefan-tecnativa hook into our fields to infer we've got the module. You can use `column_exists` https://github.com/OCA/openupgradelib/blob/d0aee908031876a95dc36c7a9a2c847186155aa2/openupgradelib/openupgrade_tools.py#L36

stefan-tecnativa: @chienandalu is it correct now? I am only checking one field, is in this case mandatory to check both?

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
```diff
@@ -3,8 +3,30 @@
 from openupgradelib import openupgrade
 
 
+def merge_pos_order_return_module(env):
+    if openupgrade.column_exist(env.cr, "pos_order_line", "returned_order_id"):
```
chienandalu: Wrong table :point_down: 
```suggestion
    if openupgrade.column_exist(env.cr, "pos_order", "returned_order_id"):
```

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
```diff
@@ -3,8 +3,30 @@
 from openupgradelib import openupgrade
 
 
+def merge_pos_order_return_module(env):
```
chienandalu, line 6: Add also a docstring summarizing the purpose

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
```diff
@@ -3,8 +3,36 @@
...
+    """
+    pos_order_return merged into point_of_sale. At this stage the module
+    is already merged. So to ensure that the module is installed we can
+    check if any of the columns that the OCA module adds exists at this
+    moment.
```
chienandalu:
```suggestion
    The pos_order_return module is now merged into point_of_sale. We can hook into
    the new core functionality just renaming the proper fields.
```

chienandalu (approved):

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/pre-migration.py
chienandalu: Yes, you can mark it as resolve and request @pedrobaeza to review again

stefan-tecnativa:
@pedrobaeza can you take a look again please?

pedrobaeza (approved):

PR #4192, opened 2023-10-27
[15.0][OU][IMP] crm: crm.team use_leads check

chienandalu:
Since v15 there are mechanisims to toggle this crm.team flag depending on the groups and reset some defaults [0]. There's also a mechanisim to automatically set as lead the crm.lead created from the website form when the assigned team has the flag use_leads on [1] making that the user wouldn't be able to reach those leads as the menu isn't available. As we could have inconsistent states from the previous installation we try to settle the states correctly.

[0] odoo/odoo/blob/15.0/addons/website_crm/models/crm_lead.py#L55
[1] odoo/odoo/tree/15.0/addons/crm/models/res_config_settings.py#L136-L141

cc @Tecnativa TT45672

please review @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/crm/15.0.1.6/post-migration.py
+++ b/openupgrade_scripts/scripts/crm/15.0.1.6/post-migration.py
@@ -10,6 +10,29 @@ def try_delete_noupdate_records(env):
     )
 
 
+def fix_crm_team_use_leads(env):
+    """Since v15 there are mechanisims to toggle this crm.team flag depending on the
+    groups and reset some defaults [0]. There's also a mechanisim to automatically set
+    as lead the crm.lead created from the website form when the assigned team has
+    the flag use_leads on [1] making that the user wouldn't be able to reach those leads
+    as the menu isn't available. As we could have inconsistent states from the previous
+    installation we try to settle the states correctly.
+    [0] odoo/odoo/blob/15.0/addons/website_crm/models/crm_lead.py#L55
+    [1] odoo/odoo/tree/15.0/addons/crm/models/res_config_settings.py#L136-L141
+    """
+    use_leads = env.ref("crm.group_use_lead") in env.ref("base.group_user").implied_ids
+    # We use leads, no problem in this case. The mechanisims in v15 will set things
+    # correctly in case the admins change the settings.
+    if use_leads:
+        return
+    # If we don't use leads, these teams aren't set correctly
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/crm/15.0.1.6/post-migration.py
```diff
@@ -10,6 +10,29 @@ def try_delete_noupdate_records(env):
...
+    """Since v15 there are mechanisims to toggle this crm.team flag depending on the
+    groups and reset some defaults [0]. There's also a mechanisim to automatically set
+    as lead the crm.lead created from the website form when the assigned team has
+    the flag use_leads on [1] making that the user wouldn't be able to reach those leads
+    as the menu isn't available. As we could have inconsistent states from the previous
+    installation we try to settle the states correctly.
+    [0] odoo/odoo/blob/15.0/addons/website_crm/models/crm_lead.py#L55
+    [1] odoo/odoo/tree/15.0/addons/crm/models/res_config_settings.py#L136-L141
+    """
+    use_leads = env["res.users"].has_group("crm.group_use_lead")
```
pedrobaeza: I prefer that this check is done if the group internal user has inherited this other group. If not, any old user can contain the group, but not be a massive option.

chienandalu: Well, this check is done only on SUPERUSER which should ok, isn't?. I haven't seen that `crm.group_use_lead` is declared as inherited in `base.group_user` or do you mean it could be a remanent of old setups?

pedrobaeza: In fact you are not doing on any user (empty recordset).

chienandalu: It goes down to self._uid: https://github.com/odoo/odoo/blob/8b552acbaf798e932b6386caaeb7cb9b6bd36bbc/odoo/addons/base/models/res_users.py#L858

pedrobaeza: But why not doing the other way? I think it's less prone to error.

chienandalu: Ok, now I get what you meant :smile: Changed :)

pedrobaeza (approved):

PR #4196, opened 2023-10-30
[15.0][OU][FIX] payment_transfer: custom acquirers redirect form

chienandalu:
We're assigning with the noupdate changes the new redirect form to the transfer acquirer but for those transfer acquirers made by the users we have to do it manually or they won't work properly.

cc @Tecnativa TT45663

please review @pedrobaeza @victoralmau

### DIFF
--- a/openupgrade_scripts/scripts/payment_transfer/15.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/payment_transfer/15.0.2.0/post-migration.py
@@ -4,4 +4,8 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "payment_transfer", "15.0.2.0/noupdate_changes.xml")
+    # Delete the old form record and manually link the custom acquirers made by
+    # the users.
     openupgrade.delete_records_safely_by_xml_id(env, ["payment_transfer.transfer_form"])
+    custom_transfer = env["payment.acquirer"].search([("provider", "=", "transfer")])
+    custom_transfer.redirect_form_view_id = env.ref("payment_transfer.redirect_form")


openupgrade_scripts/scripts/payment_transfer/15.0.2.0/post-migration.py
```diff
@@ -4,4 +4,10 @@
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "payment_transfer", "15.0.2.0/noupdate_changes.xml")
+    # Delete the old form record and link manually link the custom acquirers made by
```
pedrobaeza:
```suggestion
    # Delete the old form record and manually link the custom acquirers made by
```

openupgrade_scripts/scripts/payment_transfer/15.0.2.0/post-migration.py
```diff
@@ -4,4 +4,10 @@
...
+    # the users.
     openupgrade.delete_records_safely_by_xml_id(env, ["payment_transfer.transfer_form"])
+    custom_transfer = env["payment.acquirer"].search(
+        [("provider", "=", "transfer"), ("redirect_form_view_id", "=", False)]
```
pedrobaeza: Shouldn't they be all of them?

chienandalu: Every acquirer type could have a different acquire redirect form

openupgrade_scripts/scripts/payment_transfer/15.0.2.0/post-migration.py
chienandalu: done

openupgrade_scripts/scripts/payment_transfer/15.0.2.0/post-migration.py
pedrobaeza: I mean all the transfer one, although they have defined `redirect_form_view_id`. It will probably be incorrect.

chienandalu: Ok, done

pedrobaeza (approved):

victoralmau (approved):

PR #4199, opened 2023-10-31
[16.0][IMP] apriori: coupon_commercial_partner_applicability merged into loyalty_partner_applicability

pilarvargas-tecnativa:
pending from: 
- [x] https://github.com/OCA/sale-promotion/pull/171

cc @Tecnativa TT45744

@chienandalu @pedrobaeza please review

### DIFF
--- a/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
@@ -12,6 +12,21 @@ from odoo.addons.openupgrade_scripts.apriori import merged_modules, renamed_modu
 _logger = logging.getLogger(__name__)
 
 
+def enable_coupon_sharing_within_entity(cr):
+    """Check before merging `coupon_commercial_partner_applicability` into
+    `loyalty_partner_applicability` if it was installed in v15 to set the parameter
+    to True to keep the same functionality"""
+    if openupgrade.is_module_installed(cr, "coupon_commercial_partner_applicability"):
+        # The value of the configuration parameter is set to True.
+        openupgrade.logged_query(
+            cr,
+            """
+            INSERT INTO ir_config_parameter (key, value)
+            VALUES ('loyalty_partner_applicability.allow_coupon_sharing', 'true')
+            """,
+        )
+
+
 def login_or_registration_required_at_checkout(cr):
     """The website_sale_require_login module is merged into website_sale. Check if the
     it was installed in v15 to set the website.account_on_checkout field as mandatory
@@ -114,6 +129,7 @@ def migrate(cr, version):
             "when migrating your database."
         )
     login_or_registration_required_at_checkout(cr)
+    enable_coupon_sharing_within_entity(cr)
     openupgrade.update_module_names(cr, renamed_modules.items())
     openupgrade.update_module_names(cr, merged_modules.items(), merge_modules=True)
     # restricting inherited views to groups isn't allowed any more
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -64,6 +64,8 @@ merged_modules = {
     "account_invoice_report_due_list": "account",
     # OCA/e-commerce
     "website_sale_require_login": "website_sale",
+    # OCA/sale-promotion
+    "coupon_commercial_partner_applicability": "sale_loyalty_partner_applicability",
     # OCA/purchase-workflow
     "product_form_purchase_link": "purchase",
     # OCA/stock-logistics-workflow
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -65,7 +65,7 @@ merged_modules = {
     # OCA/e-commerce
     "website_sale_require_login": "website_sale",
     # OCA/sale-promotion
-    "coupon_commercial_partner_applicability": "sale_loyalty_partner_applicability",
+    "coupon_commercial_partner_applicability": "loyalty_partner_applicability",
     # OCA/purchase-workflow
     "product_form_purchase_link": "purchase",
     # OCA/stock-logistics-workflow


openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -12,6 +12,22 @@
 _logger = logging.getLogger(__name__)
 
 
+def enable_coupon_sharing_within_entity(cr):
+    """The coupon_commercial_partner_applicability module is merged into
+    loyalty_partner_applicability. It is checked if it was installed in v15
+    to set the configuration parameter to True to keep the same functionality."""
+    # Check if the module is installed and its status is "installed".
+    if openupgrade.is_module_installed(cr, "coupon_commercial_partner_applicability"):
```
chienandalu, line 19: No valid in this case. Check this discussion: https://github.com/OCA/OpenUpgrade/pull/4191#discussion_r1371547207

pilarvargas-tecnativa: In that case the script was being made in the module itself, there the module is already merged, I have added the script in base, before the merge method is executed. 
https://github.com/OCA/OpenUpgrade/blob/bd4f489b6b976c703f26de8a509be362d238bb25/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py#L133
In this case there would be two options, do it as I have done including it in apriori or do it in the module itself by calling the update_module_names method from the script but in this case it would not be included in apriori and it is clearer to always have it in apriori.

chienandalu: Thanks for the explanations :)

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -12,6 +12,22 @@
...
+        # The value of the configuration parameter is set to True.
+        openupgrade.logged_query(
+            cr,
+            """
+            INSERT INTO ir_config_parameter (key, value)
+            VALUES ('loyalty_partner_applicability.allow_coupon_sharing', 'true')
```
chienandalu, line 25: <del>But this column doesn't exist yet, does it?</del>

Forget it :facepalm:

chienandalu (approved):
Is it ok for you @pedrobaeza ?

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
pedrobaeza: Don't comment the "what" of obvious things:

```suggestion
    if openupgrade.is_module_installed(cr, "coupon_commercial_partner_applicability"):
```

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -12,6 +12,22 @@
 _logger = logging.getLogger(__name__)
 
 
+def enable_coupon_sharing_within_entity(cr):
+    """The coupon_commercial_partner_applicability module is merged into
+    loyalty_partner_applicability. It is checked if it was installed in v15
+    to set the configuration parameter to True to keep the same functionality."""
```
pedrobaeza:
```suggestion
    """Check before merging `coupon_commercial_partner_applicability` into 
    `loyalty_partner_applicability` if it was installed in v15 to set the parameter
    to True to keep the same functionality.
    """
```

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -12,6 +12,22 @@
...
+    # Check if the module is installed and its status is "installed".
+    if openupgrade.is_module_installed(cr, "coupon_commercial_partner_applicability"):
+        # The value of the configuration parameter is set to True.
+        openupgrade.logged_query(
```
pedrobaeza, line 21: Better to use `env["ir.config_parameter"].sudo().set_param(...)` to avoid possible conflicts.

pilarvargas-tecnativa: But in pre-migration in the base module we don't have "env" available. And I'm not sure if in this case the parameter "config[" "]" could be used.

pedrobaeza: Yes, you have `env` in both pre and post. What you don't have in pre is access to models defined in the module, but it's not the case: `env["ir.config_parameter"]` is accessible except in `base` pre-migration.

pedrobaeza: Oh, sorry, I didn't see that it's `base` pre-migration.

openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -12,6 +12,15 @@
 _logger = logging.getLogger(__name__)
 
 
+def enable_coupon_sharing_within_entity(cr):
+    """Check before merging `coupon_commercial_partner_applicability` into
+    `loyalty_partner_applicability` if it was installed in v15 to set the parameter
+    to True to keep the same functionality"""
+    if openupgrade.is_module_installed(cr, "coupon_commercial_partner_applicability"):
+        # The value of the configuration parameter is set to True.
+        cr["ir.config_parameter"].sudo().set_param("allow_coupon_sharing", "True")
```
pedrobaeza: Please revert to the previous SQL, which was correct. It has been my fault.

pilarvargas-tecnativa: It was the silly test of the day and it slipped in, sorry.

pedrobaeza (approved):

PR #4203, opened 2023-10-31
[16.0][OU-IMP] account: Use openupgradelib~fill_analytic_distribution

pedrobaeza:
It was extracted from here, which was generalized before merging the account scripts.


legalsylvain (approved):

pedrobaeza:
Current error is being fixed in https://github.com/odoo/odoo/pull/140565

PR #4204, opened 2023-10-31
[16.0][UPD] copier template

sergiocorato:


pedrobaeza:
This repo can't apply the update as is. I already applied the patch manually, so this shouldn't be needed.

.github/workflows/test.yml
sergiocorato, line 1: I don't know how to manage this file.

PR #4205, opened 2023-10-31
[16.0][OU-IMP] website: bootstrap 5 migration

chienandalu:
Highly reccomended to run along with https://github.com/OCA/openupgradelib/pull/349 which should improve the conversion avoiding false positives (content changed which had no bootstrap stuff to treat)

- Fix previous script
- Add script to migrate html fields content

cc @Tecnativa TT45734

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
@@ -1,38 +1,62 @@
 from openupgradelib import openupgrade
-from openupgradelib.openupgrade_160 import convert_string_bootstrap_4to5
-from psycopg2.extras import Json
+from openupgradelib.openupgrade_160 import (
+    _convert_field_bootstrap_4to5_sql,
+    convert_field_bootstrap_4to5,
+)
 
 
-def boostrap_5_migration(env):
+def convert_custom_qweb_templates_bootstrap_4to5(env):
     """Convert customized website views to Bootstrap 5."""
     backup_column = openupgrade.get_legacy_name("arch_db_bs4")
     openupgrade.logged_query(
         env.cr, f"ALTER TABLE ir_ui_view ADD COLUMN {backup_column} TEXT"
     )
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        UPDATE ir_ui_view SET {backup_column} = arch_db
+        WHERE type = 'qweb' AND website_id IS NOT NULL
+        """,
+    )
     # Find views to convert
     env.cr.execute(
-        """
-        SELECT iuv.id, iuv.arch_db
-        FROM ir_ui_view iuv
-        WHERE iuv.type = 'qweb'
-        """
+        "SELECT id FROM ir_ui_view WHERE type = 'qweb' AND website_id IS NOT NULL"
     )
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
```diff
@@ -1,38 +1,62 @@
...
+        """,
+    )
     # Find views to convert
     env.cr.execute(
-        """
-        SELECT iuv.id, iuv.arch_db
-        FROM ir_ui_view iuv
-        WHERE iuv.type = 'qweb'
-        """
+        "SELECT id FROM ir_ui_view WHERE type = 'qweb' AND website_id IS NOT NULL"
```
pedrobaeza, line 23: Why only applying to those with `website_id` assigned? I thought on this too, but I finally decided to apply it to all the existing templates.

pedrobaeza:
See the CI error:

```
   File "/home/runner/work/OpenUpgrade/OpenUpgrade/openupgrade/openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py", line 61, in migrate
    convert_custom_qweb_templates_bootstrap_4to5(env)
  File "/home/runner/work/OpenUpgrade/OpenUpgrade/openupgrade/openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py", line 26, in convert_custom_qweb_templates_bootstrap_4to5
    _convert_field_bootstrap_4to5_sql(env.cr, "ir_ui_view", "arch_db", ids=view_ids)
  File "/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/openupgradelib/openupgrade_160.py", line 393, in _convert_field_bootstrap_4to5_sql
    cr.execute(sql, params)
  File "/home/runner/work/OpenUpgrade/OpenUpgrade/odoo/odoo/sql_db.py", line 321, in execute
psycopg2.errors.SyntaxError: syntax error at or near "ARRAY"
LINE 1: SELECT id, arch_db FROM ir_ui_view WHERE id IN ARRAY[1914,21...
```

openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
chienandalu: Well, templates without website_id will be noupdate=0 records so it would be useless to update them or am I missing a case where there could be such custom views?

chienandalu:
> See the CI error:

It should be fixed by now

openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
pedrobaeza: Not sure if there can be any manually created view. If you are sure that it couldn't, then it's OK to continue this way.

chienandalu: I won't say I'm 100% sure but I can't figure out any exception :sweat_smile:

pedrobaeza: OK, let's keep it this way for now.

pedrobaeza (approved):

PR #4212, opened 2023-11-05
[16.0][OU-ADD] sale_loyalty

pilarvargas-tecnativa:
cc @Tecnativa TT44353

ping @chienandalu

### DIFF
--- a/openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/post-migration.py
@@ -1,0 +1,47 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_deleted_xml_records = [
+    "sale_loyalty.sale_coupon_apply_code_rule",
+]
+
+
+def convert_applied_coupons_from_sale_order_to_many2many(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO loyalty_card_sale_order_rel (sale_order_id, loyalty_card_id)
+        SELECT sales_order_id, id
+        FROM loyalty_card
+        WHERE sales_order_id IS NOT NULL
+        """,
+    )
+
+
+def fill_code_enabled_rule_ids_from_sale_order(env):
+    """This field relates the id of a rule to the id of
+    a sales order to which it has been applied, provided that the rule belongs to an
+    applicable promotion with code."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO loyalty_rule_sale_order_rel (sale_order_id, loyalty_rule_id)
+        SELECT lc.order_id, lr.id
+        FROM loyalty_card lc
+        JOIN loyalty_rule lr ON lc.program_id = lr.program_id
+        JOIN loyalty_program lp ON lr.program_id = lp.id
+        WHERE lr.mode = 'with_code'
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "sale_loyalty", "16.0.1.0/noupdate_changes.xml")
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        _deleted_xml_records,
+    )
+    convert_applied_coupons_from_sale_order_to_many2many(env)
+    fill_code_enabled_rule_ids_from_sale_order(env)
--- a/openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
@@ -1,0 +1,255 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+
+import random
+
+from openupgradelib import openupgrade
+
+_xmlids_renames = [
+    (
+        "sale_loyalty.sale_coupon_apply_code_action",
+        "sale_loyalty.sale_loyalty_coupon_wizard_action",
+    ),
+    (
+        "sale_loyalty.menu_coupon_type_config",
+        "sale_loyalty.menu_discount_loyalty_type_config",
+    ),
+    (
+        "sale_loyalty.menu_promotion_type_config",
+        "sale_loyalty.menu_gift_ewallet_type_config",
+    ),
+    (
+        "sale_loyalty.sale_coupon_view_form",
+        "sale_loyalty.loyalty_card_view_form_inherit_sale_loyalty",
+    ),
+    (
+        "sale_loyalty.sale_coupon_apply_code_view_form",
+        "sale_loyalty.sale_loyalty_coupon_wizard_view_form",
+    ),
+    (
+        "sale_loyalty.sale_order_view_form",
+        "sale_loyalty.sale_order_view_form_inherit_sale_loyalty",
+    ),
+]
+
+
+def _generate_random_reward_code():
+    """This function, _generate_random_reward_code, provides a random reward code. It is
+    used to create unique codes that identify rewards within the sale_loyalty module."""
+    return str(random.getrandbits(32))
+
+
+def generate_sale_order_coupon_points(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        CREATE TABLE IF NOT EXISTS sale_order_coupon_points (
+            coupon_id INT,
+            order_id INT,
+            points FLOAT,
+            create_uid INT,
+            write_uid INT,
+            create_date DATE,
+            write_date DATE
+        )
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO sale_order_coupon_points (
+            coupon_id, order_id, points, create_uid, write_uid, create_date, write_date
+        )
+        SELECT
+            id AS coupon_id,
+            order_id,
+            points,
+            create_uid,
+            write_uid,
+            create_date,
+            write_date
+        FROM loyalty_card
+        WHERE order_id IS NOT NULL;
+        """,
+    )
+
+
+def update_loyalty_program_data(env):
+    # Set sale_ok default values
+    if not openupgrade.column_exists(env.cr, "loyalty_program", "sale_ok"):
+        openupgrade.add_fields(
+            env,
+            [
+                (
+                    "sale_ok",
+                    "loyalty.program",
+                    "loyalty_program",
+                    "boolean",
+                    "bool",
+                    "sale_loyalty",
+                    True,
+                )
+            ],
+        )
+
+
+def update_sale_order_line_data(env):
+    # Fill reward_id field
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS reward_id INT
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line sol
+        SET reward_id = lr.id
+        FROM loyalty_reward lr
+        WHERE sol.product_id = lr.discount_line_product_id
+        """,
+    )
+    # Fill coupon_id field(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS coupon_id INT
+        """,
+    )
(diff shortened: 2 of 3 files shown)


legalsylvain:
/ocabot migration sale_loyalty

<dependency>Depends on : 

- loyalty : #4187</dependency>

OpenCode (approved):

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,297 @@
...
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS points_cost FLOAT
+        """,
+    )
+    openupgrade.logged_query(
+        openupgrade.logged_query(
+            env.cr,
+            """
+            UPDATE sale_order_line
+            SET point_cost = ABS(untaxed_amount_to_invoice)
```
cormaza: Should be points_cost
```suggestion
            SET points_cost = ABS(untaxed_amount_to_invoice)
```

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,297 @@
...
+        FROM loyalty_card
+        WHERE sale_order_line.order_id = loyalty_card.sales_order_id
+        """,
+    )
+    # Fill reward_identifier_code
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS reward_identifier_code CHAR
```
cormaza: Should be VARCHAR?
```suggestion
        ADD COLUMN IF NOT EXISTS reward_identifier_code VARCHAR
```

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    # Fill points_cost field
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS points_cost FLOAT
+        """,
+    )
+    openupgrade.logged_query(
```
cormaza, line 165: Should keep only one logged query, you have nested call

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,297 @@
...
+    # loyalty_card table based on certain criteria and relationships established between
+    # the loyalty_card, loyalty_reward and gift_card tables
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line AS sol
+        SET coupon_id = lc.id
+        FROM loyalty_card AS lc
+        JOIN loyalty_reward AS lr ON lc.program_id = lr.program_id
+        JOIN gift_card AS gc ON lc.program_id = gc.program_id
```
cormaza, line 225: Gift card model doesn't have program_id column, can you check what module add these field?

https://github.com/odoo/odoo/blob/15.0/addons/gift_card/models/gift_card.py

https://github.com/OCA/OCB/blob/15.0/addons/gift_card/models/gift_card.py

cormaza (changes requested):
Please check comments

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: This field is not added in any module, it is added in the migration scripts of the loyalty module in order to migrate the gift card data:

https://github.com/OCA/OpenUpgrade/blob/5912a50fa93234fe37643c1091919f780d718dd1/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py#L615-L621

I have indicated this in a comment to avoid confusion. Thanks for the corrections.

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+            create_date,
+            write_date
+        FROM loyalty_card;
+        """,
+    )
+
+
+def update_loyalty_program_data(env):
+    # Set sale_ok default values
+    openupgrade.logged_query(
```
pedrobaeza: Do this with `add_fields` with default value.

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+    )
+
+
+def merge_sale_gift_card_to_sale_loyalty_card(env):
+    # Update the coupon_id column in the sale_order_line table with the ID of the
+    # loyalty_card table based on certain criteria and relationships established between
+    # the loyalty_card, loyalty_reward and gift_card tables
+    # program_id is added to the gift_card table in the loyalty migration script.
+    table = openupgrade.get_model2table("gift.card")
+    if not openupgrade.table_exists(env.cr, table):
```
pedrobaeza: No need to do workarounds, as we know the table name:

```suggestion
    table = openupgrade.get_model2table("gift.card")
    if not openupgrade.table_exists(env.cr, "gift_card"):
```

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+def delete_sql_constraints(env):
+    # Delete constraints to recreate it
+    openupgrade.delete_sql_constraint_safely(
+        env, "sale_loyalty", "sale_order_coupon_points", "order_coupon_unique"
+    )
+
+
+def update_template_keys(env):
+    """Update template keys of the merged sale_gift_card module in loyalty_sale"""
+    openupgrade.logged_query(
```
pedrobaeza, line 185: I don't see the need of this, as they are not noupdate=1 records, and in any case, it should be `sale_gift_card`, not `sale_gif_card`.

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+        ADD COLUMN IF NOT EXISTS reward_id INT
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line sol
+        SET reward_id = lr.id
+        FROM loyalty_reward lr
+        WHERE sol.product_id = lr.discount_line_product_id
```
pedrobaeza, line 110: This seems a weak way to link rewards only through product. Isn't there another option?

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO sale_order_coupon_points (
+            coupon_id, order_id, points, create_uid, write_uid, create_date, write_date
+        )
+        SELECT
+            id AS coupon_id,
+            order_id,
+            1 AS points,
```
pedrobaeza: Why putting 1 point per coupon?

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
...
+from openupgradelib import openupgrade
+
+_deleted_xml_records = [
+    "sale_loyalty.sale_coupon_apply_code_rule",
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(
```
pedrobaeza, line 42: Load the noupdate_changes.xml

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+    )
+    reward_code = _generate_random_reward_code()
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line
+        SET reward_identifier_code = %s
+        WHERE reward_id IS NOT NULL
+        """,
+        (reward_code,),
```
pedrobaeza: Only one reward code is going to be generated, not one per record. The only way to do that is to iterate per record. And why is this needed? Shouldn't it be one identifier per reward?

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+
+def fill_code_enabled_rule_ids_from_sale_order(env):
+    """This is an m2m field, the relationship table is created in case it does not
+    exist and filled with the data. This field relates the id of a rule to the id of
+    a sales order to which it has been applied, provided that the rule belongs to an
+    applicable promotion with code."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        CREATE TABLE IF NOT EXISTS loyalty_rule_sale_order_rel (
```
pedrobaeza: This is not correct, as this is missing all the foreign keys, indexes, etc, and they won't be recreated. Do this on post if the field is not a computed, and if computed, ask me for more details on how to do it.

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,299 @@
...
+def convert_applied_coupons_from_sale_order_to_many2many(env):
+    """This conversion from One-to-Many (o2m) to Many-to-Many (m2m) involves the
+    creation of a relationship table between sale_order and loyalty_card, allowing
+    multiple sale_order records to be associated with multiple loyalty_card records.
+    This is done by inserting the ids corresponding to sale_order and loyalty_card
+    into the loyalty_card_sale_order_rel table."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        CREATE TABLE IF NOT EXISTS loyalty_card_sale_order_rel (
```
pedrobaeza: This is not correct, as this is missing all the foreign keys, indexes, etc, and they won't be recreated. Do this on post if the field is not a computed, and if computed, ask me for more details on how to do it.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: I had confused it with the default points of the points program, in this case the points come from loyalty_card.

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: discount_line_product_id is a unique product.product that is created for that reward_id, so it is the only link to get that reward_id. In addition, each program has its own rewards, so there can be no confusion as they are unique data. 
image

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: At the moment a promotion can have several rewards, but only one will be applied, this is chosen in the new wizard where all the reward options will appear. The reward line is the one that will have the identifier and it will be unique for each reward line, taking into account that only one reward per program can be applied to the same order. In short, there can be several programs applied to the same order but only one reward per program, so each line will have its own reward_id and identifier. No two lines in the same order will have the same reward_id.
image

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
pilarvargas-tecnativa: All changes done

openupgrade_scripts/scripts/sale_loyalty/16.0.1.0/pre-migration.py
pedrobaeza: But then this is not correct, as it puts the same identifier to all the records.

pedrobaeza (approved):

PR #4214, opened 2023-11-06
[14.0][FIX] payment: Delete template payment.account_invoice_payment_acquirer_onboarding

arif-m150:
This template view is no longer available in the version 14.0

### DIFF
--- a/openupgrade_scripts/scripts/payment/14.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/payment/14.0.1.0/pre-migration.py
@@ -13,3 +13,7 @@ def migrate(env, version):
         ],
         True,
     )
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (changes requested):
This view should be automatically deleted when you do an update=all with odoo (after migrating) because this view is noupdate=0.

What issues do you have?

(Also, this kind of things, should be done in post-migration file)

arif-m150:
Sorry when I do that, it doesn't automatically delete and it produces an error.

pedrobaeza:
That's because you have inherited views dangling from this view, and that's the result. You are going to have the same problem trying it programmatically. Just fix the inner problem, deleting such views before migrating if they are manually created ones, or remove/fix the modules that are adding such views.

PR #4023, opened 2023-07-04
[16.0][OU-ADD] website_partner: Nothing to do

duyquyen96:

PR #4216, opened 2023-11-06
[16.0][OU-FIX] website: rename group_website_publisher in custom templates

chienandalu:
Reviewing https://github.com/OCA/OpenUpgrade/pull/4023 I detected that this can be an issue in custom views that make use of that group.

cc @Tecnativa TT45244

### DIFF
--- a/openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
@@ -56,7 +56,28 @@ def convert_field_html_string_bootstrap_4to5(env):
                     )
 
 
+def rename_t_group_website_restricted_editor(env):
+    """Locate group directives in custom views with the former `group_website_publisher`
+    group which is now renamed to `group_website_restricted_editor`. For example, the
+    website_partner.partner_detail template"""
+    views = env["ir.ui.view"].search(
+        [
+            (
+                "arch_db",
+                "ilike",
+                "website.group_website_publisher",
+            ),
+            ("website_id", "!=", False),
+        ]
+    )
+    for view in views:
+        view.arch_db = view.arch_db.replace(
+            "website.group_website_publisher", "website.group_website_restricted_editor"
+        )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     convert_custom_qweb_templates_bootstrap_4to5(env)
     convert_field_html_string_bootstrap_4to5(env)
+    rename_t_group_website_restricted_editor(env)


openupgrade_scripts/scripts/website/16.0.1.0/end-migration.py
```diff
@@ -56,7 +56,28 @@ def convert_field_html_string_bootstrap_4to5(env):
...
+def rename_t_group_website_restricted_editor(env):
+    """Locate group directives in custom views with the former `group_website_publisher`
+    group which is now renamed to `group_website_restricted_editor`. For example, the
+    website_partner.partner_detail template"""
+    views = env["ir.ui.view"].search(
+        [
+            (
+                "arch_db",
+                "ilike",
+                "website.group_website_publisher",
```
pedrobaeza, line 68:
```suggestion
            (
                "arch_db",
                "ilike",
                "website.group_website_publisher"
            ),
```

chienandalu:
fixed :facepalm:

pedrobaeza:
Although it's a very distant possibility, you may have an string `website.group_website_publisher<whatever>` that is not for this use, so is it a way to directly check the string with the delimiters (are they single `'`  or double `"` quotes?) and maybe the while definition with the `groups` thing?

chienandalu:
> so is it a way to directly check the string with the delimiters (are they single `'` or double `"` quotes?) and maybe the while definition with the `groups` thing?

Normally it's double but it could be both. And also there could be commas (e.g: `groups='base.group_none,website.group_website_publisher'`).

To consider that possibility we could maybe use `openupgradelib.convert_html_fragment` or hook this change into the general list of replacements (https://github.com/OCA/openupgradelib/blob/c00a6662cbc9470cca8e0192cdda58ecfd805144/openupgradelib/openupgrade_160.py#L283-L295)

pedrobaeza:
The second one sounds good then.

chienandalu:
> The second one sounds good then.

I'm seeing that is going to be a little bit harder than expected. We can match by xpath the group but only remove it and add it later. That would spoil those cases which had several more groups along with the one we want to replace...

pedrobaeza:
So fragment conversions is not handling correctly the removal and addition of elements in list separated by commas?

chienandalu:
> So fragment conversions is not handling correctly the removal and addition of elements in list separated by commas?

It simply isn't supported and that case is hardcoded for classes and styles

pedrobaeza (commented):
OK, it's very limiting IMO.

Let's merge this patch here then.

pedrobaeza (approved):

chienandalu:
Yeah, maybe for a more important future migration we'll need to refactor that part

PR #4226, opened 2023-11-09
[16.0][OU-ADD] website_sale_loyalty

pilarvargas-tecnativa:
cc @Tecnativa TT44377

@chienandalu please review

### DIFF
--- a/openupgrade_scripts/scripts/website_sale_loyalty/16.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/website_sale_loyalty/16.0.1.0/pre-migration.py
@@ -1,0 +1,103 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_xmlids_renames = [
+    (
+        "website_sale_loyalty.coupon_view_tree",
+        "website_sale_loyalty.loyalty_card_view_tree_inherit_website_sale_loyalty",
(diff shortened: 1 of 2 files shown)


legalsylvain:
/ocabot migration website_sale_loyalty

<dependency>Depends on : 

- sale_loyalty : #4212</dependency>

openupgrade_scripts/scripts/website_sale_loyalty/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,102 @@
...
+        WHERE lr.program_id = gc.program_id
+        """,
+    )
+
+
+def update_template_keys(env, _key_updates):
+    for old_key, new_key in _key_updates:
+        openupgrade.logged_query(
+            env.cr,
+            "UPDATE ir_ui_view SET key = %s WHERE key = %s" % (new_key, old_key),
```
cormaza: It should be enclosed in quotes, to be treated as text
```suggestion
            "UPDATE ir_ui_view SET key = '%s' WHERE key = '%s'" % (new_key, old_key),
```

cormaza (changes requested):

openupgrade_scripts/scripts/website_sale_loyalty/16.0.1.0/pre-migration.py
pedrobaeza: That's not the way to do it. The parameters should be passed as arguments.

pedrobaeza (approved):

PR #4228, opened 2023-11-12
[14.0][OU-FIX] mass_mailing: Avoid duplicated mailing list name

pedrobaeza:
If you come from v12 or before, there was a mailing list called 'Newsletter' as data. This list dissapeared in v13, so on OpenUpgrade v13, we removed the associated XML-ID for preserving it. As now in v14 it's being introduced again (although with other XML-ID), the unique name contraint makes the migration to crash, so we need to detect if a mailing list with such name exists, and give it in advance the required XML-ID for avoiding the duplicity.

@Tecnativa TT45308


pedrobaeza:
Tested on a real migration, so merging.

PR #4238, opened 2023-11-23
[13.0][OU-ADD] stock_picking_responsible: merged into stock

robinkeunen:
responsible_id field in `stock_picking_responsible` is replaced by user_id in `stock`. The script migrates data from

```py
# stock_picking_responsible/models/stock_picking.py
responsible_id = fields.Many2one(
    comodel_name="res.partner",
    string="Responsible",
    default=lambda self: self.env.user.partner_id.id,
)
```

to

```py
# stock/models/stock_picking.py
user_id = fields.Many2one(
    'res.users', 'Responsible', tracking=True,
    domain=lambda self: [('groups_id', 'in', self.env.ref('stock.group_stock_user').id)],
    states={'done': [('readonly', True)], 'cancel': [('readonly', True)]},
    default=lambda self: self.env.user)
```

First PR on OpenUpgrade, comments welcome.

### DIFF
--- a/addons/stock/migrations/13.0.1.1/post-migration.py
+++ b/addons/stock/migrations/13.0.1.1/post-migration.py
@@ -185,6 +185,62 @@ def map_stock_location_usage(env):
     )
 
 
+def map_stock_picking_responsible_responsible_id_to_user_id(env):
+    """
+    responsible_id (partner_id) field in stock_picking_responsible is replaced by user_id (res.users)
+    We create a deactivated user for partners without user and then
+    map the partner to their user in the stock picking.
+    """
+    if not openupgrade.column_exists(env.cr, "stock_picking", "responsible_id"):
+        return
+
+    env.cr.execute(
+        f"""
+        SELECT distinct rp.id,
+                        rp.name,
+                        rp.company_id,
+                        rp.email
+        FROM stock_picking sp
+                 JOIN res_partner rp ON rp.id = sp.responsible_id
+                 LEFT JOIN res_users ru ON rp.id = ru.partner_id
+        WHERE ru.id IS NULL
+    """
+    )
+    partners_wo_user = env.cr.fetchall()
+
+    user_vals_list = []
+    for partner_id, name, company_id, email in partners_wo_user:
+        login = email if email else name
+        login = openupgrade.get_legacy_name(login).replace(" ", "_")
+        user_vals_list.append({
+            "login": login,
+            "partner_id": partner_id,
+            "company_id": company_id,
+            "active": False,
+        })
+
+    if user_vals_list:
+        env["res.users"].create(user_vals_list)
+
+    # map responsible_id to user_id
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        WITH partner_user AS (
+            SELECT sp.id AS picking_id,
+                   rp.id AS partner_id,
+                   ru.id AS user_id
+            FROM stock_picking sp
+                JOIN res_partner rp ON rp.id = sp.responsible_id
+                LEFT join res_users ru ON rp.id = ru.partner_id)
+        UPDATE stock_picking
+        SET user_id = partner_user.user_id
+        FROM partner_user
+        WHERE stock_picking.id = partner_user.picking_id;
+    """
+    )
+
+
 def fill_stock_picking_type_sequence_code(env):
     """Deduce sequence code from current sequence pattern """
     picking_types = env["stock.picking.type"].with_context(active_text=False).search([])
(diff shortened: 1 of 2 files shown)


addons/stock/migrations/13.0.1.1/pre-migration.py
```diff
@@ -85,6 +85,18 @@ def prefill_stock_picking_type_sequence_code(env):
     )
 
 
+def merge_stock_picking_responsible_module(env):
+    """
+    The stock_picking_responsible module is now merged in stock.
+    We copy the legacy column to convert data in post-migration script.
+    """
+    if openupgrade.column_exists(env.cr, "stock_picking", "responsible_id"):
+        openupgrade.copy_columns(
```
pedrobaeza: No need to copy or rename the column. It's going to be there in post.

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -185,6 +187,53 @@ def map_stock_location_usage(env):
...
+               rp.name as partner_name
+        FROM stock_picking sp
+                 JOIN res_partner rp ON rp.id = sp.{legacy_column_name}
+                 LEFT JOIN res_users ru ON rp.id = ru.partner_id
+        WHERE ru.id IS NULL
+    """
+    )
+    partner_wo_user = env.cr.fetchall()
+    for picking_id, picking_name, partner_id, partner_name in partner_wo_user:
+        _logger.warning(
```
pedrobaeza: Instead of this, what about creating a disabled user and fill it with it?

addons/stock/migrations/13.0.1.1/pre-migration.py
robinkeunen: I removed it as you suggested, should we just leave the `responsible_id` in the database ?

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -185,6 +185,63 @@ def map_stock_location_usage(env):
     )
 
 
+def map_stock_picking_responsible_responsible_id_to_user_id(env):
+    """
+    responsible_id (partner_id) field in stock_picking_responsible is replaced by user_id (res.users)
+    We create a deactivated user for partners without user and then
+    map the partner to their user in the stock picking.
+    """
+
```
pedrobaeza: **nitpicking:** Don't put empty lines inside a method.

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -185,6 +185,63 @@ def map_stock_location_usage(env):
...
+                        rp.email
+        FROM stock_picking sp
+                 JOIN res_partner rp ON rp.id = sp.responsible_id
+                 LEFT JOIN res_users ru ON rp.id = ru.partner_id
+        WHERE ru.id IS NULL
+    """
+    )
+    partners_wo_user = env.cr.fetchall()
+
+    user_values = []
```
pedrobaeza: **suggestion**: for following conventions on variables names:

```suggestion
    user_vals_list = []
```

addons/stock/migrations/13.0.1.1/post-migration.py
```diff
@@ -185,6 +185,63 @@ def map_stock_location_usage(env):
...
+
+    if user_values:
+        env["res.users"].create(user_values)
+
+    # map responsible_id to user_id
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        WITH partner_user AS (
+                      SELECT sp.id AS picking_id,
```
pedrobaeza: **suggestion**: Don't indent too much this part.

addons/stock/migrations/13.0.1.1/post-migration.py
robinkeunen: I assumed you mean only the first empty line, the others give some air and enhance readability.

robinkeunen:
Here you go Pedro, thank you for the review and suggestions :-)

addons/stock/migrations/13.0.1.1/post-migration.py
pedrobaeza: I mean all, as syntax highlighting already makes the needed visual separations. It's also recommended since PEP8 to not do it (although tolerated), and conceptually, if you need to have a code block, put a comment before that acts as separator and explaining what the block does. Anyway, it's not blocking by my part.

robinkeunen: Ok, I'll leave it as is then :-)

pedrobaeza (approved):

PR #4239, opened 2023-11-27
[13.0][OU-FIX] account_check_printing: Proper SQL string

pedrobaeza:
The formatting should be done regularly, or the value that will be used will be literally `openupgrade_legacy_13_0_check_number`.

@Tecnativa TT46186


legalsylvain (approved):
Good catch ! 
Thanks !

PR #4240, opened 2023-11-28
[16.0][OU-ADD] sale: migration

daiduongnguyen-odoo:
replace the work of: https://github.com/OCA/OpenUpgrade/pull/4004


legalsylvain:
/ocabot migration sale

<dependency>Depends on : 

- account_payment : #3920</dependency>

rvalyi:
great you introduced `_create_ir_model_data_sale_default_invoice_email_template`, `default_invoice_email_template` was indeed an issue in the other script I tried, I'll try this new one soon and report.

daiduongnguyen-odoo:
> great you introduced `_create_ir_model_data_sale_default_invoice_email_template`, `default_invoice_email_template` was indeed an issue in the other script I tried, I'll try this new one soon and report.

Oki thank you

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,77 @@
...
+    openupgrade.rename_xmlids(env.cr, _renames_xmlids)
+    _drop_sql_views(env)
+    _noupdate_switch(env)
+    _remove_table_constraints(env)
+    openupgrade_160.fill_analytic_distribution(
+        env,
+        "sale_order_line",
+        "account_analytic_tag_sale_order_line_rel",
+        "sale_order_line_id",
+    )
```
marielejeune: This method fails because of this line:
https://github.com/OCA/openupgradelib/blob/fdc1a1ab7d58b0c2850be77b761b0942c965b941/openupgradelib/openupgrade_160.py#L449

``` psycopg2.errors.UndefinedColumn: column line.analytic_account_id does not exist ```

Indeed there is no `analytic_account_id` field on `sale_order_line`. This field is on the related `sale_order`.

No idea of how to fix it right now though...

**Review by marielejeune — COMMENTED** [review]
Thanks a lot for taking this huge work again!
Here is an error that I noticed.

File: openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py (hunk 1 above)
Comment (daiduongnguyen-odoo) [reply openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py]: @marielejeune in that case, i will implement in a different way not using that method. Fill free to ping if i forget hehe, i'm a little busy right now

Comment (marielejeune) [reply openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py]: Hi @duong77476-viindoo, thinking about another solution: why not creating a temporary column `analytic_account_id` on `sale_order_line`, as the analytic account of the SOL is the one of the related SO? This would allow to still use this openupgrade method.

**Review by legalsylvain — COMMENTED** [review]
Hi @duong77476-viindoo. Thanks for finish this work.

Could you update the file ./docsource/modules150-160.rst, marking the ``sale`` module as done ?

File: openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py (hunk 1 above)
Comment (daiduongnguyen-odoo) [reply openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py]: @marielejeune Never mind, i restore back to inline implement not using the `openupgrade_160` any more, please retest if needed

**Comment by daiduongnguyen-odoo** [comment]
> Hi @duong77476-viindoo. Thanks for finish this work.
> 
> Could you update the file ./docsource/modules150-160.rst, marking the `sale` module as done ?

Done

File: openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py (hunk 1 above)
Comment (marielejeune) [reply openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py]: It works, thanks!

File: openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py [hunk 2]
```diff
@@ -0,0 +1,147 @@
...
+                sub.sale_order_line_id,
+                jsonb_object_agg(sub.analytic_account_id::text, sub.percentage)
+                    AS analytic_distribution
+            FROM sub
+            GROUP BY sub.sale_order_line_id
+        )
+
+        UPDATE sale_order_line so_line SET analytic_distribution = dist.analytic_distribution
+        FROM distribution_data dist WHERE so_line.id = dist.sale_order_line_id
+        """,
```
fkantelberg, line 105: This might be problematic to fill for most cases. I tested it with a sale order which had a `analytic_account_id` set before the migration. After the migration I see the `analytic_distribution` correctly (100% for the analytic account). and `analytic_account_id` is still set.

When I create an invoice from it this results in 200% for the analytic account because both are set in the sale.order. In `_prepare_invoice_line` Odoo actually adds 100%. This is the reason why it would work without a migration for such a case. Alternativly you might want to clear the `analytic_account_id` on the sale.order to prevent the +100%.

```sql
UPDATE sale_order SET analytic_account_id = NULL
WHERE analytic_account_id IS NOT NULL
```

fkantelberg (changes requested):

openupgrade_scripts/scripts/sale/16.0.1.2/pre-migration.py
pedrobaeza: I think then it's better to not do anything at all and leave it as it was.

fkantelberg: I rechecked the query. It might also be worth to remove the lines 80-90 (first part of UNION ALL) of the query because the 100% are added there to the distribution. Fully skipping it would mean that you also skip the 2nd part of the query and don't work against the removal of `analytic_tag_ids` on `sale.order.line`.

daiduongnguyen-odoo: @fkantelberg Can you provide an example that this query doesn't work please, a video would help if you have one.

fkantelberg: I don't know about video. Here is the an example how it looks in the database. Following is 15.0:
```
>>> so.read(["name", "state", "analytic_account_id"])
[{'id': 1,
  'name': 'AG-2024-01-0001',
  'state': 'sale',
  'analytic_account_id': (1, 'Intern')}]

>>> so.order_line.read(["order_id", "product_id", "analytic_tag_ids"])
[{'id': 1,
  'order_id': (1, 'AG-2024-01-0001'),
  'product_id': (4, 'Test'),
  'analytic_tag_ids': []}]
```

After the migration it looks like this:
```
>>> so.read(["name", "analytic_account_id"])
[{'id': 1, 'name': 'AG-2024-01-0001', 'analytic_account_id': (1, 'Intern')}]
>>> so.order_line.read(["order_id", "product_id", "analytic_distribution"])
[{'id': 1,
  'order_id': (1, 'AG-2024-01-0001'),
  'product_id': (4, 'Test'),
  'analytic_distribution': {'1': 100}}]
```

If I confirm the order and create an invoice out of it I see that the invoice has a wrong analytic distribution because the `analytic_account_id` on the `sale.order` is adding +100.
```
>>> so.mapped("invoice_ids.invoice_line_ids").read(["move_id", "analytic_distribution"])
[{'id': 1,
  'move_id': (1, 'Draft RE-2023-12-0001'),
  'analytic_distribution': {'1': 200.0}}]
```

daiduongnguyen-odoo: @fkantelberg ok you seem right although i think odoo should remove the `analytic_account_id` too just like what they did with purchase. I have removed the code from line 80 - 90 as you suggested. Please retest if needed
cc @pedrobaeza

fkantelberg: Works flawlessly and results in a correct invoice.

Regarding `analytic_account_id` there is still code in `action_confirm` which creates an analytic account for expense products for a sale.order.

fkantelberg (approved):

pedrobaeza:
Continue in #4301 for last touches before merging. Thanks for the hard work, @duong77476-viindoo 

The attribution has been respected.

PR #4179, opened 2023-10-20
[16.0][OU-ADD] payment_paypal: Migration to 16.0

victoralmau:
Migration to 16.0

Depends on:
- [x] https://github.com/OCA/OpenUpgrade/pull/4193

@Tecnativa TT44209

PR #4241, opened 2023-11-28
[16.0][OU-IMP] payment: Copy the column provider to code from payment.provider

victoralmau:
Copy the column `provider` to `code` from `payment.provider`

Related to https://github.com/OCA/OpenUpgrade/pull/4179#discussion_r1407365126

@Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/payment/16.0.2.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/payment/16.0.2.0/pre-migration.py
@@ -69,6 +69,10 @@ _field_renames = [
     ),
 ]
 
+_columns_copies = {
+    "payment_provider": [("provider", "code", "varchar")],
+}
+
 _xmlid_renames = [
     (
         "payment.action_invoice_order_generate_link",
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/payment/16.0.2.0/pre-migration.py
```diff
@@ -210,4 +214,5 @@ def migrate(env, version):
     openupgrade.rename_models(env.cr, _model_renames)
     openupgrade.rename_tables(env.cr, _table_renames)
     openupgrade.rename_fields(env, _field_renames)
+    openupgrade.copy_columns(env, _columns_copies)
```
pedrobaeza:
```suggestion
    openupgrade.copy_columns(env.cr, _columns_copies)
```

pedrobaeza:
On the analysis work file, you should move the `code` line apart from the block of "Nothing to do". And this is `[OU-IMP]`, not `[IMP]`.

pedrobaeza (approved):

PR #4242, opened 2023-11-28
[15.0][OU-ADD] website_crm: Migrate website_crm views

pilarvargas-tecnativa:
Migrates website_crm views to replace references to 'mail.mail' with 'crm.lead'. In versions 13 and 14, the creation of opportunities was configured by replacing the default form view with one that had a specific action to create a lead. In version 15, this configuration is no longer present, and the default behavior is to use the base view for sending emails. To ensure a seamless transition and retain the previous configuration of creating opportunities, this migration script modifies the relevant views associated with the website_crm module. It replaces instances of 'mail.mail' with 'crm.lead', preserving the user's preference for creating opportunities.

cc @Tecnativa  TT35811

Tested locally in a shell in v15

### DIFF
--- a/openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
+++ b/openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
@@ -1,0 +1,27 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def migrate_website_crm_views(env):
+    """Update the arch_db field by replacing the occurrences of "mail.mail" with "crm.lead"
+    in "website.contactus". This is done to adapt the functionality in Odoo v15 where the
+    default action is used to send an email and to ensure that the configuration is not
+    lost for the that the configuration is not lost to create an opportunity that in previous
+    versions was configured by default by a was configured by default by a replacement in
+    the form view."""
+    for website in env["website"].search([]):
+        website_contactus_view = website.with_context(website_id=website.id).viewref(
+            "website.contactus"
+        )
+        website_contactus_arch_db = website_contactus_view.arch_db
+        if 'data-model_name="mail.mail"' in website_contactus_arch_db:
+            new_arch = website_contactus_arch_db.replace(
+                'data-model_name="mail.mail"', 'data-model_name="crm.lead"'
+            )
+            website_contactus_view.with_context(website_id=website.id).arch = new_arch
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    migrate_website_crm_views(env)


openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    """Migrate website_crm views in the post-migration process so that the new contact form
+    views have been defined.
+    This function checks if the "website_crm" module is installed. If it is, it looks for
+    the views associated with the website and updates the arch_db field replacing the
+    "mail.mail" occurrences with "crm.lead". This is done to adapt the
+    functionality in Odoo v15 where the default action is used to send an email and ensure
+    that it does not lose configuration to create an opportunity which in previous versions
+    was set by default by a replace in the form view."""
+    if env["ir.module.module"].search(
+        [("name", "=", "website_crm"), ("state", "=", "installed")]
```
pedrobaeza: This script is inside `website_crm`, so no need to check this.

openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    ):
+        # Search views with associated website
+        website_views = env["ir.ui.view"].search(
```
pedrobaeza: I think this is not the case. The problem is if there's no custom view, the default one (website_id=False) is doing another action.

pedrobaeza (changes requested):
You have to go through each website, get `website.contactus` for each website with the ORM and the context (Odoo will fetch the master or the custom one automatically), and change the action, writing back.

openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
pilarvargas-tecnativa: Done in the correct way so that it considers whether there is a customised view or not.

openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def migrate_website_crm_views(env):
+    """Migrate website_crm views in the post-migration process so that the new contact form
+    views have been defined.
+    This function checks if the "website_crm" module is installed. If it is, it looks for
+    the views associated with the website and updates the arch_db field replacing the
+    "mail.mail" occurrences with "crm.lead". This is done to adapt the
```
chienandalu: The only view changing is the contact form, so there's no need to generalize
```suggestion
    """This is done to adapt the
```

openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+    for website in env["website"].search([]):
+        website_contactus_view = website.with_context(website_id=website.id).viewref(
+            "website.contactus"
+        )
+        website_contactus_arch_db = website_contactus_view.arch_db
+        if 'data-model_name="mail.mail"' in website_contactus_arch_db:
+            new_arch = website_contactus_arch_db.replace(
+                'data-model_name="mail.mail"', 'data-model_name="crm.lead"'
+            )
+            website_contactus_view.with_context(website_id=website.id).arch = new_arch
```
chienandalu, line 22: Didn't you already gather the record with that context?

pilarvargas-tecnativa: First the website view with the context is retrieved, this retrieves the custom view (with associated website) and in case it does not exist it will return the base view.
Then you have to update the view with the context given that in the case of not existing and get the base view you have to create the corresponding custom views with associated website. This is done because the base view is configured by default to send an email and this should not be modified in any case, it should only be updated in the copies with associated website.

openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
pilarvargas-tecnativa: thanks

openupgrade_scripts/scripts/website_crm/15.0.2.1/post-migration.py
chienandalu: I know that, but you already injected the context so it should not be needed to do it again. Did you try it?

pilarvargas-tecnativa: Tested, I thought the context was lost, it doesn't cause an error but it is redundant. Thanks

chienandalu (approved):

pedrobaeza (approved):

PR #4243, opened 2023-11-28
[15.0][OU-IMP] website: Update Contact Us Form with Company Description

pilarvargas-tecnativa:
cc @Tecnativa TT35810

### DIFF
--- a/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
@@ -1,5 +1,6 @@
 import re
 
+from markupsafe import Markup
 from openupgradelib import openupgrade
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Check CI

pilarvargas-tecnativa:
> Check CI

It is complaining that the field "work_location" does not exist in v15 and it is true, it does not exist but the table must continue to maintain that data for the sake of data preservation. In the analysis I have seen that it is replaced by a field "many2one" and precisely the error is given by the script that adapts the data to the new. Should we review the script and maybe move it to pre? or is it that the IC has got its wires crossed?

pedrobaeza:
No, that's not the error:

```
TypeError: update_contact_form_company_description() missing 1 required positional argument: 'env'
```

you are simply not calling correctly the method.

Anyway, please don't repeat code. Only put as conditional the `company.google_map_link()` part, not whole expression. And why not doing this on `website_crm`? And another thing: you are still using the master template instead of the possible custom view for the website. You have to get the view for the website using the ORM methods for it (passing the website_id as context if I don't remember bad).

pilarvargas-tecnativa:
> And why not doing this on website_crm?

This has to be done in website as any website has a contact page, as it is added by the website itself and the company information was included in this view, previously it was added with a t-call to the company description view, this has nothing to do with the type of action that can be configured when sending the form. If this was done in the website_crm module, websites that don't use that module would be left with the fake description.

> You have to get the view for the website using the ORM methods for it (passing the website_id as context if I don't remember bad).

I get the base view because there may not be custom copies but when the architecture is modified it is done with the context, so if there is a custom view it will do it on it and if not it will create it. 

https://github.com/OCA/OpenUpgrade/blob/92ea0c14eb1f901d5cfa80ed541c49eab31f0429/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py#L109

pedrobaeza:
> I get the base view because there may not be custom copies but when the architecture is modified it is done with the context, so if there is a custom view it will do it on it and if not it will create it.

ORM will handle this, returning the standard view if no custom one is found, or the custom one. Please don't do incorrect things.

pilarvargas-tecnativa:
> ORM will handle this, returning the standard view if no custom one is found, or the custom one. Please don't do incorrect things

Done correctly

openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
```diff
@@ -48,10 +49,67 @@ def update_website_form_call(env):
...
+            company.city,
+            company.zip,
+            company.country_id.name,
+            company.phone,
+            company.email,
+            google_maps_link,
+        )
+        company_description_pattern = r'<div class="col-lg-4 mt-4 mt-lg-0">(.*?)<\/div>'
+        website_contactus_arch = website_contactus_view.arch_db
+        website_contactus_matches, *_ = re.findall(
```
pedrobaeza: Instead of `findall` + `replace`, you have `sub` method for doing the replacement AFAIK.

pilarvargas-tecnativa: Yes, that option is also available. I have already made the change, thanks

pedrobaeza (approved):

PR #4250, opened 2023-12-10
[FIX][12.0 -> 13.0]Duplicate move when invoice is draft

dansanti:
In some cases, when the invoice is in draft status and has an associated move line, the previous code associated the invoice with the invoice ID. This fix associates the invoice with the associated move line's move_id.
to reproduce:
- create invoice in draft
- create a move in draft
- create a move line for this move
- reference that line to draft invoice

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -201,7 +201,7 @@ def migration_invoice_moves(env):
         vendor_display_name, cash_rounding_id, id, create_uid, create_date,
         write_uid, write_date
         FROM account_invoice ai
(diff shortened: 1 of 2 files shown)


pedrobaeza:
By definition, a draft invoice can't have an associated move, so that should be a customization on your end, so you should handle this on your side.

dansanti:
The original code try to support that case, so this means that is not a customization, I only fix the reference, because originally reference to invoice_id, but must be referenced to move id,, maybe happens when a invoice is open and late was closed.

With the original code, get not found id or is associate to another record (because invoice_id)

pedrobaeza:
Well, unless you put some clear steps to reproduce the problem, this code has been working since a lot without issues reported, so if it's something without extra custom flows, nobody has seen it.

PR #4253, opened 2023-12-13
[15.0][OU-FIX] hr_expense: mark credit account move lines excluded from invoice tab

### LINKED ISSUE #4251: [15.0] hr_expense: wrong amount update to 0.0 for some expense journal items
Hello,

I've recently found that seems to be a problem with journal entries that belong to expenses. A lot of journal items amounts were set to `0.0`, and every journal entry amount as well.

I've debugged and found this: starting from v15, as journal entries are now similar to invoices, when creating journal items, some of those items are marked as "excluded from invoice tab":

https://github.com/odoo/odoo/blob/15.0/addons/hr_expense/models/hr_expense.py#L618

In v14 this is different, as a pure "entry" journal entry.

This difference is not taken in account during migration AFAIK. Journal entries are recomputed in order to update payment status:

https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py#L24

and that method, since now expenses journal entries are processed as invoices, sets total amount to 0:

https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L1525-L1530

that also set journal items amount to 0 if such journal entry have only two items:

https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L1593-L1607

This could be an important issue IMO, maybe easy to be detected before in other migrations, so I made an in-depth review if I've lost something in my configuration (I'm making a v11 to v15 migration, not starting from v14), but everything seems to be fine. @hbrunn @legalsylvain @pedrobaeza what do you think, is it a genuine bug or a wrong local configuration?

If it's finally a migration bug, I'll made a PR. What could be the best approach, adding to pre or post migration script a SQL properly setting `exclude_from_invoice_tab` for destination expense journal items? Which could be the exact identification method for them within the journal entry?

Thanks in advance!

dalonsod:
Starting from v15, credit journal items linked to expenses will be marked as excluded from invoice tab. Existing journal entries will be recomputed due to payment status, and this information should be properly set, otherwise amounts will be unset.

Fixes #4251

### DIFF
--- a/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
@@ -10,6 +10,23 @@ def _fill_payment_state(env):
         WHERE account_move_id IS NULL
         """,
     )
+    # Starting from v15, credit journal items linked to expenses will be marked
+    # as excluded from invoice tab. Existing journal entries will be recomputed
+    # below, and this information should be properly set, otherwise amounts
+    # will be unset
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_move_line
+        SET exclude_from_invoice_tab=(
+            CASE WHEN credit > 0.0 THEN true
+            WHEN debit > 0.0 THEN false
+            ELSE exclude_from_invoice_tab
+            END
+        )
+        WHERE expense_id IS NOT NULL
+        """,
+    )
     # Recompute payment_state for the moves associated to the expenses, as on
     # v14 these ones were not computed being of type `entry`, which changes now
     # on v15 if the method `_payment_state_matters` returns True, which is the


openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
```diff
@@ -10,6 +10,23 @@ def _fill_payment_state(env):
...
+        env.cr,
+        """
+        UPDATE account_move_line
+        SET exclude_from_invoice_tab=(
+            CASE WHEN credit > 0.0 THEN true
+            WHEN debit > 0.0 THEN false
+            ELSE exclude_from_invoice_tab
+            END
+        )
+        WHERE expense_id IS NOT NULL
```
pedrobaeza, line 27: For touching less rows:

```suggestion
        UPDATE account_move_line
        SET exclude_from_invoice_tab=true
        WHERE expense_id IS NOT NULL
        AND credit > 0
```

dalonsod: Yes, that's correct, and I've verified that such account move lines have `exclude_from_invoice_tab` unset for a v14 database, indeed .

But I've found this problem: used OpenUpgrade with a database from v11 to v15, and every v14 account move line linked to expenses were migrated with `exclude_from_invoice_tab=true`, I still don't know why. That's why I've preferred this approach.

Anyway, if you think is more accurate your approach, no problem changing it.

pedrobaeza: So is it a problem that both lines have `exclude_from_invoice_tab=true`?

And do you have `hr_expense_invoice` installed to know what happen to these items?

dalonsod: > So is it a problem that both lines have `exclude_from_invoice_tab=true`?

Yes, according to

https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L1525-L1530 

for this kind of moves, total_amount is set to 0.0 if `exclude_from_invoice_tab` is the same value (no matter `true` or `false` because with `false` both balances are added and sum is `0.0`, and with `true` both are ignored), one of them must be `true` and the other `false`, as v15 new code indicates.

> And do you have `hr_expense_invoice` installed to know what happen to these items?

Not installed in origin. Should I install it?

pedrobaeza: The `hr_expense_invoice` is to check if the journal entries linked to both the invoice and the expense are correctly kept, as for example the tax line should be excluded from invoice tab. I'm having now the doubt if your expense contains taxes, that this is the correct approach. Please check how it's done in v15.

dalonsod: OK, I'll take a look at v15 implementation and will update. And in my case, every expense comes without taxes.

Thanks for all Pedro.

pedrobaeza:
Any news on this?

dalonsod:
Not yet, in this case migration tasks took my whole time and I couldn't test it before migration was completed (matter of priorities and deadlines, as you can guess), so I simply applied the proposed patch during final migration stuff. Sorry 😞

Anyway, I'll analyze `hr_expense_invoice` addon in this context, and try to update. Unfortunately, I have no new migrations scheduled at this time.

pedrobaeza:
Fixed in #4386

PR #4258, opened 2023-12-15
[MIG] sale_product_set_layout: Rename to sale_product_set

CRogos:
Related to:
https://github.com/OCA/sale-workflow/pull/2786
https://github.com/OCA/product-attribute/pull/1452


CRogos:
@pedrobaeza I think we need to add this to the upgrade as well?

pedrobaeza (approved):
It's merged, not renamed, but the code is correct.

Issue #4257, opened 2023-12-15
account.payment.method.line not connecting when migrating from 14 -> 15


Meisolsson:
## Module
payment

## Describe the bug
Between version 14 and 15 the `journal_id` value became a computed field updated via `account.payment.method.line` objects. Currently this is not correctly migrated as the *Payment Journal* for the payment provider we installed it is empty after migration.

There is a migration for this in `post-migration.py` for the `payment` module but it doesn't do anything. Since the updated `account.payment.method` are created in the provider modules this migration runs before the required values are exists in the database. Current migration for reference [post-migration](https://github.com/OCA/OpenUpgrade/blob/a5e43bfeb5e96f44105d0e59d9507ee5bc6529b9/openupgrade_scripts/scripts/payment/15.0.2.0/post-migration.py#L47).

I propose to move this migration step to a `end-migration.py` so that all values exists before it runs.

## To Reproduce
**Affected versions**: 15

1. Create a new database with payment and one of the included payment provider. (I've used payment_adyen)
2. Run migration for v14 -> 15

**Expected behavior**
If we open the payment acquirer and look at the *Payment Journal* it should be set to the payment journal created by the payment provider we installed.

**Additional context**

pedrobaeza:
Hi, Henrik, are you sure that's the problem? The SQL should match all the existing payment providers, no matter if later the code is changed.

Meisolsson:
Well what we are matching here is not the provider. The connection is `payment.acquirer` -> `account.payment.method` -> `account.payment.method.line` -> `account.journal`. In version 14 the `account.payment.method.line` is not there and the `account.payment.method` is only created by the system.

In version 15 each provider creates a new `account.payment.method` via the XML. Then in the migration we find all `account.payment.method` that matches the `payment.acquirer` and that matches the `account.journal`. So without the `account.payment.method` the connection can never be done by the migration as this point.

So at the point that the migration runs it will not find the newly created `account.payment.method`.

pedrobaeza:
OK, please do a PR moving the method to end-migration. Add in the docstring of the method the explanation why it should be there.

PR #4259, opened 2023-12-15
[15.0][FIX] payment: Wait for payment methods to be created

Meisolsson:
Since other modules create account.payment.method in this migration we need to wait for them before finding all payment methods and and connecting them to the journal and acquirer.

Discussed in #4257

### DIFF
--- a/openupgrade_scripts/scripts/payment/15.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/payment/15.0.2.0/post-migration.py
@@ -44,31 +44,12 @@ def fill_payment_transaction_partner_state_id(env):
     )
 
 
-def create_account_payment_method_line(env):
-    # Create account payment method lines from account payment methods
-    openupgrade.logged_query(
-        env.cr,
-        """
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change the commit name to include the module name.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

openupgrade_scripts/scripts/payment/15.0.2.0/post-migration.py
```diff
@@ -68,7 +68,6 @@ def migrate(env, version):
     fill_payment_transaction_tokenize(env)
     fill_payment_transaction_partner_state_id(env)
     fill_payment_transaction_last_state_change(env)
-    create_account_payment_method_line(env)
```
pedrobaeza, line 71: Remove the method from here.

Meisolsson:
My bad, I was sure I had cleaned up the old migration and I've must have missed the note about adding the module name. Should be fixed now.

pedrobaeza (approved):

PR #4264, opened 2023-12-19
[OU-ADD] calendar_sms module

legalsylvain:


legalsylvain:
@pedrobaeza : I have a doubt. Is it a nothing todo ( like https://github.com/OCA/OpenUpgrade/pull/4034), or do we have to do something in that case ? 

thanks.

pedrobaeza:
If it's noupdate=0, then you don't have to explicitly anything, as they are removed naturally by the ORM, so here it's "Nothing to do".

PR #4270, opened 2023-12-21
[14.0][OU-IMP] account: Post statements with reconciled items

pedrobaeza:
If you have statements in new state (open), but you have already performed some reconciliations on v13, you need to post the statements to continue doing reconciliations, and for having a consistent state between the lines of such statement, so we call the post button for achieving that, as only changing the statement state won't post the remaining journal entries.

@Tecnativa TT40209

### DIFF
--- a/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
@@ -313,7 +313,9 @@ def create_account_tax_report_lines(env):
             break
 
 
-def post_statements_with_unreconciled_lines(env):
+def post_statements(env):
+    """These 2 inconsistent use cases require to put the statement as posted."""
+    # Confirmed statements that have unreconciled items
     openupgrade.logged_query(
         env.cr,
         """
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -324,6 +326,16 @@ def post_statements_with_unreconciled_lines(env):
...
+    # New statements with reconciled items
+    openupgrade.logged_query(
+        env.cr,
+        """SELECT bst.id FROM account_bank_statement bst
+        JOIN account_bank_statement_line bstl ON bstl.statement_id = bst.id
+        WHERE bst.state = 'open' AND bstl.is_reconciled IS TRUE
+        """,
+    )
+    stmt_ids = list({x[0] for x in env.cr.fetchall()})
+    env["account.bank.statement"].browse(stmt_ids).button_post()
```
MiquelRForgeFlow, line 338: why not do this by query?

pedrobaeza: Already explained in the main comment. The remaining moves won't be posted.

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/account/14.0.1.1/post-migration.py
```diff
@@ -324,6 +326,16 @@ def post_statements_with_unreconciled_lines(env):
             AND bstl.is_reconciled IS DISTINCT FROM TRUE
         """,
     )
+    # New statements with reconciled items
+    openupgrade.logged_query(
+        env.cr,
+        """SELECT bst.id FROM account_bank_statement bst
```
MiquelRForgeFlow, line 332:
```suggestion
        """SELECT DISTINCT bst.id FROM account_bank_statement bst
```

pedrobaeza: Not needed, as I deduplicate later doing a Python set.

MiquelRForgeFlow: Oh, right

PR #4271, opened 2023-12-22
[15.0][OU-FIX] don't fail when there's an unknown field in a view

hbrunn:
given we don't raise in `raise_view_error`, we get an unbound local error after https://github.com/OCA/OCB/blob/15.0/odoo/addons/base/models/ir_ui_view.py#L1838 if a field used in a view doesn't exist. Seems fine to me to just ignore this then. If you agree I'll cherry pick to v16

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
@@ -61,9 +61,21 @@ def _raise_view_error(
             )
 
 
+def _check_field_paths(self, node, field_paths, model_name, use):
+    """Ignore UnboundLocalError when we squelched the raise about missing fields"""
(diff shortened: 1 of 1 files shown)


legalsylvain (commented):
Not sure to understand : on which variable there is an UnboundLocalError error ?

hbrunn:
in the line just below my link, it accesses `field`, but field hasn't been bound because the assignment failed with a key error. Without our patched `_raise_view_error` we'd never get there because the original raises

legalsylvain (approved):
thanks for the explanation.

Makes sense.

pedrobaeza (approved):
I have never faced this one, but it's OK from a technical PoV.

hbrunn:
thanks, here the same for v16: https://github.com/OCA/OpenUpgrade/pull/4272

PR #4276, opened 2024-01-04
[15.0][FIX] website_event_track: contact_email

chienandalu:
If we don't set parenthesis in this query condition we'll be getting the same contact every time.

cc @Tecnativa TT46915

please check @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
:facepalm:

PR #4278, opened 2024-01-18
[OU-IMP] apriori : pos_order_line_no_unlink features has been merged into Point of Sale. 

legalsylvain:
See related PR : https://github.com/OCA/pos/pull/989

cc reviewer of the previous PR : @enriquemartin, @rafaelbn, @carmenbianca, @FrancoMaxime

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -79,6 +79,8 @@ merged_modules = {
     "coupon_commercial_partner_applicability": "loyalty_partner_applicability",
     # OCA/sales-workflow
     "sale_product_set_layout": "sale_product_set",
+    # OCA/pos
+    "pos_order_line_no_unlink": "point_of_sale",
     # OCA/purchase-workflow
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/apriori.py
```diff
@@ -79,6 +79,8 @@
     "coupon_commercial_partner_applicability": "loyalty_partner_applicability",
     # OCA/sales-workflow
     "sale_product_set_layout": "sale_product_set",
+    # OCA/pos
```
pedrobaeza, line 78: Please insert this in alphabetically order (of repos)

legalsylvain: The original file is not ordered correctly. I fixed it.

pedrobaeza (approved):

openupgrade_scripts/apriori.py
FrancoMaxime: @legalsylvain Nice PR. Thanks for the work!

PR #4282, opened 2024-01-23
[FIX][16.0] website: sync access token of partner on website visitor

daiduongnguyen-odoo:
Following pr https://github.com/odoo/odoo/pull/110870 this is needed

### DIFF
--- a/openupgrade_scripts/scripts/website/16.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/website/16.0.1.0/pre-migration.py
@@ -124,6 +124,21 @@ def _mig_s_progress_steps_contents(env):
         view.arch_db = env["ir.ui.view"]._pretty_arch(arch)
 
 
+def _sync_website_visitor_access_token(env):
+    """
+    Following pr https://github.com/odoo/odoo/pull/110870
+    this is needed
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE website_visitor
+            SET access_token = partner_id::text
+        WHERE partner_id IS NOT NULL
+        """,
+    )
+
+
 @openupgrade.migrate()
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
The access token is not computed that way:

https://github.com/odoo-dev/odoo/blob/98163e998763a954477e9785af81a8de48b57444/addons/website/models/website_visitor.py#L35

and you are here overwriting all the existing access tokens, no matter if they were previously generated or not.

Searching on previous versions, the access token exists since the beginning of this model (v13), so there's nothing to do on this side IMO.

daiduongnguyen-odoo:
@pedrobaeza Did you check the PR i mention, just try install website module , migrate it to v16  ,  create some visitor using same portal account, then try to merge partner and you will get error. Or you need me to record a video for you ?

daiduongnguyen-odoo:
@pedrobaeza This happened will our production database after we migrated it to v16 , of course we have numerous record of visitor which has access_token is not partner_id , therefor we have to do a fix using migration script (Mean that if your database is alreay migrate to v16 and you haven't had this code in this PR , you will need to do something like merge the visitor yourself using another migration script)

pedrobaeza:
The question is that this is not an OU fix (although we can maybe handle it here). It's something that they screwed up when adding the token feature, not considering the effect when merging partners. Apart of judging if it's correct to put `partner_id` as the token in that cases (very different from the normal generated tokens), and not knowing the real utility of this token, this can happen the same in any DB from 13.0 that has partners merged. Thus, this should be added to all OU versions to get fixed "the next time you migrate to a major version".

Coming again to this PR with all the previous considerations, the code here is incorrect, reassigning the token for all the records, not only those without the token.

pedrobaeza:
Closing as no answer.

PR #4284, opened 2024-01-28
[OU-IMP] apriori. OCA/pos_invoicing module is now native in 'point_of_sale'

legalsylvain:
due to the call of _apply_invoice_payments during order / invoice creation.

https://github.com/odoo/odoo/blob/15.0/addons/point_of_sale/models/pos_order.py#L648


rvalyi (commented):
LGTM

PR #4289, opened 2024-01-31
[16.0][OU-ADD] project: migration script

remytms:
Cortesy @SonCrits 

Replace #3989 

Speed improvements can be done by computing some new computed fields via SQL request. But I will not do that in this PR. As it is not mandatory.


remytms:
I don't remember if I should modify the file `noupdate_changes.xml`.

pedrobaeza:
Yes if something shouldn't be imported, but commenting the lines.

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,99 @@
...
+obsolete model project.delete.wizard [transient]
+# NOTHING TO DO: wizard removed
+
+---Fields in module 'project'---
+project      / account.analytic.tag     / task_ids (many2many)          : DEL relation: project.task
+project      / project.task             / analytic_tag_ids (many2many)  : DEL relation: account.analytic.tag
+# NOTHING TO DO: lost feature
+
+project      / project.milestone        / task_ids (one2many)           : NEW relation: project.task
+project      / project.project          / allow_milestones (boolean)    : NEW hasdefault: default
```
pedrobaeza: The default of this one is not sure to be `True`, so better to initialize the column manually with True.

openupgrade_scripts/scripts/project/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,139 @@
...
+        ],
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE project_task task
+        SET is_closed = CASE
+            WHEN stage.fold = TRUE THEN TRUE
+            ELSE FALSE
+        END
```
pedrobaeza:
```suggestion
        SET is_closed = stage.fold
```

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,99 @@
...
+project      / project.task             / is_closed (boolean)           : is now stored
+# DONE: pre-migration: Add new column & set value for it
+
+project      / project.task             / ancestor_id (many2one)        : NEW relation: project.task, isfunction: function, stored
+project      / project.task             / is_analytic_account_id_changed (boolean): NEW isfunction: function, stored
+# DONE: pre-migration: pre-compute value for new computed fields
+
+project      / project.task             / is_blocked (boolean)          : NEW isfunction: function, stored
+project      / project.task             / milestone_id (many2one)       : NEW relation: project.milestone, hasdefault: compute
+# TODO (speed improvement): pre-migration: pre-compute value for new computed fields
```
pedrobaeza, line 45: Is this going to be done?

pedrobaeza (changes requested):
Delete translations for noupdate=1 records.

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
remytms: No, I'm not going to do this. I focus on getting something that works.

remytms:
I will work on this tomorrow. :)

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
remytms: I've tested a bit. The default value is set to conform to the current value of the `project.group_project_milestone`.

On a fresh install of `project` in 16.0, the milestone option is set to False. Which leads to a default value set to False when a project is created.

With demo data in 16.0 the milestone option is set to True, which leads to projects with a default value set to True.

Doing nothing in migration script leads to a null value for the `allow_milestone` column in database.

I made a proposal of setting the `allow_milestone` field to False in pre-migration script. It seams to work correctly with demo data also.

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,101 @@
...
+# NOTHING TO DO: new features
+
+project      / project.tags             / project_ids (many2many)       : NEW relation: project.project
+project      / project.tags             / task_ids (many2many)          : NEW relation: project.task
+# NOTHING TO DO: reverse relation that already exists on project.project and project.task
+
+pad_project  / project.project          / description_pad (char)        : DEL
+pad_project  / project.project          / use_pads (boolean)            : DEL
+pad_project  / project.task             / description_pad (char)        : DEL
+# TODO: Module pad_project has been removed
```
remytms, line 23: @pedrobaeza I'm not sure what to do with this. Should I rename the table with a legacy name to ensure that data are not deleted ? Or should I just do nothing ?

openupgrade_scripts/scripts/project/16.0.1.2/pre-migration.py
remytms: Thanks. Fixed.

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
pedrobaeza: I don't think that should be the analysis. The question is that on v15, there's no check for choosing if showing milestones or not, so a user always sees them all. So if you migrate from v15, you should mark it as True for preserving that the milestones are shown for all the existing projects. Later on, you will be able to unmark some of them if you want to, or the default will apply according the group.

Do you agree with my reasoning?

remytms: Ok, I understand your analysis. I agree with you.

remytms: Done, `allow_milestones` is created and set to True in pre-migration script.

openupgrade_scripts/scripts/project/16.0.1.2/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+    "mail_template_data_project_task",
+    "project_manager_all_project_tasks_rule",
+    "project_message_user_assigned",
+    "rating_project_request_email_template",
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "project", "16.0.1.2/noupdate_changes.xml")
```
robinkeunen, line 17: I'd suggest not  the content of `rating_project_request_email_template` : what is the best practice on this ?

robinkeunen (changes requested):

openupgrade_scripts/scripts/project/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2023 Trần Trường Sơn
+# Copyright 2023 Rémy Taymans
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import openupgradelib as openupgrade
```
Linekio: It should be `from openupgradelib import openupgrade`

remytms: Indeed. Fixed. :)

openupgrade_scripts/scripts/project/16.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.add_fields(env, _new_fields)
+    _set_task_type_fold_if_is_closed(env)
+    _fill_project_task_is_closed(env)
+    _fill_project_last_update_status_if_null(env)
+    _compute_project_task_ancestor_id(env)
+    _compute_project_task_is_analytic_account_id_changed(env)
```
remi-filament, line 165: Hi @remytms 
as per #4313 you may want to also drop view project_task_burndown_chart_report:
```suggestion
    _compute_project_task_is_analytic_account_id_changed(env)
    
    # Remove SQL view project_task_burndown_chart_report not used anymore in Odoo v16.0
    openupgrade.logged_query(
        env.cr, "DROP VIEW IF EXISTS project_task_burndown_chart_report CASCADE"
    )
```

remytms: @remi-filament Thank you. :) It is added. :)

openupgrade_scripts/scripts/project/16.0.1.2/post-migration.py
remytms: @legalsylvain Do you have an opinion about loading new text for mail template? Should we remove it from `noupdate_changes.xml`, or should we keep it and rewrite user data?

pedrobaeza: New texts are always added (and they usually imply changes in variables as well). And translations should be removed as well.

openupgrade_scripts/scripts/project/16.0.1.2/upgrade_analysis_work.txt
pedrobaeza: I would say "Nothing to do", as the content is dumped on save to the description fields, and the collaboration comes now from web editor itself.

pedrobaeza (approved):

PR #4293, opened 2024-02-02
[16.0][OU-ADD] stock_account: set display_type of cogs account move line

robinkeunen:
superseeds https://github.com/OCA/OpenUpgrade/pull/3971

Nothing to do in stock_account regarding the models and views.

Fill the display type for journal items corresponding to Cost of Good Sold lines (COGS) for customer invoices. In v15, the lines were set to is_anglo_saxon_line == True, in v16 field display_display is used.

openupgrade account migration script already sets display_type in [_account_move_fast_fill_display_type](https://github.com/ngochung207-viindoo/OpenUpgrade/blob/0b2ea483314d32366bb3138336fe0b3ea584ab1f/openupgrade_scripts/scripts/account/16.0.1.2/pre-migration.py#L178) (pre-migration)

I struggled setting up a database issuing cogs lines so I relied on my understanding of the code to write the sql command.


docsource/modules150-160.rst
```diff
@@ -770,7 +770,7 @@ Module coverage 15.0 -> 16.0
 +-------------------------------------------------+----------------------+-------------------------------------------------+
 | stock                                           |                      |                                                 |
 +-------------------------------------------------+----------------------+-------------------------------------------------+
-| stock_account                                   |                      |                                                 |
+| stock_account                                   |  Nothing to do       |                                                 |
```
remytms:
```suggestion
| stock_account                                   |  Done                |                                                 |
```

remytms (changes requested):

docsource/modules150-160.rst
robinkeunen: fixed

remytms (approved):
LGTM. Tests does not pass, but it seams related to the `stock` module.

andreampiovesana (approved):
LGTM

openupgrade_scripts/scripts/stock_account/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Models in module 'stock_account'---
+---Fields in module 'stock_account'---
+stock_account / account.move.line        / is_anglo_saxon_line (boolean) : DEL
+stock_account / account.move.line        / stock_valuation_layer_ids (one2many): NEW relation: stock.valuation.layer
+stock_account / stock.valuation.layer    / account_move_line_id (many2one): NEW relation: account.move.line
+stock_account / stock.valuation.layer    / price_diff_value (float)      : NEW
+# NOTHING TO DO: new feature
```
chienandalu, line 7: Shouldn't it be pointed that this would be addresed in `purchase_stock` (I'm not sure if it's necessary/possible/worth it) to rebuild past values

pedrobaeza:
The explanations on the work file are very few, but let's move on with this. `price_diff_value` should be reviewed later.

PR #3994, opened 2023-07-03
[16.0][OU-ADD] hr_recruitment

nguyenvietlam0640:

PR #4296, opened 2024-02-08
[16.0][MIG] hr_recruitment: Migration scripts

robinkeunen:
Superseeds https://github.com/OCA/OpenUpgrade/pull/3994

@victoralmau I deleted the fields from the noupdate if the odoo15 template was still valid. I'm not sure this is the way to do, I can reintroduce it if needed.


openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
```diff
@@ -7,121 +7,27 @@
...
-                <t t-if="object.job_id.address_id.zip"><t t-out="object.job_id.address_id.zip or ''">98450</t><t t-set="location" t-value="'%s, %s' % (location, object.job_id.address_id.zip)"/></t>
-                <br/>
-                <t t-if="object.job_id.address_id.country_id.name"><t t-out="object.job_id.address_id.country_id.name or ''">Argentina</t><br/><t t-set="location" t-value="'%s, %s' % (location, object.job_id.address_id.country_id.name)"/></t>
-                <br/>
-            </div>
-          </td>
-        </tr>
-      </table>
-    </field>
-    <field name="name">Recruitment: Application Acknowledgement</field>
```
remytms: Comment these lines instead of removing it from the file.

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
```diff
@@ -7,121 +7,27 @@
...
-                <t t-if="object.job_id.address_id.zip"><t t-out="object.job_id.address_id.zip or ''">98450</t><t t-set="location" t-value="'%s, %s' % (location, object.job_id.address_id.zip)"/></t>
-                <br/>
-                <t t-if="object.job_id.address_id.country_id.name"><t t-out="object.job_id.address_id.country_id.name or ''">Argentina</t><br/><t t-set="location" t-value="'%s, %s' % (location, object.job_id.address_id.country_id.name)"/></t>
-                <br/>
-            </div>
-          </td>
-        </tr>
-      </table>
-    </field>
-    <field name="name">Recruitment: Interest</field>
```
remytms: Idem, comment instead of deleting. Keep the explanation you put at the end. :)

remytms (changes requested):

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
robinkeunen: done

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
robinkeunen: done

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,13 @@
+from openupgradelib import openupgrade
+
+translations_to_delete = [
+    "hr_recruitment.group_hr_recruitment_user",
```
victoralmau:
```suggestion
    "group_hr_recruitment_user",
```

victoralmau (changes requested):

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/post-migration.py
robinkeunen: done

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
```diff
@@ -7,7 +7,9 @@
     <field name="description">The user interacting with the application as interviewer don't need any specific access. They'll have access thanks to their interviewer assignation.</field>
   </record>
   <record id="email_template_data_applicant_congratulations" model="mail.template">
-    <field name="body_html" type="html">
+  <!--    The default template from version 15.0 should still work in version 16.0
+          No need to erase user data -->
+    <!-- <field name="body_html" type="html">
```
pedrobaeza: If the template has changed, although the old one works, we should overwrite it, so we benefit from the new format/redaction/etc introduced in the new version. I know that this is counter-productive for customized templates, but it's not the usual (most of the installations keep the standard templates). One fine-tuning would be to detect if we have the original, and only overwrite in such case, but that's something very hard to be implemented, so the decision was some time ago to always overwrite.

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,65 @@
...
+# NOTHING TO DO: new transient model
+
+---Fields in module 'hr_recruitment'---
+hr_recruitment / hr.applicant             / interviewer_ids (many2many)   : NEW relation: res.users
+hr_recruitment / hr.applicant             / linkedin_profile (char)       : NEW
+hr_recruitment / hr.job                   / _order                        : _order is now 'sequence, name asc' ('sequence, state desc, name asc')
+hr_recruitment / hr.job                   / extended_interviewer_ids (many2many): NEW relation: res.users, isfunction: function, stored
+hr_recruitment / hr.job                   / interviewer_ids (many2many)   : NEW relation: res.users
+
+# NOTHING TO DO: new fields
```
pedrobaeza: For readability, it's better to not leave the upper empty line, as this way, the comment stands in the middle of both the above and bottom blocks, not having clear to which one belongs:

```suggestion
# NOTHING TO DO: new fields
```

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,65 @@
...
+
+hr_recruitment / hr.recruitment.source    / _inherits                     : DEL _inherits: {'utm.source': 'source_id'}
+
+# NOTHING TO DO
```
pedrobaeza: Always indicate why is nothing to do (for your future you and for the rest):

```suggestion
# NOTHING TO DO: Now the inheritance is over `utm.source.mixin`, which in fact adds the same `source_id` required field + related name
```

pedrobaeza (changes requested):

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/upgrade_analysis_work.txt
robinkeunen: :+1:

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
robinkeunen: i ended up uncommenting the whole suggestion as per perdro's suggestion

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
robinkeunen: Thanks for clarifying :ok_hand:

robinkeunen:
Thanks for the review, I fixed accordingly.

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/post-migration.py
```diff
@@ -0,0 +1,13 @@
+from openupgradelib import openupgrade
+
+translations_to_delete = [
+    "group_hr_recruitment_user",
```
pedrobaeza, line 4: The translations you have to delete are the mail template ones:

```suggestion
    "email_template_data_applicant_congratulations",
    "email_template_data_applicant_interest", 
```

openupgrade_scripts/scripts/hr_recruitment/16.0.1.1/noupdate_changes.xml
pedrobaeza: @legalsylvain is this rationale included somehow in the migration scripts guide you wrote? If not, it would be interested to be added.

pedrobaeza:
As we need to move forward with dependent modules, I'm superseding this in #4333 with my latest comment attended. Thanks for the work.

PR #4301, opened 2024-02-13
[16.0][OU-ADD] sale: Migration scripts

pedrobaeza:
Supersedes #4240 adding last touches, as I can't push on their branch, always respecting existing attribution of course.

@Tecnativa TT42276


daiduongnguyen-odoo:
Thanks @pedrobaeza

PR #4291, opened 2024-02-01
[16.0][OU-ADD] stock

marielejeune:
This Pull Request cherry-picks commits from https://github.com/OCA/OpenUpgrade/pull/3966 where a huge work was already carried (thanks to hoangtiendung070797 for that) but where the last review comments are waiting since a long time.

More precisely, these 2 changes were performed, following the comments on the other PR:
* Method `rename_fields` already performs the column renaming, so calling `rename_columns` is useless.
* Fixed SQL query in `_compute_stock_location_replenish_location`.

PR #4070, opened 2023-07-08
[16.0][OU-ADD] account: Migration

ngochung207:
Video test localhost:
https://youtu.be/f1vUk11DpiU

PR #4302, opened 2024-02-14
[16.0][OU-ADD] repair: Migration scripts for v16

remi-filament:


pedrobaeza:
/ocabot migration repair

<dependency>Depends on : 

- stock : #4291
- sale_management : #4014

carlos-lopez-tecnativa (approved):
Please rebase the branch to run the CI.
Tested locally and it’s working fine.

carlos-lopez-tecnativa:

remi-filament:
Thanks @carlos-lopez-tecnativa rebase done

MiquelRForgeFlow (approved):
LGTM

openupgrade_scripts/scripts/repair/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def _compute_repair_price_total(env, model):
+    fees = env[model].search([])
+    fees._compute_price_total_and_subtotal()
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env.cr, "repair", "16.0.1.0/noupdate_changes.xml")
```
pedrobaeza, line 9: Reset the translations for the fields "name" and "description" of `mail_template_repair_quotation`

openupgrade_scripts/scripts/repair/16.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+    _compute_repair_price_total(env, "repair.fee")
+    _compute_repair_price_total(env, "repair.line")
```
pedrobaeza: This is totally useless.

```suggestion
```

Odoo is already computing these fields:

image

In fact, the improvement will be to do this by SQL instead.

openupgrade_scripts/scripts/repair/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,26 @@
...
+
+repair       / repair.order             / location_id (many2one)        : now a function
+repair       / repair.order             / product_uom (many2one)        : now a function
+# NOTHING TO DO: field already existed (was a onchange)
+
+repair       / repair.order             / picking_id (many2one)         : NEW relation: stock.picking
+repair       / stock.picking            / repair_ids (one2many)         : NEW relation: repair.order
+# NOTHING TO DO: new relation
+
+repair       / stock.picking.type       / is_repairable (boolean)       : NEW hasdefault: compute
```
pedrobaeza, line 19: Is the default a good value?

pedrobaeza (changes requested):

openupgrade_scripts/scripts/repair/16.0.1.0/upgrade_analysis_work.txt
remi-filament: Hi @pedrobaeza thanks for your review
I think it is, it is checking whether the type is not incoming and has other types pointing to it with for return.
Anyhow, this new field allows then to create repair from the incoming picking so I suppose it is a good thing to have it possible by default ?

openupgrade_scripts/scripts/repair/16.0.1.0/post-migration.py
remi-filament: Fine I remove it then.
these methods are calling account.tax compute_all() function which is quite complex and I think is better done in python rather than in SQL.

openupgrade_scripts/scripts/repair/16.0.1.0/post-migration.py
remi-filament: Done

pedrobaeza (approved):

PR #4309, opened 2024-02-15
[14.0] Rename module delivery_roulier_laposte to delivery_roulier_laposte_fr

alexis-via:
Rename module delivery_roulier_laposte to delivery_roulier_laposte_fr (module from OCA/delivery-carrier)

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -95,6 +95,8 @@ merged_modules = {
     "theme_graphene_blog": "theme_graphene",
     # odoo/enterprise
     "hr_holidays_gantt_calendar": "hr_holidays_gantt",
+    # OCA/delivery-carrier
+    "delivery_roulier_laposte": "delivery_roulier_laposte_fr",
     # OCA/intrastat-extrastat
     "hs_code_link": "product_harmonized_system_delivery",
     # OCA/event


florian-dacosta (approved):

MiquelRForgeFlow (changes requested):
I don't see where `delivery_roulier_laposte` module is (repo? version?), I cannot find it.

rvalyi:
> I don't see where `delivery_roulier_laposte` module is (repo? version?), I cannot find it.

The module is here https://github.com/OCA/delivery-carrier/tree/14.0/delivery_roulier_laposte_fr now I cannot find the former delivery_roulier_laposte. Was it ever publised in the OCA in some branch before 14.0?? If not, I'm not sure this rename should be part of the script. cc @alexis-via @florian-dacosta

legalsylvain:
The module is available here : https://github.com/akretion/delivery-carrier/tree/12-demo-laposte/delivery_roulier_laposte

ref : https://odoo-code-search.com/ocs/search?q=delivery_roulier_laposte

Indeed, never shared in OCA. 

@alexis-via, for such situation (previous version custom / new version OCA), you can use a dedicated branch like this one : 
https://github.com/OCA/OpenUpgrade/pull/4299

Issue #4281, opened 2024-01-19
[16.0][account] Missing field and field information for account


mholthausen:
## Module
account (account.move and account.payment)

## Describe the bug
I have installed a fresh odoo15 and installed the app "account". After the upgrade to odoo16 I get the following error when I try to open the view of the Invoicong app in odoo:

```text
Error: Missing field string information for the field 'authorized_transaction_ids' from the 'account.move' model
    at http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6753:258
    at traverse (http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6748:200)
    at http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6748:242
    at _.each._.forEach (http://localhost:8070/web/assets/134-6323ddf/web.assets_common.min.js:63:404)
    at traverse (http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6748:211)
    at http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6748:242
    at _.each._.forEach (http://localhost:8070/web/assets/134-6323ddf/web.assets_common.min.js:63:404)
    at traverse (http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6748:211)
    at processArch (http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6752:180)
    at generateLegacyLoadViewsResult (http://localhost:8070/web/assets/135-057a973/web.assets_backend.min.js:6750:106)
```

I tried to run an upgrade of the app "account" from the App view. But during this process, I get the following error
(it says: the field "amount_available_for_refund" does not exist in model "account.payment"):

```text
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1591, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/usr/lib/python3/dist-packages/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1618, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/usr/lib/python3/dist-packages/odoo/http.py", line 1822, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
[...]

PR #4310, opened 2024-02-16
[16.0][OU-FIX] payment: install account_payment module

remi-filament:
This PR proposes to force install of account_payment in this migration script  since in v16 account_payment is auto_install = ["account"]

Without this, you may end up with v16 database without account_payment installed (if it was not in v15) and you cannot access account moves.

See #4281 

cc @legalsylvain, @pedrobaeza 

Sylvain, if you prefer I can do that in account module instead, but since most of fields for account_payment are moved in payment migration script, it makes sense to me to have it in that one.

### DIFF
--- a/openupgrade_scripts/scripts/payment/16.0.2.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/payment/16.0.2.0/pre-migration.py
@@ -209,6 +209,23 @@ _xmlid_renames = [
 ]
 
 
+def _module_account_payment_to_install(env):
+    """
+    This function force installation of account_payment in case it is not yet installed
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks !

pedrobaeza (approved):

pedrobaeza:
I'm thinking you should put a bit of background as docstring for the method. If not, then with time, we don't remember why these choices, and they can appear to be random.

remi-filament:
Thanks Sylvain, Pedro,
I just added some explanation on function.

pedrobaeza (approved):
Great, thanks.

PR #4312, opened 2024-02-19
[ADD] apriori: rename account_invoice_tier_validation to account_move_tier_validation

kaynnan:
cc @marcelsavegnago @OCA/accounting-maintainers


MiquelRForgeFlow (changes requested):
Firstly, it should be in v13 in any case. Lastly, there is no direct correspondence in the data from one module to another. I mean, one would need migration scripts in account_move_tier_validation (in order to copy all the tier.validation info and tier.definition info from account.invoice to account.move). And I don't see it feasible. Until that is done, I close this PR.

Issue #4313, opened 2024-02-21
[14.0 -> 16.0] SQL views removed in Odoo


remi-filament:
## Module
account

## Describe the bug
After running migration from version 15.0 to 16.0, I tried deleting obsolete columns and tables from 16.0 DB with database_cleanup module and it reports that some cannot be deleted since still in use in account_invoice_report view.

I then checked that account_invoice_report view and it still holds the query from old version

## To Reproduce
**Affected versions**: 16.0 (actually I see that view query from v14 is also not updated when migration to v15)

1. Get a v15 db with account module installed
2. Migrate to v16
3. Check your account_invoice_report view query

**Expected behavior**
View should be recreated by Odoo if code changes, if not, it should be done by OpenUpgrade script.

**Additional context**
```
v16_mig=# \d+ account_invoice_report 
                                 View "public.account_invoice_report"
         Column          |       Type        | Collation | Nullable | Default | Storage  | Description 
-------------------------+-------------------+-----------+----------+---------+----------+-------------
 id                      | integer           |           |          |         | plain    | 
 move_id                 | integer           |           |          |         | plain    | 
 product_id              | integer           |           |          |         | plain    | 
 account_id              | integer           |           |          |         | plain    | 
 analytic_account_id     | integer           |           |          |         | plain    | 
 journal_id              | integer           |           |          |         | plain    | 
 company_id              | integer           |           |          |         | plain    | 
 currency_id             | integer           |           |          |         | plain    | 
 commercial_partner_id   | integer           |           |          |         | plain    | 
 name                    | character varying |           |          |         | extended | 
 state                   | character varying |           |          |         | extended | 
 type                    | character varying |           |          |         | extended | 
 partner_id              | integer           |           |          |         | plain    | 
 invoice_user_id         | integer           |           |          |         | plain    | 
 fiscal_position_id      | integer           |           |          |         | plain    | 
 invoice_payment_state   | character varying |           |          |         | extended | 
 invoice_date            | date              |           |          |         | plain    | 
 invoice_date_due        | date              |           |          |         | plain    | 
 invoice_payment_term_id | integer           |           |          |         | plain    | 
 invoice_partner_bank_id | integer           |           |          |         | plain    | 
 residual                | numeric           |           |          |         | main     | 
 amount_total            | numeric           |           |          |         | main     | 
 product_uom_id          | integer           |           |          |         | plain    | 
 product_categ_id        | integer           |           |          |         | plain    | 
[...]

PR #4315, opened 2024-02-22
[14.0][OU-IMP] *: remove obsolete SQL views

remi-filament:
As per #4313 this PR removes obsolete views from concerned modules


legalsylvain (approved):
Thanks !

rvalyi (approved):

pedrobaeza (approved):

pedrobaeza (changes requested):
Wait a moment: why the `CASCADE` thing? It can have side effects

remi-filament:
I just kept the Odoo default behaviour : https://github.com/odoo/odoo/blob/b7777c50d40cc56e094f43d8f35753d713e17f77/odoo/tools/sql.py#L244

I could remove the cascade although I do not see how it would get side effect except if extra module would be making external link to a view ?

pedrobaeza (approved):
OK then. I was fearing that the FK constraints provokes legit data tables to be removed as well, but I suppose is only on the other way.

PR #4316, opened 2024-02-22
[16.0][OU-IMP] sale: remove obsolete SQL views

remi-filament:
As per https://github.com/OCA/OpenUpgrade/issues/4313 this PR removes obsolete views from sale module


legalsylvain (approved):
Thanks !

rvalyi (approved):

pedrobaeza (approved):

PR #4318, opened 2024-02-23
[16.0][OU-ADD] mail_preview_base + mail_preview_audio: Merged into mail

victoralmau:
Merged `mail_preview_base` + `mail_preview_audio` to mail.

@Tecnativa TT44204


pedrobaeza:
@etobella can you confirm that `mail_preview_base` is not needed anymore?

etobella (approved):
It was only used for audio (the idea was to open other kind of fields) If audio is managed natively by Odoo we can  probably remove the module

pedrobaeza (approved):

PR #4294, opened 2024-02-02
[ADD] 17.0 analysis

StefanRijnhart:

PR #4322, opened 2024-02-27
[17.0][FIX] define openupgradelib git url in test-requirements.txt

legalsylvain:
Ref : https://github.com/OCA/OpenUpgrade/pull/4294

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -19,7 +19,7 @@ org_slug: OCA
 rebel_module_groups: []
 repo_description: OpenUpgrade
 repo_name: 'Open Source upgrade path for Odoo/OpenERP '
-repo_slug: openupgrade
+repo_slug: OpenUpgrade
 repo_website: https://github.com/OCA/OpenUpgrade
 use_pyproject_toml: true
 use_ruff: true
--- a/openupgrade_scripts/__manifest__.py
+++ b/openupgrade_scripts/__manifest__.py
@@ -11,10 +11,6 @@
     "license": "AGPL-3",
     "depends": ["base"],
     "images": ["static/description/banner.jpg"],
-    "external_dependencies": {
-        "python": [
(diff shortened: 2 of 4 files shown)


openupgrade_scripts/README.rst
```diff
@@ -16,14 +16,14 @@ Openupgrade Scripts
 .. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
-.. |badge3| image:: https://img.shields.io/badge/github-OCA%2FOpenUpgrade-lightgray.png?logo=github
-    :target: https://github.com/OCA/OpenUpgrade/tree/17.0/openupgrade_scripts
-    :alt: OCA/OpenUpgrade
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fopenupgrade-lightgray.png?logo=github
+    :target: https://github.com/OCA/openupgrade/tree/17.0/openupgrade_scripts
```
pedrobaeza: Why this change in casing? It was correct before

legalsylvain: I didn't do that. Pre-commit did it.
Something should be wrong in readme generation. 

I'll take a look.

pedrobaeza: Maybe the folder where you download this is in lower case?

legalsylvain: > Maybe the folder where you download this is in lower case?

renamed into Camelcase, and doesn't change anything.

pedrobaeza: Here it is the problem: https://github.com/OCA/OpenUpgrade/blob/843d412c7f62e58227249051cb226d7e6788dccc/.copier-answers.yml#L20

PR #4325, opened 2024-02-28
[16.0][OU-ADD] purchase_picking_state: Merged into purchase_stock

victoralmau:
Related to https://github.com/OCA/purchase-workflow/pull/2081

Merged into purchase_stock (`receipt_status` field)

@Tecnativa TT47855

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -101,6 +101,7 @@ merged_modules = {
     # OCA/purchase-workflow
     "product_form_purchase_link": "purchase",
     "purchase_order_line_price_history": "purchase",
+    "purchase_picking_state": "purchase_stock",
     # OCA/sale-promotion
     "coupon_commercial_partner_applicability": "loyalty_partner_applicability",
     "sale_coupon_selection_wizard": "sale_loyalty_order_suggestion",


chienandalu (commented):
Shouldn't there be a migration script renaming the column and mapping the option values to the former ones?

victoralmau:
> Shouldn't there be a migration script renaming the column and mapping the option values to the former ones?

Interesting, but I'm not sure if doing so is the best thing to do (case of all cancelled pickings) https://github.com/OCA/purchase-workflow/pull/2081#issuecomment-1792087353

pedrobaeza:
I think we can plan it this way:

- Add the merge in apriori as it is here.
- Add in the `purchase_stock` migration scripts that if we find the old column from this module, copy it with the name, and transform values. Cancelled status is passed to NULL, as stated in my comment on the PR.
- Still on `purchase_stock`, if we don't find the column, we pre-create it and fill it according the rules of the computed field.

I'm going to speed up `purchase` migration scripts PR for merging it, so that we add the `purchase_stock` ones.

chienandalu (approved):

legalsylvain:
hi @victoralmau. could you rebase this one ?

legalsylvain:
Hi. It looks like the current script define here https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py#L4 does the job. 
So I'm in favor to merge this PR. or do you think something left to do ?

Thanks ! 
CC : @pedrobaeza

pedrobaeza:
This point is missing:

- Add in the `purchase_stock` migration scripts that if we find the old column from this module, copy it with the name, and transform values. Cancelled status is passed to NULL, as stated in my comment on the PR.

legalsylvain:
>  Add in the purchase_stock migration scripts that if we find the old column from this module, copy it with the name, and transform values. Cancelled status is passed to NULL, as stated in my comment on the PR.

Naive question : is it usefull ? I mean, 
- if the purchase_picking_state is not installed, the current pre-migration script will populate the data.
- if the purchase_picking_state is installed, the current pre-migration script will also populate the data. 

Do we save a lot of time, to implement an alternative computation ? 
in the meantime, it adds extra complexity.

pedrobaeza:
Yes, there can be differences with the new computation, and thus some orders to change its state, so I prefer to keep old values (preservation principle). There's barely no performance penalty, as it's a direct UPDATE with mapping.

legalsylvain:
> thus some orders to change its state, so I prefer to keep old values (preservation principle).

Thanks for your answer, Pedro . 
I'm not sure to understand your point of view. If values are differents, two possibilities :
- previous values was wrong. In that case, it should be fixed.
- it's différent concept, in that case, it should be two fields. 

But as a développer, I expect that the new field contains the value according to the new way to compute. 

Receipt_status is readonly, so values should be prédictive.

https://github.com/odoo/odoo/blob/16.0/addons/purchase_stock/models/purchase.py#L35

pedrobaeza:
Imagine you have a filter showing the number of orders received, and if the computation was wrong in the past, now that you migrate, you get different numbers. You will say that the migration is not correct.

But it can be the contrary case: our code is not correct, and we are replacing correct results by the wrong ones.

In any case, we should keep what was computed, not to "recompute" them.

victoralmau:
Taking this up again, it is still not clear to me that anything needs to be changed, changing something would be wrong and not doing it as well.

The problem is in https://github.com/OCA/purchase-workflow/pull/2081#issuecomment-1792087353 

Do we then want to “_convert_” those cases to `False` according to what `purchase_stock` does (as indicated by https://github.com/OCA/purchase-workflow/pull/2081#issuecomment-1798205806)?

Keeping the behavior of `purchase_picking_state` with the cancel _state_ in `purchase_stock` is not possible because it does not exist, moreover, a call to `_compute_receipt_status()` method would change the value.

pedrobaeza:
Yes, you have to map the states in `purchase_picking_state` to the ones in 16 core:

draft > False
cancel > False
not_received > pending
partially_received > partial
done > full

avoiding to call costly ORM compute method and preserving the most old data.

legalsylvain:
I disagree.

> Imagine you have a filter showing the number of orders received, and if the computation was wrong in the past, now that you migrate, you get different numbers. You will say that the migration is not correct.

If data was wrong before migration, that's great if the data are fixed during the migration.

> But it can be the contrary case: our code is not correct, and we are replacing correct results by the wrong ones.

You can't base a strategy on the fact that the migration scripts are wrong (especially for this script, which is fairly simple).

> Yes, there can be differences with the new computation

That's the reason we should not take into account previous column. For me [that script](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py#L4) + this apriori change is enought.

Note : I'm basing my migration process on this current PR. Please, @victoralmau, if you add extra scripts that @pedrobaeza is asking, please ping me, I'll disable it in a custom branch. Thanks  !

pedrobaeza:
Sylvain, the most important thing is performance, as your migration will cost more time to process all the purchase orders, so I insist, this have to be done.

legalsylvain:
> Sylvain, the most important thing is performance, as your migration will cost more time to process all the purchase orders, so I insist, this have to be done.

I'm surprised. I don't think that's the logic behind OpenUpgrade. The priority for me is : 

1) the data should be correct and similar to if we had entered the old data in a V16 system.
2) the migration script should be as fast as possible.

Switching the 2 priorities seems dangerous to me.

pedrobaeza:
No, the priority is that the data should be preserved, which is this case.

pedrobaeza:
And the module `purchase_picking_state` has been always the same way, so there's no possibility of having incorrect data there.

legalsylvain:
> No, the priority is that the data should be preserved, which is this case.

You're right, let me rephrase.  The priority for me is :

1) the **old** data should be preserved. In that case, we can call ``rename_columns(cr, ["old_column", None])``
2) the **new** data should be correct and similar to if we had entered the old data in a V16 system.
3) the migration script should be as fast as possible.


> And the module purchase_picking_state has been always the same way, so there's no possibility of having incorrect data there.

I had understood the complete opposite in this comment : https://github.com/OCA/OpenUpgrade/pull/4325#issuecomment-2405741831 and in this comment : https://github.com/OCA/OpenUpgrade/pull/4325#issuecomment-2404151577

pedrobaeza:
Yes, because we are doing philosophy, not real cases. But I state the same: old data in this exact case should be preserved, even in they were wrong computed, being only informational purpose. Don't take it as a general mood for every field. If some values of a new/converted field affect other business processes, then it could be considered to be recomputed if they are not correct. What I'm never going to do is to recompute the whole recordset for the case of any non correct data, which is also what you want to do here.

legalsylvain:
> What I'm never going to do is to recompute the whole recordset for the case of any non correct data, which is also what you want to do here.

Oh I can see our difference of opinion ! IMO, I dont propose to recompute a data. My point is  : 
- the apriori file says that the purchase_picking_state feature are now implemented in purchase_stock and so, purchase_picking_state is obsolete.
- the field purchase_stock implement a new field, named receipt_status that has to be computed.

For me the data of the previous obsolete module has to be ignored. It is stored in the db, but it's a coincidence. It could just as easily have been a field calculated on the fly.

Anyway, I don't think we'll agree ! and this is turning into a long discussion thread for both of us.

As I said, I would manually deactivate this recovery script, which I find dangerous, if it is developed.

note: i'm curious about the performance gains. @victoralmau if you develop this recovery script, can you make a comparison? ([that script](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py#L4) vs new one).

thanks !

pedrobaeza:
@legalsylvain we can't calculate that, because we don't have a sample DB and it depends a lot on the number of purchase orders and pickings that you have, but I can assure you that it's going to always be faster to perform an SQL update over populating by ORM the recordsets of purchase.order and stock.picking, read their information, and store data back in the database.

legalsylvain:
@pedrobaeza : it's not an ORM computation I propose, it's an SQL one. Please. https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/purchase_stock/16.0.1.2/pre-migration.py#L4

pedrobaeza:
Yes, sorry, I lost in this long road that context. Anyway, a simple update is better than composing with groups and counting. As said, it will depend on the DB, but you can try easily both queries in any of your DBs.

pedrobaeza (approved):
For not having diverged code, let's not perform that mapping, as it's not so important the possible differences both in performance and data, so I decline from my request. Let's close this :smile:

PR #4327, opened 2024-02-28
[17.0][OU-ADD] openupgrade_framework: framework patches

acpMicrocom:


legalsylvain (changes requested):
Hi @acpMicrocom 

Thanks a lot for working on openupgrade V17.
Could you preserve git history, cherry picking previous commit present in V16 ? 
- it respects authors and history
- it makes easier the diff review between 16.0 and 17.0 

see : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#technical-method-to-migrate-a-module-from-160-to-170-branch

Thanks ! 

/ocabot migration openupgrade_framework

acpMicrocom:
Ok Thanks !

legalsylvain (approved):
thanks !

pedrobaeza:
@legalsylvain have you checked that each patch method is still the same? It's for not doing 2 times the same work.

legalsylvain:
> @legalsylvain have you checked that each patch method is still the same? It's for not doing 2 times the same work.

I just check that the function still exists, with the same arguments list.

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_model.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+_module_data_uninstall._original_method = IrModelData._module_data_uninstall
+IrModelData._module_data_uninstall = _module_data_uninstall
+
+
+@api.model
+def _process_end(self, modules):
+    """Don't warn about upgrade conventions from Odoo
+    ('fields should be explicitly removed by an upgrade script')
```
StefanRijnhart, line 70: Warning was removed from Odoo in https://github.com/odoo/odoo/pull/56657
This patch can be removed here.

openupgrade_framework/odoo_patch/odoo/tests/loader.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+
+from odoo.tests import loader
+from odoo.tools import config
+
+
+def _get_tests_modules(mod):
+    """
+    Make odoo load tests from openupgrade_scripts/scripts/$module/$version/tests
+    instead of the standard location
```
StefanRijnhart, line 15: It looks like https://github.com/odoo/odoo/blob/8c62f4a/odoo/tests/loader.py#L38-L57 makes it possible to load tests from upgrade paths, which is also the aim of this patch. Do you agree @hbrunn?

StefanRijnhart (changes requested):

openupgrade_framework/odoo_patch/odoo/tests/loader.py
StefanRijnhart: Played around with this a bit, and I got it to load a test from the `--upgrade-path` (subpath `base/tests/test_migration_test.py`). Because it has a module path in `odoo.upgrade` instead of `odoo.addons` it does not get assigned the expected attributes in https://github.com/odoo/odoo/blob/aa84ccf/odoo/tests/common.py#L222-L227. The attributes therefore have to be assigned manually:

```
from odoo.tests import tagged, TransactionCase


@tagged("at_install", "openupgrade")
class TestMigrationTest(TransactionCase):
    def test_migration_test(self):
        self.assertEqual(1, 0)

TestMigrationTest.test_module = "base"
TestMigrationTest.test_class = "TestMigrationTest"
TestMigrationTest.test_sequence = 0
```

After that, the test runs fine and can be selected with test tags `base:TestMigrationTest` or `openupgrade`.

Personally, I think that is acceptable given that tests are still very much underused and not for the faint of heart at this point.

openupgrade_framework/readme/CREDITS.md
```diff
@@ -0,0 +1,4 @@
+Many developers have contributed to the OpenUpgrade framework in its
+previous incarnation. Their original contributions may no longer needed,
```
rvalyi, line 2: s/may no longer needed/may no longer be needed/

openupgrade_framework/odoo_patch/odoo/tests/loader.py
hbrunn: agreed the patch becomes obsolete in v17. We could provide a base class for running openupgrade tests, but that's for another PR

openupgrade_framework/readme/CREDITS.md
legalsylvain: hi @rvalyi. I think you can use the "propose change" github feature. It is more readable than the regex syntax ! (and can be approved if desired).
thanks !

rvalyi: are you sure it can be used? It adds extra commits noise that many dislike. I don't see the feature used anymore in the OCA. Can you confirm it's easy enough to remove the commit noise vs doing it this way?

hbrunn: I don't use the github feature for committing myself, because indeed it creates extra commits I'd have to squash again. But still I appreciate reviewers using the feature, because that makes it perfectly clear what they mean.

legalsylvain: >  But still I appreciate reviewers using the feature, because that makes it perfectly clear what they mean.

:+1: That is my point. Regex expression is not very easy to read.

> I don't use the github feature for committing myself, because indeed it creates extra commits I'd have to squash again

Matter of taste. Once the change proposal is done you can : 
- merge then squash
- or make changes locally and push

StefanRijnhart:
@iTecan There are no migration scripts yet for OpenUpgrade 17. These are only the supporting framework methods.

StefanRijnhart:
@acpMicrocom The pandoc-3.1.12.1-1-amd64.deb is an artifact of pre-commit initialization. Please remove it from this PR. Can you follow-up the other comments?

MiquelRForgeFlow:
@StefanRijnhart would you like to supersede this PR?

StefanRijnhart:
Superseded by https://github.com/OCA/OpenUpgrade/pull/4427

PR #4331, opened 2024-03-05
[16.0][OU-FIX] mail: missing ir_rule to be updated

remi-filament:
Rule ir_rule_mail_channel_partner_group_user from mail, needs to be force updated (since it is noupdate = True, it will not get updated otherwise).

Without this update, you get an error when trying to open chatter (from top bar when on a different menu than Discuss) when it tries to resolve access rights to mail_channel_member object since public field does not exist anymore on mail_channel model.

### DIFF
--- a/openupgrade_scripts/scripts/mail/16.0.1.10/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/mail/16.0.1.10/noupdate_changes.xml
@@ -13,6 +13,19 @@
                             ('group_public_id', 'in', [g.id for g in user.groups_id])]
             </field>
   </record>
+  <record id="ir_rule_mail_channel_member_group_user" model="ir.rule">
+    <field name="domain_force">[
+                '|',
+                    '&amp;',
+                        ('channel_id.channel_type', '!=', 'channel'),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not able to find the record `ir_rule_mail_channel_member_group_user` in 15.0:

https://github.com/odoo/odoo/blob/15.0/addons/mail/security/mail_security.xml

May it be you have it due to other reasons?

pedrobaeza:
OK, found it. It has been renamed in pre:

https://github.com/OCA/OpenUpgrade/blob/4111d4ff22c72838e16c18107f4d9bbede99af87/openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py#L30

Then the same problem applies to the other renamed rule, isn't it?

remi-filament:
Hi @pedrobaeza the other renamed rules keep the same domain [(1, "=", 1)] so for me there is no issue here, since renaming XML and model is already done in these migration scripts.

(only name also was updated but I am not sure if necessary to update name for both rules ?)

pedrobaeza (approved):
OK then. Thanks for spotting this.

PR #4340, opened 2024-03-08
[15.0][OU-ADD] purchase_requisition_stock: purchase_requisition_grouped_by_procurement compatibility

victoralmau:
Rename `group_id` column from `purchase.requisition` to procurement_group_id.
Related to https://github.com/OCA/purchase-workflow/pull/2196

@Tecnativa TT41760

### DIFF
--- a/openupgrade_scripts/scripts/purchase_requisition_stock/15.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/purchase_requisition_stock/15.0.1.2/pre-migration.py
@@ -1,0 +1,23 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """The module purchase_requisition_grouped_by_procurement previously added
+    group_id column in previous version. Now procurement_group_id field exists."""
+    if openupgrade.is_module_installed(
+        env.cr, "purchase_requisition_grouped_by_procurement"
+    ):
+        openupgrade.rename_fields(
+            env,
+            [
+                (
+                    "purchase.requisition",
+                    "purchase_requisition",
+                    "group_id",
+                    "procurement_group_id",
+                ),
+            ],
+        )


openupgrade_scripts/scripts/purchase_requisition_stock/15.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """The module purchase_requisition_grouped_by_procurement previously added
+    group_id column in previous version. Now procurement_group_id field exists."""
+    if openupgrade.column_exists(env.cr, "purchase_requisition", "group_id"):
```
pedrobaeza: As the module `purchase_requisition_grouped_by_procurement` is not merged, it's better to check if it's installed also for avoiding to rename an old column or other purpose one.

victoralmau: The related PR (https://github.com/OCA/purchase-workflow/pull/2196) is already merged, so I think it is enough to check if the column exists.

pedrobaeza: I mean that the module is not merged with any other module here in OpenUpgrade in apriori file.

victoralmau: The module `purchase_requisition_grouped_by_procurement` does not have to be merged, it should be maintained. The only thing here is that it now uses a column that already exists.

pedrobaeza: That's what I'm saying...

My suggestion:

```suggestion
    if openupgrade.is_module_installed(env.cr, "purchase_requisition_grouped_by_procurement"):
```

MiquelRForgeFlow (approved):

PR #4355, opened 2024-03-16
[16.0][OU-FIX] digest: Load upstream latest template

pedrobaeza:
As the template is constantly changing upstream:

https://github.com/odoo/odoo/commits/16.0/addons/digest/data/digest_data.xml

keeping a copy here and loading it is not practical and can lead to errors, like the one that triggers this PR, having an inheritance in mass_mailing that crashes due to obsolete content.

Thus, the best strategy is to switch to noupdate=0 before updating contents, and then restore it to noupdate.


rvalyi (approved):
smart!

Issue #4356, opened 2024-03-20
[14.0] update all multi-company security rules


remi-filament:
Odoo added yesterday on v14 a big change (in terms of numbers of rules concernés, should not change anything functionnally) by updating all multi-company to improve performances : https://github.com/odoo/odoo/pull/157861
(This will most probably be forward ported to mater versions in the coming days)

Since a lot of these rules are noupdate=true, I guess it should be added to migration scripts (and maybe even d'or v14.0 to v14.0 updates). Because of the number of impacted rules, I am not sure what would be the n'est way-forward ? Maybe the easiest way would be to replace regex on all rules containing company_ids ?

remi-filament:
Your point of view on this one would be highly appreciated @legalsylvain @StefanRijnhart @pedrobaeza 

Thanks !

pedrobaeza:
Yes, we should add them. The best way may be to relaunch an analysis that compares all the noupdate data and rewrite the files.

And for existing DBs, we may do a general replace with an script.

remi-filament:
Thank you @pedrobaeza 
I am wondering if we rewrite noupdate files, is there a chance we could lose updates that were made manually to those files (for instance removing some fields that should not be updated) while writing migration scripts ?

pedrobaeza:
Yes, the diff should then be checked manually to comment again what has been disabled across modules.

remi-filament:
### **v14.0** fixed with #4358 


### **for v15**
I have analyzed changes between v14 commit and v15 commit, and only new rules (which were not present in v14) differ, so from an OpenUpgrade perspective there should be no changes since these should be created during upgrade). The following XML ids are modified in v15 commit with respect to v14 commit : 

- NEW fleet.ir_rule_fleet_report
- NEW hr_timesheet_attendance.hr_timesheet_attendance_report_restricted_company_rule
- NEW hr_work_entry_contract.ir_rule_hr_work_entry_multi_company
- NEW lunch.ir_rule_lunch_order_multi_company
- DEL lunch.ir_rule_lunch_report_multi_company
- NEW resource.resource_resource_multi_company
- NEW resource.resource_calendar_leaves_rule_multi_company
- NEW stock.stock_storage_category_rule

### **for v16**
there are new changes so, I should create PR for the following XML IDs:
- account.account_bank_statement_comp_rule
- account.account_bank_statement_line_comp_rule
- hr_holidays.hr_leave_rule_multicompany
[...]

PR #4358, opened 2024-03-25
[14.0][OU-UPD] rerun analysis for noupdate_changes.xml

remi-filament:
As per #4356 multi-company security rules have changed.
This PR is the outcome of rerunning analysis, then checking manually each file to keep only what is relevant (and not overwriting noupdate_changes.xml which were manually modified while writing migration scripts)


legalsylvain (approved):
Thanks !

openupgrade_scripts/scripts/crm/14.0.1.2/noupdate_changes.xml
```diff
@@ -1,8 +1,12 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
-  <!-- <record id="crm_activity_report_rule_multi_company" model="ir.rule">
+  <record id="crm_activity_report_rule_multi_company" model="ir.rule">
+    <field name="domain_force">[('company_id', 'in', company_ids + [False])]</field>
     <field name="global"/>
```
MiquelRForgeFlow: comment this line

openupgrade_scripts/scripts/product/14.0.1.2/noupdate_changes.xml
```diff
@@ -4,18 +4,23 @@
     <!-- <field name="digits" eval="2"/> -->
   </record>
   <record id="product_comp_rule" model="ir.rule">
-<!--    <field name="global"/>-->
+    <field name="domain_force"> [('company_id', 'in', company_ids + [False])]</field>
+    <field name="global"/>
```
MiquelRForgeFlow: comment global in this file

openupgrade_scripts/scripts/sale/14.0.1.1/noupdate_changes.xml
```diff
@@ -146,10 +148,4 @@
     <field name="name">Personal Invoices</field>
     <field name="domain_force">[('move_type', 'in', ('out_invoice', 'out_refund')), '|', ('invoice_user_id', '=', user.id), ('invoice_user_id', '=', False)]</field>
   </record>
-  <record id="account_invoice_line_rule_see_all" model="ir.rule">
-    <field name="domain_force">[('move_id.move_type', 'in', ('out_invoice', 'out_refund'))]</field>
-  </record>
-  <record id="account_invoice_line_rule_see_personal" model="ir.rule">
-    <field name="domain_force">[('move_id.move_type', 'in', ('out_invoice', 'out_refund')), '|', ('move_id.invoice_user_id', '=', user.id), ('move_id.invoice_user_id', '=', False)]</field>
-  </record>
```
MiquelRForgeFlow: don't remove this due to https://github.com/OCA/OpenUpgrade/commit/17e556d243fd4f7bb8960105ab1c884fbf16c160

remi-filament:
Thank @MiquelRForgeFlow for review, I indeed missed a few points ! All resolved

MiquelRForgeFlow (approved):
LGTM

pedrobaeza:
Thanks for this, Rémi. Remember that `UPD` is not a valid tag. It should be `IMP` or `FIX`. I think that something similar should be done for v15, as there will be some new company rules there. And maybe it's interesting to force the rewriting unconditionally for those coming from a v14 with old rules.

remi-filament:
> And maybe it's interesting to force the rewriting unconditionally for those coming from a v14 with old rules.

I tried asking Odoo to get update script directly in v14 core, but they answered that there is no need, they would get it when upgrading to next version...
I am not sure what would be the best way-forward for existing DB in v14+ : 
1. should we enforce in OpenUpgrade in v15, v16, v17 these noupdate changes for all ir_rules ?
2. should we create a script that could be run directly in existing DB outside of a migration and that could be called by OpenUpgrade (or listed in the doc as prerequisite) when upgrading to v15 / v16 / v17 ?

Solution 1 is probably cleaner, but solution 2 would allow to get performance improvements for existing DB without waiting for migration... (also I am not sure how to make solution 2 since we would need to check for each module if module is installed before updating rules)

pedrobaeza:
FYI, these code snippets serves for updating existing old instances with the new record rules:

v14: https://gist.github.com/pedrobaeza/81047d9d529eae6fc64dff277f68c291
v15: https://gist.github.com/pedrobaeza/4030e735757571632ae457ed6001d6c5
v16: https://gist.github.com/pedrobaeza/c67693945a4e50607670541cee5cfd6c
v17: https://gist.github.com/pedrobaeza/2ce14433272be5a00d89adaab0f5545d

PR #4361, opened 2024-03-27
[16.0][OU-IMP] account, hr_holidays: update multi-company ir_rule

remi-filament:
Update multi-company ir_rule to align with https://github.com/odoo/odoo/pull/158437
Related to #4356


remi-filament:
I have some doubt about this PR, since in the mean time hr_holidays was updated to v1.6 with https://github.com/odoo/odoo/commit/d59e645b2edf167cb7f108fd7db563799b3db2f1
upgrade_analysis module therefore created me a 16.0.1.6 directory in hr_holidays (same applied to project with version 1.3)

I am not sure what should be done here, should I update in 16.0.1.5 still or should I create a new 16.0.1.6 with potentially only this change inside ?

@legalsylvain @pedrobaeza you probably know : How will OpenUpgrade behave if we try to migrate hr_holidays from 15.0.1.5 to 16.0.1.6 ? Will it apply scripts from 16.0.1.5 still or would it be looking only for scripts in a hr_holidays/16.0.1.6 directory ?

pedrobaeza:
I think we should move everything to the latest core available version. Odoo is executing each pre- scripts from older version to newer one, and after update module, all the post- scripts, so it's better to reduce side effects putting everything in the same latest version.

remi-filament:
Thanks @pedrobaeza so I updated hr_holidays version to 1.6 as well in this PR (and made another PR for updating project version)

pedrobaeza (approved):

PR #4370, opened 2024-04-01
[16.0][OU-ADD] delivery: Migration scripts

pedrobaeza:
Completes #4112 

@Tecnativa TT43606


huguesdk (commented):
@pedrobaeza i encountered 2 issues with the numerical ranges:

1.  when a numerical zip code starts with a `0` (which exists in some countries), this information is lost.
2.  odoo [converts all prefixes to capitals](https://github.com/OCA/OCB/blob/16.0/addons/delivery/models/delivery_zip_prefix.py#L20), and the prefixes returned by `numerical_range_to_prefixes()` sometimes contain `r"\d"` expressions, which become `r"\D"` and thus don’t work.

pedrobaeza:
Well, I haven't been in the need of such conversions for now, so feel free to propose improvements/exceptions on current scripts.

PR #4374, opened 2024-04-04
[13.0][OU-IMP] base: vacuum of transient models

MiquelRForgeFlow:
Selection_2936
Selection_2937

### DIFF
--- a/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
+++ b/odoo/addons/base/migrations/13.0.1.3/pre-migration.py
@@ -799,6 +799,7 @@ def migrate(env, version):
         env.cr, apriori.renamed_modules.items())
     openupgrade.update_module_names(
         env.cr, apriori.merged_modules.items(), merge_modules=True)
+    openupgrade.clean_transient_models(env.cr)
     openupgrade.copy_columns(env.cr, column_copies)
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

StefanRijnhart:
For my curiousity, what problem does this solve?

pedrobaeza:
Hi @StefanRijnhart doing this, the same as enterprise migration scripts seems to be doing, we preventively avoid possible issues due to chained data. Imagine for example a many2one with ondelete="restrict" defined on a transient model, and you try to delete the linked record on your migration scripts.

@MiquelRForgeFlow I'm thinking that you should add specific code like this for assuring everything is deleted, as if not, there may be some row not older than 5 minutes:

```python
    for mname in env:
        model = env[mname]
        if model.is_transient():
            try:
                with env._cr.savepoint():
                    model._transient_clean_rows_older_than(0)
                env._cr.commit()
            except Exception as e:
                _logger.warning("Failed to clean transient model %s\n%s", model, str(e))
```

This may be put in `openupgradelib` as a method `clean_transient_models`.

MiquelRForgeFlow:
It's ok now?

pedrobaeza (approved):

StefanRijnhart (approved):

PR #4378, opened 2024-04-05
[16.0][IMP] base: vacuum of transient models (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/OpenUpgrade/pull/4377.


pedrobaeza (approved):

pedrobaeza:
@legalsylvain you may add to the documentation to call to this method when doing the `base` migration script.

legalsylvain:
> @legalsylvain you may add to the documentation to call to this method when doing the base migration script.

could be great. however, if I remember correctly, there is no section in the documentation that mention what to do for the "base" module. (use apriori, etc...)
It is, for the time being, a documentation to contribute, not to initialize. 
I don't say it could not be added, but it's not just a line somewhere. 

Note for the time being the "initialization documentation" is here : https://github.com/OCA/OpenUpgrade/wiki/

PR #4381, opened 2024-04-05
[16.0][OU-ADD] mrp: Migration scripts

pedrobaeza:
@Tecnativa TT43622


diggy128:
Great! Thank you for the effort.

If possible, take a look at the failing test. It's a double equal sign in an SQL query.

pedrobaeza:
Yes, Python deformation...

PR #4385, opened 2024-04-08
[13.0][FIX] account: don't set exclude_from_invoice_tab for type entry

hbrunn:
when we create an entry in v13, this flag isn't set on move lines for type entry, so the migration shouldn't set it either. Otherwise, we get weird side effects when recomputing ie the balance of such entries

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -296,7 +296,11 @@ def migration_invoice_moves(env):
         env.cr, """
         UPDATE account_move_line
         SET exclude_from_invoice_tab = TRUE
-        WHERE old_invoice_line_id IS NULL""",
+        FROM account_move
+        WHERE old_invoice_line_id IS NULL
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't get the need of this. `exclude_from_invoice_tab=True` is the default value for everything except the invoice lines, so when move_type = entry, they should be that value too.

hbrunn:
this one I found on the way to https://github.com/OCA/OpenUpgrade/pull/4386 - we don't exactly *need* it, but being closer to what the target version does is better, right?

For demonstration, I've just created a move on [runboat](http://oca-account-invoicing-13-0-f9bd9aa6e789.runboat.odoo-community.org/web?debug=1#action=301&cids=1&id=10&menu_id=165&model=account.move&view_type=form)

pedrobaeza (approved):
OK, I see that indeed Odoo doesn't care about such flag when the move type is entry:

imagen

hbrunn:
I added the column at the beginning ;-)

pedrobaeza:
Hehe, didn't see it...

PR #4388, opened 2024-04-09
[16.0][OU-FIX] mail: Avoid mail_notification duplicate constraint

### LINKED ISSUE #4277: [16.0] mail: Delete incorrect rows before applying unique constraint
## Module
mail

## Describe the bug
Prior to Odoo 16, there is no unique constraint on the mail_notification table, so if your table contains duplicate `mail_message_id` and `res_partner_id` combo, the upgrade will fail with the following error:

```
bad query: CREATE UNIQUE INDEX IF NOT EXISTS unique_mail_message_id_res_partner_id_if_set
                                              ON mail_notification (mail_message_id, res_partner_id)
                                           WHERE res_partner_id IS NOT NULL
ERROR: could not create unique index "unique_mail_message_id_res_partner_id_if_set"
DETAIL:  Key (mail_message_id, res_partner_id)=(xxx, yyy) is duplicated.
```

## Expected behavior
Add the following to pre-migration script:

```
DELETE FROM mail_message_res_partner_needaction_rel T1
    USING   mail_message_res_partner_needaction_rel T2
WHERE   T1.ctid < T2.ctid 
    AND T1.mail_message_id = T2.mail_message_id 
    AND T1.res_partner_id  = T2.res_partner_id;
```

pedrobaeza:
Prior to Odoo 16, there is no unique constraint on the mail_notification table, so if there's any duplicate, the upgrade will fail with the following message:

```
bad query: CREATE UNIQUE INDEX IF NOT EXISTS unique_mail_message_id_res_partner_id_if_set
                                              ON mail_notification (mail_message_id, res_partner_id)
                                           WHERE res_partner_id IS NOT NULL
ERROR: could not create unique index "unique_mail_message_id_res_partner_id_if_set"
DETAIL:  Key (mail_message_id, res_partner_id)=(xxx, yyy) is duplicated.
```

Let's preventively delete the duplicated entries, as at the end, they are not needed.

Fixes #4277

@Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
+++ b/openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
@@ -37,6 +37,27 @@ _columns_copies = {
 }
 
 
+def _avoid_mail_notification_new_constraint(env):
+    """Prior to Odoo 16, there is no unique constraint on the mail_notification table,
+    so if there's any duplicate, the upgrade will fail.
+
+    Let's preventively delete the duplicated entries, as at the end, they are not
+    needed.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """DELETE FROM mail_notification
+        WHERE id IN (
+            SELECT id FROM (
+                SELECT id, row_number()
+                OVER (
+                    partition BY res_partner_id, mail_message_id ORDER BY id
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks.

PR #4397, opened 2024-04-17
[15.0][OU-IMP] project: merge project_milestone

remytms:
Parts of the project_milestone feature are merged into the `project` module and another part of the features are moved to the `project_task_milestone` module.

This PR is linked to this issue: https://github.com/OCA/project/issues/1267

### DIFF
--- a/openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
@@ -60,9 +60,21 @@ def adapt_project_task_dependency(env):
     )
 
 
+def rename_project_milestone_target_date(env):
+    """If project_milestone is installed then rename column target_date
+    to deadline.
+    """
+    if openupgrade.column_exists(env.cr, "project_milestone", "target_date"):
+        openupgrade.rename_columns(
+            env.cr,
+            {"project_milestone": [("target_date", "deadline")]},
+        )
+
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
```diff
@@ -60,9 +60,22 @@ def adapt_project_task_dependency(env):
...
+def rename_project_milestone_target_date(env):
+    """If project_milestone is installed then rename column target_date
+    to deadline.
+    """
+    if openupgrade.column_exists(env.cr, "project_milestone", "target_date"):
+        openupgrade.rename_columns(
+            env.cr,
+            {"project_milestone": [("target_date", "deadline")]},
+        )
+        # TODO: activate milestone in configuration
```
pedrobaeza: This is done adding to internal user group the inherit.

remytms: Ok, I will look at that. :)

Thank you very much.

remytms: In fact, there is a configuration to activate milestone, but only in 16.0. Here we are migrating to 15.0, and milestone seams to be enabled by default. So there is nothing to do.

pedrobaeza (approved):

PR #4410, opened 2024-04-18
[16.0][OU-FIX] l10n_mx: Delete account.tag safely

pedrobaeza:
If the tags are used, when Odoo removes them on the regular update process, it crashes:

```
INFO prod odoo.addons.base.models.ir_model: Deleting nnn@account.account.tag (l10n_mx.account_tag_899_01)
ERROR prod odoo.sql_db: bad query: DELETE FROM "account_account_tag" WHERE id IN (nnn)
ERROR: update or delete on table "account_account_tag" violates foreign key constraint "account_account_account_tag_account_account_tag_id_fkey" on table "account_account_account_tag"
DETAIL:  Key (id)=(nnn) is still referenced from table "account_account_account_tag".
````

@Tecnativa


pedrobaeza:
Tested on real migration.

PR #4414, opened 2024-04-23
[15.0][OU-IMP] project: merge project_status

remytms:
Merge `project_status` into `project` module. (https://github.com/OCA/project/issues/1269)

### DIFF
--- a/openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
@@ -71,6 +71,53 @@ def rename_project_milestone_target_date(env):
         )
 
 
+def migrate_project_status(env):
+    """Migrate project_status module if installed to
+    project.project.stages.
+    """
+    if openupgrade.table_exists(env.cr, "project_status"):
+        openupgrade.rename_models(env.cr, [("project.status", "project.project.stage")])
+        openupgrade.rename_tables(env.cr, [("project_status", "project_project_stage")])
+        openupgrade.rename_fields(
+            env,
+            [
+                (
+                    "project.project.stage",
+                    "project_project_stage",
+                    "status_sequence",
+                    "sequence",
+                ),
+                (
+                    "project.project",
+                    "project_project",
+                    "project_status",
+                    "stage_id",
+                ),
+            ],
+        )
+        openupgrade.rename_columns(
+            env.cr,
+            {
+                "project_project_stage": [
+                    ("company_id", None),
+                    ("description", None),
+                    ("is_closed", None),
+                ],
+            },
+        )
+        # Delete project_status views
+        xmlids_views = [
+            "project.view_project",
+            "project.edit_project",
+            "project.project_project_view_form_simplified",
+            "project.project_view_kanban",
+            "project.view_project_project_filter",
+        ]
+        openupgrade.delete_records_safely_by_xml_id(
+            env, xmlids_views, delete_childs=True
+        )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.rename_columns(env.cr, _column_renames)
@@ -78,3 +125,4 @@ def migrate(env, version):
     fill_project_project_allow_task_dependencies(env)
     fill_project_project_last_update_status(env)
     rename_project_milestone_target_date(env)
+    migrate_project_status(env)
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -91,6 +91,7 @@ merged_modules = {
     "survey_description": "survey",
     # OCA/project
     "project_mail_chatter": "project",
+    "project_status": "project",
     "project_task_dependency": "project",
     "project_timeline_task_dependency": "project_timeline",
     # OCA/timesheet
--- a/openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/project/15.0.1.2/post-migration.py
@@ -1,4 +1,8 @@
+import logging
+
 from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
 
 
 def _convert_project_task_assigned_users(env):
@@ -70,6 +74,19 @@ def _fill_project_task_display_project_id(env):
     )
 
 
+def _activate_project_stages_conditionally(env):
+    """Activate project stages if module project_status was installed"""
+    if openupgrade.column_exists(
+        env.cr,
+        "project_project_stage",
+        openupgrade.get_legacy_name("description"),
+    ):
+        _logger.info("project_status module was installed so enable project stages")
+        internal_group = env.ref("base.group_user")
+        stage_group = env.ref("project.group_project_stages")
+        internal_group.implied_ids += stage_group
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     _convert_project_task_assigned_users(env)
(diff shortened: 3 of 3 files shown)


openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
```diff
@@ -71,10 +71,56 @@ def rename_project_milestone_target_date(env):
...
+                    ("description", None),
+                    ("is_closed", None),
+                ],
+            },
+        )
+        # We need to delete old project_status views
+        openupgrade.logged_query(
+            env.cr,
+            "DELETE FROM ir_ui_view WHERE model = 'project.project';"
+        )
```
remytms: This part may be a bit overkill. But I need to delete `project_status` views because there is references on old project.status model in it, and this causes the database to fail showing project views.

But views where renamed during updade of the base module, so I didn't find a way to distinguish between views that are part of existing project module and views coming from project_status and that were renamed.

I tried to remove old views by xmlids but this does not work. I think that this is because some xmlids ends up being the same after the renaming (e.g. `project_status.edit_project` and `project.edit_project`).

By using this, I force the update of project module to recreate views of the project.project model.

Any better way of doing it is welcome.

pedrobaeza: Indeed this is not OK. Regular update process (the second pass without OpenUpgrade) must remove them being noupdate="0".

remytms: I tried not to delete `project.project` views, and perform a regular update after migration in 15.0.

I got the following logs:

```
2024-04-23 08:39:02,435 7 INFO brupower_upgrade odoo.addons.base.models.ir_module: ALLOW access to module.button_immediate_upgrade on ['Projet'] to user __system__ #1 via n/a 
2024-04-23 08:39:02,436 7 INFO brupower_upgrade odoo.addons.base.models.ir_module: ALLOW access to module.button_upgrade on ['Projet'] to user __system__ #1 via n/a 
2024-04-23 08:39:02,436 7 INFO brupower_upgrade odoo.addons.base.models.ir_module: ALLOW access to module.update_list on ['Projet'] to user __system__ #1 via n/a 
2024-04-23 08:39:09,805 7 INFO brupower_upgrade odoo.addons.base.models.ir_module: ALLOW access to module.button_install on [] to user __system__ #1 via n/a 
...
2024-04-23 08:39:19,078 7 INFO brupower_upgrade odoo.modules.loading: loading project/views/project_views.xml 
2024-04-23 08:39:19,253 7 WARNING brupower_upgrade odoo.addons.openupgrade_framework.odoo_patch.odoo.addons.base.models.ir_ui_view: Can't render custom view project.project_view_kanban for model project.project. Assuming you are migrating 
between major versions of Odoo. Please review the view contents manually after the migration.
Error: Element '<kanban class="o_kanban_mobile">' cannot be located in parent view 
2024-04-23 08:39:19,286 7 WARNING brupower_upgrade odoo.addons.openupgrade_framework.odoo_patch.odoo.addons.base.models.ir_ui_view: Can't render custom view project.project_view_kanban for model project.project. Assuming you are migrating 
between major versions of Odoo. Please review the view contents manually after the migration.
Error: Element '<kanban class="o_kanban_mobile">' cannot be located in parent view 
2024-04-23 08:39:19,325 7 INFO brupower_upgrade odoo.models.unlink: User #1 deleted ir.actions.act_window.view records with IDs: [336, 337] 
2024-04-23 08:39:19,593 7 WARNING brupower_upgrade odoo.addons.openupgrade_framework.odoo_patch.odoo.addons.base.models.ir_ui_view: Can't render custom view project.project_view_kanban for model project.project. Assuming you are migrating 
between major versions of Odoo. Please review the view contents manually after the migration.
Error: Element '<kanban class="o_kanban_mobile">' cannot be located in parent view 
2024-04-23 08:39:19,598 7 INFO brupower_upgrade odoo.modules.loading: loading project/views/res_partner_views.xml 
...
```

So it seams that I need to delete some views, but may be not all.

pedrobaeza: That's not a regular update, as `openupgrade_framework` is still around. Regular update will remove them.

remytms: Ho, yes indeed.

I've tested to deactivate views instead of deleting it, and it works also.

remytms: To perform a regular update after this migration in 15.0, I need to add a step to my migration process. Without deactivating views, if I go directly from a upgrade in 15.0 with openupgrade to an upgrade in 16.0 with openupgrade then upgrade in 16.0 with openupgrade fails.

pedrobaeza: Well, that shouldn't happen. It will emit the same warnings, but no crash.

remytms: I fixed it by just removing views that were overridden in the `project_status` module. And it works well.
Some standard views of `project` module were deleted in `pre-migration` script and then created fresh new by the update.

remytms: @pedrobaeza What do you think about the fix I proposed? It has been tested on a database for a real migration.

openupgrade_scripts/scripts/project/15.0.1.2/pre-migration.py
```diff
@@ -71,10 +71,58 @@ def rename_project_milestone_target_date(env):
...
+            {
+                "project_project_stage": [
+                    ("company_id", None),
+                    ("description", None),
+                    ("is_closed", None),
+                ],
+            },
+        )
+        # Delete project_status views
+        xmlids_views = [
```
pedrobaeza, line 109: I still don't see this as correct. Views are not a blocker for migration processes, and more deleting the whole inheritance chain. Please remove this part.

enricostano:
Hi @remytms ! How can we help to move this forward? Thanks!

cc/ @juanmaGHub

remytms:
> Hi @remytms ! How can we help to move this forward? Thanks!
> 
> cc/ @juanmaGHub

@enricostano For me what needs to be done is setting up a migration test process like described by Pedro.

- remove lines from this PR that [removes views in pre-migration script](https://github.com/OCA/OpenUpgrade/pull/4414/files#diff-df77a1ab34ddcbe6402091462cbd731c3a79209a3534165cd2437a4c83fcf5a7R108-R118)
- start from a database in 14.0 with project_status installed and some data.
- migrate from 14.0 to 15.0 with openupgrade
- perform an update all in 15.0 without openupgrade
- migrate from 15.0 to 16.0 with openupgrade
- perform an update all in 16.0 without openupgrade

Each step should work without ERRORS and the data of `project_status` must be correct in 16.0.

I've not found the time to perform such a test yet. Feel free to do it and report here.

PicchiSeba:
@remytms @pedrobaeza @enricostano

Hello everyone, I'm in the process of migrating an existing `14.0` installation to `16.0`.

I created a test installation with just `project_status` in order to test this PR.
I also included the suggestions given by Pedro (I just commented out the elimination of the views from the migration script).

After the OpenUpgrade step I also upgraded the other modules as [suggested](https://github.com/OCA/OpenUpgrade/pull/4414#discussion_r1576002137).
The data is correctly migrated but the second upgrade fails because of an error in this incriminated view

```
Error: while parsing /opt/odoo/auto/addons/project/views/project_views.xml:552
Error while validating view:

Element '<kanban class="o_kanban_mobile">' cannot be located in parent view

View error context:
{'file': '/opt/odoo/auto/addons/project/views/project_views.xml',
 'line': 2,
 'name': 'project.project.kanban',
 'view': ir.ui.view(525,),
 'view.model': 'project.project',
 'view.parent': ir.ui.view(495,),
 'xmlid': 'project_view_kanban'}
```

I managed to fix the error by renaming the xmlid of the view to something else. I'm not sure why this fixes the error. The only thing that comes to mind is the fact the views involved have the same xmlid except for the module part.

In our case we might consider actually removing this specific view.

PR #4417, opened 2024-04-26
[16.0][OU-IMP] point_of_sale: Complete migration scripts

pedrobaeza:
@Tecnativa TT45176

### DIFF
--- a/openupgrade_scripts/scripts/point_of_sale/16.0.1.0.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/point_of_sale/16.0.1.0.1/pre-migration.py
@@ -1,0 +1,36 @@
+# SPDX-FileCopyrightText: 2024 Tecnativa - Pedro M. Baeza
+# SPDX-License-Identifier: AGPL-3.0-or-later
+import uuid
+
+from openupgradelib import openupgrade
+
+
+def _assign_pos_order_token(env):
+    openupgrade.add_fields(
+        env,
+        [("access_token", "pos.order", "pos_order", "char", False, "point_of_sale")],
+    )
+    env.cr.execute("SELECT id FROM survey_survey")
+    for row in env.cr.fetchall():
+        env.cr.execute(
+            "UPDATE survey_survey SET access_token = %s WHERE id = %s",
+            (str(uuid.uuid4()), row[0]),
+        )
+
+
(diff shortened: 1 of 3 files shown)


pedrobaeza:
FYI @carmenbianca

openupgrade_scripts/scripts/point_of_sale/16.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    openupgrade.add_fields(
+        env,
+        [("access_token", "pos.order", "pos_order", "char", False, "point_of_sale")],
+    )
+    env.cr.execute("SELECT id FROM survey_survey")
+    for row in env.cr.fetchall():
+        env.cr.execute(
+            "UPDATE survey_survey SET access_token = %s WHERE id = %s",
+            (str(uuid.uuid4()), row[0]),
+        )
```
legalsylvain, line 18: I installed point of sale in a V15 database and a V16 database, and in each case, the request fail ``SELECT id FROM survey_survey;``
ERREUR:  la relation « survey_survey » n'existe pas
LIGNE 1 : SELECT id FROM survey_survey;

That is quite normal as survey.survey is defined in survey module that is not a dependency of point_of_sale module.

1) Could you elaborate why you introduced that code ? 
~2) I can not understand why the CI is green.~

EDIT : The CI is green, because survey is marked as done, so it is installed by the CI.

pedrobaeza: Ouch, copy/paste...

Fixing it. The problem is that the module was not put on the doc (which I fixed in f52ca3f5595f21153846385d367e1b4980888dfd), and thus CI won't warn me, so now I'm reacting and merging all the needed pieces (as this didn't serve without `pos_restaurant` modules).

pedrobaeza: Fixed in https://github.com/OCA/OpenUpgrade/commit/dea8b8a70af3e9b3cc29a5b472d2699b0ba8f212 and merging `pos_sale` for normalizing CI.

carmenbianca:
Thanks @pedrobaeza ! you have an amazingly good eye for this

PR #4423, opened 2024-05-01
[16.0][OU-FIX] fill new mail.template template_fs field

huguesdk:
the `mail.template` `template_fs` field must contain the path of the data file where it is defined (to allow for it to be reset to its default value). fill empty values by searching for the mail templates (by xml id) in the xml files of the module that defines it.


pedrobaeza (approved):
Working great! Thanks for the improvement.

PR #4428, opened 2024-05-21
[17.0][OU] New analysis

royle-vietnam:
It solves the problems found in [#4042](https://github.com/OCA/OpenUpgrade/pull/4294):
- the most constraints still exist but are considered deleted

With latest changes in Odoo.


MiquelRForgeFlow:
> It solves the problems found in https://github.com/OCA/OpenUpgrade/pull/4294:

I think this link is wrong. Could you link correctly to where the problem of constraints is mentioned?

StefanRijnhart:
Thanks, changes are looking good. Do you have a fix on upgrade_analysis to arrive to these results?

pedrobaeza:
Isn't this simply a result of changes upstream in Odoo since the release?

StefanRijnhart:
Ah yes, it must be this one: https://github.com/odoo/odoo/commit/b278241a719264

StefanRijnhart (approved):

PR #4430, opened 2024-05-22
[17.0][OU-ADD] base: migration to 17.0

daiduongnguyen-odoo:


openupgrade_scripts/scripts/base/17.0.1.3/post-migration.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 Viindoo Technology Joint Stock Company (Viindoo)
+# Copyright 2023 Hunki Enterprises BV (https://hunki-enterprises.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+_deleted_xml_records = [
+    "base.icp_mail_bounce_alias",
+    "base.icp_mail_catchall_alias",
+    "base.icp_mail_default_from",
```
xtanuiha: These three pieces of data seem to have been deleted during the pre-migration phase.

daiduongnguyen-odoo: Are you sure ?, normally those xml-id that have noupdate need to be manually deleted using migration script, in lower version (15, 16) i usually do that

xtanuiha: > Are you sure ?, normally those xml-id that have noupdate need to be manually deleted using migration script, in lower version (15, 16) i usually do that


Sorry, it seems I didn’t express myself clearly before; What I meant was that the three _deleted_xml_records and their corresponding deletion methods appear repeatedly in post_migration.py and end_migration.py. I think it’s only necessary to clean them up once in either place.

daiduongnguyen-odoo: oh yeah,sorry i'll update now

xtanuiha (approved):

daiduongnguyen-odoo:
Dah it green before :))

rvalyi:
Hello, I'm creating a new analysis tool called odoo-module-diff, it's explained here: https://github.com/akretion/odoo-module-diff

To review or complete the PR, you might be interested by these key commits of the base module migration to 17.0:
https://github.com/akretion/odoo-module-diff-analysis/tree/main/17.0/base

openupgrade_scripts/scripts/base/17.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,151 @@
...
+                    ELSE NULL
+                END,
+                update_boolean_value = CASE
+                    WHEN tmp.ttype = 'boolean' then 'true'
+                    ELSE NULL
+                END
+        FROM tmp
+        WHERE ias.id = tmp.id
+        """,
+    )
```
remi-filament, line 109: This one is tricky, since ir_act_server has been modified in v17 to update only 1 field per action (for 'object_create' / 'object_write' state).
The above script does not cover the case where you ir_act_server was updating multiple fields in v16.

I have been scratching my head to find proper way-forward for these cases, but I could not find a solution I am satisfied with.

As a matter of fact with default installation (even with demo), there are only cases for state = 'code' so I guess these other types are mostly linked to base automation module ?

I suppose that for 'object_write' state, we could create multiple actions and link them together in a new ir_act_server with state 'multi'.
For 'object_create' it is more tricky, since you can only define a value (static one) now for 1 field only, so previous action would probably need to be converted to 'code' to get it working in v17. There I am not sure what should be done ? Any idea @pedrobaeza @StefanRijnhart @legalsylvain @MiquelRForgeFlow ?

daiduongnguyen-odoo: Well now look back,  i don't understand why i wrote it lol @@@

pedrobaeza:
Superseded by #4582

openupgrade_scripts/scripts/base/17.0.1.3/pre-migration.py
pedrobaeza: Handled in #4582 creating always "multi" action server, and converting lines to children actions.

daiduongnguyen-odoo: Thank you, i really want to contribute more but my boss keep doing nonsense :)

pedrobaeza: No worries, it can be a collaborative work.

PR #4438, opened 2024-05-28
[15.0][FIX] mail: remove `safe` filters from email templates

chienandalu:
We need to remove `safe` filters from email templates placeholders as they will crash in rendering as qweb can't deal with them.

cc @Tecnativa TT49382

please check @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
+++ b/openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
@@ -104,7 +104,9 @@ def _migrate_placeholder_html(string):
     """
     Replace dynamic placeholders in HTML fields:
     Example: 'Your name is ${object.name}' -> 'Your name is <t t-out="object.name"></t>'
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/mail/15.0.1.5/post-migration.py
```diff
@@ -104,7 +104,9 @@ def _migrate_placeholder_html(string):
     """
     Replace dynamic placeholders in HTML fields:
     Example: 'Your name is ${object.name}' -> 'Your name is <t t-out="object.name"></t>'
+    Example:'${object.html_content | safe}' -> '<t t-out="object.html_content"></t>'
     """
+    string = re.sub(r"\s?\|\s?safe\s?", "", string)
```
pedrobaeza: This is not filtering `'${object.html_content|    safe}'` for example.

chienandalu: Try now :)

pedrobaeza (approved):

PR #4448, opened 2024-06-03
[15.0][OU-FIX] hr_expense: Get payment_state directly

pedrobaeza:
Previous approach was too much aggressive, calling a very generic method that does a lot of stuff apart from recomputing payment_state, launching a rewrite of the lines with debit=0 and credit=0, which totally wrecks the accounting.

Now we have extracted the code for computing the payment state inside the migration script and manually written for each of the expenses moves.

@Tecnativa


pedrobaeza:
Tested on real migration.

huguesdk:
this only sets the value of the `payment_state` field, but other fields must be updated too: `always_tax_exigible`, `amount_residual`, `amount_residual_signed`, `amount_untaxed` and `amount_untaxed_signed`. created #4577 to address this.

PR #4449, opened 2024-06-03
[13.0][OU-FIX] pos_sale: do not delete point_of_sale.pos_config_main

remytms:
In the analysis, the record point_of_sale.pos_config_main appears to be deleted. It's true that it is not overridden in the 13.0 version. But this record is not deleted, it still exists as it main definition is in the module point_of_sale.

In 12.0 the field crm_team_id is added to point_of_sale.pos_config_main and in 13.0 this field rely only on the _get_default function defined in the pos.config model in pos_sale.

There is no need to delete point_of_sale.pos_config_main, and there is no need to modify the crm_team_id value of this record. As this may have been modified by the users.


pedrobaeza (approved):

remytms:
Thanks Pedro. :)

PR #4451, opened 2024-06-05
[14.0][OU-ADD] base: web_disable_export_group migration

chienandalu:
Take the users from the web_disable_export_group.group_export_data group to the new base.group_allow_export which covers the same features.

As now Odoo supports disabling data exports, which is the main feature that the module web_disable_export_group provided we want to move the existing users in that module's groups to the new core one.

The module's functionality isn't completly merged into core as it allows to differentiate which type of export users can use.

This might be unnecessary for some module users that would drop the module while others might want to keep it. To make the transition transparent for both cases, we put this migration script here.

cc @Tecnativa TT49468

please check @pedrobaeza 

fyi @etobella

### DIFF
--- a/openupgrade_scripts/scripts/base/14.0.1.3/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/noupdate_changes.xml
@@ -22,9 +22,13 @@
   <record id="ca" model="res.country">
     <field name="state_required">1</field>
   </record>
-  <record id="default_user" model="res.users">
-    <field name="groups_id" eval="[(4, ref('base.group_partner_manager')), (4, ref('base.group_allow_export'))]"/>
-  </record>
+  <!-- Update only if web_disable_export_group wasn't installed. See post-migration-->
+  <!-- <record id="default_user" model="res.users">
+    <field
+            name="groups_id"
+            eval="[(4, ref('base.group_partner_manager')), (4, ref('base.group_allow_export'))]"
+        />
+  </record> -->
   <record id="es" model="res.country">
     <field eval="'%(street)s\n%(street2)s\n%(zip)s %(city)s (%(state_name)s)\n%(country_name)s'" name="address_format"/>
   </record>
--- a/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
+++ b/openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
@@ -143,3 +143,10 @@ def migrate(env, version):
     binary_conversion_for_db_attachments(env)
     # Load noupdate changes
     openupgrade.load_data(env.cr, "base", "14.0.1.3/noupdate_changes.xml")
+    # By default Odoo grants new users export access. If we had web_disable_export_group
+    # installed is very likely that this isn't the desired behavior and that we want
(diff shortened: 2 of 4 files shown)


StefanRijnhart:
Wondering if we should keep the new group out of the default user's group if this module is installed. What do you think?

https://github.com/OCA/OpenUpgrade/blob/da5e89c3cd377f49c6cb846579a2383418599f61/openupgrade_scripts/scripts/base/14.0.1.3/noupdate_changes.xml#L25-L27

pedrobaeza:
cc @chienandalu

chienandalu:
Thanks for the hint @StefanRijnhart I considered it in the script

openupgrade_scripts/scripts/base/14.0.1.3/noupdate_changes.xml
```diff
@@ -22,9 +22,6 @@
   <record id="ca" model="res.country">
     <field name="state_required">1</field>
   </record>
-  <record id="default_user" model="res.users">
```
StefanRijnhart, line 25: Thanks for the update! Please don't remove this entry but `<!--` uncomment it instead. That way it will be respected when the analysis is updated in the future.

StefanRijnhart (approved):
Thanks!

openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
```diff
@@ -143,3 +143,12 @@ def migrate(env, version):
     binary_conversion_for_db_attachments(env)
     # Load noupdate changes
     openupgrade.load_data(env.cr, "base", "14.0.1.3/noupdate_changes.xml")
+    # By default Odoo grants new users export access. If we had web_disable_export_group
+    # installed is very likely that this isn't the desired behavior and that we want
+    # to keep the access to that feature restricted by default.
+    if not env.ref(
+        "web_disable_export_group.group_export_data", raise_if_not_found=False
```
pedrobaeza: AFAIK, this is always going to be False, as you rename it in pre-migration.

StefanRijnhart: Ah, better use https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2542-L2552 then.

StefanRijnhart (approved):

openupgrade_scripts/scripts/base/14.0.1.3/post-migrate.py
```diff
@@ -143,3 +143,10 @@ def migrate(env, version):
     binary_conversion_for_db_attachments(env)
     # Load noupdate changes
     openupgrade.load_data(env.cr, "base", "14.0.1.3/noupdate_changes.xml")
+    # By default Odoo grants new users export access. If we had web_disable_export_group
+    # installed is very likely that this isn't the desired behavior and that we want
+    # to keep the access to that feature restricted by default.
+    if not openupgrade.is_module_installed("web_disable_export_group"):
```
MiquelRForgeFlow:
```suggestion
    if not openupgrade.is_module_installed(env.cr, "web_disable_export_group"):
```

pedrobaeza (approved):

PR #4452, opened 2024-06-05
[13.0][IMP] apriori: edi_oca -> edi

MiquelRForgeFlow:
Was done in https://github.com/OCA/edi/pull/877.

It was commented in https://github.com/OCA/OpenUpgrade/pull/4384:
Selection_3378

### DIFF
--- a/odoo/addons/openupgrade_records/lib/apriori.py
+++ b/odoo/addons/openupgrade_records/lib/apriori.py
@@ -9,6 +9,8 @@ renamed_modules = {
     'payment_ogone': 'payment_ingenico',
     # OCA/delivery-carrier
     'delivery_carrier_label_ups': 'delivery_ups_oca',
+    # OCA/edi
+    'edi_oca': 'edi',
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't know the history about this, so others are more suitable to judge this move. I see it weird, as AFAIK everything was moved to `_oca` suffix.

MiquelRForgeFlow:
Yeah, in v14 was done the rename from edi to edi_oca. But later, when we made the backport to v12 we used v14 as reference, and we didn't change the module name.

pedrobaeza:
Please @hbrunn can you confirm this is OK?

hbrunn (approved):
I think we need this indeed for migrating to v13, and then we'll have to rename it back for migrating to v14:

https://github.com/OCA/edi/tree/12.0/edi_oca
https://github.com/OCA/edi/tree/13.0/edi
https://github.com/OCA/edi/tree/14.0/edi_oca

PR #4455, opened 2024-06-11
[IMP] website: BS4->BS5 performance refinement

chienandalu:
Depends on:

- [x] https://github.com/OCA/openupgradelib/pull/371

Now that openupgradelib checks implicitly if a record value has a class attribute, we can refine the conversion further applying some extra logical filtering. Mainly when we're dealing with large tables.

We want also to log conversion performance to detect where the bottlenecks might be.

As a performance indicator, in a database with a very large `account_move` table this step passed **from ~14 hours to less than a minute** :upside_down_face: 

cc @Tecnativa TT46020


pedrobaeza:
Please rebase

chienandalu:
> Please rebase

done

pedrobaeza (approved):

PR #4458, opened 2024-06-13
[17.0][OU-ADD] analytic: migration to 17.0

daiduongnguyen-odoo:


acpMicrocom:
Hi @duong77476-viindoo ,

First of all, thank you for the work you've done on this pull request. While testing in my environment, I encountered a small issue related to property creation in the `post-migration` file, specifically on line 66.

The problem seemed to be that some properties already existed, causing an error when trying to recreate them. I suggest a solution to check if the property exists before creating it. The fix I applied is as follows:

```python
  if vals_list:
      for vals in vals_list:
          property = env['ir.property'].search([('fields_id', '=', vals['fields_id']),
                                                ('company_id', '=', vals['company_id']),
                                                ('res_id', '=', vals['res_id'])])
          if property:
              property.write(vals)
          else:
              env["ir.property"].create(vals) 
```
This prevents recreating existing properties and only applies changes when necessary.

I'm not sure of the best way to submit this modification, so I wanted to ask if I should create another pull request with these changes or if you'd prefer to integrate them directly into this one.

Thanks again for your work and help!

daiduongnguyen-odoo:
@acpMicrocom Thanks, i will apply your fix soon, i don't have time to test though because we stop migration campaign for now

openupgrade_scripts/scripts/analytic/17.0.1.1/post-migration.py
```diff
@@ -0,0 +1,75 @@
...
+                "fields_id": field_id,
+                "company_id": company_id,
+                "res_id": "account.analytic.plan,%s" % plan_id,
+                "name": "default_applicability",
+                "value": default_applicability,
+                "type": "selection",
+            }
+        )
+    if vals_list:
+        env["ir.property"].create(vals_list)
```
hbrunn, line 65: I suggest to use [`ir.property#_set_multi`](https://github.com/OCA/OCB/blob/17.0/odoo/addons/base/models/ir_property.py#L324) here, this also takes care of @acpMicrocom's problem

hbrunn (changes requested):

openupgrade_scripts/scripts/analytic/17.0.1.1/post-migration.py
daiduongnguyen-odoo: Yeah indeed hehe

hbrunn:
@duong77476-viindoo I don't manage to push commits onto your branch, so I'll start superseding your PRs with new ones

PR #4469, opened 2024-06-13
[17.0][OU-ADD] project: migrate to 17.0

daiduongnguyen-odoo:

PR #4470, opened 2024-06-13
[17.0][OU-ADD] project_todo: migrate to 17.0

daiduongnguyen-odoo:


pedrobaeza:
/ocabot migration project_todo

<dependency>Depends on : 

- project : #4469</dependency>

openupgrade_scripts/scripts/project_todo/17.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,115 @@
...
+        ALTER TABLE project_task ADD COLUMN openupgrade_legacy_note_id INTEGER;
+        INSERT INTO project_task(
+            create_uid, write_uid, create_date, write_date,
+            active, name, description, priority, sequence, state, project_id,
+            display_in_project, company_id, color, openupgrade_legacy_note_id
+        )
+        SELECT create_uid, write_uid, create_date, write_date,
+               open, name, memo, '0', sequence, '01_in_progress', null,
+               true, company_id, color, id
+        FROM note_note
```
carlos-lopez-tecnativa, line 41: Please migrate followers from `note.note` to `project.task`, as this is currently missing.

image

carlos-lopez-tecnativa (changes requested):
Tested and working, just a minor observation.

TT50089

Update: The followers should be handled by [openupgradelib](https://github.com/OCA/openupgradelib/blob/95e59ab331c312ac5bdaa399aaa6ddf18ffe616a/openupgradelib/openupgrade.py#L1231), but for some reason, I noticed data loss in `res.partner`.

@pedrobaeza I mentioned this to you a few days ago—did you have a chance to check it?

openupgrade_scripts/scripts/project_todo/17.0.1.0/pre-migration.py
pedrobaeza: As I'm understanding this, the problem is not with `openupgradelib`, but the fact that is not a renaming, but an INSERT INTO which is not caring about the followers.

carlos-lopez-tecnativa: > As I'm understanding this, the problem is not with `openupgradelib`, but the fact that is not a renaming, but an INSERT INTO which is not caring about the followers.

I’m sorry, please disregard my previous comment. I tested with a database containing demo data, and Odoo deleted the address that was removed from the code (as a private address). But as you said, `openupgradelib` handles the followers [here](https://github.com/OCA/openupgradelib/blob/95e59ab331c312ac5bdaa399aaa6ddf18ffe616a/openupgradelib/openupgrade.py#L1231). And here, it calls `merge_models` https://github.com/OCA/OpenUpgrade/pull/4470/files#diff-1937c9b8d895188c69985aa078d667cde3ed4c8f72ff658c214a66c1307c4b95R111

V16
image

V17 after migration
image

The manual contacts are kept, but the contacts like Deco or Azure are removed. I suspect this is because some demo data was dropped, and this caused the cascade removal.

carlos-lopez-tecnativa:
@ndd-odoo please rebase the branch

openupgrade_scripts/scripts/project_todo/17.0.1.0/pre-migration.py
pedrobaeza: openupgradelib `merge_models` method already cares about the followers move, so nothing more to do here.

pedrobaeza:
Superseded by #4914

PR #4475, opened 2024-06-13
[16.0][IMP] stock: pre-compute quantity_done

chienandalu:
The field stock.move.quantity_done has passed to be stored.

In large databases (millions of records) it takes a good amount of time to compute it by the ORM, so it's worth it to pre-compute its values as much as possible.

In a database with 12M stock_move records we go from 1h15m with pure ORM to 18minutes

cc @Tecnativa TT46020

### DIFF
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
@@ -42,8 +42,25 @@ def _handle_stock_picking_backorder_strategy(env):
         )
 
 
+def _complete_stock_move_quantity_done_with_orm(env):
+    """In pre-migration we left out the moves with lines with different units of
+    measure to be treated by the ORM"""
+    env.cr.execute(
+        """
+        SELECT move_id
+        FROM stock_move_line
+        GROUP BY move_id
+        HAVING COUNT(DISTINCT product_uom_id) > 1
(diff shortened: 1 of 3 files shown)


openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
```diff
@@ -1,6 +1,27 @@
...
+        """
+        SELECT
+            move_id
+        FROM
+            stock_move_line
+        GROUP BY
+            move_id
+        HAVING
+            COUNT(DISTINCT product_uom_id) > 1 AND
+            SUM(qty_done) <> 0
```
pedrobaeza: Please avoid this new SQL form that requires a lot of lines, and more for queries that are even logged (the same for the rest):

```suggestion
        SELECT move_id
        FROM stock_move_line
        GROUP BY move_id
        HAVING COUNT(DISTINCT product_uom_id) > 1
            AND SUM(qty_done) <> 0
```

legalsylvain:
Hi @chienandalu. What is the state of this PR ? could you apply  @pedrobaeza proposal to get this one merged ? 
thanks !

legalsylvain (approved):

chienandalu:
I had the other query to reformat to a more compact version. Now it should be ok

pedrobaeza (approved):

PR #4476, opened 2024-06-13
[RFC][IMP] POC: async execution

chienandalu:
Maybe this makes no sense :sweat_smile: but here's my rationale

Normally we're interested in executing the migration tasks in order, but there are some cases where that isn't that important. If one of those cases is blocking the next tasks for a long time, wouldn't it be great to just leave it running until it ends?

In this example we've got one of those cases. In v15 the field `narration` in `account.move` goes from a `Text` field to an `Html`. Nothing is going to change in the info of the record values aside from wrapping it in html tags. In the other hand, in a large table (millions of records) it can take a good amount of time to finish this task.

cc @Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/account/15.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/15.0.1.2/pre-migration.py
@@ -1,3 +1,5 @@
+import asyncio
+
 from openupgradelib import openupgrade
 
 from odoo.tools import sql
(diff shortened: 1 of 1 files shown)


legalsylvain (commented):
Hi. Interesting to introduce async / await python features.

However, I don't see await. I guess that the post-migration will be executed, even if _convert_field_to_html is still running. 
My question is : what occures when the upgrade is finished, but the _convert_field_to_html is still running ?

chienandalu:
Yes, this is lacking the `await` indeed :sweat_smile:  The way it is right now, it's limited to the scope of the script itself. Maybe a global approach could be made but I didn't research further  :)

For this case, another way could be to be able to chunk into multiple tasks the `convert_field_to_html` update query, as every update is idempotent

legalsylvain:
> The way it is right now, it's limited to the scope of the  script itself

Not sure to understand. Do you mean the execution of the pre-migration will "await" at the end of the ``def migrate(env, version):`` ?

chienandalu:
Yes, that's the idea. Although the scope would be limited we could improve the final spent time balance with that parallelization

legalsylvain:
@chienandalu. Could you provide benchmark of time consumption (with / without parallelization) ?

thanks !

PR #4477, opened 2024-06-13
[16.0][FIX] base: keep translations for null src while upgrade

chienandalu:
The method ``update_translatable_fields`` is inspired in the core ``_get_translation_upgrade_queries`` method, which had a bug that has been recently addressed in https://github.com/odoo/odoo/pull/168038

Also, we're avoiding removing any original ir_translation record so we can address any further issue easly without needing to redump the table from the former DB.

Let's use that core method with openupgradelib so we can be up to date with any further bugfix.

Depends on

- [x] https://github.com/OCA/openupgradelib/pull/374

cc @Tecnativa TT49615

please review @pedrobaeza 

fyi @hbrunn

### DIFF
--- a/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
@@ -86,6 +86,7 @@ def update_translatable_fields(cr):
             continue
         # borrowed from odoo/tools/translate.py#_get_translation_upgrade_queries
         translation_name = "%s,%s" % (model, field)
+        emtpy_src = """'{"en_US": ""}'::jsonb"""
         openupgrade.logged_query(
             cr,
             f"""
@@ -98,8 +99,10 @@ def update_translatable_fields(cr):
                 GROUP BY it.res_id
             )
             UPDATE {table} m
-            SET "{field}" = CASE WHEN t.noupdate IS FALSE THEN t.value || m."{field}"
-                                 ELSE m."{field}" || t.value END
+            SET "{field}" = CASE
+                WHEN m."{field}" IS NULL THEN {emtpy_src} || t.value
+                WHEN t.noupdate IS FALSE THEN t.value || m."{field}"
+                ELSE m."{field}" || t.value END
             FROM t
             WHERE t.res_id = m.id
             """,
@@ -107,11 +110,6 @@ def update_translatable_fields(cr):
                 "model": model,
                 "name": translation_name,
             },
-        )
-        openupgrade.logged_query(
-            cr,
-            "DELETE FROM ir_translation WHERE type = 'model' AND name = %s",
-            [translation_name],
         )
 
 
--- a/openupgrade_scripts/scripts/base/16.0.1.3/end-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/end-migration.py
@@ -1,0 +1,44 @@
+# Copyright 2024 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade, openupgrade_160
+
+from odoo.tools import column_exists
+
+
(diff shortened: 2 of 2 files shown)


openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -98,8 +99,9 @@ def update_translatable_fields(cr):
                 GROUP BY it.res_id
             )
             UPDATE {table} m
-            SET "{field}" = CASE WHEN t.noupdate IS FALSE THEN t.value || m."{field}"
-                                 ELSE m."{field}" || t.value END
+               SET "{field}" = CASE WHEN m."{field}" IS NULL THEN {emtpy_src} || t.value
```
pedrobaeza: It seems to have more indentation than needed.

pedrobaeza (commented):
Why this is not using existing Odoo method for building the query? There's https://github.com/OCA/openupgradelib/blob/5646c07e598c411ae77b630e19122efc410415ae/openupgradelib/openupgrade_160.py#L29 to do it in theory.

chienandalu:
Mmm, yes, that's a good question :sweat_smile: I'll check it tomorrow

pedrobaeza:
Did you see why the Odoo core method can't be used?

hbrunn (approved):
we can't use Odoo function directly because it needs a registry, while this function might be called before some field exists there

pedrobaeza:
And why not doing it in post-migration?

chienandalu:
I will make a performance comparison and see

pedrobaeza:
```
 psycopg2.errors.DatatypeMismatch: column "title" is of type jsonb but expression is of type character varying
LINE 4:                 UPDATE utm_campaign SET title=name;
                                                      ^
HINT:  You will need to rewrite or cast the expression.
```

chienandalu:
Yes, looks like it will be needed to fix some scripts

pedrobaeza:
Why do it at end-migration? That way, if the translatable fields are involved in any computed operation, this will fail.

chienandalu:
Yes I see. The issue here is that only the `base` models are populated in the environment `env` at the moment of post-migrate... so a lower level approach will be needed...

pedrobaeza:
Yeah, that's true... so the core method is requiring that the model is loaded? It seems a bit restricting... Then we should turn back to the pure SQL approach that it was before. I think we have the answer to why this was done this way originally. Please document it for not going again this way in the future.

chienandalu:
I think we can work with just the field, that we can access in post but adapt openupgradelib to avoid reliying on env

pedrobaeza:
Seeing the core method, it relies on the registry model to get the table, so that's a non-go one...

chienandalu:
Ok, but this would leave behind html translatable fields, which weren't covered by OU: https://github.com/odoo/odoo/blob/3f9eab47809da86eacf0e5a89914aea7531ab44f/odoo/tools/translate.py#L1593

My propose would be:

- Let's do in `pre` an initial pass
- And then in `end` do an additional pass on html translatable records to let them right

pedrobaeza:
I still think that this is not correct, as if any computed field requires a translatable field, it will do the computation with empty strings.

chienandalu:
Only translatable fields with a callable (html_translate and xml_translate) are left for end-migration. Anyway, are you thinking in some case in particular?

In the other hand, Odoo themselves must use this in end-migration if they rely in the registry...

pedrobaeza (approved):
OK, I see

@hbrunn can you give your final blessing?

hbrunn:
looks good to me. What do you mean in your [comment](https://github.com/OCA/OpenUpgrade/pull/4477#issuecomment-2179047145)?

pedrobaeza:
Nevermind, they were just wrong thoughts: first of all, a computed stored field depending on translations is incorrect by definition. Second, the end-migration script is only for callable translations (`translate=html_translate`), so they don't have also computed fields depending on them. Merging then.

PR #4457, opened 2024-06-13
[17.0][OU-ADD] account: migration to 17.0

daiduongnguyen-odoo:

PR #4488, opened 2024-06-26
[17.0][OU-ADD] account: migration to 17.0

flachica:
I make a [PR](https://github.com/duong77476-viindoo/OpenUpgrade/pull/2) in order to add a fix in [account](https://github.com/OCA/OpenUpgrade/pull/4457) migrating module

Open this PR in order to facilitate the work of others colaborators that need the account module be migrated


daiduongnguyen-odoo:
Thanks but why dont u suggest change, i'll fix :v

pedrobaeza:
Yes, let's honor the first proposal.

flachica:
> Yes, let's honor the first proposal.

In the description of the PR I already warned that I had made a PR suggestion to the original repository and that the reason for opening this one was to facilitate the migration to whoever might need it. At no time I tried to take the authorship of this work or to take the honors of the proposal or disrespect or honor the original author who has done so much good work.

To avoid future confusion, is there any guide to collaborate? I could not find it

pedrobaeza:
You should wait a reasonable amount of time to see if the original contributor answers (or in this case, merges the PR). If not, then you can supersede the work for not being stuck. In this case, you have made the PR in OCA at the same time you propose the other PR (and you can also propose in a code review instead of doing a full PR).

PR #4506, opened 2024-07-09
[16.0][MIG] sale_loyalty_delivery: Migration to version 16.0

pilarvargas-tecnativa:
cc @Tecnativa TT44345

### DIFF
--- a/openupgrade_scripts/scripts/sale_loyalty_delivery/16.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/sale_loyalty_delivery/16.0.1.0/pre-migration.py
@@ -1,0 +1,38 @@
+# Copyright 2024 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+_column_copy = {
+    "loyalty_reward": [
+        ("reward_type", None, None),
+    ],
+}
+
+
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/sale_loyalty_delivery/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,38 @@
...
+        ("reward_type", None, None),
+    ],
+}
+
+
+def check_for_free_shipping_promotions(env):
+    """
+    Since the values of the selection reward_type field only change in the
+    sale_loyalty_delivery module, a copy of the column will only be made in case there
+    is a free delivery promotion. In case of such a promotion, a copy of the column will
```
chienandalu, line 16: Shouldn't we copy the column in the `coupon` `pre-mig`? Isn't it very likely that at this point we won't have this info?

pilarvargas-tecnativa: This field comes as it is from v15, only the selection value is modified in this module so there is no loss of previous information :)

chienandalu (approved):
Ok :+1:

pedrobaeza:
/ocabot migration sale_loyalty_delivery

<dependency>Depends on : 

- loyalty_delivery : TODO</dependency>

pedrobaeza (approved):

PR #4512, opened 2024-07-13
[15.0][OU-FIX] website: Escape information renered into website templates

### LINKED ISSUE #4511: Unescaped parameters in website 15.0 post-migration script
## Module
website

## Describe the bug
In website module's 15.0 migration scripts, more specifically in the post-migration.py we execute the following functions:
* extract_footer_copyright_company_name
* update_contact_form_company_description

In both cases, the company name is formatted into XML that is then parsed and saved into the arch_db parameter. Example:

```python
f"Copyright © {website.company_id.name}</span>",
```

```python
company_info_html = common_html_block % (
            company.name,
            ...
)
...
new_arch = re.sub(
            company_description_pattern,
            lambda match: company_info_html,
            website_contactus_arch,
            flags=re.DOTALL,
        )
```

In both cases if the company name or other company information contains a special character like `&` or `<>` the migration fails due to XML parsing error:

```
2024-07-11 18:54:53,015 54944 ERROR db-15 OpenUpgrade: website: error in migration script /var/opt/odoo/OpenUpgrade/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py: EntityRef: expecting ';', line 4, column 74 (<string>, line 4)
```

## To Reproduce
**Affected versions**:
15.0 website post-migration script

1. Try to set the company name with an HTML/XML char that must be escaped.
2. Execute the migration from 14.0 to 15.0 with that database.

**Expected behavior**
We should pass the parameters through `Markup.escape` for the migration to be successful.

**Additional context**
-

PauMAVA:
Fixes #4511 .

### DIFF
--- a/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/15.0.1.0/post-migration.py
@@ -38,7 +38,7 @@ def extract_footer_copyright_company_name(env):
                 website_layout_matches[0]
                 if website_layout_matches
                 else f'<span class="o_footer_copyright_name mr-2">'
-                f"Copyright © {website.company_id.name}</span>",
+                f"Copyright © {Markup.escape(website.company_id.name)}</span>",
                 main_copyright_arch,
             )
             main_copyright_view.with_context(website_id=website.id).arch = new_arch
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Thanks for the patch

PR #4518, opened 2024-07-16
[16.0][MIG] web_tree_image_tooltip: Merged to web

CarlosRoca13:
The module web_tree_image_tooltip is not necessary anymore because odoo has added the zoom option which, if set to True, adds the tooltip with the image.

Example of how to use the feature:
`<field name="image_512" options="{'zoom': True}" widget="image" string="Image"/>`

zoom

cc @Tecnativa TT49818

ping @pedrobaeza @chienandalu

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -113,6 +113,7 @@ merged_modules = {
     # OCA/web
     "web_drop_target": "web",
     "web_ir_actions_act_view_reload": "web",
+    "web_tree_image_tooltip": "web",
 }
 
 # only used here for upgrade_analysis


chienandalu (approved):
:+1:

pedrobaeza (approved):

legalsylvain:
Are the small images loaded by défaut ?

pedrobaeza:
No, the field you decide to put in the tree is the one that is loaded.

legalsylvain:
FYI, it's one of the feature of the current module web_tree_image_tooltip to avoid useless huge data load.

Maybe we can develop a new lighter module to do the same thing.

pedrobaeza:
Yeah, indeed.

PR #4521, opened 2024-07-16
Update pre-migration.py

jjsamayoa:
UPDATE A QRYS


pedrobaeza:
Please give more details about what you are trying to improve. At first sight, these changes are not OK and put them with PL/pgSQL is not needed and it's not adding anything.

PR #4525, opened 2024-07-17
[15.0][OU-IMP] stock: pre-compute stock.quant inventory_date

chienandalu:
This computed field is part of the new inventory system. As we made we the location next_inventory_date, we'll leave it empty by default.

cc @Tecnativa TT38303

please review @pedrobaeza @sergio-teruel


pedrobaeza:
Please change the work done in the work file and the reasons to do this:

https://github.com/OCA/OpenUpgrade/blob/d72e96898adda3ca52e296cc7742bb93e39f4402/openupgrade_scripts/scripts/stock/15.0.1.1/upgrade_analysis_work.txt#L148

chienandalu:
Done :+1:

pedrobaeza (approved):

PR #4529, opened 2024-07-17
[OU-FIX] stock: We don't need to unset name

etobella:
Actually, people might have changed the name

### DIFF
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/noupdate_changes.xml
@@ -11,7 +11,10 @@
     <field name="model_id" ref="model_stock_route"/>
     <field name="name">stock_route multi-company</field>
   </record>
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/stock/16.0.1.1/noupdate_changes.xml
```diff
@@ -11,7 +11,4 @@
     <field name="model_id" ref="model_stock_route"/>
     <field name="name">stock_route multi-company</field>
   </record>
-  <record id="warehouse0" model="stock.warehouse">
-    <field name="name"/>
-  </record>
```
legalsylvain, line 16: Better to comment. So we know what changed.
Otherwise. Lgtm.

etobella: Sure!

pedrobaeza (approved):

PR #4531, opened 2024-07-22
[16.0][OU-FIX] base: in order to load translation, we should check if the value is empty

etobella:
This is a weird case, but if the translation state is "translated" and the value is empty, on 15, we would not use that value

https://github.com/odoo/odoo/blob/15.0/odoo/fields.py#L1602-L1603

with the current change, we are setting the value as an empty string in the database and it might give some issues.

### DIFF
--- a/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
@@ -95,7 +95,10 @@ def update_translatable_fields(cr):
                     bool_or(imd.noupdate) AS noupdate
                 FROM ir_translation it
                 LEFT JOIN ir_model_data imd ON imd.model = %(model)s AND imd.res_id = it.res_id
-                WHERE it.type = 'model' AND it.name = %(name)s AND it.state = 'translated'
+                WHERE it.type = 'model'
+                    AND it.name = %(name)s
+                    AND it.state = 'translated'
+                    AND COALESCE(it.value, '') != ''
                 GROUP BY it.res_id
             )
             UPDATE {table} m


openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py
```diff
@@ -95,7 +95,11 @@ def update_translatable_fields(cr):
                     bool_or(imd.noupdate) AS noupdate
                 FROM ir_translation it
                 LEFT JOIN ir_model_data imd ON imd.model = %(model)s AND imd.res_id = it.res_id
-                WHERE it.type = 'model' AND it.name = %(name)s AND it.state = 'translated'
+                WHERE it.type = 'model'
+                    AND it.name = %(name)s
+                    AND it.state = 'translated'
+                    AND it.value IS NOT NULL
+                    AND length(it.value) > 0
```
pedrobaeza: Shouldn't this be this way?

```suggestion
                    AND it.value != ""
```

zamberjo: Hi, you can simplify the 2 conditions (not null + != '') with:
```suggestion
                    AND COALESCE(it.value, '') != ''
```

pedrobaeza: Yes, this way can be more compact. In terms of performance, I think it's more or less the same. @etobella do you change it?

pedrobaeza: @etobella do you perform this last change (joining the 2 criteria)?

etobella: Sure! Changes applied.

pedrobaeza (approved):

PR #4534, opened 2024-07-28
[FIX] account: reversed_entry_id migration was reversed

### LINKED ISSUE #4376: [13.0] wrong rename for account.move.reverse_entry_id
I don't know if it's a bug or I misunderstand something

In 12.0 `reverse_entry_id` is set in the **original move** and point to the **reverse move**
https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_move.py#L469

In 13.0 `reversed_entry_id` is set in the **reverse move** and point to the **original move**
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L2247

But in the pre-migration we rename  `reverse_entry_id` to `reversed_entry_id` and so point in the wrong direction
https://github.com/OCA/OpenUpgrade/blob/13.0/addons/account/migrations/13.0.1.1/pre-migration.py#L37

sbidoul:
fixes #4376

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -581,14 +581,24 @@ def migration_voucher_moves(env):
 
 
 def fill_account_move_reversed_entry_id(env):
+    # copy refund_invoice_id to reverse_entry_id
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_move am
+        SET reverse_entry_id = am2.id
+        FROM account_invoice ai
+        JOIN account_invoice ai2 ON ai.refund_invoice_id = ai2.id
+        JOIN account_move am2 ON am2.old_invoice_id = ai2.id
+        WHERE am.reverse_entry_id IS NULL AND am.old_invoice_id = ai.id"""
+    )
+    # copy reverse_entry_id to reversed_entry_id (the relation is reversed in 13.0)
     openupgrade.logged_query(
         env.cr, """
         UPDATE account_move am
         SET reversed_entry_id = am2.id
-        FROM account_invoice ai
-        JOIN account_invoice ai2 ON ai.refund_invoice_id = ai2.id
-        JOIN account_move am2 ON am2.old_invoice_id = ai2.id
-        WHERE am.reversed_entry_id IS NULL AND am.old_invoice_id = ai.id"""
+        FROM account_move am2
+        WHERE am.reversed_entry_id IS NULL AND am2.reverse_entry_id = am.id
+        """
     )
 
 
--- a/addons/account/migrations/13.0.1.1/openupgrade_analysis_work.txt
+++ b/addons/account/migrations/13.0.1.1/openupgrade_analysis_work.txt
@@ -331,8 +331,9 @@ account      / account.move             / matched_percentage (float)    : DEL
 account      / account.move             / reverse_entry_id (many2one)   : DEL relation: account.move
 account      / account.move             / reversed_entry_id (many2one)  : NEW relation: account.move
 account      / account.invoice          / refund_invoice_id (many2one)  : DEL relation: account.invoice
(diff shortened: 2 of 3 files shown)


addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -585,10 +585,9 @@ def fill_account_move_reversed_entry_id(env):
         env.cr, """
         UPDATE account_move am
         SET reversed_entry_id = am2.id
-        FROM account_invoice ai
-        JOIN account_invoice ai2 ON ai.refund_invoice_id = ai2.id
-        JOIN account_move am2 ON am2.old_invoice_id = ai2.id
-        WHERE am.reversed_entry_id IS NULL AND am.old_invoice_id = ai.id"""
+        FROM account_move am2
+        WHERE am.reversed_entry_id IS NULL AND am2.reverse_entry_id = am.id
```
pedrobaeza, line 600: `reverse_entry_id` is not filled in this case AFAIK, so you should go to invoice table for this information.

sbidoul: @pedrobaeza I don't understand your remark. in 13.0, `reverse_entry_id` became `reversed_entry_id` with a reversed relation. This is what this update does. Since reverse_entry_id did exist in 12, we can assume it was populated no?

Regarding the invoices, the update before this PR did not work, and you suggested in https://github.com/OCA/OpenUpgrade/issues/4376#issuecomment-2053617507 that there was nothing to do.

In any case I can confirm this PR works for me and correctly set the Reversal Entry field.

pedrobaeza: But in v12, invoices were another thing, and this is filling the reverse link between invoices through the `refund_invoice_id` field. I don't think the `reverse_entry_id` was filled for the journal entries generated from the invoices that are refunded. Or am I wrong? Maybe both operations should be done.

sbidoul: Ok, I updated to preserve the existing update but changing `reverse_entry_id` instead, then let the second update reverse the relation.

I can't test this as the database I'm working with does not have refund_invoice_id for some reason.

rvalyi: Hello just in case this might help you to inspect or remember what Odoo SA did for v13 commit by commit (with their explanations): https://github.com/akretion/odoo-module-diff-analysis/tree/main/13.0/account

sbidoul: Thanks Raphael. Well in this case, the relevant commit is imp-ref-accounting-pocalypse-yeaaahh, so... yeah. :)

sbidoul: But your analysis tool is definitely super useful!

rvalyi: Once I'll have cleaned up it a bit more I'll propose to integrate these key change commits along the OpenUgrade analysis files. Then it will be up to the OpenUpgrade leaders to accept it inside the repo or not...

addons/account/migrations/13.0.1.1/post-migration.py
```diff
@@ -581,14 +581,24 @@ def migration_voucher_moves(env):
...
+    )
+    # copy reverse_entry_id to reversed_entry_id (the relation is reversed in 13.0)
+    openupgrade.logged_query(
+        env.cr, """
+        UPDATE account_move am
+        SET reversed_entry_id = am2.id
+        FROM account_move am2
+        WHERE am.reversed_entry_id IS NULL AND am2.reverse_entry_id = am.id
+        """
     )
```
MiquelRForgeFlow, line 602:
```suggestion
        """
    )
    openupgrade.lift_constraints(env.cr, "account_move", "reverse_entry_id")
```
this is needed. Or better if you add it at pre-migration, as you prefer.

PR #4538, opened 2024-08-09
[15.0][OU-FIX] account: payment_method_line

remytms:
Not all payment_method should generate payment_method_line, but SEPA method from module account_banking_sepa_credit_transfer should generate one.

See comments in https://github.com/OCA/OpenUpgrade/pull/4537


pedrobaeza:
I think this should be done on the OCA module itself, to keep the encapsulation

legalsylvain:
@remytms  can we close this one ?

MiquelRForgeFlow:
Done in https://github.com/OCA/bank-payment/pull/1388.

PR #4545, opened 2024-09-09
[16.0][FIX] payment: is_published state

chienandalu:
This new fields should be set on for payment providers which were in state enabled before the migration. Otherwise they're not visible on the portal or the website.

cc @Tecnativa TT50766

please check @victoralmau @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
@@ -65,7 +65,11 @@ payment      / payment.provider         / code (selection)              : NEW re
 payment      / payment.provider         / allow_express_checkout (boolean): NEW
 payment      / payment.provider         / available_country_ids (many2many): NEW relation: res.country
 payment      / payment.provider         / express_checkout_form_view_id (many2one): NEW relation: ir.ui.view
+# NOTHING TO DO: new features
+
 payment      / payment.provider         / is_published (boolean)        : NEW
+# DONE: Set it on enabled providers to keep the former behavior.
+
 payment      / payment.provider         / maximum_amount (float)        : NEW
 payment      / payment.provider         / token_inline_form_view_id (many2one): NEW relation: ir.ui.view
 # NOTHING TO DO: new features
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -65,7 +65,11 @@ payment      / payment.provider         / code (selection)              : NEW re
 payment      / payment.provider         / allow_express_checkout (boolean): NEW
 payment      / payment.provider         / available_country_ids (many2many): NEW relation: res.country
 payment      / payment.provider         / express_checkout_form_view_id (many2one): NEW relation: ir.ui.view
+# NOTHING TO DO: new features
+
 payment      / payment.provider         / is_published (boolean)        : NEW
+# Set it on on enabled providers to keep the former behavior.
```
pedrobaeza:
```suggestion
# Set it on enabled providers to keep the former behavior.
```

pedrobaeza (changes requested):
You have to map the publishing state from the previous `state` field:

imagen

chienandalu:
> You have to map the publishing state from the previous state field:

What do mean? Isn't that what I'm doing already?

pedrobaeza:
Ah, OK, I read it bad.

openupgrade_scripts/scripts/payment/16.0.2.0/upgrade_analysis_work.txt
```diff
@@ -65,7 +65,11 @@ payment      / payment.provider         / code (selection)              : NEW re
 payment      / payment.provider         / allow_express_checkout (boolean): NEW
 payment      / payment.provider         / available_country_ids (many2many): NEW relation: res.country
 payment      / payment.provider         / express_checkout_form_view_id (many2one): NEW relation: ir.ui.view
+# NOTHING TO DO: new features
+
 payment      / payment.provider         / is_published (boolean)        : NEW
+# Set it on enabled providers to keep the former behavior.
```
pedrobaeza:
```suggestion
# DONE: post-migration: Set it according previous state field.
```

chienandalu: done :+1:

pedrobaeza (approved):

PR #4551, opened 2024-09-12
[16.0][OU-IMP] mail: Compatibility with v15 private channels

victoralmau:
Compatibility with v15 private channels

In v15 you could set a private group that only members could access, now in v16 this is not possible.
Defining the `group_public_id` empty will make everyone can access, and leaving the value that has group_public_id with 'base.group_user' by default is totally wrong.
In order to maintain the compatibility of the existing channels, two transformations are necessary:
- If a channel is private, but had only one auto-subscription group (group_ids), those group become the authorized one (group_public_id) of the channel, and it remains as such as a channel, since at the end, the auto-subscription groups adds as members all the users of those groups, acting in the same way.
- If the list of members was manual, then the channel becomes a 'Direct Messages'
(channel_type='group'), keeping the channel name and authorized members.

@Tecnativa TT50831

### DIFF
--- a/openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
+++ b/openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
@@ -137,6 +137,73 @@ def mail_channel_unset_wrong_group_public_id(env):
     )
 
 
+def mail_channel_private_compatibility(env):
+    """In v15 you could set a private group that only members could access, now in v16
+    this is not possible.
+    Defining the group_public_id empty will make everyone can access, and leaving the
+    value that has group_public_id with 'base.group_user' by default is totally wrong.
+    In order to maintain the compatibility of the existing channels, two transformations
+    are necessary:
+    - If a channel is private, but had only one auto-subscription group (group_ids), those
+    group become the authorized one (group_public_id) of the channel, and it remains as such
+    as a channel, since at the end, the auto-subscription groups adds as members all the
+    users of those groups, acting in the same way.
+    - If the list of members was manual, then the channel becomes a 'Direct Messages'
+    (channel_type='group'), keeping the channel name and authorized members.
+    """
+    env.cr.execute(
+        """
+        SELECT
+            c.id,
+            c.name,
+            (
+                SELECT COUNT(res_groups_id)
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
```diff
@@ -137,6 +137,55 @@ def mail_channel_unset_wrong_group_public_id(env):
...
+    env.cr.execute(
+        """
+        SELECT c.id, c.name, (
+            SELECT res_groups_id
+            FROM mail_channel_res_groups_rel
+            WHERE mail_channel_id = c.id
+            LIMIT 1
+        ) AS auto_subscribe_group_0
+        FROM mail_channel AS c
+        WHERE c.channel_type = 'channel'
```
pedrobaeza: Limit this to only private channels

openupgrade_scripts/scripts/mail/16.0.1.10/pre-migration.py
```diff
@@ -137,6 +137,55 @@ def mail_channel_unset_wrong_group_public_id(env):
...
+        """
+    )
+    for channel_id, _name, auto_subscribe_group_0 in env.cr.fetchall():
+        if auto_subscribe_group_0:
+            openupgrade.logged_query(
+                env.cr,
+                """
+                UPDATE mail_channel
+                SET group_public_id = %s
```
pedrobaeza, line 180: Only one group can be set? If so, this formula doesn't serve.

victoralmau: Yes, only one group can be defined (this has not changed) https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_channel.py#L78

I believe that most often only one group is set, I propose the following:
If the channel has +1 group set, show a log indicating that x group has been set in the channel and to be modified if deemed necessary. 
What do you think?

pedrobaeza: Well, we continue with this strategy if only one group is set. If more than 1, convert it to channel_type = group

pedrobaeza (approved):

PR #4552, opened 2024-09-13
[15.0][ADD] missed default partial account_reconcile_model

dansanti:
By default , must be created 2 reconcile method, in odoo 15.0 was introduced the partial one:

https://github.com/odoo/odoo/blob/ca9db797e37e29030d6f727bd0c0947d09aea203/addons/account/models/chart_template.py#L1063

### DIFF
--- a/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
@@ -74,6 +74,7 @@ def _handle_website_legal_page(env):
             )
             view_temp.unlink()
 
+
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please provide more info about this and greenify CIs.

dansanti:
Hello, in odoo 15.0 was introduced by default partial reconcile. I have to do a PR too, with Total reconcile, introduced between 11.0 and 12.0 , without that and module account_accountant installed ( or some like these), keeps moves open after payment https://github.com/odoo/odoo/blob/ba1ba88874e2ae9716e59ec199eb739d58c3f771/addons/account/data/account_data.xml#L189

PR #4554, opened 2024-09-13
[17.0][OU-ADD] crm: Migration to 17.0

acpMicrocom:


rvalyi (approved):
from my understanding (I apologize if I'm wrong) we should complete the migration by including a noupdate

rvalyi:
> from my understanding (I apologize if I'm wrong) we should complete the migration by including a noupdate

Sorry, I posted this by mistake, the 2 noupdate records are new:

```
NEW ir.rule: crm.mail_plan_rule_group_sale_manager_lead (noupdate)
NEW ir.rule: crm.mail_plan_templates_rule_group_sale_manager_lead (noupdate)
```

So they will be created anyway, right? So in fact I think the PR is OK already, sorry for the noise.

openupgrade_scripts/scripts/crm/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,28 @@
+---Models in module 'crm'---
+---Fields in module 'crm'---
+crm          / crm.lead                 / activity_user_id (many2one)   : not related anymore
+crm          / crm.lead                 / activity_user_id (many2one)   : now a function
+crm          / crm.lead                 / date_action_last (datetime)   : DEL
+crm          / crm.lead                 / date_automation_last (datetime): NEW
```
legalsylvain, line 8: Hi @acpMicrocom. When you see DEL and NEW on a model, there is sometime a renaming behind this point. To know that, you can git blame the line that define the new field.

here, you'll find this commit : https://github.com/odoo/odoo/commit/0a744accc2aaa965d5353e854317895d822ad954#diff-595d3dbbabdc4f766a380a320c1c1a43b143385bc7487c7275e80f76a9fbabc2R165

legalsylvain (changes requested):
Thanks for your contribution. Some remarks inline.

legalsylvain:
/ocabot migration crm

<dependency>Depends on : 

- sales_team : #4685</dependency>

hbrunn (approved):

hbrunn:
as @legalsylvain's comment has been fixed:

/ocabot merge nobump

PR #4555, opened 2024-09-14
[12.0][ADD]Missed default account_reconciliation_model

dansanti:
This is needed for modules like account_accountant, without it , invoices will never get marked as paid, will stod in in_payment


addons/account/migrations/12.0.1.1/noupdate_changes.xml
```diff
@@ -110,4 +110,15 @@
   <record id="digest.digest_digest_default" model="digest.digest">
     <field name="kpi_account_total_revenue">True</field>
   </record>
+  <record id="reconciliation_model_default_rule" model="account.reconcile.model">
```
MiquelRForgeFlow, line 114: add `<!--manually added -->` kind of comment, to know it was manually added

pedrobaeza:
@dansanti any news?

PR #4561, opened 2024-09-19
[17.0][OU-ADD] sale: Migration to 17.0

acpMicrocom:


legalsylvain:
/ocabot migration sale

<dependency>Depends on : 

- sales_team : #4685
- account_payment : #4650
- utm : #4664</dependency>

pedrobaeza:
For each line (or batch of lines) of the analysis file where there's a nothing to do, you have to indicate why is nothing to do. It seems a bit weird that there's nothing to do on the DB schema part.

rvalyi:
for the record, here are the 3 important commits that changed the database for the v17 migration of the sale module:
https://github.com/akretion/odoo-module-diff-analysis/tree/main/17.0%2Fsale

EDIT: Looking quickly it seems there is actually no change requiring migration in the 3 commits. But there might still be XML changes or minor changes missed by my commit analyser and present in the analysis files.

legalsylvain (changes requested):
Indeed, as said by pedro, please analyse the difference between V16 and V17. 

I took a quick look. For exemple, new 'locked' and removal of 'done' state should be handled.

Thanks !

acpMicrocom:
> for the record, here are the 3 important commits that changed the database for the v17 migration of the sale module: https://github.com/akretion/odoo-module-diff-analysis/tree/main/17.0%2Fsale
> 
> EDIT: Looking quickly it seems there is actually no change requiring migration in the 3 commits. But there might still be XML changes or minor changes missed by my commit analyser and present in the analysis files.

Thanks !

legalsylvain:
> for the record, here are the 3 important commits that changed the database for the v17 migration of the sale module:

This analysis looks wrong. Don't you think ? I don't see "locked" changes in your files.

rvalyi:
> > for the record, here are the 3 important commits that changed the database for the v17 migration of the sale module:
> 
> This analysis looks wrong. Don't you think ? I don't see "locked" changes in your files.

Indeed, these changes were not identified by odoo-module-diff and the detected commits are false positives. The commit analyzer is text based and not something structured, so the heuristics are imperfect. I don't try to detect field addition in general, because 99% of the time field additions are new features and not breaking changes. As for selection changes: I didn't try to detect their changes either but eventually that's a good idea to hunt them too https://github.com/akretion/odoo-module-diff/blob/main/odoo_module_diff/main.py#L16  I'll try it and report if it will catch the commit were it changes.

EDIT: hum selections typically spread over several lines and are very sensitive to style formatting, so tracking their changes in plain text will probably give too many false positive and harm the overall relevance. Sadly we will need to rely on the OpenUprade analysis changes for this. Eventually I could use OpenUpgrade analysis files to help hunting the related commits, but that would be quite some work...

being text based, odoo-module-diff is imperfect, but in general it gives a useful complementary vision to OpenUpgrade migration files. So for instance if you look at the analysis for the account module, it ill will give 10 very relevant commits with all the Odoo explanations out of around 1000 commits between v16 and v17: https://github.com/akretion/odoo-module-diff-analysis/tree/main/17.0/account

legalsylvain:
Hi @rvalyi. Thanks for the explanation. I allowed me to put all the odoo-module-diff comment in off-topic, as the current PR is about "sale" migration. Feel free to open dedicated issue if you want to talk about the akretion tool. thanks !

openupgrade_scripts/scripts/sale/17.0.1.2/pre-migration.py
```diff
@@ -13,6 +13,33 @@ def _remove_table_constraints(env):
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order
+        SET state = 'sale'
+        WHERE state = 'done'
+        """,
+    )
+
+
```
legalsylvain: Hi. Thanks for your patch ! 

question: why create a temporary field. I mean, you can simply add locked field, update the value (based on state = "done"), then update state = sale, where state = done). ?

convention: could change the name of the function, for a more explicit one.  "def _sale_sale_order" is not very explicit. something like "def _sale_order_populate_locked_field" looks better. (or something like that.)

thanks !

acpMicrocom:
During the execution of the tests in the pull request for the `sale` module migration script, the following error occurs:
``` 
ERROR: column res_partner.complete_name does not exist
```
This error occurs because the `complete_name` field of the `res_partner` model is not available until the `base` addon is installed or updated. To resolve the problem, it is necessary to make sure that the `base` module is installed and updated before running the tests.

dansanti:
Hello!

I think with this query, we can merge it with website_sale_digital


```
openupgrade.logged_query(
    env.cr,
    """
    UPDATE product_document
    SET attached_on = 'sale_order'
    FROM ir_attachment ir
    WHERE product_document.ir_attachment_id = ir.id
      AND ir.product_downloadable = TRUE
    """
)```

File: openupgrade_scripts/scripts/sale/17.0.1.2/pre-migration.py [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+from openupgradelib import openupgrade
+
+
+def _remove_table_constraints(env):
+    openupgrade.delete_sql_constraint_safely(
+        env, "sale", "res_company", "res_company_check_quotation_validity_days"
+    )
+    openupgrade.delete_sql_constraint_safely(
+        env, "sale", "sale_order", "sale_order_date_order_conditional_required"
+    )
```
hbrunn, line 13: the ORM does this automatically

openupgrade_scripts/scripts/sale/17.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,37 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order
+        SET locked = True, state = 'sale'
+        WHERE state = 'done'
+        """,
+    )
```
hbrunn, line 31: locking should be conditional on the creating user having the `sale.group_auto_done_setting` group.

Better do this in post-migration anyways, then you don't need to create the column manually

hbrunn (changes requested):

openupgrade_scripts/scripts/sale/17.0.1.2/pre-migration.py
alexisadrianc: Hi, @hbrunn 

Thank you for your comment. You can proceed with the changes in this Pull Request, as we currently don’t have any hours allocated for contributions

acpMicrocom: Hi, @hbrunn,

Thank you for your comment. You can proceed with the changes in this Pull Request, as we currently don’t have any hours allocated for contributions

hbrunn:
@acpMicrocom done in https://github.com/OCA/OpenUpgrade/pull/4691

@dansanti I think this belongs into the `website_sale` migration, I left a [comment](https://github.com/OCA/OpenUpgrade/pull/4687#pullrequestreview-2489441991) there mentioning your input

PR #4575, opened 2024-10-01
[ADD] documentation for OpenUpgrade manager. how to create reference database.

legalsylvain:
replace old wiki. see https://github.com/grap/openupgrade_wiki_backup/blob/master/How-to-create-a-reference-database.md

What I did.
- copy paste old wiki page into OpenUpgrade documentation
- split into a ``rst`` part and a ``sh`` part
- add an extra comment to mention that we have to make some retries, if some odoo modules requires extra python lib. see for exemple : asn1crypto remarks.
- I changed shell command, using cat & pipe. it avoid to have to copy paste orm commands.

**blocking point** 
I have a blocking problem regarding force_storage. here is the log of the error, generated by l10n_de module in V17.

```
  File "/home/sylvain/grap_dev/grap-odoo-env-17.0/src/odoo/odoo/addons/base/models/ir_attachment.py", line 81, in force_storage
    ]))._migrate()
  File "/home/sylvain/grap_dev/grap-odoo-env-17.0/src/odoo/odoo/addons/base/models/ir_attachment.py", line 89, in _migrate
    attach.write({'raw': attach.raw, 'mimetype': attach.mimetype})
  File "/home/sylvain/grap_dev/grap-odoo-env-17.0/src/odoo/addons/l10n_de/models/ir_attachment.py", line 31, in write
    self._except_audit_trail()
  File "/home/sylvain/grap_dev/grap-odoo-env-17.0/src/odoo/addons/l10n_de/models/ir_attachment.py", line 27, in _except_audit_trail
    raise UserError(_("You cannot remove parts of the audit trail."))
odoo.exceptions.UserError: You cannot remove parts of the audit trail.

```
-> What is the best way to move forward ? I'd like to avoid to have to write / maintain odoo patches in the database creation process.
-> is this step mandatory ?

@OCA/openupgrade-maintainers : thanks for your help here !

### DIFF
--- a/docsource/100_maintain_repository.rst
+++ b/docsource/100_maintain_repository.rst
@@ -11,6 +11,25 @@ Set up the branch for a new Odoo release
 
 Finally, ``git push`` the branch on your fork, and make a pull request against OCA/$NEW branch.
 
+Create a Reference Database
+---------------------------
+
+To run a test migration, the Github `Test` workflow downloads a preinstalled database of the previous version
+from https://github.com/OCA/OpenUpgrade/releases/tag/databases.
+This database has all Odoo core modules installed with demo data.
+
+* Checkout a clean Odoo up-to-date base code
+* Make sure your addons_path only contains odoo repo, (if it contains OCA / custom repo) the modules will be installed.
+* Odoo version 11.0 may require this patch: https://github.com/odoo/odoo/pull/28620
+* Generate the database
+
+.. literalinclude:: maintainer_scripts/create_reference_database.sh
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
You could consider moving towards an Odoo style backup package that includes the filestore. Alternatively, you can just comment out this code temporarily while you move the filestore into the database: https://github.com/odoo/odoo/blob/17.0/addons/l10n_de/models/ir_attachment.py#L29-L32

legalsylvain:
Thanks !

PR #4577, opened 2024-10-04
[15.0][OU-FIX] fix hr.expense account.move and lines

huguesdk:
fix the value of several fields on `account.move` and `account.move.line` records linked to `hr.expense` records.

this computes `account.move.line.exclude_from_invoice_tab` differently than #4386 did. the goal remains the same: it should be true for destination lines. the code was assuming that destination lines where the ones with a quantity of 0, but this does not work, as odoo 14 [does not set a quantity on those lines](https://github.com/OCA/OCB/blob/14.0/addons/hr_expense/models/hr_expense.py#L500) and [the default value for this field is 1](https://github.com/OCA/OCB/blob/14.0/addons/account/models/account_move.py#L3318). instead, this field is computed by checking that the account is of type `"payable"`.

with `exclude_from_invoice_tab` correctly set, it is now safe to call `account.move._compute_amount()` to correctly recompute several fields (`always_tax_exigible`, `amount_residual`, `amount_residual_signed`, `amount_untaxed`, `amount_untaxed_signed`, `payment_state`) of the `account.move` without changing accounting data. `account.move.write()` checks that reconciled moves are not changed, and with this code, it succeeds.

this code also recomputes these fields for `account.move` records not directly linked to `hr.expense.sheet` records, but linked to an `hr.expense` record through their `account.move.line` records, since this is how `account.move._payment_state_matters()` selects the records.

### DIFF
--- a/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
@@ -14,44 +14,26 @@ def _fill_payment_state(env):
     openupgrade.logged_query(
         env.cr,
         """
-        UPDATE account_move_line
-        SET exclude_from_invoice_tab=(coalesce(quantity, 0) = 0)
-        WHERE expense_id IS NOT NULL
+        UPDATE account_move_line AS aml
+        SET exclude_from_invoice_tab = true
+        FROM account_account AS aa
+        INNER JOIN account_account_type AS aat ON
+            aa.user_type_id = aat.id
+        WHERE
+            aml.account_id = aa.id AND
+            aml.expense_id IS NOT NULL AND
+            aat.type = 'payable'
         """,
     )
-    # Recompute payment_state for the moves associated to the expenses, as on
-    # v14 these ones were not computed being of type `entry`, which changes now
-    # on v15 if the method `_payment_state_matters` returns True, which is the
-    # case for the expense moves
-    for move in env["hr.expense.sheet"].search([]).account_move_id:
-        # Extracted and adapted from _compute_amount() in account.move
-        new_pmt_state = "not_paid" if move.move_type != "entry" else False
-        total_to_pay = total_residual = 0.0
-        for line in move.line_ids:
-            if line.account_id.user_type_id.type in ("receivable", "payable"):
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/hr_expense/15.0.2.0/post-migration.py
```diff
@@ -14,36 +14,36 @@ def _fill_payment_state(env):
...
+        UPDATE account_move_line AS aml
+        SET exclude_from_invoice_tab = true
+        FROM
+            account_account AS aa,
+            account_account_type AS aat
+        WHERE
+            aml.account_id = aa.id AND
+            aa.user_type_id = aat.id AND
+            aml.expense_id IS NOT NULL AND
+            aat.type = 'payable'
```
MiquelRForgeFlow:
```suggestion
        FROM
            account_account AS aa
        JOIN account_account_type AS aat
            ON aa.user_type_id = aat.id
        WHERE
            aml.account_id = aa.id AND
            aml.expense_id IS NOT NULL AND
            aat.type = 'payable'
```
it's nicer with JOIN

huguesdk: @MiquelRForgeFlow thanks for the review! i didn’t know we could use joins there. i added your suggestion, but used `inner join` instead of just `join` (even though it means the same) because i prefer explicit over implicit.

pedrobaeza:
I need why this is needed with an example dataset that fails previously. In general, I don't want to call the general method `_compute_amount`, which can have lots of side effects.

huguesdk:
@pedrobaeza could you tell me in which form this dataset should be presented?

about `_compute_amount()`, since these are (read-only) computed fields, isn’t it supposed to be a model invariant that the fields should be able to be recomputed at any time if any of their dependencies have changed (or not)? if calling this method results in incorrect data, it means that there is a problem elsewhere. i see the fact of being able to call the compute method more as a safety check, actually. also, what would be the point of (possibly incorrectly) recomputing all of these fields with custom code here while the compute method already exists and does it correctly? what do you think?

pedrobaeza:
I mean the records with their field values you have previous the migration and the result after for requiring this recomputation.

PR #4582, opened 2024-10-11
[17.0][OU-ADD] base: Migration scripts

pedrobaeza:
Supersede of #4430, as there are no push rights there, and Viindoo has said that they are not working more on the migration scripts for now.

Changes performed over existing PR:

- Comment some non desirable noupdate=1 changes
- Handle properly ir.actions.server update conversion
- Handle properly the private type removal
- Reformat analysis work file
- Other minor adjustments

@Tecnativa TT49290


pedrobaeza:
OK, let's merge this as a "base" (wordgame, hehe) for continuing doing the rest of the scripts, and with real tests, we can refine it.

remi-filament:
Thanks @pedrobaeza for taking over this base module migration.

From what I can see, the only point that might be problematic during migrations is about ir_act_server with state = 'action_create'. I understand we lose functionalities to set multiple values with new Odoo way of using ir_act_server, maybe we should make it clear somewhere (in the logs or elsewhere ?), I know you point it in upgrade_analysis_work.txt but I am not sure it is visible enough ?
Also you chose to use only field = "name", which should cover most of the cases, also there are probably corner cases where main field of model is not called "name" but should be the one to fill in, yet it should apply to most cases so maybe we could also cover this limitation in documentation somewhere ?

pedrobaeza:
Yes, right now we don't have a centralized place where informing all of this. Odoo enterprise is putting a chat conversation for system administrators. We may mimic that approach, but sometimes the people reading that may not be suitable to act.

And also specifically about actions servers, you shouldn't based your developments on actions servers if you want to be a serious integrator.

remi-filament:
> Yes, right now we don't have a centralized place where informing all of this. Odoo enterprise is putting a chat conversation for system administrators. We may mimic that approach, but sometimes the people reading that may not be suitable to act.

Yes maybe, I have seen it indeed only when updating localisation that they left a message in chat. Maybe all people reading that would not be suitable to act, but in general after a migration I always connect with administrator account to check a few things so I would get the information which seems a good solution !

> And also specifically about actions servers, you shouldn't based your developments on actions servers if you want to be a serious integrator.

:+1:

pedrobaeza:
One of the OpenUpgrade principles is to provide a full migration process, including opinionated decisions when there are several choices, so the number of these entries to check should be minimum.

PR #4583, opened 2024-10-11
[17.0][OU-ADD] web: Nothing to do

remi-filament:


remi-filament:
Hi @pedrobaeza I have a doubt on this one as well, since I have no case where qweb is used in view_mode for ir.actions.act_window.view (I also scanned the full OCA repo and did not find any).

However, I suppose that in order to be more conservative, we could remove any view of type qweb from ir.actions.act_window.view and probably from view_mode string in ir.actions.act_window.
Let me know if you think that we should go that way ?

pedrobaeza (approved):
These kind of views was very weird, and also old, like the project status one in `sale_timesheet` in v13:

https://github.com/odoo/odoo/blob/4cc086d330b73514fbc64a7fcf22d8a7a9f1b691/addons/sale_timesheet/views/project_task_views.xml#L121

If there's a remaining one, it's OK to have an error, as you must replace it properly.

PR #4466, opened 2024-06-13
[17.0][OU-ADD] payment: migrate to 17.0

daiduongnguyen-odoo:

PR #4584, opened 2024-10-11
[17.0][OU-ADD] payment: Migration scripts

remi-filament:
Take over payment migration from PR #4466 upon @duong77476-viindoo request (https://github.com/OCA/OpenUpgrade/pull/4466#discussion_r1785528577)

Extra tasks performed : 

- Remove payment.payment_method_acss_debit from noupdate_changes since this module does not exist anymore
- Rename model payment.icon to payment.method
- Reload payment methods


rvalyi:
for reference, the 5 key commits for v17: https://github.com/akretion/odoo-module-diff-analysis/tree/main/17.0/payment

remi-filament:
Thanks @rvalyi it helps understanding where the changes come from (although in that case many changes were on data so there are more commits to take into account).
I added migration script to transform payment.icon in payment.method, even if I am not sure it was worth the effort since they changed pretty much everything, it would probably have been better to just drop everything and let ORM recreate proper values...

legalsylvain:
/ocabot migration payment

<dependency></dependency>

openupgrade_scripts/scripts/payment/17.0.2.0/post-migration.py
```diff
@@ -0,0 +1,34 @@
...
+        payment_token.payment_method_id = (
+            PaymentMethod._get_from_code(payment_token.provider_id.code)
+            or unknown_payment_method
+        ).id
+
+    for transaction in PaymentTransaction.search([("payment_method_id", "=", False)]):
+        transaction.payment_method_id = (
+            PaymentMethod._get_from_code(transaction.provider_id.code)
+            or unknown_payment_method
+        ).id
```
hbrunn, line 25: can we group those by provider_id and do only one write per provider?

openupgrade_scripts/scripts/payment/17.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,189 @@
...
+    ("payment_icon", "payment_method"),
+]
+
+_field_renames = [
+    (
+        "payment.provider",
+        "payment_provider",
+        "payment_icon_ids",
+        "payment_method_ids",
+    ),
```
hbrunn, line 153: for this to work we'll also have to rename `payment_icon_payment_provider_rel` I think, and in there `payment_icon_id` to `payment_method_id` (I wonder if we should actually add this to rename_fields, seems a common enough task)

hbrunn (commented):
Totally agree with renaming `payment.icon` to `payment.method`, as that will be the base for whatever migrations of modules depending on `payment` do.

Not so sure yet about the full reload of the payment methods, shouldn't we only force load new fields like `primary_payment_method_id`? Or the other way around, not touch `active` and `sequence`?

hbrunn:
@remi-filament I've implemented my suggestions in https://github.com/OCA/OpenUpgrade/pull/4692

openupgrade_scripts/scripts/payment/17.0.2.0/post-migration.py
remi-filament: Thanks for you review @hbrunn 
Sorry I did not come back on this before, I have had a crazy loaded month !
Thanks for fixing this in https://github.com/OCA/OpenUpgrade/pull/4692/commits/4ec754cc2bc785841ec95388a3cf4df403b1901c

openupgrade_scripts/scripts/payment/17.0.2.0/pre-migration.py
remi-filament: You are right of course, however since link between provider and methods completely changed in v17, I thought that maybe it was better to let ORM recreate the full table.
Now I see that you may have extra payment provider (not provided by Odoo) that would need to keep their corresponding methods, so it makes sense to migrate those as you did in https://github.com/OCA/OpenUpgrade/pull/4692/commits/4ec754cc2bc785841ec95388a3cf4df403b1901c

remi-filament:
> Not so sure yet about the full reload of the payment methods, shouldn't we only force load new fields like `primary_payment_method_id`? Or the other way around, not touch `active` and `sequence`?

The problem here is that before you only had 3 fields in data : sequence, name, image (https://github.com/odoo/odoo/blob/16.0/addons/payment/data/payment_icon_data.xml)
Now you have a lot more, and payment method data becomes noupdate=1 (https://github.com/odoo/odoo/blob/17.0/addons/payment/data/payment_method_data.xml).

Note that active and primary_payment_method_id did not exist before in v16.

Therefore I believe there were 2 ways to handle this : 

1. drop all existing payment methods from v16 and let ORM recreate them (and maybe keep only the ones created by modules outside Odoo core ?)
2. try to transform old payment.icon XML-IDs to new payment.method XML-IDs and set force update on these new XML-IDs and update data afterwards (which is what I tried to achieve here, letting Odoo create all the new methods that did not map to icons in v16).

PR #4597, opened 2024-10-21
 [17.0][OU-ADD] analytic: migration to 17.0

hbrunn:
https://github.com/OCA/OpenUpgrade/pull/4458 with review followup


hbrunn:
/ocabot migration analytic

<dependency></dependency>

hbrunn:
reminder: This needs to load the noupdate changes to be introduced in #4606

openupgrade_scripts/scripts/analytic/17.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,26 @@
+---Models in module 'analytic'---
+---Fields in module 'analytic'---
+analytic     / account.analytic.account / message_main_attachment_id (many2one): DEL relation: ir.attachment
+analytic     / account.analytic.account / root_plan_id (many2one)       : not a function anymore
+analytic     / account.analytic.account / root_plan_id (many2one)       : now related
+analytic     / account.analytic.plan    / _order                        : _order is now 'sequence asc, id' ('complete_name asc')
```
pedrobaeza, line 10: I'm checking now about this point, as we should preserve the order shown in previous version, but I'm not finding right now where the widget is taking the order. Investigating now...

pedrobaeza:
Finish migration scripts with these changes:

- Set account.analytic.plan sequence according previous order (by id)
- Don't put default value for account.analytic.plan.applicability~company_id for preserving previous behavior.
- Better work file explanations.

pedrobaeza (approved):

PR #4431, opened 2024-05-22
[17.0][OU-ADD] mail: migration to 17.0

daiduongnguyen-odoo:

PR #4600, opened 2024-10-21
[17.0][OU-ADD] mail: migration to 17.0

hbrunn:
https://github.com/OCA/OpenUpgrade/pull/4431 with review followup


hbrunn:
/ocabot migration mail

<dependency>Depends on : 

- bus : TODO
- web_tour : TODO</dependency>

openupgrade_scripts/scripts/mail/17.0.1.15/pre-migration.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+
+
+def _mail_gateway_allowed(env):
+    """Set some dummy value so that the not null constraint can be created"""
+    env.cr.execute(
+        """
+        UPDATE mail_gateway_allowed SET email='admin@example.com'
+        WHERE email IS NULL"
+        """
```
hbrunn, line 80: @remi-filament I think we should fill this with a dummy value after all, otherwise Odoo won't be able to set up the not null constraint

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,488 @@
...
+# NOTHING TO DO
+
+DEL ir.model.access: mail.access_res_users_settings_all [renamed to base module]
+DEL ir.model.access: mail.access_res_users_settings_user [renamed to base module]
+# NOTHING TO DO: handled in base module
+
+DEL ir.model.access: mail.access_res_users_settings_volumes_all
+DEL ir.model.constraint: mail.constraint_bus_presence_partner_or_guest_exists
+DEL ir.model.constraint: mail.constraint_mail_activity_check_res_id_is_set
+DEL ir.model.constraint: mail.constraint_mail_alias_alias_unique
```
hbrunn, line 294: @remi-filament I don't think we need to remove this manually - that's necessary when we do something during migration that would violate it, and afterwards it's deleted anyways

hbrunn:
to discuss: `message_main_attachment_id` has been moved to its own [mixin](https://github.com/OCA/OCB/blob/17.0/addons/mail/models/mail_thread_main_attachment.py) and most models don't inherit that one. As I assume there's a plethora of private code out there doing $deity knows what with this field, should we loop through all models in end-migration and copy away the column to its legacy name for further processing?

docsource/modules160-170.rst
```diff
@@ -86,7 +86,7 @@ Module coverage 16.0 -> 17.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | base_install_request                              |                      |No DB layout changes.                            |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| base_setup                                        |                      |                                                 |
+| base_setup                                        | Nothing to do        |                                                 |
```
pedrobaeza: I think this should be done in its own PR, not here.

docsource/modules160-170.rst
```diff
@@ -872,7 +872,7 @@ Module coverage 16.0 -> 17.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | |del| web_kanban_gauge                            |                      |Merged into web.                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| web_tour                                          |                      |                                                 |
+| web_tour                                          | Nothing to do        |                                                 |
```
pedrobaeza: I think this should be done in its own PR, not here.

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,488 @@
...
+# DONE: renamed to discuss.channel in pre-migration
+
+obsolete model mail.channel.member
+# DONE: renamed to discuss.channel.member in pre-migration
+
+obsolete model mail.channel.rtc.session
+# DONE: renamed to discuss.channel.rtc.session in pre-migration
+
+model res.users.settings (moved to base)
+new model discuss.channel
```
pedrobaeza: Please group together both the `new model` entry with the `obsolete model` for logical grouping blocks for having a compact analysis work file.

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,488 @@
...
+new model mail.partner.device
+new model mail.tracking.duration.mixin [abstract]
+# NOTHING TO DO: new feature
+
+---Fields in module 'mail'---
+mail         / discuss.channel          / active (boolean)              : NEW hasdefault: default
+# NOTHING TO DO: renamed model
+
+mail         / discuss.channel          / allow_public_upload (boolean) : NEW hasdefault: default
+# DONE pre-migration: set True for channel_type = 'livechat' because customer need to upload some documents
```
pedrobaeza: Why on pre? I think it doesn't matter, and in post you save the column addition.

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,488 @@
...
+
+new model mail.message.translation
+new model mail.notification.web.push
+new model mail.partner.device
+new model mail.tracking.duration.mixin [abstract]
+# NOTHING TO DO: new feature
+
+---Fields in module 'mail'---
+mail         / discuss.channel          / active (boolean)              : NEW hasdefault: default
+# NOTHING TO DO: renamed model
```
pedrobaeza: Grouped all the lines that corresponds to the renaming together, or include the model renamings and provide a new cleaned analysis file.

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,488 @@
...
+mail         / mail.activity.plan       / template_ids (one2many)       : NEW relation: mail.activity.plan.template
+mail         / mail.activity.plan.template / activity_type_id (many2one)   : NEW relation: mail.activity.type, required, hasdefault: default
+mail         / mail.activity.plan.template / note (html)                   : NEW
+mail         / mail.activity.plan.template / plan_id (many2one)            : NEW relation: mail.activity.plan, required
+mail         / mail.activity.plan.template / responsible_id (many2one)     : NEW relation: res.users, hasdefault: compute
+mail         / mail.activity.plan.template / responsible_type (selection)  : NEW required, selection_keys: ['on_demand', 'other'], hasdefault: default
+mail         / mail.activity.plan.template / sequence (integer)            : NEW hasdefault: default
+mail         / mail.activity.plan.template / summary (char)                : NEW hasdefault: compute
+mail         / mail.activity.type       / keep_done (boolean)           : NEW
+# NOTHING TO DO: new feature
```
pedrobaeza: If you want, you can mention that this exists in OCA module for v16 :wink: : https://github.com/OCA/social/tree/16.0/mail_activity_plan

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,488 @@
...
+
+mail         / mail.alias               / alias_full_name (char)        : NEW isfunction: function, stored
+# DONE pre and post migration: create column in pre and fill value in post
```
pedrobaeza, line 137: Why on pre? I think it doesn't matter, and in post you save the column addition.

And you can group all the entries belonging to this.

openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
```diff
@@ -0,0 +1,90 @@
...
+        env["mail.template"],
+        "mail_template",
+        "report_template_ids",
+        "report_template",
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "mail", "17.0.1.15/noupdate_changes.xml")
```
pedrobaeza, line 94: I think 

https://github.com/OCA/OpenUpgrade/blob/8e6c601e90489cbbaddd7b801e2180975e8c6c6c/openupgrade_scripts/scripts/mail/17.0.1.15/noupdate_changes.xml#L10

should be commented

pedrobaeza (commented):
In general, the work file is very hard to be followed, due to not proper logical blocks grouping, and the models renaming that are not taken into account for it.

May you consider to re-run analysis to get a cleaner one and do the groupings (or at least this second one).

pedrobaeza:
> to discuss: `message_main_attachment_id` has been moved to its own [mixin](https://github.com/OCA/OCB/blob/17.0/addons/mail/models/mail_thread_main_attachment.py) and most models don't inherit that one. As I assume there's a plethora of private code out there doing $deity knows what with this field, should we loop through all models in end-migration and copy away the column to its legacy name for further processing?

I don't think we should rename. The goal of OU is the preservation, and this is achieved not doing anything, as the column is preserved with `main_attachment_id` name. In fact, I only rename when there's a risk of collapsing with other thing. If not, I prefer to keep the original column name. This way, any module restoring the feature don't need to re-rename again the column in pre_init_hook.

hbrunn:
@pedrobaeza thanks for your points, I'll pick them up next Monday. Or some member of @OCA/openupgrade-maintainers before, I suspect I couldn't push to the original PRs because they come from a fork of a fork, while my copy is forked directly from here.

pedrobaeza:
I have pushed a rebased branch with my comments applied. I see that the new analysis doesn't contain the model renaming for shortening the analysis file, but I have shorten it myself.

hbrunn:
@pedrobaeza thanks for taking over, but I'm not sure I agree with shortening/reordering the work file. What I do currently to verify all analysis points are covered is `diff upgrade_analysis.txt upgrade_analysis_work.txt`, and if any non-comment line shows up there, something is missing. That's especially crucial now we have changed analysis files.

How do you do that? Compare manually?

pedrobaeza:
I don't need to do anything with the `_work` file. What I do is to check the diff of the regular file in the new analysis PR, and incorporate what needed both in the migration scripts and in the work file.

In your PR https://github.com/OCA/OpenUpgrade/pull/4606/files, I already checked that there's nothing touched in `mail` analysis file.

PR #4601, opened 2024-10-21
[16.0][OU-ADD] l10n_es_irnr/l10n_es_irnr_sii: Merged into l10n_es/l10n_es_aeat_sii_oca

pedrobaeza:
Following https://github.com/OCA/l10n-spain/pull/3783, those migrating from 15 to 16 with these modules don't need them, as they are useless now.

@Tecnativa

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -88,6 +88,9 @@ merged_modules = {
     "account_invoice_report_due_list": "account",
     # OCA/e-commerce
     "website_sale_require_login": "website_sale",
+    # OCA/l10n-spain
+    "l10n_es_irnr": "l10n_es",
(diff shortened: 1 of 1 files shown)


etobella:
@pedrobaeza A migration scripts is missing for this part, as the change of id is not handled :sob:  (at least for l10n_es_irnr to rename everything)

I will check how to do this

pedrobaeza:
Isn't the XML-IDs the same?

etobella:
but changed the modules, also the groups...

pedrobaeza:
OK, this is not v17, so each company noupdate XML-ID created on the fly will point to `l10n_es_irnr`, not `l10n_es`. Go ahead then with the programmatic renaming.

PR #4603, opened 2024-10-21
[17.0][OU-ADD] contacts: Nothing to do

acpMicrocom:
- No file generated by `upgrade_analysis`


legalsylvain:
/ocabot migration contacts

<dependency></dependency>

rvalyi (approved):
the tests probably fail because of the mail module which isn't migrated yet

PR #4625, opened 2024-10-23
[17.0][OU-ADD] gamification: Migration to 17.0

acpMicrocom:


legalsylvain:
/ocabot migration gamification

<dependency></dependency>

rvalyi (commented):
Not sure it is complete.

I believe the `ir_cron_consolidate_last_month` xml reference of this v16 cron https://github.com/odoo/odoo/blob/16.0/addons/gamification/data/ir_cron_data.xml#L15 should be renamed to ir_cron_consolidate in the migration, see https://github.com/odoo/odoo/blob/17.0/addons/gamification/data/ir_cron_data.xml#L15 .

I'm also not sure, but may be the karma computation should be triggered, See the change here https://github.com/akretion/odoo-module-diff-analysis/blob/main/17.0/gamification/c000__________-%23_76430_imp-gamification-website-forum-website-slides-karma-tracking-improveme.patch
I'm not sure about this last one.

acpMicrocom:
> I'm also not sure, but may be the karma computation should be triggered, See the change here https://github.com/akretion/odoo-module-diff-analysis/blob/main/17.0/gamification/c000__________-%23_76430_imp-gamification-website-forum-website-slides-karma-tracking-improveme.patch
> I'm not sure about this last one.

@rvalyi Thank you for the feedback.

I noticed that the karma field update doesn't appear in the upgrade_analysis.txt. Could you advise on the best approach to ensure the computation is triggered in this case? Should we add it manually, or is there another recommended method to handle fields that aren't included in the analysis file?

Thanks in advance for your guidance.

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/4717.

PR #4637, opened 2024-10-25
[17.0][OU-ADD] delivery: Migration to 17.0

acpMicrocom:


legalsylvain:
/ocabot migration delivery

<dependency>Depends on : 

- sale : #4691</dependency>

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,76 @@
...
+        FROM stock_picking sp
+        WHERE sml.picking_id = sp.id;
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_xmlids(env.cr, _xmlids_renames)
+    openupgrade.add_fields(env, _new_fields)
```
MiquelRForgeFlow, line 74: no need, as you will fill the carrier_id by sql. so, better to create column by sql in this case

openupgrade_scripts/scripts/delivery/17.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,95 @@
+---Models in module 'delivery'---
+model choose.delivery.package (moved to stock_delivery) [column['model_type']]
+# DONE: in migration script of module stock_delivery
```
MiquelRForgeFlow, line 3: no migration script will be executed, as stock_delivery is new module.

let's install here explicitly this module.

MiquelRForgeFlow (commented):
Let's supersede it in https://github.com/OCA/OpenUpgrade/pull/4751.

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
pedrobaeza: No. Why create it by SQL if we have an API for that?

openupgrade_scripts/scripts/delivery/17.0.1.0/upgrade_analysis_work.txt
pedrobaeza: Migration scripts are also executed in newly installed modules thanks to the patches in `openupgrade_framework`.

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
MiquelRForgeFlow: At least, the field should be added in `stock_delivery` instead :S I prefer let ORM do the work, to avoid messing anything.

openupgrade_scripts/scripts/delivery/17.0.1.0/upgrade_analysis_work.txt
MiquelRForgeFlow: I didn't know, nice. But unneeded in this case.

PR #4641, opened 2024-10-27
[OU-ADD] spreadsheet. Nothing to do

legalsylvain:
https://github.com/OCA/OpenUpgrade/blob/17.0/openupgrade_scripts/scripts/spreadsheet/17.0.1.0/upgrade_analysis.txt


legalsylvain:
/ocabot migration spreadsheet

<dependency></dependency>

hbrunn:
given the actual data lives in a json blob, did you check if we need to convert that? As in, can we open a spreadsheet created in v16 with v17?

legalsylvain:
> did you check if we need to convert that? 

No, I didn't checked. I'll give it a try and tell you.

Thanks for your question !

pedrobaeza:
The spreadsheet library includes their own migration engine, and it's included in the main community module. My only doubt is if it's auto-run on opening or not.

hbrunn (approved):
I just diffed a couple of spreadsheet definitions between v16 and v17, and they all have the same version with just some data changes, so I think this is fine.

PR #4461, opened 2024-06-13
[17.0][OU-ADD] digest: migration to 17.0

daiduongnguyen-odoo:

PR #4642, opened 2024-10-28
[17.0][OU-ADD] digest: migration to 17.0

hbrunn:
#4461 with review followup


hbrunn:
/ocabot migration digest

<dependency></dependency>

pedrobaeza:
I have pushed some fixes:

- `load_data` should be done on post, as the record may be deleted on 16, and then the new record creation will fail due to missing required fields.
- We have to reset translations for the template that has changed to have the new definition for other languages than `en_US`.

pedrobaeza (approved):

PR #4562, opened 2024-09-19
[17.0][OU-ADD] website: Migration to 17.0

acpMicrocom:

PR #4645, opened 2024-10-29
[17.0][OU-ADD] website_mail: Nothing to do

acpMicrocom:
- While migrating the `website_mail` script to version 17, I found that there is no migration directory or file specifically for version 17. I would appreciate guidance on the correct approach here. Should a new migration directory and script be created for this version, or is there another standard procedure for handling this?


legalsylvain:
/ocabot migration website_mail

<dependency>Depends on : 

- website : #4562</dependency>

MiquelRForgeFlow (approved):

PR #4655, opened 2024-11-07
[17.0][OU-FIX] resource: Reset sequence field in resource.calendar.attendance

pedrobaeza:
As the field `sequence` exists previously in 16, and you were able to change it on screen:

imagen

there's a chance that your DB have some sequence values that alters the previous order when migrating.

Thus, the safest thing is to reset the sequence value to its default value for preserving always the previous order.

@Tecnativa TT49327


legalsylvain (approved):
Nice catch !

PR #4660, opened 2024-11-08
[17.0][OU-ADD] web_pwa_oca: Merged into web

victoralmau:
Merged into `web`

@Tecnativa TT51658

### DIFF
--- a/openupgrade_scripts/scripts/web/17.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/web/17.0.1.0/pre-migration.py
@@ -1,0 +1,29 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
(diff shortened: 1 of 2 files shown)


legalsylvain:
/ocabot migration web_pwa_oca

In fact, I think it's great if the main 17.0 issue mentions the changes made by OpenUpgrade also for OCA modules.

What do you think ?  CC : @pedrobaeza 

here, there is an apriori change + script.

pedrobaeza:
I don't think so. It should be manually added to OCA/web migration issue, not here.

openupgrade_scripts/scripts/web/17.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    web_pwa_oca_installed = env["ir.module.module"].search(
```
pedrobaeza: The module doesn't "exist" at this level, as it has been merged. You have to detect it by any column existence or similar.

rvalyi (approved):

pedrobaeza (approved):

PR #4663, opened 2024-11-14
[17.0][OU-ADD] event: Migration scripts

victoralmau:
Migration to 17.0

Depends on:
- [x] `utm`: https://github.com/OCA/OpenUpgrade/pull/4664

@Tecnativa TT49909


legalsylvain:
/ocabot migration event

<dependency></dependency>

victoralmau:
I am not sure if the analysis is up to date, I just saw that (for example) the `seats_expected` field does not exist in v17. 
Can you update the analysis?

openupgrade_scripts/scripts/event/17.0.1.8/pre-migration.py
```diff
@@ -0,0 +1,22 @@
...
+        ADD COLUMN IF NOT EXISTS company_name VARCHAR
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE event_registration AS r
+        SET company_name = p2.company_name
+        FROM res_partner AS p
+        LEFT JOIN res_partner AS p2 ON (p2.parent_id = p.id AND p2.type = 'contact')
+        WHERE r.partner_id IS NOT NULL AND p2.company_name IS NOT NULL""",
```
MiquelRForgeFlow:
```suggestion
        SET company_name = p.company_name
        FROM res_partner AS p
        WHERE r.partner_id IS NOT NULL and p.type = 'contact' AND p.company_name IS NOT NULL;
        UPDATE event_registration AS r
        SET company_name = p2.company_name
        FROM res_partner AS p
        JOIN res_partner AS p2 ON (p2.parent_id = p.id AND p2.type = 'contact')
        WHERE r.partner_id IS NOT NULL AND p2.company_name IS NOT NULL AND r.company_name IS NULL;""",
```
In fact, the two queries can be fused, but it would get uglier. I prefer clarity.

openupgrade_scripts/scripts/event/17.0.1.8/pre-migration.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
```
MiquelRForgeFlow:
```suggestion
@openupgrade.migrate()
def migrate(env, version):
    fill_event_registration_company_name(env)
```
please, group them in a separated method

MiquelRForgeFlow (approved):
LGTM

openupgrade_scripts/scripts/event/17.0.1.8/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+
+_deleted_xml_records = [
+    "event.event_type_data_conference",
+    "event.event_type_data_ticket",
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "event", "17.0.1.8/noupdate_changes.xml")
```
pedrobaeza, line 13: Here it's missing the reset of the translations for the mail templates.

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
+---Models in module 'event'---
+---Fields in module 'event'---
+event        / event.event              / _order                        : _order is now 'date_begin, id' ('date_begin')
+event        / event.event              / activity_user_id (many2one)   : not related anymore
+event        / event.event              / activity_user_id (many2one)   : now a function
+# NOTHING TO DO
+event        / event.event              / auto_confirm (boolean)        : DEL
+# NOTHING TO DO: Attendees are now created confirmed (state=open)
+event        / event.event              / badge_format (selection)      : NEW required, selection_keys: ['A4_french_fold', 'A6', 'four_per_sheet'], hasdefault: default
```
pedrobaeza, line 9: Next time, please document this nothing to do, saying that the default value is the same as the previous behavior.

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
...
+event        / event.event              / badge_image (binary)          : NEW attachment: True
+# NOTHING TO DO
+event        / event.event              / lang (selection)              : NEW selection_keys: function
+# NOTHING TO DO: It is not required, it can be empty, we do not set the value of the old records
+event        / event.event              / message_main_attachment_id (many2one): DEL relation: ir.attachment
+event        / event.event              / rating_ids (one2many)         : NEW relation: rating.rating
+event        / event.event              / registration_properties_definition (properties_definition): NEW
+# NOTHING TO DO
+event        / event.event.ticket       / _order                        : _order is now 'event_id, sequence, name, id' ('id')
+event        / event.event.ticket       / color (char)                  : NEW hasdefault: default
```
pedrobaeza, line 19: How was the colors handled previously? All the tickets with the same color are OK?

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
...
+event        / event.event.ticket       / sequence (integer)            : NEW hasdefault: default
```
pedrobaeza, line 20: The sequence should be equal to the id to respect the same ordering as before.

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
+---Models in module 'event'---
+---Fields in module 'event'---
+event        / event.event              / _order                        : _order is now 'date_begin, id' ('date_begin')
+event        / event.event              / activity_user_id (many2one)   : not related anymore
+event        / event.event              / activity_user_id (many2one)   : now a function
+# NOTHING TO DO
+event        / event.event              / auto_confirm (boolean)        : DEL
```
pedrobaeza, line 7: As this is plain text, you need to leave a blank line between each block where you insert comments, or it's very difficult to be read.

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
...
+event        / event.event.ticket       / color (char)                  : NEW hasdefault: default
+event        / event.event.ticket       / sequence (integer)            : NEW hasdefault: default
+# NOTHING TO DO
+event        / event.registration       / activity_user_id (many2one)   : not related anymore
+event        / event.registration       / activity_user_id (many2one)   : now a function
+# NOTHING TO DO
+event        / event.registration       / barcode (char)                : NEW
+# NOTHING TO DO: There is no correct way to do it in postgres, we let the ORM set the value of the old records
+event        / event.registration       / company_name (char)           : NEW hasdefault: compute
+# DONE: Define the value of existing records in pre-migration
```
pedrobaeza, line 28: The format for this is:

```
# DONE: pre-migration: Pre-create and fill the column for avoiding the ORM computation
```

pedrobaeza (commented):
@victoralmau please fix my comments in an extra PR.

@MiquelRForgeFlow it's convenient to not merge this so lightly.

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
...
+event        / event.registration       / message_main_attachment_id (many2one): DEL relation: ir.attachment
+event        / event.registration       / mobile (char)                 : DEL
+event        / event.registration       / rating_ids (one2many)         : NEW relation: rating.rating
+event        / event.registration       / registration_properties (properties): NEW hasdefault: compute
+# NOTHING TO DO
+event        / event.type               / auto_confirm (boolean)        : DEL
+# NOTHING TO DO: Explained in auto_confirm field of event.event
+event        / event.type.ticket        / _order                        : _order is now 'sequence, name, id' ('id')
+event        / event.type.ticket        / sequence (integer)            : NEW hasdefault: default
```
pedrobaeza, line 37: This also requires to be set for not changing the previous order.

openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt
victoralmau: Before there was no color, and now it is not used anywhere except in the event form view to define the tickets (i have not seen it used elsewhere). Also, defining random colors I don't think was a good decision either.

pedrobaeza: How this works if you create the same tickets in this version?

victoralmau: I think this is only used in the form view of the event. 
test

Is it a problem that old tickets have the same color defined?

PR #4664, opened 2024-11-14
[17.0][OU-ADD] utm: Nothing to do

victoralmau:
Nothing to do

@Tecnativa TT49337


hbrunn (changes requested):
why do you do anything manually here? This should be taken care of by the ORM, so I'd file the module as nothing to do

hbrunn (approved):

PR #4666, opened 2024-11-18
[17.0][OU-ADD] fleet: Nothing to do

victoralmau:
Nothing to do

@Tecnativa TT49301


hbrunn (changes requested):
I think this is nothing to do, as the column is created and filled automatically by the ORM

hbrunn (approved):

PR #4667, opened 2024-11-18
[16.0][FIX] base: avoid ignoring right translations

chienandalu:
DBs with a long versions history could have right translations with wrong to_translate states. We should try to fix them in advance to avoid missing them.

cc @Tecnativa TT51774

please check @pedrobaeza


pedrobaeza:
The CI has been fixed in 17.0 this way: https://github.com/OCA/OpenUpgrade/pull/4669/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R88

MiquelRForgeFlow:
Could you rebase? Error is solved.

pedrobaeza:
Let's use ocabot for that:

/ocabot merge nobump

PR #4668, opened 2024-11-18
[17.0][OU-ADD] account: Migration to 17.0

hbrunn:
continuation of #4662

I tested this with my own DB, a few customers', and the demo database and couldn't find issues any more, so from my side I think it's done.

PR #4669, opened 2024-11-18
[17.0][OU-ADD] enable tests, add decorator to make them convenient to use

hbrunn:
the decorator is used in the [account migration](https://github.com/OCA/OpenUpgrade/pull/4668).

I made this a decorator so that it doesn't interfere with the tests' metaclass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,8 +91,8 @@ jobs:
           pip install asn1crypto
       - name: Test data
         run: |
-          if test -n "$(ls openupgrade/openupgrade_scripts/scripts/*/*/tests/data*.py 2> /dev/null)"; then
-             for snippet in openupgrade/openupgrade_scripts/scripts/*/*/tests/data*.py; do
+          if test -n "$(ls openupgrade/openupgrade_scripts/scripts/*/tests/data*.py 2> /dev/null)"; then
+             for snippet in openupgrade/openupgrade_scripts/scripts/*/tests/data*.py; do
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -91,8 +91,8 @@ jobs:
           pip install asn1crypto
       - name: Test data
         run: |
-          if test -n "$(ls openupgrade/openupgrade_scripts/scripts/*/*/tests/data*.py 2> /dev/null)"; then
-             for snippet in openupgrade/openupgrade_scripts/scripts/*/*/tests/data*.py; do
+          if test -n "$(ls openupgrade/openupgrade_scripts/scripts/*/tests/data*.py 2> /dev/null)"; then
+             for snippet in openupgrade/openupgrade_scripts/scripts/*/tests/data*.py; do
```
hbrunn, line 96: this is because tests must live in the base directory of the addon's path

.github/workflows/test.yml
```diff
@@ -131,14 +131,16 @@ jobs:
           echo Execution of Openupgrade with the update of the following modules : $MODULES_NEW
           # Silence redundant logs from unlinking records (1 line is enough)
           # to prevent log overflow
-          OPENUPGRADE_TESTS=1 $ODOO \
+          $ODOO \
```
hbrunn, line 135: the env var was a leftover from the pre-openupgrade_framework test mechanism

pedrobaeza (approved):

PR #4672, opened 2024-11-21
[16.0][OU-ADD] enable tests, add decorator to make them convenient to use

chienandalu:
backport of

- #4669 

please review @pedrobaeza @hbrunn

### DIFF
--- a/.github/workflows/documentation-commit.yml
+++ b/.github/workflows/documentation-commit.yml
@@ -52,6 +52,7 @@ jobs:
               unixodbc-dev
       - name: Requirements Installation
         run: |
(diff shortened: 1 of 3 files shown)


chienandalu:
looks like it doesn't work as it is... :sweat_smile:

hbrunn:
you'll need to replace https://github.com/OCA/OpenUpgrade/pull/4672/commits/5c38e0500de43cfd73bea50c7d19910493127ad5 with the version for v16 from https://github.com/Tecnativa/doodba/pull/636

And better don't backport the other commit, as in v16 we use the [patch](https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_framework/odoo_patch/odoo/tests/loader.py) I added before there was https://github.com/OCA/OCB/blob/16.0/odoo/tests/loader.py#L38, and we already have tests working with this mechanism, ie base

MiquelRForgeFlow:
Superseded in https://github.com/OCA/OpenUpgrade/pull/4701.

PR #4680, opened 2024-11-28
[15.0][OU-FIX] web: Set the correct report_footer value

victoralmau:
Set the correct report_footer value

Related to https://github.com/OCA/OpenUpgrade/pull/4095

@Tecnativa TT52003


pedrobaeza (changes requested):
This is missing the decorations to be exactly as it was (`<ul>`, `<li>`, icons, etc).

pedrobaeza:
Have you checked that these classes are correctly exported to PDF?

victoralmau:
Yes, it is checked.

**Configure document layout wizard**
ejemplo-1

**Pdf footer**
ejemplo-2

Still, I think an **extra validation** of this change would be nice (for example in a migration).

pedrobaeza:
Great, let's test it in the migration we are going to perform, cc @chienandalu 

I'm merging now and possible amendments will be done through subsequent PRs.

PR #4692, opened 2024-12-09
[17.0][OU-ADD] payment: Migration to 17.0

hbrunn:
continuation of https://github.com/OCA/OpenUpgrade/pull/4584


openupgrade_scripts/scripts/payment/17.0.2.0/noupdate_changes.xml
```diff
@@ -1,5 +1,355 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo>
+    <record id="payment_method_bancontact" model="payment.method">
```
MiquelRForgeFlow: add a note saying that these changes are generated manually.

hbrunn: done, and I moved the manual changes to a `_work` file as proposed in https://github.com/OCA/OpenUpgrade/issues/4707#issuecomment-2610570291

openupgrade_scripts/scripts/payment/17.0.2.0/noupdate_changes_work.xml
```diff
@@ -0,0 +1,406 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo>
+    <!--
+    these methods are created manually
+    /-->
```
legalsylvain: I think that if we copy paste this file with a ``_work`` suffix, we should allways use a very explicit tag in the XML description. so we can very quickly know what is the original code, and what is added by Openugprade.

I propose something like : 

    

and just before the ``  <record id="payment_provider_adyen" model="payment.provider">``

something like : 

    

what do you think ? 

CC : @pedrobaeza, @StefanRijnhart, @MiquelRForgeFlow .

hbrunn: fine with me, added that

MiquelRForgeFlow:
Could you rebase?

hbrunn:
done

MiquelRForgeFlow (approved):

remi-filament:
Hi @hbrunn @MiquelRForgeFlow @pedrobaeza 
I have a problem with this migration script when trying to migrate a DB from v16 where I have multiple companies and in my case Stripe on a secondary company.

In this scenario, payment_provider stripe has been duplicated for the second company, and therefore does not hold XML-id. Therefore the new payment methods (and in particular "card" one were not applied on this provider) and when running end-migration script I get the following UserError from payment_method write() method (https://github.com/odoo/odoo/blob/a73211ad089c238d193d928a70d2586cb92e4707/addons/payment/models/payment_method.py#L184) : 
`This payment method needs a partner in crime; you should enable a payment provider supporting this method first`

I am not sure how to proceed from there, and would appreciate your insights !
Should we look at all payment_providers without XML id based on their code and load method_ids for them ? 
That makes sense to me but I am not sure how to do that, maybe copying these from the one with XML id after no_update_changes have been loaded on them ?

pedrobaeza:
Shouldn't you adjust your DB to have the primary provider enabled in the second company?

remi-filament:
You mean using the payment_provider stripe with XML id on my secondary company ?
I cannot do that since I need it also on my primary company (with different configuration) so I need two, but only one carries the XML id.

pedrobaeza:
Then I read bad the code... The other part of the condition is about payment methods. Maybe you need to set them properly?

remi-filament:
Exactly and this is my point, payment methods changed a lot between 16.0 and 17.0 and were reloaded on payment_provider with this file : https://github.com/OCA/OpenUpgrade/blob/cb3a56f1557b50cc7de2b6f775379d0bdc73c351/openupgrade_scripts/scripts/payment/17.0.2.0/noupdate_changes_work.xml#L395

The problem is that during migration we are only updating those payment methods on records with XML id (which were noupdate here), but we are not on the paymment provider created manually from those (which should have the same code as original ones but no XML id).

I need to figure out the best way to reload these updated methods on the payment providers without XML id...

remi-filament:
I could solve the issue by only looking at existing payment methods on payment provider and adding the primary_method from those if missing.
That should solve the UserError, but if feels like it would be an incomplete solution since payment provider would not have been updated with all new payment methods / removing the ones not supported anymore.

pedrobaeza:
OK, let's do that assignment for all the copies of each payment provider. Can you do the PR?

remi-filament:
Yes I will, I would need confirmation on a couple of things to ease the process :

1. Do you know a way to get all objects without an external id ? (I was thinking making a SQL request for all IDs not in ir_model_data, but maybe there is a more elegant way to do ?)
2. Should I rely on the code of the payment provider to determine which one was the source ?

Thanks !

pedrobaeza:
Isn't there the provider code? And I would use the noupdate changes file as the template, and even write the same data for all of them at once and don't load that noupdate data.

PR #4693, opened 2024-12-10
[17.0][OU-ADD] hr_attendance_geolocation: Merged into hr_attendance

victoralmau:
Merged into hr_attendance

Locked by:
- [x] `hr_attendance`: https://github.com/OCA/OpenUpgrade/pull/4801

@Tecnativa TT51930

### DIFF
--- a/openupgrade_scripts/scripts/hr_attendance/17.0.2.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_attendance/17.0.2.0/pre-migration.py
@@ -1,4 +1,5 @@
 # Copyright 2025 ForgeFlow S.L. (https://www.forgeflow.com)
+# Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from openupgradelib import openupgrade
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@legalsylvain this is not an Odoo core module, so please don't raise the bot command.

legalsylvain:
@pedrobaeza : sorry ! didn't know ! (i don't use this module).

chienandalu (approved):

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/pre-migration.py
```diff
@@ -42,5 +69,6 @@ def fill_res_company_attendance_from_systray(env):
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.rename_xmlids(env.cr, _xmlid_renames)
+    openupgrade.rename_fields(env, _fields_renames)
```
pedrobaeza: Encapsulate this into a module explaining that this comes from `hr_attendance_geolocation`, and only being called if you find the SQL column.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/pre-migration.py
```diff
@@ -12,6 +13,38 @@
         "hr_attendance.group_hr_attendance_own_reader",
     ),
 ]
+_fields_renames = [
```
pedrobaeza: Put this inside the method, as this is not something general, and having it here can be misleading.

pedrobaeza (approved):

PR #4694, opened 2024-12-11
[15.0][OU-IMP] new analysis (December 2024)

MiquelRForgeFlow:
It's time.


pedrobaeza:
`[UPD]` tag doesn't exist. I don't know why a lot of people uses it.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

It's the same like saying "I do something... What? You don't know".

MiquelRForgeFlow:
The `[OU]` tag also doesn't exist 😏

pedrobaeza:
It's not `[OU]`, it's `[OU-ADD]`, `[OU-IMP]`, etc, as an extension of the current tag system for indicating that is restricted to the OpenUpgrade scope, as if we put `[ADD]`, we are not adding a whole module, just the migration scripts. It should be indeed reflected in the documentation though. Can you do the corresponding PR?

For this one, I think it should be `[OU-IMP]`.

openupgrade_scripts/scripts/hr/15.0.1.1/noupdate_changes.xml
```diff
@@ -4,6 +4,9 @@
     <field name="address_id" ref="base.main_partner"/>
     <field name="department_id" ref="dep_administration"/>
   </record>-->
+  <record id="offboarding_setup_compute_out_delais" model="hr.plan.activity.type">
```
MiquelRForgeFlow, line 7: not sure if I should comment this record

PR #4695, opened 2024-12-11
[17.0][IMP] base: vacuum of transient models (forwardport)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/OpenUpgrade/pull/4378.


pedrobaeza (approved):

pedrobaeza:
I think this should be added to the documentation of putting a new version of OU in place. cc @legalsylvain @hbrunn @StefanRijnhart

PR #4697, opened 2024-12-12
[17.0][OU-IMP] new analysis (december 2024)

MiquelRForgeFlow:
It hasn't changed much, as the last one was done on October (https://github.com/OCA/OpenUpgrade/commit/9bde91fbf9356f10077870e815292d8083e43f8c).

I added `"sale_quotation_builder": "sale_pdf_quote_builder`" in apriori, because expectations. People having the old one expects having the new one after the migration, independently if the code has changed between the modules.


openupgrade_scripts/scripts/base_import_module/17.0.1.0/upgrade_analysis_work.txt
hbrunn, line 1: what's up with this rename?

hbrunn (changes requested):
I think the rename should go into another commit than the analysis

openupgrade_scripts/scripts/base_import_module/17.0.1.0/upgrade_analysis_work.txt
MiquelRForgeFlow: It's an error from https://github.com/OCA/OpenUpgrade/commit/60f18195d9c9db8b620a2610cbc150a1aaae3312, that was put the file in the wrong folder.

pedrobaeza: cc @legalsylvain please take care of these things being the coordinator.

legalsylvain: Sorry. My bad.

hbrunn:
bu still I think the addition to apriori.py should go in its own commit

hbrunn (approved):

PR #4560, opened 2024-09-18
[17.0][OU-ADD] purchase: Migration to 17.0

acpMicrocom:

PR #4699, opened 2024-12-16
[17.0][OU-ADD] purchase: Migration to 17.0

hbrunn:
continuation of https://github.com/OCA/OpenUpgrade/pull/4560


hbrunn:
/ocabot migration purchase

<dependency>Depends on : 

- account : #4668</dependency>

openupgrade_scripts/scripts/purchase/17.0.1.2/post-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 Viindoo Technology Joint Stock Company (Viindoo)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "purchase", "17.0.1.2/noupdate_changes.xml")
```
MiquelRForgeFlow, line 9: delete template translations

openupgrade_scripts/scripts/purchase/17.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version=None):
+    if not openupgrade.column_exists(env.cr, "purchase_order_line", "discount"):
+        openupgrade.logged_query(
+            env.cr, "ALTER TABLE purchase_order_line ADD COLUMN discount FLOAT"
+        )
+        openupgrade.logged_query(env.cr, "UPDATE purchase_order_line SET discount=0")
```
MiquelRForgeFlow: not sure if it's better this way or doing "ADD COLUMN discount FLOAT set default 0" and later drop the default instead.

openupgrade_scripts/scripts/purchase/17.0.1.2/post-migration.py
hbrunn: done

openupgrade_scripts/scripts/purchase/17.0.1.2/pre-migration.py
hbrunn: I think we'd need a *lot* of record for it to make much difference, but then it doesn't hurt either

openupgrade_scripts/scripts/purchase/17.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,11 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version=None):
+    if not openupgrade.column_exists(env.cr, "purchase_order_line", "discount"):
+        openupgrade.logged_query(
+            env.cr,
+            "ALTER TABLE purchase_order_line ADD COLUMN discount FLOAT " "DEFAULT 0",
```
MiquelRForgeFlow:
```suggestion
            "ALTER TABLE purchase_order_line ADD COLUMN discount FLOAT DEFAULT 0",
```

MiquelRForgeFlow (approved):

PR #4701, opened 2024-12-18
[16.0][OU-FIX] github workflows: apply doodba's patch to fix building gevent

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4672.


pedrobaeza (approved):

sbidoul:
Could we apply this to OCB's requirements.txt?

PR #4702, opened 2024-12-24
[17.0][OU-ADD] hr_holidays: Migration to 17.0

remi-filament:


legalsylvain:
/ocabot migration hr_holidays

<dependency></dependency>

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
```diff
@@ -0,0 +1,121 @@
...
+}
+
+_field_renames = [
+    # (
+    #     "payment.provider",
+    #     "payment_provider",
+    #     "payment_icon_ids",
+    #     "payment_method_ids",
+    # ),
+]
```
Murtaza-SerpentCS: Remove unused code.

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
```diff
@@ -0,0 +1,121 @@
...
+from openupgradelib import openupgrade
+
+_xmlids_renames = [
+    # This one has been removed with https://github.com/odoo/odoo/pull/120446
+    #
+    # (
+    #     "payment.payment_icon_cc_webmoney",
+    #     "payment.payment_method_webmoney",
+    # ),
+]
```
Murtaza-SerpentCS: Remove unused code

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/post-migration.py
```diff
@@ -0,0 +1,42 @@
...
+    allocations = env["hr.leave.allocation"].search(
+        [
+            ("allocation_type", "=", "accrual"),
+            ("state", "=", "validate"),
+            ("accrual_plan_id", "!=", False),
+            ("employee_id", "!=", False),
+            ("number_of_days", ">", 0),
+        ]
+    )
+    allocations.write({"already_accrued": True})
```
Murtaza-SerpentCS: Convert to Query.

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/noupdate_changes.xml
```diff
@@ -12,6 +13,7 @@
   <record id="holiday_status_unpaid" model="hr.leave.type">
     <field name="responsible_ids" eval="[(4, ref('base.user_admin'))]"/>
   </record>
+  -->
```
Murtaza-SerpentCS, line 16: Remove unused code.

remi-filament: Here, I think it is better to keep these so that we see what we should not update.

remi-filament:
Done as per your comments, thanks for your review @Murtaza-SerpentCS

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/noupdate_changes.xml
Murtaza-SerpentCS: I don't think it's useful.

remi-filament: IMO it is, and it actually what is usually done on other scripts as well, see for instannce : https://github.com/OCA/OpenUpgrade/blob/c31d4d644143b5339498ca1215727842fa866de5/openupgrade_scripts/scripts/account/16.0.1.2/noupdate_changes.xml#L9-L30
The use case where it is useful is for when we run analysis again, (like it was done with https://github.com/OCA/OpenUpgrade/pull/4697) by keeping the comments, we know that these data were commented out willingly, and it was not new data which was added after last analysis or data which was forgotten.

carlos-lopez-tecnativa (changes requested):
@remi-filament  could you please rebase the branch?

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,184 @@
+---Models in module 'hr_holidays'---
+obsolete model hr.leave.stress.day
+new model hr.leave.mandatory.day
+# DONE pre-migration: rename model
```
MiquelRForgeFlow, line 4: could you add this in apriori file too?

remi-filament:
@carlos-lopez-tecnativa @MiquelRForgeFlow thanks for your review, rebase done and apriori updated.

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,184 @@
...
+
+---Fields in module 'hr_holidays'---
+hr_holidays  / hr.leave                 / activity_user_id (many2one)   : not related anymore
+hr_holidays  / hr.leave                 / activity_user_id (many2one)   : now a function
+# NOTHING TO DO: not stored field
+
+hr_holidays  / hr.leave                 / category_id (many2one)        : not a function anymore
+# NOTHING TO DO: states and validate removed only
+
+hr_holidays  / hr.leave                 / company_id (many2one)         : NEW relation: res.company, isfunction: function, stored
```
carlos-lopez-tecnativa, line 14: I think using SQL instead of ORM to fill this field is a good option, especially for databases with a large number of records. What do you think?

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
```diff
@@ -0,0 +1,97 @@
...
+            ADD COLUMN IF NOT EXISTS is_based_on_worked_time BOOLEAN;
+        """,
+    )
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE hr_leave_accrual_plan plan
+        SET is_based_on_worked_time = level.is_based_on_worked_time
+        FROM  hr_leave_accrual_level level
+        WHERE level.accrual_plan_id = plan.id;
```
carlos-lopez-tecnativa: I think that if any level is set to `True`, then the new field in the plan must also be set to `True`.

In the current script, the field in the plan is set to `False` because there is no defined order—it takes the first, last, or a random level?.
V16, Set to `True` on the second level
image

V17
image

```suggestion
        UPDATE hr_leave_accrual_plan plan
        SET is_based_on_worked_time = subquery.max_value
        FROM (
            SELECT accrual_plan_id, MAX(is_based_on_worked_time::INT) AS max_value
            FROM hr_leave_accrual_level
            GROUP BY accrual_plan_id
        ) subquery
        WHERE subquery.accrual_plan_id = plan.id;
```

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
```diff
@@ -0,0 +1,97 @@
...
+]
+
+
+def _map_leave_accrual_level_action(cr):
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE hr_leave_accrual_level
+        SET action_with_unused_accruals = 'all'
+        WHERE action_with_unused_accruals = 'postponed';
```
carlos-lopez-tecnativa: I think setting this field to `all` always is incorrect; it must depend on the `maximum_leave` field.
Otherwise, it can never be applied according to [this code](https://github.com/odoo/odoo/blob/7bf18616e4ae39c5e712339cfd560b57018a93de/addons/hr_holidays/models/hr_leave_allocation.py#L521-L524
)

V16
image
V17
image

V17 expected behavior
image

```suggestion
        UPDATE hr_leave_accrual_level
        SET action_with_unused_accruals = CASE 
                WHEN maximum_leave > 0 THEN 'maximum' 
                ELSE 'all' 
                END
        WHERE action_with_unused_accruals = 'postponed';
```

carlos-lopez-tecnativa (changes requested):
Tested and working, but I think we need to adjust some SQL to maintain the current behavior in V16. Please let me know if you have any doubts.
TT49870
cc: @pedrobaeza

carlos-lopez-tecnativa:
ping @remi-filament

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
```diff
@@ -0,0 +1,129 @@
...
+        cr,
+        """UPDATE hr_leave AS leave
+        SET company_id = CASE
+            WHEN employee_company_id IS NOT NULL THEN employee_company_id
+            WHEN mode_company_id IS NOT NULL THEN mode_company_id
+            WHEN leave.department_id IS NOT NULL THEN d.company_id
+            END
+        FROM hr_department AS d
+        WHERE leave.company_id IS NULL AND leave.department_id = d.id""",
+    )
```
MiquelRForgeFlow:
```suggestion
    openupgrade.logged_query(
        cr,
        """UPDATE hr_leave AS leave
        SET company_id = CASE
            WHEN employee_company_id IS NOT NULL THEN employee_company_id
            WHEN mode_company_id IS NOT NULL THEN mode_company_id
            ELSE NULL
            END
        WHERE leave.company_id IS NULL""",
    )
    openupgrade.logged_query(
        cr,
        """UPDATE hr_leave AS leave
        SET company_id = d.company_id
        FROM hr_department AS d
        WHERE leave.company_id IS NULL AND leave.department_id = d.id""",
    )
```

remi-filament:
Thanks @carlos-lopez-tecnativa and @MiquelRForgeFlow I implemented the changes as per your review !

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _map_leave_accrual_level_action(cr):
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE hr_leave_accrual_level
+        SET action_with_unused_accruals = CASE
+                WHEN maximum_leave > 0 THEN 'maximum'
```
carlos-lopez-tecnativa: Please fine-tune this to avoid considering it if the value is zero.
```suggestion
                WHEN maximum_leave > 0 AND postpone_max_days > 0  THEN 'maximum'
```
V16
image

V17: If the value is zero, it is better to keep the default value.
image

carlos-lopez-tecnativa (changes requested):
Just a minor change, and I think it's ready to merge.

carlos-lopez-tecnativa (approved):
LGTM

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,186 @@
...
+# DONE: pre-migration propagate value from hr.leave.accrual.level to hr.leave.accrual.plan
+
+hr_holidays  / hr.leave.accrual.level   / parent_id (many2one)          : DEL relation: hr.leave.accrual.level
+# NOTHING TO DO: removed field without replacement
+
+hr_holidays  / hr.leave.accrual.plan    / accrued_gain_time (selection) : NEW required, selection_keys: ['end', 'start'], hasdefault: default
+# NOTHING TO DO: new field with default value = v16 behaviour
+
+hr_holidays  / hr.leave.accrual.plan    / active (boolean)              : NEW hasdefault: default
+# NOTHING TO DO: new field, default = active
```
MiquelRForgeFlow: please, do like `_pre_create_account_report_active` in pre-migration of `account`.

remi-filament: Done, what is the rationale to create columns in SQL instead of letting Odoo do it, is it so much performance gain ?
Should we do this for all new fields with default value ? (there are lots of them on hr.leave.accrual.plan : accrued_gain_time, added_value_type, carryover_date, carryover_day, carryover_month).

carlos-lopez-tecnativa: My logic is that if a model is transactional (with many records, such as `account.move`, `account.move.line`, `stock.move`, `stock.move.line`, `sale.order`, `sale.order.line`, etc.), it should be handled using SQL instead of the ORM. However, if the model is a master record (with only a few records, typically settings), then using the ORM is sufficient.

It would be great to hear the logic from others, such as @pedrobaeza @MiquelRForgeFlow, to take different perspectives into account.

MiquelRForgeFlow (approved):

PR #4705, opened 2024-12-25
[16.0][OU-FIX] stock: do not try to compute quantity_done for canceled stock moves

em230418:
Otherwise UserError will be raised [1]

[1] https://github.com/odoo/odoo/blob/0043634d87fdde1fa427eda90e444fe76767bacb/addons/stock/models/stock_move.py#L675-L676

### DIFF
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
@@ -49,6 +49,7 @@ def _complete_stock_move_quantity_done_with_orm(env):
         """
         SELECT move_id
         FROM stock_move_line
+        WHERE state != 'cancel'
         GROUP BY move_id
         HAVING COUNT(DISTINCT product_uom_id) > 1
             AND SUM(qty_done) <> 0


pedrobaeza:
Is it possible to have stock.move.line with quantity done when the move is cancelled?

em230418:
> Is it possible to have stock.move.line with quantity done when the move is cancelled?

Yes.

1. Make sure your user is in group "Manage Multiple Stock Locations"
2. Inventory -> Configuration -> Operation Types
3. For example choose "Delivery orders"
4. Add tick on "Show Detailed Operations" checkbox, save
5. Inventory -> Operations -> Transfers
6. Create transfer with "Delivery orders" type, add product that exist in stock. Demand value for example 1. Save
7. "Mark as Todo" button
8. "Set quantities" button
9. "Cancel" button

In "detailed operations" tab you will see red line with value 1 quantity done.

Reproduced here.
http://oca-event-15-0-a74d199cc222.runboat.odoo-community.org/web?debug=1#cids=1&menu_id=108&action=234&model=stock.picking&view_type=form&id=32

In production database on 14.0 I have 2180 canceled stock move lines from 2022 to 2024.

em230418:
Today I checked, that my patch worked so marked as ready for review.,

pedrobaeza (approved):
OK, it doesn't harm either.

PR #4710, opened 2025-01-14
[16.0][FIX] stock: handle improper variable declaration

PicchiSeba:
In the current implementation `view` will always be truthy, even when no view is found

With this PR the intended behavior is restored

### DIFF
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/post-migration.py
@@ -17,7 +17,7 @@ def _handle_multi_location_visibility(env):
             "stock.stock_location_view_tree2_editable",
             "stock.stock_location_view_form_editable",
         ):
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

PicchiSeba:
Hello @pedrobaeza , I can see that the merge button is available to me.

Screenshot From 2025-01-14 11-17-40

I'm reporting this to you because I am not a maintainer, I don't think I should have access to it

pedrobaeza:
I don't get you. It's already merged.

PicchiSeba:
@pedrobaeza 
You merged before I could finish my message lol.

The thing is before you merged this, I could have too. But I'm not a maintainer.
I don't think I should be able to

pedrobaeza:
I don't think so. The dropdown you are showing is not for merging, but for updating your branch if I'm not mistaken.

PicchiSeba:
oh my, sorry for the trouble

PR #4713, opened 2025-01-22
[17.0][OU-ADD]  stock_lot_filter_available: Merged into stock

CarlosRoca13:
cc @Tecnativa TT54326

Option to filter on hand lots added on this commit to stock module:
https://github.com/odoo/odoo/commit/03b6d593eac4a3d9c977687619b4da7c7386a3fd#diff-2730b4e72907db5b9b8ae730655aefc27ea15f2a884111490527d3cdadcbfc7fR129

ping @sergio-teruel @pedrobaeza

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -40,6 +40,8 @@ merged_modules = {
     "l10n_es_irnr_sii": "l10n_es_aeat_sii_oca",
     # OCA/purchase-workflow
     "purchase_discount": "purchase",
+    # OCA/stock-logistics-warehouse
+    "stock_lot_filter_available": "stock",
     # OCA/web
     "web_advanced_search": "web",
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

pedrobaeza:
Aborting, `stock` is not merged. And anyway, no need to use ocabot when it's not outdated.

StefanRijnhart:
@pedrobaeza Does it matter for apriori mappings whether a core module is covered with migration scripts in OpenUpgrade?

pedrobaeza:
Well, it can make the CI to fail under some circumstances, so it's better to wait for having the dependency and then check if there's any problem. This happens specially when there are also migration scripts for doing some stuff after merging the module. It's not the case in this PR, but it's convenient to always treat all cases the same way IMO.

PR #4718, opened 2025-01-22
[17.0][OU-IMP] base: avoid pylint relative-import

MiquelRForgeFlow:
Selection_4222


openupgrade_scripts/scripts/gamification/17.0.1.0/pre-migration.py
```diff
@@ -17,7 +17,7 @@ def new_tracking_fields(env):
         ALTER TABLE gamification_karma_tracking
             ADD COLUMN IF NOT EXISTS origin_ref varchar;
         UPDATE gamification_karma_tracking
-        SET origin_ref = 'res.users,' || create_uid;
+        SET origin_ref = 'res.users,' || user_id;
```
pedrobaeza, line 20: This change is unrelated

PR #4735, opened 2025-02-12
[15.0][OU-FIX] web: fix report_footer migration

HekkiMelody:
There were two issues with the previous migration scripts

1. A missing += meant half the script was overwritten
2. Creating a `base.document.layout` model meant the default report footer was immediately applied to the company, so the result for companies without a pre-existing footer would effectively be duplicated information


pedrobaeza (approved):

HekkiMelody:
Wait, pre-commit is red.

EDIT: Was just some formatting, should be good now.

HekkiMelody:
Thanks @pedrobaeza !

PR #4738, opened 2025-02-13
[16.0][IMP] copier and github actions

MiquelRForgeFlow:
...


openupgrade_scripts/readme/CONFIGURE.rst
pedrobaeza, line 1: I don't see the.md counterparts

MiquelRForgeFlow: They were deleted by the pre-commit automatically. I will add them again.

pedrobaeza: They are transformed into .md. Maybe you have forgotten `git -a` or you don't have locally the proper tools to do it, but in any case, this is incorrect.

MiquelRForgeFlow: Done!

pedrobaeza (approved):

PR #4739, opened 2025-02-13
[16.0][OU-IMP] purchase_requisition: link purchase orders based on deprecated requisition type

MiquelRForgeFlow:
In v16, the call for tenders flow has disappeared, changed by the purchase.order.group link between purchases. This PR handles this by using the deprecated requisition type.

Otherwise, you won't be able to delete the deprecated requisition type because `type_id` of `purchase.requisition` is a `required` field.


pedrobaeza (approved):

openupgrade_scripts/apriori.py
```diff
@@ -11,6 +11,7 @@
     "pos_sale_gift_card": "pos_sale_loyalty",
     "sale_coupon": "sale_loyalty",
     "sale_coupon_delivery": "sale_loyalty_delivery",
+    "purchase_requisition_stock_dropshipping": "purchase_requisition_stock",
```
pedrobaeza: Oh, wait, incorrect section. It should be in `merged_modules`.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/purchase_requisition/16.0.0.1/post-migration.py
```diff
@@ -3,8 +3,66 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        INSERT INTO purchase_order_group (old_purchase_requisition_id,
+            create_date, create_uid, write_date, write_uid)
+        SELECT DISTINCT ON (pr.id) pr.id,
+            pr.create_date, pr.create_uid, pr.write_date, pr.write_uid
+        FROM purchase_requisition pr
+        JOIN purchase_order po1 ON po1.requisition_id = pr.id
+        JOIN purchase_order po2 ON po2.requisition_id = pr.id AND po1.id != po2.id
```
pedrobaeza, line 24: Can you link here `po1`?

MiquelRForgeFlow: Yes.

PR #4709, opened 2025-01-13
[18.0] initialize+openupgrade_framework+analysis+base

hbrunn:
Left to do before we can sensibly start with migrating modules:

- [x] move the [function](https://github.com/OCA/OpenUpgrade/pull/4709/files#diff-3de28e94184027666025e88c51d1f3c2c1c74981392d9c610d432c4a5be396d6R7) to convert company dependent fields and the [patch](https://github.com/OCA/OpenUpgrade/pull/4709/files#diff-efdc354a0d25840c0b9822891d88e771cbf28e9dcfb0198a22e2399852fd35a9R28) to allow (env, cr) signatures in migration functions to openupgradelib: https://github.com/OCA/openupgradelib/pull/396
- [x] make upgrade_analysis flag company dependent fields also if nothing changed (alternative would be to look at ir_properties in base's end-migration and just migrate whatever fields exist there. Seems error prone to me): https://github.com/OCA/server-tools/pull/3172
- [x] analyze the [base analysis file](https://github.com/OCA/OpenUpgrade/pull/4709/files#diff-b89082dcf791c14ebb38550eb37ecd9244b407f16a9e24168b8f64909190bbc3) more deeply

PR #4742, opened 2025-02-17
[18.0][OU-ADD] bus: nothing do do

hbrunn:
depends on

- [x] #4709
- [x] #4741


legalsylvain:
/ocabot migration bus

<dependency>Depends on : 

- web : #4741</dependency>

PR #4743, opened 2025-02-17
[18.0][OU-ADD] html_editor: nothing to do

hbrunn:
depends on

- [x] #4709
- [x] #4741
- [x] #4742


legalsylvain:
/ocabot migration html_editor

<dependency></dependency>

openupgrade_scripts/scripts/html_editor/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,11 @@
+---Models in module 'html_editor'---
+---Fields in module 'html_editor'---
+html_editor  / ir.attachment            / image_height (integer)        : previously in module web_editor
+html_editor  / ir.attachment            / image_src (char)              : previously in module web_editor
+html_editor  / ir.attachment            / image_width (integer)         : previously in module web_editor
+html_editor  / ir.attachment            / local_url (char)              : previously in module web_editor
+html_editor  / ir.attachment            / original_id (many2one)        : previously in module web_editor
+
+# NOTHING TO DO: odoo handles this transition automatically
```
MiquelRForgeFlow, line 9: During migration, `html_editor` will be installed automatically if `web_editor` was installed, right?

hbrunn: it will be pulled by the dependencies mechanism. We could also rename the module, that would be a nice test case for https://github.com/OCA/OpenUpgrade/pull/4873

pedrobaeza:
I think it should be renamed.

hbrunn:
ah right, that's why I didn't do this, they split this off web_editor, which still exists in v18: https://github.com/OCA/OCB/tree/18.0/addons/web_editor

pedrobaeza:
OK, then as you have said, it will be installed by dependencies.

MiquelRForgeFlow (approved):

PR #4744, opened 2025-02-17
[18.0][OU-ADD] web_tour

hbrunn:
depends on

- [x] #4709
- [x] #4741


legalsylvain:
/ocabot migration web_tour

<dependency></dependency>

openupgrade_scripts/scripts/web_tour/18.0.1.0/end-migration.py
```diff
@@ -0,0 +1,25 @@
...
+        FROM
+        {openupgrade.get_legacy_name('web_tour_tour')} legacy_table,
+        web_tour_tour
+        WHERE web_tour_tour.name=legacy_table.name
+        ON CONFLICT DO NOTHING
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        f"DROP TABLE {openupgrade.get_legacy_name('web_tour_tour')}",
```
MiquelRForgeFlow: why? that should be done we database cleanup, right?

I think you should call `openupgrade.remove_tables_fks(env.cr, [openupgrade.get_legacy_name('web_tour_tour')])` instead. Also, better to do it in pre-migration.

hbrunn: done

openupgrade_scripts/scripts/web_tour/18.0.1.0/end-migration.py
```diff
@@ -0,0 +1,21 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        INSERT INTO res_users_web_tour_tour_rel
+        (res_users_id, web_tour_tour_id)
+        SELECT legacy_table.user_id, web_tour_tour.id
+        FROM
+        {openupgrade.get_legacy_name('web_tour_tour')} legacy_table,
+        web_tour_tour
+        WHERE web_tour_tour.name=legacy_table.name
```
MiquelRForgeFlow, line 18: what if a tour changed the name?

hbrunn: see comment in pre-migration, doesn't happen in core modules and if it happens in some other module, it's the job of this module's migration script to change the name in the legacy table anyways

MiquelRForgeFlow (approved):

PR #4747, opened 2025-02-19
[17.0][OU-FIX] sale: Order state + locked

pedrobaeza:
- Preserve a copy of the column `state` for any future need, as we modify its values.
- We should only set as locked orders in "done" state. The locking group doesn't matter.

@Tecnativa TT49328


MiquelRForgeFlow:
> The locking group doesn't matter

Why?

pedrobaeza:
Because that group is only for auto-locking when confirming the sales order or to show the buttons, but the only important thing now is the current order state. You may have the locking group, but have unlocked a specific sales order, or you may not have now the locking group, but lock in the past some orders. Simply `state='sale'` means `locked=True`.

MiquelRForgeFlow (approved):

PR #4749, opened 2025-02-19
[17.0][OU-FIX] sale_stock: Missing migration script

pedrobaeza:
Honor https://github.com/OCA/OpenUpgrade/blob/44ca5988eb34de588b28c00342a6d54c5ce4e71d/openupgrade_scripts/scripts/account/17.0.1.2/upgrade_analysis_work.txt#L138

@Tecnativa TT49892


MiquelRForgeFlow:
Shouldn't you modify the analysis work file?

pedrobaeza:
No, as this field is not logged in this module. It's just the overridden computed.

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/sale_stock/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Fill delivery_date and incoterm_location according
+    # https://github.com/odoo/odoo/blob/00818b7bfaf22635b0c40b3b9c7e37e0c9789da1/
+    # addons/sale_stock/models/account_move.py#L116-L134
+    openupgrade.logged_query(
```
MiquelRForgeFlow, line 12: I think we should apply some WHERE conditions... I mean, what happens if you have a LOT of sale order lines performance-wise?

pedrobaeza: This SQL is already optimized, so it won't suffer in case of a big number of lines. You can try in a DB with lots of them (or use the populate tool for it).

pedrobaeza: The JOIN conditions are the "WHEREs" :wink:

openupgrade_scripts/scripts/sale_stock/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
...
+            SELECT
+                MAX(so.effective_date) as effective_date,
+                MAX(COALESCE(so.incoterm_location, '')) as incoterm_location,
+                am.id as move_id
+            FROM sale_order so
+            JOIN sale_order_line sol ON sol.order_id = so.id
+            JOIN sale_order_line_invoice_rel rel ON rel.order_line_id = sol.id
+            JOIN account_move_line aml ON rel.invoice_line_id = aml.id
+            JOIN account_move am ON aml.move_id = am.id
+            GROUP BY am.id
```
MiquelRForgeFlow, line 25:
```suggestion
            WHERE am.state != 'cancel'
            GROUP BY am.id
```

pedrobaeza: I don't think so. The date and incoterm are filled even if the invoices are canceled. They may come back to draft in any moment, and you want that information there.

PR #4750, opened 2025-02-19
[16.0][OU-ADD] delivery_carrier_multi_zip: Merged into delivery

victoralmau:
Merged into `delivery`

@Tecnativa TT52409


pedrobaeza:
I don't think we should remove this module, as the prefix is very worst user experience than selecting specific ZIPs.

victoralmau:
https://github.com/OCA/delivery-carrier/pull/990

PR #4758, opened 2025-02-21
[17.0][OU-ADD] mail_activity_unlink_log: Rename to mail_activity_cancel_tracking

victoralmau:
Renamed to `mail_activity_cancel_tracking`

Related to https://github.com/OCA/social/pull/1599#issuecomment-2673963038

@Tecnativa

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -8,7 +8,8 @@ renamed_modules = {
     "note": "project_todo",
     "website_sale_delivery_mondialrelay": "website_sale_mondialrelay",
     # odoo/enterprise
-    # OCA/...
+    # OCA/social
+    "mail_activity_unlink_log": "mail_activity_cancel_tracking",
 }
 
 # Merged modules contain a mapping from old module names to other,


openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
```diff
@@ -10,6 +10,8 @@
     "mail.mail_channel_rule",
     "mail.channel_all_employees",
     "mail.channel_member_general_channel_for_admin",
+    "mail_activity_unlink_log.mt_activities_unlink",
+    "mail_activity_unlink_log.message_activity_unlink",
```
MiquelRForgeFlow:
```suggestion
    "mail_activity_cancel_tracking.mt_activities_unlink",
    "mail_activity_cancel_tracking.message_activity_unlink",
```

pedrobaeza:
It should be rename, not merge.

openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
MiquelRForgeFlow: Oh, then this should be added in mail_activity_cancel_tracking module instead, right?

pedrobaeza: Yes

MiquelRForgeFlow (approved):

pedrobaeza (approved):

openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
MiquelRForgeFlow: Done in https://github.com/OCA/social/pull/1602.

PR #4762, opened 2025-02-21
[17.0][OU-FIX] mail: set outgoing emails

MiquelRForgeFlow:
See https://github.com/odoo/odoo/pull/139814:

Selection_4337

### DIFF
--- a/openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
+++ b/openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
@@ -89,6 +89,25 @@ def _mail_template_convert_report_template_m2o_to_m2m(env):
     )
 
 
+def _fill_mail_message_outgoing(env):
+    # set outgoing messages
+    group = env.ref("base.group_user")
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        UPDATE mail_message mm
+        SET message_type = 'email_outgoing'
+        FROM mail_mail mail, res_partner rp
+        JOIN res_users ru ON ru.partner_id = rp.id
+        JOIN res_groups_users_rel rel ON rel.uid = ru.id
+            AND rel.gid = {group.id}
+        WHERE mm.message_type = 'email'
+        AND mm.message_id like '%-openerp-' || mm.res_id || '-' || mm.model || '@%'
+        AND (mm.author_id = rp.id OR mail.mail_message_id = mm.id)
+        """,
+    )
+
+
 def _mail_activity_plan_template(env):
     """If the OCA mail_activity_plan module was installed, we create the
     mail.activity.plan.template records.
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -120,7 +120,8 @@ mail         / mail.shortcode           / message_ids (many2one)        : DEL re
 # NOTHING TO DO: following commit https://github.com/odoo/odoo/pull/138938/commits/94b71f74ba20733110c123e1c02df9a384fe9d74
 
 mail         / mail.message             / message_type (selection)      : selection_keys is now '['auto_comment', 'comment', 'email', 'email_outgoing', 'notification', 'user_notification']' ('['comment', 'email', 'notification', 'user_notification']')
-# NOTHING TO DO: well when odoo backport https://github.com/odoo/odoo/pull/149576 in 15, nothing to do either because there is no way to detect and fill for 'auto_common' and 'email_outgoing'
+# NOTHING TO DO (auto_comment): well when odoo backport https://github.com/odoo/odoo/pull/149576 in 15, nothing to do because there is no way to detect and fill for 'auto_common'
```
pedrobaeza:
```suggestion
# NOTHING TO DO (auto_comment): well when odoo backport https://github.com/odoo/odoo/pull/149576 in 15, nothing to do because there is no way to detect and fill for 'auto_comment'
```

pedrobaeza (commented):
Not sure about the criteria for selecting outgoing messages, as replies will contain as well the message_id. What about having an internal user as sender?

MiquelRForgeFlow:
> What about having an internal user as sender?

I have put this as one option. The other option is if a related mail.mail record exists.

openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
```diff
@@ -89,6 +89,28 @@ def _mail_template_convert_report_template_m2o_to_m2m(env):
...
+        JOIN res_groups_users_rel rel ON rel.uid = ru.id
+            AND rel.gid IN (SELECT imd.res_id
+                            FROM ir_model_data imd
+                            WHERE imd.module='base'
+                                AND imd.name='group_user'
+                                AND imd.model='res.groups')
+        WHERE mm.message_type = 'email'
+        AND mm.message_id like '%-openerp-' || mm.res_id || '-' || mm.model || '@%'
+        AND (mm.author_id = rp.id OR mail.mail_message_id = mm.id)
+        """,
```
pedrobaeza:
```suggestion
    group = env.ref("base.group_user")
    openupgrade.logged_query(
        env.cr,
        f"""
        UPDATE mail_message mm
        SET message_type = 'email_outgoing'
        FROM mail_mail mail, res_partner rp
        JOIN res_users ru ON ru.partner_id = rp.id
        JOIN res_groups_users_rel rel ON rel.uid = ru.id
            AND rel.gid = {group.id}
        WHERE mm.message_type = 'email'
        AND mm.message_id like '%-openerp-' || mm.res_id || '-' || mm.model || '@%'
        AND (mm.author_id = rp.id OR mail.mail_message_id = mm.id)
        """,
```

pedrobaeza (approved):

PR #4766, opened 2025-02-25
[17.0][OU-ADD] base_vat: Migration to 17.0.

carlos-lopez-tecnativa:
TT49295
@Tecnativa @pedrobaeza @chienandalu @sergio-teruel could you please review this?


rvalyi:
the relevant commit in odoo/odoo if it helps https://github.com/akretion/odoo-module-diff-analysis/blob/main/17.0/base_vat/c000___________-_116391_imp-base-vat-vies-validation-on-fiscal-position.patch

openupgrade_scripts/scripts/base_vat/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,50 @@
...
+
+from openupgradelib import openupgrade
+
+
+def _partner_fill_vies_valid(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        WITH check_vies_enabled AS (
+            SELECT COUNT(*) AS count FROM res_company WHERE vat_check_vies = TRUE
```
MiquelRForgeFlow:
```suggestion
            SELECT COUNT(*) AS count FROM res_company WHERE vat_check_vies
```
or IS TRUE

openupgrade_scripts/scripts/base_vat/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,50 @@
...
+                    )
+                    IN (SELECT code FROM eu_countries)
+                )
+        )
+        UPDATE res_partner p
+        SET vies_valid = CASE
+            WHEN (SELECT count FROM check_vies_enabled) = 0 THEN FALSE
+            WHEN p.id IN (SELECT id FROM valid_partners) THEN TRUE
+            ELSE FALSE
+        END;
```
MiquelRForgeFlow:
```suggestion
        UPDATE res_partner p
        SET vies_valid = CASE
            WHEN (SELECT count FROM check_vies_enabled) = 0 THEN FALSE
            WHEN LENGTH(p.vat) > 1
                AND (
                    -- Extract VAT prefix
                    -- 1 or 2 characters according to the logic in _split_vat
                    LEFT(p.vat,
                        CASE WHEN SUBSTRING(p.vat, 2, 1) ~ '[A-Za-z]'
                        THEN 2 ELSE 1 END
                    ) IN (SELECT code FROM eu_countries)
                ) THEN TRUE
            ELSE FALSE
        END;
```
avoid WHEN p.id IN (SELECT id FROM valid_partners) as can be a big list

carlos-lopez-tecnativa: I applied your suggestions. Please update your review. Thanks.

MiquelRForgeFlow (approved):

PR #4592, opened 2024-10-18
[17.0][OU-ADD] payment_custom: Migration to 17.0

remi-filament:

PR #4587, opened 2024-10-16
[17.0][OU-ADD] account_edi: Migration to 17.0

acpMicrocom:

PR #4771, opened 2025-02-25
[17.0][OU-ADD] account_edi: Migration to 17

carlos-lopez-tecnativa:
TT49276
@Tecnativa @pedrobaeza


MiquelRForgeFlow:
@carlos-lopez-tecnativa could you add the commit in https://github.com/OCA/OpenUpgrade/pull/4587 before your commit? (this way, we can supersede that PR)

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa could you add the commit in #4587 before your commit? (this way, we can supersede that PR)

Hi @MiquelRForgeFlow, sorry, but I don't understand what commit or change I must add.
The `_sql_constraints` is not removed; it is present in V16:
**Edi format**
https://github.com/odoo/odoo/blob/f7db1775af3cb641f6cc88607b2d9dce611fb049/addons/account_edi/models/account_edi_format.py#L34
and V17:
https://github.com/odoo/odoo/blob/f4f7f74950b1a36eb0b4fb85529cc65816808879/addons/account_edi/models/account_edi_format.py#L29

**Edi document**
V16
https://github.com/odoo/odoo/blob/f7db1775af3cb641f6cc88607b2d9dce611fb049/addons/account_edi/models/account_edi_document.py#L45C14-L45C51
V17
https://github.com/odoo/odoo/blob/f4f7f74950b1a36eb0b4fb85529cc65816808879/addons/account_edi/models/account_edi_document.py#L44

I think the analysis is incorrect. In any case, a similar issue was reported as a comment here: https://github.com/OCA/OpenUpgrade/pull/4592#discussion_r1969971891

MiquelRForgeFlow:
Oh, right. The analysis file was updated in https://github.com/OCA/OpenUpgrade/commit/9bde91fbf9356f10077870e815292d8083e43f8c, and other PR is deprecated then.

MiquelRForgeFlow (approved):

pedrobaeza:
This is missing the deletion of the translations for the cron name loaded in noupdate.

MiquelRForgeFlow:
The "name" field of `ir.cron` model is not translatable.

pedrobaeza:
Uhm, that's not true...

imagen

MiquelRForgeFlow:
Ok, I see that it happens for ir.actions.actions. Then ir.action.server inherits from ir.actions.actions and ir.cron inherits from ir.action.server.

Selection_4351

pedrobaeza:
OK, but the end is the same: translations must be deleted for forcing the system on next language load to get the new terms. If not, you will see the name in English, and the rest of the languages will be with the old term.

@carlos-lopez-tecnativa there's an openupgradelib method for doing that `delete_translations_...`.

MiquelRForgeFlow:
Done in https://github.com/OCA/OpenUpgrade/pull/4783.

PR #4782, opened 2025-02-27
[17.0][OU-ADD] product_expiry: Nothing to do

CarlosRoca13:
cc @Tecnativa TT52341

ping @pedrobaeza @sergio-teruel


openupgrade_scripts/scripts/product_expiry/17.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,13 @@
+---Models in module 'product_expiry'---
+---Fields in module 'product_expiry'---
+product_expiry / stock.quant              / expiration_date (datetime)    : NEW isrelated: related, stored
+NOTHING TO DO: ORM already filled this field by SQL being a related stored with one level deep
```
MiquelRForgeFlow, line 4: I am not sure if it's better to fill it in pre-migration by SQL, what's your opinion @pedrobaeza ?

pedrobaeza: It's correct this way due to the explanation given. No gain doing it manually in pre-migraiton.

MiquelRForgeFlow (approved):

PR #4536, opened 2024-08-08
[15.0][OU-FIX] account: should not merge website_legal_page module

remi-filament:
Issue raised while migrating database with both account and website_legal_page installed.

website_legal_page is merged into account model with post-migration.py script since PR #3728 

The problem is that merging module in post-migration does not reload the list of modules to be updated, so when you run the command with -u all, it then tries to update website_legal_page which does not exist anymore, leading to CacheMiss exception. (example of error in the first comment on this PR to not overload the description here).

After giving it some thoughts, I do not think merging website_legal_page into account is a good solution, because in many cases legal page from that module are not the same as the one we want to link to from account move. (for instance in France, legal page is mandatory on every website with few information to identify owner, but the terms you will need on account move are related to the sale you performed).
Also, website_legal_page adds a link to legal page in website footer which is then lost with this merging (since account is not doing that).

Still it can make sense to initialize new invoice_terms_html with legal page content as long as the new mode is not forced (already fixed/reverted with PR #4110 )

Therefore, I propose here to not merge website_legal_page into account but keeping the initialization of the company invoice_terms_html field.

One would then need to uninstall website_legal_page manually if it is not needed anymore.

PR #4784, opened 2025-02-27
[15.0][OU-IMP] account: install again website_legal_page if installed

MiquelRForgeFlow:
An alternative (?) of https://github.com/OCA/OpenUpgrade/pull/4536.

### DIFF
--- a/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
@@ -73,6 +73,15 @@ def _handle_website_legal_page(env):
                 }
             )
             view_temp.unlink()
+    # let's install again the module for the extra needs
+    env["ir.module.module"].update_list()
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE ir_module_module
+        SET state='to install'
+        WHERE name = 'website_legal_page' AND state='uninstalled'""",
+    )
 
 
 @openupgrade.migrate()


pedrobaeza:
I still haven't found any problem with previous code, and this won't be correct at all, as you may not want to have the module re-installed, but merged.

MiquelRForgeFlow:
> you may not want to have the module re-installed, but merged.

Why? It gives you extra things.

pedrobaeza:
It's not needed for most of the uses.

MiquelRForgeFlow:
🤔 "most of" means that there are cases where it may be needed (although very few). Thus, it seems to me that it's preferable install the module, and let's people be able to uninstall when they desire than not installing and letting people with the option to install.

pedrobaeza:
No, you can install it in any moment manually. We should favor the most common migration path, which is to have it merged and forgotten.

MiquelRForgeFlow:
@remi-filament maybe you prefer something like this.

remi-filament:
Hi @MiquelRForgeFlow I am not sure about that one, although I still think we should keep website_legal_page.
Why should we merge it in account to reinstall it afterwards ?
If we go that way, it is probably simpler to not merge it at all no (and just copying the terms in invoice_terms_html) ?

The point here of keeping the module would be to keep the modified terms, I am not sure that by reinstalling it after merging would not reinitialize terms ?

MiquelRForgeFlow:
Ok, I close this 🤷‍♂️

But maybe the `update_list()` part solves the error you mentioned in the other PR 😉

PR #4671, opened 2024-11-21
[17.0][OU-ADD] mass_mailing: Migration to 17.0

acpMicrocom:

PR #4785, opened 2025-02-27
[17.0][OU-ADD] mass_mailing: Migration to 17.0.

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4671
TT49916

### DIFF
--- a/openupgrade_scripts/scripts/mass_mailing/17.0.2.7/post-migration.py
+++ b/openupgrade_scripts/scripts/mass_mailing/17.0.2.7/post-migration.py
@@ -1,0 +1,39 @@
+# Copyright 2025 Carlos Lopez - Tecnativa
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def _utm_campaign_fill_ab_testing_winner_mailing_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        WITH winner_mailing AS (
+            SELECT
+                campaign_id,
+                MAX(id) AS mailing_id
+            FROM
+                mailing_mailing
+            WHERE
+                ab_testing_enabled IS TRUE
+                AND state = 'done'
+                AND ab_testing_pc = 100
+            GROUP BY
+                campaign_id
+        )
+        UPDATE utm_campaign
+            SET ab_testing_winner_mailing_id = winner_mailing.mailing_id
+        FROM winner_mailing
+        WHERE utm_campaign.id = winner_mailing.campaign_id
+            AND utm_campaign.ab_testing_completed IS TRUE
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "mass_mailing", "17.0.2.7/noupdate_changes.xml")
+    openupgrade.delete_record_translations(
+        env.cr, "mass_mailing", ["mass_mailing_mail_layout"], field_list=["arch_db"]
+    )
+    _utm_campaign_fill_ab_testing_winner_mailing_id(env)
--- a/openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
@@ -1,0 +1,145 @@
+---Models in module 'mass_mailing'---
+obsolete model mailing.contact.subscription
+new model mailing.subscription
+# DONE: pre-migration: Renamed to mailing.subscription according https://github.com/odoo/odoo/commit/46ab1e38fcae2054a54bd0081528f3f05dd79e10
+new model mailing.subscription.optout
+# NOTHING TO DO: new feature
+
+---Fields in module 'mass_mailing'---
+mass_mailing / mail.blacklist           / opt_out_reason_id (many2one)  : NEW relation: mailing.subscription.optout
+# NOTHING TO DO: new feature
+mass_mailing / mailing.contact          / _order                        : _order is now 'name ASC, id DESC' ('email')
+# NOTHING TO DO:
+
+mass_mailing / mailing.contact          / list_ids (many2many)          : table is now 'mailing_subscription' ('mailing_contact_list_rel')
+mass_mailing / mailing.contact          / message_main_attachment_id (many2one): DEL relation: ir.attachment
+mass_mailing / mailing.contact          / rating_ids (one2many)         : NEW relation: rating.rating
+mass_mailing / mailing.contact          / subscription_ids (one2many)   : NEW relation: mailing.subscription
+mass_mailing / mailing.contact          / subscription_list_ids (one2many): DEL relation: mailing.contact.subscription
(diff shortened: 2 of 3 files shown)


MiquelRForgeFlow:
Could you maintain the commit in https://github.com/OCA/OpenUpgrade/pull/4671 and then add you a second commit with your improvements?

carlos-lopez-tecnativa:
> Could you maintain the commit in #4671 and then add you a second commit with your improvements?

Could you specify the relevant changes in this PR? I started from scratch because, from my point of view, these changes are incorrect. The author is dropping a SQL unique constraint on unique_contact_list, but this constraint is still present. So, if I preserve this change, I would need to revert it again in my commit. Additionally, most of the changes in the `upgrade_analysis_work.txt` file are marked as "NOTHING TO DO," whereas, in reality, there is still a lot of work to be done.

MiquelRForgeFlow:
Ok, I was asking for to make it easier the review process. With your way, I will need to compare the two PRs 🤷‍♂️

carlos-lopez-tecnativa:
> Ok, I was asking for to make it easier the review process. With your way, I will need to compare the two PRs 🤷‍♂️

Ok, sorry. I need to add "superseded" because the intention is to close the other PR without merging.

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+        FROM winner_mailing
+        WHERE utm_campaign.id = winner_mailing.campaign_id
+            AND utm_campaign.ab_testing_completed IS TRUE
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "mass_mailing", "17.0.2.7/noupdate_changes.xml")
```
MiquelRForgeFlow, line 35: delete translations of mass_mailing_mail_layout

MiquelRForgeFlow:
please rebase (and handle above comment)

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/post-migration.py
carlos-lopez-tecnativa: Done

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+mass_mailing / mailing.subscription.optout / sequence (integer)            : NEW hasdefault: default
+# NOTHING TO DO: new feature
+
+mass_mailing / mailing.trace            / failure_reason (text)         : NEW
+mass_mailing / mailing.trace            / failure_type (selection)      : selection_keys is now '['mail_bl', 'mail_bounce', 'mail_dup', 'mail_email_invalid', 'mail_email_missing', 'mail_from_invalid', 'mail_from_missing', 'mail_optout', 'mail_smtp', 'unknown']' ('['mail_bl', 'mail_dup', 'mail_email_invalid', 'mail_email_missing', 'mail_optout', 'mail_smtp', 'unknown']')
+mass_mailing / mailing.trace            / trace_status (selection)      : selection_keys is now '['bounce', 'cancel', 'error', 'open', 'outgoing', 'pending', 'process', 'reply', 'sent']' ('['bounce', 'cancel', 'error', 'open', 'outgoing', 'reply', 'sent']')
+# NOTHING TO DO: new feature
+
+mass_mailing / utm.campaign             / ab_testing_completed (boolean): now a function
+# NOTHING TO DO: now is compute with store=True
```
pedrobaeza: Initialize the column in pre and do a SQL for assigning the value on post after processing `ab_testing_winner_mailing_id`. If not, the ORM computation will take place before and this won't have a valid value.

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+mass_mailing / utm.campaign             / ab_testing_total_pc (integer) : DEL
+# NOTHING TO DO: Removed in https://github.com/odoo/odoo/commit/37348521ced0ae094962661d84994a147c9b884c
+
+mass_mailing / utm.campaign             / ab_testing_winner_mailing_id (many2one): NEW relation: mailing.mailing
+# DONE: post-migration: fill the field where mailing_mailing related to campain with ab_testing_pc=100
```
pedrobaeza: Specify that we can only deduce a winning A/B in the case only one is done, as the rest of the cases can't be deduced.

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+NEW ir.model.access: mass_mailing.access_mailing_subscription_optout_mm_user
+# NOTHING TO DO: new access right
+
+NEW ir.model.access: mass_mailing.access_mailing_subscription_mm_user
+DEL ir.model.access: mass_mailing.access_mailing_contact_subscription_mm_user
+# DONE: pre-migration: Rename xml-id
+
+NEW ir.model.constraint: mass_mailing.constraint_mailing_subscription_unique_contact_list
+DEL ir.model.constraint: mass_mailing.constraint_mailing_contact_list_rel_unique_contact_list
+# DONE: pre-migration: handle in rename_tables
```
pedrobaeza, line 92: But the constraint name is not renamed AFAIK.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
carlos-lopez-tecnativa: This field already exist in V16 https://github.com/odoo/odoo/blob/34947c01623a9fc753197bd55d6121d9c6fa682f/addons/mass_mailing/models/utm_campaign.py#L25, so recomputation is not triggered. Why is it necessary to initialize the column? Please clarify.

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Done

openupgrade_scripts/scripts/mass_mailing/17.0.2.7/upgrade_analysis_work.txt
carlos-lopez-tecnativa: The `rename_tables` function handles this, please see https://github.com/OCA/openupgradelib/blob/73f7b8fda0f5102f7bd58df831b39e4e1c5a180c/openupgradelib/openupgrade.py#L885-L903 
Are you referring to a different renaming process? Please clarify.

pedrobaeza: OK, didn't remember.

pedrobaeza (approved):

PR #4792, opened 2025-02-28
[17.0][OU-FIX] account: Rename XML-IDs for installed CoA elements

pedrobaeza:
On v17, when you load a CoA into a company, the CoA elements are still given an XML-ID with the company ID + `_` + the original template XML-ID, but now, instead of putting the module containing the template, all of them are put with the module `account`.

Thus, we need to rename the module for all the existing CoA elements XML-IDs with this pattern.

Reference: https://github.com/odoo/odoo/blob/b9abe46c1492b09e369434e76ec8196c6b02dd19/addons/account/models/chart_template.py#L608

@Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
@@ -518,6 +518,36 @@ def _map_chart_template_id_to_chart_template(
     )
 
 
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):
LGTM

christian-ramos-tecnativa (approved):

PR #4796, opened 2025-03-03
[17.0][OU-ADD] exclude analytic 1.2 migration

hbrunn:
[this](https://github.com/odoo/odoo/blob/17.0/addons/analytic/migrations/1.2/pre-migrate.py) migration script will fail in an openupgrade context because at the point it is run, the column to be indexed doesn't exist yet.

Openupgrade creates the column and the correct index at a later point in the post migration script.


MiquelRForgeFlow (approved):

pedrobaeza (approved):
Thanks for looking for a solution, Holger.

PR #4564, opened 2024-09-27
[17.0][OU-ADD] hr_attendance: Nothing to do

acpMicrocom:

PR #4801, opened 2025-03-06
[17.0][OU-ADD] hr_attendance

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4564.


openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
```diff
@@ -234,15 +234,17 @@ def _pre_create_account_report_active(env):
...
         """
         ALTER TABLE account_report ADD COLUMN active boolean DEFAULT true
-        """
+        """,
     )
-    env.cr.execute(
+    openupgrade.logged_query(
+        env.cr,
         """
         ALTER TABLE account_report ALTER COLUMN active SET DEFAULT false
```
pedrobaeza:
```suggestion
        ALTER TABLE account_report ALTER COLUMN active DROP DEFAULT
```

or you can use `add_fields` from openupgradelib

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
```diff
@@ -0,0 +1,32 @@
...
+
+def fill_res_company_attendance_kiosk_key(env):
+    companies = env["res.company"].search([])
+    for company in companies:
+        company.attendance_kiosk_key = uuid.uuid4().hex
+
+
+def fill_hr_attendance_overtime_hours(env):
+    attendances = env["hr.attendance"].search([])
+    attendances._compute_overtime_hours()
```
pedrobaeza, line 52: Have you tracked times for this? And what is the difference on pre-creating the column and then launch this instead of letting the ORM to do it on the update process?

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,106 @@
...
+hr_attendance / hr.employee              / attendance_manager_id (many2one): NEW relation: res.users
+# NOTHING TO DO
+
+hr_attendance / res.company              / attendance_from_systray (boolean): NEW hasdefault: default
+# DONE: pre-migration: fill default values
+
+hr_attendance / res.company              / attendance_kiosk_key (char)   : NEW hasdefault: default
+# DONE: post-migration: fill default values
+
+hr_attendance / res.company              / attendance_kiosk_use_pin (boolean): NEW
```
pedrobaeza, line 39: Shouldn't you need to transfer this from the previous system parameter? (the same for the following).

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,106 @@
...
+
+NEW ir.rule: hr_attendance.hr_attendance_rule_attendance_admin (noupdate)
+NEW ir.rule: hr_attendance.hr_attendance_rule_attendance_manager_restrict (noupdate)
+NEW ir.rule: hr_attendance.hr_attendance_rule_attendance_officer_overtime_restrict (noupdate)
+NEW ir.rule: hr_attendance.hr_attendance_rule_attendance_overtime_admin (noupdate)
+NEW ir.rule: hr_attendance.hr_attendance_rule_attendance_overtime_simple_user (noupdate)
+NEW ir.rule: hr_attendance.hr_attendance_rule_attendance_simple_user (noupdate)
+# NOTHING TO DO
+
+DEL ir.rule: hr_attendance.hr_attendance_report_rule_multi_company (noupdate)
```
pedrobaeza, line 75: Aren't some of these rules the equivalent of the new ones?

pedrobaeza (changes requested):
And please remove extra commits that don't belong here.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
MiquelRForgeFlow: `_compute_overtime_hours` does a query at the beginning. If I don't do nothing, it will do this for every record. By doing my way, only is executed once.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Maybe. But why rename them when we can add new and deleted old ones? The result is the same.

openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
MiquelRForgeFlow: I want to avoid the part of ir_model_fields and ir_model_data inside `add_fields`

pedrobaeza: Why? The ORM is doing later the same.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
pedrobaeza: OK. Please explain it on the work file.

openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
MiquelRForgeFlow: Exactly. Thus, why bother?

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
pedrobaeza: If users have modified the rules, they will lose that customizations. It's an edge case, but for being the most conservative possible, and only if they are not rewritten via noupdate changes of course.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Which system parameters? I couldn't find them.

openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
pedrobaeza: For not having 8 lines of code where you can have 1.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
pedrobaeza: Maybe they are other thing. In the previous version, you were able to do the same, so I suppose they were stored at system level instead of company level.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: The problem is that they are noupdate. Thus, if I rename, I should manually add a noupdate_changes file the changes of this rules.

pedrobaeza: But they have changed?

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Ok, I found it.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Yes.

pedrobaeza: OK, then it doesn't worth.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Done :)

pedrobaeza (approved):

PR #4803, opened 2025-03-07
[16.0][OU-IMP] mrp: Fill origin of stock.moves that has a production related

CarlosRoca13:
In v15, there was an issue with the origin field when adding a new stock move to a production order after it was completed. This script ensures that the origin field, which was not filled in the previous version, is now properly populated in this new one.

cc @Tecnativa TT55383

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/scripts/mrp/16.0.2.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/mrp/16.0.2.0/pre-migration.py
@@ -31,7 +31,30 @@ def _fill_stock_move_manual_consumption(env):
     )
 
 
+def _fill_origin_not_set_on_stock_move(env):
+    """In v15, there was an issue with the origin field when adding
+    a new stock move to a production order after it was completed.
+    This script ensures that the origin field, which was not filled
+    in the previous version, is now properly populated in this new
+    one.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move sm
+        SET origin = mrp.name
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/mrp/16.0.2.0/pre-migration.py
```diff
@@ -31,7 +31,38 @@ def _fill_stock_move_manual_consumption(env):
...
+        env.cr,
+        """
+        UPDATE stock_move sm
+        SET origin = mrp.name
+        FROM mrp_production mrp
+        WHERE sm.state = 'done'
+            AND sm.origin IS NULL
+            AND (
+                (
+                    sm.raw_material_production_id IS NOT NULL
```
pedrobaeza: Not needed this and the following one "IS NOT NULL"

openupgrade_scripts/scripts/mrp/16.0.2.0/pre-migration.py
```diff
@@ -31,7 +31,31 @@ def _fill_stock_move_manual_consumption(env):
...
+    in the previous version, is now properly populated in this new
+    one.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move sm
+        SET origin = mrp.name
+        FROM mrp_production mrp
+        WHERE sm.state = 'done'
```
pedrobaeza: why only this state? It should be for all IMO.

pedrobaeza (approved):

PR #4804, opened 2025-03-09
[17.0][OU-FIX] product: Fixing table name ir_model_data

mgomezleon:


openupgrade_scripts/scripts/product/17.0.1.2/post-migration.py
```diff
@@ -46,7 +46,7 @@ def _remove_if_applicable_base_pricelist(env):
             openupgrade.delete_records_safely_by_xml_id(env, ["product.list0"])
             return
     openupgrade.logged_query(
-        env.cr, "DELETE FROM ir_model_date WHERE module = 'product' AND name='list0'"
+        env.cr, "DELETE FROM ir_model_data WHERE module = 'product' AND name='list0'"
```
legalsylvain, line 49: Thanks for the patch.

The fonction _remove_if_applicable_base_pricelist is not called. 

I guess it should, dont you think ?

legalsylvain (changes requested):

openupgrade_scripts/scripts/product/17.0.1.2/post-migration.py
mgomezleon: I think we should run it! 
In my database, I did it manually.

pedrobaeza:
Please rewrite the commit message to `[OU-FIX] product: ir_model_data table typo` and include the call to the method.

pedrobaeza:
You have introduced a merge commit, doing a pull instead of a rebase. Please do the rebase, and squash both commits into one.

legalsylvain (approved):

pedrobaeza (changes requested):
Setting this until the commits are fixed.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #4805, opened 2025-03-10
[17.0][OU-ADD] account_invoice_fiscal_position_update: Merged into account

juancarlosonate-tecnativa:
Merged into account: https://github.com/OCA/account-invoicing/pull/1935#issuecomment-2704882999

@Tecnativa

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -37,6 +37,8 @@ merged_modules = {
     "website_sale_delivery": "website_sale",
     "website_sale_loyalty_delivery": "website_sale_loyalty",
     "website_sale_stock_product_configurator": "website_sale_product_configurator",
+    # OCA/account-invoicing
+    "account_invoice_fiscal_position_update": "account",
     # OCA/l10n-spain
     "l10n_es_irnr": "l10n_es",
     "l10n_es_irnr_sii": "l10n_es_aeat_sii_oca",


legalsylvain (changes requested):
Hi.

I'm not agree that ``account_invoice_fiscal_position_update`` feature is covered by ``account`` native module.
- ``account`` module add a button and user has to click on it manually. As a result, taxes can be wrong, if user doesn't click on it.
- ``account_invoice_fiscal_position_update`` changes taxes onchange(fiscal_position). As a result, taxes will be right without any actions.

CC : @StefanRijnhart, @pedrobaeza

Thanks !

pedrobaeza:
Indeed the behavior is not exactly the same, but it's better to just indicate and let the user decides. Most users will accept this "change". And for those that not, an extra module can be developed on top doing it automatically, but reusing the existing code and not doing the refresh on its own IMO.

legalsylvain:
> And for those that not, an extra module can be developed on top doing it automatically

Sorry, I didn't understood. what could be the name and the feature of this extra module ?

pedrobaeza:
Hehe, indeed that extra module may be very similar on the name, but it should be different for letting user choose, and to not be "enabled" by default. For example,  `account_invoice_fiscal_position_update_auto_apply` or `account_invoice_fiscal_position_update_auto` or similar. And in the code, it should be just the call to the method of the button.

juancarlosonate-tecnativa:
Is it really necessary to make this logic automatic again? Right now, there's a button that performs the action, which means it's one less thing we have to maintain. Is it worth creating a module to make this automatic? How often do fiscal changes occur to the point that the user would 'waste so much time pressing a button'? What’s the real benefit of this? In my opinion, it seems unnecessary.

legalsylvain:
Well, as soon as I can develop a new module named `` account_invoice_fiscal_position_update_auto_apply``, that's OK for me !

> How often do fiscal changes occur to the point that the user would 'waste so much time pressing a button' ?

that's really not about time wasted, it's just about to avoid accounting misconfiguration, if user avoid to click on "update taxes".


> but it should be different for letting user choose

I don't figure out the use case where we want to change the fiscal position, but keep taxes unchanged. If some people can elaborate ? 


Note 2 : How do you handle the following use case  ? 
- change fiscal position
- the button appears
- click on save (or go out and come back on invoice)
- the button disappeared
So : 
- user doesn't know that taxes are incorrect.
- to apply the correct taxes, we have to change fiscal position twice. that's not really intuitive.

StefanRijnhart (approved):
Approving, given that any follow-up in the OCA will be under a different module name. For the record, for the purposes of the project I'm currently working on, the standard Odoo functionality suffices.

legalsylvain (approved):

pedrobaeza:
> I don't figure out the use case where we want to change the fiscal position, but keep taxes unchanged. If some people can elaborate ?

There are some complex cases where you switch fiscal positions for applying certain taxes to the lines that you are introducing in that moment, then switch to another FP, and continue introducing other lines. It happens in Spain with certain cases.

> click on save (or go out and come back on invoice)
> the button disappeared

The new always edit mode makes the save something that few people do it in a way that would miss the button click, although it's something to take into account indeed.

legalsylvain:
> There are some complex cases where you switch fiscal positions for applying certain taxes to the lines that you are introducing in that moment, then switch to another FP, and continue introducing other lines. It happens in Spain with certain cases.

Thanks ! (Nothing similar in France, IFaik.)

PR #4806, opened 2025-03-10
[17.0][FIX] account: copy imd entry without ir.model.data's random name mechanism

hbrunn:
this avoids the [override](https://github.com/OCA/OCB/blob/17.0/odoo/addons/base/models/ir_model.py#L2231) that causes tests to fail randomly in docker container, probably some entropy issue

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
@@ -3,6 +3,8 @@
 
 from openupgradelib import openupgrade
 
+from odoo import models
+
 from odoo.addons.base.models.ir_property import TYPE2FIELD as ir_property_TYPE2FIELD
 
 _deleted_xml_records = [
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
```diff
@@ -418,8 +420,13 @@ def _account_tax_group_migration(env):
             )
 
             if tax_group_name:
-                new_imd = imd.copy({"res_id": new_tax_group.id})
-                new_imd.write({"name": f"{company_id}_{tax_group_name}"})
+                models.BaseModel.copy(
+                    imd,
```
MiquelRForgeFlow, line 424:
```suggestion
                imd.copy(
```
Why not this?

hbrunn: because the override linked above overwrites the name with a randomly generated one, and the write in the first version only happens when the transaction is flushed. When the next create generates a new name, there doesn't seem to be enough randomness in the container to create a new name

MiquelRForgeFlow (approved):

PR #4807, opened 2025-03-11
[16.0][OU-ADD] web_chatter_position: Ensure correct values for 'chatter_position'

baguenth:
Field 'chatter_position' was defined in module 'web_responsive' before it was moved to 'web_chatter_position'. 
In 'web_responsive' the allowed values have been ['normal', 'sided'] while the allowed values in 'web_chatter_position' are ['auto', 'bottom', 'sided']

In case users installed 'web_chatter_position' along with 'web_responsive', field 'chatter_position' potentially was already in the db with old values, which are not supported.

This migration covers this scenario by correcting unsupported values.

@pedrobaeza  Kindly asking for a review


MiquelRForgeFlow:
This is not needed. With https://github.com/OCA/web/pull/3118 is enough.

PR #4651, opened 2024-11-05
[17.0][OU-ADD] account_edi_ubl_cii: Migrate to 17.0

mgomezleon:

PR #4812, opened 2025-03-12
[17.0][OU-ADD] account_edi_ubl_cii: Migrate to 17.0

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4651.


openupgrade_scripts/scripts/account_edi_ubl_cii/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+def _res_partner_compute_fields_values(env):
+    res_partner = env["res.partner"].with_context(active_test=False).search([])
+    res_partner._compute_ubl_cii_format()
```
pedrobaeza: Shouldn't this be done on end-migration after all modules has been loaded?

MiquelRForgeFlow: Interesting. Yeah, also the other methods too, because they use `_deduce_country_code` which is also modified in several localization modules.

Thanks, done.

pedrobaeza: Yeah, I mean all of them.

pedrobaeza (approved):
/ocabot merge nobump

pedrobaeza:
I have canceled mine.

PR #4821, opened 2025-03-13
[17.0][MIG] website

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4562.


openupgrade_scripts/scripts/website/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 Viindoo Technology Joint Stock Company (Viindoo)
+# Copyright 2025 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "website", "17.0.1.0/noupdate_changes.xml")
```
pedrobaeza, line 10: @MiquelRForgeFlow I think we shouldn't overwrite that texts in noupdate, as they are very generic and may be updated by users concerned by SEO, specially the homepage one.

MiquelRForgeFlow: I though about it too, but I didn't have strong arguments. But voicing your opinion helps. Let's avoid the overwrite then.

pedrobaeza: Yes, nobody wants an un-personalized homepage description saying `This is the homepage of the website`, so it will be for sure changed, and putting this generic text silently is not good.

MiquelRForgeFlow: Done in https://github.com/OCA/OpenUpgrade/pull/4825.

PR #4827, opened 2025-03-15
[17.0][OU-FIX] website_event: No XML-ID renaming needed

pedrobaeza:
The merge of website_event_questions into website_event already performs the XML-IDs renaming, and if you really see the renamings, the source and target are the same, so no real renaming in the script.

@Tecnativa


pedrobaeza:
FYI @pilarvargas-tecnativa @MiquelRForgeFlow

MiquelRForgeFlow:
> the source and target are the same, so no real renaming in the script.

Oh, I didn't see that haha

pilarvargas-tecnativa:
> Oh, I didn't see that haha

It seems that this time 4 eyes didn't see more than 2 :face_with_diagonal_mouth:

MiquelRForgeFlow:
Then 6 eyes see more than 4.

PR #4836, opened 2025-03-17
[17.0][MIG] event_sale

MiquelRForgeFlow:
.


hbrunn (approved):
I don't think precomputing is necessary

openupgrade_scripts/scripts/event_sale/17.0.1.3/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,21 @@
+---Models in module 'event_sale'---
+---Fields in module 'event_sale'---
+event_sale   / event.event.ticket       / _order                        : _order is now 'event_id, sequence, price, name, id' ('event_id, price')
+# NOTHING TO DO
+
+event_sale   / event.event.ticket       / price_incl (float)            : NEW hasdefault: compute
+# NOTHING TO DO: not sure if needed to precreate it.
```
pedrobaeza, line 7: If you have a lot of events, this will be heavy, but let's keep it for now until someone feels is too slow.

openupgrade_scripts/scripts/event_sale/17.0.1.3/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,21 @@
...
+
+event_sale   / event.registration       / is_paid (boolean)             : DEL
+event_sale   / event.registration       / sale_status (selection)       : NEW selection_keys: ['free', 'sold', 'to_pay'], isfunction: function, stored
```
pedrobaeza, line 10: This one is a bit more critical, as the number of registrations for sure will be high, so you have to pre-create it and only assign when there are payed events.

openupgrade_scripts/scripts/event_sale/17.0.1.3/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,21 @@
...
+event_sale   / event.registration       / state (False)                 : NEW mode: modify, hasdefault: compute
+# NOTHING TO DO: sale_status (before was payment_status) is now stored, but is computed. not sure if needed to precreate it.
+
+event_sale   / product.template         / detailed_type (False)         : selection_keys is now '['combo', 'consu', 'event', 'product', 'service']' ('['consu', 'event', 'product', 'service']')
+event_sale   / sale.order.line          / registration_ids (one2many)   : NEW relation: event.registration
+# NOTHING TO DO
```
pedrobaeza, line 16: You are still not giving enough reasons why is nothing to do, and joining fields together that has nothing to do one with the other.

pedrobaeza (changes requested):
Check my comments and at least one field should be pre-filled.

pedrobaeza:
Superseded by #4936 attending the comments I put.

PR #4837, opened 2025-03-17
[15.0][OU-IMP] account: Improve amount_total_in_currency_query

etobella:
This query should be faster, as we are only updating the required records.


pedrobaeza:
@etobella the tag should be `[OU-IMP]`. Don't rip out the second (and important) part of it.

pedrobaeza (changes requested):
You are not writing the value in the rest of entries, so it doesn't seem correct.

etobella:
You are completely right. I assumed that the we were rewriting the field, not filling a new field. Sorry for the misunderstanding and the error on the commit name :sob: 

Thanks for reviewing it.

PR #4832, opened 2025-03-17
[18.0][OU-ADD] mail

hbrunn:
depends on:

- [x] #4709
- [x] #4740
- [x] #4742
- [x] #4743
- [x] #4741
- [x] #4744

PR #4740, opened 2025-02-17
[18.0][OU-ADD] base_setup

hbrunn:
depends on

- [x] #4709
- [x] #4741

PR #4839, opened 2025-03-17
[18.0][OU-ADD] product

hbrunn:
depends on:

- [x] #4709
- [x] #4740
- [x] #4742
- [x] #4743
- [x] #4741
- [x] #4744
- [x] #4832
- [x] #4876


hbrunn:
/ocabot migration product

<dependency></dependency>

MiquelRForgeFlow:
Depends on module `uom`.

hbrunn:
> Depends on module `uom`.

[No DB layout changes](https://github.com/OCA/OpenUpgrade/pull/4839/files#diff-720ddfab388b3d6aaeec9924b155ca52dd0ceeab92efc99415332d917f088bacR1027)

rvalyi:
FYI the 2 main change commits in product for 18.0 https://github.com/akretion/odoo-module-diff-analysis/tree/main/18.0/product

- https://github.com/odoo/odoo/pull/166497
- https://github.com/odoo/odoo/pull/159746

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,100 @@
...
+new model product.combo.item
+
+# NOTHING TO DO: moved in from pos
+
+new model update.product.attribute.value [transient]
+
+# NOTHING TO DO: new functionality
+
+---Fields in module 'product'---
+product      / product.attribute        / active (boolean)              : NEW hasdefault: default
```
MiquelRForgeFlow, line 10: better to autocreate column with default in pre-migration as in https://github.com/OCA/OpenUpgrade/blob/17.0/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py#L233

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,100 @@
...
+product      / product.attribute        / template_value_ids (one2many) : NEW relation: product.template.attribute.value
+product      / product.attribute.value  / active (boolean)              : NEW hasdefault: default
```
MiquelRForgeFlow: same here

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,100 @@
...
+
+# NOTHING TO DO
+
+product      / product.template         / is_favorite (boolean)         : NEW
+product      / product.template         / priority (selection)          : DEL selection_keys: ['0', '1']
+
+# DONE: set is_favorite flag based on priority
+
+product      / product.template         / service_tracking (selection)  : previously in module sale_project
+product      / product.template         / type (selection)              : now required
```
MiquelRForgeFlow, line 76: I am sure this doesn't happen, but maybe just in case, we should fill if empty using detailed_type

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,100 @@
...
+product      / product.document         / sequence (integer)            : NEW hasdefault: default
+
+# DONE: create sequence matching name
+
+product      / product.pricelist        / _order                        : _order is now 'sequence, id, name' ('sequence asc, id asc')
+
+# NOTHING TO DO
+
+product      / product.pricelist        / discount_policy (selection)   : DEL required, selection_keys: ['with_discount', 'without_discount']
+product      / product.pricelist.item   / display_applied_on (selection): NEW required, selection_keys: ['1_product', '2_product_category'], hasdefault: default
```
MiquelRForgeFlow, line 41: I have a BD with 342375 pricelist items. not sure if we should prefill this too.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,100 @@
+---Models in module 'product'---
+new model product.combo
+new model product.combo.item
+
+# NOTHING TO DO: moved in from pos
```
MiquelRForgeFlow: why not rename `pos.combo` to `product.combo` and `pos.combo.line` to `product.combo.item`?

hbrunn: I didn't want to dive into this yet as we don't need it for pure product migration. Left a note that this should be revisited when when doing the pos migration

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
hbrunn: well measure it? In my experience, the impact is negligible for indexed columns on less than millions of rows tables

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,50 @@
...
+]
+
+column_creates = [
+    ("product_attribute", "active", "boolean", "TRUE"),
+    ("product_attribute_value", "active", "boolean", "TRUE"),
+    ("product_pricelist_item", "display_applied_on", "varchar", "'1_product'"),
+]
+
+
+def create_columns(env, column_spec):
```
hbrunn: probably makes sense to add this to openupgradelib if we want to precreate all columns with default values?

MiquelRForgeFlow: 🤷‍♂️ Well, it already exists `add_fields`. But it is a more complex method, which creates not only the column but the field too, and if it isn't necessary, I prefer to avoid it in this kind of cases, as they are added with an OU description.

MiquelRForgeFlow: - https://github.com/OCA/openupgradelib/pull/412

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+
+product      / product.pricelist        / discount_policy (selection)   : DEL required, selection_keys: ['with_discount', 'without_discount']
+product      / product.pricelist.item   / display_applied_on (selection): NEW required, selection_keys: ['1_product', '2_product_category'], hasdefault: default
+
+# DONE: added in pre-migration
+
+product      / product.pricelist.item   / price_markup (float)          : NEW isfunction: function, stored
+product      / product.product          / _order                        : _order is now 'is_favorite desc, default_code, name, id' ('priority desc, default_code, name, id')
+product      / product.product          / combo_ids (many2many)         : previously in module point_of_sale
+product      / product.product          / service_tracking (selection)  : previously in module sale_project
```
MiquelRForgeFlow: service_tracking now is required (see in sale_project analysis file). Not sure if we should assure somehow this field is filled or delete those that have it empty.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
+---Models in module 'product'---
+new model product.combo
+new model product.combo.item
```
MiquelRForgeFlow, line 3: add a check in pre-migration, like:

`if openupgrade.table_exists(env.cr, "pos.combo"):`

then, add the model and table rename for both (`pos.combo` -> `product.combo` and `pos.combo.line` -> `product.combo.item`).

A part from doing this, I don't think anything else is necessary, as new fields are new features, computes or related. You can check analysis file if you do a rebase.

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
MiquelRForgeFlow: Is it already merged :)

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
hbrunn: it had a default in v17 too, so I don't think that's necessary

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
hbrunn: I intended to do this when doing looking into point_of_sale, but now looked at those models anyways. We have to rename combo_price to extra_price in addition to the above.

Difficult to test though as it is now.

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
hbrunn: and used here now

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,53 @@
...
+
+def rename_pos_models(env):
+    """
+    pos.combo and pos.combo.line have been moved to product from point_of_sale and
+    renamed to product.combo and product.combo.item respectively
+    """
+    if not openupgrade.table_exists(
+        env.cr, "pos_combo"
+    ) or not openupgrade.table_exists(env.cr, "pos_combo_line"):
+        return
```
MiquelRForgeFlow:
```suggestion
    if not openupgrade.table_exists(
        env.cr, "pos_combo"
    ):
        return
```
If they have combo, they will have combo line. And if they have combo line, they will have combo.

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,53 @@
...
+            ("pos.combo", "product.combo"),
+            ("pos.combo.line", "product.combo.item"),
+        ],
+    )
+    openupgrade.rename_fields(
+        env,
+        [
+            ("product.combo.item", "product_combo_item", "combo_price", "extra_price"),
+        ],
+    )
```
MiquelRForgeFlow, line 46:
```python
openupgrade.rename_fields(
        env,
        [
            ("product.combo", "product_combo", "combo_line_ids", "combo_item_ids"),
        ],
    )
```
to add

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
hbrunn: done

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
hbrunn: done

MiquelRForgeFlow (approved):
LGTM

@pedrobaeza do you plan to do a final review?

openupgrade_scripts/scripts/product/18.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,52 @@
...
+from openupgradelib import openupgrade
+
+xmlid_renames = [
+    ("product.group_discount_per_so_line", "sale.group_discount_per_so_line"),
+]
+
+column_creates = [
+    ("product.attribute", "active", "boolean", "TRUE"),
+    ("product.attribute.value", "active", "boolean", "TRUE"),
+    ("product.pricelist.item", "display_applied_on", "varchar", "'1_product'"),
```
pedrobaeza:
```suggestion
    ("product.pricelist.item", "display_applied_on", "char", "'1_product'"),
```

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+
+# DONE: rename models and fields from point_of_sale models if installed
```
pedrobaeza: All these references miss the usual format `# DONE: pre-migration:` or `# DONE: post-migration:` to indicate where this is done.

And we don't usually put an empty line in between for being more easily linkable to the modifications block.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+product      / product.combo.item       / product_id (many2one)         : NEW relation: product.product, required
+product      / product.document         / _order                        : _order is now 'sequence, name' ('name')
+
+# NOTHING TO DO
+
+product      / product.document         / sequence (integer)            : NEW hasdefault: default
+
+# DONE: create sequence matching name
+
+product      / product.pricelist        / _order                        : _order is now 'sequence, id, name' ('sequence asc, id asc')
```
pedrobaeza, line 37: This is changing the pricelist order, so we should assign new `sequence` values to preserve the previous one.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+
+# NOTHING TO DO
+
+product      / product.pricelist        / discount_policy (selection)   : DEL required, selection_keys: ['with_discount', 'without_discount']
+product      / product.pricelist.item   / display_applied_on (selection): NEW required, selection_keys: ['1_product', '2_product_category'], hasdefault: default
```
pedrobaeza, line 41: This should be adjusted according `applied_on` if I'm not wrong.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+
+# DONE: added in pre-migration
+
+product      / product.pricelist.item   / price_markup (float)          : NEW isfunction: function, stored
```
pedrobaeza, line 42: Please pre-create this in pre and fill it with `-price_discount`:

https://github.com/odoo/odoo/blob/0d396533336f4bcb55a18c961f81c1fbf80e0bf4/addons/product/models/product_pricelist_item.py#L237

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+product      / product.product          / _order                        : _order is now 'is_favorite desc, default_code, name, id' ('priority desc, default_code, name, id')
```
pedrobaeza, line 45: Specify that is nothing to do because is a related field to template, and this is done in template.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+# NOTHING TO DO
+
+product      / product.product          / standard_price (float)        : needs conversion to v18-style company dependent
+
+# DONE
+
+product      / product.product          / type (selection)              : now required
+product      / product.tag              / _order                        : _order is now 'sequence, id' ('id')
+product      / product.tag              / sequence (integer)            : NEW hasdefault: default
+product      / product.template         / _order                        : _order is now 'is_favorite desc, name' ('priority desc, name')
```
pedrobaeza, line 62: This can be reorder with the following block, as it's related.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+product      / product.template         / combo_ids (many2many)         : previously in module point_of_sale
+product      / product.template         / detailed_type (selection)     : DEL required, selection_keys: ['consu', 'service']
```
pedrobaeza: Here it should be mentioned that this will require some adaptation on subsequent modules, like `event`.

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,109 @@
...
+
+product      / product.template         / service_tracking (selection)  : previously in module sale_project
+product      / product.template         / type (selection)              : now required
+product      / product.template         / type (selection)              : selection_keys is now '['combo', 'consu', 'service']' ('['consu', 'service']')
+
+# NOTHING TO DO
+
+product      / res.partner              / specific_property_product_pricelist (many2one): NEW relation: product.pricelist
+
+# DONE
```
pedrobaeza: What is done here?

openupgrade_scripts/scripts/product/tests/data.py
```diff
@@ -0,0 +1,27 @@
+env = locals().get("env")
+# set specific pricelist
+pricelist_main = env["product.pricelist"].create(
+    {
+        "name": "Pricelist for main company",
+    }
```
pedrobaeza: Removing trailing comma the auto-formatter saves you a lot of lines:

```suggestion
    {"name": "Pricelist for main company"}
```

pedrobaeza (changes requested):

openupgrade_scripts/scripts/product/18.0.1.2/upgrade_analysis_work.txt
hbrunn: I don't see how this change impacts order out of joins given asc is the default and ids are unique?

pedrobaeza: OK, you are right. They remove the explicit word. That's something you can put also in the `NOTHING TO DO` comment. `name` is totally useless, as ID is always going to be unique.

hbrunn: done, the others too

pedrobaeza (approved):

PR #4840, opened 2025-03-17
[15.0][OU-FIX] mail: channel_all_employees doesn't need to be modified

etobella:
We don't need to change this as the change between 14 and 15 is the same (they changed the nomenclature but it is technically the same).

https://github.com/odoo/odoo/blob/14.0/addons/mail/data/mail_channel_data.xml#L29-L31
https://github.com/odoo/odoo/blob/15.0/addons/mail/data/mail_channel_data.xml#L28-L30

### DIFF
--- a/openupgrade_scripts/scripts/mail/15.0.1.5/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/mail/15.0.1.5/noupdate_changes.xml
@@ -9,22 +9,32 @@
   <!--<record id="base.partner_root" model="res.partner">
     <field name="email"/>
   </record>-->
+  <!--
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

MiquelRForgeFlow:
Have you checked the other entries in the file?

etobella:
No, but I will do it today, :wink:

etobella:
@MiquelRForgeFlow there were other records that didn't need a change

MiquelRForgeFlow:
Thanks.

pedrobaeza (approved):

PR #4841, opened 2025-03-17
[18.0][OU-ADD] analytic

hbrunn:
depends on:

- [x] #4709
- [x] #4740
- [x] #4742
- [x] #4743
- [x] #4741
- [x] #4744
- [x] #4832


hbrunn:
/ocabot migration analytic

<dependency></dependency>

rvalyi:
FYI The main change commit for v18: https://github.com/odoo/odoo/pull/156488

openupgrade_scripts/scripts/analytic/18.0.1.1/post-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade, openupgrade_180
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "analytic", "18.0.1.1/noupdate_changes.xml")
```
MiquelRForgeFlow: please, rebase and move everything to 18.0.1.2

hbrunn: will have to check how this interacts with the exclusion of that migration from upstream

hbrunn: done

openupgrade_scripts/scripts/analytic/18.0.1.2/post-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade, openupgrade_180
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "analytic", "18.0.1.2/noupdate_changes.xml")
```
MiquelRForgeFlow, line 9:
```suggestion
    openupgrade.load_data(env, "analytic", "18.0.1.2/noupdate_changes.xml", mode="init_no_create")
```
This is because now the record has `forcecreate="0"`

hbrunn: but that's fine. we don't want to create it if it doesn't exist, just rename it if it does

MiquelRForgeFlow (approved):

PR #4851, opened 2025-03-18
[17.0][MIG] account_payment_term

MiquelRForgeFlow:
.


openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
```diff
@@ -488,9 +488,8 @@ def _force_install_account_payment_term_module_module(env):
             env.cr,
             """
             UPDATE account_payment_term_line
-            SET
-            nb_days = nb_days - days_after,
-            days_next_month = days_after
+            SET nb_days = nb_days - days_after,
```
pedrobaeza, line 491: Please don't introduce random changes on PRs with other scope.

PR #4852, opened 2025-03-18
[17.0][OU-IMP] *: use of "IF NOT EXISTS"

MiquelRForgeFlow:
.


pedrobaeza (changes requested):
I don't agree with this. This is masking other possible problems. If you already have the column, something bad will happen later.

PR #4859, opened 2025-03-20
[17.0][OU-ADD] google_calendar: Nothing to do

carlos-lopez-tecnativa:
TT50059


MiquelRForgeFlow:
Fill [docsource](https://github.com/OCA/OpenUpgrade/blob/17.0/docsource/modules160-170.rst).

MiquelRForgeFlow (approved):

PR #4865, opened 2025-03-20
[17.0][OU-ADD] website_links: Nothing to do

carlos-lopez-tecnativa:
TT51556

I noticed that there is no migration folder for this module, even though it is present in V17. Why? However, I have done the analysis and found no changes. Is it correct to just edit `docsource/modules160-170.rst`, or do I need to add a migration folder for this module?


MiquelRForgeFlow:
> Why?

When there are absolute no changes. I didn't check the code why sometimes it creates an empty analysis file and why sometimes it doesn't create the folder. I am not sure what is the difference between this two cases. I will investigate.

> do I need to add a migration folder for this module?

No.

MiquelRForgeFlow (approved):

PR #4867, opened 2025-03-20
[17.0][IMP] new analysis (20th March 2025)

MiquelRForgeFlow:
~~If anyone want do to the migration of website_sale_digital or sale_quotation_builder, it will be easier this way.~~

Update analysis and fix modules due to analysis.


pedrobaeza:
Please split the unrelated changes in several PRs.

MiquelRForgeFlow:
> unrelated changes

Please, define "unrelated changes". First commit is PR's main purpose, to modify apriori. Second commit is the analysis update as a consequence of modifying apriori. It doesn't make sense to do a new PR for the analysis and put this PR as a PR dependency of the other. Last commit comes from the analysis. I could put the third commit in another PR, but it would lack the analysis part. And I don't want to do a new analysis without first commit and then do a new analysis with the first commit.

pedrobaeza (approved):

PR #4868, opened 2025-03-20
[16.0][OU-ADD] base_address_extended

etobella:
Adds the scripts for the migration.

Otherwise, we do it record by record (really slow if we have a lot of partners)


MiquelRForgeFlow:
Selection_4389

etobella:
The problem is that we cannot use the regex_match inside the UPDATE statement. I opted for a subquery.

MiquelRForgeFlow (approved):

PR #4870, opened 2025-03-21
[18.0][OU-ADD] sale_order_qty_change_no_recompute: Merged into sale

victoralmau:
Merged into `sale` https://github.com/odoo/odoo/commit/c584a61697e

@Tecnativa TT54289

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -15,6 +15,8 @@ renamed_modules = {
 merged_modules = {
     # odoo
     # odoo/enterprise
+    # OCA/sale-workflow
+    "sale_order_qty_change_no_recompute": "sale",
     # OCA/...
 }
 


legalsylvain:
Finally ! 

Note: Thanks for your PR.

hbrunn:
why do we block this kind of PRs? I think we can merge module renames right away.

legalsylvain:
> why do we block this kind of PRs? I think we can merge module renames right away.

Indeed. @MiquelRForgeFlow. I don't think we should wait for "sale" migration. Don't you think ?

MiquelRForgeFlow:
I am fine with the merge, but I know @pedrobaeza is not fine with it.

pedrobaeza:
The idea of blocking this is to avoid any possible problem derived from the merge. It would be an edge case, but just in case. If you think there won't be any problem (and this exact case it won't), go ahead.

legalsylvain:
> It would be an edge case, but just in case

I'm not sure to understand the edge case. I propose to merge such kind of PRs. (leaving PRs open for months on end also has its drawbacks.)
If edge cases occures, please ping me, I'll try to write some explanation somewhere over there https://oca.github.io/OpenUpgrade/100_maintain_repository.html 

thanks ! 

/ocabot merge nobump

MiquelRForgeFlow (approved):

legalsylvain:
looks quite unrelated. 
manual merge.

PR #4872, opened 2025-03-24
[17.0][ADD] update analysis bot

hbrunn:
this creates a cronjob to run the analysis for 18.0 every week. Needs to be in the 17.0 branch because cronjobs only run on the default branch.

This is written to allow easy addition of other versions, but for now I only added v18 because there we need it most, and so we can see how well it works there before adding a bunch of other branches.

You can check out the [workflows](https://github.com/hbrunn/OpenUpgrade/actions/workflows/generate-analysis-cron.yml) on my fork where you can inspect the logs, the [open PR](https://github.com/hbrunn/OpenUpgrade/pull/16) it just created there, and the [PRs for previous versions](https://github.com/hbrunn/OpenUpgrade/pulls?q=is%3Apr+is%3Aclosed+%5BUPD%5D+analysis+files) I created to test support for older versions.

### DIFF
--- a/.github/workflows/generate-analysis.yml
+++ b/.github/workflows/generate-analysis.yml
@@ -1,0 +1,226 @@
+name: Generate analysis files
+
+on:
+  workflow_call:
+    inputs:
+      from_version:
+        required: true
+        type: string
+      from_version_requirements:
+        default: 'gevent==22.10.2 greenlet==2.0.2'
+        type: string
+      from_version_requirements_sed:
+        default: '/^gevent\>/d; /^greenlet\>/d'
+        type: string
+      to_version:
+        required: true
+        type: string
+      to_version_requirements:
+        default: 'gevent==22.10.2 greenlet==2.0.2'
+        type: string
+      to_version_requirements_sed:
+        default: '/^gevent\>/d; /^greenlet\>/d'
+        type: string
+      python_version:
+        default: "3.10"
+        type: string
+      postgres_version:
+        default: "14"
+        type: string
+
+jobs:
+  generate-analysis:
+    name: Generate analysis ${{ inputs.to_version }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/generate-analysis.yml
```diff
@@ -0,0 +1,224 @@
...
+    runs-on: ubuntu-latest
+    env:
+      PGHOST: "localhost"
+      PGPASSWORD: "odoo"
+      PGUSER: "odoo"
+      FROM_VERSION: "${{ inputs.from_version }}"
+      TO_VERSION: "${{ inputs.to_version }}"
+    steps:
+      - name: Configure Postgres
+        uses: harmon758/postgresql-action@v1
```
hparfr:
```suggestion
        uses: harmon758/postgresql-action@0be19fa37850b22cb4c9bbf28a03abbf44abd863 #v1
```
To limit supply-chain attacks

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

See also the recent "tj-actions" affair.

hparfr (changes requested):

.github/workflows/generate-analysis.yml
hparfr: there is also `peter-evans/create-pull-request@v7` line 215

hbrunn: thanks, you might be interested in https://github.com/OCA/OpenUpgrade/pull/4901

hparfr (approved):

hbrunn:
this has worked nicely. You can find the created PR [here](https://github.com/OCA/OpenUpgrade/pull/5053) and the workflow run [here](https://github.com/OCA/OpenUpgrade/actions/workflows/generate-analysis-cron.yml)

PR #4873, opened 2025-03-24
[18.0][ADD] update analysis bot

hbrunn:
This is https://github.com/OCA/OpenUpgrade/pull/4872 for v18, and causes the analysis to be run whenever we push changes to `apriory.py`.

### DIFF
--- a/.github/workflows/generate-analysis.yml
+++ b/.github/workflows/generate-analysis.yml
@@ -1,0 +1,226 @@
+name: Generate analysis files
+
+on:
+  workflow_call:
+    inputs:
+      from_version:
+        required: true
+        type: string
+      from_version_requirements:
+        default: 'gevent==22.10.2 greenlet==2.0.2'
+        type: string
+      from_version_requirements_sed:
+        default: '/^gevent\>/d; /^greenlet\>/d'
+        type: string
+      to_version:
+        required: true
+        type: string
+      to_version_requirements:
+        default: 'gevent==22.10.2 greenlet==2.0.2'
+        type: string
+      to_version_requirements_sed:
+        default: '/^gevent\>/d; /^greenlet\>/d'
+        type: string
+      python_version:
+        default: "3.10"
+        type: string
+      postgres_version:
+        default: "14"
+        type: string
+
+jobs:
+  generate-analysis:
+    name: Generate analysis ${{ inputs.to_version }}
+    runs-on: ubuntu-latest
+    env:
+      PGHOST: "localhost"
+      PGPASSWORD: "odoo"
+      PGUSER: "odoo"
+      FROM_VERSION: "${{ inputs.from_version }}"
+      TO_VERSION: "${{ inputs.to_version }}"
+    services:
+      postgres:
+        image: postgres:${{ inputs.postgres_version }}
+        env:
+          POSTGRES_USER: odoo
+          POSTGRES_PASSWORD: odoo
+        ports:
+          - 5432:5432
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "${{ inputs.python_version }}"
+      - name: Install required packages
+        run: |
+          sudo apt update
+          sudo apt install \
+              expect \
+              expect-dev \
+              libevent-dev \
+              libldap2-dev \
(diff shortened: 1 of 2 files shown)


.github/workflows/generate-analysis.yml
```diff
@@ -0,0 +1,226 @@
...
+          EOF
+          $ODOO_SHELL <<EOF
+          env['upgrade.generate.record.wizard'].create({}).generate()
+          env.cr.commit()
+          EOF
+      - name: Delete all files to be generated from repository
+        env:
+          OPENUPGRADE_DIR: openupgrade-${{ env.TO_VERSION }}
+        run: |
+          for file in $OPENUPGRADE_DIR/openupgrade_scripts/scripts/*/*/{noupdate_changes.xml,upgrade_analysis.txt}; do
```
MiquelRForgeFlow, line 156:
```suggestion
          for file in $OPENUPGRADE_DIR/openupgrade_scripts/scripts/*/*/{noupdate_changes.xml,upgrade_analysis.txt,upgrade_general_log.txt}; do
```

.github/workflows/generate-analysis.yml
```diff
@@ -0,0 +1,226 @@
...
+      - name: Create analysis
+        env:
+          FROM_ODOO: env-${{ env.FROM_VERSION }}/bin/odoo -c odoo-${{ env.FROM_VERSION }}.cfg -d ${{ env.FROM_VERSION }}
+          TO_ODOO_SHELL: env-${{ env.TO_VERSION }}/bin/odoo shell -c odoo-${{ env.TO_VERSION }}.cfg -d ${{ env.TO_VERSION }}
+        run: |
+          $FROM_ODOO --max-cron-threads=0 &
+          ODOO_PID=$!
+          $TO_ODOO_SHELL <<EOF
+          config = env['upgrade.comparison.config'].create({
+              'database': '${{ env.FROM_VERSION }}',
```
MiquelRForgeFlow, line 171:
```suggestion
              'database': '${{ env.FROM_VERSION }}',
              'version':  '${{ env.FROM_VERSION }}',
```

.github/workflows/generate-analysis.yml
```diff
@@ -0,0 +1,226 @@
...
+          git -C $OPENUPGRADE_DIR add .
+
+          echo "Analysis or noupdate changes for modules marked as done:
+
+          " > "$PR_BODY"
+
+          for module in $(
+            git -C $OPENUPGRADE_DIR status --short | awk '{print $2}' | awk -F/ '{print $3}' | sort -u
+          ); do
+            if grep -qE "\<$module\>.*\<((Done)|(No))" $OPENUPGRADE_DIR/docsource/modules*; then
```
MiquelRForgeFlow, line 197: When new analysis is done, it deletes all the content in Status column of docsource, so you will not detect anything here if I understood correctly your code.

hbrunn: this looks at the rst file, which this code doesn't touch at all. Check out https://github.com/hbrunn/OpenUpgrade/pull/16 where finds some modules mentioned there

.github/workflows/generate-analysis.yml
hbrunn: the git rm is for the case that a module changes version, and the general log is only created in the base module. Will this ever change version?

.github/workflows/generate-analysis.yml
hbrunn: that's preset to the current version, isn't it?

.github/workflows/generate-analysis.yml
MiquelRForgeFlow: It can change if we add more rename/merge modules in apriori, for example.

.github/workflows/generate-analysis.yml
MiquelRForgeFlow: Not sure. It's better to make it explicit, I think.

pedrobaeza:
Is this finished?

hbrunn:
yes, it's the same as v17, and will cause an analysis to be run immediately when a change to apriori.py is merged

pedrobaeza (approved):

PR #4878, opened 2025-03-25
[ADD] reference to migration 16.0 -> 17.0 and 17.0 -> 18.0

legalsylvain:
Try to fix https://github.com/OCA/OpenUpgrade/pull/4870 CI error.


MiquelRForgeFlow (approved):

MiquelRForgeFlow:
It seems this didn't solve that issue :S

pedrobaeza:
The default branch is 17.0, so shouldn't that one be the one having all references? Or put as default 18.0 directly.

PR #4883, opened 2025-03-26
[18.0][IMP] apriori: fill it accordingly

MiquelRForgeFlow:
.


docsource/modules170-180.rst
```diff
@@ -988,7 +988,7 @@ Module coverage 17.0 -> 18.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | |del| spreadsheet_dashboard_purchase              |                      |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| spreadsheet_dashboard_purchase_stock        |                      |                                                 |
+| |del| spreadsheet_dashboard_purchase_stock        |                      |Moved to Enterprise                              |
```
pedrobaeza, line 991: Does they really move them to enterprise?

openupgrade_scripts/apriori.py
```diff
@@ -4,8 +4,11 @@
 
 # Renamed modules is a mapping from old module name to new module name
 renamed_modules = {
-    "account_accountant": "accountant",
     # odoo
+    "l10n_es_pos_tbai": "l10n_es_edi_tbai_pos",
+    "mrp_subonctracting_landed_costs": "mrp_subcontracting_landed_costs",
```
pedrobaeza, line 9: Odoo spelled it bad?

docsource/modules170-180.rst
MiquelRForgeFlow: Yes. (`spreadsheet_dashboard_purchase` too but is it merged now in `spreadsheet_dashboard_purchase_stock`)

openupgrade_scripts/apriori.py
MiquelRForgeFlow: Yes.

docsource/modules170-180.rst
pedrobaeza: Ouch... I hope they won't block that we rescue it from previous versions in OCA.

legalsylvain: Not possible this time. It's not an error in a commit in an unreleased version.

pedrobaeza (approved):

PR #4884, opened 2025-03-27
[13.0][FIX] account: avoid infinite loop when searching for child taxes

TelmoSuarez:
Control infinite recursive loop while searching for child taxes. This happens when a child's child is it's parent (A->B->A)

### DIFF
--- a/addons/account/migrations/13.0.1.1/post-migration.py
+++ b/addons/account/migrations/13.0.1.1/post-migration.py
@@ -861,12 +861,13 @@ def create_account_tax_repartition_lines(env):
 def move_tags_from_taxes_to_repartition_lines(env):
     openupgrade.logged_query(
         env.cr, """
-        WITH RECURSIVE tax2parent(tax_id, parent_id) AS (
-            SELECT id, id FROM account_tax
+        WITH RECURSIVE tax2parent(tax_id, parent_id, is_cycle, all_parents) AS (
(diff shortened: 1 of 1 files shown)


legalsylvain:
Hi @TelmoSuarez 
> his happens when a child's child is it's parent (A->B->A)

This configuration looks weird. could you mention wich localization has such circular settings ? 

thanks !

TelmoSuarez:
Hi @legalsylvain.

I found this issue in l10_es. In my case i'm migrating a client from 11.0 to 16.0, that had been previously migrated from 8.0.
I'm not 100% sure if this is specific to my case or is a more general issue, will test with clean db. 
In l10n_es's migration to 13.0 found [this](https://github.com/OCA/OpenUpgrade/blob/ad672c8d8f5974729f5e3387a8ce63dc26ec713b/addons/l10n_es/migrations/13.0.4.0/post-migration.py#L77) query that would "fix" this issue, but due to migration order(account -> l10n_es) it gets stuck in account's query.

legalsylvain:
@MiquelRForgeFlow  / @pedrobaeza : any point of view regarding recursive taxes ?

pedrobaeza:
Well, the error is due to a previous incorrect data handling, so not sure if we should handle here. Anyway, if the SQLs cost is simple and effective, we can do it. I'm not qualified right now to say if the change on the SQLs are OK or not.

PR #4751, opened 2025-02-20
[17.0][MIG] stock_delivery

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4637.

PR #4885, opened 2025-03-27
[17.0][OU-IMP] delivery: Finish migration scripts

carlos-lopez-tecnativa:
Continuation of https://github.com/OCA/OpenUpgrade/pull/4751
TT52315

### DIFF
--- a/openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
@@ -37,6 +37,98 @@ _xmlids_renames = [
         "delivery.model_choose_delivery_package",
         "stock_delivery.model_choose_delivery_package",
     ),
+    (
+        "delivery.choose_delivery_package_view_form",
+        "stock_delivery.choose_delivery_package_view_form",
+    ),
+    (
+        "delivery.delivery_stock_report_delivery_no_package_section_line",
+        "stock_delivery.delivery_stock_report_delivery_no_package_section_line",
+    ),
+    (
+        "delivery.delivery_tracking_url_warning_form",
+        "stock_delivery.delivery_tracking_url_warning_form",
+    ),
+    (
+        "delivery.label_package_template_view_delivery",
+        "stock_delivery.label_package_template_view_delivery",
+    ),
+    (
+        "delivery.product_template_hs_code",
+        "stock_delivery.product_template_hs_code",
+    ),
+    (
+        "delivery.report_delivery_document2",
+        "stock_delivery.report_delivery_document2",
+    ),
+    (
+        "delivery.report_package_barcode_delivery",
+        "stock_delivery.report_package_barcode_delivery",
+    ),
+    (
+        "delivery.report_package_barcode_small_delivery",
+        "stock_delivery.report_package_barcode_small_delivery",
+    ),
+    (
+        "delivery.report_shipping2",
+        "stock_delivery.report_shipping2",
+    ),
+    (
+        "delivery.sale_order_portal_content_inherit_sale_stock_inherit_website_sale_delivery",
+        "stock_delivery.sale_order_portal_content_inherit_sale_stock_inherit_website_sale",
+    ),
+    (
+        "delivery.stock_move_line_view_search_delivery",
+        "stock_delivery.stock_move_line_view_search_delivery",
+    ),
+    (
+        "delivery.stock_package_type_form_delivery",
+        "stock_delivery.stock_package_type_form_delivery",
+    ),
+    (
+        "delivery.stock_package_type_tree_delivery",
+        "stock_delivery.stock_package_type_tree_delivery",
+    ),
+    (
+        "delivery.stock_report_delivery_aggregated_move_lines_inherit_delivery",
+        "stock_delivery.stock_report_delivery_aggregated_move_lines_inherit_delivery",
+    ),
+    (
+        "delivery.stock_report_delivery_has_serial_move_line_inherit_delivery",
+        "stock_delivery.stock_report_delivery_has_serial_move_line_inherit_delivery",
+    ),
+    (
+        "delivery.stock_report_delivery_package_section_line_inherit_delivery",
+        "stock_delivery.stock_report_delivery_package_section_line_inherit_delivery",
+    ),
+    (
+        "delivery.view_move_line_tree_detailed_delivery",
+        "stock_delivery.view_move_line_tree_detailed_delivery",
+    ),
+    (
+        "delivery.view_picking_type_form_delivery",
+        "stock_delivery.view_picking_type_form_delivery",
+    ),
+    (
+        "delivery.view_picking_withcarrier_out_form",
+        "stock_delivery.view_picking_withcarrier_out_form",
+    ),
+    (
+        "delivery.view_picking_withweight_internal_move_form",
+        "stock_delivery.view_picking_withweight_internal_move_form",
+    ),
+    (
+        "delivery.view_quant_package_weight_form",
+        "stock_delivery.view_quant_package_weight_form",
+    ),
+    (
+        "delivery.view_stock_rule_form_delivery",
+        "stock_delivery.view_stock_rule_form_delivery",
+    ),
+    (
+        "delivery.vpicktree_view_tree",
+        "stock_delivery.vpicktree_view_tree",
+    ),
 ]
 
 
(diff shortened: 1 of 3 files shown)


openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -47,6 +139,39 @@ def _delete_sql_constraints(env):
     )
 
 
+def _fill_sale_order_shipping_weight(env):
+    """Field `shipping_weight` on sale.order is a new stored field."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order
+        ADD COLUMN IF NOT EXISTS shipping_weight float8;
```
MiquelRForgeFlow, line 148:
```suggestion
        ADD COLUMN IF NOT EXISTS shipping_weight numeric;
```

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -37,6 +37,98 @@
         "delivery.model_choose_delivery_package",
         "stock_delivery.model_choose_delivery_package",
     ),
+    (
+        "delivery.choose_delivery_package_view_form",
+        "stock_delivery.choose_delivery_package_view_form",
```
MiquelRForgeFlow, line 42: analysis didn't detect it, thanks :(

but views are noupdate=0, so in theory, they are deleted and created again by ORM 🤷‍♂️

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -47,6 +139,39 @@ def _delete_sql_constraints(env):
...
+        """
+        WITH weight_data AS (
+            SELECT
+                COALESCE(SUM(sol.product_uom_qty * p.weight), 0) AS shipping_weight,
+                sol.order_id
+            FROM sale_order_line sol
+            JOIN product_product p ON sol.product_id = p.id
+            JOIN product_template pt ON p.product_tmpl_id = pt.id
+            WHERE pt.type IN ('product', 'consu')
+                AND (sol.is_delivery IS NULL OR sol.is_delivery = FALSE)
```
MiquelRForgeFlow, line 164:
```suggestion
                AND NOT sol.is_delivery
```

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -47,6 +139,39 @@ def _delete_sql_constraints(env):
...
+                AND sol.display_type IS NULL
```
MiquelRForgeFlow, line 165:
```suggestion
                AND NOT sol.display_type
```

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -47,6 +139,39 @@ def _delete_sql_constraints(env):
...
+        ALTER TABLE sale_order
+        ADD COLUMN IF NOT EXISTS shipping_weight float8;
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        WITH weight_data AS (
+            SELECT
+                COALESCE(SUM(sol.product_uom_qty * p.weight), 0) AS shipping_weight,
```
MiquelRForgeFlow:
```suggestion
                SUM(COALESCE(sol.product_uom_qty,0) * COALESCE(p.weight,0)) AS shipping_weight,
```

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
```diff
@@ -47,6 +139,39 @@ def _delete_sql_constraints(env):
...
+                AND (sol.is_delivery IS NULL OR sol.is_delivery = FALSE)
+                AND sol.display_type IS NULL
+                AND sol.product_uom_qty > 0
+            GROUP BY sol.order_id
+        )
+        UPDATE sale_order so
+        SET shipping_weight = wd.shipping_weight
+        FROM weight_data wd
+        WHERE so.id = wd.order_id;
+        """,
```
MiquelRForgeFlow, line 173: add
```python
    openupgrade.logged_query(
        env.cr,
        """UPDATE sale_order so
           SET shipping_weight = 0
           WHERE so.shipping_weight IS NULL
        """)
```

openupgrade_scripts/scripts/delivery/17.0.1.0/upgrade_analysis_work.txt
```diff
@@ -8,7 +8,7 @@ delivery     / delivery.carrier         / invoice_policy (selection)    : select
 # NOTHING TO DO
 
 delivery     / sale.order               / shipping_weight (float)       : NEW hasdefault: compute
-# NOTHING TO DO: new features
+# DONE: pre-migration: Add new column & set value for it
```
MiquelRForgeFlow, line 11: you should also do the stock_delivery part too in stock_delivery.
Selection_4392

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
pedrobaeza: Are you sure `NOT` works with non boolean columns?

MiquelRForgeFlow: Oh, my bad, it's not a boolean.

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
carlos-lopez-tecnativa: The default type chosen for Odoo is `float8` because this field does not have `digits` like other fields such as `amount_total`. So, I chose the same type based on the same logic. Please confirm if I should change the field type?.
https://github.com/odoo/odoo/blob/a0542b13f679d4dc20d61eb62e4eb5106d6abe5a/odoo/fields.py#L1530

https://github.com/odoo/odoo/blob/a0542b13f679d4dc20d61eb62e4eb5106d6abe5a/addons/delivery/models/sale_order.py#L16

image

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
carlos-lopez-tecnativa: What happens if another view in a different module inherits any view from this list? According to this code, it should be restricted from unlinking, or did I miss something?
https://github.com/odoo/odoo/blob/a0542b13f679d4dc20d61eb62e4eb5106d6abe5a/odoo/addons/base/models/ir_ui_view.py#L184

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
carlos-lopez-tecnativa: This is not working correctly.
With your proposal, the result is:
image

With the current code, the result is:
image

Because you have `true`, `false`, `null`:
image

openupgrade_scripts/scripts/delivery/17.0.1.0/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Sorry, but I don't understand. This function belongs to `stock.picking`, and my code only takes into account `sale.order`. Please, are you sure your suggestion is correct?
https://github.com/odoo/odoo/blob/17.0/addons/stock_delivery/models/stock_picking.py#L317

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
MiquelRForgeFlow: Oh sorry.

FYI Equivalent options:

```suggestion
                AND NOT COALESCE(sol.is_delivery, FALSE)
```
```suggestion
                AND sol.is_delivery IS DIFFERENT FROM TRUE
```
```suggestion
                AND (sol.is_delivery IS NULL OR sol.is_delivery = FALSE)
```

openupgrade_scripts/scripts/delivery/17.0.1.0/pre-migration.py
MiquelRForgeFlow: Oh, I didn't say you should avoid the rename, just that is not strictly needed. framework avoid this thing you comment if I am not wrong.

MiquelRForgeFlow (approved):

PR #4898, opened 2025-03-31
[18.0][ADD] openupgrade_framework: add scripts' model renames in apriori

hbrunn:
this way we don't have to administer things at two places

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -1,6 +1,10 @@
 """ Encode any known changes to the database here
 to help the matching process
 """
+import os
+import os.path
+
+from odoo.modules.migration import load_script
 
 # Renamed modules is a mapping from old module name to new module name
 renamed_modules = {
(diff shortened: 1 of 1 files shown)


legalsylvain (changes requested):
At a first sight, it looks good to reduce places where thing are defined. 

However, this is breaking the order of the work of Openupgrade.
this file has to be written well before the migration scripts to be used by upgrade_anlysis module. see : https://github.com/OCA/server-tools/blob/18.0/upgrade_analysis/compare.py#L30-L46
With the refactoring you propose, I think it will not be possible to generate correct analysis files at the beginning of the work of OpenUpgrade.

(subjective point) also this is less readeable : this apriori.py file is interesting to see all the big changes in a single place.

pedrobaeza:
And this is assuming you will have a variable `model_renames` that performs the job, being some kind of "black magic", so I don't like it too much, as it's very fragile.

Starting with, I prefer to name that variables as internal (`_model_renames`), but others may think that are static data (which they are), and name it as `MODEL_RENAMES` or `_MODEL_RENAMES`. And others may not know this "convention", as call it other way, and it's very easy that we bypass it on the reviews, as nothing will fail in the CI.

legalsylvain:
Indeed, @pedrobaeza.

also such valid syntax will be skipped : https://github.com/OCA/account-financial-reporting/blob/12.0/partner_statement/migrations/12.0.1.0.0/pre-migration.py#L21

pedrobaeza:
Maybe another option is to parse by introspection the `rename_models` calls and read the arguments.

legalsylvain:
> Maybe another option is to parse by introspection the rename_models calls and read the arguments.

I don't get how ``upgrade_analysis`` could work with your proposal ? (see : https://github.com/OCA/OpenUpgrade/pull/4898#pullrequestreview-2728402174)

hbrunn:
for addressing the initial phase, I just need to add another isdir check at the top. Trying multiple variable names also isn't a problem imho.

But if the general consensus is duplicating this is better anyways (before you could const_eval this file, with this, you can't), I'll ditch this and keep copy&pasting stuff.

MiquelRForgeFlow:
I am against, by the arguments already exposed. I don't thing the "duplication issue" is truly such a big issue that needs this kind of complex code in order to be "solved". Maybe someday exists a model rename/merge we may want the analysis part to consider it, but in the migration script maybe we are not interested in calling the `model_renames` but a query in post-migration with `insert into`, or something similar. Here in apriori, the model rename/merge only serves for the analysis purpose, as they are not called in pre-migration base like the module rename/merges. Your approach is mixing the migration and analysis...

hbrunn:
so closing this and adding the renamed modules manually in the mail migration you asked for

PR #4899, opened 2025-03-31
[15.0][OU][FIX] point_of_sale: empty journal in bank payment method

chienandalu:
From now on, a journal is required for payment methods when transactions aren't splitted. If we don't set a journal in these case, the session closing won't be made properly for these methods.

cc @Tecnativa @pedrobaeza  TT54971

### DIFF
--- a/openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/post-migration.py
+++ b/openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/post-migration.py
@@ -1,0 +1,31 @@
+# Copyright 2025 Tecnativa
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def _update_pos_payment_method_journal(env):
+    """From now on, a journal is required when the transactions aren't splitted. If we
+    don't set a journal in these cases, the session closing won't be made properly for
+    these methods. Cash payment methos already had a jornal set, but bank ones didn't."""
+    payment_methods = env["pos.payment.method"].search(
+        [
+            ("journal_id", "=", False),
+            ("split_transactions", "=", False),
+            ("type", "=", "bank"),
+        ]
+    )
+    for method in payment_methods:
+        method.journal_id = env["account.journal"].create(
+            {
+                "type": "bank",
+                "name": f"[openupgrade] Journal for {method.name}",
+                "code": "POS_BANK",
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2025 Tecnativa
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def _update_pos_payment_method_journal(env):
+    """From now on, a journal is required when the transactions aren't splitted. If we
```
MiquelRForgeFlow, line 7: is there a link to odoo code to see/undertand this kind of feature?

chienandalu: Here's the rule by UI that makes it required: https://github.com/OCA/OCB/blob/4e770afbd0c253c28555a1e71a933568787fdcf6/addons/point_of_sale/views/pos_payment_method_views.xml#L20

A side effect of this is that the invoices payed with these methods aren't correctly reconciled with their payments.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

openupgrade_scripts/scripts/point_of_sale/15.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+            ("split_transactions", "=", False),
+            ("type", "=", "bank"),
+        ]
+    )
+    for method in payment_methods:
+        method.journal_id = env["account.journal"].create(
+            {
+                "type": "bank",
+                "name": f"[openupgrade] Journal for {method.name}",
+                "code": "POS_BANK",
```
legalsylvain, line 22: Hi @chienandalu. Could you fix this point ? https://github.com/OCA/OpenUpgrade/commit/bde6c18907e1daa4a03c00e9bcb6dc34201f18d2#r154676459


(Thanks to @em230418)

pedrobaeza: I will publish a patch soon

pedrobaeza: Solved in #4917

PR #4910, opened 2025-04-01
[17.0][OU-ADD] sale_timesheet: Migration scripts to 17.0

carlos-lopez-tecnativa:
TT50096


openupgrade_scripts/scripts/sale_timesheet/17.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,33 @@
...
+            )
+        ],
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE project_project
+        SET billing_type = 'not_billable';
+        """,
+    )
```
pedrobaeza: Anyway, being a static value, I'm not sure the gain doing this over the ORM counterpart.

```suggestion
                "not_billable",
            )
        ],
    )
```

PR #4913, opened 2025-04-01
[17.0][OU-ADD] hr_timesheet_attendance: Migration scripts to 17.0

carlos-lopez-tecnativa:
TT50078
@Tecnativa @pedrobaeza @MiquelRForgeFlow  could you please review this?


openupgrade_scripts/scripts/hr_timesheet_attendance/17.0.1.1/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env, "hr_timesheet_attendance", "17.0.1.1/noupdate_changes.xml"
+    )
+    openupgrade.delete_record_translations(
```
MiquelRForgeFlow: why?

carlos-lopez-tecnativa: Because the action name changed and this field is translated, or is it not necessary?

https://github.com/odoo/odoo/blob/d5b7f5fc46f3775e508417883849d931c8cc073e/addons/hr_timesheet_attendance/report/hr_timesheet_attendance_report_view.xml#L47

MiquelRForgeFlow: It's noupdate=0. So ORM should handle it.

pedrobaeza (approved):

PR #4920, opened 2025-04-07
[17.0][OU-IMP] analytic: Improved project plan selection + clean folder duplication

pedrobaeza:
We allow to pre-fill in the system parameter or the ir.model.data entry which is going to be the plan for projects.

@Tecnativa


openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
```diff
@@ -1,12 +1,29 @@
...
+    Finally, if not found, we get the first existing plan.
+    """
+    if env["ir.config_parameter"].get_param("analytic.project_plan", False):
+        return
+    imd = env["ir.model.data"].search(
+        [("module", "=", "analytic"), ("name", "=", "analytic_plan_projects")]
+    )
+    plan_id = imd.res_id
+    if not plan_id:
+        env.cr.execute("SELECT min(id) FROM account_analytic_plan")
```
hparfr, line 24: @pedrobaeza 
there is a regression on this line.
If there is no `account_analytic_plan`  it returns nothing 


I then got an issue 
```
  File "/odoo/src/addons/analytic/models/analytic_plan.py", line 107, in __get_all_plans
    project_plan = self.browse(int(self.env['ir.config_parameter'].sudo().get_param('analytic.project_plan', 0)))
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'None'

```

I don't get why I got None instead of 0 in get_param

Maybe 
```sql
SELECT COALESCE(min(id),0) min  FROM account_analytic_plan;
```
Is a bit better because we expect plan_id to have a value in line 26

pedrobaeza: Solved other way in #4921. At the end, we will always end up with an existing `analytic.analytic_plan_projects` record sooner or later, so we have to stick to it.

PR #4921, opened 2025-04-08
[17.0][OU-FIX] analytic: Fine-tune "Projects" plan finding

pedrobaeza:
Instead of selecting the first available plan, that can lead to an error if there's none, if we are on that part of the flow, we are sure the plan will be created through XML data, so we set the next available ID.

From the comment https://github.com/OCA/OpenUpgrade/pull/4920#discussion_r2032931116

@Tecnativa


hparfr (approved):
Tested successfully

Thanks for this fast fix @pedrobaeza  !

PR #4922, opened 2025-04-10
[16.0][OU-ADD] survey_placeholder: Merge into survey

pilarvargas-tecnativa:
cc @Tecnativa TT51954

ping @pedrobaeza

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -123,6 +123,8 @@ merged_modules = {
     "mail_preview_base": "mail",
     # OCA/stock-logistics-workflow
     "stock_picking_backorder_strategy": "stock",
+    # OCA/survey
+    "survey_placeholder": "survey",
     # OCA/web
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/apriori.py
```diff
@@ -123,6 +123,8 @@
     "mail_preview_base": "mail",
     # OCA/stock-logistics-workflow
     "stock_picking_backorder_strategy": "stock",
+    # OCA/survey
+    "question_placeholder": "survey",
```
pedrobaeza:
```suggestion
    "survey_placeholder": "survey",
```

pedrobaeza (changes requested):
Is the field name the same?

openupgrade_scripts/apriori.py
pilarvargas-tecnativa: Yes, it is the same

pedrobaeza (approved):

PR #4925, opened 2025-04-12
[17.0][IMP] mail: improve query performance when setting outgoing messages

royle-vietnam:


pedrobaeza:
Can you publish some stats about the performance gain? And also the `EXPLAIN ANALYZE` outputs of both queries in the system you are struggling with. Seeing the query, it's not easy to deduce the gain.

pedrobaeza (approved):
OK, I have directly checked in a DB with 721481 records in mail.message, and we get 5 seconds against 3 minutes, so the performance gain is noticeable and no need of more proofs. Thanks for the improvement.

PR #4932, opened 2025-04-16
[14.0][OU-FIX] fix posting bank statements

huguesdk:
fix posting bank statements during account post-migration: call `.button_post()` on each statement one by one instead of on all statements to avoid an error.


pedrobaeza (approved):
I didn't face this, but it's not a problem to do it this way.

PR #4942, opened 2025-04-22
[15.0][OU-FIX] point_of_sale: Non-stored field pos.payment.method.typ…

FernandoRomera:
…e cannot be searched


FernandoRomera:
@pedrobaeza
Can you review?

pedrobaeza (approved):

PR #4956, opened 2025-04-23
[17.0][OU-FIX] account: Restict name deduplication

pedrobaeza:
It's OK to be duplicated. In fact, all the draft moves have "/" as name.

@Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
@@ -114,6 +114,7 @@ def _am_uniquify_name(env):
         UPDATE account_move SET name=name || ' [' || id || ']'
         FROM (
             SELECT array_agg(id) ids FROM account_move
+            WHERE state = 'posted' AND name != '/'
             GROUP BY journal_id, name HAVING COUNT(id)>1
         ) duplicate_names
         WHERE account_move.id=ANY(duplicate_names.ids);


pedrobaeza:
@hbrunn I don't get why you introduced this, but seeing in native 17 DB, I don't see such constraint.

openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
```diff
@@ -104,23 +104,6 @@ def _am_create_incoterm_location_column(env):
...
-    """
-    openupgrade.logged_query(
-        env.cr,
-        """
-        UPDATE account_move SET name=name || ' [' || id || ']'
-        FROM (
-            SELECT array_agg(id) ids FROM account_move
-            GROUP BY journal_id, name HAVING COUNT(id)>1
-        ) duplicate_names
-        WHERE account_move.id=ANY(duplicate_names.ids);
```
MiquelRForgeFlow, line 119: `AND account_move.state = 'posted'`

pedrobaeza: OK, now I see the constraint is put via unique index:

https://github.com/odoo/odoo/blob/5fdef9019db65749b6f1f80d828fda6bc2280f55/addons/account/models/account_move.py#L641

so the conditions of the deduplication are changed to met .

PR #4957, opened 2025-04-24
[18.0][OU-ADD] stock

MiquelRForgeFlow:
Depends on:

- [x] https://github.com/OCA/openupgradelib/pull/412

<dependency>Depends on : 

- product : #4839
- barcodes_gs1_nomenclature : #4951
- digest : #4950</dependency>


MiquelRForgeFlow:
/ocabot migration stock

<dependency></dependency>

rvalyi:
for reference the 2 breaking changes and main feature in the stock module for 18.0: https://github.com/akretion/odoo-module-diff-analysis/tree/main/18.0/stock

openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,81 @@
+# Copyright 2025 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+_field_renames = [
+    ("stock.move", "stock_move", "location_dest_id", "location_final_id"),
```
hbrunn, line 6: @MiquelRForgeFlow did you test this with a multi step route? Now working on the mrp migration, I very much doubt this is correct.

I think we need to keep location_dest_id what it is, and reconstruct location_final_id from the procurement that cause the move

MiquelRForgeFlow: But `location_dest_id` is filled in `fill_stock_move_location_dest_id()`.

hbrunn: yes, but still location_final_id will be the original value of location_dest_id.

Consider a route:

A -> B
B -> C

then we want the v17 move with

```python
{
    'location_src_id': A,
    'location_dest_id': B,
}
```

to end up in v18 as

```python
{
    'location_src_id': A,
    'location_dest_id': B,
    'location_final_id': C,
}
```

while now we get

```python
{
    'location_src_id': A,
    'location_dest_id': B,
    'location_final_id': B,
}
```

MiquelRForgeFlow: Oh, then it lacks some script to update the `location_final_id` at some point in the migration. I suppose it has to check the locations of move_dest_ids, right? Do you have any concrete idea?

hbrunn: I'm now working on a function that takes all stock moves from the procurement group, and determines the "greatest" location_dest_id in the sense that there's no other stock move that goes out of this location within the group. will probably need some fine tuning re scrap locations and the like

hbrunn: this is more tricky then it looked like in the beginning, will revisit next week

MiquelRForgeFlow:
```sql
UPDATE stock_move sm
SET location_final_id = COALESCE(sr.location_dest_id, sm.location_final_id)
FROM stock_rule sr
WHERE sm.rule_id = sr.id AND sr.location_dest_from_rule
```

What do you think of this?

MiquelRForgeFlow: See https://github.com/OCA/OpenUpgrade/pull/5088.

hbrunn: I'm still not entirely clear on the correct behavior of this for chained moves, will have to do more code analysis next week

PR #4977, opened 2025-05-05
[18.0][OU-FIX] base: add missing SQL to fix view mode in ir_act_window

carlos-lopez-tecnativa:
Previous actions had a default value of `tree,form`, but now the default is `list,form`. If any records do not define this field explicitly, the default value from the previous version is kept, which causes an error.

image

image

### DIFF
--- a/openupgrade_scripts/scripts/base/18.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/18.0.1.3/pre-migration.py
@@ -23,6 +23,21 @@ def _fix_list_view_type(cr):
     openupgrade.logged_query(cr, "UPDATE ir_ui_view SET type='list' WHERE type='tree'")
 
 
+def _fix_list_view_mode(cr):
+    """
+    Previous actions had a default value of tree,form, but now the default is list,form.
+    If any records do not define this field explicitly,
+    the default value from the previous version is kept, which causes an error.
+    """
+    openupgrade.logged_query(
+        cr,
+        r"""UPDATE ir_act_window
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/base/18.0.1.3/pre-migration.py
```diff
@@ -23,6 +23,21 @@ def _fix_list_view_type(cr):
...
+def _fix_list_view_mode(cr):
+    """
+    Previous actions had a default value of tree,form, but now the default is list,form.
+    If any records do not define this field explicitly,
+    the default value from the previous version is kept, which causes an error.
+    """
+    openupgrade.logged_query(
+        cr,
+        """UPDATE ir_act_window
+            SET view_mode=REPLACE(view_mode, 'tree', 'list')
```
pedrobaeza: This is dangerous, as may exist a view `tree_hierarchy` (I think there's an OCA module like this), and it will be replaced by `list_hierarchy`. You should look for `tree,` or `tree$`, being `$` the end of the string (with regexp can be done).

carlos-lopez-tecnativa: I think you were referring to this module, but it hasn't been merged yet: https://github.com/OCA/web/pull/2970.
Anyway, I updated the SQL to be more specific, please let me know if it's OK for you.

legalsylvain (approved):
Thanks !

pedrobaeza (approved):

PR #4979, opened 2025-05-06
[16.0][OU-ADD] website_sale_slides: Migration to 16.0

carlos-lopez-tecnativa:
TT55808


openupgrade_scripts/scripts/website_sale_slides/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+    Adapt all products using into a course to the new detailed_type=course
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE product_template pt
+            SET detailed_type = 'course'
+        FROM product_product pp
+            JOIN slide_channel sc ON sc.product_id = pp.id
+        WHERE pp.product_tmpl_id = pt.id
+            AND sc.product_id IS NOT NULL;
```
MiquelRForgeFlow:
```suggestion
        WHERE pp.product_tmpl_id = pt.id;;
```
the `join` already assures that is not null

carlos-lopez-tecnativa: Done, thanks

openupgrade_scripts/scripts/website_sale_slides/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Adapt all products using into a course to the new detailed_type=course
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE product_template pt
+            SET detailed_type = 'course'
+        FROM product_product pp
+            JOIN slide_channel sc ON sc.product_id = pp.id
+        WHERE pp.product_tmpl_id = pt.id;
```
MiquelRForgeFlow, line 16:
```suggestion
        WHERE pp.product_tmpl_id = pt.id AND pt.type = 'service';
```
not sure if this is needed...

carlos-lopez-tecnativa: The field doesn't have a specific domain, so you can select any product. Therefore, I think the additional WHERE clause is not necessary.

image

image

MiquelRForgeFlow: Interesting.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #4980, opened 2025-05-06
[16.0][OU-ADD] website_slides_survey: Migration to 16.0

carlos-lopez-tecnativa:
TT51975


openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2025 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def _delete_sql_constraints(env):
+    # Delete constraints to recreate it
```
MiquelRForgeFlow: not sure if this is needed at v16. Do you know @pedrobaeza ?

carlos-lopez-tecnativa: I saw a similar case in the `account` module:
https://github.com/OCA/OpenUpgrade/blob/c158fd44fcf15bc090a14c2e50863789bb4657e5/openupgrade_scripts/scripts/account/16.0.1.2/pre-migration.py#L133-L135
Why is it not necessary in version 16? Could you please explain it to me?

MiquelRForgeFlow: Because at some point, I don't remember the version, odoo introduced a fix that automatically makes the ORM update these constrains. But let's merge as it is to assure the constrains are reset.

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
+---Models in module 'website_slides_survey'---
+---Fields in module 'website_slides_survey'---
+website_slides_survey / slide.slide              / name (False)                  : NEW mode: modify, hasdefault: compute
+# NOTHING TO DO: The field exists in the website_slides module as storable; just changed it to computed.
+
+website_slides_survey / slide.slide              / slide_category (False)        : NEW selection_keys: ['article', 'certification', 'document', 'infographic', 'quiz', 'video'], mode: modify
+# DONE pre-migration: The field exists in the website_slides module, the migration script is done on website_slides
```
MiquelRForgeFlow:
```suggestion
# NOTHING TO DO: The field exists in the website_slides module, the migration script is done on website_slides
```

carlos-lopez-tecnativa: Done, thanks

carlos-lopez-tecnativa:
To add context to the latest changes: `slide_type` is set to `null` after the recompute triggered by `rename_columns`
https://github.com/OCA/OpenUpgrade/blob/c158fd44fcf15bc090a14c2e50863789bb4657e5/openupgrade_scripts/scripts/website_slides/16.0.2.6/pre-migration.py#L29

Before the migration process (database from V15):
image

After migrating to V16 without this change, the records are not set to `null` because the computed field is inherited from another module:
https://github.com/odoo/odoo/blob/acbd2da7828090753a9ee3dc85c44bc8eca2aaa8/addons/website_slides_survey/models/slide_slide.py#L69

image

After migrating to V16 with this change:
image

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2025 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+_translations_to_delete = [
```
pedrobaeza: Being strict, these are used as constants, so they can be put in upper and with a meaningful name, but you can pass directly the value in the method call instead of putting it here.

```suggestion
XMLIDS_TO_DELETE_TRANSLATIONS = [
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/pre-migration.py
pedrobaeza: I don't remember either when and if the constraint are correctly recreated.

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,50 @@
...
+    "mail_template_user_input_certification_failed",
+]
+
+
+def _delete_sql_constraints(env):
+    # Delete constraints to recreate it
```
pedrobaeza: It can be a docstring.

```suggestion
    """Delete constraints to recreate it"""
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,50 @@
...
+
+    https://github.com/OCA/OpenUpgrade/blob/c158fd44fcf15bc090a14c2e50863789bb4657e5/openupgrade_scripts/scripts/website_slides/16.0.2.6/pre-migration.py#L29
+    https://github.com/odoo/odoo/blob/acbd2da7828090753a9ee3dc85c44bc8eca2aaa8/addons/website_slides_survey/models/slide_slide.py#L69
+
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE slide_slide
+        SET slide_type = slide_category
```
pedrobaeza: To be more verbose:

```suggestion
        SET slide_type = 'certification'
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
+---Models in module 'website_slides_survey'---
+---Fields in module 'website_slides_survey'---
+website_slides_survey / slide.slide              / name (False)                  : NEW mode: modify, hasdefault: compute
+# NOTHING TO DO: The field exists in the website_slides module as storable; just changed it to computed.
+
+website_slides_survey / slide.slide              / slide_category (False)        : NEW selection_keys: ['article', 'certification', 'document', 'infographic', 'quiz', 'video'], mode: modify
+# NOTHING TO DO: The field exists in the website_slides module, the migration script is done on website_slides
+
+website_slides_survey / slide.slide              / slide_type (False)            : selection_keys is now '['article', 'certification', 'doc', 'google_drive_video', 'image', 'pdf', 'quiz', 'sheet', 'slides', 'vimeo_video', 'youtube_video']' ('['certification', 'document', 'infographic', 'presentation', 'quiz', 'video', 'webpage']')
+# Done: Fill the correct value when slide_category=certification
```
pedrobaeza:
```suggestion
# DONE: pre-migration: Filled with "certification" value when slide_category=certification
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+
+---XML records in module 'website_slides_survey'---
+DEL ir.actions.act_window: website_slides_survey.survey_survey_action_slides_report
+DEL ir.ui.menu: website_slides_survey.website_slides_menu_report_certification
+# NOTHING TO DO: Handle by ORM
```
pedrobaeza:
```suggestion
# NOTHING TO DO: Handled by ORM
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+
+NEW ir.actions.act_window.view: website_slides_survey.survey_survey_action_slides_view_form
+NEW ir.actions.act_window.view: website_slides_survey.survey_survey_action_slides_view_kanban
+NEW ir.actions.act_window.view: website_slides_survey.survey_survey_action_slides_view_tree
+# NOTHING TO DO: New features
```
pedrobaeza:
```suggestion
# NOTHING TO DO: it's noupdate=0 stuff
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+
+ir.model.constraint: website_slides_survey.constraint_slide_slide_check_certification_preview (changed definition: is now 'check(slide_category != 'certification' or is_preview = false)' ('check(slide_type != 'certification' or is_preview = false)'))
+ir.model.constraint: website_slides_survey.constraint_slide_slide_check_survey_id (changed definition: is now 'check(slide_category != 'certification' or survey_id is not null)' ('check(slide_type != 'certification' or survey_id is not null)'))
+# DONE pre-migration: safely delete constraint to recreate it
```
pedrobaeza:
```suggestion
# DONE: pre-migration: safely delete constraints for letting ORM to recreate them
```

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+
+NEW ir.ui.view: website_slides_survey.res_config_settings_view_form
+NEW ir.ui.view: website_slides_survey.slide_sidebar_done_button
+NEW ir.ui.view: website_slides_survey.slide_slide_partner_view_form
+NEW ir.ui.view: website_slides_survey.slide_slide_partner_view_search
+NEW ir.ui.view: website_slides_survey.slide_slide_partner_view_tree
+NEW ir.ui.view: website_slides_survey.survey_survey_view_kanban
+NEW ir.ui.view: website_slides_survey.survey_survey_view_search_slides
+NEW ir.ui.view: website_slides_survey.survey_survey_view_tree_slides
+# NOTHING TO DO: New features
```
pedrobaeza: This can be grouped with the previous ir noupdate=0 stuff.

pedrobaeza:
> To add context to the latest changes: `slide_type` is set to `null` after the recompute triggered by `rename_columns`
> 
> https://github.com/OCA/OpenUpgrade/blob/c158fd44fcf15bc090a14c2e50863789bb4657e5/openupgrade_scripts/scripts/website_slides/16.0.2.6/pre-migration.py#L29
> 
> Before the migration process (database from V15): image
> 
> After migrating to V16 without this change, the records are not set to `null` because the computed field is inherited from another module: https://github.com/odoo/odoo/blob/acbd2da7828090753a9ee3dc85c44bc8eca2aaa8/addons/website_slides_survey/models/slide_slide.py#L69
> 
> image
> 
> After migrating to V16 with this change: image

Maybe it's better to not renaming, and force in end-migration script the overall recomputation?

carlos-lopez-tecnativa:
> Maybe it's better to not renaming, and force in end-migration script the overall recomputation?

I’ve updated the code, please let me know if it’s OK for you.

openupgrade_scripts/scripts/website_slides_survey/16.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr, "website_slides_survey", "16.0.1.0/noupdate_changes.xml"
+    )
+    openupgrade.delete_record_translations(
+        env.cr,
+        "website_slides_survey",
+        ["mail_template_user_input_certification_failed"],
+    )
```
MiquelRForgeFlow: `load_data` (+`delete_record_translations`) is always called in post-migration.

carlos-lopez-tecnativa: Thanks so much for the correction. I'll take it into account next time.

pedrobaeza (approved):

openupgrade_scripts/scripts/website_slides/16.0.2.6/end-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Tecnativa - Carlos Lopez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def _recompute_slide_slide_type(env):
```
pedrobaeza, line 7: For having a trace of what's going on.

```suggestion
@openupgrade.logging()
def _recompute_slide_slide_type(env):
```

carlos-lopez-tecnativa: Done

PR #4982, opened 2025-05-06
[16.0][OU-ADD] sale_quotation_builder : Nothing to do

celm1990:
TT51943


pedrobaeza (approved):
Well, the ORM computation may be slow if there's a lot of template lines, but being html_translate makes very difficult to handle this by SQL, so let's continue this way.

PR #4987, opened 2025-05-07
[15.0][OU-FIX] hr_expense: if account_journal_lock_date, skip lock date

FernandoRomera:


pedrobaeza:
Maybe it's better to write the value by SQL?

FernandoRomera:
@pedrobaeza changed.

pedrobaeza (approved):

PR #4995, opened 2025-05-07
[17.0][OU-ADD] sale_quotation_builder > sale_pdf_quote_builder: Nothing to do

carlos-lopez-tecnativa:
I don't see any equivalence between these modules. Odoo removed one and added another that is completely different. Please let me know if there's anything else that needs to be done in this PR. Do I need to mark the new module `sale_pdf_quote_builder` as `to install` or mark the older one as `to uninstall`?

https://github.com/odoo/odoo/pull/133773

TT51944
@Tecnativa @pedrobaeza @MiquelRForgeFlow 

Depends on:
- [x] https://github.com/OCA/sale-reporting/pull/327


MiquelRForgeFlow:
I know they are not technically a continuity, but they are similar in functionality. If you had installed the old one, you should expect to have installed the new one after the migration. That's why I put them in apriori (see https://github.com/OCA/OpenUpgrade/blob/17.0/openupgrade_scripts/apriori.py#L36).

carlos-lopez-tecnativa:
I closed this PR. We decided to recover these features in an OCA module.
Please see this PR https://github.com/OCA/sale-reporting/pull/327 for more information.

carlos-lopez-tecnativa:
I reopened this PR to remove the module from apiori.py.
@pedrobaeza @MiquelRForgeFlow, please let me know if these changes are OK or if I’m missing anything.

openupgrade_scripts/scripts/sale_pdf_quote_builder/17.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
...
+NEW ir.actions.report: sale_pdf_quote_builder.action_report_saleorder_raw
+NEW ir.ui.view: sale_pdf_quote_builder.res_config_settings_view_form
+NEW ir.ui.view: sale_pdf_quote_builder.sale_order_template_form
+DEL ir.ui.view: sale_quotation_builder.brand_promotion
+DEL ir.ui.view: sale_quotation_builder.res_config_settings_view_form_inherit
+DEL ir.ui.view: sale_quotation_builder.sale_order_form_quote_design
+DEL ir.ui.view: sale_quotation_builder.sale_order_portal_content_inherit_sale_quotation_builder
+DEL ir.ui.view: sale_quotation_builder.sale_order_template_view_form_inherit_sale_quotation_builder
+DEL ir.ui.view: sale_quotation_builder.so_template
+DEL sale.order.template: sale_quotation_builder.sale_order_template_default (noupdate)
```
MiquelRForgeFlow: Independently of the port to v17, shouldn't this be safely deleted?

carlos-lopez-tecnativa: Why? The records exist with the same ID.

docsource/modules160-170.rst
```diff
@@ -916,7 +916,7 @@ Module coverage 16.0 -> 17.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | sale_purchase_stock                               | Nothing to do        |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| sale_quotation_builder                      |                      |Merged into sale_pdf_quote_builder.              |
+| |del| sale_quotation_builder                      | Nothing to do        |Merged into sale_pdf_quote_builder.              |
```
pedrobaeza: Remove `Nothing to do` here. It shouldn't be checked in CI. 

Replace the text `Merged into..` by `Forward-ported into OCA/sale-reporting`

openupgrade_scripts/scripts/sale_pdf_quote_builder/17.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,31 @@
+---Models in module 'sale_pdf_quote_builder'---
```
pedrobaeza: This file and the source one doesn't make sense, as the renaming has been undone. And similar for `sale_pdf_quote_builder`, as all the possible renamings shouldn't be taken into account in the analysis file.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/sale_pdf_quote_builder/17.0.1.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Yeah, better to remove it. The `upgrade_analysis.txt` too.

openupgrade_scripts/scripts/sale_pdf_quote_builder/17.0.1.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Well, do as you see fit, but maybe someone doesn't want the forward port to v17 in OCA...

MiquelRForgeFlow:
Maybe a pre-migration script with:

```python
openupgrade.logged_query(
        env.cr,
        """
        UPDATE ir_module_module
        SET state='to install'
        WHERE name = 'sale_pdf_quote_builder' AND state='uninstalled'""",
    )
```
    
 could be added.

docsource/modules160-170.rst
carlos-lopez-tecnativa: Done

openupgrade_scripts/scripts/sale_pdf_quote_builder/17.0.1.0/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Done

carlos-lopez-tecnativa:
> Maybe a pre-migration script with:
> 
> ```python
> openupgrade.logged_query(
>         env.cr,
>         """
>         UPDATE ir_module_module
>         SET state='to install'
>         WHERE name = 'sale_pdf_quote_builder' AND state='uninstalled'""",
>     )
> ```
> 
> could be added.

I think this is not necessary. @pedrobaeza, what do you think?

pedrobaeza:
That's it. We shouldn't force that installation. It's user choice.

docsource/modules160-170.rst
```diff
@@ -916,7 +916,7 @@ Module coverage 16.0 -> 17.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | sale_purchase_stock                               | Nothing to do        |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| sale_quotation_builder                      |                      |Merged into sale_pdf_quote_builder.              |
+| |del| sale_quotation_builder                      | Done                 |Forward-ported into OCA/sale-reporting           |
```
pedrobaeza: We should leave blank the past of "Done", do the module is not installed in the DB for the CI.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #5001, opened 2025-05-08
[18.0][OU-ADD] iap

MiquelRForgeFlow:
.


openupgrade_scripts/scripts/iap/18.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,34 @@
...
+iap          / iap.account              / service_id (many2one)         : NEW relation: iap.service, required
+iap          / iap.service              / description (char)            : NEW required
+iap          / iap.service              / integer_balance (boolean)     : NEW required
+iap          / iap.service              / name (char)                   : NEW required
+iap          / iap.service              / technical_name (char)         : NEW required
+iap          / iap.service              / unit_name (char)              : NEW required
+# DONE: pre-migration: create new service for each iap.account record and link them
+
+---XML records in module 'iap'---
+NEW iap.service: iap.iap_service_reveal
```
MiquelRForgeFlow, line 31: not sure if I should use this service instead of create new ones...
https://github.com/odoo/odoo/commit/ae35752ce7aeb6e62e7ef0ec46c15dd9d23571cc

openupgrade_scripts/scripts/iap/18.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,48 @@
...
+    openupgrade.rename_models(env.cr, model_renames)
+    openupgrade.rename_tables(env.cr, table_renames)
+    openupgrade.add_columns(env, [("iap.service", "integer_balance", "boolean")])
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE iap_account ia
+        SET name = service_name
+        WHERE COALESCE(name, '') != ''""",
+    )
+    openupgrade.logged_query(
```
hbrunn: this needs to go to post-migration

hbrunn (changes requested):

openupgrade_scripts/scripts/iap/18.0.1.1/pre-migration.py
MiquelRForgeFlow: why?

hbrunn: https://github.com/OCA/OpenUpgrade/actions/runs/14910707744/job/41884075008?pr=5001#step:13:666

MiquelRForgeFlow: Thanks, it was a bit incomplete 😅

pedrobaeza:
I uninstall this module from every instance.

MiquelRForgeFlow:
> I uninstall this module from every instance.

It's in the paid oca project... 🤷‍♂️

pedrobaeza:
As @hbrunn has started the review, I let him to finish it.

openupgrade_scripts/scripts/iap/18.0.1.1/upgrade_analysis_work.txt
hbrunn: migration scripts of subsequent addons like sms will have to create xmlids for the [services](https://github.com/OCA/OCB/blob/18.0/addons/sms/data/iap_service_data.xml) they define in case they exist already as a result of this addon's migration script

hbrunn (changes requested):
reading [the code](https://github.com/OCA/OCB/blob/18.0/addons/iap/models/iap_account.py#L53-L54) again I think you should do nothing with the transient model (those records might or might not exist), just rename iap.account#service_name to a legacy column in pre-migration, and use that to fill iap.service in post-migration. Then call iap.account#_add_sql_constraints() to create the requiredness constraint for service_id.

Your migration scripts are very pre-migration heavy anyways in situations where I'd let the ORM do more of the work and do my stuff in post-migration. Am I missing something here?

openupgrade_scripts/scripts/iap/18.0.1.1/upgrade_analysis_work.txt
hbrunn: this is what I mean: https://github.com/OCA/OpenUpgrade/pull/5055

MiquelRForgeFlow:
> the transient model (those records might or might not exist)

In fact, they don't exist because we cleaned them in `base` pre-migration.

openupgrade_scripts/scripts/iap/18.0.1.1/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+            FROM iap_account ia
+            WHERE ia.service_id IS NULL
+            RETURNING id, ia.id as account_id
+        )
+        UPDATE iap_account ia
+        SET service_id = sub.id
+        FROM sub
+        WHERE ia.id = sub.account_id
+        """,
+    )
```
hbrunn: this needs grouping by `openupgrade.get_legacy_name("service_name")`, and you should do the matching by service_name/technical_name

openupgrade_scripts/scripts/iap/18.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+        WHERE COALESCE(name, '') != ''
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DELETE FROM iap_account
+        WHERE service_name = 'reveal'
+        """,
+    )
```
hbrunn: instead of deleting those accounts, you should rather exclude `reveal` in the post migration query and link accounts with this service name to the service created in xml data

hbrunn (changes requested):
I wonder if it might be simpler to not touch iap_service at all in pre/post, have the service records be created by xml data, and add a end-migration script that sets iap.account#service_id based on iap.service#technical_name and iap_account#{get_legacy_name('service_name')}.

Then there would be nothing to do for the iap related modules I've been working on this morning

openupgrade_scripts/scripts/iap/18.0.1.1/end-migration.py
```diff
@@ -0,0 +1,40 @@
...
+            WHERE ia.service_id IS NULL
+                AND ia.{openupgrade.get_legacy_name("service_name")} IS NOT NULL
+            RETURNING id, technical_name
+        )
+        UPDATE iap_account ia
+        SET service_id = sub.id
+        FROM sub
+        WHERE ia.{openupgrade.get_legacy_name("service_name")} = sub.technical_name
+        """,
+    )
```
hbrunn:
```suggestion
```

whatever service exists should have been added by sms, partner_autocomplete etc, right?

MiquelRForgeFlow: Maybe you have old iap accounts with an old service_name that doesn't match any of these new services.

hbrunn: but this will be a defunct account then if there's no module handing this service, so I don't see why we should create bogus service entries for them

MiquelRForgeFlow: haha I don't know, but you have a point.

MiquelRForgeFlow: I will delete them instead.

MiquelRForgeFlow:
@hbrunn ready

hbrunn (approved):

PR #5004, opened 2025-05-09
[17.0][OU-IMP] website_event: Add script to insert default event questions for old events

pilarvargas-tecnativa:
In version 16, the default questions in the registration form, such as ‘Name’, ‘Email’ and ‘Phone’, were directly visible in the registration form view for users. However, in Odoo v17, although those questions are automatically created in new events as records in the event.question template, they are not inserted in existing events, causing a lack of consistency in the registration form.
This script solves that problem by adding those questions by default to all existing events, ensuring that they have the same behaviour as new events. For the translations, instead of defining them manually, the script takes advantage of existing translations in the ir.model.fields table for the name, email and phone fields of the res.partner model.

cc @Tecnativa TT51546

@pedrobaeza @victoralmau please review


victoralmau (approved):
The change seems coherent.

pedrobaeza (approved):
Well done!

PR #4958, opened 2025-04-24
[18.0][OU-ADD] payment

MiquelRForgeFlow:
Depends on:

- [x] https://github.com/OCA/openupgradelib/pull/417

PR #4931, opened 2025-04-16
[18.0][OU-ADD] account

MiquelRForgeFlow:
Depends on:

- [x] https://github.com/OCA/openupgradelib/pull/412

<dependency>Depends on : 

- product : #4839
- analytic : #4841
- onboarding : #4944
- portal : #4949
- digest : #4950</dependency>

PR #5005, opened 2025-05-09
[18.0][OU-ADD] account_payment

MiquelRForgeFlow:
<dependency>Depends on: 

- account : #4931
- payment : #4958</dependency>


MiquelRForgeFlow:
/ocabot migration account_payment

<dependency>Depends on : 

hbrunn (approved):

PR #5009, opened 2025-05-10
[16.0][OU-FIX] base_address_extended: regexp fix

rvalyi:
when trying to run OpenUpgrade to 16.0, I got this error:

```
    )   and I get this error:   File "/odoo/external-src/openupgrade/openupgrade_framework/odoo_patch/odoo/modules/migration.py", line 18, in migrate_module
    MigrationManager.migrate_module._original_method(self, pkg, stage)
  File "/odoo/src/odoo/modules/migration.py", line 189, in migrate_module
  File "/env/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 2400, in wrapped_function
    func(
  File "/odoo/external-src/openupgrade/openupgrade_scripts/scripts/base_address_extended/16.0.1.1/pre-migration.py", line 41, in migrate
    openupgrade.logged_query(
  File "/env/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 1750, in logged_query
    cr.execute(query, args)
  File "/odoo/src/odoo/sql_db.py", line 321, in execute
psycopg2.errors.InvalidRegularExpression: invalid regular expression: invalid escape \ sequence 
```

Deepseek suggested me several solutions and the one that could run for me was the one of this patch. Deepseek explanations:

-  Used E'' string literal which properly interprets backslash escapes
-  Doubled the backslashes for Python string interpretation
-  Simplified digit matching with \d

Disclaimer: I made no functional test. I will see if I have data I could check eventually.

versions:

Postgres server 13
psycopg2==2.9.5
Python 3.10.12

cc @etobella

### DIFF
--- a/openupgrade_scripts/scripts/base_address_extended/16.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/base_address_extended/16.0.1.1/pre-migration.py
@@ -49,7 +49,7 @@ def migrate(env, _version):
         FROM (
             SELECT
                 id,
-                regexp_match(street, '^(.*?)(?:\\s([0-9][0-9\\S]*))?(?: - (.+))?$') AS rgx
+                regexp_match(street, E'^(.*?)(?:\\s+(\\d[\\d\\S]*))?(?:\\s+-\\s+(.+))?$') AS rgx
             FROM res_partner
             WHERE street IS NOT NULL
         ) AS sub


etobella:
It is weird, as this worked for me a month ago. Maybe it depends on the postgres version 🤔?

MiquelRForgeFlow:
@etobella which postgres version you used?

etobella:
15 in my case

openupgrade_scripts/scripts/base_address_extended/16.0.1.1/pre-migration.py
```diff
@@ -49,7 +49,7 @@ def migrate(env, _version):
         FROM (
             SELECT
                 id,
-                regexp_match(street, '^(.*?)(?:\\s([0-9][0-9\\S]*))?(?: - (.+))?$') AS rgx
```
MiquelRForgeFlow, line 52: I think the problem is with `\S` inside the bracket, see:

In v13: https://www.postgresql.org/docs/13/functions-matching.html
```
\S 	[^[:space:]]
Within bracket expressions, \d, \s, and \w lose their outer brackets, and \D, \S, and \W are illegal. (So, for example, [a-c\d] is equivalent to [a-c[:digit:]]. Also, [a-c\D], which is equivalent to [a-c^[:digit:]], is illegal.)
```

In v15: https://www.postgresql.org/docs/15/functions-matching.html
```
\S 	matches any non-whitespace character, like [^[:space:]]
The class-shorthand escapes also work within bracket expressions, although the definitions shown above are not quite syntactically valid in that context. For example, [a-c\d] is equivalent to [a-c[:digit:]].
```

etobella: Following @MiquelRForgeFlow comments, we can use `regexp_match(street, '^(.*?)(?:\\s([0-9][0-9^\\s]*))?(?: - (.+))?$') AS rgx`

When tested, both queries where not giving the same result (tested on 13)

etobella: and was the same result expected on postgres 15 (with the original regex and the new regex)

MiquelRForgeFlow: @rvalyi will you update the code?

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/5078.

PR #5015, opened 2025-05-12
[18.0][OU-ADD] crm

BhaveshHeliconia:


etobella:
Shouldn't you apply the no-update changes?

https://github.com/OCA/OpenUpgrade/blob/18.0/openupgrade_scripts/scripts/crm/18.0.1.8/noupdate_changes.xml

BhaveshHeliconia:
@etobella, Just wondering — could you please elaborate a bit in case I’m missing something?

etobella:
openupgrade says that there a change that will not be changed automatically when updating because it is a noupdate record. Usually, this needs are applied automatically on the migration

BhaveshHeliconia:
Thanks for the clarification, @etobella. I understand the point about the noupdate record changes. Just to confirm — did I miss applying anything specific in this PR that you'd recommend I adjust?

etobella:
Well, on the folder of the scripts there is an xml file with noupdate_changes. You should read it and decide if we need an script to import it (there are examples in other OpenUpgrade scripts)

pedrobaeza:
@BhaveshHeliconia you should document yourself a bit and check previous works for doing an optimal process in this repository and avoid to make reviewers to spend time teaching you about the basics. Check https://oca.github.io/OpenUpgrade/080_migration_script_development.html

BhaveshHeliconia:
@pedrobaeza, I’ve reviewed the previous implementations and documentation to align with the expected process for this repository. However, I’d appreciate it if you could confirm whether I’m missing anything, just to be sure I’m on the right track and not overlooking any basics.

I always value the time and feedback provided on my PRs and truly appreciate your guidance in helping me improve.

pedrobaeza:
I don't think so. This is not "Nothing to do" AFAIK, and you are not filling `openupgrade_analysis_work.txt`, not loading noupdate_changes.xml...

https://oca.github.io/OpenUpgrade/use_cases/noupdate_xml_entry_changed.html

BhaveshHeliconia:
@pedrobaeza, I checked the existing PRs and have used this one as a reference:  

- https://github.com/OCA/OpenUpgrade/pull/4976

but that module not referenced here: https://github.com/OCA/OpenUpgrade/tree/18.0/openupgrade_scripts/scripts
So  I'll go ahead and update it now based on your feedback.

BhaveshHeliconia:
@etobella, I verified in Odoo 18, and just like in version 17, the record gets created successfully even without loading noupdate changes. That’s why I didn’t load the file here.

odoo 17
Screenshot from 2025-05-14 12-16-22

odoo 18
Screenshot from 2025-05-14 12-17-11

MiquelRForgeFlow:
@BhaveshHeliconia an example on how to load the noupdate_changes: https://github.com/OCA/OpenUpgrade/pull/5039

Please, check all your other Pull Requests and load it in the ones they have this file. In cases where the content of the file is not needed, you should modify the file adding `<!--` and `-->` inside the content.

openupgrade_scripts/scripts/crm/18.0.1.8/post-migrate.py
```diff
@@ -0,0 +1,11 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "crm", "18.0.1.8/noupdate_changes.xml")
+    openupgrade.delete_record_translations(
```
MiquelRForgeFlow: don't use this, only used for email templates or records that change their translation fields

BhaveshHeliconia:
@MiquelRForgeFlow, Please Review.it's done.

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
@BhaveshHeliconia thanks, notice next time that you have to change the commit and PR title, as it isn't "nothing to do" anymore.

PR #5021, opened 2025-05-13
[17.0][OU-FIX] point_of_sale: group account move, duplicate with more…

FernandoRomera:
… than one pos.order.line


legalsylvain (approved):
Makes sense.

pedrobaeza (approved):

PR #4459, opened 2024-06-13
[17.0][OU-ADD] base_automation: migration to 17.0

daiduongnguyen-odoo:

PR #5028, opened 2025-05-13
[17.0][OU-ADD] base_automation: migration to 17.0

petrus-v:
attempt to take care some reviews from https://github.com/OCA/OpenUpgrade/pull/4459

- [x] https://github.com/OCA/OpenUpgrade/pull/4459#discussion_r1791552948 done by https://github.com/OCA/OpenUpgrade/pull/5028/commits/9490f090cbddea8129a841872ed04d234efaa491
- [x] https://github.com/OCA/OpenUpgrade/pull/4459#discussion_r1976230947 done by https://github.com/OCA/OpenUpgrade/pull/5028/commits/9f394bd4935163b50b4c11c65e2cac83fc75fb21
- [x] https://github.com/OCA/OpenUpgrade/pull/4459#discussion_r1986749830 done by https://github.com/OCA/OpenUpgrade/pull/5028/commits/2dc89db0ecb2bae36c83afd5ba81b3b665bf21dd
- [x] https://github.com/OCA/OpenUpgrade/pull/4459#discussion_r1986754025 not yet : I've only 4 records in my context and not sure what would be the proper sql code setup values as far I understand only create column with Null values should do the trick, **any advices welcomes** done by https://github.com/OCA/OpenUpgrade/pull/5028/commits/7857148f586a36cae751eceaa89c1ab1c90d4826 (thanks Miquel for the help)
- [x] https://github.com/OCA/OpenUpgrade/pull/4459#discussion_r1986755757 I'm wondering what's wrong moving forwards ? it is to gives more times to sub modules to change code as well and take care new values ? done by https://github.com/OCA/OpenUpgrade/pull/5028/commits/2d3653c893989be5c667d36230b5eed8ec2253d3


openupgrade_scripts/scripts/base_automation/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,84 @@
...
+        env.cr,
+        env["base.automation"],
+        "base_automation",
+        "action_server_ids",
+        "action_server_id",
+    )
+
+
+def _base_automation_rotate_webhook_uuid(env):
+    env["base.automation"].with_context(active_test=False).action_rotate_webhook_uuid()
```
MiquelRForgeFlow:
```suggestion
    env["base.automation"].with_context(active_test=False).search([]).action_rotate_webhook_uuid()
```

MiquelRForgeFlow (commented):
> I understand only create column with Null values should do the trick, **any advices welcomes**

 

```python
_new_columns = [
    ("base.automation", "trg_field_ref", "many2one_reference"),
    ("base.automation", "trg_selection_field_id", "many2one"),
]

openupgrade.add_columns(env, _new_columns)
```

> I'm wondering what's wrong moving forwards ? it is to gives more times to sub modules to change code as well and take care new values ?

I don't know. That's something only @pedrobaeza should answer, as he made that comment.

petrus-v:
@MiquelRForgeFlow  thanks for review and advices.

> > I understand only create column with Null values should do the trick, **any advices welcomes**
> 
> ```python
> _new_columns = [
>     ("base.automation", "trg_field_ref", "many2one_reference"),
>     ("base.automation", "trg_selection_field_id", "many2one"),
> ]
> 
> openupgrade.add_columns(env, _new_columns)
> ```

Here my question was not how to create those empty fields, but: if I only create those fields leaving them empty is it enough in this specific case ?

As far I understand, generally we do that for performance reason by populate the field using SQL query re-implementing the business python code in SQL. here the python code that compute `trg_selection_field_id` was not clear to me what it really does:
```python
    @api.depends('trigger', 'trigger_field_ids')
    def _compute_trg_selection_field_id(self):
        to_reset = self.filtered(lambda a: a.trigger not in ['on_priority_set', 'on_state_set'] or len(a.trigger_field_ids) != 1)
        to_reset.trg_selection_field_id = False
        for automation in (self - to_reset):
            # always re-assign to an empty value to make sure we have no discrepencies
            automation.trg_selection_field_id = self.env['ir.model.fields.selection']
```

as far I understand the value is False or an empty recordset which is also a falsy value or I missed interpret the following  instruction ?
```
>>> self.env['ir.model.fields.selection'].id
False
```

also when [I read the doc](https://oca.github.io/OpenUpgrade/use_cases/field_fast_precomputation.html#notes)

> These scripts are about optimization. As a contributor of openupgrade for a module, if you do not have a problem of excessive > duration, you can propose migration scripts _without_ such optimizations, especially when the SQL queries are complex to write.
>
>    Another contributor can always propose a PR for performance improvement, if he faces this problem.

And seen only 4 records in my database I was thinking should I take this risks !

MiquelRForgeFlow:
> Here my question was not how to create those empty fields, but: if I only create those fields leaving them empty is it enough in this specific case ?

Yeah, in this specific case, leaving them empty is fine 😊 (it's a new functionality, they are supposed to be filled using the form view)

MiquelRForgeFlow (approved):
LGTM

openupgrade_scripts/scripts/base_automation/17.0.1.0/end-migration.py
```diff
@@ -0,0 +1,21 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+def _ir_act_server_update_name_if_base_automation(env):
+    # Liste des langues actives dans la base
```
pedrobaeza: Please put comments in English.

openupgrade_scripts/scripts/base_automation/17.0.1.0/end-migration.py
```diff
@@ -0,0 +1,21 @@
...
+    langs = [lang.code for lang in env["res.lang"].search([("active", "=", True)])]
```
pedrobaeza:
```suggestion
    langs = env["res.lang"].search([]).mapped("code")
```

openupgrade_scripts/scripts/base_automation/17.0.1.0/end-migration.py
```diff
@@ -0,0 +1,21 @@
...
+
+    act_servers = (
+        env["ir.actions.server"]
+        .with_context(active_test=False)
+        .search([("base_automation_id", "!=", False)])
+    )
+    if act_servers:
+        for lang in langs:
+            act_servers.with_context(lang=lang)._compute_name()
```
pedrobaeza, line 15: Are you sure this stores in the same field all the translations?

pedrobaeza:
And please reduce the commit history to one commit.

openupgrade_scripts/scripts/base_automation/17.0.1.0/end-migration.py
MiquelRForgeFlow: It seems correct. It is handled by ORM the storing.

pedrobaeza:
Well, 2 commits, the other and yours, but don't put so many commits of your authorship.

openupgrade_scripts/scripts/base_automation/17.0.1.0/end-migration.py
petrus-v: double checked, it works, here result on base_automation after the migration
```sql
oca-ba=# \x
Affichage étendu activé.
oca-ba=# select * from base_automation;
-[ RECORD 1 ]----------+--------------------------------------------------------------------------------------
id                     | 1
action_server_id       | 135
trg_date_id            | 
trg_date_range         | 0
trg_date_calendar_id   | 
create_uid             | 2
write_uid              | 1
trigger                | on_create
trg_date_range_type    | 
filter_pre_domain      | 
filter_domain          | 
active                 | t
last_run               | 
create_date            | 2025-05-27 15:34:09.526755
write_date             | 2025-05-27 16:24:22.906395
trg_field_ref          | 0
trg_selection_field_id | 
model_id               | 85
name                   | {"en_US": "PV TEST ENGLISH", "es_ES": "TEST PV ESPANOL", "fr_FR": "TEST PV Français"}
webhook_uuid           | b4976149-0bc5-4993-b893-974b3cd316e4
record_getter          | model.env[payload.get('_model')].browse(int(payload.get('_id')))
description            | 
``

petrus-v:
I've squashed commits ! (I kept it in the mean time with the hope to help reviews)

pedrobaeza (approved):

PR #5029, opened 2025-05-13
[18.0] [OU-ADD] apriori: document_page_group merged into document_page_access_group

luisDIXMIT:
Due to https://github.com/OCA/knowledge/issues/502 document_page_group is not needed and can be merged into document_page_access_group which has already been migrated

@etobella


etobella (approved):

pedrobaeza:
Isn't there any XML-ID renaming or so?

etobella:
Well, the original module has no noupdate xml files. However, it might be necessary to add a small openupgrade script. However, the right place would be on the module, isn't it?

pedrobaeza (approved):
Yeah, we can put it in the module itself

PR #5032, opened 2025-05-13
[17.0][OU-FIX] hr_attendance: Add the 'User: Read his own attendances' permission to the appropriate users.

victoralmau:
Add the 'User: Read his own attendances' permission to the appropriate users.

The new permission 'User: Read his own attendances' is required for all existing users, otherwise, when trying to access 'Last month' smart-button of their employee tab they will get an access error.

The old `hr_attendance.group_hr_attendance_kiosk` permission does not replace this new one exactly.
https://github.com/odoo/odoo/blob/97d8fdbcd363e17569ac420049989959258aa0b7/addons/hr_attendance/security/hr_attendance_security.xml#L14

**Before**
ejemplo

@Tecnativa


pedrobaeza:
According to the code you link, you only need to remove the renaming. Please check CI, as this may have associated tests.

pedrobaeza (approved):

Issue #4550, opened 2024-09-12
[15.0][mail]Duplicated entry in update mail_group_moderation_rel


dansanti:
If 2 mail channels haves the same moderator, and when is migrated to mail_group, the group id could by the same of channel, then when update is performed , could cause  unique constraint error

## Module
mail

## Describe the bug
current 14.0  table mail_channel_moderator_rel:
mail_channel_id, res_users_id
7,	2
1,	2

when migration is performed , may create mail_channel 7 with mail_group id 1, then  the table will be renamed to mail_group_moderator_rel and  mail_channel_id will be renamed to mail_group_id, so, in the SQL update mail_group_id 7 will be updated according old channel_id 1 . This cause unique error, because already exists that tuple 1, 2 ( not update yet)

here the update line:

https://github.com/OCA/OpenUpgrade/blob/9af4744ada76ae405cd290d04c052360ed66acb9/openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py#L231

pedrobaeza:
Please propose a patch to fix this.

PR #5040, opened 2025-05-15
[15.0][OU-FIX] mail: avoid duplication error in mail_group_moderation_rel query

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/issues/4550.

### DIFF
--- a/openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
+++ b/openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
@@ -170,13 +170,12 @@ def migration_to_mail_group(env):
             ("mail_channel_moderator_rel", "mail_group_moderator_rel"),
         ],
     )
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/mail/15.0.1.5/pre-migration.py
```diff
@@ -234,7 +233,16 @@ def migration_to_mail_group(env):
         UPDATE mail_group_moderator_rel rel
         SET mail_group_id = mg.id
         FROM mail_group mg
-        WHERE mg.old_channel_id = rel.mail_group_id""",
+        WHERE mg.old_channel_id = rel.mail_channel_id""",
+    )
+    openupgrade.logged_query(
```
pedrobaeza, line 238: Please only one query per call.

pedrobaeza (approved):

PR #4761, opened 2025-02-21
[17.0][MIG] repair

MiquelRForgeFlow:
I will do a new analysis to better match repair lines with stock moves.

PR #5044, opened 2025-05-16
[17.0][OU-ADD] repair: Migration to 17.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/OpenUpgrade/pull/4761

TT49887

Note: This migration does not create a Sales Order for Repair Orders that are in progress and not invoiced. This decision should be left to the user.

TODO: Define how to handle Repair Orders that were invoiced before the repair was completed. The relationship between these two models no longer exists.
Is it necessary to generate a Sales Order to cover this case and link it to the Repair Order? Or should it be handled another way?


openupgrade_scripts/scripts/repair/17.0.1.0/post-migration.py
```diff
@@ -18,77 +21,198 @@ def fill_product_template_create_repair(env):
...
+            JOIN product_product pp ON pp.id = rl.product_id
+            JOIN product_template pt ON pt.id = pp.product_tmpl_id
+            JOIN uom_uom rl_uom ON rl_uom.id = rl.product_uom
+            JOIN uom_uom pt_uom ON pt_uom.id = pt.uom_id
         WHERE rl.move_id IS NULL AND rl.type IS NOT NULL
         """,
     )
 
 
-def create_default_repair_type_for_all_warehouses(env):
```
MiquelRForgeFlow: why don't use this method? It was correct.

carlos-lopez-tecnativa: Because I use the Odoo method that they use when the module is installed:
https://github.com/odoo/odoo/blob/4baf55a5d108063b0b60beddf332b6ae367f1871/addons/repair/__init__.py#L10

Why repeat the code if Odoo already does it?

MiquelRForgeFlow: First, because my method was not wrong, so no need to change something that was not incorrect. Second, because odoo method is updating all picking type values, not only creating the new picking type, and we should avoid that.

carlos-lopez-tecnativa: Ok, I understand, my apologies. I updated the code to restore your function. I just added the domain `[("repair_type_id", "=", False)]` because Odoo loads the picking type for the main warehouse via `data`, and your script creates the record again, so I end up with two picking types for this warehouse. Please let me know if this is OK with you.

openupgrade_scripts/scripts/repair/17.0.1.0/end-migration.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Remove the temporary location and picking type
+    # created during the migration
+    env["stock.location"].search(
```
MiquelRForgeFlow: why do this here instead of doing in `fill_repair_picking_type()`?

carlos-lopez-tecnativa: Done

MiquelRForgeFlow (approved):
LGTM

openupgrade_scripts/scripts/repair/17.0.1.0/post-migration.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    Map the states of the repair orders to the new ones.
+    The mapping is as follows:
+    - ready -> confirmed. Order not repaired but already invoiced.
+        TODO: Decide how to handle the invoice
+            when it has been created before the repair is completed.
+    - 2binvoiced and not repaired -> confirmed. Order not repaired and not invoiced
+    - 2binvoiced and repaired -> done. Order repaired but not invoiced
+    """
+    openupgrade.logged_query(
```
pedrobaeza: Use `map_values` openupgradelib method, and you have to copy the column in pre for preserving old values.

carlos-lopez-tecnativa: Done

pedrobaeza: You should do now the SQLs and map_values with the other column

pedrobaeza (approved):

PR #5007, opened 2025-05-09
[18.0][OU-ADD] sale

BhaveshHeliconia:

PR #5045, opened 2025-05-19
[18.0][OU-ADD] sale_crm: Nothing to do

BhaveshHeliconia:
Depends on 

- https://github.com/OCA/OpenUpgrade/pull/5007


legalsylvain:
/ocabot migration sale_crm

<dependency>Depends on : 

- sale : #5071</dependency>

PR #5047, opened 2025-05-19
[15.0][OU-ADD] repair: Migration to 15.0

carlos-lopez-tecnativa:
TT47859


openupgrade_scripts/scripts/repair/15.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,25 @@
...
+
+repair       / repair.order             / priority (selection)          : NEW selection_keys: ['0', '1'], hasdefault: default
+# NOTHING TO DO: new field for a new feature. The default value is coherent and preserves the behavior of the previous version
+
+repair       / repair.order             / internal_notes (text)         : type is now 'html' ('text')
+repair       / repair.order             / quotation_notes (text)        : type is now 'html' ('text')
+# DONE: pre-migration: convert text to html
+
+repair       / repair.order             / state (selection)             : selection_keys is now '['2binvoiced', 'cancel', 'confirmed', 'done', 'draft', 'ready', 'under_repair']' ('['2binvoiced', 'cancel', 'confirmed', 'done', 'draft', 'invoice_except', 'ready', 'under_repair']')
+# NOTHING TO DO: the option invoice_except is unused in the code
```
MiquelRForgeFlow: merge `invoice_except` into `2binvoiced`.

The `invoice_except` state was used to indicate that the creation of the invoice failed. I don't know in which version was deprecated.

carlos-lopez-tecnativa: I searched for the use of this option, but up to Odoo 12, this option was never used. So, at this point, I think the value is not present in any database. I believe changing the value is unnecessary. @pedrobaeza what do you think?

MiquelRForgeFlow: I found https://github.com/odoo/odoo/commit/53b4c561beed2a2943f8f2ff4b61658c4d0f3f1f. It seems in v6 was used, but not anymore in v7. I don't think someone will migrate directly from v6 up to v15... But who knows...

carlos-lopez-tecnativa: Done

pedrobaeza (approved):

MiquelRForgeFlow (approved):
LGTM

PR #5048, opened 2025-05-19
[17.0][FIX] analytic: rename account.analytic.plan#default_applicability

### LINKED ISSUE #5041: [16.0 -> 17.0 -> 18.0] Cannot load the analytic module Openupgrade
## Module
Assuming the problem is with the analytic module.

## Describe the bug
While upgrading my Odoo database from version 16.0 to 17.0 and then to 18.0, I encountered a critical error that prevents the web interface from loading. During the upgrade process, although the upgrade scripts complete without initial errors, the server fails to initialize the database.

Upon reviewing the logs, the following traceback appears:

```
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type json
DETAIL:  Token "optional" is invalid.
CONTEXT:  JSON data, line 1: optional
```

## To Reproduce
**Affected versions**:

1. Migrate the database using [OpenUpgrade](https://github.com/OCA/OpenUpgrade) from Odoo 16.0 to 17.0.
2. Then migrate the same database from Odoo 17.0 to 18.0 using OpenUpgrade.
3. Start the Odoo 18.0 server with the upgraded database.
4. Observe that the server fails to initialize and logs a psycopg2.errors.InvalidTextRepresentation error related to invalid JSON syntax.

hbrunn:
as leaving the column will cause problems in the v18 migration.

Then we can also use openupgrade.convert_to_company_dependent

Fixes #5041

### DIFF
--- a/openupgrade_scripts/scripts/analytic/17.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/analytic/17.0.1.2/post-migration.py
@@ -60,19 +60,12 @@ def _analytic_plan_update_applicability_into_property(env):
     in all companies as the company_id field was shifted from account.analytic.plan
     to account.analytic.applicability
     """
-    env.cr.execute(
-        """
-        SELECT id, default_applicability FROM account_analytic_plan
-        WHERE default_applicability != 'optional'
-        """
+    openupgrade.convert_to_company_dependent(
+        env,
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Thanks !

pedrobaeza (approved):

PR #5059, opened 2025-05-26
[17.0][OU-FIX] project_todo: if the OCA project_task_code was install…

FernandoRomera:
…ed, code is required

### DIFF
--- a/openupgrade_scripts/scripts/project_todo/17.0.1.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/project_todo/17.0.1.0/pre-migration.py
@@ -23,19 +23,23 @@ def _convert_note_note_to_project_task(env):
     openupgrade.logged_query(
         env.cr, "ALTER TABLE project_task ADD COLUMN old_note_id INTEGER"
     )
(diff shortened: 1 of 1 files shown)


hbrunn (approved):
you could have used f-strings and do something like

```python
INSERT ... {column_exists and ", code" or ""}
SELECT ... {column_exists and "('OU' || id::VARCHAR)" or ""}
```

to make it clearer that this is the same query with just the code added, but I won't quibble about that

FernandoRomera:
@hbrunn 
Done, thanks.

pedrobaeza:
Shouldn't the module `project_task_code` covers the case when `project_todo` is installed? What is happening with the "todos" created once you are on 17?

FernandoRomera:
@pedrobaeza 
Ok, module project_task_code needs to be revised, but we can merge this pr. To make the database migration work.

MiquelRForgeFlow (approved):
LGTM

FernandoRomera:
@MiquelRForgeFlow 
Can you mege this PR, please?.

PR #5062, opened 2025-05-26
[9.0][FIX] purchase.trans_confirmed_double_gt: may not exist in db

sbejaoui:


MiquelRForgeFlow:
> purchase.trans_confirmed_double_gt: may not exist in dbtrans_confirmed_double_gt

why it should not exist?

sbejaoui:
> > purchase.trans_confirmed_double_gt: may not exist in dbtrans_confirmed_double_gt
> 
> why it should not exist?

removed by users, 
image

PR #5067, opened 2025-05-26
[OU-FIX] product: avoid reserved SQL keyword "sequence"

MaPaLo76:
When running OpenUpgrade 18.0 on PostgreSQL, queries using  `row_number(...) sequence` raise a syntax error due to unquoted alias names with postgres version 13.0

This PR fixes:
- `sequence` → `"sequence"`

to ensure compatibility with PostgreSQL reserved keywords and avoid `syntax error at or near ...`.

This patch unblocks migration from Odoo 17.0 to 18.0 for affected modules (, `product`) with postgres 13.0.

### DIFF
--- a/openupgrade_scripts/scripts/product/18.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/product/18.0.1.2/post-migration.py
@@ -12,7 +12,7 @@ def product_document_sequence(env):
         env.cr,
         """
         UPDATE product_document
-        SET sequence=seq.sequence
+        SET sequence=seq.seq
         FROM
         (
             SELECT
--- a/openupgrade_scripts/scripts/product/18.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/product/18.0.1.2/post-migration.py
@@ -12,22 +12,23 @@ def product_document_sequence(env):
         env.cr,
         """
         UPDATE product_document
-        SET sequence=sequence.sequence
+        SET sequence = "sequence".sequence
         FROM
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Maybe it's better to simply not call it `sequence`, but `sub` or any other name.

hbrunn (changes requested):
please write a commit message describing what the change does (it has nothing to do with json_object_agg btw) and remove the trailing semicolon.

the patch also contains a lot of unnecessary changes, but i don't mind them.

you might also consider @pedrobaeza's proposal to just not use a non-reserved keyword here

MiquelRForgeFlow:
I also think the best way is not using the reserved keywords.

Also, commit title should be `[OU-FIX] product: brief explanation`.

MaPaLo76:
sorry for the inconvenience, i made a mistake during this PR. I thought I was in my fork.

I'm relatively inexperienced at submitting pull requests. This is my second one. The first one, in openupgradelib, worked better.
Do I need to do anything else?

openupgrade_scripts/scripts/product/18.0.1.2/post-migration.py
```diff
@@ -12,21 +12,21 @@ def product_document_sequence(env):
         env.cr,
         """
         UPDATE product_document
-        SET sequence=sequence.sequence
+        SET sequence=seq.sequence
```
MiquelRForgeFlow:
```suggestion
        SET sequence=seq.seq
```

PR #5071, opened 2025-05-27
[18.0][OU-ADD] sale

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration sale

<dependency>Depends on : 

- account_payment : #5005</dependency>

PR #5072, opened 2025-05-27
[18.0][OU-ADD] stock_account

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration stock_account

<dependency>Depends on : 

- stock : #4957

PR #5073, opened 2025-05-27
[18.0][OU-ADD] l10n_es

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration l10n_es

<dependency>Depends on : 

- account : #4931
- base_iban : #5081
- base_vat : #5080</dependency>

PR #5086, opened 2025-06-02
[17.0][OU-FIX] analytic: search for existing default project analytic plan

### LINKED ISSUE #5085: [17.0] analytic: creation of a duplicate project plan
## Module
analytic

## Describe the bug
OpenUpgrade script would recreate a new analytic plan for projects, even if one already existed, causing a mess afterwards in analytic, since you cannot delete the new plan (because is the default plan, used in account_id in account_analytic_line and you get the old plan as x_plan1_id column in account_analytic_line)

## To Reproduce
**Affected versions**: 17.0

1. Migrate Odoo v16.0 database with hr_timesheet installed but no demo data to v17.0

**Expected behavior**
When you already have a default plan for projects configured in v16, you do not need to create a new one.

**Additional context**
In v16, without demo, when you install project and hr_timesheet, you get a default analytic plan created and used.
You can then find this plan in field analytic_plan_id of res_config_settings from Project page.
This field is related to company_id.analytic_plan_id which is computed here : https://github.com/odoo/odoo/blob/685931dccf93d9c5d27282273583f25101ead0cc/addons/project/models/company.py#L17, if none exist it is created by calling self.env['account.analytic.plan'].with_company(company)._get_default()

This default plan does not have an external id and is usually not configured in ir.config_parameter analytic.project_plan key, therefore when you run openupgrade script, while upgrading analytic, in pre-migration, you will not find it and create a new one here : https://github.com/OCA/OpenUpgrade/blob/e5e8c83106ee8fb8812909e02d9d35d2e6aeb5c8/openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py#L28 

I suppose that if we set analytic.analytic_plan_projects XML id on that plan, it would therefore not be recreated by OpenUpgrade script since found by the search here : https://github.com/OCA/OpenUpgrade/blob/e5e8c83106ee8fb8812909e02d9d35d2e6aeb5c8/openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py#L21

Also, in v16 you can have a different default plan per company, which is not possible anymore in v17 which is using the unique default ir.config_parameter "analytic_plan_projects". Therefore I am not sure which plan should be used there in case you have more than 1 company using different default plans ?

remi-filament:
Fixes #5085

### DIFF
--- a/openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
@@ -12,6 +12,10 @@ def _fill_config_parameter_analytic_project_plan(env):
     If not, we check if there's already an existing ir.model.data entry for projects
     according standard data (also pre-filled externally).
 
+    If not, we look for either a specific plan set on a company or first plan
+    (as per v16 project module, company.py, _compute_analytic_plan_id)
+    and we set external ID accordingly.
+
     Finally, if not found, we will put the next available ID, as regular ORM update
     process will load the record "analytic.analytic_plan_projects", and creates the
     record that belongs to the "Projects" plan.
@@ -22,6 +26,27 @@ def _fill_config_parameter_analytic_project_plan(env):
         [("module", "=", "analytic"), ("name", "=", "analytic_plan_projects")]
     )
     plan_id = imd.res_id
+    if not plan_id:
+        default_plans = (
+            env["ir.config_parameter"]
+            .search([("key", "like", "default_analytic_plan_id_%")], order="value")
+            .mapped("value")
+        )
+        plan_id = default_plans[0] if default_plans else False
+        if not plan_id:
+            env.cr.execute("SELECT id FROM account_analytic_plan ORDER BY id LIMIT 1;")
+            fetch_plan = env.cr.fetchone()
+            plan_id = fetch_plan[0] if fetch_plan else False
+        if plan_id:
+            env["ir.model.data"].create(
+                {
+                    "module": "analytic",
+                    "name": "analytic_plan_projects",
+                    "model": "account.analytic.plan",
+                    "res_id": plan_id,
+                    "noupdate": True,
+                }
+            )
     if not plan_id:
         env.cr.execute("SELECT last_value + 1 FROM account_analytic_plan_id_seq;")
         plan_id = env.cr.fetchone()[0]


openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
```diff
@@ -22,6 +25,30 @@ def _fill_config_parameter_analytic_project_plan(env):
         [("module", "=", "analytic"), ("name", "=", "analytic_plan_projects")]
     )
     plan_id = imd.res_id
+    if not plan_id and openupgrade.column_exists(
+        env.cr, "res_company", "internal_project_id"
```
hbrunn: that's the id of a project, I think you'll want https://github.com/odoo/odoo/blob/16.0/addons/project/models/company.py#L9 here?

openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
```diff
@@ -22,6 +25,30 @@ def _fill_config_parameter_analytic_project_plan(env):
...
+    ):
+        env.cr.execute(
+            "SELECT internal_project_id "
+            "FROM res_company "
+            "WHERE internal_project_id IS NOT NULL "
+            "LIMIT 1;"
+        )
+        plan_id = env.cr.fetchone()[0]
+        if plan_id:
+            env.cr.execute(
```
hbrunn: no reason not to use the orm here I think

hbrunn (changes requested):

openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
```diff
@@ -22,6 +25,30 @@ def _fill_config_parameter_analytic_project_plan(env):
...
     )
     plan_id = imd.res_id
+    if not plan_id and openupgrade.column_exists(
+        env.cr, "res_company", "internal_project_id"
+    ):
+        env.cr.execute(
+            "SELECT internal_project_id "
+            "FROM res_company "
+            "WHERE internal_project_id IS NOT NULL "
+            "LIMIT 1;"
```
hbrunn, line 35:
```suggestion
            "LIMIT 1"
            "ORDER BY id"
```

hbrunn:
```suggestion
            """SELECT analytic_plan_id
            FROM res_company
            WHERE analytic_plan_id IS NOT NULL
            LIMIT 1
            ORDER BY id"""
```

openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
remi-filament: Of course you are right, I got mistaken !
Since the field was computed it was not present in SQL so I had to make something a bit more complicated (the same as done by compute here : https://github.com/odoo/odoo/blob/17f3ae162e4350cdc59503ae57bda05c382de58f/addons/project/models/company.py#L17)

openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
remi-filament: Turns out this column does not exist in DB...

openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
```diff
@@ -22,6 +26,26 @@ def _fill_config_parameter_analytic_project_plan(env):
...
     plan_id = imd.res_id
+    if not plan_id:
+        default_plans = (
+            env["ir.config_parameter"]
+            .search([("key", "like", "default_analytic_plan_id_%")], order="value")
+            .mapped("value")
+        )
+        plan_id = default_plans[0] if default_plans else False
+        if not plan_id:
+            env.cr.execute("SELECT id FROM account_analytic_plan ORDER BY id;")
```
MiquelRForgeFlow:
```suggestion
            env.cr.execute("SELECT id FROM account_analytic_plan ORDER BY id LIMIT 1;")
```

remi-filament: Thanks @MiquelRForgeFlow good catch !

hbrunn (approved):
ah right, that's not a db column

openupgrade_scripts/scripts/analytic/17.0.1.2/pre-migration.py
```diff
@@ -22,6 +26,26 @@ def _fill_config_parameter_analytic_project_plan(env):
...
+    if not plan_id:
+        default_plans = (
+            env["ir.config_parameter"]
+            .search([("key", "like", "default_analytic_plan_id_%")], order="value")
+            .mapped("value")
+        )
+        plan_id = default_plans[0] if default_plans else False
+        if not plan_id:
+            env.cr.execute("SELECT id FROM account_analytic_plan ORDER BY id LIMIT 1;")
+            plan_id = env.cr.fetchone()[0]
```
MiquelRForgeFlow:
```suggestion
            fetch_plan = env.cr.fetchone()
            plan_id = fetch_plan[0] if fetch_plan else False
```

remi-filament: Thanks, implemented as per suggestion

MiquelRForgeFlow (approved):

PR #5087, opened 2025-06-02
[17.0][OU-FIX] project: set default analytic_plan_id for projects

remi-filament:
Continuation of #5086 for project to set default value for analytic_plan_id


petrus-v (approved):
not openupgrade expert but code seems good to me

hbrunn (approved):

PR #5088, opened 2025-06-02
[18.0][OU-FIX] stock: adapt location_final_id

MiquelRForgeFlow:
Fixes `stock_move.location_final_id`.

### DIFF
--- a/openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
@@ -76,24 +76,42 @@ def fill_stock_move_location_dest_id(env):
     openupgrade.logged_query(
         env.cr,
         """
-        UPDATE stock_move sm
-        SET location_final_id = sr.location_dest_id
-        FROM stock_rule sr
-        WHERE sm.rule_id = sr.id AND sr.location_dest_id IS NOT NULL
-            AND sr.location_dest_id != sm.location_final_id
-            AND sr.location_dest_id != sm.location_dest_id
-            AND sr.action IN ('pull', 'pull_push')
+        WITH sub AS (
+            UPDATE stock_move sm
+            SET location_final_id = sr.location_dest_id
+            FROM stock_rule sr
+            WHERE sm.rule_id = sr.id AND sr.location_dest_id IS NOT NULL
+                AND sr.location_dest_id != sm.location_final_id
+                AND sr.location_dest_id != sm.location_dest_id
+                AND sr.action IN ('pull', 'pull_push')
+            RETURNING rule_id
+        ), sub2 AS (
+            SELECT rule_id
+            FROM sub
+            GROUP BY rule_id
+        )
+        UPDATE stock_rule sr
+        SET location_dest_from_rule = TRUE
+        FROM sub2
+        WHERE sub2.rule_id = sr.id
         """,
     )
     openupgrade.logged_query(
         env.cr,
         """
+        WITH sub AS (
+            SELECT sm.rule_id
+            FROM stock_move sm
+            JOIN stock_move_move_rel rel ON
+                rel.move_orig_id = sm.id OR rel.move_dest_id = sm.id
+            JOIN stock_rule sr ON sm.rule_id = sr.id
+            WHERE sr.action IN ('pull', 'pull_push')
+            GROUP BY sm.rule_id
+        )
         UPDATE stock_rule sr
         SET location_dest_from_rule = TRUE
-        FROM stock_move sm
-        WHERE sm.rule_id = sr.id
-            AND sm.location_dest_id != sm.location_final_id
-            AND sr.action IN ('pull', 'pull_push')
+        FROM sub
+        WHERE sub.rule_id = sr.id
         """,
     )
 
--- a/openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
@@ -2,6 +2,10 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from openupgradelib import openupgrade
 
+_columns_copy = {
+    "stock_move": [("location_dest_id", None, None)],
+}
+
 _field_renames = [
     ("stock.move", "stock_move", "location_dest_id", "location_final_id"),
     (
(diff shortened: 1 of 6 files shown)


hbrunn:
see https://github.com/ForgeFlow/OpenUpgrade/pull/27

hbrunn:
@MiquelRForgeFlow are you going to continue here or should I?

MiquelRForgeFlow:
I plan to continue ~~when finishing point_of_sale~~, but feel free to continue it if you want.

MiquelRForgeFlow:
CI is green.

openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
```diff
@@ -48,6 +52,30 @@ def fill_stock_move_location_dest_id(env):
         WHERE sm2.id = sm.id
         """,
     )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move sm2
+        SET location_final_id = sm.location_final_id
+        FROM stock_move sm
+        JOIN stock_move_move_rel rel ON rel.move_dest_id = sm.id
```
hbrunn: this should be recursive I think, you can have chains of push moves of arbitrary length

openupgrade_scripts/scripts/stock/tests/test_migration.py
```diff
@@ -0,0 +1,135 @@
...
+        self.assertEqual(from_intermediate.location_dest_id, customer_location)
+        self.assertEqual(from_intermediate.location_final_id, customer_location)
+
+        self.env["stock.rule"].search(
+            [
+                "|",
+                ("location_src_id", "=", intermediate_location.id),
+                ("location_dest_id", "=", intermediate_location.id),
+            ]
+        ).location_dest_from_rule = True
```
hbrunn: I don't think it should be necessary to set this here, should come out of the migration like that

hbrunn (changes requested):

openupgrade_scripts/scripts/stock/tests/test_migration.py
MiquelRForgeFlow: I mean, the migration expects to maintain compatibility. Odoo, in their comments, explain this:

https://github.com/odoo/odoo/commit/11e69870db1c49d9a6af79ffd263e4e162b34b6b:
Selection_4625
https://github.com/odoo/odoo/commit/76b7d335a0a04d59705ac2e99a08778bb45681b7:
Selection_4626

openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
MiquelRForgeFlow: recursion done!

openupgrade_scripts/scripts/stock/tests/test_migration.py
hbrunn: and you see in the tests that it's not the same behavior post migration if you don't set the flag

MiquelRForgeFlow: What do you mean? Your tests are not testing `location_dest_from_rule` flag.

hbrunn: but they test its effect. remove the line I'm commenting on, and you get different (wrong) location_dest_ids below

MiquelRForgeFlow: Ok, I changed that line by an assert, to check rule has `location_dest_from_rule = True`, an adapt queries.

MiquelRForgeFlow:
Let's merge this, as the module was done already. If some error is detected in the future,we will fix it later.

openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
```diff
@@ -51,12 +55,63 @@ def fill_stock_move_location_dest_id(env):
...
+            JOIN sub ON sub.move_orig_id = rel.move_dest_id)
+        )
+        UPDATE stock_move sm2
+        SET location_final_id = sm.location_final_id
+        FROM stock_rule sr, stock_move sm
+        JOIN sub ON sub.move_dest_id = sm.id
+        WHERE sm2.rule_id = sr.id AND sub.move_orig_id = sm2.id
+            AND sr.action IN ('push', 'pull_push')
+        """,
+    )
```
carlos-lopez-tecnativa, line 75: Hi, I have a case where a database from v17 was migrated to v18.

The `stock.rule` records are marked with `location_dest_from_rule=True`, but I don’t understand why this is done. Could you please explain the reason? I see that when Odoo creates a new database and generates the warehouse routes, this field is not marked as True.

I imagine this should only be applied when the warehouse is configured to use more than one delivery step. However, **if you create a delivery and then return it**, the moves are related through `move_origin_ids` and `move_dest_ids`, so this SQL query returns values even though the warehouse uses a single-step delivery flow.

Database migrated from v17
Image

Fresh database in v18

Image

What is the problem with `location_dest_from_rule=True`?

For example, when I install the `purchase_sale_stock_inter_company` module and create a purchase from C1 to C2:

When reviewing the delivery in C2, the `stock.picking` becomes `Stock -> Customer`, but the `stock.move` is `Stock -> Intercompany Location`.
If you perform the same test in a fresh database, or set `location_dest_from_rule=False`, the `stock.picking` becomes `Stock -> Intercompany Location`, and the `stock.move` uses the same locations.
Image

This inconsistency between the locations in `stock.picking` and `stock.move` causes problems, for example with the `stock_barcodes` module, which expects both records to have the same location. **When you scan a product, instead of modifying the current line, it tries to create a new stock.move**.

Please explain the reason for setting this field to True based on stock_move_move_rel. I think a better approach would be to set it to True only when the warehouse uses two- or three-step routes (although Odoo itself does not set this field in those cases either).

carlos-lopez-tecnativa (commented):
ping @MiquelRForgeFlow @hbrunn

openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py
hbrunn: my intention was to keep previous behavior (as that was my analysis of this flag at the time) in cases where it's used. If that doesn't work for you, that's a problem indeed.

Please provide a PR for adding [v18 test data + a v19 test](https://github.com/OCA/OpenUpgrade/tree/18.0/openupgrade_scripts/scripts/stock/tests) that fails with the current code, then we can discuss there how to fix it

PR #5089, opened 2025-06-02
[18.0][OU-ADD] purchase_stock

MiquelRForgeFlow:
.


openupgrade_scripts/scripts/purchase_stock/18.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,86 @@
...
+from openupgradelib import openupgrade
+
+_new_columns = [
+    ("purchase.order.line", "group_id", "many2one"),
+    ("purchase.order.line", "location_final_id", "many2one"),
+    ("stock.warehouse.orderpoint", "product_supplier_id", "many2one"),
+]
+
+
+def fill_purchase_order_line_group_id(env):
```
hbrunn, line 12: why the restrictions on group propagation? I'd set this unconditionally from COALESCE(po.group_id, stock_move.group_id)

openupgrade_scripts/scripts/purchase_stock/18.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,86 @@
...
+        """,
+    )
+
+
+def fill_purchase_order_line_location_final_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE purchase_order_line pol
+        SET location_final_id = sm.location_dest_id
```
hbrunn: this is not what you say you do below, I think the work file is right

hbrunn (changes requested):

openupgrade_scripts/scripts/purchase_stock/18.0.1.2/pre-migration.py
MiquelRForgeFlow: Yes, it's a small slip.

openupgrade_scripts/scripts/purchase_stock/18.0.1.2/pre-migration.py
MiquelRForgeFlow: The explanation is the following:

The purchase order line only receives the group in the procurement:
Selection_4582

But the procurement data is not saved anywhere. So, we need to go heuristic here. You cannot use freely the group of the purchase order because the group of the purchase is only the group of the procurement when the rule has 'propagate', otherwise it's the group of the stock rule.
Selection_4583

So, when the rule is not 'propagate', we don't have a clean way to set the group in the purchase order line.

MiquelRForgeFlow: @hbrunn what's your opinion of my comment?

hbrunn: agreed. But judging from https://github.com/odoo/odoo/commit/f9381a0611207e0fec6b957aef3b160437ecbafb and https://github.com/odoo/odoo/pull/177185/files, I think all those statements should do nothing if there's destination moves.

PS: I find links to code much more helpful than screenshots of code, makes it much more convenient to see some code in context

MiquelRForgeFlow: I added a last query to set group as stock_move.group_id.

pedrobaeza (approved):

PR #5090, opened 2025-06-05
[18.0][OU-ADD] l10n_fr

remi-filament:


remi-filament:
Hi @legalsylvain @hbrunn @MiquelRForgeFlow @pedrobaeza 

While looking at migrating l10n_fr which has largely been moved to l10n_fr_account (all accounting data : chart of account, taxes, etc.). I realize that module l10n_fr_fec has also been moved to l10n_fr_account.

In apriori.py file this module is marked as merged in l10n_fr which is incorrect (and therefore analysis is also incorrect since it considers that FEC related stuff has been removed from l10n_fr and added in l10n_fr_account).

However, I am not sure how to handle that since l10n_fr_account did not exist before, should we consider that l10n_fr_fec has been renamed to l10n_fr_account or does it imply side effects since l10n_fr_account is much more than only l10n_fr_fec ?
If we go with rename, then how should I handle the fact that a noupdate ir_rule from l10n_fr_fec has been removed ? I suppose this should be done in l10n_fr_account migration script, but since this is a new module, will this migration script be run ?

For now I have added the removal of ir_rule from l10n_fr_fec in this module migration script.

Thanks for your insight !

openupgrade_scripts/scripts/l10n_fr/18.0.2.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,384 @@
+---Models in module 'l10n_fr'---
+obsolete model account.fr.fec [transient]
+# NOTHING TO DO: removed transient model from l10n_fr_fec recreated in l10n_fr_account by ORM
+
+---Fields in module 'l10n_fr'---
+l10n_fr      / res.company              / l10n_fr_rounding_difference_loss_account_id (many2one): module is now 'l10n_fr_account' ('l10n_fr')
+l10n_fr      / res.company              / l10n_fr_rounding_difference_profit_account_id (many2one): module is now 'l10n_fr_account' ('l10n_fr')
+# DONE: pre-migration: rename fields
```
MiquelRForgeFlow: where is the script?

MiquelRForgeFlow:
@remi-filament `l10n_fr_fec` (that only contained a wizard) was merged into `l10n_fr` (see https://github.com/odoo/odoo/commit/0e71e2a380d51503f8a5bb2ece182eebcb2f696d) before `l10n_fr_account` was created. But what you say is true, when `l10n_fr_account` was later created, they moved the wizard there. So, please, in apriori merge `l10n_fr_fec` to `l10n_fr_account`. Then you will have to do in the migration script of `l10n_fr_account`:
- delete of `l10n_fr_fec.account_fr_fec_rule`
- rename of wizard table and models

remi-filament:
Thanks @MiquelRForgeFlow , I will do as proposed !

openupgrade_scripts/scripts/l10n_fr/18.0.2.1/upgrade_analysis_work.txt
remi-filament: I forgot it in the first commit.
I have added these 2 lines at the beginning of XML ids to be renamed in pre-migration. I am not sure if there is a better way to do that ?

docsource/modules170-180.rst
```diff
@@ -380,13 +380,13 @@ Module coverage 17.0 -> 18.0
...
+| l10n_fr                                           |Done                  |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |new| l10n_fr_account                             |                      |                                                 |
+| |new| l10n_fr_account                             |Done                  |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | l10n_fr_facturx_chorus_pro                        |                      |No DB layout changes.                            |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| l10n_fr_fec                                 |                      |Merged into l10n_fr.                             |
+| |del| l10n_fr_fec                                 |                      |Renamed to l10n_fr_account.                      |
```
MiquelRForgeFlow:
```suggestion
| |del| l10n_fr_fec                                 |                      |Merged into l10n_fr_account.                      |
```

openupgrade_scripts/apriori.py
```diff
@@ -43,7 +44,6 @@
     "l10n_es_edi_facturae_adm_centers": "l10n_es_edi_facturae",
     "l10n_es_edi_facturae_invoice_period": "l10n_es_edi_facturae",
     "l10n_es_edi_tbai_multi_refund": "l10n_es_edi_tbai",
-    "l10n_fr_fec": "l10n_fr",
```
MiquelRForgeFlow, line 46: "l10n_fr_fec": "l10n_fr_account".

openupgrade_scripts/scripts/l10n_fr_account/18.0.2.2/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_models(env.cr, model_renames)
+    openupgrade.rename_tables(env.cr, table_renames)
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "l10n_fr_fec.account_fr_fec_rule",
+        ],
+    )
```
MiquelRForgeFlow, line 24:
```python
    openupgrade.logged_query(  # just in case `account_fr_fec` was not installed
        env.cr,
        """
        UPDATE ir_module_module
        SET state='to install'
        WHERE name = 'l10n_fr_account' AND state='uninstalled'""",
    )
```

MiquelRForgeFlow (commented):
I think it's better to put into merge modules. It doesn't change anything technically, but semantically.

openupgrade_scripts/apriori.py
remi-filament: You would put it both in renamed_modules and merged_modules ?

docsource/modules170-180.rst
remi-filament: Fine

openupgrade_scripts/scripts/l10n_fr_account/18.0.2.2/pre-migration.py
remi-filament: Thanks, indeed I should have kept that one

remi-filament:
While testing, I had an issue since Odoo tries to run end_migration script from l10n_fr_account version 2.1
The problem I had was linked to account.account which did not have the proper account_type and non_trade not set.

Therefore I made some extra post-migration script to handle those. I hope this is OK ?

openupgrade_scripts/scripts/l10n_fr/18.0.2.1/upgrade_analysis_work.txt
MiquelRForgeFlow: In fact, this kind of thing is done by the ORM. Just delete what you did here and leave as NOTHING TO DO.

docsource/modules170-180.rst
```diff
@@ -380,13 +380,13 @@ Module coverage 17.0 -> 18.0
...
+| l10n_fr                                           |Done                  |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |new| l10n_fr_account                             |                      |                                                 |
+| |new| l10n_fr_account                             |Done                  |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | l10n_fr_facturx_chorus_pro                        |                      |No DB layout changes.                            |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| l10n_fr_fec                                 |                      |Merged into l10n_fr.                             |
+| |del| l10n_fr_fec                                 |                      |Merged into l10n_fr_account.                     |
```
MiquelRForgeFlow:
```suggestion
| |del| l10n_fr_fec                                 |Done                  |Merged into l10n_fr_account.                     |
```

openupgrade_scripts/scripts/l10n_fr_account/18.0.2.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,3185 @@
+---Models in module 'l10n_fr_account'---
+new model l10n_fr.fec.export.wizard [transient]
+# DONE: pre-migration: rename model from account.fr.fec
+
+---Fields in module 'l10n_fr_account'---
+l10n_fr_account / res.company              / l10n_fr_rounding_difference_loss_account_id (many2one): previously in module l10n_fr
+l10n_fr_account / res.company              / l10n_fr_rounding_difference_profit_account_id (many2one): previously in module l10n_fr
+# NOTHING TO DO: handled in l10n_fr migration script
```
MiquelRForgeFlow:
```suggestion
# NOTHING TO DO: handled by ORM
```

remi-filament:
Thanks @MiquelRForgeFlow I implemented your remarks !

MiquelRForgeFlow (approved):

openupgrade_scripts/apriori.py
pedrobaeza: Why is this in merged modules instead on renamed ones?

MiquelRForgeFlow: `l10n_fr_account` is a split of `l10n_fr` in which `l10n_fr_fec` was merged into. Is there a problem?

pedrobaeza: That conceptualization doesn't matter at this level. The fact is that `l10n_fr_account` didn't exist previously, and `l10n_fr_fec` did, but not now, so it's a clear rename. The rest of the operations (renames and so on) should be done on scripts, but not on apriori.py.

There's a risk (although minimal), that putting it in `merged_modules`, some information is lost (as there are deletes), or that some important crashes are ignored.

And also talking about that, if this is a split as you say, you should force its installation in `l10n_fr` scripts.

MiquelRForgeFlow: > if this is a split as you say, you should force its installation in l10n_fr scripts.

This was already done in pre-migration of `l10n_fr`.

openupgrade_scripts/scripts/l10n_fr/18.0.2.1/pre-migration.py
```diff
@@ -0,0 +1,687 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_xmlids(env.cr, _xmlids_renames)
+    openupgrade.logged_query(  # just in case `account_fr_fec` was not installed
+        env.cr,
+        """
+        UPDATE ir_module_module
+        SET state='to install'
+        WHERE name = 'l10n_fr_account' AND state='uninstalled'""",
```
pedrobaeza, line 686: This is where this can be conflicting, as merging the other module may have an undesired state "to install" while it should be "to upgrade".

MiquelRForgeFlow: https://github.com/OCA/OpenUpgrade/pull/5118.

PR #5091, opened 2025-06-06
[17.0][OU-FIX] base: error on pre-migration file

adasatorres:
Good morning, while migrating a database to version 17.0, I encountered an error because the code field in the res_country model has a defined length of 2 characters. However, the update that assigns the value adds the string 'UO' followed by the ID (as a varchar), which exceeds the allowed length and causes an error. 

I’ve modified the update to assign only 'UO' to the field, as appending the ID as a varchar could trigger the same error if the ID exceeds the allowed length.

### DIFF
--- a/openupgrade_scripts/scripts/base/17.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/17.0.1.3/pre-migration.py
@@ -157,9 +157,26 @@ def _fill_empty_country_codes(cr):
     openupgrade.logged_query(
         cr,
         """
-        UPDATE res_country
-        SET code = 'OU' || id::VARCHAR
-        WHERE code IS NULL
+        WITH dummy_codes AS (
+            SELECT LPAD(i::text, 2, '0') AS code
+            FROM generate_series(0, 99) AS i
+        ),
+        available_codes AS (
+            SELECT dc.code, ROW_NUMBER() OVER () AS rn
+            FROM dummy_codes dc
+            LEFT JOIN res_country rc ON rc.code = dc.code
+            WHERE rc.code IS NULL
+        ),
+        null_countries AS (
+            SELECT id, ROW_NUMBER() OVER () AS rn
+            FROM res_country
+            WHERE code IS NULL
+        )
+        UPDATE res_country AS rc
+        SET code = ac.code
+        FROM available_codes ac
+        JOIN null_countries nc ON nc.rn = ac.rn
+        WHERE rc.id = nc.id
         """,
     )
 


pedrobaeza:
I'm afraid this is not valid, as the code should be unique, so having more than one country the same code will fail as well. From which version do you come? I think the size limits in chars was removed some time ago and this can be a leftover.

adasatorres:
Good night @pedrobaeza,

I hadn’t realized the field was set as unique, as I ran the script during some initial testing of the changes and it didn’t raise any errors.

I’m currently in the process of migrating from version 16.0 to 17.0. Additionally, based on what I’ve seen in the Odoo repository, the code field still appears to have a size of 2. I’ve also confirmed this behavior on a local test machine.

Would you happen to have any suggestions?

adasatorres:
Good morning, I’ve been doing some research, and the only idea I can think of is to delete all those records that don't have a code.

hbrunn:
do something creative like

```sql
select dummy_code::text from generate_series(0, 99) as dummy_code where not exists (select code from res_country where code=dummy_code::text) limit 1
```

which will be fine for a hundred countries without code

hbrunn:
@pedrobaeza 

```

openupgrade18=> \d res_country
                                           Table "public.res_country"
     Column      |            Type             | Collation | Nullable |                 Default                 
-----------------+-----------------------------+-----------+----------+-----------------------------------------
 id              | integer                     |           | not null | nextval('res_country_id_seq'::regclass)
 address_view_id | integer                     |           |          | 
 currency_id     | integer                     |           |          | 
 phone_code      | integer                     |           |          | 
 create_uid      | integer                     |           |          | 
 write_uid       | integer                     |           |          | 
 code            | character varying(2)        |           | not null |
```

pedrobaeza:
The size limit is enforced in the base.sql file...

adasatorres:
Good afternoon! Adding that query inside the update could be dangerous, right? I don't have deep knowledge of SQL, but since the update runs over a snapshot of the table, that query will always return '00'.

hbrunn:
you're supposed to do that in a loop. or in a cte. doesn't matter, you just need to produce a unique two character value here

openupgrade_scripts/scripts/base/17.0.1.3/pre-migration.py
```diff
@@ -157,9 +157,26 @@ def _fill_empty_country_codes(cr):
...
+        null_countries AS (
+            SELECT id, ROW_NUMBER() OVER () AS rn
+            FROM res_country
+            WHERE code IS NULL
+        )
+        UPDATE res_country AS rc
+        SET code = ac.code
+        FROM available_codes ac
+        JOIN null_countries nc ON nc.rn = ac.rn
+        WHERE rc.id = nc.id;
```
hbrunn:
```suggestion
        WHERE rc.id = nc.id
```

hbrunn (approved):

MiquelRForgeFlow (approved):
LGTM

PR #5034, opened 2025-05-14
[18.0][OU-ADD] mrp

rvalyi:
currently rebased on #4957 

The migration has been generated entirely with Gemini AI. Let's see how it works now...
More on the process soon... But if it works this is big deal...

PR #5092, opened 2025-06-09
[18.0][OU-ADD] mrp

hbrunn:
/ocabot migration mrp


MiquelRForgeFlow:
Have you checked https://github.com/OCA/OpenUpgrade/pull/5034?

hbrunn:
> Have you checked #5034?

seeing the [hallucinations](https://github.com/OCA/OpenUpgrade/actions/runs/15017386557/job/42198382997#step:13:1340) I consider my time too valuable to look into it in depth

*Of course I shouldn't have written "hallucination", that's the anthropomorphizing bullshit the ai bro hype machine spits out

rvalyi:
@MiquelRForgeFlow @hbrunn for the record I admittedly just scratched the surface with 4 Gemini 2.5 pro prompts in the mentioned PR above. But I can see it got a few things right (disclaimer I didn't check deeply)...

I made this little tool to build a focused Odoo context for the AI https://github.com/akretion/akaidoo

It basically uses manifestoo to spit the models code (and migration and module-diff-analysis) of the module and all the dependent modules. So it keeps the hallucination pretty low! In these prompts I fogot to to pass the code of openupgradelib.py in the 1st prompt and apparently this is important to fully avoid hallucinations. Unlike other models Gemini has a huge context (1 millions token) and can be tested for free (for how long?) in aistudio.google.com . You can test that such "akaidoo" mrp migration prompt easily consumes 500k tokens, so the 1 million token of Gemini is the game changer here for Odoo. 

BTW I don't endorse all ethics implications of AI, but I think we will have no other choice than deal with it to stay relevant. cc @pedrobaeza @etobella @sebastienbeau @renatonlima @sbidoul

pedrobaeza:
I'm also not discarding AI in any context, but right now, I think it takes more time to ours (understood as the core OpenUpgrade developers) to do that research and experiments than to write actual working migration scripts.

rvalyi:
> I'm also not discarding AI in any context, but right now, I think it takes more time to ours (understood as the core OpenUpgrade developers) to do that research and experiments than to write actual working migration scripts.

I can imagine that. On my side it was rather to test the AI capabilities on something quite hard. And frankly I think it will be very useful on simpler modules...

Also I can assure you akaidoo and Gemini helps me a lot for the 80% of the simpler code I do daily. Yes it should still be watched carefully and corrected, but all in all I can guarantee you it's a huge productivity boost. Specially to write tests or simple customizations. At least for me.

openupgrade_scripts/scripts/mrp/18.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,103 @@
...
+
+# NOTHING TO DO: we can't detect if a document has been attached via the bom or the product itself, so we keep the default 'hidden'
+
+mrp          / stock.move               / manual_consumption (boolean)  : not a function anymore
+
+# NOTHING TO DO: can be edited manually now
+
+mrp          / stock.picking.type       / use_auto_consume_components_lots (boolean): DEL
+
+# NOTHING TO DO: replaced by manual_consumption being manually editable above
```
MiquelRForgeFlow, line 73: shouldn't `manual_consumption` be filled as per `use_auto_consume_components_lots`?

openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    openupgrade.add_columns(env, [("mrp.workorder", "sequence", "integer", 100)])
+    env.cr.execute(
+        """
+        UPDATE mrp_workorder
+        SET sequence=COALESCE(mrp_routing_workcenter.sequence, mrp_workorder.sequence)
+        FROM mrp_routing_workcenter
+        WHERE mrp_workorder.operation_id=mrp_routing_workcenter.id
+        """
+    )
```
MiquelRForgeFlow, line 29:
```suggestion
    openupgrade.add_columns(env, [("mrp.workorder", "sequence", "integer")])
    env.cr.execute(
        """
        UPDATE mrp_workorder
        SET sequence=COALESCE(mrp_routing_workcenter.sequence, mrp_workorder.sequence)
        FROM mrp_routing_workcenter
        WHERE mrp_workorder.operation_id=mrp_routing_workcenter.id
        """
    )
    env.cr.execute(
        """
        SELECT COALESCE(max(sequence), 0) AS max_sequence
        FROM mrp_workorder
        WHERE sequence IS NOT NULL
        """
    )
    max_sequence = env.cr.fetchone()[0]
    openupgrade.logged_query(
        env.cr,
        f"""
        UPDATE mrp_workorder
        SET sequence = {max_sequence} + id::int
        WHERE sequence IS NULL""",
    )
```

openupgrade_scripts/scripts/mrp/18.0.2.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2025 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def mrp_document_to_product_document(env):
+    """
+    mrp.document has been replaced by product.product.
```
MiquelRForgeFlow:
```suggestion
    mrp.document has been replaced by product.document.
```

openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+
+from openupgradelib import openupgrade
+
+add_columns = [
+    ("product.document", "attached_on_mrp", "char", "hidden"),
+]
+
+copy_columns = {
+    "mrp_production": [
+        ("location_dest_id", "location_final_id", "integer"),
```
MiquelRForgeFlow, line 12: are you sure we don't need to do some adjustment to `location_dest_id` or `location_final_id`? :)

openupgrade_scripts/scripts/mrp/18.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: hehe
Selection_4677

openupgrade_scripts/scripts/mrp/18.0.2.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    link_column = openupgrade.get_legacy_name("mrp_document_id")
+    openupgrade.add_columns(
+        env,
+        [("product.document", link_column, "integer")],
+    )
+    openupgrade.logged_query(
+        env.cr,
+        f"""
+        INSERT INTO product_document
```
MiquelRForgeFlow, line 20: It think https://github.com/OCA/OpenUpgrade/blob/17.0/openupgrade_scripts/scripts/product/17.0.1.2/post-migration.py#L61 also included the ones in mrp.document. So, in fact, I think we should just check `product_document.ir_attachment_id` with `mpr_document.ir_attachment_id` and where is match set `attached_on_mrp = 'bom'`.

MiquelRForgeFlow: IA says:
Selection_4675

hbrunn: how does it matter what the v17 migration did? This needs to work for just 17->18 too, so if there's anything to do, it's excluding attachments that are mrp.documents in the v17 migration.

Not sure how the word soup is helpful? We can't do anything about the field because this is [set](https://github.com/odoo/odoo/blob/18.0/addons/mrp/models/mrp_bom.py#L716) depending on the attachment being created via a bom or not, which we simply don't know, so I think defaulting to `hidden` is the right thing to do

openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
hbrunn: what's the point of this? In my version, work orders without operation_id are ordered by `leave_id, date_start, id` thanks to the default value you propose to delete. How is moving them after all orders with operation_id and enforcing an ordering by id only an improvement?

openupgrade_scripts/scripts/mrp/18.0.2.0/upgrade_analysis_work.txt
hbrunn: computing this from use_auto_consume_components_lots has already happened in [v17](https://github.com/odoo/odoo/blob/17.0/addons/mrp/models/stock_move.py#L720)

openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
hbrunn: well not entirely, I find the way this is implemented in v18 somewhat confusing. My understanding is that `location_final_id` is mainly used for reporting and UI purposes, while `location_dest_id` is the relevant one for stock. Will add a few tests comparing migrated v17 and v18 behavior

hbrunn: now I remember the reasoning: Whatever production orders exist, should go exactly where the v17 code pointed them to by location_dest_id.

The [v18](https://github.com/odoo/odoo/blob/18.0/addons/mrp/models/stock_rule.py#L164-L165) code allows for detours if the picking type is configured accordingly, but whatever users did in v17 will rely on the resulting move to end up in location_dest_id, so no, no adjustment needed.

openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
MiquelRForgeFlow: So, should we update them by `leave_id, date_start, id`? 😅

openupgrade_scripts/scripts/mrp/18.0.2.0/post-migration.py
MiquelRForgeFlow: > it's excluding attachments that are mrp.documents in the v17 migration.

Exactly, just not set `hidden` for them.

pedrobaeza (approved):
Check existing comments, and everything seems to be OK.

PR #5094, opened 2025-06-11
[18.0][OU-ADD] account_edi_ubl_cii: Nothing to do

remi-filament:


openupgrade_scripts/scripts/account_edi_ubl_cii/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,27 @@
...
+account_edi_ubl_cii / res.company              / invoice_is_ubl_cii (boolean)  : DEL
+# NOTHING TO DO: removed parameter not used anymore
+
+account_edi_ubl_cii / res.partner              / invoice_edi_format (False)    : NEW selection_keys: ['facturx', 'nlcius', 'ubl_a_nz', 'ubl_bis3', 'ubl_sg', 'xrechnung'], mode: modify
+account_edi_ubl_cii / res.partner              / ubl_cii_format (selection)    : DEL selection_keys: ['facturx', 'nlcius', 'ubl_a_nz', 'ubl_bis3', 'ubl_sg', 'xrechnung']
+# NOTHING TO DO: field renamed but already added in account and computed not stored
+
+account_edi_ubl_cii / res.partner              / is_peppol_edi_format (boolean): previously in module account_peppol
+account_edi_ubl_cii / res.users                / is_peppol_edi_format (boolean): previously in module account_peppol
+# NOTHING TO DO: fields moved from account_peppol, to be handled in account_peppol migration script
```
hbrunn, line 12: actually, I think it's this addon's responsibility to handle this.

The computation function has changed from [v17](https://github.com/odoo/odoo/blob/17.0/addons/account_peppol/models/res_partner.py#L151) to [v18](https://github.com/odoo/odoo/blob/18.0/addons/account_edi_ubl_cii/models/res_partner.py#L144-L149) (it looks at this dict's on_peppol key), so I think you should detect if `is_peppol_edi_format` exists in pre-migration, copy it to a legacy column, and in end-migration, recompute the field to give localizations the chance to add their own formats

openupgrade_scripts/scripts/account_edi_ubl_cii/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,27 @@
...
+# NOTHING TO DO: values will be removed by ORM and should be handled manually if where used (0212, 0215, 9901)
```
hbrunn, line 15: how does the ORM do that? I'd expect you have to map the deleted values to some existing value/set null

hbrunn (changes requested):

pedrobaeza:
Superseded by #5143 attending @hbrunn's comments.

openupgrade_scripts/scripts/account_edi_ubl_cii/18.0.1.0/upgrade_analysis_work.txt
pedrobaeza: This is a non stored field, so finally this is indeed nothing to do.

PR #5099, opened 2025-06-11
[18.0][OU-ADD] sale_management

remi-filament:


openupgrade_scripts/scripts/sale_management/18.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+from openupgradelib import openupgrade
+
+
+def _fill_sale_order_template_sequence(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+         ALTER TABLE sale_order_template ADD COLUMN IF NOT EXISTS sequence INTEGER
+         """,
+    )
```
MiquelRForgeFlow, line 13: LGTM, but FYI: we use now `openupgrade.add_columns(env, [("sale.order.template", "sequence", "integer")])`

openupgrade_scripts/scripts/sale_management/18.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,26 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_template
+        SET sequence = id
+        WHERE sequence IS NULL
```
MiquelRForgeFlow, line 19: LGTM, but are you sure this does the trick?

hbrunn: why wouldn't it?

hbrunn (approved):

openupgrade_scripts/scripts/sale_management/18.0.1.0/pre-migration.py
MiquelRForgeFlow: I don't know, I just asked haha

pedrobaeza (approved):

Issue #5105, opened 2025-06-17
[17.0] mail: question about field_info in mail.tracking.value.


adasatorres:
Good morning,

I’ve opened this issue because I have a question regarding certain fields in the `mail.tracking.value` model. In version 17.0, the `field_desc` and `field_type` fields have been removed, and the information is now stored in the JSON field `field_info`. I’ve reviewed the scripts in the mail folder but haven’t found any reference to this change. Is there a specific reason why this update hasn’t been reflected or documented in the scripts?

Thank you in advance for the clarification.

pedrobaeza:
This haven't been done because of unacknowledged of this change, but indeed it should be done.

pedrobaeza:
Check https://github.com/OCA/OpenUpgrade/pull/5106#issuecomment-2990084470

PR #5106, opened 2025-06-17
[17.0][OU-IMP] mail: field_desc and field_type to field_info

adasatorres:
Possible solution for the following issue #5105 .


rrebollo (changes requested):
Looks good but You need to reflect the changes in the upgrade_analysis_work.txt file. Example: [openupgrade_scripts/scripts/website_sale/17.0.1.1/upgrade_analysis_work.txt](https://github.com/OCA/OpenUpgrade/pull/5103/files)

MiquelRForgeFlow:
> You need to reflect the changes in the upgrade_analysis_work.txt file

Exactly.

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
```diff
@@ -395,3 +395,6 @@ DEL ir.ui.view: mail.view_mail_alias_tree
 
 NEW res.groups: mail.group_mail_notification_type_inbox (noupdate)
 # NOTHING TO DO
+
+mail         / mail.tracking.value      / field_info (jsonb)            : NEW isfunction: function, stored
```
MiquelRForgeFlow: you have to do something with lines 170-175.

adasatorres: @MiquelRForgeFlow Adding the DONE line below line 175 should be enough, right? Also, line 172 has the wrong field type — the field is actually of type jsonb. 
image

MiquelRForgeFlow: > Adding the DONE line below line 175 should be enough, right?

No, because someone looking into this file will get confused, for duplicated/contradictory information. It's better to amend line 175 and explain your approach there.

> Also, line 172 has the wrong field type — the field is actually of type jsonb.

Column type is `jsonb`, but odoo type is `json`. The same way you see `many2one` odoo types there instead of `int4`.

pedrobaeza: @adasatorres can you group together the lines and put for all of them the `DONE` message for continuing with this?

openupgrade_scripts/scripts/mail/17.0.1.15/post-migration.py
```diff
@@ -151,6 +151,20 @@ def _mail_activity_plan_template(env):
...
+def _mail_tracking_value_field_desc_type_on_field_info(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE mail_tracking_value
+        SET field_info = jsonb_build_object(
+            'type', field_type,
+            'desc', field_desc
+        )
+        WHERE field_info IS NULL;
```
pedrobaeza:
```suggestion
        WHERE field_info IS NULL
            AND field_type IS NOT NULL
            AND field_desc IS NOT NULL
```

openupgrade_scripts/scripts/mail/17.0.1.15/upgrade_analysis_work.txt
adasatorres: @pedrobaeza I was also thinking of adding the commit ID to the NOTHING TO DO.

pedrobaeza:
Anyway, I'm checking source code, and it seems that part is only filled when a ir.model.fields is removed:

https://github.com/odoo/odoo/blob/ee1e91ec69c71e783e7f291f8d19a2610c9070a3/addons/mail/models/ir_model_fields.py#L51

but no other reference to that in the rest of tracking references.

I have tried to set the value in one message, and I get an error:

imagen

as `name` key is mandatory, which is not being filled.

Filling that information on a regular tracking value, is not showing nothing extra on the chatter, as `field_id` is filled and the information is get from there.

On summary, this change is not correct and we can't recreate the information for ir.model.fields unlink trackings, so let's keep it as is.

Which thing brought you to this initially?

adasatorres:
The migration of project_task_report to version 17.0 is what led me to investigate the mail.tracking.value model. While reviewing the code, I saw that some methods were using that field to store information. I suppose I didn’t review it thoroughly enough — sorry about that. @pedrobaeza

PR #5118, opened 2025-06-25
[18.0][OU-FIX] l10n_fr: let's upgrade l10n_fr_account

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/pull/5090.

Let's see the cases:

Case 1:
- v17: `l10n_fr` installed, `l10n_fr_fec` not installed.
- v18: `l10n_fr` migrates, and sets 'to install' `l10n_fr_account`. Then `l10n_fr_account` is installed (but scripts of `l10n_fr_account` are not executed, which is wrong. We need them to fix tax accounts). **Fixed** by putting "to upgrade". We also check if `l10n_fr_fec` was installed.

Case 2:
- v17: `l10n_fr` installed, `l10n_fr_fec` is installed.
- v18: In base, `l10n_fr_fec` data is merged into `l10n_fr_account`. `l10n_fr` migrates and then `l10n_fr_account` migrates.

### DIFF
--- a/openupgrade_scripts/scripts/l10n_fr_account/18.0.2.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/l10n_fr_account/18.0.2.2/pre-migration.py
@@ -14,11 +14,9 @@ table_renames = [
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.rename_models(env.cr, model_renames)
-    openupgrade.rename_tables(env.cr, table_renames)
-    openupgrade.delete_records_safely_by_xml_id(
-        env,
-        [
-            "l10n_fr_fec.account_fr_fec_rule",
-        ],
-    )
+    if openupgrade.table_exists(env.cr, "account_fr_fec"):
+        openupgrade.rename_models(env.cr, model_renames)
+        openupgrade.rename_tables(env.cr, table_renames)
+        openupgrade.delete_records_safely_by_xml_id(
+            env, ["l10n_fr_fec.account_fr_fec_rule"]
+        )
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/l10n_fr_account/18.0.2.2/pre-migration.py
```diff
@@ -14,11 +14,12 @@
...
-        ],
-    )
+    if openupgrade.table_exists(env.cr, "account_fr_fec"):
+        openupgrade.rename_models(env.cr, model_renames)
+        openupgrade.rename_tables(env.cr, table_renames)
+        openupgrade.delete_records_safely_by_xml_id(
+            env,
+            [
+                "l10n_fr_fec.account_fr_fec_rule",
+            ],
```
pedrobaeza: Let's not add lines by the formatter (the key is to not add trailing commas):

```suggestion
            env, ["l10n_fr_fec.account_fr_fec_rule"]
```

MiquelRForgeFlow: That's pure nitpicking 😅

pedrobaeza: Yeah, I agree, but as we are changing the lines...

pedrobaeza (approved):

remi-filament:
Thanks @MiquelRForgeFlow I tested both use cases (with or without l10n_fr_fec installed) and it works fine !

PR #5121, opened 2025-06-26
[13.0][OU-FIX] stock: convert responsible_id to company_dependent

MiquelRForgeFlow:
Handle conversion to company-dependent.

Before:
Selection_4705

After:
Selection_4706


pedrobaeza (commented):
Why not using the openupgradelib method?

MiquelRForgeFlow:
> Why not using the openupgradelib method?

Selection_4708

pedrobaeza:
You have to rename the column in pre, and convert in post.

pedrobaeza:
Or do you mean other thing? And if so, let's improve the method.

MiquelRForgeFlow:
Let's avoid renames in old versions.

pedrobaeza:
OK on my side.

PR #5123, opened 2025-06-26
[18.0][OU-ADD] sale_purchase

MiquelRForgeFlow:
CI error is due to `purchase_stock`, so let's wait to:

- [x] https://github.com/OCA/OpenUpgrade/pull/5089.

PR #5124, opened 2025-06-26
[16.0][OU-IMP] sale_purchase: rename deprecated column

MiquelRForgeFlow:
It's better this way. If not, when you migrate to v18, you will get an error during the module load, as the column will exist and ORM will try to set it as jsonb (See https://github.com/OCA/OpenUpgrade/pull/5123).

### DIFF
--- a/openupgrade_scripts/scripts/sale_purchase/16.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/sale_purchase/16.0.1.0/post-migration.py
@@ -71,6 +71,10 @@ def convert_service_to_purchase_to_company_dependent(env):
         """,
         {"field_id": service_to_purchase_field_id},
     )
+    openupgrade.rename_columns(
+        env.cr,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not sure about this, as already migrated instances will have the problem the same if a cleanup has been done yet. I think it's better to tackle it on 18 detecting if the column exists and with which type, and doing there the renaming or whatever is needed.

MiquelRForgeFlow:
Cannot it be done in both places?

pedrobaeza:
Well, why having so much code for the same problem? And I see it more targeted in 18.0, as it's where the problem arises.

pedrobaeza:
That column may be used currently for any extra migration scripts (inside or outside OpenUpgrade), so renaming it will break them.

MiquelRForgeFlow:
Ok, implemented in https://github.com/OCA/OpenUpgrade/pull/5123.

PR #5128, opened 2025-06-28
[18.0][OU-ADD] event_product: Migration scripts

pedrobaeza:
@Tecnativa TT56885


docsource/modules170-180.rst
```diff
@@ -142,7 +142,7 @@ Module coverage 17.0 -> 18.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | event_crm_sale                                    |                      |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |new| event_product                               |                      |                                                 |
+| |new| event_product                               |Done                  |                                                 |
```
MiquelRForgeFlow, line 145: This is new, so it will be installed instead of upgraded. Does the scripts will be executed?

pedrobaeza: It should thanks to the corresponding patch in `openupgrade_framework`, but it's true that I'm not seeing it to work:

https://github.com/OCA/OpenUpgrade/actions/runs/15969669440/job/45037815900#step:13:2681

@hbrunn @StefanRijnhart isn't the patch working?

MiquelRForgeFlow: The patch was https://github.com/OCA/OpenUpgrade/commit/e06c908652e139b41fdf64d0d16c2df6539ad0d6. Maybe something changed in odoo.

hbrunn: looking into this

MiquelRForgeFlow: I have a theory, but it would need proof.

My hypothesis is as follows:
Selection_4713
New modules doesn't have "installed version".
Selection_4712
By default, odoo uses "0.1" if not found.
Selection_4710
Later odoo does comparisons. As new `event_product` has version "0.1" and I suppose parsed_installed_version is assigned v18 too, then in comparison, it's the same version and script is not executed.

StefanRijnhart: An issue with the patch was already observed in 16: https://github.com/OCA/OpenUpgrade/pull/3286#issuecomment-2067923444

MiquelRForgeFlow: Shouldn't we patch `load_module_graph`?

hbrunn: https://github.com/OCA/OpenUpgrade/pull/5147

PR #5146, opened 2025-06-30
[17.0][IMP] account_edi_ubl_cii: Optimize end-script

pedrobaeza:
- Don't do the compute if the initial conditions (VAT or country) are not set.
- Don't compute 2 times peppol_endpoint, as it's already triggered by the computation of peppol_eas.

@Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/account_edi_ubl_cii/17.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/account_edi_ubl_cii/17.0.1.0/end-migration.py
@@ -3,10 +3,13 @@ from openupgradelib import openupgrade
 
 
 def _res_partner_compute_fields_values(env):
-    res_partner = env["res.partner"].with_context(active_test=False).search([])
-    res_partner._compute_ubl_cii_format()
-    res_partner._compute_peppol_endpoint()
-    res_partner._compute_peppol_eas()
+    partners = (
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@MiquelRForgeFlow @hbrunn do you see this reasonable?

hbrunn (commented):
to me copying the computation functions' code screams tacit weird bugs whenever this changes upstream.

Agreed with restricting this to the partners where it makes sense, I'd however use [_peppol_eas_endpoint_depends](https://github.com/odoo/odoo/blob/17.0/addons/account_edi_ubl_cii/models/res_partner.py#L166) to generate a list of fields that should be non-null.

For tackling performance issues, I'd rather remove `_compute_peppol_endpoint` (it's recomputed by changing peppol_eas), possibly run the three functions on chunks of partners, and look out if there are any unexpected flushes

pedrobaeza:
@hbrunn I have reduced the scope to 2 points for not overcomplicating:


- Don't do the compute if the initial conditions (VAT or country) are not set.
- Don't compute 2 times peppol_endpoint, as it's already triggered by the computation of peppol_eas.

With this, we get a slight better computation time in my current case (~7 s), but touching other way, could be overkill.

hbrunn (approved):

PR #5147, opened 2025-06-30
[18.0] openupgrade_framework: several issues concerning module installation

hbrunn:
see https://github.com/OCA/OpenUpgrade/pull/5128/files#r2174688424

This fixes several issues:
- auto_install addons weren't actually installed by OpenUpgrade (the addon in the link above is only installed because sale_event pulls it)
- pre scripts weren't run for newly installed addons
- no migration scripts were run for newly installed addons

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_module.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_module.py
@@ -1,0 +1,33 @@
+# Copyright Odoo Community Association (OCA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.addons.base.models.ir_module import Module
+
+
+def update_list(self):
+    """
+    Mark auto_install modules as to install if all their dependencies are some kind of
+    installed.
+    Ignore localization modules that are set to auto_install
+    """
+    Module.update_list._original_method(self)
+    new_auto_install_modules = self.browse([])
+    for module in self.env["ir.module.module"].search(
+        [
+            ("auto_install", "=", True),
+            ("state", "=", "uninstalled"),
+            ("name", "not like", ("l10n_%")),
+        ]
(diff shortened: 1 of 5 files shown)


StefanRijnhart (approved):
Great!

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_module.py
```diff
@@ -0,0 +1,33 @@
+# Copyright Odoo Community Association (OCA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.addons.base.models.ir_module import Module
+
+
+def update_list(self):
+    """
+    Mark auto_install modules as to install if all their dependencies are some kind of
```
MiquelRForgeFlow, line 10: Why this is needed? I thought `auto_install` was already working.

hbrunn: try it: Migrate a database where you said `-i account,event,product` and verify that without this patch, you don't get `event_product` after the migration. Or remove the auto_install flag in your code, install the three modules, re-add the auto_install flag and see you don't get this installed automatically.

The Odoo code dealing with auto_install is run when modules are installed, not when they're updated.

MiquelRForgeFlow: Nice explanation.

MiquelRForgeFlow:
@hbrunn could you make backport of these fixes in the versions that need them?

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (approved):
Thanks for the patches. Checked that the install one works: https://github.com/OCA/OpenUpgrade/actions/runs/15974698237/job/45053999890?pr=5147#step:13:2685

/ocabot merge patch

PR #5148, opened 2025-07-01
[15.0][OU-ADD] website_crm_privacy_policy: Merged into website_form_require_legal

victoralmau:
Merged into `website_form_require_legal`

@Tecnativa TT54869

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -113,6 +113,7 @@ merged_modules = {
     # OCA/web
     "web_decimal_numpad_dot": "web",
     # OCA/website
+    "website_crm_privacy_policy": "website_form_require_legal",
     "website_google_analytics_4": "website",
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Shouldn't you need to mark the contact us form with the check that requires the legal?

victoralmau:
I believe that according to the existing script https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/scripts/website_crm/15.0.2.1/pre-migration.py this is not necessary, right?

pedrobaeza:
With that code, you can't do this merge, as it won't be detected in that code. I think the module would be just installed.

victoralmau:
> With that code, you can't do this merge, as it won't be detected in that code. I think the module would be just installed.

I agree, that's why I say that with the existing migration script this PR is no longer needed.

PR #5151, opened 2025-07-04
[18.0][REV+OU-FIX] openupgrade_framework: Revert version assignation + event_product: remove decorator

pedrobaeza:
- Revert "[FIX] openupgrade_framework: manipulate versions to actually run scripts"

- event_product: Adapt to run it always
    
  `@openupgrade.migrate` decorator avoids to execute the script if version is none. As we have reverted the openupgrade_framework patch that modifies the version, we avoid the side effect and get this script working removing the decorator.

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/modules/migration.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/migration.py
@@ -15,11 +15,9 @@ def migrate_module(self, pkg, stage):
     to_install = pkg.state == "to install"
     if to_install:
         pkg.state = "to upgrade"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
And that's it! Migration scripts are still being executed for the split module: https://github.com/OCA/OpenUpgrade/actions/runs/16076314034/job/45372272328?pr=5151#step:13:2683

@hparfr can you try with your case if the error is gone?

hparfr:
Thanks @pedrobaeza  !

- the migration pass  !
- account_edi_ubl_cii is installed during the migration

PR #5152, opened 2025-07-05
[18.0][OU-IMP] account: fill account_bank_statement_line company_id

jguenat:
seems new company_id field on account_bank_statement_line was not filled.
(already specifed in _work.txt so maybe a oversight)

### DIFF
--- a/openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
@@ -169,6 +169,25 @@ def fill_account_payment(env):
     )
 
 
+def fill_statement_line_fields(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line sl
+        SET company_id = mv.company_id
+        FROM account_move mv
+        WHERE sl.move_id = mv.id""",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line sl
+        SET journal_id = mv.journal_id
+        FROM account_move mv
+        WHERE sl.move_id = mv.id""",
+    )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     if openupgrade.column_exists(env.cr, "account_cash_rounding", "profit_account_id"):
@@ -199,6 +218,7 @@ def migrate(env, version):
     adapt_account_move_sending_data(env)
     rename_selection_option(env)
     fill_account_payment(env)
+    fill_statement_line_fields(env)
     openupgrade.convert_field_to_html(
         env.cr,
         "account_tax",


openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
```diff
@@ -169,6 +169,17 @@ def fill_account_payment(env):
...
 
+def fill_statement_line_company(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line sl
+        SET company_id = mv.company_id
+        FROM account_move mv
+        WHERE sl.move_id = mv.id""",
+    )
```
MiquelRForgeFlow, line 180: Please, also add this:
```python
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE account_bank_statement_line sl
        SET journal_id = mv.journal_id
        FROM account_move mv
        WHERE sl.move_id = mv.id""",
    )
```

openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
```diff
@@ -169,6 +169,17 @@ def fill_account_payment(env):
     )
 
 
+def fill_statement_line_company(env):
```
MiquelRForgeFlow:
```suggestion
def fill_statement_line_fields(env):
```

MiquelRForgeFlow (commented):
Oh, my bad, I forgot to add this. Thank you.

MiquelRForgeFlow (approved):

jguenat:
I rebased

openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
```diff
@@ -169,6 +169,25 @@ def fill_account_payment(env):
...
 
 
+def fill_statement_line_fields(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line sl
+        SET company_id = mv.company_id
+        FROM account_move mv
+        WHERE sl.move_id = mv.id""",
```
pedrobaeza, line 179:
```suggestion
        WHERE sl.move_id = mv.id
        AND sl.company_id IS NULL
        """,
```

openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
```diff
@@ -169,6 +169,25 @@ def fill_account_payment(env):
...
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_bank_statement_line sl
+        SET journal_id = mv.journal_id
+        FROM account_move mv
+        WHERE sl.move_id = mv.id""",
```
pedrobaeza, line 187:
```suggestion
        WHERE sl.move_id = mv.id
        AND sl.journal_id IS NULL
        """,
```

openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
jguenat: Is there any reason company_id might not be NULL ? Field was not stored in 17.0.
Or just a good practice ?

pedrobaeza: A good practice, as you may have that field filled by other means.

pedrobaeza: Oh, forget it. You are creating the columns upper, so no need.

pedrobaeza (approved):

PR #5154, opened 2025-07-07
[18.0][OU-FIX] hr_expense: handle rule

MiquelRForgeFlow:
Caught in https://github.com/OCA/OpenUpgrade/pull/5144:
Selection_4749


pedrobaeza:
:+1:

Issue #5150, opened 2025-07-03
Migration scripts running for un-installed modules


hparfr:
Hello,

There is some migration script running for not installed modules.

For instance, in v18, account_edi_ubl_cii is not installed in my db. And the migration fails since the addition of https://github.com/OCA/OpenUpgrade/pull/5143  (introducing an SQL update on not-existing column)

```
odoo.modules.loading: Loading module account_edi_ubl_cii (52/106) 
odoo.modules.migration: module account_edi_ubl_cii: Running migration [>18.0.1.0] pre-migration 
OpenUpgrade: account_edi_ubl_cii: pre-migration script called with version 0.0 
odoo.sql_db: bad query: b"UPDATE res_partner SET peppol_eas = NULL WHERE peppol_eas IN ('0212', '0215', '9901')"
ERROR: ERREUR:  la colonne « peppol_eas » n'existe pas
LINE 1: UPDATE res_partner SET peppol_eas = NULL WHERE peppol_eas IN...
                                                       ^
```

`account_edi_ubl_cii` is `autoinstall=True`  and it has been uninstalled in a previous version in this db.

Is it expected to run migrations scripts for uninstalled modules ?

hbrunn:
that's ultimately caused by https://github.com/OCA/OpenUpgrade/pull/5147

@OCA/openupgrade-maintainers the more I think about it, the more I don't think we should be running pre-migration scripts for newly installed modules. If we do this, we'll have to rewrite all our migration scripts to be much more defensive on what assumptions to make.

StefanRijnhart:
I agree it is a painful subversion of the concept of a migration script. But is there an alternative to make the migration work in the case of split off glue modules?

pedrobaeza:
Well, the need is there and we need to run that scripts, but the issue is another thing, as it's executing end-migration script even if not installed, so we should handle that case independently from the installation one.

MiquelRForgeFlow:
> `account_edi_ubl_cii` is `autoinstall=True` and it has been uninstalled in a previous version in this db.

@hparfr if you don't want `account_edi_ubl_cii`, you have to put `autoinstall=False` in your custom migration environment. And you can manually install it later without using OU if you want it.

hbrunn:
what is the need though? what currently happens for event_product can just as well happen in sale's migration script after detecting the conditions. what would be a case not solvable in the migration script of the migration script the installed module depends on?

hparfr:
I don't know OU internals.

[...]

PR #5155, opened 2025-07-07
[UPD] faq: precise behavior about uninstalled auto_install modules

hparfr:
Related to #5150

Inform that some modules may be reinstalled during a migration

### DIFF
--- a/docsource/060_odoo_migration_manager.rst
+++ b/docsource/060_odoo_migration_manager.rst
@@ -48,3 +48,10 @@ Are migration scripts fired when installing new modules?
    would never ever do so anyway. Developers are free to corrupt the regular
    workings of Odoo if it helps the migration. For instance, from OpenUpgrade
    9.0 on, the workflow engine is disabled during field recomputation.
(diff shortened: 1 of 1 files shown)


docsource/060_odoo_migration_manager.rst
```diff
@@ -48,3 +48,7 @@ Are migration scripts fired when installing new modules?
    would never ever do so anyway. Developers are free to corrupt the regular
    workings of Odoo if it helps the migration. For instance, from OpenUpgrade
    9.0 on, the workflow engine is disabled during field recomputation.
+
+How are handled auto-installed modules that have been manually uninstalled?
+   Modules with auto_install=True (like IAP) that have been manually uninstalled
+   are automatically installed during the migration.
```
legalsylvain, line 54: We could add a text like "if you dont want to install such module you can use module_change_auto_install in each migration step to force auto_install to False.

https://github.com/OCA/server-tools/tree/18.0/module_change_auto_install

PR #5163, opened 2025-07-10
[17.0][OU-FIX] account: assign xmlid to journals

### LINKED ISSUE #5108: [17.0] account try_loading chart template creates duplicate journal
## Module
Account

## Describe the bug
Chart template reload has changed a lot between version 16.0 and 17.0. In particular on the account.journal part, it will singe 17.0 assign XML-ids to default journals which was not the case before (as part of https://github.com/odoo/odoo/pull/110016).

It tries to map existing journals with either type+name or code (https://github.com/odoo/odoo/pull/128139) but if both were changed manually it will not find those and recreate them when reloading chart template.

The issue was discovered when testing migration to version 18.0 on our own database where purchase journal code had been changed and name was different (extra s at the end for plural). Also, while migrating french localized database to 18.0, because of new module l10n_fr_account renamed from l10n_fr_fec, Odoo end migration script for version 2.1 was run which reloads chart_template (but the issue is linked to v17 adding this reload, but was not spotted before because it was not called)

## To Reproduce
**Affected versions**: 17.0

1. Install l10n_fr on v16 database
2. Update account journal names and codes for FR Company
3. Migrate to version 17.0
4. Reload chart of account
5. Check for duplicate journals

**Expected behavior**
Journals should not ne recreated

**Additional context**
I found the issue with our database using Fr localization but the problem is likely to ne the same on all localizations unless properly handled in specific locale scripts

remi-filament:
This PR fixes #5108 for duplicate journals

When testing this PR I also realized there was a problem assigning chart_template from chart_template_id since company_id was retrieved instead of chart_template_id. I fixed that as well.

I had to copy a few lines from Odoo account/models/chart_template.py code, I added information about that lines in comments, let me know if you think this is not acceptable ?

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/end-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/end-migration.py
@@ -1,0 +1,43 @@
+# Copyright 2025 Le Filament (https://le-filament.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def _assign_journal_xmlids(env):
+    """
+    Starting in Odoo 17, Odoo expects journals to get external id
+    like account.1_sale for sale journal on company 1
+    We therefore assign xmlids to existing journals so that they will not be recreated
+    """
+    for company in env["res.company"].search([("chart_template", "!=", False)]):
+        template_data = env["account.chart.template"]._get_chart_template_data(
+            company.chart_template
+        )
+        existing_journals = (
+            env["account.journal"]
+            .with_context(active_test=False)
+            .search(env["account.journal"]._check_company_domain(company), order="id")
+        )
+        for xmlid, journal_data in list(
+            template_data.get("account.journal", {}).items()
+        ):
+            if not env.ref(xmlid, raise_if_not_found=False) and "type" in journal_data:
+                journal = existing_journals.filtered(
+                    lambda j: j.type == journal_data["type"]  # noqa: B023
+                )[:1]
+                if journal:
+                    existing_journals -= journal
+                    env["ir.model.data"]._update_xmlids(
+                        [
+                            {
+                                "xml_id": f"account.{company.id}_{xmlid}",
+                                "record": journal,
+                                "noupdate": True,
+                            }
+                        ]
+                    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    _assign_journal_xmlids(env)


openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
```diff
@@ -558,6 +558,104 @@ def _rename_coa_elements_xmlids(env):
...
+    Template = env["account.chart.template"]
+    env.cr.execute(
+        """SELECT id AS company_id, chart_template
+            FROM res_company
+            WHERE chart_template IS NOT NULL
+        """,
+    )
+    for company_id, chart_template in env.cr.fetchall():
+        company = env["res.company"].browse(company_id)
+        template_data = Template._get_chart_template_data(chart_template)
```
MiquelRForgeFlow: It seems this script has to be an end-migration.

remi-filament: I did not get that error while testing locally, but yes it makes sense. I just fixed that !
Thanks !

pedrobaeza (commented):
I think you may be overcomplicating the code. I think you just need to look for the first journal of each type and assign the XML-ID. Matching per name doesn't worth, as you may rename the journals, install them in another language, or a lot of reasons.

pedrobaeza:
And please put the other fix in a separate PR, or at least commit.

remi-filament:
Hi @pedrobaeza thanks for your comments, I moved fix on assign chart_template to #5166.
Regarding the complex code, I believe it is worth it for a few reasons :
- this is what Odoo core is trying to achieve when updating chart of account (https://github.com/odoo/odoo/blob/7df78a28dd7c05b1ea246e9f0735d8da0cb5486e/addons/account/models/chart_template.py#L257-L284), with an additional fallback to first not yet used journal of correct type
- for type = "general", you get at least 3 journals (Miscellaneous Operations, Exchange Difference, Cash Basis Taxes) + 1 extra if stock_account is installed (Inventory Valuation) - for this specific case, I believe it is better to try to find the correct one instead of assigning them randomly, falling back to assigning in order if you get no match on code / name.

openupgrade_scripts/scripts/account/17.0.1.2/end-migration.py
```diff
@@ -0,0 +1,107 @@
...
+                        .with_context(active_test=False)
+                        .search(
+                            [
+                                *env["account.journal"]._check_company_domain(company),
+                                ("type", "=", journal_data["type"]),
+                                ("id", "not in", assigned_journals.ids),
+                            ],
+                            limit=1,
+                        )
+                    )
```
MiquelRForgeFlow, line 90: maybe you could do two calls of this loop:

```python
for xmlid, journal_data in list(
            template_data.get("account.journal", {}).items()
        ):
```
the first loop containing the exact odoo code. and the second loop only containing this part you are adding if xmlid doesn't exist yet.

pedrobaeza:
@remi-filament I still think it should be simplified, as all that code Odoo core is if not XML-ID is found, but here we want to just assign one for not triggering that code, so it's enough assigning one journal, not matter which.

remi-filament:
Fine @pedrobaeza, I was afraid that maybe somewhere in the code Odoo is trying to get account_journal by XML-id and would fail if we set the incorrect one, but maybe it does not happen.
I simplified the code to assign first of correct type instead !

Let me know if this is OK for you ?

openupgrade_scripts/scripts/account/17.0.1.2/end-migration.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+        for xmlid, journal_data in list(
+            template_data.get("account.journal", {}).items()
+        ):
+            if not env.ref(xmlid, raise_if_not_found=False) and "type" in journal_data:
+                journal = (
+                    existing_journals.filtered(lambda j: j.type == journal_data["type"])  # noqa: B023
+                    or [False]
+                )[0]
+                if journal:
```
MiquelRForgeFlow:
```suggestion
                    existing_journals.filtered(lambda j: j.type == journal_data["type"])  # noqa: B023
                )[:1]
                if journal:
```
empty record already is not True

remi-filament: Thanks @MiquelRForgeFlow 
My problem here is that I only want the first record (when the filtered() may return a recordset of multiple records)
So I added a [0], but if filtered() returns empty recordset, the [0] is returning an error, which is why I replaced empty recordset by [False] !
There is probably a more elegant way to do that that I do not know !

MiquelRForgeFlow: `[:1]` does the trick.

openupgrade_scripts/scripts/account/17.0.1.2/end-migration.py
```diff
@@ -0,0 +1,44 @@
...
+            .search(env["account.journal"]._check_company_domain(company), order="id")
+        )
+        for xmlid, journal_data in list(
+            template_data.get("account.journal", {}).items()
+        ):
+            if not env.ref(xmlid, raise_if_not_found=False) and "type" in journal_data:
+                journal = (
+                    existing_journals.filtered(lambda j: j.type == journal_data["type"])  # noqa: B023
+                    or [False]
+                )[0]
```
pedrobaeza:
```suggestion
                journal = existing_journals.filtered(
                    lambda j: j.type == journal_data["type"]
                )[:1]
```

remi-filament: Thanks for the tip ! I have added it !

MiquelRForgeFlow (approved):
Hope this fix is fine and we don't need to add more conditions (like the `code`) in the filtered later.

pedrobaeza (approved):

PR #5167, opened 2025-07-11
[18.0][FIX] openupgrade_framework: fix return value for update_list method patch

kafai-lam:
since [`update_list` will be called by other module](https://github.com/odoo/odoo/blob/c0183a81a9150c987522a7803a6e4598c0e72b7c/odoo/addons/base/wizard/base_module_update.py#L16), it will be better to keep the original return value


pedrobaeza:
If the returned result is handled, I think you should add the auto installed modules to the added returned value, isn't it?

hbrunn (approved):
shouldn't we merge this first and talk about the details later?

pedrobaeza:
OK. I think it's everything related, but let's merge this for now. 

/ocabot merge patch

PR #5035, opened 2025-05-14
[17.0][FIX] hr_attendance: Define the corresponding groups in the menus

victoralmau:
Define the corresponding groups in the menus

**Attendances**

16.0: `<menuitem id="menu_hr_attendance_root" groups="hr_attendance.group_hr_attendance,hr_attendance.group_hr_attendance_kiosk" `
17.0: `<menuitem id="menu_hr_attendance_root" groups="hr_attendance.group_hr_attendance_officer"`

**Attendances > Kiosk Mode**

16.0: `<menuitem id="menu_hr_attendance_kiosk_no_user_mode" groups="hr_attendance.group_hr_attendance_kiosk"`
17.0: `<menuitem id="menu_hr_attendance_kiosk_no_user_mode" groups="hr_attendance.group_hr_attendance_manager"`

**Note**: Attendances > Overview menu does not need any change, because although the group is not the same (`hr_attendance.group_hr_attendance_user` vs `hr_attendance.group_hr_attendance_officer`), that group has been renamed in pre-migration.

Please @pedrobaeza can you review it?

@Tecnativa

PR #5169, opened 2025-07-14
[17.0][FIX] hr_attendance: omitted group

FernandoRomera:
NEW res.groups: hr_attendance.group_hr_attendance_own_reader
DEL res.groups: hr_attendance.group_hr_attendance_kiosk
DONE: pre-migration: rename group

NEW res.groups: hr_attendance.group_hr_attendance_own_reader
DEL res.groups: hr_attendance.group_hr_attendance
DONE: pre-migration: join to same group

### DIFF
--- a/openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
+++ b/openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
@@ -67,6 +67,16 @@ def hr_attendance_menus(env):
             ]
         }
     )
+    openupgrade.rename_xmlids(
+        env.cr,
+        [
+            (
+                "hr_attendance.group_hr_attendance",
+                "hr_attendance.group_hr_attendance_own_reader",
+            ),
+        ],
+        allow_merge=True,
+    )
     # Remove the groups of 16.0 from the Attendances > Kiosk Mode menu
     env.ref("hr_attendance.menu_hr_attendance_kiosk_no_user_mode").write(
         {
@@ -74,6 +84,16 @@ def hr_attendance_menus(env):
                 Command.unlink(group_hr_attendance_kiosk.id),
             ]
         }
+    )
+    openupgrade.rename_xmlids(
+        env.cr,
+        [
+            (
+                "hr_attendance.group_hr_attendance_kiosk",
+                "hr_attendance.group_hr_attendance_own_reader",
+            ),
+        ],
+        allow_merge=True,
     )
 
 
--- a/openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/hr_attendance/17.0.2.0/upgrade_analysis_work.txt
@@ -102,7 +102,8 @@ DEL res.groups: hr_attendance.group_hr_attendance_user
 
 NEW res.groups: hr_attendance.group_hr_attendance_own_reader
 DEL res.groups: hr_attendance.group_hr_attendance_kiosk
-# DONE: pre-migration: rename group
+# DONE: post-migration: merge group
 
+NEW res.groups: hr_attendance.group_hr_attendance_own_reader
 DEL res.groups: hr_attendance.group_hr_attendance
-# NOTHING TO DO
+# DONE: post-migration: merge group


FernandoRomera:
@MiquelRForgeFlow 
Can you review this PR, please?
The first part I think is correct, the second part I am not so sure.

MiquelRForgeFlow:
Please, update `upgrade_analysis_work.txt` file.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
```diff
@@ -56,25 +56,35 @@ def fill_hr_attendance_overtime_hours(env):
...
-        }
-    )
-    # Remove the groups of 16.0 from the Attendances > Kiosk Mode menu
-    env.ref("hr_attendance.menu_hr_attendance_kiosk_no_user_mode").write(
-        {
-            "groups_id": [
-                Command.unlink(group_hr_attendance_kiosk.id),
-            ]
-        }
+    group_hr_attendance = env.ref("hr_attendance.group_hr_attendance", False)
```
MiquelRForgeFlow: This doesn't exist nor group_hr_attendance_kiosk, because you merged into `group_hr_attendance_own_reader`.

I think you should avoid doing changes in this method. The way to proceed should be doing _xmlid_merge of the two xmlids at the last step of post-migration instead of doing in pre-migration.

FernandoRomera: Thanks.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
```diff
@@ -56,25 +56,56 @@ def fill_hr_attendance_overtime_hours(env):
...
+    if group_hr_attendance:
```
MiquelRForgeFlow: I don't know why this and group_hr_attendance_kiosk should be False. They should always exist.

FernandoRomera: Thanks.

MiquelRForgeFlow (approved):
LGTM

AntoniRomera (approved):
LGTM

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
```diff
@@ -57,12 +57,29 @@ def fill_hr_attendance_overtime_hours(env):
 
 def hr_attendance_menus(env):
     group_hr_attendance = env.ref("hr_attendance.group_hr_attendance")
-    group_hr_attendance_kiosk = env.ref("hr_attendance.group_hr_attendance_kiosk")
```
MiquelRForgeFlow: nitpick 1: could you avoid reorganizing this?

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
```diff
@@ -57,12 +57,29 @@ def fill_hr_attendance_overtime_hours(env):
...
+            (
+                "hr_attendance.group_hr_attendance",
+                "hr_attendance.group_hr_attendance_own_reader",
+            ),
+        ],
+        allow_merge=True,
+    )
+    group_hr_attendance_kiosk = env.ref("hr_attendance.group_hr_attendance_kiosk")
+    # Remove the groups of 16.0 from the Attendances menu
+    env.ref("hr_attendance.menu_hr_attendance_root").write(
```
MiquelRForgeFlow: nitpick 2: there is no point in doing two write on same record. Let's do only 1 write.

MiquelRForgeFlow (commented):
Some nitpickings...

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
FernandoRomera: Thanks.

openupgrade_scripts/scripts/hr_attendance/17.0.2.0/post-migration.py
FernandoRomera: Thanks.

pedrobaeza:
@victoralmau you were the original author of this. Do you consider it correct?

victoralmau:
Related to https://github.com/OCA/OpenUpgrade/pull/5035

While I don't consider it “totally wrong” IMO it can be confusing. The reason?

https://github.com/odoo/odoo/blob/1f0af705eb2c1c39f35e4327cf9e5e4df35bdd91/addons/hr_attendance/security/hr_attendance_security.xml#L14C65-L14C95 any user belonging to `base.group_user` will already have the `hr_attendance.group_hr_attendance_own_reader` (User: Read his own attendances) permission, so I don't think this was necessary. The same happens in the case of `group_hr_attendance_kiosk` (in v16 https://github.com/odoo/odoo/blob/cdeca0417ee05f05360d2ea0fb56c4b90654c16d/addons/hr_attendance/security/hr_attendance_security.xml#L11)

PR #5191, opened 2025-07-24
[18.0][IMP] apriori.py: document modules that moved to a different repo

thienvh332:
This PR introduces a `moved_modules` dictionary to the `apriori.py` script for the `Odoo 18.0` migration, following the same approach as in https://github.com/OCA/OpenUpgrade/pull/5189 for v16.0.


openupgrade_scripts/apriori.py
```diff
@@ -92,3 +92,35 @@
     "mrp.document": "product.document",
     # OCA/...
 }
+
+# List of OCA modules that were moved to other repositories .
+# Format: {"module_name": ("old_repo_name", "new_repo_name")}
+# Note: The key is the module_name for easier reference
+moved_modules = {
+    # OCA/social
```
MiquelRForgeFlow, line 100: `mail_debrand` too.

pedrobaeza:
Uhm, I don't really agree to put this here, as this is not affecting anything to OpenUpgrade.

nilshamerlinck:
Hi @pedrobaeza 
- it's just adding an innocuous variable, no downside;
- it also strengthens the status of `apriori.py` as the de facto source of reference for migration-related information;
- if not here, where would you see it fit?

pedrobaeza:
This is maintenance burden for the OpenUpgrade project, and there are a lot of chances of this to be outdated/incomplete.

This information should be irrelevant for any deployment. If you deploy with pip, the package name is the same being in one repo or another. If you deploy with GitHub (as me), you should have a system doing this task for you. We do it through modules in https://github.com/OCA/interface-github/

PR #5193, opened 2025-07-24
[18.0][OU-ADD] mass_mailing

MiquelRForgeFlow:
.


openupgrade_scripts/scripts/mass_mailing/18.0.2.7/noupdate_changes_work.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='utf-8' ?>
```
pedrobaeza, line 1: If no change is performed in the file, not sure if we should create a copy always. What do you think, @hbrunn ?

hbrunn: I agree. If the `noupdate_changes.xml` file changes, chances are we can just leave it at that, while with loading the work file, this always needs manual intervention

PR #5195, opened 2025-07-24
[18.0][OU-FIX] stock: fill picking type default locations in end-migration

MiquelRForgeFlow:
Other modules like `stock_dropshipping` and `repair` customize the default.


MiquelRForgeFlow:
@hbrunn how can we execute the `stock` tests **after** **end**-migration is executed?

hbrunn (approved):
change makes sense I think.

About tests: Never tried that, but chances are just tagging with `post_install` could work.

MiquelRForgeFlow:
> About tests: Never tried that, but chances are just tagging with post_install could work.

It didn't work :(

hbrunn:
well that's a shame, will have to look into this eventually. Deactivate the test for now?

PR #5036, opened 2025-05-15
[18.0][OU-ADD] project: Nothing to do

BhaveshHeliconia:

PR #5203, opened 2025-07-25
[18.0][OU-ADD] project

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/5036.


openupgrade_scripts/scripts/project/18.0.1.3/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _field_renames)
+    rule = env.ref(
+        "project_todo.task_visibility_rule_project_user", raise_if_not_found=False
+    )
+    if rule:
```
pedrobaeza, line 14: @MiquelRForgeFlow you don't need to do this check, as simply calling the rename method will perform the same: if it doesn't exist, nothing happens.

openupgrade_scripts/scripts/project/18.0.1.3/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,67 @@
+---Models in module 'project'---
+new model project.share.collaborator.wizard [transient]
+---Fields in module 'project'---
+project      / project.collaborator     / limited_access (boolean)      : NEW hasdefault: default
+# NOTHING TO DO
```
pedrobaeza, line 5: Please put why is nothing to do.

openupgrade_scripts/scripts/project/18.0.1.3/pre-migration.py
MiquelRForgeFlow: It's to avoid this log: Selection_4853

I mean, if somebody doesn't have project_todo and see this log, they may wrongly worry why it's missing an xmlid.

openupgrade_scripts/scripts/project/18.0.1.3/upgrade_analysis_work.txt
MiquelRForgeFlow: boolean with default False

pedrobaeza: But please put it on the work file for other times, as stated several times. We don't know from this file which is the default and why it's correct to let it that way.

openupgrade_scripts/scripts/project/18.0.1.3/pre-migration.py
pedrobaeza: It's a info one, so no problem, and it's done this way across whole OpenUpgrade.

PR #5220, opened 2025-07-28
[18.0][OU-ADD] account_invoice_mass_sending: Merged into account

victoralmau:
Merged into `account`

The [cron job](https://github.com/odoo/odoo/blob/7c95e2d3d6e55976abbafaa296378cb8ff44bef9/addons/account/data/service_cron.xml#L17C58-L17C67) "Send invoices automatically" already exists in the `account` and will automatically send pending invoices; therefore, you simply need to set the appropriate frequency for the cron job (every 5 minutes, for example).

@Tecnativa TT54853

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -62,6 +62,7 @@ merged_modules = {
     "website_sale_product_configurator": "website_sale",
     # odoo/enterprise
     # OCA/account-invoicing
+    "account_invoice_mass_sending": "account",
     "account_invoice_supplierinfo_update_discount": "account_invoice_supplierinfo_update",  # noqa: E501
     # OCA/e-commerce
     "website_sale_product_attachment": "website_sale",


pedrobaeza:
Please solve conflicts. Can you explain a bit why this module is not needed anymore in 18?

MiquelRForgeFlow:
The `account_invoice_mass_sending` uses `queue_job`. How everything in that module is replaced by a simple cron?

victoralmau:
> The `account_invoice_mass_sending` uses `queue_job`. How everything in that module is replaced by a simple cron?

The method https://github.com/odoo/odoo/blob/7c95e2d3d6e55976abbafaa296378cb8ff44bef9/addons/account/models/account_move.py#L5487 takes care of everything. I haven't looked into it too deeply, but it seems pretty consistent. In other modules, the dependency on `queue_job` has already been removed, so I know it's possible: https://github.com/OCA/l10n-spain/pull/3737

pedrobaeza:
@victoralmau please solve conflicts.

pedrobaeza (approved):

PR #5234, opened 2025-07-29
[18.0][OU-ADD] sale_expense: Nothing to do

carlos-lopez-tecnativa:
TT55915
@Tecnativa  @pedrobaeza @MiquelRForgeFlow could you please review this?


pedrobaeza (approved):
@legalsylvain the "Blocked by dependency" tag is not correct.

PR #5239, opened 2025-07-29
[17.0][OU-FIX] sale: Update also sale.order.line~state

pedrobaeza:
This is a related stored field that is not updated in SQL, so we have to sync it as well.

@Tecnativa


sergio-teruel (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #5242, opened 2025-07-30
[18.0][OU-FIX] payment_paypal: archive paypal provider

MiquelRForgeFlow:
Fixes https://github.com/OCA/OpenUpgrade/commit/c190b0d4b45a56f9e55a147241aec9a0fda5e8f6.


pedrobaeza:
This is too heavy IMO... I think it's better that UI stops people complaining about the missing data, than having to look for it in the archived providers.

PR #5249, opened 2025-07-30
[18.0][OU-ADD] website_blog: Nothing to do

carlos-lopez-tecnativa:
TT54376


openupgrade_scripts/scripts/website_blog/18.0.1.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,12 @@
+---Models in module 'website_blog'---
+---Fields in module 'website_blog'---
+website_blog / blog.tag                 / color (integer)               : NEW
```
pedrobaeza, line 3: Is the default empty value equivalent to previously how they were viewed in 17?

carlos-lopez-tecnativa: Yes, `Tag 1` and `Tag 2` are records from a V17 database, while `New Tag 18` was created after the migration.
They all look identical, so the default value appears to be equivalent.

image

image

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/website_blog/18.0.1.1/upgrade_analysis_work.txt
pedrobaeza: That's on backend, and in frontend?

carlos-lopez-tecnativa: V17
image

V18
image

pedrobaeza: So "Tag1" and "Tag2" are not seeing the same in v18, and we may need to put them in white or similar.

carlos-lopez-tecnativa: Sorry, but I don't understand. The three tags look the same to me. What did you expect?

carlos-lopez-tecnativa: The font color that you see is from V17, but in V18 I think Odoo removed this font color, so all the tags look identical.

Only if I set a color manually in the backend, that color is displayed in the frontend.

Otherwise, the colors are all the same.

image

image

pedrobaeza: But the idea is that when migrating, things look the same as in previous version, not with the v18 defaults, but well, let's move in this case, as it's not very significant.

pedrobaeza (approved):

PR #5251, opened 2025-07-31
[18.0][OU-ADD] Some project related modules

MiquelRForgeFlow:
.


openupgrade_scripts/scripts/project_purchase/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,9 @@
+---Models in module 'project_purchase'---
+---Fields in module 'project_purchase'---
+project_purchase / purchase.order           / project_id (many2one)         : NEW relation: project.project
```
carlos-lopez-tecnativa, line 3: Just a doubt, does it make sense to fill the `project_id` based on the analytic account of the `purchase.order.line`?

MiquelRForgeFlow: It makes sense, but I am no sure. You should analyze properly and check if it's true or not, with a proper valid argument. If you have it, please do a PR.

PR #5257, opened 2025-07-31
[18.0][OU-ADD] website_project: Migration to 18.0. Renamed from website_form_project

carlos-lopez-tecnativa:
TT54680


openupgrade_scripts/scripts/website_project/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,18 @@
+---Models in module 'website_project'---
+---Fields in module 'website_project'---
+website_project / project.task             / email_from (char)             : NEW
+# TODO: pre-create this field and populate it from the partner's email?
```
MiquelRForgeFlow: I would say... No. It seems a technical field, where in some cases they need to save the email, bot not always, if I am not wrong 🤔
Let's wait a second opinion.

pedrobaeza: This was removed in 17.0 from `project`:

https://github.com/OCA/OpenUpgrade/blob/a1ee86c71709a2762dca4d0e39bed6051f25498c/openupgrade_scripts/scripts/project/17.0.1.3/upgrade_analysis.txt#L30

and according to the comment, it's a mechanism for checking possible email loops (embedded in `mail`), so this field can be the same useful for any gateway entry, including `project` itself. This seems an ugly patch embedded here for the specific case where you have an automatic reply to the form.

Anyway, this should be `NOTHING TO DO` indeed.

carlos-lopez-tecnativa: Thanks to both of you for your comments. I investigated the original commit and found this:
https://github.com/odoo/odoo/commit/1aa5cbb06be85e94c01f8f55ef57b9415b9bb50f
It makes things clearer for me.
Please let me know if this is OK for you or if I need to make any further changes.

pedrobaeza (approved):

PR #5261, opened 2025-08-01
[18.0][OU-ADD] purchase_requisition : Migration to 18.0

carlos-lopez-tecnativa:
TT54474


openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
...
+purchase_requisition / purchase.requisition.line / schedule_date (date)          : DEL
+# NOTHING TO DO: These fields are not used.
+
+purchase_requisition / purchase.requisition.type / active (boolean)              : DEL
+purchase_requisition / purchase.requisition.type / exclusive (selection)         : DEL required, selection_keys: ['exclusive', 'multiple']
+purchase_requisition / purchase.requisition.type / line_copy (selection)         : DEL required, selection_keys: ['copy', 'none']
+purchase_requisition / purchase.requisition.type / name (char)                   : DEL required
+purchase_requisition / purchase.requisition.type / quantity_copy (selection)     : DEL required, selection_keys: ['copy', 'none']
+purchase_requisition / purchase.requisition.type / sequence (integer)            : DEL
+# NOTHING TO DO: Odoo replaced this model with a new selection field
```
MiquelRForgeFlow: call openupgrade.remove_tables_fks on this table

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
...
+---Fields in module 'purchase_requisition'---
+purchase_requisition / purchase.order.line      / price_total_cc (float)        : NEW isfunction: function, stored
+# DONE: pre-migration: pre-create and filled
+
+purchase_requisition / purchase.requisition     / active (boolean)              : NEW hasdefault: default
+purchase_requisition / purchase.requisition     / date_end (datetime)           : type is now 'date' ('datetime')
+# NOTHING TO DO: Handled by ORM
+
+purchase_requisition / purchase.requisition     / schedule_date (date)          : DEL
+purchase_requisition / purchase.requisition     / date_start (date)             : NEW
```
MiquelRForgeFlow, line 12: image

I would say that `ordering_date` -> `date_start`

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
+---Models in module 'purchase_requisition'---
+obsolete model purchase.requisition.type
+---Fields in module 'purchase_requisition'---
+purchase_requisition / purchase.order.line      / price_total_cc (float)        : NEW isfunction: function, stored
+# DONE: pre-migration: pre-create and filled
+
+purchase_requisition / purchase.requisition     / active (boolean)              : NEW hasdefault: default
```
MiquelRForgeFlow, line 7: active = type_id.active

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
...
+purchase_requisition / purchase.requisition     / date_end (datetime)           : type is now 'date' ('datetime')
```
MiquelRForgeFlow, line 8: I am not sure. Have you tested it to assure the column has been converted correctly?

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Why? This is a new field in `purchase.requisition`. I don't think that just because a requisition type is archived, all records linked to that type must also be archived. This new field is independent of the type.
Please explain to me why we mark active according to the requisition type.

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
carlos-lopez-tecnativa: I see the database, and PostgreSQL truncates the date. I think it's correct, or do you suggest another way to do it?
image

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
carlos-lopez-tecnativa: I based my reasoning on the logic in the code, not the search view.
When a purchase line is created from a requisition:
In V17, the field used is `schedule_date`:
https://github.com/odoo/odoo/blob/6dba451dc50bfbb4ec4f85b168bc542b0d66cf41/addons/purchase_requisition/models/purchase_requisition.py#L265
In V18, the field used is `date_start`:
https://github.com/odoo/odoo/blob/b20aa39b71635bb071ced480fb80dfa8f464dfb5/addons/purchase_requisition/models/purchase_requisition.py#L268

So, for me, `schedule_date` maps to `date_start`.
Additionally, in V17 the field `ordering_date` is not used in the code.

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Sorry for not indicating this in the file, but yes, I called this function in the `pre-migration`.
I've updated the file, please let me know if it's okay now or if I need to do anything else.

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
MiquelRForgeFlow: Ok, makes sense what you said.

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
MiquelRForgeFlow: > PostgreSQL truncates the date

Yes, but that part filled with 00:00:00s is ugly. Date columns don't have it:
Selection_4860

> do you suggest another way to do it?

rename_column + add_columns (with date type), and then doing a query update using old_value::date? 😅

carlos-lopez-tecnativa: I think it's a visual effect depending on the Postgres client used.
My first image was with `PgWeb`; now I checked with `PgAdmin` and it appears as a Date without 00:00:00.
Even so, do you consider any changes necessary according to your suggestions, or is the current code OK?

image

MiquelRForgeFlow: No, it's ok :)

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/pre-migration.py
```diff
@@ -0,0 +1,68 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version=None):
+    openupgrade.copy_columns(
+        env.cr,
+        {
+            "purchase_requisition": [
+                ("state", None, None),
+                ("schedule_date", None, None),
+                ("origin", None, None),
```
pedrobaeza: On a simple renaming, you don't have to copy the column, as the values are not overwritten. For `state` one, you have to, as the original values are changed.

```suggestion
```

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
...
+# DONE: pre-migration: copied column for preservation and rename fields
+
+purchase_requisition / purchase.requisition     / requisition_type (selection)  : NEW required, selection_keys: ['blanket_order', 'purchase_template'], hasdefault: default
+# DONE: pre-migration: pre-created and filled according to this logic
+# if quantity_copy = 'none' then 'blanket_order', else 'purchase_template'
+
+purchase_requisition / purchase.requisition     / state (selection)             : selection_keys added: [confirmed], removed: [in_progress, ongoing, open]
+# DONE: pre-migration: copied column for preservation and mapped values: in_progress -> confirmed. ongoing -> confirmed. open -> confirmed.
+
+purchase_requisition / purchase.requisition     / type_id (many2one)            : DEL relation: purchase.requisition.type, required
```
pedrobaeza, line 21: You should move this row next to `requisition_type` as a block and put the explanation together.

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
...
+# NOTHING TO DO: Odoo replaced this model with a new selection field
+# See https://github.com/odoo/odoo/pull/129114
+
+purchase_requisition / purchase.requisition.line / analytic_distribution_search (json): DEL stored: False
+# NOTHING TO DO
```
pedrobaeza: Remember to always put why is a nothing to do:

```suggestion
# NOTHING TO DO: Helper field needed in previous version for searches
```

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,61 @@
...
+
+purchase_requisition / purchase.requisition     / ordering_date (date)          : DEL
+purchase_requisition / purchase.requisition.line / schedule_date (date)          : DEL
+# NOTHING TO DO: These fields are not used.
+
+purchase_requisition / purchase.requisition.type / active (boolean)              : DEL
```
pedrobaeza, line 24: You may want to move these rows together with the block about the new selection field.

openupgrade_scripts/scripts/purchase_requisition/18.0.0.1/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Done. Please let me know if it’s Ok for you now.

carlos-lopez-tecnativa:

pedrobaeza (approved):

PR #5217, opened 2025-07-26
[18.0][OU-ADD] website_sale: Migration scripts

pedrobaeza:
It also includes a change in the prettier configuration for not touching the new noupdate_changes_work.xml files.

TT54385 @Tecnativa

PR #5262, opened 2025-08-01
[18.0][OU-ADD] stock_picking_batch: Migration to 18.0

carlos-lopez-tecnativa:
TT54330


openupgrade_scripts/scripts/stock_picking_batch/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,22 @@
+---Models in module 'stock_picking_batch'---
+---Fields in module 'stock_picking_batch'---
+stock_picking_batch / stock.picking            / batch_sequence (integer)      : NEW
```
pedrobaeza, line 3: Isn't affecting this to the previous order?

You have to put for each new field, the detailed explanation of why is nothing to do. Example: the "nothing to do" of this module: https://github.com/OCA/OpenUpgrade/pull/5217/files#diff-c0136719296715e18716a7d7ee7e8368f55abefb85c05dfc4000d8d6e28354d0

carlos-lopez-tecnativa: > Isn't affecting this to the previous order?
> 

Thanks for the observation. I have updated the migration script to set a default value in order to preserve the same order as in the previous version.

pedrobaeza (approved):

PR #5022, opened 2025-05-13
[18.0][OU-ADD] hr_recruitment

BhaveshHeliconia:
Depends on

- https://github.com/OCA/OpenUpgrade/pull/5006

PR #5274, opened 2025-08-04
[18.0][OU-ADD] hr_recruitment : Migration to 18.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/OpenUpgrade/pull/5022

TT54654


openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,109 @@
...
+    Pre-create the minimal hr_candidate table
+    and populate it with data from hr_applicant.
+    Create a temporary column hr_applicant_id to link the two tables.
+    This is done to avoid errors with NOT NULL constraints,
+    since candidate_id is a required field in hr.applicant.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        CREATE TABLE IF NOT EXISTS hr_candidate (
```
MiquelRForgeFlow, line 49: I am not a fan of creating table in pre-migration. There is nothing wrong will filling this table in post-migration.

carlos-lopez-tecnativa: > I am not a fan of creating table in pre-migration. There is nothing wrong will filling this table in post-migration.

I discussed this with @pedrobaeza and explored some options, but the limitation is that this new field in `hr_applicant` is required.
So, I created the table with minimal fields and linked both tables.
https://github.com/odoo/odoo/blob/5191d99026101393e9ab68892bebb9659af58531/addons/hr_recruitment/models/hr_applicant.py#L40

Another alternative, as Pedro suggested, is to create a fictitious record in `hr_candidate` and populate the table in `post-migration`, but for me, this approach is cleaner.
Please let me know if you have any other suggestions.

MiquelRForgeFlow: > this new field in `hr_applicant` is required.

Which one?

pedrobaeza: This one: https://github.com/OCA/OpenUpgrade/pull/5274/files#diff-f6da79626b08347cb81a40992cf28156d22b89345367759fede3b14c1b74dc8bR20

MiquelRForgeFlow: Ah, do you mean that you cannot access post-migration of this module without filling that field?

pedrobaeza: There will be an error in the log of `NOT NULL constraint violation` (and maybe a crash, not sure).

MiquelRForgeFlow: Okay. I prefer then to fill it with `0` in pre-migration and later the creation and fill in post-migration.

pedrobaeza: If I don't remember bad, putting 0 will give a FK constraint, isn't it?

MiquelRForgeFlow: I think we did in another module something like this. I will try to find it.

MiquelRForgeFlow:
BTW, I will do a supersede, as I have more to add here.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/pre-migration.py
MiquelRForgeFlow: Selection_4863

It is not a problem to load data without the candidate_id filled.

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/5275.

PR #5275, opened 2025-08-05
[18.0][OU-ADD] hr_recruitment

MiquelRForgeFlow:
Supersedes https://github.com/OCA/OpenUpgrade/pull/5274.


openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
```diff
@@ -0,0 +1,172 @@
...
+            user_id, company_id, old_applicant_id,
+            create_date, write_date, create_uid, write_uid
+        )
+        SELECT partner_id, partner_name, email_from, email_cc, email_normalized,
+               COALESCE(partner_mobile, partner_phone),
+               COALESCE(partner_mobile_sanitized, partner_phone_sanitized),
+               phone_sanitized, emp_id, linkedin_profile, type_id, priority,
+               availability, color, message_bounce, TRUE,
+               user_id, company_id, id,
+               NOW(), NOW(), create_uid, write_uid
```
pedrobaeza: Why not putting the hr_applicant dates?

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
```diff
@@ -0,0 +1,172 @@
...
+        WHERE ce.applicant_id = ha.id
+        """,
+    )
+    # unify candidates
+    env.cr.execute(
+        """
+        SELECT array_agg(id order by create_date desc), partner_id
+        FROM hr_candidate
+        GROUP BY partner_id
+        HAVING COUNT(id) > 1
```
pedrobaeza: Why doing this instead of having the `INSERT` with a select with group by and `MAX`/ `MIN` for auto-selecting proper values from the duplicates?

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
```diff
@@ -0,0 +1,172 @@
...
+    recruitment_manager.implied_ids = [
+        Command.unlink(group_hr_user.id),
+        Command.unlink(group_mail_template.id),
+    ]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "hr_recruitment", "18.0.1.1/noupdate_changes_work.xml")
+    openupgrade.load_data(env, "hr_recruitment", "18.0.1.1/noupdate_changes.xml")
```
pedrobaeza: You should remove this second one.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
carlos-lopez-tecnativa: I think you must filter only if the partner is set; otherwise, you will merge records with an empty partner, and this is incorrect.

```suggestion
        SELECT array_agg(id order by create_date desc), partner_id
        FROM hr_candidate
        WHERE partner_id IS NOT NULL
        GROUP BY partner_id
        HAVING COUNT(id) > 1
```

carlos-lopez-tecnativa (changes requested):

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
MiquelRForgeFlow: They are complementary. Do you mean, copying what is inside to the other file?

pedrobaeza: Yes, we agreed on that when modifying the original file, we do a copy, not to have 2 files.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
MiquelRForgeFlow: And then I would have to do a lot of COALESCE in the values and such. And the helper is also implicated. I preferred to avoid over-complicate things. My mind considers simply by link correctly everything and then merging records if needed.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,46 @@
...
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, columns_copy)
+    openupgrade.add_columns(env, column_creates)
+    openupgrade.rename_fields(env, field_renames)
+    openupgrade.rename_xmlids(env.cr, xmlid_renames)
+    openupgrade.rename_xmlids(
+        env.cr,
+        [
+            ("refuse_reason_3", "refuse_reason_2"),
+            ("refuse_reason_4", "refuse_reason_2"),
```
carlos-lopez-tecnativa: `Cannot rename XMLID refuse_reason_3 to refuse_reason_2: need the module reference to be specified in the IDs`

```suggestion
            ("hr_recruitment.refuse_reason_3", "hr_recruitment.refuse_reason_2"),
            ("hr_recruitment.refuse_reason_4", "hr_recruitment.refuse_reason_2"),
```

carlos-lopez-tecnativa (changes requested):

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
pedrobaeza: No, as said, in this case they are aggregation functions, like `MAX`, `SUM`, etc. It's simpler, and more efficient. Please consider cases where there are thousands of applicants.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/pre-migration.py
MiquelRForgeFlow: Thanks.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
MiquelRForgeFlow: Please, do the explicit proposal of the insert.

pedrobaeza: I have already told you: `GROUP BY` and `MAX`. You were the one superseding our PR :wink:

MiquelRForgeFlow: Haha Feel free to supersede. I don't see clearly an easy way to do it without bumping into some issues.

pedrobaeza: Which issues? Please explain them and we figure out a solution.

MiquelRForgeFlow: If I understood correctly, you want to do something like:

```python
INSERT INTO hr_candidate (
            partner_id, partner_name, email_from, email_cc, email_normalized,
            partner_phone, partner_phone_sanitized, phone_sanitized,
            employee_id, linkedin_profile, type_id, priority,
            availability, color, message_bounce, active,
            user_id, company_id, old_applicant_id,
            create_date, write_date, create_uid, write_uid
        )
        SELECT partner_id, COALESCE(partner_name), COALESCE(email_from), 
              (...), company_id, max(id),
               min(create_date), max(write_date), COALESCE(create_uid), COALESCE(write_uid)
        FROM hr_applicant
        WHERE candidate_id IS NULL
        GROUP BY partner_id, company_id;
```

Which is wrong:
- Maybe partner_id IS NULL.
- Here you are not controlling which `old_applicant_id` you get. That's why I did `SELECT array_agg(id order by create_date desc)`.
- The lot of COALESCEs/min/max will look awful.

pedrobaeza: - Are `hr.candidate` created if no `partner_id`?
- You don't want to control that, because you already have the proper link through `partner_id` and `company_id` AFAIK.
- Which is the problem with `COALESCE`? For me it's perfect.

MiquelRForgeFlow: > - You don't want to control that, because you already have the proper link through partner_id and company_id AFAIK.

Suppose you have one applicant with linkedin_profile A, and another applicant with same partner and company with linkedin_profile B. You are not controling if you will get linkedin_profile A or B.

MiquelRForgeFlow: Also, this kind of suggestion, is a blocking requirement? I would prefer to avoid it.

pedrobaeza: > Suppose you have one applicant with linkedin_profile A, and another applicant with same partner and company with linkedin_profile B. You are not controling if you will get linkedin_profile A or B.

You have the same kind of control using aggregators in the SQL than later doing the merge. At the end, only one is kept. Which? The one you decide in the SQL or in the merge operation.

> Also, this kind of suggestion, is a blocking requirement? I would prefer to avoid it.

Yes, it's blocking because the performance is in play.

MiquelRForgeFlow: > - Are `hr.candidate` created if no `partner_id`?

It seems that yes, it is possible, as `partner_id` is not a required field in v17 nor in v18.

> Yes, it's blocking because the performance is in play.

I have to say it's kind of sad. The code now works, and a nitpicking shouldn't block it. I mean, "performance" in this case should not be an issue, as I think in any company there should be **very few** cases of more than one applicant having same partner_id per company.

pedrobaeza: > It seems that yes, it is possible, as partner_id is not a required field in v17 nor in v18.

Then what is the criteria in the code for "grouping" the applicants around candidates when there don't have partner? If no grouping, have the same query for both, just varying the group by (`partner_id` if partner_id IS NOT NULL, and the `id` if partner_id IS NULL). It's simple IMO.

MiquelRForgeFlow: Ok, I tried this "elegant" query:

```sql
    WITH sub as (
        SELECT partner_id, partner_name,
               email_from, email_cc, email_normalized,
               COALESCE(partner_mobile, partner_phone) AS partner_phone,
               COALESCE(partner_mobile_sanitized, partner_phone_sanitized) AS partner_phone_sanitized,
               phone_sanitized, emp_id, linkedin_profile, type_id, priority,
               availability, color, message_bounce, TRUE,
               user_id, company_id, id,
               create_date, write_date, create_uid, write_uid
        FROM hr_applicant
        WHERE candidate_id IS NULL AND partner_id IS NOT NULL
        ORDER BY partner_id, company_id, create_date DESC
    )
    INSERT INTO hr_candidate (
        partner_id, partner_name, email_from, email_cc, email_normalized,
        partner_phone, partner_phone_sanitized, phone_sanitized,
        employee_id, linkedin_profile, type_id, priority,
        availability, color, message_bounce, active,
        user_id, company_id, old_applicant_id,
        create_date, write_date, create_uid, write_uid
    )
    SELECT partner_id, COALESCE(partner_name),
           COALESCE(email_from), COALESCE(email_cc), COALESCE(email_normalized),
           COALESCE(partner_phone), COALESCE(partner_phone_sanitized),
           COALESCE(phone_sanitized), COALESCE(emp_id), COALESCE(linkedin_profile),
           COALESCE(type_id), COALESCE(priority),
           MAX(availability), COALESCE(color), COALESCE(message_bounce), TRUE,
           COALESCE(user_id), company_id, COALESCE(id),
           MIN(create_date), MAX(write_date), COALESCE(create_uid), COALESCE(write_uid)
    FROM sub
    GROUP BY partner_id, company_id
```

But this doesn't work.

> - Which is the problem with COALESCE? For me it's perfect.

>> Suppose you have one applicant with linkedin_profile A, and another applicant with same partner and company with linkedin_profile B. You are not controling if you will get linkedin_profile A or B.
>
> You have the same kind of control using aggregators in the SQL than later doing the merge. At the end, only one is kept. Which? The one you decide in the SQL or in the merge operation.


COALESCE(linkedin_profile) without an aggregation function just returns a random linkedin_profile per group. It will get an arbitrary one, not necessarily the newest filled. Same for COALESCE(id) and others. COALESCE() is not deterministic. The ORDER BY inside the subquery is ignored. In a subquery or CTE, ORDER BY has no effect unless you use LIMIT, OFFSET, or are inside a window function or DISTINCT ON. So, that “elegant” query is syntactically valid, and it may appear to work, but it's not semantically correct for the expected purpose, and the results will be unreliable.

Finally, I found a solution using windows function and FIRST_VALUE.

MiquelRForgeFlow:
Ready.

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
```diff
@@ -0,0 +1,239 @@
...
+    # update attachments
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE ir_attachment ia
+        SET res_model = 'hr.candidate', res_id = ha.candidate_id
+        FROM hr_applicant ha
+        WHERE ia.res_model = 'hr.applicant' AND ia.res_id = ha.id
+        """,
+    )
```
carlos-lopez-tecnativa: I'm not sure if it's correct to change the model of the attachments.

In v17, the dashboard points to `hr.applicant`:
image

And you can see the attachments there:
image

In v18, the same dashboard still points to the same model, so I would expect to see the same data:
image

But now I don't see the attachment, so the principle of preserving previous behavior is not being applied here.
Remember that `hr.applicant` has not been removed, so I believe the attachment model should not be changed.

Or, if you want to display the attachment in `hr.candidate`, perhaps it should be duplicated.
image

Also, if you use the `website_hr_recruitment` module in v18 and upload an attachment, you'll see that it's still associated with `hr.applicant`, not with `hr.candidate`.
So, what was the reasoning behind changing the attachment model?
image

carlos-lopez-tecnativa (changes requested):

openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/post-migration.py
MiquelRForgeFlow: In the commit that created the `hr.candidate` model (https://github.com/odoo/odoo/commit/75b385bae036031b8fef651977bad71691b1e7a2), they moved the attachments to the candidate in their demo data. That's why I moved them too.

Selection_4868

MiquelRForgeFlow: What do you think @pedrobaeza about this?

MiquelRForgeFlow: At the end, a duplication has been applied to attachments.

MiquelRForgeFlow:
All the reviews have been attended.

PR #5282, opened 2025-08-05
[18.0][OU-FIX] l10n_fr: some XML-ids where not properly moved

remi-filament:
There were too many XML-ids to rename and some which were not listed in upgrade_analysis, so I took a more aggressive approach for renaming all XML ids from models which were moved from l10n_fr to l10n_fr_account.

For instance, id="box_26_external" which is in nested line_ids / child_ids : 
```xml
    <record id="tax_report" model="account.report">
       ...
       <field name="line_ids">
            ...
            <record id="tax_report_determination" model="account.report.line">
                <field name="name">Determining the amount to be paid and/or VAT and/or TIC credits</field>
                <field name="hierarchy_level">0</field>
                <field name="children_ids">
                    <record id="tax_report_26_external" model="account.report.line">
                        <field name="name">26 - Repayment of credit requested on form n°3519 attached</field>
                        <field name="code">box_26_external</field>
```
is not reported in https://github.com/OCA/OpenUpgrade/blob/18.0/openupgrade_scripts/scripts/l10n_fr/18.0.2.1/upgrade_analysis.txt when it was moved from https://github.com/odoo/odoo/blob/fa1d2e4982683062d5b8913b10e20a893838a5af/addons/l10n_fr/data/tax_report_data.xml#L1950 to https://github.com/odoo/odoo/blob/18.0/addons/l10n_fr_account/data/tax_report_data.xml#L1951

Maybe because analysis tool does not handle this many nested records ?


legalsylvain (approved):

pedrobaeza (approved):

openupgrade_scripts/scripts/l10n_fr/18.0.2.1/pre-migration.py
```diff
@@ -3,682 +3,21 @@
...
-    openupgrade.logged_query(  # to update 2 problematic tax accounts in l10n_fr_account
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE ir_model_data
+        SET module = 'l10n_fr_account'
+        WHERE module = 'l10n_fr' AND model IN (
+            'account.account.tag', 'account.report', 'account.report.column',
+            'account.report.expression', 'account.report.line',
+            'ir.config_parameter', 'res_bank')
```
MiquelRForgeFlow, line 17: res.bank

remi-filament: Thanks for spotting that, I just created another PR for this typo  : https://github.com/OCA/OpenUpgrade/pull/5290

PR #5283, opened 2025-08-06
[17.0][OU-FIX] payment: assign payment_method_ids on duplicated providers

remi-filament:
Follow-up of https://github.com/OCA/OpenUpgrade/pull/4692#issuecomment-3156119755
When payment providers are duplicated (for being used on another company most of the time), the previous script implementation did not set method_ids on these ones (since it did it only on payment providers with an External ID so created at module installation).

This PR fixes that behaviour by setting payment_method_ids on payment providers based on (in the following order) :
1. external ID
2. code
3. module_id (this one should not be needed since code is required)

Also, I realize that payment_method_ids were not updated for 4 modules (payment_asiapay, payment_ogone, payment_payulatam and payment_payumoney) (since these are not even present in openupgrade_scripts / scripts directory) so I handled those also here.

I also changed readme to set as done all the payment modules which had no other changes.

Eventually, I also took the opportunity to update payment_custom and payment_demo migration scripts (I could not handled these ones in payment directly since only when installed they install their own payment_method).


openupgrade_scripts/scripts/payment/17.0.2.0/post-migration.py
```diff
@@ -4,6 +4,537 @@
...
+from odoo import Command
+
+_PROVIDER_CODE_MAPPING = {
+    "adyen": "payment.payment_provider_adyen",
+    "alipay": "payment_alipay.payment_provider_alipay",
+    "aps": "payment.payment_provider_aps",
+    "asiapay": "payment.payment_provider_asiapay",
+    "authorize": "payment.payment_provider_authorize",
+    "buckaroo": "payment.payment_provider_buckaroo",
+    # This one should be handed in payment_transfer OpenUpgrade script
```
MiquelRForgeFlow:
```suggestion
    # This one should be handed in payment_custom OpenUpgrade script
```

MiquelRForgeFlow (commented):
I have no idea but LGTM 😄

openupgrade_scripts/scripts/payment/17.0.2.0/post-migration.py
remi-filament: Thanks, good catch !

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #5285, opened 2025-08-07
[18.0][OU-FIX] mrp: Force attached_on_mrp on db insert

HviorForgeFlow:
When applying an DB INSERT INTO, default values are not considered.
Not adding attached_on_mrp end up failing.
As mention in https://github.com/OCA/OpenUpgrade/blob/18.0/openupgrade_scripts/scripts/mrp/18.0.2.0/upgrade_analysis_work.txt#L65C3-L65C132 defaulting to 'hidden'

CC @ForgeFlow @MiquelRForgeFlow @hbrunn


MiquelRForgeFlow (approved):
Makes sense.

PR #5286, opened 2025-08-07
[18.0][OU-FIX] account: fill hard_lock_date

MiquelRForgeFlow:
...

### DIFF
--- a/openupgrade_scripts/scripts/account/18.0.1.3/post-migration.py
+++ b/openupgrade_scripts/scripts/account/18.0.1.3/post-migration.py
@@ -4,7 +4,7 @@
 from openupgradelib import openupgrade, openupgrade_180
 
 
-def replace_period_lock_date(env):
+def handle_lock_dates(env):
     openupgrade.logged_query(
         env.cr,
         """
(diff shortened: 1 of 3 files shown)


openupgrade_scripts/scripts/account/18.0.1.3/post-migration.py
```diff
@@ -4,14 +4,21 @@
...
-def replace_period_lock_date(env):
+def handle_lock_dates(env):
     openupgrade.logged_query(
         env.cr,
         """
         UPDATE res_company
         SET sale_lock_date = period_lock_date, purchase_lock_date = period_lock_date
         WHERE period_lock_date IS NOT NULL""",
     )
+    openupgrade.logged_query(
```
MiquelRForgeFlow: In fact, I should add the condition if `account_lock` was installed or not. How can we check it?

pedrobaeza: Yes, this can't be unconditionally. Check for any trace of that module.

MiquelRForgeFlow: It's difficult. No fields, no xmlids.

MiquelRForgeFlow: @StefanRijnhart @hbrunn any magical idea?

hbrunn: nothing better than setting a flag in base's pre-migration ie in ir.config_parameter, and reading and deleting it here.

or maybe more general: do `select name, state from ir_module_module into {get_legacy_name('ir_module_module')}` at the very beginning of base's pre-migration, then we've a general way of inspecting module state before migration?

MiquelRForgeFlow: Thanks

PR #5302, opened 2025-08-13
[18.0][OU-ADD] hr_recruitment_skills : Migration to 18.0

carlos-lopez-tecnativa:
TT57486


openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+model_renames = [("hr.applicant.skill", "hr.candidate.skill")]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.table_exists(env.cr, "hr_applicant_skill"):
+        openupgrade.rename_models(env.cr, model_renames)
```
MiquelRForgeFlow: if you don't rename tables, it doesn't make sense to rename the models

openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,39 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+def _fill_hr_candidate_skill(env):
+    # Create candidate skill records from applicant skill
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO hr_candidate_skill (
```
MiquelRForgeFlow, line 11: you should try to avoid duplicating skills for each candidate in case a candidate has more than one applicant.

carlos-lopez-tecnativa: I don’t understand your point. I’m getting the data from `hr_applicant_skill` and joining it with `hr_applicant`, not with `hr_candidate`, so `hr_applicant` shouldn’t have duplicates because it only has a `many2one` relation with `hr_candidate`. Could you please provide an example?

openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,48 @@
+---Models in module 'hr_recruitment_skills'---
+obsolete model hr.applicant.skill (renamed to hr.candidate.skill)
```
MiquelRForgeFlow, line 2: call openupgrade.remove_tables_fks for this table and `hr_applicant_hr_skill_rel`

openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/post-migration.py
MiquelRForgeFlow: In `hr_recruitment` migration, we didn't create a candidate for each applicant (we grouped by partner and company), so it may happen that some candidates have more than one applicant. Suppose that for each applicant you had several hr_applicant_skill. If for two applicants than have same candidate they have both also an applicant skill that point to the same skill, then you don't want to have a candidate with two candidate_skills than point the same skill (in the same or different level).

carlos-lopez-tecnativa: Okay, now it’s clearer. But which skill should I keep? What should the criteria be?
I think we should preserve the one with the highest level, what do you think?

openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Done

openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/post-migration.py
carlos-lopez-tecnativa: I updated the code to consider this use case.

In V17 
Applicant 1
image

Applicant 2
image

In V18, The unique candidate has the same skills with a higher level.
image

Please let me know if this is okay for you.?

MiquelRForgeFlow: Yes, group by skill and get the highest level.

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/hr_recruitment_skills/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        INSERT INTO hr_candidate_hr_skill_rel (
+            hr_candidate_id, hr_skill_id
+        )
+        SELECT ha.candidate_id, m2m.hr_skill_id
+        FROM hr_applicant_hr_skill_rel as m2m
+            JOIN hr_applicant AS ha ON ha.id = m2m.hr_applicant_id
+        ON CONFLICT DO NOTHING
+        """,
+    )
```
MiquelRForgeFlow:
```suggestion
    # fill many2many field skill_ids
    env["hr.candidate"].search([])._compute_skill_ids()
```

carlos-lopez-tecnativa: Why? What is the reason for using the ORM here, if SQL has better performance and the final result is the same?

MiquelRForgeFlow: Because your query is adding all skills of all applicant_skills of all applicants. And this is wrong. When you filled the `hr_candidate_skill`, you filtered `hr_applicant_skill`. So, now, to fill `hr_candidate_hr_skill_rel`, you have the maintain the same filter. Of course, if you do by query, it would be better but then you have to modify the query. I just thought this compute is not that bad, but as you prefer.

MiquelRForgeFlow: Or maybe I am confused, and your query is fine.

carlos-lopez-tecnativa: The query is working, and the `ON CONFLICT DO NOTHING` clause handles any possible duplicates.
However, I’ve updated the query to follow the same logic as the compute method, using the previously filled `hr_candidate_skill` records.
Please let me know if it looks good to you now.

MiquelRForgeFlow (approved):

PR #5104, opened 2025-06-13
[18.0][OU-ADD] point_of_sale

MiquelRForgeFlow:
.

PR #5305, opened 2025-08-13
[18.0][OU-FIX] point_of_sale: Prevent error when sale module is not installed

carlos-lopez-tecnativa:
This case is already handled in the migration script of the sale module. See: https://github.com/OCA/OpenUpgrade/blob/8ff8356dda92caefb532c5a37f718dada0682c5d/openupgrade_scripts/scripts/sale/18.0.1.2/pre-migration.py#L21-L31

Complementary to https://github.com/OCA/OpenUpgrade/pull/5104

image


pedrobaeza (commented):
Is there any entry for this in upgrade_analysis_work.txt? Why it was done this way?

MiquelRForgeFlow:
It's from the `sale` module. My bad.

MiquelRForgeFlow (approved):

PR #5306, opened 2025-08-13
[15.0][IMP] account: handle better website_legal_page

MiquelRForgeFlow:
Better approach of https://github.com/OCA/OpenUpgrade/pull/4536.

### DIFF
--- a/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/15.0.1.2/post-migration.py
@@ -45,6 +45,19 @@ def _handle_website_legal_page(env):
     the information added in the page created by 'website_legal_page' has to be
     transferred to the invoice_terms_html field of the company.
     """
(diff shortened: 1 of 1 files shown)


christian-ramos-tecnativa:
Hi @MiquelRForgeFlow What’s the benefit of merging the module and then reinstalling it? The function uses the key `website_legal_page.legal_page`, so the merge itself doesn’t seem to matter, unless I’m missing something. I assume the merge is mainly to “uninstall” the module.
As it stands, it assigns the legal info to the company and then performs the regular update. I’m not sure why your option is better. From my perspective, the main question was whether we keep the module installed or not. In its current form, it keeps it installed with the same result you’re proposing.
Are we not just adding more steps here?

MiquelRForgeFlow:
@Christian-RB sorry, you are right.

christian-ramos-tecnativa:
@MiquelRForgeFlow no worries!

PR #5307, opened 2025-08-13
[18.0][OU-ADD] pos_hr : Migration to 18.0

carlos-lopez-tecnativa:
TT57491


openupgrade_scripts/scripts/pos_hr/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,25 @@
+---Models in module 'pos_hr'---
+new model report.pos_hr.multi_employee_sales_report [abstract]
+new model report.pos_hr.single_employee_sales_report [abstract]
+# NOTHING TO DO: new feature added in https://github.com/odoo/odoo/commit/5193e458ac0dc165757fa0f3fc17796376ad1b3b
+
+---Fields in module 'pos_hr'---
+pos_hr       / account.bank.statement.line / employee_id (many2one)        : NEW relation: hr.employee
+# DONE: post-migration: Fill this field based on the employee related to the user who created the record.
```
pedrobaeza, line 8: There can be multiple cashiers per user. Isn't there the possibility to obtain it from the `pos.order`?

carlos-lopez-tecnativa: The same user cannot have more than one employee in the same company, so this restriction prevents a potential issue.

image

This case happens only when creating a `Cash In/Out`. In that case, no `pos.order` or `pos.payment` is generated, the only record created is a statement line. However, I haven’t found another reliable way to fill this field.

One possible approach is to use the session’s tracking, but I’ve seen some transactions with the same value, so it’s not a reliable method.
What other solution would you suggest?
image

pedrobaeza (approved):

openupgrade_scripts/scripts/pos_hr/18.0.1.0/upgrade_analysis_work.txt
pedrobaeza: OK then.

PR #5310, opened 2025-08-14
[18.0][OU-ADD] fleet: Migration to 18.0

jguenat:


openupgrade_scripts/scripts/fleet/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,63 @@
...
+DEL fleet.vehicle.model: fleet.model_seriex
+DEL fleet.vehicle.model: fleet.model_seriez4
+DEL fleet.vehicle.model: fleet.model_tt
+DEL fleet.vehicle.model: fleet.model_zafira
+DEL fleet.vehicle.model: fleet.model_zafiratourer
+NEW ir.actions.server: fleet.action_fleet_vehicle_send_mail
+NEW ir.model.access: fleet.access_fleet_vehicle_send_mail
+NEW ir.ui.view: fleet.fleet_vehicle_send_mail_view_form
+NEW ir.ui.view: fleet.fleet_vehicle_state_view_form
+# DONE fleet model deleted safely
```
carlos-lopez-tecnativa: I think the `Done` comment should apply only to line 58. The following lines are new features.

openupgrade_scripts/scripts/fleet/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,63 @@
+---Models in module 'fleet'---
+new model fleet.vehicle.send.mail [transient]
+# NOTHING TO DO
+---Fields in module 'fleet'---
+fleet        / fleet.vehicle            / power_unit (selection)        : NEW required, selection_keys: ['horsepower', 'power'], hasdefault: default
+fleet        / fleet.vehicle            / vehicle_range (integer)       : NEW
+fleet        / fleet.vehicle.model      / activity_ids (one2many)       : NEW relation: mail.activity
+fleet        / fleet.vehicle.model      / message_follower_ids (one2many): NEW relation: mail.followers
+fleet        / fleet.vehicle.model      / message_ids (one2many)        : NEW relation: mail.message
+fleet        / fleet.vehicle.model      / power_unit (selection)        : NEW required, selection_keys: ['horsepower', 'power'], hasdefault: default
```
carlos-lopez-tecnativa, line 10: Please add a note to indicate that the default value is consistent with the existing data.

carlos-lopez-tecnativa (changes requested):
LGTM, just a minor change for a more detailed explanation of the work done.

openupgrade_scripts/scripts/fleet/18.0.0.1/upgrade_analysis_work.txt
jguenat: In fact after checking the view the fields power and horsepower are hidden depending on the new field power_unit chosen.
In version 17.0 it was possible to fill both in the view, now it's not possible.
By defining power_unit default to power, it will hide horsepower values.
I made a migration script but it's opiniated because in version 17.0 it was possible to have both horsepower and power displayed at the same time and it's not possible anymore.

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def _set_power_unit(env):
+    # if horspower is defined, set power_unit to horsepower
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE fleet_vehicle_model
+        SET power_unit='horsepower'
+        WHERE horsepower IS NOT NULL
```
carlos-lopez-tecnativa: Since the field is numeric, it does not contain NULL values. With your code, the column `power_unit` is set to `horsepower` in all rows, including rows where it should not apply (see the image).
```suggestion
        WHERE horsepower > 0
```
image

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE fleet_vehicle
+        SET power_unit='horsepower'
+        WHERE horsepower IS NOT NULL
```
carlos-lopez-tecnativa:
```suggestion
        WHERE horsepower > 0
```

carlos-lopez-tecnativa (changes requested):

MiquelRForgeFlow:
\ocabot migration fleet

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
jguenat: It worked when I tested it but you are right in odoo once a value is applied to an integer you cannot go back to NULL in database even if you set it to 0 in Odoo. Integer fields are NULL when they have never been assigned a value so > 0 is indeed better.
odoo_int

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def _set_power_unit(env):
+    # if horspower is defined, set power_unit to horsepower
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE fleet_vehicle_model
+        SET power_unit='horsepower'
+        WHERE horsepower > 0
```
carlos-lopez-tecnativa:
```suggestion
        WHERE COALESCE(horsepower, 0) > 0
```

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE fleet_vehicle
+        SET power_unit='horsepower'
+        WHERE horsepower > 0
```
carlos-lopez-tecnativa:
```suggestion
        WHERE COALESCE(horsepower, 0) > 0
```

carlos-lopez-tecnativa (approved):
If you want to be sure it also covers cases where the value is NULL or 0, you could use `COALESCE`, but LGTM.

TT54429

@pedrobaeza @MiquelRForgeFlow could you please review this?

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    _set_power_unit(env)
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
```
MiquelRForgeFlow:
```suggestion
    data = env["ir.model.data"].search(
                [("module", "=", "fleet"), ("name", "in", 
```

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+        [
+            "fleet.model_a1",
```
MiquelRForgeFlow:
```suggestion
            "model_a1",
```
for all

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,79 @@
...
+            "fleet.model_serie6",
+            "fleet.model_serie7",
+            "fleet.model_seriehybrid",
+            "fleet.model_seriem",
+            "fleet.model_seriex",
+            "fleet.model_seriez4",
+            "fleet.model_tt",
+            "fleet.model_zafira",
+            "fleet.model_zafiratourer",
+        ],
```
MiquelRForgeFlow: `data.unlink()`

MiquelRForgeFlow (commented):
I think it's better to save the fleet models than to delete them.

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
jguenat: Yes you are right kind of strange to loose models with an upgrade even if not used.

openupgrade_scripts/scripts/fleet/18.0.0.1/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,64 @@
...
+fleet        / fleet.vehicle.model      / vehicle_range (integer)       : NEW
+# DONE set power_unit to horsepower default is power) if horsepower is > 0. Other fields are new with no consequence on existing data
+---XML records in module 'fleet'---
```
pedrobaeza, line 15: You have to separate lines for better reading, and the syntax is `DONE: ` + `pre-migration/post-migration/end-migration` + `: `.

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
```diff
@@ -0,0 +1,86 @@
...
+                    "model_seriex",
+                    "model_seriez4",
+                    "model_tt",
+                    "model_zafira",
+                    "model_zafiratourer",
+                ],
+            ),
+        ]
+    )
+    data.unlink()
```
pedrobaeza, line 86: Why don't you use `delete_records_safely_by_xml_id` that tries to remove data if not used, or the XML-ID if not able to remove?

pedrobaeza (changes requested):

openupgrade_scripts/scripts/fleet/18.0.0.1/upgrade_analysis_work.txt
jguenat: will do :)

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
jguenat: I did it like that first but then @MiquelRForgeFlow thought that it was better to keep the models even if not used. I tend to agree it should be the decision of the user not the migration script to remove or keep models.

jguenat: Also side technical question : delete_records_safely_by_xml_id does not remove the XML-ID but change it to no-update if unable to unlink the record. Why not delete the XML-ID ?

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
MiquelRForgeFlow:
```suggestion
        WHERE horsepower > 0 AND COALESCE(power,0) = 0
```

I think it's better to prioritize `power` in case both fields are filled.
Selection_4892

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/fleet/18.0.0.1/post-migration.py
MiquelRForgeFlow: > it was better to keep the models

In fact, Odoo didn't delete them. They moved them to demo data.

pedrobaeza: Yeah, but in a practical side, they are not there on non demo databases, so trying to removing them is the logical way to handle it. Users may not remove them because they don't know they can.

MiquelRForgeFlow: > they are not there on non demo databases

Well, in v17, they were as non-demo data. So clients coming from v17 with `fleet` installed, will have this data of vehicle models. And they could be used, so it's not practical their removal.

pedrobaeza: If they are used, they can't be removed.

MiquelRForgeFlow: Suppose that from those vehicle models, you used 3. Thus, following you rationale, the other vehicle models would be deleted except those 3. It doesn't make much sense. Or you have them all, or you remove them all.

pedrobaeza: Well, I see it as the correct behavior

MiquelRForgeFlow: That's a personal point of view, not necessary the "correct" behavior, right? In this case, I believe there doesn't exist a "true correct one" behavior, and thus, I believe we should respect the option the author of this PR chooses.

> Also side technical question : delete_records_safely_by_xml_id does not remove the XML-ID but change it to no-update if unable to unlink the record. Why not delete the XML-ID ?

BTW, the use of `delete_records_safely_by_xml_id` in this case would be incorrect, as it doesn't delete the model data:
Selection_4902

🤔 Maybe we should create another method where the exception is:

`imd.unlink()` instead of `imd.noupdate = true`.

jguenat: if we look it in a positive way meaning user are well enough savy with Odoo to archive/delete model they don't use then they can do the cleaning themself if they want/need to.
Also there is a menu entry (not in technical) to open models (tree, form, search view and all) so we might imagine someone is using this natively without linking models to vehicles but just do do statistics or something. We can also thing about customizations that might use models do do other things that might lead to bad outcome if we "safely" delete them.
I know this is far stretched and it was certainly a mistake made by odoo to have these models in module data and not in demo data.
My opinion is to keep them :)

pedrobaeza: OK, I think the usage of this module is niche, so it won't affect too much. Let's continue!

pedrobaeza (approved):

PR #5312, opened 2025-08-14
[16.0][OU-FIX] purchase_requisition: Proper order group creation (and global) + no delete

pedrobaeza:
The PO groups weren't created properly before, so the new query is doing the correct job, and doing it for all the requisitions, not only those of one type.

The previous code was also deleting all the requisitions of a deleted type, but doing that, apart from being partial, it's counter-productive, requiring a lot of work with no benefit, so let's keep them.

@Tecnativa


pedrobaeza:
Tested and working.

PR #5314, opened 2025-08-15
[17.0][OU-FIX] l10n_de_mis_*_report: Rename XML-IDs pre-merge

pedrobaeza:
Both l10n_de_mis_skr03_report and l10n_de_mis_skr04_report have the same XML-IDs for their data records, so merging them into l10n_de_mis_report generates duplicated XML-IDs if both are installed, crashing the system.

But only if you have one of them installed, all the previous MIS definition records will be tried to be removed, giving the corresponding error if they have been used.

Let's fix it doing a previous renaming of all its XML-IDs.

@Tecnativa TT41859


pedrobaeza:
Tested and working.

PR #5347, opened 2025-08-25
[18.0][OU-FIX] Remove spreadsheet_dashboard_purchase* as they were moved to EE

etobella:
https://github.com/odoo/odoo/pull/179875

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -60,6 +60,8 @@ merged_modules = {
     "sale_product_configurator": "sale",
     "stock_landed_costs_company": "stock_landed_costs",
     "website_sale_product_configurator": "website_sale",
+    "spreadsheet_dashboard_purchase": "spreadsheet_dashboard",
+    "spreadsheet_dashboard_purchase_stock": "spreadsheet_dashboard",
     # odoo/enterprise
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
This is not correct. You are losing such features, so they shouldn't be merged seamlessly to the parent module. What we should do is to migrate the existing 17 modules to OCA instead.

legalsylvain:
hi @etobella. in the PR you mention, I don't see the removal done by Odoo. Did you put wrong number ? Or did I missed something ?

etobella:
Sorry, you are right: https://github.com/odoo/odoo/pull/179875
I used the wrong PR :cry:

pedrobaeza:
@etobella @legalsylvain this is what I mean: https://github.com/OCA/spreadsheet/pull/67

pedrobaeza:
The OpenUpgrade PR will be to rename `spreadsheet_dashboard_purchase` to `spreadsheet_dashboard_purchase_oca` (with tag `[OU-ADD]` by the way, not `[OU-FIX]` :wink:).

MiquelRForgeFlow:
Superseded by https://github.com/OCA/OpenUpgrade/pull/5360.

PR #5349, opened 2025-08-26
[18.0][OU-ADD] website_sale_collect: Migration scripts. Renamed from website_sale_picking

carlos-lopez-tecnativa:
TT57380


openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,43 @@
...
+
+website_sale_collect / stock.warehouse          / opening_hours (many2one)      : NEW relation: resource.calendar
+# NOTHING TO DO: New feature, the location selector only displays the opening hours info if it is set.
+# added in https://github.com/odoo/odoo/commit/087c48c4ed2eff55361a2b9fa925800dbdc46fab
+
+website_sale_picking / delivery.carrier         / delivery_type (False)         : selection_keys added: [gelato, in_store], removed: [onsite]
+website_sale_picking / payment.provider         / custom_mode (False)           : selection_keys added: [on_site], removed: [onsite]
+# DONE: post-migration: map values
+# delivery.carrier: onsite > in_store.
+# payment.provide: onsite > on_site
```
pedrobaeza:
```suggestion
# DONE: pre-migration: copy columns for preserving values.
# DONE: post-migration: map values: delivery.carrier: onsite > in_store. payment.provider: onsite > on_site
```

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,43 @@
...
+DEL delivery.carrier: website_sale_picking.default_onsite_carrier (noupdate)
+NEW payment.provider: website_sale_collect.payment_provider_on_site
+DEL payment.provider: website_sale_picking.payment_provider_onsite
+NEW product.product: website_sale_collect.product_pick_up_in_store (noupdate)
+DEL product.product: website_sale_picking.onsite_delivery_product (noupdate)
+NEW ir.ui.view: website_sale_collect.delivery_carrier_form
+DEL ir.ui.view: website_sale_picking.view_delivery_carrier_form_with_onsite_picking
+NEW ir.ui.view: website_sale_collect.res_config_settings_form
+DEL ir.ui.view: website_sale_picking.res_config_settings_view_form
+# DONE: pre-migration: rename XML-IDs for this data according this commit https://github.com/odoo/odoo/commit/087c48c4ed2eff55361a2b9fa925800dbdc46fab
```
pedrobaeza, line 36: Have you taking into account possible fields that were filled in previous version and in this one they are missing (so preserving old value)?

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,48 @@
...
+    openupgrade.m2o_to_x2m(
+        env.cr,
+        env["delivery.carrier"],
+        "delivery_carrier",
+        "warehouse_ids",
+        "warehouse_id",
+    )
+    _delivery_carrier_map_delivery_type(env)
+    _payment_provider_map_custom_mode(env)
+    openupgrade.delete_record_translations(
```
pedrobaeza, line 68: Why this?

pedrobaeza (commented):
Have you checked if there's anything of noupdate data due to the module renaming? Not sure if the system handles this case.

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/upgrade_analysis_work.txt
carlos-lopez-tecnativa: These lines don’t have `noupdate="1"` and they are also views. What are you referring to? Sorry, but I don’t completely understand what you mean.

carlos-lopez-tecnativa:
> Have you checked if there's anything of noupdate data due to the module renaming? Not sure if the system handles this case.

Yes, I checked, the records have `noupdate="1"`, and only two records are like that. However, the XML-IDs are different. Please see:
V17 
https://github.com/odoo/odoo/blob/17.0/addons/website_sale_picking/data/website_sale_picking_data.xml

V18 
https://github.com/odoo/odoo/blob/18.0/addons/website_sale_collect/data/delivery_carrier_data.xml
https://github.com/odoo/odoo/blob/18.0/addons/website_sale_collect/data/product_product_data.xml

So I think that, for the analysis, these are completely different records. But since I consider them to be the same records, to prevent Odoo from creating new ones, I call `rename_xmlids` in the `pre-migration` and `delete_record_translations` in the `post-migration`.

Should I manually add these changes into a `noupdate_changes_work.xml` file, or what would be the best way?

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/post-migration.py
carlos-lopez-tecnativa: Do you mean the `map_values` or the translations?

If you are asking about translations, please check my answer here: https://github.com/OCA/OpenUpgrade/pull/5349#issuecomment-3225316274

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/upgrade_analysis_work.txt
carlos-lopez-tecnativa: Done, Thanks

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/upgrade_analysis_work.txt
pedrobaeza: There are four lines in this block that are noupdate.

pedrobaeza: Sorry, I have mixed things as you have answered at the same time.

I mean for example this:

```
<record id="old_id">
  <field name="field_x">VALUE</field>
</record>
```

```
<record id="new_id">
</record>
```

so the field `field_x` remains with `VALUE` while it should be empty.

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/post-migration.py
pedrobaeza: Yes, translations, but then it means there are different values that you need to fill, so yes, noupdate_changes.xml.

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/upgrade_analysis_work.txt
carlos-lopez-tecnativa: I manually generated the `noupdate_changes.xml` file. Could you please let me know if it’s OK, or if I need to adapt something?

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/noupdate_changes.xml
pedrobaeza, line 1: Please name it `noupdate_changes_work.xml` for not being replaced in the future with a hypothetical automatically generated file (if we fix the problem in `upgrade_analysis` module).

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    _delivery_carrier_map_delivery_type(env)
+    _payment_provider_map_custom_mode(env)
+    openupgrade.load_data(env, "website_sale_collect", "18.0.1.0/noupdate_changes.xml")
+    openupgrade.delete_record_translations(
+        env.cr,
+        "website_sale_collect",
+        [
+            "carrier_pick_up_in_store",
+            "payment_provider_on_site",
```
pedrobaeza, line 73: I don't see this one in noupdate_changes.xml

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/noupdate_changes.xml
carlos-lopez-tecnativa: Done

openupgrade_scripts/scripts/website_sale_collect/18.0.1.0/post-migration.py
carlos-lopez-tecnativa: Since the record does not have the `noupdate` attribute, I did not add it to `noupdate_changes_work.xml`.
I renamed the XML-ID to prevent Odoo from creating another `payment.provider`.
I also used `delete_record_translations` because the `name` field is `translatable`.
Or is this step not necessary for `payment.provider`?

V17
https://github.com/odoo/odoo/blob/7c3d34f5ef4d111fbbc89a07e207fe5add4d0275/addons/website_sale_picking/data/payment_provider_data.xml#L4
V18
https://github.com/odoo/odoo/blob/08b2573f2ac2b812a0e7f8ebaf90db431124ac69/addons/website_sale_collect/data/payment_provider_data.xml#L4

carlos-lopez-tecnativa: If I don’t use `delete_record_translations`, the `payment.provider` keeps the old name.
I tested it, so I think it is necessary.

pedrobaeza: That may be because you didn't load translations after the upgrade, but if it's noupdate=0, it shouldn't be a problem.

pedrobaeza (approved):

PR #5355, opened 2025-08-27
[17.0][OU-FIX] account: Use migrate_translations_to_jsonb to migrate translations from the field

victoralmau:
Use migrate_translations_to_jsonb to migrate translations from the field

Related to https://github.com/OCA/OpenUpgrade/commit/3d5ebac00ea9aa9d2126b7988af2c21fbd943d64 

Similar to https://github.com/OCA/OpenUpgrade/blob/4af8572c7f3e313773db558ba382b5ffd19a157c/openupgrade_scripts/scripts/base/16.0.1.3/end-migration.py#L39

@Tecnativa TT57778

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
@@ -1,8 +1,6 @@
 # Copyright 2024 Viindoo Technology Joint Stock Company (Viindoo)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from openupgradelib import openupgrade
-
-from odoo.tools.sql import convert_column_translatable
+from openupgradelib import openupgrade, openupgrade_160
 
 _fields_renames = [
     (
@@ -73,7 +71,12 @@ def _convert_account_tax_description(env):
     openupgrade.rename_columns(
         env.cr, {"account_tax": [("description", "invoice_label")]}
     )
-    convert_column_translatable(env.cr, "account_tax", "invoice_label", "jsonb")
+    # Deliberately call migrate_translations_to_jsonb instead of
+    # convert_column_translatable to support the case you have translate=True on
+    # the field in v16 (l10n_multilang compatibility).
+    openupgrade_160.migrate_translations_to_jsonb(
+        env, [("account_tax", "invoice_label")]
+    )
 
 
 def _am_create_delivery_date_column(env):


pedrobaeza:
Isn't this covered by the base module and the fix should be to not perform anything here?

victoralmau:
I don't quite understand what you mean by the base module. 
Let me try to explain what is happening:
- In v16, there was a **non**-translatable `description` field.
- In v17, the `invoice_label` field was added as translatable.

Simply renaming it would not be enough, because the database column will not be in json (with the appropriate translations) and Odoo will auto-set the empty json column (without translations).

IMO, using the `migrate_translations_to_jsonb` method after renaming would be the right solution, although I think this needs to be thoroughly tested and confirmed before merging.

pedrobaeza (approved):
OK, you have answered me indirectly. The key is that there's a renaming in between: description > invoice_label. That can be the reason why the method in the `base` module doesn't handle the transition to JSONB.

I wait for @hbrunn to confirm the change as well.

hbrunn (commented):
what does this change? `migrate_translations_to_jsonb` calls `
convert_column_translatable` [internally](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_160.py#L71-L73), and only adds running the translation queries afterwards. But as you rightly pointed out the field isn't translatable in [v16](https://github.com/OCA/OCB/blob/16.0/addons/account/models/account_tax.py#L118), so I wouldn't expect this to make any difference?

Which problem are you solving here?

victoralmau:
Oops, I think I forgot to clarify something important: the `l10n_multilang` module (no longer available in v17). 
Anyone who has the `l10n_multilang` module installed will have this problem. I'm going to improve this code so that only those who have defined that field as translate need to apply this (this should apply to other fields in other models).

hbrunn:
aaah yes, now that makes sense. No, just keep the code as it is, as the translation queries shouldn't do anything for the case where the field isn't translated.

Just add a comment that you deliberately call `migrate_translations_to_jsonb` instead of `convert_column_translatable` to support the case you have translate=True on the field in v16

MiquelRForgeFlow (approved):

hbrunn (changes requested):
please revert to your previous version, because the way it's now, it's wrong and only doesn't fail tests because the multilang addon is installed in the CI database.

As written above, I think it's perfectly safe to just call `migrate_translations_to_jsonb` unconditionally

hbrunn (approved):
/ocabot merge nobump

PR #5358, opened 2025-09-02
[REF] 'OU-' prefixes are used only for old OpenUpgrade versions.

legalsylvain:
See : https://github.com/OCA/odoo-community.org/pull/221#issuecomment-3227193378

Rational : In V13 and before, OpenUpgrade was a fork, so it was important to distinguish Odoo commits and Openupgraders commits. Nowadays, OpenUpgrade is just 'two modules', so using classic 'IMP' / 'ADD' / 'REF' / 'FIX' prefix are OK.

### DIFF
--- a/docsource/090_contribute.rst
+++ b/docsource/090_contribute.rst
@@ -77,15 +77,19 @@ Examples: all the noupdate=0 new and del ir* records, or the NEW and DEL lines f
   .. code-block:: shell
 
       git add .
-      git commit -am "[OU-ADD] account"
+      git commit -am "[ADD] account"
       git push MY_REMOTE 16.0-mig-account
 
 
   **Note:**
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
I don't agree with this change. They should be distinguished on all versions.

luisg123v (approved):

StefanRijnhart (approved):

MiquelRForgeFlow (approved):

hbrunn (approved):

pedrobaeza:
@hbrunn I didn't agree on the change.

hbrunn:
and you've been outvoted, several times

pedrobaeza:
But this is not a voting. The way OCA usually works is consensus.

PR #5365, opened 2025-09-06
[18.0][web_tour] Obsolete security rules block login after migration

demodooit:
Impacted versions:
- OpenUpgrade 18.0
- Module web_tour

Steps to reproduce:
1. Migrate a DB from 17.0 to 18.0 with `web_tour` installed.
2. After migration, login fails due to obsolete security rules for model `web_tour.tour`.

Current behavior:
- Login blocked after migration.

Expected behavior:
- Obsolete security rules should be removed during `end-migration.py`.

Proposed fix:
- Update `openupgrade_scripts/scripts/web_tour/18.0.1.0/end-migration.py` to drop security rules for model `web_tour.tour`.

@openupgrade.migrate()
def migrate(env, version=None):
    """
    Set consumed tours from legacy table after migration and
    remove obsolete security rules.
    """
    openupgrade.logged_query(
        env.cr,
        f"""
        INSERT INTO res_users_web_tour_tour_rel
        (res_users_id, web_tour_tour_id)
        SELECT legacy_table.user_id, web_tour_tour.id
        FROM
        {openupgrade.get_legacy_name('web_tour_tour')} legacy_table,
        web_tour_tour
        WHERE web_tour_tour.name=legacy_table.name
        ON CONFLICT DO NOTHING
        """,
    )

    openupgrade.logged_query(
        env.cr,
        """
        DELETE FROM ir_rule
        WHERE model_id = (
            SELECT id FROM ir_model WHERE model = 'web_tour.tour'
        )
        AND domain_force LIKE '%user_id%';
        """,
    )

### DIFF
--- a/openupgrade_scripts/scripts/web_tour/18.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/web_tour/18.0.1.0/end-migration.py
@@ -4,7 +4,8 @@ from openupgradelib import openupgrade
 @openupgrade.migrate()
 def migrate(env, version=None):
     """
-    Set consumed tours from legacy table after migration
+    Set consumed tours from legacy table after migration and
+    remove obsolete security rules.
     """
     openupgrade.logged_query(
         env.cr,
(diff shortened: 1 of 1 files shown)


hbrunn (changes requested):
please figure out where the problematic rule comes from, because `web_tour.own_tours` is supposed to be [deleted](https://github.com/OCA/OpenUpgrade/actions/runs/17838991581/job/50723339257#step:13:9113) during the migration. Chances are the issue is with your database or the way you run OU.

Then your sql statement is pretty broad, and given ir.rules usually have xmlids, if we actually need this, you should use [delete_records_safely_by_xml_id](https://oca.github.io/openupgradelib/API.html#openupgradelib.openupgrade.delete_records_safely_by_xml_id) - but as of above, I doubt it

celm1990:
I have the same issue

PR #5367, opened 2025-09-08
[18.0][OU-FIX] hr_holidays: rename private_name to name throw an error

FernandoRomera:
In 17.0 exists private_name and name

### DIFF
--- a/openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
@@ -174,6 +174,8 @@ def split_employee_leaves(env):
 
 @openupgrade.migrate()
 def migrate(env, version):
+    if openupgrade.column_exists(env.cr, "hr_leave_allocation", "name"):  # from 13.0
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (changes requested):
> In 17.0 exists private_name and name

Wrong. Proof:
Selection_4970
Selection_4971

But in fact in Odoo 13.0 this field was stored. So what you have to do instead is to add at the beginning:
```python
if openupgrade.column_exists(env.cr, "hr_leave_allocation", "name"): # from 13.0
    openupgrade.rename_columns(env.cr, {"hr_leave_allocation": [("name", False)]})
```

MiquelRForgeFlow (approved):

Lukwos:
I have an error with this because the column "name" is renamed to "False". I think you ( @MiquelRForgeFlow ) mean to use None instead : 
```python
openupgrade.rename_columns(env.cr, {"hr_leave_allocation": [("name", None)]})
```

MiquelRForgeFlow:
Oh, yes, it should be `None` instead of `False`.

pedrobaeza:
@FernandoRomera please do the correction.

MiquelRForgeFlow:
Fixed.

PR #5369, opened 2025-09-09
[17.0][OU-FIX] account: Avoid the error when executing the migrate_translations_to_jsonb method

### LINKED ISSUE #5368: [17.0]account: key_error: 'account_tax' on method migrate_translations_to_jsonb
## Module
account

## Describe the bug
https://github.com/OCA/OpenUpgrade/commit/b097c5d6a42562a89be28032667eb699b2aa0fbb#diff-406c7e73ba75feeb54738e35feff14fb8fe419b838b4461e864cf033b2fa0f52R78
I believe the JSON function should use pure SQL instead of the Odoo API.

## To Reproduce
**Affected versions**: 17.0

1. Launch migration in branch 17.0

**Expected behavior**
No errors should occur.

**Additional context**
The error appears in a private project with custom data but yesterday, before the referenced commit, the migration ran without issues. I haven’t deeply investigated the error, but I attempted to replace account_tax with account.tax in the tuple, as it seems to be processed by the Odoo API. However, the error persists with account.tax. This may be because it’s a pre-migration script, and the model is not yet available in the registry.

victoralmau:
Avoid the error when executing the migrate_translations_to_jsonb method

Related to https://github.com/OCA/OpenUpgrade/pull/5355

Fixes #5368 

Please @pedrobaeza and @hbrunn can you review it?

@Tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
@@ -1,7 +1,7 @@
 # Copyright 2024 Viindoo Technology Joint Stock Company (Viindoo)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openupgradelib import openupgrade
+from openupgradelib import openupgrade, openupgrade_160
 
 from odoo import models
 
@@ -558,6 +558,15 @@ def _rename_coa_elements_xmlids(env):
         )
 
 
+def _convert_account_tax_description(env):
+    # Deliberately call migrate_translations_to_jsonb instead of
+    # convert_column_translatable to support the case you have translate=True on
+    # the field in v16 (l10n_multilang compatibility).
+    openupgrade_160.migrate_translations_to_jsonb(
+        env, [("account.tax", "invoice_label")]
+    )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     _account_payment_term_migration(env)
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Isn't Odoo in between pre and post doing anything with the column or giving an error?

javierjcf:
:+1: It is working for me and fixing #5368

Vicenteagf:
lgtm

pedrobaeza (approved):
OK then

hbrunn:
wait a second here, are we sure this actually does something?

The ir_translation->jsonb transition was 15->16, so if in v16 the field was translatable, it was a json field all along.

@victoralmau is it possible you're migrating a database that has been migrated from v15 already, and what you're "fixing" here is actually a followup issue from the v16 migration?

Because I don't see how `migrate_translations_to_jsonb` ever [does](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_160.py#L59) anything in a new v16 database

pedrobaeza:
Remember that this happens only with `l10n_multilang` installed.

hbrunn:
yes, but still `l10n_multilang` installed in v16 makes the fields in question json fields, so I don't understand how `migrate_translations_to_jsonb` can do anything for the general case.

The only way I can imagine is that you had translations in v15, migrated to v16 and for whatever reason then the translations didn't end up in the json columns.

pedrobaeza:
I can't say right now. @victoralmau can you answer?

PR #5377, opened 2025-09-15
[17.0][OU-FIX] account: Remove the unnecessary _convert_account_tax_description() method in post-migration.

victoralmau:
Remove the unnecessary _convert_account_tax_description() method in post-migration.

Related to https://github.com/OCA/OpenUpgrade/pull/5369#issuecomment-3284403042

This change is not necessary because when migrating from 15 to 16, the description column in account_tax will already be json (https://github.com/OCA/OpenUpgrade/blob/4af8572c7f3e313773db558ba382b5ffd19a157c/openupgrade_scripts/scripts/base/16.0.1.3/end-migration.py#L39), therefore, in v17, you only need to rename the column (this is already done in pre-migration).

Please @pedrobaeza and @hbrunn can you review it?

@Tecnativa TT57778

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/post-migration.py
@@ -1,7 +1,7 @@
 # Copyright 2024 Viindoo Technology Joint Stock Company (Viindoo)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openupgradelib import openupgrade, openupgrade_160
+from openupgradelib import openupgrade
 
 from odoo import models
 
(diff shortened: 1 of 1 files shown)


hbrunn (approved):

hbrunn:
we should restore the state before your first pr though I think, with [converting the column](https://github.com/OCA/OpenUpgrade/pull/5355/files#diff-406c7e73ba75feeb54738e35feff14fb8fe419b838b4461e864cf033b2fa0f52L76), right?

victoralmau:
> we should restore the state before your first pr though I think, with [converting the column](https://github.com/OCA/OpenUpgrade/pull/5355/files#diff-406c7e73ba75feeb54738e35feff14fb8fe419b838b4461e864cf033b2fa0f52L76), right?

There is no need to convert anything, the column is already json in the database (since 16).

pedrobaeza (approved):

hbrunn:
not when it's not translated

PR #5378, opened 2025-09-15
[18.0] [OU-ADD] l10n_es_igic: Merged into l10n_es

luisDIXMIT:
Migration script added. Four XML IDs are empty because I missed them in v18. I have to test it in a local database when my doubts have been solved.


pedrobaeza:
Have you checked if there's any need of XML-ID renaming of the existing taxes?

luisDIXMIT:
> Have you checked if there's any need of XML-ID renaming of the existing taxes?

No, sorry. I will check. Thanks!

etobella (approved):
LGTM. Thanks

openupgrade_scripts/scripts/l10n_es/18.0.5.4/pre-migration.py
```diff
@@ -0,0 +1,302 @@
...
+    (["account_pymes_canary_2405"], "account_pymes_2405"),
+    (["account_pymes_canary_2493"], "account_pymes_2493"),
+    (["account_pymes_canary_2494"], "account_pymes_2494"),
+    (["account_pymes_canary_2495"], "account_pymes_2495"),
+    (["account_pymes_canary_255"], "account_pymes_255"),
+    (["account_pymes_canary_2935"], "account_pymes_2935"),
+    (["account_pymes_canary_296"], "account_pymes_296"),
+    (["account_pymes_canary_490"], "pgc_pyme_490"),
+    (["account_pymes_canary_551"], "pgc_pyme_551"),
+    (["account_pymes_canary_4707"], ""),
```
pedrobaeza: You have to remove this and the following as they are doing nothing.

luisDIXMIT: I did this because I couldn’t find this in v18, so I don’t know what I must do. I added a comment in the description of this PR asking what I should do in this case. Is it a missing account, or should I mark it as removed?

etobella: I think this account doesn't make sense anymore.... it should be ported to the existent account account_common_canary_47071, but I am not sure which is the best way to handle this specific migration.

luisDIXMIT: I found this, and I think it’s still valid. WDYT? https://www.boe.es/buscar/doc.php?id=BOE-A-1992-28829

openupgrade_scripts/scripts/l10n_es/18.0.5.4/pre-migration.py
```diff
@@ -0,0 +1,302 @@
...
+    (["account_full_canary_2404"], "account_full_2404"),
+    (["account_full_canary_2405"], "account_full_2405"),
+    (["account_full_canary_2493"], "account_full_2493"),
+    (["account_full_canary_2494"], "account_full_2494"),
+    (["account_full_canary_2495"], "account_full_2495"),
+    (["account_full_canary_2550"], "account_full_2550"),
+    (["account_full_canary_2553"], "account_full_2553"),
+    (["account_full_canary_257"], "account_full_257"),
+    (["account_full_canary_466"], "account_full_466"),
+    (["account_full_canary_4707"], ""),
```
pedrobaeza: This doesn't make sense: or you remove them, or you map to something, as keeping them this way won't make anything.

luisDIXMIT: I'm with you on this, I'm just asking for help to know what to map it against, since there's nothing in 18. As I see we're all in the same situation and from what you mentioned, I can just leave it unmapped, so I'll do that and not block the topic any further. Thanks.

PR #5383, opened 2025-09-19
[FIX] hr_expense is Done

SecretAgentNull:
According to [this](https://github.com/OCA/OpenUpgrade/pull/5130) merged PR, "hr_expense" migration scripts are Done, however they are not marked as Done on the page detailing coverage.


pedrobaeza (approved):
Thanks for this. It was missed.

PR #5385, opened 2025-09-20
chefk if project_task field user_id exists

agent-z28:


pedrobaeza:
Can you explain why it shouldn't exist?

agent-z28:
@pedrobaeza 
I just happend that me and @PCatinean had the Problem in completely diffrent databases and systems.
When he told me that he had the same problem, i just thought it would be nice to share the fix.

Also i'm aware that the field is there in:
https://github.com/odoo/odoo/blob/14.0/addons/project/models/project.py#L603

Maybe it happens when the OpenUpgrade get's started a second time for the same version.

pedrobaeza:
Well, it we don't know the reason/root, it's hard to determine if this is the proper solution. For now, this is only omitting a renaming of a column, but maybe it's because the existing column coming from other module that is adding it, and contains incorrect data, and then the problem is at another level. Do you understand my point?

agent-z28:
@pedrobaeza Yes, I understand.
The idea was simply to check if the column exists first, and then proceed with the rename — this ensures the operation runs correctly every time. 

@PCatinean did you find anthing more ?

PR #5396, opened 2025-10-07
[18.0][OU-FIX] hr_holidays: do not duplicate multi employee leaves

remi-filament:
Multi-employee leaves and allocations once validated already created duplicate leave / allocation for each employee.
Current script by rewriting multi-employee leave / allocation to final employees was creating duplicate leave / allocation causing all leave counters to be wrong.

The duplication is only done in case leave / allocation is in 'to approve' state, otherwise the multi-employee leave / allocation is deleted.

### DIFF
--- a/openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
@@ -51,7 +51,7 @@ def update_states(env):
         env.cr,
         """
         UPDATE hr_leave_allocation
-        SET state = 'cancel'
+        SET state = 'refuse'
         WHERE active IS DISTINCT FROM TRUE
         """,
     )
@@ -77,7 +77,26 @@ def update_allocation_validation_type(env):
 
 
 def split_employee_leaves(env):
+    """
+    Confirmed leave or allocation already have duplicated leave for each employee
+    Since multi employee leave / allocation are not allowed anymore,
+    these should be removed
+    Draft (= 'confirm' state) are the only ones which have not yet
+    been duplicated and therefore should be transformed
+    """
     for table in ["hr_leave", "hr_leave_allocation"]:
+        # First remove all multi-employee leaves / allocations
+        # already confirmed or refused
+        openupgrade.logged_query(
+            env.cr,
+            f"""
+            DELETE FROM {table}
+            WHERE state IN ('refuse', 'validate1', 'validate', 'cancel') AND (
+                (holiday_type = 'employee' AND multi_employee)
+                OR holiday_type IN ('company', 'category', 'department'))
+            """,
+        )
+        # Then for the remaining ones we transform existing data for each employee
         env.cr.execute(
             f"""
             SELECT column_name
@@ -106,7 +125,7 @@ def split_employee_leaves(env):
             SELECT leave.id, 'company', array_agg(he.id) as employee_ids
             FROM {table} AS leave
             JOIN hr_employee he ON he.company_id = leave.mode_company_id
-            WHERE leave.holiday_type = 'company'
+            WHERE leave.holiday_type = 'company' and he.active = True
             GROUP BY leave.id
             """
         )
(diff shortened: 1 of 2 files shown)


remi-filament:
@MiquelRForgeFlow would you mind having a look at that one ?

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
```diff
@@ -51,7 +51,7 @@ def update_states(env):
         env.cr,
         """
         UPDATE hr_leave_allocation
-        SET state = 'cancel'
+        SET state = 'refuse'
```
MiquelRForgeFlow, line 54: why?

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
```diff
@@ -106,7 +125,7 @@ def split_employee_leaves(env):
             SELECT leave.id, 'company', array_agg(he.id) as employee_ids
             FROM {table} AS leave
             JOIN hr_employee he ON he.company_id = leave.mode_company_id
-            WHERE leave.holiday_type = 'company'
+            WHERE leave.holiday_type = 'company' and he.active = True
```
MiquelRForgeFlow, line 128: this should be unnecessary, because if active != true, then by `update_states()` the state is cancel/refuse and thus it is already deleted by your first query in `split_employee_leaves()`

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
remi-filament: Cancel state was only added to hr_leave but does not exist on hr_leave_allocation

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
remi-filament: Hère I am checking whether employee is active (in order to avoid creating leave/allocation for all past employees), the update_states is updating inactive leave/allocation, not employees

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
MiquelRForgeFlow: Good point

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
MiquelRForgeFlow: Good point

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
```diff
@@ -77,7 +77,26 @@ def update_allocation_validation_type(env):
 
 
 def split_employee_leaves(env):
+    """
+    Confirmed leave or allocation already have duplicated leave for each employee
+    Since multi employee leave / allocation are not allowed anymore,
+    these should be removed
```
MiquelRForgeFlow, line 83: Could you explain me a bit more or point me to the relevant code in v17? I don't see what you meant by duplicated leave. The point of `split_employee_leaves` is to transform multi employee leave into single ones, not to delete them.

remi-filament: Yep, when calling action_validate on hr_leave or on hr_leave_allocation, Odoo will first create a leave / allocation for each employee : 
- on hr_leave : https://github.com/odoo/odoo/blob/9466d11c8764885ef7b1208a255bf89a04914e2b/addons/hr_holidays/models/hr_leave.py#L1394-L1405 and then validate these new leaves.
- on hr_leave_allocation : https://github.com/odoo/odoo/blob/9466d11c8764885ef7b1208a255bf89a04914e2b/addons/hr_holidays/models/hr_leave_allocation.py#L792

Therefore in most of the cases you already have splitted leaves / allocations for each employee, except in the case where leave / allocation is in draft since it creates splitted only on validation action.

Let me know if you need more information ?

MiquelRForgeFlow: So why Odoo also creates this multi employee leave if it also creates the single ones? why the duplication?

remi-filament: It was actually simpler in terms of UX : you could for instance assign a 25 days leave allocation to all employees from company without having to create them one by one.
When you validated the company leave allocation, Odoo would then create and validate allocation for each employee of the company.

MiquelRForgeFlow: Nice, good to know.

MiquelRForgeFlow (approved):

PR #5397, opened 2025-10-08
[18.0][OU-FIX] account: drop account_root view

remi-filament:
account_root view is not stored in DB anymore in v18. account.root model was moved from account_account.py to its own file account_root.py and the init function creating VIEW in DB was removed at the same time : 
https://github.com/odoo/odoo/commit/854c3b27aa5476c208572f19e64f8f3364bfc381

Not sure why this it not detected in upgrade_analysis.txt though (probably because model still exists but is not using SQL VIEW anymore).

I am not sure if I should add information somewhere (adding entries in upgrade_analysis_work.txt for instance) ?

### DIFF
--- a/openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
@@ -42,6 +42,15 @@ _new_columns = [
 ]
 
 
+def _drop_sql_views(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DROP VIEW IF EXISTS account_root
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (commented):
> adding entries in upgrade_analysis_work.txt for instance

Yes, it would be nice.

remi-filament:
> > adding entries in upgrade_analysis_work.txt for instance
> 
> Yes, it would be nice.

Thanks @MiquelRForgeFlow I have added a line in upgrade_analysis_work.txt

MiquelRForgeFlow (approved):

PR #5398, opened 2025-10-08
[18.0][OU-FIX] hr_holidays: remove if employee is inactive

MiquelRForgeFlow:
Follow-up of https://github.com/OCA/OpenUpgrade/pull/5396.

### DIFF
--- a/openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
@@ -125,7 +125,7 @@ def split_employee_leaves(env):
             SELECT leave.id, 'company', array_agg(he.id) as employee_ids
             FROM {table} AS leave
             JOIN hr_employee he ON he.company_id = leave.mode_company_id
-            WHERE leave.holiday_type = 'company' and he.active = True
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
@remi-filament what do you think?

openupgrade_scripts/scripts/hr_holidays/18.0.1.6/pre-migration.py
```diff
@@ -125,7 +125,7 @@ def split_employee_leaves(env):
             SELECT leave.id, 'company', array_agg(he.id) as employee_ids
             FROM {table} AS leave
             JOIN hr_employee he ON he.company_id = leave.mode_company_id
```
remi-filament, line 127: I think you can also remove that line which I added so that I can filter on employees

remi-filament (changes requested):
I think it will take more time to do a browse then filtered (that will need to fetch fields from hr_employee) rather that doing it in SQL, but it should work the same !

MiquelRForgeFlow:
> I think it will take more time to do a browse then filtered

Yes, but I am afraid that if in some case all employees are inactive in a leave, then the leave will not be selected, right?

remi-filament (approved):

PR #5409, opened 2025-10-20
[OU_ADD] website_sale_product_description: rename field

luisDIXMIT:
Merged module website_sale_product_description into website_sale.

Renamed field public_description to description_ecommerce.

### DIFF
--- a/openupgrade_scripts/scripts/website_sale/18.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/website_sale/18.0.1.1/pre-migration.py
@@ -15,3 +15,10 @@ def migrate(env, version):
     openupgrade.logged_query(
         env.cr, "UPDATE product_ribbon SET text_color='' WHERE text_color IS NULL"
     )
+    # If public_description exists, copy its content to description_ecommerce because
+    # this field has been deprecated in favor of description_ecommerce in 18.0
+    if openupgrade.column_exists(env.cr, "product_template", "public_description"):
+        openupgrade.logged_query(
+            env.cr,
+            "UPDATE product_template SET description_ecommerce=public_description",
+        )
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -73,6 +73,7 @@ merged_modules = {
     # OCA/e-commerce
     "website_sale_product_attachment": "website_sale",
     "website_sale_product_attribute_filter_collapse": "website_sale",
+    "website_sale_product_description": "website_sale",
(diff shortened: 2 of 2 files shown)


openupgrade_scripts/scripts/website_sale/18.0.1.1/pre-migration.py
```diff
@@ -15,3 +15,6 @@ def migrate(env, version):
     openupgrade.logged_query(
         env.cr, "UPDATE product_ribbon SET text_color='' WHERE text_color IS NULL"
     )
+    openupgrade.logged_query(
```
etobella: No, you should add here a check if the field public_description exists. Imagine that the module is not installed 😉

etobella (changes requested):

openupgrade_scripts/scripts/website_sale/18.0.1.1/pre-migration.py
luisDIXMIT: Thanks, is correct now?

openupgrade_scripts/scripts/website_sale/18.0.1.1/pre-migration.py
```diff
@@ -15,3 +15,8 @@ def migrate(env, version):
     openupgrade.logged_query(
         env.cr, "UPDATE product_ribbon SET text_color='' WHERE text_color IS NULL"
     )
+    if openupgrade.column_exists(env.cr, "product_template", "public_description"):
+        openupgrade.logged_query(
```
etobella, line 21: Add a comment to know why it is here, but yes, this is how it should be.

etobella (approved):

hbrunn (changes requested):
product.template#description_ecommerce existed in v17 already, so I don't think it should be overwritten unconditionally

etobella:
Probably that means that the migration script should be in 17, isn't it?

hbrunn:
but the module that's merged in providing the field has been migrated to v17. probably that shouldn't have happened?

etobella:
Probably... At least in my opinion, that need can be filled with the field provided by odoo :thinking:

luisDIXMIT:
Hi @pedrobaeza, is this no longer necessary?

pedrobaeza:
According comments, it should be in 17

luisDIXMIT:
So, I will move it, thanks! @pedrobaeza

pedrobaeza:
Please talk first with Enric to be sure about the proper solution.

PR #5410, opened 2025-10-20
[18.0][OU-ADD] apriori: Rename account_statement_import_paypal to acc…

javierjcf:
…ount_statement_import_online_paypal


hbrunn (approved):

pedrobaeza:
Can you reference the PR where the rename is done?

javierjcf:
My apologies, I think I was wrong. After checking, I realized it's not a rename, but rather two different modules. The account_statement_import_paypal is for import from csv file. This module is missing in v18.

PR #5432, opened 2025-10-29
[19.0] MIG openupgrade_scripts

legalsylvain:
Use https://github.com/OCA/server-tools/pull/3395 to generate first analysis to estimate workload for the new V19 project.

Note : weird data generated. See : https://github.com/OCA/server-tools/pull/3395#issuecomment-3459483379


MiquelRForgeFlow:
`openupgrade_scripts` module is already merged. Only it needs the analysis files.

hbrunn:
no, what we need is merging https://github.com/OCA/OpenUpgrade/pull/5389 so that this happens automatically

hbrunn:
@legalsylvain we had the first successful [run](https://github.com/OCA/OpenUpgrade/pull/5499) of the analysis cronjob for v19, so I suggest to close this

PR #5434, opened 2025-10-29
[18.0][OU-ADD] stock_multi_warehouse_security: Renamed to stock_warehouse_s…

christian-ramos-tecnativa:
…ecurity
Depends on:

- [ ] https://github.com/OCA/stock-logistics-warehouse/pull/2445

@Tecnativa TT58828


pedrobaeza:
Was the original module ever merged in OCA. If not, then this is not needed.

christian-ramos-tecnativa:
It seems like that wasn't the case. Thanks

PR #5435, opened 2025-10-29
[18.0][OU][ADD] web_editor: Fix snippets

etobella:
Odoo has changed the format of snippets between 17 and 18.

https://github.com/odoo/odoo/blob/17.0/addons/web_editor/models/ir_ui_view.py#L489-L504

IT needs to be adapted or edition will fail.

https://github.com/odoo/odoo/blob/18.0/addons/web_editor/models/ir_ui_view.py#L492-L504


pedrobaeza:
Change the rst to put `Done` in the module instead of `Nothing to do`.

openupgrade_scripts/scripts/web_editor/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,32 @@
+import logging
+
+from lxml import etree
+from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate()
+def migrate(env, cr):
```
luisDIXMIT:
```suggestion
def migrate(env, version):
```
To fix checks

pedrobaeza (approved):

pilarvargas-tecnativa:
I think #5575 may be related to the execution order of this migration script.

website depends on web_editor, so this post-migration runs before website is migrated. At this point, website.snippets may still have the v17 structure, while this script updates the custom views through the ORM (view.write()) to already reference the v18 structure:

<xpath expr="//snippets[@id='snippet_custom']" position="inside">

Would it make more sense to move this migration to website, so it runs after website.snippets has been updated, or alternatively perform this particular arch_db adaptation directly in SQL?

What do you think?

etobella:
No to website, as not everyone has website... Maybe on end-migration?

pilarvargas-tecnativa:
> No to website, as not everyone has website... Maybe on end-migration?

Yes, end-migration makes more sense and avoids depending on website being installed. I can move this transformation there so the ORM write is performed once all installed modules have been migrated.

Issue #5437, opened 2025-11-03
No update template (QWeb) views updated by OpenUpgrade get openupgrade_scripts path in arch_fs


remi-filament:
## Module
openupgrade_framework

## To Reproduce
Get a database in v17 with at least digest installed and upgrade to v18 (same occurs between v16 and v17)
When updating 
The arch_fs field of template ir.ui.views is set to e.g. openupgrade_scripts/scripts/digest/17.0.1.1/noupdate_changes.xml
With this arch_fs, it is no longer possible to Compare / Reinit view to file based view since it cannot be found anymore once DB is moved to final server (without openupgrade_scripts)

**Expected behavior**
QWeb views should keep arch_fs from odoo modules, not be replaced by openupgrade_scripts source

For instance on a database migrated from v16 to v17 then to v18 : 

`# select id,name,arch_fs,type from ir_ui_view where arch_fs like '%openupgrade%';`
  id  |             name              |                                    arch_fs                                    | type 
------|-------------------------------|-------------------------------------------------------------------------------|------
 3277 | digest_mail_layout            | openupgrade_scripts/scripts/digest/17.0.1.1/noupdate_changes.xml              | qweb
 3647 | project_message_user_assigned | openupgrade_scripts/scripts/project/17.0.1.3/noupdate_changes.xml             | qweb
 3280 | digest_tool_kpi               | openupgrade_scripts/scripts/digest/17.0.1.1/noupdate_changes.xml              | qweb
 3460 | candidate_hired_template      | openupgrade_scripts/scripts/hr_recruitment/18.0.1.1/noupdate_changes_work.xml | qweb
 2472 | portal_share_template         | openupgrade_scripts/scripts/portal/17.0.1.0/noupdate_changes.xml              | qweb
 3279 | digest_section_mobile         | openupgrade_scripts/scripts/digest/17.0.1.1/noupdate_changes.xml              | qweb
 3278 | digest_mail_main              | openupgrade_scripts/scripts/digest/18.0.1.1/noupdate_changes.xml              | qweb

remi-filament:
Hi @hbrunn do you think this one can be fixed in openupgrade_framework somewhere ?

hbrunn:
[...]

PR #5445, opened 2025-11-07
[18.0][FIX] don't set install_filename when loading views from openupgrade_scripts

hbrunn:
Remove install_filename from context if it's from openupgrade_scripts.
Without this, arch_fs will point to openupgrade_scripts' file which most likely
won't exist when the migrated database is deployed, which breaks resetting views

see #5437

@remi-filament please verify this fixes your issue for v18, and then cherry pick to v17 without the test

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
@@ -5,6 +5,7 @@ import logging
 
 from odoo import api
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


StefanRijnhart (approved):

legalsylvain (approved):
Thanks ! 

makes sense.

/ocabot merge nobump

pedrobaeza:
Isn't this something to do also on 17.0?

hbrunn:
yes, probably even older ones

lorenzomorandini:
Hi @hbrunn can this be merged into 17.0 too please? Currently trying to upgrade 16->17->18 and 18 can not start because of this problem.

PR #5447, opened 2025-11-08
[18.0][IMP] account: Pre-compute fields in order to make it faster

etobella:
- made_sequence_gap field in account.move

### DIFF
--- a/openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/18.0.1.3/pre-migration.py
@@ -39,6 +39,7 @@ _new_columns = [
     ("account.payment", "memo", "char"),
     ("account.payment", "state", "selection"),
     ("account.payment", "is_sent", "boolean"),
+    ("account.move", "made_sequence_gap", "boolean", True),
 ]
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

remi-filament:
Hi @etobella as reported here : https://github.com/OCA/account-financial-tools/issues/2133#issuecomment-3528129252
Initializing the field made_sequence_gap to True causes issues during migration of database with account_move_name_sequence installed (which disable the fields sequence_prefix and sequence_number)

That would mean that in case these fields are not used (for instance in account_move_name_sequence) we would need to reset all made_sequence_gap back to False and therefore would not make migration faster...
Does anyone have a better idea ?

Best Regards

etobella:
@remi-filament 

It will be faster for just one reason: Now it is done with SQL,  if you have that module, it will be done by python code.
In any case, there are 2 options: 
1. Add a verification of having that module installed in order to skip it. 
2. Follow the script and then make the changes directly in the module.

pedrobaeza:
For the record, I strongly advice to not use that module.

remi-filament:
Thanks @etobella I have made a PR on account_move_name_sequence to add a new migration script to set field to False.

I understand your point of view @pedrobaeza , yet there are some corner cases which are not properly handled by Odoo default sequence-guessing, like if you want to have year or year and month in your sequence but still do not want to reset sequence to 1 on each year or year-month change... In these cases, I have not found any other solution than to bring back ir.sequence where you can choose to not use range_year but only year so that sequence is never reset.

pedrobaeza:
@remi-filament IMO, that customer whims shouldn't be satisfied, as they are more harmless than other thing.

remi-filament:
Well, this is what we are using internally, and it was working properly before Odoo 14. Yet it is true that we could change that to accomodate with Odoo limitations added in v14 :)

Although, for other customers in the same case, it is a bit harsh to explain to them it is a whim when it was working properly on previous version and they have to change the way they are doing accounting because of migration to newer version...

pedrobaeza:
I won't say "whim" to him, but only tell them that now it's not possible because now the system is better in this and that about sequences (like not having to manipulate the next value if removing the last invoice), but they have to resign from that numbering. And they usually accept. And if not, then "motivate" them with a substantial increase in the maintenance fee (which is justified, as there can be a lot of problems with that module).

PR #4648, opened 2024-10-29
[17.0][OU-ADD] sale_product_configurator: Nothing to do

acpMicrocom:

PR #4489, opened 2024-06-26
[17.0][OU-ADD] point_of_sale: migration to 17.0

flachica:

PR #5456, opened 2025-11-18
[17.0][OU-ADD] pos_sale_product_configurator: Nothing to do

pedrobaeza:
Even no analysis file.

@Tecnativa TT57495


pedrobaeza:
/ocabot migration pos_sale_product_configurator

<dependency>Depends on : 

- point_of_sale : #4489
- sale_product_configurator : #4648</dependency>

PR #5463, opened 2025-12-01
[17.0][OU-IMP]hr: Add computation of some fields

etobella:
Otherwise, an error might be rise if we have [hr_holidays_natural_period](https://github.com/OCA/hr-holidays/tree/18.0/hr_holidays_natural_period) installed because the option might have not been selected yet.

### DIFF
--- a/openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
+++ b/openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
@@ -29,8 +29,10 @@ hr_holidays  / hr.leave                 / mode_company_id (many2one)    : not a 
 # NOTHING TO DO: states removed only
 
 hr_holidays  / hr.leave                 / number_of_days (float)        : now a function
+# NOTHING TO DO
+
 hr_holidays  / hr.leave                 / number_of_hours (float)       : NEW isfunction: function, stored
-# NOTHING TO DO: computed fields
+# DONE: pre-migration: add column for avoiding the computation
 
 hr_holidays  / hr.leave                 / rating_ids (one2many)         : NEW relation: rating.rating
 hr_holidays  / hr.leave                 / resource_calendar_id (many2one): NEW relation: resource.calendar, hasdefault: compute
--- a/openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_holidays/17.0.1.6/pre-migration.py
@@ -20,6 +20,18 @@ _column_renames = {
     ],
 }
 
+_field_add = [
+    (
+        "number_of_hours",
+        "hr.leave",
+        "hr_leave",
+        "float",
+        None,
+        "hr_holidays",
+        0.0,
+    ),
+]
+
 
 def _pre_create_accrual_plan_active(cr):
     """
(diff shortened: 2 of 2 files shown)


etobella:
@MiquelRForgeFlow May I ask for your opinion? :pray:

MiquelRForgeFlow:
Hehe An alternative should be adding migration scripts in `hr_holidays_natural_period`, by recomputing those fields...

etobella:
The problem is that the compute is failing because it has a value that doesn't exist yet

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (changes requested):
My concern on this is the increase of time on the execution, or to have different data in past leaves. Why are you recomputing the duration? Remember that you may adjust manually the duration. I don't think it should be recomputed.

etobella:
After a full review of the error I found that the problem comes from the fact that number_of_hours was not stored in v16, but stored in v17. If we don't apply something, the function is executed no matter what, and it might raise errors if we have other modules like natural_days.

Options I see:

- Create the field and leave it blank (reexecute only for the leaves needed?)
- Reexecute for everyone as we did it before but on the end-migration, to ensure that other changes will work.

Leaving as it is will not work if you are using an OCA module and there is no natural way to solve it.

pedrobaeza:
I agree on pre-adding the field, but we shouldn't trigger the regular computation, as that one computes both `number_of_days` and `number_of_hours`, overwriting manual changes done on `number_of_days` field.

etobella:
legit for me. I will manage it

pedrobaeza:
@etobella do  you plan to add anything more?

MiquelRForgeFlow (approved):

etobella:
@pedrobaeza no, I think I made all the changes that you asked for.

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
```diff
@@ -30,7 +30,7 @@ hr_holidays  / hr.leave                 / mode_company_id (many2one)    : not a
 
 hr_holidays  / hr.leave                 / number_of_days (float)        : now a function
```
pedrobaeza, line 31:
```suggestion
hr_holidays  / hr.leave                 / number_of_days (float)        : now a function
# NOTHING TO DO

```

openupgrade_scripts/scripts/hr_holidays/17.0.1.6/upgrade_analysis_work.txt
```diff
@@ -30,7 +30,7 @@ hr_holidays  / hr.leave                 / mode_company_id (many2one)    : not a
...
 hr_holidays  / hr.leave                 / number_of_hours (float)       : NEW isfunction: function, stored
-# NOTHING TO DO: computed fields
+# Pre-migration script and end-migration script to recompute
```
pedrobaeza:
```suggestion
# DONE: pre-migration: add column for avoiding the computation
```

pedrobaeza (approved):

lorenzomorandini:
Hi everyone, sorry if I reply in a merged PR. I just came across the field `number_of_hours` not being set while upgrading 16->17->18. I understand why you decided to do so but how am I supposed to populate that field for rows already in DB? Thank you!

pedrobaeza:
@lorenzomorandini you should launch by an script the recomputation if you are hit by this, because the problem is that the recomputation may overwrite legit different data (due to manually changed days).

PR #5465, opened 2025-12-05
[PERF] account: optimize migration performance by pre-creating column and bulk transferring invoice_date

royle-vietnam:
Pre-create the target column to avoid triggering compute() during update,
and perform a direct bulk data transfer from account_move to account_move_line.
This avoids ORM overhead, reduces heap rewrites, and prevents unnecessary
recomputation on 15M move lines.

Performance result:
- Previous approach: ~2 hours
- Optimized approach: ~15 minutes (~8x faster)

### DIFF
--- a/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
+++ b/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py
@@ -195,6 +195,26 @@ def _pre_create_early_pay_discount_computation(env):
     )
 
 
+def _pre_account_move_line_invoice_date_computation(env):
+    """Avoid triggering the computed method"""
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE account_move_line
+        ADD COLUMN IF NOT EXISTS invoice_date DATE;
+        """,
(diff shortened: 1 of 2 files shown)


hbrunn (approved):

pedrobaeza (commented):
Are you sure this improves the performance? Odoo already does a SQL for related fields with only one level of depth.

royle-vietnam:
> Are you sure this improves the performance? Odoo already does a SQL for related fields with only one level of depth.

Yes, I'm aware that Odoo ORM already optimizes SQL for related fields with one level of depth.
However, the SQL generated by the ORM still includes all account.move records, including those with `invoice_date IS NULL`.

In my custom SQL, I explicitly exclude moves where invoice_date is NULL, which reduces the dataset before aggregation. This difference helps avoid unnecessary rows being processed and therefore improves performance in our specific case.

So the improvement does not come from bypassing the related field resolution, but from pruning irrelevant records earlier in the query.

pedrobaeza (approved):
I have proposed https://github.com/odoo/odoo/pull/238812, but meanwhile, let's merge it.

PR #5446, opened 2025-11-07
[19.0][MIG] base

hbrunn:
depends:

- [x] review https://github.com/OCA/OpenUpgrade/pull/5395 first, this only makes sense afterwards
- [x] test will only work after a test db is created by https://github.com/OCA/OpenUpgrade/pull/5375

/ocabot migration base

PR #5395, opened 2025-10-03
[19.0] initialize+openupgrade_framework

hbrunn:
My goal here is to reduce the amount of manual work when dealing with OpenUpgrade.

I [proposed](https://github.com/OCA/OpenUpgrade/pull/5394) a somewhat modified script for branch initialization and following steps, which I used for creating this one.

Note that I renamed the migration test script to test-migration.yml, and keep the standard OCA test script for running [unit tests](https://github.com/OCA/OpenUpgrade/commit/be146f948c0a5a2cc09db17dc31832540272fa48). This should make it easier in future versions to determine which patches we still need, and if they're still effective.

You can see the full thing working on https://github.com/hbrunn/OpenUpgrade/pull/18, but I'm not done yet with the base migration, so no PR for this yet.

Meanwhile for making the migration test work as soon as the base module is migrated, please review https://github.com/OCA/OpenUpgrade/pull/5389 and https://github.com/OCA/OpenUpgrade/pull/5375

PR #5375, opened 2025-09-12
[18.0][IMP] add a github workflow that generates test databases used in CI

hbrunn:
In https://github.com/OCA/OpenUpgrade/issues/5370 I claimed that https://github.com/OCA/OpenUpgrade/pull/5355 not failing in CI is because the test databases don't include translations, but that's wrong. Anyways, I spent the morning writing this workflow to have tests databases regenerated regularly, that's a good thing to have in its own right, as there are relevant changes from time to time which might break existing scripts.

You can inspect the results in https://github.com/hbrunn/OpenUpgrade/releases/tag/databases

- [ ] update https://github.com/OCA/OpenUpgrade/blob/documentation/docsource/100_maintain_repository.rst to point to this workflow instead of a non-existing wiki page

Fixes #5370

PR #5466, opened 2025-12-05
[19.0][MIG] mail+dependencies

hbrunn:
depends:

- [x] review https://github.com/OCA/OpenUpgrade/pull/5395 first, this only makes sense afterwards
- [x] test will only work after a test db is created by https://github.com/OCA/OpenUpgrade/pull/5375

/ocabot migration base_setup
/ocabot migration bus
/ocabot migration html_editor
/ocabot migration mail
/ocabot migration web
/ocabot migration web_editor
/ocabot migration web_tour


MiquelRForgeFlow:
please rebase.

docsource/modules180-190.rst
```diff
@@ -1114,13 +1114,13 @@ Module coverage 18.0 -> 19.0
 +---------------------------------------------------+----------------------+-------------------------------------------------+
 | utm                                               |                      |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| web                                               |                      |                                                 |
+| web                                               |Nothing to do         |                                                 |
 +---------------------------------------------------+----------------------+-------------------------------------------------+
-| |del| web_editor                                  |                      |                                                 |
+| |del| web_editor                                  |                      |Merged into html_editor                          |
```
MiquelRForgeFlow, line 1119: FYI: When I have sometime, I will try to do a manual analysis just to update this second column.

openupgrade_scripts/scripts/base_setup/19.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,6 @@
+---Models in module 'base_setup'---
+---Fields in module 'base_setup'---
+---XML records in module 'base_setup'---
+DEL ir.config_parameter: base_setup.setup_default_user_rights (noupdate)
```
MiquelRForgeFlow, line 4: safely delete

openupgrade_scripts/scripts/mail/19.0.1.19/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,299 @@
...
+
+mail         / mail.notification        / failure_type (selection)      : selection_keys added: [mail_bl, mail_dup, mail_optout, mail_spam] (most likely nothing to do)
+
+# NOTHING TO DO
+
+mail         / mail.notification        / mail_email_address (char)     : NEW
+
+# NOTHING TO DO: new feature
+
+mail         / mail.presence            / guest_id (many2one)           : NEW relation: mail.guest
```
MiquelRForgeFlow, line 160: I think this matches with
```
mail         / bus.presence             / guest_id (many2one)           : DEL relation: mail.guest
```

openupgrade_scripts/apriori.py
```diff
@@ -43,6 +42,7 @@
     "pos_self_order_epson_printer": "pos_self_order",
     "sale_async_emails": "sale",
     "website_membership": "website_crm_partner_assign",
+    "web_editor": "html_editor",
```
MiquelRForgeFlow, line 45: why?

hbrunn: because modules that used to depend on web_editor (like mail) now depend on html_editor, not html_builder

pedrobaeza: But the main question to do this is if the resources are attached to this new XML-ID or the old. If there are no records nor resources, then it's the same.

MiquelRForgeFlow: Ok. Fine to merge it to `html_editor`.

openupgrade_scripts/scripts/mail/19.0.1.19/upgrade_analysis_work.txt
hbrunn: but neither of those models matter much because they'll be upserted whenever odoo comes online after the migration

openupgrade_scripts/scripts/base_setup/19.0.1.0/upgrade_analysis_work.txt
hbrunn: done

MiquelRForgeFlow (approved):

PR #5472, opened 2025-12-12
[19.0][MIG] uom

hbrunn:
depends:

- [x] review https://github.com/OCA/OpenUpgrade/pull/5395 first, this only makes sense afterwards
- [x] test will only work after a test db is created by https://github.com/OCA/OpenUpgrade/pull/5375


openupgrade_scripts/scripts/uom/19.0.1.0/noupdate_changes_work.xml
MiquelRForgeFlow, line 1: Is it the same as the `noudpate_changes.xml` or you changed something?

openupgrade_scripts/scripts/uom/19.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,70 @@
...
+DEL ir.ui.view: uom.product_uom_categ_form_view
+DEL ir.ui.view: uom.product_uom_categ_tree_view
+DEL ir.ui.view: uom.uom_categ_view_search
+DEL uom.category: uom.product_uom_categ_energy (noupdate)
+DEL uom.category: uom.product_uom_categ_kgm (noupdate)
+DEL uom.category: uom.product_uom_categ_unit (noupdate)
+DEL uom.category: uom.product_uom_categ_vol (noupdate)
+DEL uom.category: uom.uom_categ_length (noupdate)
+DEL uom.category: uom.uom_categ_surface (noupdate)
+DEL uom.category: uom.uom_categ_wtime (noupdate)
```
MiquelRForgeFlow, line 59: safely delete?

openupgrade_scripts/scripts/uom/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,65 @@
...
+
+        old_reference = uoms.filtered(
+            lambda x, uom2type=uom2type: uom2type[x] == "reference"
+        )
+
+        for uom in uoms - old_reference:
+            if has_xmlid(uom):
+                # uoms with xmlid will be updated by the migration of the module
+                # providing them
+                continue
```
MiquelRForgeFlow: why cannot we have:


```suggestion
        uoms_without_xmlid = some_method(uoms - old_reference)
        for uom in uoms_without_xmlid:
```
?

openupgrade_scripts/scripts/uom/19.0.1.0/noupdate_changes_work.xml
hbrunn: the work file doesn't touch the `active` and `factor` fields

openupgrade_scripts/scripts/uom/19.0.1.0/post-migration.py
hbrunn: what would that change apart from iterating through the list twice?

openupgrade_scripts/scripts/uom/19.0.1.0/upgrade_analysis_work.txt
hbrunn: same as with res.partner.title, this does nothing on nonexisting models

openupgrade_scripts/scripts/uom/19.0.1.0/post-migration.py
MiquelRForgeFlow: I mean, it's to avoid doing a search_count inside the loop, and only do 1 search in total. Only asking if it's feasible (to obtain the best script performance-wise).

openupgrade_scripts/scripts/uom/19.0.1.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Oh, right.

openupgrade_scripts/scripts/uom/19.0.1.0/noupdate_changes_work.xml
MiquelRForgeFlow: Ok. FYI, in my personal taste, I prefer adding those lines commented instead of deleting them. (not a blocker)

openupgrade_scripts/scripts/uom/19.0.1.0/post-migration.py
hbrunn: it's feasible, but pointless imho for the amount of records we're talking about

MiquelRForgeFlow: I prefer the fastest script. We have a client with 30 uoms extra (besides the ones from odoo). If we can improve some running seconds, it's the best!

openupgrade_scripts/scripts/uom/19.0.1.0/noupdate_changes_work.xml
hbrunn: I do it like that myself usually, updated accordingly

openupgrade_scripts/scripts/uom/19.0.1.0/post-migration.py
hbrunn: done

MiquelRForgeFlow (approved):

PR #5476, opened 2025-12-20
Add direct migration scripts v13 → v19

bhanquier:
Supports multiple migration paths:
- v13 → v19 (direct, all-in-one)
- v13 → v14
- v14 → v16 (includes v15)
- v14 → v19
- v15 → v19
- v16 → v19
- v17 → v19
- v18 → v19

### DIFF
--- a/openupgrade_scripts/direct_migration/DIRECT_v13_to_v19.sql
+++ b/openupgrade_scripts/direct_migration/DIRECT_v13_to_v19.sql
@@ -1,0 +1,734 @@
+-- =============================================================================
+-- MIGRATION DIRECTE v13 → v19 (Skip Intermédiaires)
+-- =============================================================================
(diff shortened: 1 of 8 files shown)


bhanquier:
⚠️ **Note about CI failure**: These SQL scripts are standalone migration tools, not Python code. They are meant to be run manually with `psql` after exporting v13 data into a v19 database. The CI failure is expected - these scripts require an existing database with real data to work.

**Tested on**: Real migration from Odoo 13.0 to 19.0 (HBBL production database, December 2024).

Note: These SQL scripts are standalone tools meant to be run manually after data export/import. They are not designed to run in CI - they require an existing database with migrated data. The CI failure is expected.

legalsylvain:
That is not how openupgrade is designed.

PR #5477, opened 2025-12-26
[19.0][MIG] product

hbrunn:
depends:

- [x] https://github.com/OCA/OpenUpgrade/pull/5466
- [x] https://github.com/OCA/OpenUpgrade/pull/5472


openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,121 @@
...
+        AND product_supplierinfo.product_id IS NULL
+        """
+    )
+    env.cr.execute(
+        """
+        UPDATE product_supplierinfo
+        SET product_uom_id=product_template.uom_id
+        FROM product_product, product_template
+        WHERE
+        product_supplierinfo.product_id=product_product.id
```
MiquelRForgeFlow:
```suggestion
        FROM product_template
        WHERE
        product_supplierinfo.product_id IS NOT NULL
```

openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+
+def _prepare_product_packaging_migration(env):
+    """
+    product.packaging is replaced by UOMs being pointed at by records of product.uom
+    Rename product_packaging here to product_uom, and amend with data in post-migration
+    Set uom_id to dummy UOM to have the ORM set up the not null constraint correctly
+    """
+    openupgrade.rename_tables(env.cr, [("product_packaging", "product_uom")])
```
MiquelRForgeFlow, line 77: call `rename_models` too.

openupgrade_scripts/scripts/product/19.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,127 @@
...
+product      / product.product          / is_favorite (boolean)         : is now stored
+
+# DONE: pre-created in pre-migration and filled from product_template
+
+product      / product.product          / is_in_selected_section_of_order (boolean): NEW
+
+# NOTHING TO DO: virtual field only used for searching
+
+product      / product.product          / packaging_ids (one2many)      : DEL relation: product.packaging
+product      / product.product          / product_uom_ids (one2many)    : NEW relation: product.uom
```
MiquelRForgeFlow, line 40: why not rename fields too?

openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,121 @@
...
+        ],
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.delete_records_safely_by_xml_id(
+        env,
+        [
+            "product.constraint_product_packaging_positive_qty",
```
MiquelRForgeFlow: better delete constraints with `delete_sql_constraint_safely`.

openupgrade_scripts/scripts/product/19.0.1.2/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,127 @@
...
+DEL ir.ui.view: product.product_packaging_tree_view2
+
+# NOTHING TO DO
+
+NEW product.category: product.product_category_expenses (noupdate)
+NEW product.category: product.product_category_goods (noupdate)
+NEW product.category: product.product_category_services (noupdate)
+DEL product.category: product.cat_expense (noupdate)
+DEL product.category: product.product_category_1 (noupdate)
+DEL product.category: product.product_category_all (noupdate)
```
MiquelRForgeFlow, line 120: safely remove this one

openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py
hbrunn: applying this suggestion would leave an incorrect sql statement. but I realized we don't need this query at all, and I modified the one above to not be conditional on `product_id` but on `product_uom_id` being `NULL`

openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py
hbrunn: done

openupgrade_scripts/scripts/product/19.0.1.2/pre-migration.py
hbrunn: did that because it doesn't hurt, but what's the purpose of this function given modern odoo drops constraints when deleting the constraint record?

openupgrade_scripts/scripts/product/19.0.1.2/upgrade_analysis_work.txt
hbrunn: good idea

openupgrade_scripts/scripts/product/19.0.1.2/upgrade_analysis_work.txt
hbrunn: I'm renaming this one to `product.product_category_goods`, which only leaves `product.product_category_1` (v18: All/Saleable).

This will be a bad fit to rename to services (the other new xmlid left), because then services would be a subcategory of goods in the migrated database if users kept the original structure.

Deleting categories will be a bad idea I think, because especially with those semi root categories, they tend to contain configuration while not being linked anywhere except their child categories. If we go and unlink that, it will succeed, but users will lose their routes and whatever else you can configure on categories recursively.

MiquelRForgeFlow (approved):

PR #5489, opened 2026-01-21
[19.0][IMP] openupgrade_framework & openupgrade_scripts

MiquelRForgeFlow:
.

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/modules/module_graph.py
+++ b/openupgrade_framework/odoo_patch/odoo/modules/module_graph.py
@@ -20,7 +20,6 @@ def _update_from_database(self, *args, **kwargs) -> None:
 
     if os.environ.get("OPENUPGRADE_USE_DEMO", "") == "yes":
         return
-
     if (
         "base" in self._modules
         and self._modules["base"].demo
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
@@ -3,7 +3,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import logging
 
-from odoo import api
 from odoo.exceptions import ValidationError
 from odoo.modules.module import get_resource_from_path
 from odoo.tools import mute_logger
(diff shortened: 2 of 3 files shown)


openupgrade_framework/odoo_patch/odoo/modules/module_graph.py
```diff
@@ -31,5 +30,40 @@ def _update_from_database(self, *args, **kwargs) -> None:
             module.demo = False
 
 
+def extend(self, names) -> None:
```
MiquelRForgeFlow: @hbrunn I see you deleted several methods in `openupgrade_framework` migration, so I suppose they are not needed anymore. But why you deleted this one? Is it really not needed?

hbrunn: I tested this by adding a new dependency in the manifest of some installed module, and checked that Odoo installs this and its dependencies on `-u all`. What case are you handling here?

And can we add a test for it?

hbrunn: PS: given we can test openupgrade_framework now isolated, it would be good to add a test that fails without this

openupgrade_framework/odoo_patch/odoo/modules/module_graph.py
```diff
@@ -6,9 +6,9 @@
 from odoo.modules.module_graph import ModuleGraph
 
 
-def _update_from_database(self, *args, **kwargs) -> None:
+def _update_from_database(self, names) -> None:
```
hbrunn: why this? my thinking is that this lowers the maintenance burden when we don't care about the arguments anyways

MiquelRForgeFlow: Although it's valid what you said, but in new versions you will still check if the method has the same name and if still it's needed, so I don't see the burden of "adding" a check to see if the arguments are correct.

BTW, if it's a blocker, I can revert it.

openupgrade_framework/odoo_patch/odoo/modules/module_graph.py
MiquelRForgeFlow: Oh, if you tested, then nice. Removed.

openupgrade_framework/odoo_patch/odoo/modules/module_graph.py
hbrunn: I'm less thinking of new versions but Odoo SA tinkering with internals within some version

MiquelRForgeFlow: Ok, reverted.

hbrunn (approved):

PR #5507, opened 2026-02-09
[19.0][FIX] pin setuptools to v81.0

hbrunn:
setuptools finally removed pgk_resources, which some package(s) in Odoo's dependency tree rely on


MiquelRForgeFlow (approved):
LGTM

hbrunn:
ah yes, this happens in its own env. looking into it

PR #5509, opened 2026-02-09
[19.0][MIG] account

hbrunn:
depends:

- [x] https://github.com/OCA/OpenUpgrade/pull/5472
- [x] https://github.com/OCA/OpenUpgrade/pull/5477
- [x] https://github.com/OCA/openupgradelib/pull/440


openupgrade_scripts/scripts/account/19.0.1.4/post-migration.py
```diff
@@ -0,0 +1,213 @@
...
+    for full_reconcile_id, exchange_move_id in env.cr.fetchall():
+        AccountFullReconcile.browse(full_reconcile_id).partial_reconcile_ids[-1:].copy(
+            {
+                "exchange_move_id": exchange_move_id,
+                "full_reconcile_id": full_reconcile_id,
+            }
+        )
+
+
+def account_move_line_is_storno(env):
```
MiquelRForgeFlow: move to end-migration, as point_of_sale also calls `_compute_is_storno`.

openupgrade_scripts/scripts/account/19.0.1.4/post-migration.py
```diff
@@ -0,0 +1,213 @@
...
+                                "partner_id": partner_id,
+                            }
+                        ),
+                    ],
+                }
+            )
+            env.cr.execute(
+                f"""
+                UPDATE account_reconcile_model
+                SET {link_column} = {mapping_id}
```
MiquelRForgeFlow, line 102: what is the use of this link_column field?

hbrunn: to know which model was created by a partner mapping. whenever the oca version of the reconciliation is done, we'll probably want to undo this, or differently, so it seems prudent to me to know where this comes from

openupgrade_scripts/scripts/account/19.0.1.4/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,386 @@
...
+DEL ir.ui.view: account.product_template_view_tree
+DEL ir.ui.view: account.view_account_reconcile_model_line_form
+DEL ir.ui.view: account.view_in_invoice_receipt_tree
+
+# NOTHING TO DO
+
+NEW mail.template: account.email_template_edi_self_billing_credit_note [renamed from account_peppol_selfbilling module] (noupdate)
+NEW mail.template: account.email_template_edi_self_billing_invoice [renamed from account_peppol_selfbilling module] (noupdate)
+
+# NOTHING TO DO: the templates have changed, so if we renamed this, we'd have to overwrite them again
```
MiquelRForgeFlow: If you don't want to do the rename, then at least safely remove the old ones.

hbrunn: I verified those were only whitespace changes after all, so just renaming is fine

MiquelRForgeFlow (approved):

PR #5511, opened 2026-02-10
[17.0][OU-IMP] hr: Private data use mobile

pedrobaeza:
On the private phone, we transfer the phone from the partner, and if
not filled, the mobile.

@Tecnativa TT60883

### DIFF
--- a/openupgrade_scripts/scripts/hr/17.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/hr/17.0.1.1/post-migration.py
@@ -28,7 +28,9 @@ def _transfer_employee_private_data(env):
                 he.private_country_id, rpp.country_id, rp.country_id
             ),
             private_email = COALESCE(he.private_email, rpp.email, rp.email),
-            private_phone = COALESCE(he.private_phone, rpp.phone, rp.phone),
+            private_phone = COALESCE(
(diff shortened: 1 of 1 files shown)


hbrunn (changes requested):
it's not guaranteed that `address_home_id` points to a private address, so we need the fallback to that.

and why don't you map mobile to the mobile field?

pedrobaeza:
> it's not guaranteed that address_home_id points to a private address, so we need the fallback to that.

both tables, `res_partner` and `ou_res_partner_private` are linked through `address_home_id`, so if it's empty, neither of both will be used. In fact, it won't be part of this update.

> and why don't you map mobile to the mobile field?

There's no mobile specific field in the employee. Just `private_phone` field.

hbrunn:
right with the mobile, I was thinking about partners.

I don't get your reply about the partners though. If address_home_id links to partner with id 42 which is not a private address, why wouldn't we want to use that data? Mind you ou_res_partner_private only contains partners that had type == private

pedrobaeza:
OK, now I get you. It's very unlikely that a home address is not private, but let's keep that possibility. Now I have just adjusted to get phone, mobile from `ou_res_partner_private`, and if not filled, the same from the partner.

hbrunn (approved):

PR #5515, opened 2026-02-10
[18.0][OU-ADD] pos_discount: Nothing to do

david-banon-tecnativa:
`point_of_sale.product_product_consumable` got moved to this module as `pos_discount.product_product_consumable`, but changing xmlid is already taken care of in point_of_sale scripts, so nothing to do here
TT59139 @tecnativa


openupgrade_scripts/scripts/pos_discount/18.0.1.0/upgrade_analysis.txt
```diff
@@ -2,3 +2,4 @@
 ---Fields in module 'pos_discount'---
 ---XML records in module 'pos_discount'---
 NEW product.product: pos_discount.product_product_consumable [renamed from point_of_sale module] (noupdate)
+# Nothing to do as xmlid is alredy beeing taken care of in point_of_sale scripts
```
pedrobaeza: We use this convention, and I have rephrased to:

```suggestion
# NOTHING TO DO: As XML-ID renaming is already taken into account in point_of_sale scripts
```

pedrobaeza (approved):

openupgrade_scripts/scripts/pos_discount/18.0.1.0/upgrade_analysis.txt
hbrunn: this should have gone to `upgrade_analysis_work.txt`, see https://github.com/OCA/OpenUpgrade/pull/5519/changes#diff-c2034d9f8edeb109b86f449319e48a96aedb8712149267e4aebf431b8cd18c5e

pedrobaeza: Ouch, true. My fault not detecting it in the review...

pedrobaeza:
Fixed in #5520

PR #5522, opened 2026-02-17
[18.0][OU-IMP] mail: Define the appropriate delay_from value (similar to v17)

victoralmau:
Define the appropriate delay_from value (similar to v17)

Related to https://github.com/OCA/OpenUpgrade/pull/5506/changes/a39f30352d48bd352aa983d50e3455081a2236d1

@Tecnativa TT61045

### DIFF
--- a/openupgrade_scripts/scripts/mail/18.0.1.18/post-migration.py
+++ b/openupgrade_scripts/scripts/mail/18.0.1.18/post-migration.py
@@ -71,12 +71,17 @@ def mail_activity_plan_template_delay(env):
     """
     Define the appropriate values for delay_count and delay_unit as
     defined in mail_activity_type to have the same behavior as in v17.
+    All records are defined with delay_from=after_plan_date because
+    that was the behavior up to v17.
     """
     openupgrade.logged_query(
         env.cr,
         """
         UPDATE mail_activity_plan_template t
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):
No idea, but LGTM

openupgrade_scripts/scripts/mail/18.0.1.18/post-migration.py
```diff
@@ -71,12 +71,17 @@ def mail_activity_plan_template_delay(env):
...
     """
     openupgrade.logged_query(
         env.cr,
         """
         UPDATE mail_activity_plan_template t
-        SET delay_count = at.delay_count, delay_unit = at.delay_unit
+        SET
+            delay_count = at.delay_count,
+            delay_unit = at.delay_unit,
+            delay_from = 'after_plan_date'
```
pedrobaeza, line 84: Is it a new field? If so, it should be in `openupgrade_analysis.txt` and should be reflected in this PR in `openupgrade_analysis_work.txt`.

pedrobaeza (approved):
For the next time, we usually put the `_work` file with the format `DONE: post/pre/end-migration: ...`, to know where to look, although I see that this file is not complying with this guideline, but let's not block this more.

PR #5526, opened 2026-02-23
[19.0][MIG] payment

hbrunn:
depends:

- [x] https://github.com/OCA/server-tools/pull/3535


openupgrade_scripts/scripts/payment/19.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,73 @@
...
+# NOTHING TO DO
+
+---XML records in module 'payment'---
+NEW ir.actions.server: payment.action_start_payment_onboarding
+DEL ir.actions.server: payment.action_activate_stripe
+DEL ir.model.access: payment.payment_provider_onboarding_wizard
+NEW ir.ui.view: payment.payment_transaction_graph
+NEW ir.ui.view: payment.payment_transaction_pivot
+DEL ir.ui.view: payment.payment_provider_onboarding_wizard_form
+DEL onboarding.onboarding.step: payment.onboarding_onboarding_step_payment_provider (noupdate)
```
MiquelRForgeFlow, line 29: safely delete

openupgrade_scripts/scripts/payment/19.0.2.0/noupdate_changes_work.xml
```diff
@@ -0,0 +1,471 @@
...
+    <field name="support_manual_capture">none</field>
+  </record>
+  <record id="payment_method_bank_reference" model="payment.method">
+    <field name="support_manual_capture">none</field>
+  </record>
+  <record id="payment_method_bank_transfer" model="payment.method">
+    <!--
+    this fails on loading
+    <field name="supported_currency_ids"/>
+    -->
```
MiquelRForgeFlow: call we look into why it fails? maybe we should do something like this:
```suggestion
    <!-- manually edited -->
    <field name="supported_currency_ids" eval="[Command.set([])]"/>
```
or force it it someway.

hbrunn: see https://github.com/OCA/server-tools/pull/3535, I'm relying on this patch now and load the generated file

openupgrade_scripts/scripts/payment/19.0.2.0/upgrade_analysis_work.txt
hbrunn: done, somehow was confused about this model being deprecated last week

MiquelRForgeFlow (approved):

Issue #5491, opened 2026-01-23
[17.0] l10n_fr UniqueViolation account_report_expression_line_label_uniq tax_report_data.xml


Lukwos:
Since this PR (https://github.com/odoo/odoo/pull/203227), the account.report.expression are explicitely set and when I run the migration from 16.0 to 17.0, I get a contraint error : 
```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "account_report_expression_line_label_uniq"
DETAIL:  Key (report_line_id, label)=(165, balance) already exists.
```

## Module
l10n_fr

remi-filament:
Thanks for reporting this @Lukwos I just faced the exact same issue.
They removed aggregation_formula which was using an inverse to compute expressions and replaced those by well defined expressions.

The problem is that these expressions did not have an XML-id and are therefore not updated but recreated instead leading to hitting unicity constraint.

Since these entries are not noupdate="1", it seems safe to remove them before Odoo will automatically recreate them properly.

The following SQL script does the job and should be added to l10n_fr migration scripts : 
```sql
DELETE FROM account_report_expression 
WHERE label='balance' 
    AND report_line_id IN (
        SELECT res_id FROM ir_model_data
        WHERE module='l10n_fr'
        AND name in ('tax_report_16', 'tax_report_23', 'tax_report_TIC_total', 'tax_report_X4', 'tax_report_Y1', 'tax_report_Y2', 'tax_report_Y3', 'tax_report_Z4', 'tax_report_32')
        AND model='account.report.line')
    AND id NOT IN (
        SELECT res_id FROM ir_model_data WHERE module='l10n_fr' AND model='account.report.expression'
);
```

remi-filament:
Should be fixed with #5533 which has just been merged !
Would you mind testing it @Lukwos and closing this issue if fixed ?
Best Regards,

eLvErDe:
Hello,

`l10n_lu` is failing with a similar error:

```
2026-03-30 21:32:01,362 1 ERROR odoo odoo.sql_db: bad query: INSERT INTO "account_report_expression" ("create_date", "create_uid", "date_scope", "engine", "formula", "green_on_positive", "label", "report_line_id
[...]

PR #5533, opened 2026-02-23
[17.0][OU-ADD] l10n_fr: Migration scripts

remi-filament:
Migration scripts for l10n_fr
Nothing to do for other l10n_fr modules.


remi-filament:
I have a question not directly related to this PR regarding l10n_fr_hr_holidays (new module added in v17 with auto_install if you have both l10n_fr and hr_holidays installed).
There is a new parameter added in res_company and related one in res.config_settings that I would like to initialize. Since this module cannot be installed in v16, how can I make sure that my migration scripts are run during v17 migration ? Where should I define those scripts ?

I am pretty sure you have the answer @hbrunn or @MiquelRForgeFlow ?
Thanks !

hbrunn:
> I am pretty sure you have the answer @hbrunn or @MiquelRForgeFlow ? Thanks !

just write a migration script for the new module, for exactly this reason we patch Odoo to also run migration scripts for newly installed modules. If that doesn't work, that's a but in openupgrade_framework

remi-filament:
Hi @hbrunn thank you for your answer.
I just published a PR with script for l10n_fr_hr_holidays, although script is never run. We should probably continue the conversation in the other PR !
https://github.com/OCA/OpenUpgrade/pull/5539

remi-filament:
Fixes https://github.com/OCA/OpenUpgrade/issues/5491

flotho (approved):
successfully tested in a migration
LGTM
Thanks

hparfr (approved):
Thanks @remi-filament 

Tested in a CI.

PR #5551, opened 2026-03-04
[17.0][FIX] account: Only set res_field=invoice_pdf_report_file on customer invoice/refunds

alexis-via:
Current behavior: On ir.attachment of invoices, 'res_field' is set to 'invoice_pdf_report_file' on all invoices that have 'message_main_attachment_id'
New behavior: only do that for customer invoices and refunds

Why ? Because supplier invoices/refunds can also have 'message_main_attachment_id', for example when you click on the "Upoad" button of the purchase journal on the accounting dashboard and you upload a PDF file. And we don't want res_field='invoice_pdf_report_file' on attachments of supplier invoice/refunds !


pedrobaeza (approved):

pedrobaeza:
The effect on this is that if the invoice is reset to draft, the attachment is auto-removed.

PR #5547, opened 2026-03-02
[19.0][MIG] stock, stock_account

hbrunn:
depends:

- [x] https://github.com/OCA/openupgradelib/pull/441

PR #5555, opened 2026-03-09
[19.0][MIG] sale_stock

hbrunn:
depends:

- [x] https://github.com/OCA/OpenUpgrade/pull/5547
- [x] https://github.com/OCA/openupgradelib/pull/442


openupgrade_scripts/scripts/sale_stock/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+    openupgrade.m2o_to_x2m(
+        env.cr,
+        env["sale.order"],
+        "sale_order",
+        "stock_reference_ids",
+        "procurement_group_id",
+    )
+    openupgrade.m2o_to_x2m(
+        env.cr, env["sale.order.line"], "sale_order_line", "route_ids", "route_id"
+    )
```
MiquelRForgeFlow, line 19:
```python
openupgrade.m2o_to_x2m(
        env.cr,
        env["stock.reference"],
        "stock_reference",
        "sale_ids",
        "sale_id",
    )
```

Is it missing?

openupgrade_scripts/scripts/sale_stock/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
...
+        "sale.order.line",
+        "route_id",
+        "route_ids",
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_field_references(env, _renamed_field_references)
```
MiquelRForgeFlow, line 27: why not copy_columns + rename_fields?

openupgrade_scripts/scripts/sale_stock/19.0.1.0/post-migration.py
hbrunn: no, that's the inverse of sale.order#procurement_group_id

openupgrade_scripts/scripts/sale_stock/19.0.1.0/pre-migration.py
hbrunn: this would create a column named ..._ids which we don't need or want

MiquelRForgeFlow (approved):
LGTM

PR #5558, opened 2026-03-16
[19.0][MIG] mrp

hbrunn:


openupgrade_scripts/scripts/mrp/19.0.2.0/post-migration.py
```diff
@@ -0,0 +1,80 @@
...
+
+from openupgradelib import openupgrade
+
+
+def mrp_production_group(env):
+    """
+    Create production groups from procurement groups
+    """
+    link_column = openupgrade.get_legacy_name("procurement_group_id")
+    env.cr.execute(
```
MiquelRForgeFlow, line 12: Not telling you to change it, but FYI: `add_columns()` exists...

openupgrade_scripts/scripts/mrp/19.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+mrp          / mrp.production.group     / parent_ids (many2many)        : NEW relation: mrp.production.group
+
+# NOTHING TO DO
+
+mrp          / mrp.production.group     / production_ids (one2many)     : NEW relation: mrp.production
+
+# DONE: see mrp.production#production_group_id
+
+mrp          / mrp.routing.workcenter   / cost_mode (selection)         : NEW selection_keys: ['actual', 'estimated'], hasdefault: default
+mrp          / mrp.routing.workcenter   / note (html)                   : DEL
```
MiquelRForgeFlow, line 53: why not save this field (with `copy`)?

openupgrade_scripts/scripts/mrp/19.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,145 @@
...
+mrp          / mrp.unbuild              / lot_id (many2one)             : not a function anymore
+mrp          / mrp.workcenter           / default_capacity (float)      : DEL
+
+# NOTHING TO DO
+
+mrp          / mrp.workcenter.capacity  / product_uom_id (many2one)     : is now stored
+mrp          / mrp.workcenter.capacity  / product_uom_id (many2one)     : not related anymore
+mrp          / mrp.workcenter.capacity  / product_uom_id (many2one)     : now required
+
+# NOTHING TO DO: was related to product_id.uom_id and product_id was required
```
MiquelRForgeFlow: then, shouldn't this be filled with `product_id.uom_id` or `mrp.workorder`.`product_uom_id`?

openupgrade_scripts/scripts/mrp/19.0.2.0/post-migration.py
hbrunn: yes, but that one breaks if the column exists. we should fix that eventually for having idempotent migration scripts at some point

openupgrade_scripts/scripts/mrp/19.0.2.0/upgrade_analysis_work.txt
hbrunn: why save that one as opposed to any of the others? in my opinion, we copy columns when we modify them in a potentially lossy way, but leave all the others alone, no?

openupgrade_scripts/scripts/mrp/19.0.2.0/upgrade_analysis_work.txt
hbrunn: i'm now precreating the field and do the compute function in sql

openupgrade_scripts/scripts/mrp/19.0.2.0/upgrade_analysis_work.txt
MiquelRForgeFlow: Yes. It's just that `note` seems to be a column one may want to save, it can have relevant content.

MiquelRForgeFlow (approved):

PR #5559, opened 2026-03-16
[19.0][MIG] purchase

hbrunn:


openupgrade_scripts/scripts/purchase/19.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2026 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+renamed_fields = [
+    ("purchase.order", "purchase_order", "note", "notes"),
```
MiquelRForgeFlow:
```suggestion
    ("purchase.order", "purchase_order", "notes", "note"),
```

openupgrade_scripts/scripts/purchase_stock/19.0.1.2/post-migration.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2026 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "purchase_stock", "19.0.1.2/noupdate_changes.xml")
+    openupgrade.m2o_to_x2m(
```
MiquelRForgeFlow: could you change the order of calling these methods?

MiquelRForgeFlow (commented):
~~Also add `sale_purchase_stock`.~~

openupgrade_scripts/scripts/purchase/19.0.1.2/pre-migration.py
hbrunn: ho thanks. now here I'd expect OU to raise, but we probably had a reason for it not to

openupgrade_scripts/scripts/purchase_stock/19.0.1.2/post-migration.py
hbrunn: did so, but why?

MiquelRForgeFlow: You never know what the noupdate_changes file is touching. And perhaps it is touching the `reference_ids`.

MiquelRForgeFlow (approved):

PR #5562, opened 2026-03-19
[MIG] sale_mrp_link merged into sale_mrp.

legalsylvain:
Ref : https://github.com/odoo/odoo/pull/43366
commit : https://github.com/odoo/odoo/commit/e1b3a17c3bda2b2d585d930386244e3f2b0bed31

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -128,6 +128,7 @@ merged_modules = {
     # OCA/sale-reporting
     "report_qweb_pdf_fixed_column": "web",
     # OCA/sale-workflow
+    "sale_mrp_link": "sale_mrp",
     "sale_order_price_recalculation": "sale",
     "sale_order_pricelist_tracking": "sale",
     # OCA/stock-logistics-warehouse


MiquelRForgeFlow:
Referenced commit adds `sale_order_count` to `mrp.production`, but `sale_mrp_link` adds `sale_order_id` to `mrp.production`. I think it's not the same.

legalsylvain:
> Referenced commit adds sale_order_count to mrp.production, but sale_mrp_link adds sale_order_id to mrp.production. I think it's not the same.

I'm not expert in mrp features. What I understand : 
- The "feature" has been merged. because the objective of the sale_mrp_link is to provide form links between sale order and mrp production.
- the implementation is not the same, because before, the link between sale order and mrp production was M2O, and in odoo 14.0 it's a M2M. So indeed, the computation is different, and the sale_order_id is now obsolete, I think.

hbrunn (approved):
I think sale_mrp does everything sale_mrp_link does, even though slightly differently, so agreed with merging this.

Left a review in https://github.com/OCA/sale-workflow/pull/4080#pullrequestreview-3989530608 to that effect too

pedrobaeza (approved):

MiquelRForgeFlow:
A retry is needed here...

MiquelRForgeFlow:
> the implementation is not the same, because before, the link between sale order and mrp production was M2O, and in odoo 14.0 it's a M2M. So indeed, the computation is different, and the sale_order_id is now obsolete, I think.

Shouldn't we fill that M2M with `mrp_production.sale_order_id`?

PR #5565, opened 2026-03-23
[19.0][MIG] point_of_sale

hbrunn:


hbrunn:
CI will keep failing until we have migrated all modules mentioned in https://github.com/odoo/odoo/blob/19.0/addons/point_of_sale/models/pos_config.py#L159 ff. Not much we can do about that

hbrunn:
CI succeeds now some pos_* modules are migrated

MiquelRForgeFlow (approved):

PR #5568, opened 2026-03-26
[16.0][OU-FIX] payment_paypal: We should unset the value also on providers with paypal code

etobella:
That might happen if you have duplicated the paypal provider for another company for example


pedrobaeza (approved):
/ocabot merge nobump

PR #5569, opened 2026-03-27
[18.0][OU-FIX] website_mass_mailing: Manage changes on newsletter

etobella:
When migrating from 17 to 18, I found out that the snippets for newsletter have changed and an error was raised because it was expecting an element that doesn't exist.

This fixes it, but maybe you know a better way for handling this :wink: 

On 17
https://github.com/odoo/odoo/blob/17.0/addons/website_mass_mailing/views/snippets_templates.xml#L24

On 18
https://github.com/odoo/odoo/blob/18.0/addons/website_mass_mailing/views/snippets_templates.xml#L25

### DIFF
--- a/openupgrade_scripts/scripts/website_mass_mailing/18.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website_mass_mailing/18.0.1.0/post-migration.py
@@ -1,0 +1,35 @@
+# Copyright 2026 Dixmit Consulting
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from lxml import etree
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    new_snippet = env.ref("website_mass_mailing.s_newsletter_subscribe_form")
+    views = env["ir.ui.view"].search(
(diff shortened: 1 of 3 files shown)


pedrobaeza:
On openupgradelib.openupgrade_tools, we have several methods like `convert_html_fragment` or `convert_xml_node` that may help. You can check a example in https://github.com/OCA/OpenUpgrade/blob/12.0/addons/website_blog/migrations/12.0.1.0/post-migration.py#L40, but not sure if it totally fits the need.

etobella:
No, it doesn't feed the need as we are replacing an item and that is not an options with that functions, however, the idea can help in order to clean up the code

etobella:
@pedrobaeza I think now it is cleaner :wink:

pedrobaeza (commented):
Isn't this replacing possible custom texts?

etobella:
Might be and this is something that I might need to review, but at least the system will not fail :cry:

pedrobaeza (approved):

pedrobaeza:
Leftover in https://github.com/OCA/OpenUpgrade/commit/f0b9aac1db530e3b8f50295533c2c1d7097121f0

PR #5572, opened 2026-03-30
[19.0][MIG] project + dependencies

hbrunn:
I'm merging project_task_add_very_high into project and propose to close the PR for migrating this to v19: https://github.com/OCA/project/pull/1672


openupgrade_scripts/scripts/project/19.0.1.3/post-migration.py
```diff
@@ -0,0 +1,84 @@
...
+            rating_status, rating_status_period
+            FROM project_project
+            WHERE id in %s
+            ORDER BY write_date DESC
+            LIMIT 1
+            """,
+            (tuple(task_type.project_ids.ids),),
+        )
+        for vals in env.cr.dictfetchall():
+            task_type.write(vals)
```
MiquelRForgeFlow, line 45: Please, could you transform all this in one query?

hbrunn: I'd rather not, because that would mean rebuilding https://github.com/OCA/OCB/blob/19.0/addons/project/models/project_task_type.py#L116-L124 and keeping it in sync with whatever changes there. Don't see this as much of a performance issue as there won't be many stages

MiquelRForgeFlow (approved):
Please rebase.

hbrunn:
done

PR #5577, opened 2026-04-06
[18.0][IMP] apriori: sale_commission_product_criteria_semaphore -> sale_commission_oca_product_criteria_semaphore

CarlosRoca13:
cc @Tecnativa TT61696

This PR depends of:

- [ ] https://github.com/OCA/commission/pull/672

ping @pedrobaeza @david-banon-tecnativa

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -18,6 +18,7 @@ renamed_modules = {
     "commission_formula": "commission_formula_oca",
     "hr_commission": "hr_commission_oca",
     "sale_commission": "sale_commission_oca",
+    "sale_commission_product_criteria": "sale_commission_oca_product_criteria",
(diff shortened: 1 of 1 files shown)


david-banon-tecnativa:
Can you also add `sale_commission_product_criteria` -> `sale_commission_oca_product_criteria`?

CarlosRoca13:
Sure

CarlosRoca13:
Done @david-banon-tecnativa

david-banon-tecnativa (approved):
LGTM I don't think the test failures are related to this PR

pedrobaeza:
The idea behind the `_oca` suffix is to go at the end. I see that one module was already merged, but I'm not even sure if this should have the suffix at all, as the intention is to avoid collisions with Odoo core names, but in this case, it's very unlikely. For example, `sale_commission_salesman` hasn't been renamed.

pedrobaeza:
As the module is merged, let's keep it as is.

pedrobaeza (approved):

PR #5584, opened 2026-04-13
[19.0][FIX] website: make sure website.snippets and website.snippet_options are not deleted

hbrunn:
Central snippet related qweb templates were provided by web_editor v18, and website
inherited from them. Now with web_editor being merged into html_editor, we need
to disentangle those templates from their parents, as in v19, website is where
snippets are introduced, and they would be deleted if inheritance wasn't removed
during cleanup of xmlids


MiquelRForgeFlow (approved):

MiquelRForgeFlow:
Is the PR finished?

hbrunn:
yes

PR #5588, opened 2026-04-16
[19.0][OU-IMP] sale_order_warn_message: Merge into sale

Andrii9090-tecnativa:
OCA module [sale_order_warn_message](https://github.com/OCA/sale-workflow/tree/18.0/sale_order_warn_message) got its functionality implemented in Odoo core.

https://github.com/odoo/odoo/blob/19.0/addons/sale/models/sale_order.py#L828

@pedrobaeza ping

@Tecnativa
TT61963

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -46,7 +46,8 @@ merged_modules = {
     # odoo/enterprise
     # OCA/project
     "project_task_add_very_high": "project",
-    # OCA/...
+    # OCA/sale-workflow
+    "sale_order_warn_message": "sale",
 }
 
 # only used here for upgrade_analysis


openupgrade_scripts/apriori.py
```diff
@@ -41,11 +41,10 @@
     "pos_epson_printer": "point_of_sale",
     "pos_self_order_epson_printer": "pos_self_order",
     "sale_async_emails": "sale",
+    "sale_order_warn_message": "sale",
```
pedrobaeza: This shouldn't go into this sections, as it's not an Odoo module. Being the first OCA module merged in this version, you can replace `# OCA/...` by `# OCA/sale-workflow`, and put this merging there.

pedrobaeza (approved):

PR #5591, opened 2026-04-16
[17.0][IMP] apriori: rename mail_template_attachment_i18n

MiquelRForgeFlow:
https://github.com/OCA/server-tools/pull/3559.


openupgrade_scripts/apriori.py
```diff
@@ -14,6 +14,8 @@
     "l10n_es_pos": "l10n_es_pos_oca",
     # OCA/social
     "mail_activity_unlink_log": "mail_activity_cancel_tracking",
+    # OCA/server-tools
```
pedrobaeza, line 15: This should go before OCA/social per alphabetical order.

pedrobaeza (approved):

PR #5598, opened 2026-04-20
[19.0][MIG] hr_holidays

hbrunn:


MiquelRForgeFlow (approved):

pedrobaeza:
This one depends on `calendar`, and that one seems to not be done.

MiquelRForgeFlow:
> This one depends on calendar, and that one seems to not be done.

Done in https://github.com/OCA/OpenUpgrade/pull/5613.

PR #5600, opened 2026-04-23
[17.0][FIX] stock: prefill picked for done moves to avoid massive rec…

aneavanzosc:
…ompute

### DIFF
--- a/openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
@@ -38,45 +38,37 @@ def fix_move_line_quantity(env):
 
 def prefill_picked(env):
     """
-    Pre-fill picked for done stock moves and their move lines
-    to reduce ORM recomputation during migration.
+    Pre-fill picked for stock moves and move lines to reduce ORM
+    recomputation during migration.
     """
-
-    openupgrade.logged_query(
-        env.cr,
-        """
-        ALTER TABLE stock_move
-        ADD COLUMN IF NOT EXISTS picked boolean
-        """,
+    openupgrade.add_columns(
+        env,
+        [
+            ("stock.move", "picked", "boolean", True, "stock_move"),
+            ("stock.move.line", "picked", "boolean", True, "stock_move_line"),
+        ],
     )
-
-    openupgrade.logged_query(
-        env.cr,
-        """
-        ALTER TABLE stock_move_line
-        ADD COLUMN IF NOT EXISTS picked boolean
-        """,
-    )
-
     openupgrade.logged_query(
         env.cr,
         """
         UPDATE stock_move_line sml
-        SET picked = TRUE
+        SET picked = FALSE
         FROM stock_move sm
         WHERE sml.move_id = sm.id
-          AND sm.state = 'done'
-          AND (sml.picked IS DISTINCT FROM TRUE)
+          AND sm.state != 'done'
         """,
     )
-
     openupgrade.logged_query(
         env.cr,
         """
-        UPDATE stock_move
-        SET picked = TRUE
-        WHERE state = 'done'
-          AND (picked IS DISTINCT FROM TRUE)
+        UPDATE stock_move sm
+        SET picked = False
+        FROM stock_move sm2
+        LEFT JOIN stock_move_line sml
+            ON sml.move_id = sm2.id
+        WHERE sm.state NOT IN ('done', 'cancel')
+            AND sml.id IS NULL
+            AND sm.id = sm2.id
         """,
     )
 
--- a/openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
@@ -35,8 +35,8 @@ def fix_move_line_quantity(env):
         """,
     )
 
-def prefill_picked(env):
 
+def prefill_picked(env):
     """
     Pre-fill picked for done stock moves and their move lines
     to reduce ORM recomputation during migration.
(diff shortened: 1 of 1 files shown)


anajuaristi (approved):

hbrunn (commented):
assuming there are much more done lines than others, you probably get a speedup for creating the column with default true and setting it to false where state isn't done.

[openupgrade.add_columns](https://oca.github.io/openupgradelib/API.html#openupgradelib.openupgrade.add_columns) makes this simple to do

anajuaristi:
@hbrunn
Hi Holger. IMHO your assumption is not correct. 
It must be by default false, since the picket field means that the move and move line has been picked (quant assigned) by default will be always false.
The only thing we know on migration is that "DONE" moves are picked always. This is a fact. If you set by default = true, it's possible that some moves stay in a picked status that it's not true. 
For me it's much more consistent if we do as we proposed the first time. The performance issues are solved with this aproach. We have tested in a 5million move and more movelines environment and it works perfectly.

I do not aprove the second aproache.
Please let me know.
Thank you!

openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
```diff
@@ -38,45 +38,42 @@ def fix_move_line_quantity(env):
...
-          AND (picked IS DISTINCT FROM TRUE)
+        UPDATE stock_move sm
+        SET picked = FALSE
+        WHERE sm.state != 'done'
+          AND NOT EXISTS (
+              SELECT 1
+              FROM stock_move_line sml
+              WHERE sml.move_id = sm.id
+                AND sml.picked = TRUE
+          )
```
hbrunn: this can better be a left join

remi-filament:
> @hbrunn Hi Holger. IMHO your assumption is not correct. It must be by default false, since the picket field means that the move and move line has been picked (quant assigned) by default will be always false. The only thing we know on migration is that "DONE" moves are picked always. This is a fact. If you set by default = true, it's possible that some moves stay in a picked status that it's not true. For me it's much more consistent if we do as we proposed the first time. The performance issues are solved with this aproach. We have tested in a 5million move and more movelines environment and it works perfectly.
> 
> I do not aprove the second aproache. Please let me know. Thank you!

Hi @anajuaristi I think Holger did not mean that default should be set to True after migration, but rather that in order to reduce compte time (which is the purpose of your PR), since on existing database there is much likely more done than on-going moves, it could be even faster to create the field everywhere with a True value, then revert default to False (for records created after migration) and also set back to False with a SQL query only the on-going moves.

You have an exemple here (done in SQL though, not using openupgradelib) : https://github.com/OCA/OpenUpgrade/blob/f9f400666c2a2ac9559613ef6a2addcb5a685e54/openupgrade_scripts/scripts/account/17.0.1.2/pre-migration.py#L251

openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
```diff
@@ -36,8 +36,65 @@ def fix_move_line_quantity(env):
...
+        """,
+    )
+
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE stock_move_line ALTER COLUMN picked DROP DEFAULT
+        """,
+    )
+
```
pedrobaeza: The method already drops the defaults.

```suggestion
```

And please don't introduce empty lines inside the method.

openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
```diff
@@ -36,8 +36,65 @@ def fix_move_line_quantity(env):
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move_line sml
+        SET picked = FALSE
+        FROM stock_move sm
+        WHERE sml.move_id = sm.id
+          AND sm.state != 'done'
+          AND sml.picked IS DISTINCT FROM FALSE
```
pedrobaeza: As you have just created the column, this is not adding value and increasing computation time.

```suggestion
```

openupgrade_scripts/scripts/stock/17.0.1.1/pre-migration.py
```diff
@@ -36,8 +36,65 @@ def fix_move_line_quantity(env):
...
+        """
+        UPDATE stock_move sm
+        SET picked = FALSE
+        WHERE sm.state != 'done'
+          AND NOT EXISTS (
+              SELECT 1
+              FROM stock_move_line sml
+              WHERE sml.move_id = sm.id
+                AND sml.picked = TRUE
+          )
```
pedrobaeza: This will exclude automatically those not having smls:

```suggestion
        UPDATE stock_move sm
        SET picked = False
        FROM stock_move_line sml
        WHERE sml.move_id = sm.id
```

pedrobaeza: Sorry, my mistake for this. You need the contrary operation, so it would be:

```suggestion
        UPDATE stock_move sm
        SET picked = False
        FROM stock_move sm2
        LEFT JOIN stock_move_line sml
            ON sml.move_id = sm2.id
        WHERE sm.state NOT IN ('done', 'cancel')
            AND sml.id IS NULL
            AND sm.id = sm2.id
```

I have included the `cancel` state, as in my tests, they are marked as picked.

pedrobaeza: Old times in a test DB:

Image

New times in the same DB:

Image

pedrobaeza (approved):

anajuaristi (approved):

PR #5605, opened 2026-04-27
[19.0][IMP] hr: set contract start date to employee's create date

hbrunn:
this one is more an rfc: When migrating a database without hr_contract, all employees end up without a contract start date. Any code looking for active versions will treat those employees as inactive, so it might be a good idea to set some date.

On the other hand, this will most likely be a wrong date, so that's an argument for not doing it and having HR managers fill in the correct date afterwards.

### DIFF
--- a/openupgrade_scripts/scripts/hr/19.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/hr/19.0.1.1/post-migration.py
@@ -68,12 +68,12 @@ def fill_hr_version(env):
         f"""
         INSERT INTO hr_version
         (
-            employee_id, date_version,
+            employee_id, date_version, contract_date_start,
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
I think it's better to not have them deactivated, although the date is not correct. I suppose this date is modified having `hr_contract` installed, isn't it?

hbrunn:
> I think it's better to not have them deactivated, although the date is not correct. I suppose this date is modified having `hr_contract` installed, isn't it?

yes. So they aren't deactivated in the active=False sense, but whenever something date dependent happens, https://github.com/odoo/odoo/blob/19.0/addons/hr/models/hr_employee.py#L1622 is called, which excludes employees without start date

PR #5608, opened 2026-05-01
[14.0][FIX] avoid overwriting uom rounding

huguesdk:
update noupdate_changes.xml and avoid overwriting uom rounding values.

the values of the `rounding` field were present for `product_uom_unit` and `product_uom_kgm` in the `uom` module data at some point, but [this commit](https://github.com/OCA/OCB/commit/460ec0402a2352a5b179d81935f7230f6bc97cb7) removed them.

### DIFF
--- a/openupgrade_scripts/scripts/uom/14.0.1.0/noupdate_changes.xml
+++ b/openupgrade_scripts/scripts/uom/14.0.1.0/noupdate_changes.xml
@@ -18,7 +18,7 @@
     <field name="name">in</field>
   </record>
   <record id="product_uom_kgm" model="uom.uom">
-    <field name="rounding" eval="0.01"/>
+    <!-- <field name="rounding"/> -->
     <!-- <field name="uom_type">reference</field> -->
   </record>
   <record id="product_uom_lb" model="uom.uom">
@@ -43,7 +43,7 @@
     <field name="name">qt (US)</field>
   </record>
   <record id="product_uom_unit" model="uom.uom">
-    <field name="rounding" eval="0.01"/>
+    <!-- <field name="rounding"/> -->
     <!-- <field name="uom_type">reference</field> -->
   </record>
 </odoo>
--- a/openupgrade_scripts/scripts/uom/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/uom/14.0.1.0/post-migration.py
@@ -3,6 +3,22 @@
 from openupgradelib import openupgrade
 
 
+def change_default_uom_rounding(env):
+    # in 13.0, the default value for the rounding field of these units is set
+    # to 0.001 in uom_data.xml. this is removed in 14.0, so it falls back to
+    # the default value of the field, which is 0.01. the rounding on these
+    # units is reset here to their default value if the value hasn’t change,
+    # to match how it would be if the database was created in 14.0, which is
+    # what is expected from openupgrade.
+    for ref in ("uom.product_uom_unit", "uom.product_uom_kgm"):
+        uom = env.ref(ref, raise_if_not_found=False)
+        if not uom:
+            continue
+        if uom.rounding == 0.001:
+            uom.rounding = 0.01
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(env.cr, "uom", "14.0.1.0/noupdate_changes.xml")
+    change_default_uom_rounding(env)


hbrunn (changes requested):
please comment instead of deleting, this makes it easier to integrate changes whenever this file is newly generated.

Note that in newer versions, we copy the file to noupdate_changes_work.xml and modify that, but I don't see a point in doing so here.

As for the change itself: I'm not so sure this will be a good choice for most users. On the other hand, we also disable updating the [decimal precision](https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/scripts/product/14.0.1.2/noupdate_changes.xml#L4), so your change would make it consistent at least.

Though to me, it seems the better way to consistency would be to also update the decimal precision to v14's default, and have users who need a different one update accordingly. Difficult to judge as I don't have a real world db that old at hand. What are you considerations? Other reviewers?

huguesdk:
i deleted it because it is not an openupgrade decision: it is only an update of the file from the odoo module, as if it was newly generated. should this still be a comment instead of a deletion?

this change has been made in odoo 5,5 years ago, and keeping this actually introduces a change that isn’t present in odoo anymore. the rounding is now only defined as a default value on the field (see [here](https://github.com/OCA/OCB/blob/14.0/addons/uom/models/uom_uom.py#L40)), not in the xml data anymore. note that this is not related to the decimal precision (which can be different per field), it is the rounding of the uom itself (which can be lower than the decimal precision used, for example 1.0 for units but 0.001 for kg).

we had the problem during our upgrades (from 12 to 16): these 2 units lost the rounding they had.

hbrunn:
> i deleted it because it is not an openupgrade decision: it is only an update of the file from the odoo module, as if it was newly generated. should this still be a comment instead of a deletion?

ah, I thought this is what fell out of `upgrade_analysis`, but it was added manually in https://github.com/OCA/OpenUpgrade/commit/e0bf271f9ec09db01fef56ddda34232ebf50debd as I see now - @pedrobaeza remember what this was about?

So indeed deleting is the right thing to do. Would like to hear which problem the commit referenced above solves before proceeding here

pedrobaeza:
The aim seems to be to get the new rounding defaults, as the previous one were not the same, but unfortunately I didn't put too many details on that time.

huguesdk:
in that case, how to prevent to overwrite the current values defined by the users? as they are only default values and not part of the data files, they shouldn’t be changed, right? if you think that the new default values should be enforced, i can work around the problem locally by storing the value somewhere and restoring it afterwards. but isn’t this a case of data loss (as is the case when overwriting `noupdate` values, actually)?

pedrobaeza:
This debate is the same for any noupdate data: to preserve possible user changes, or to adapt them to new values. On OpenUpgrade, we usually do the second one, as not having that values may lead to inconsistent things. That's why I put the new default rounding for those that didn't touch it for having the new expected value. If you have touched it and want to preserve that one, I think you should do extra scripts or change it manually afterwards. Not doing this change will keep the old values in case no change was performed.

huguesdk:
makes sense, thank you for the explanation, @pedrobaeza. what i don’t understand here is that the default value of the `rounding` field of `uom.uom` did not change: it was `0.01` [in 13.0](https://github.com/OCA/OCB/blob/13.0/addons/uom/models/uom_uom.py#L48) as well as [in 14.0](https://github.com/OCA/OCB/blob/14.0/addons/uom/models/uom_uom.py#L40) (even [in 8.0](https://github.com/OCA/OCB/blob/8.0/addons/product/product.py#L159) it was already the same default value).

i think that here it is really a leftover of an old value from a `noupdate` xml data, which has been removed in odoo after `noupdate_changes.xml` has been generated, and the file hasn’t been updated since.

pedrobaeza:
@huguesdk but that's not the problem... The problem is that in 13, the specific rounding of such records was `0.001`:

https://github.com/OCA/OCB/blob/e473a797f8f5df16901efa3e099c1e06187e8db1/addons/uom/data/uom_data.xml#L47

and now, being removed from data, it takes the default one (`0.01`), so that's why we explicitly add this second one into noupdate changes for switching to the new default these records coming from 13-.

pedrobaeza (approved):

MiquelRForgeFlow (commented):
In `noupdate_changes.xml` file, please comment lines instead of deleting them.

huguesdk:
oh, ok, i missed that change from 13.0 to 14.0, thanks! now i get it. do you think that `noupdate_changes.xml` is the correct place to handle this kind of change (a reset to the default value)? what about handling this in `post-migration.py` like i’ve just done here? i added a comment explaining what has happened. moreover, the code only changes the value if it is still at the previous default value. (that last part could be dropped if you think that it’s too strange; my goal is to explain better what has happened and why the value is changed.) what do you think?

pedrobaeza:
Already approved it, but in general, IMO this can go in noupdate_changes.xml.

huguesdk:
i didn’t see your comments before posting my last comment. ☺️

i’ve re-added the lines in `noupdate_changes.xml`, but as they were before the change made in e0bf271f9ec09db01fef56ddda34232ebf50debd. ok like this?

edit: sorry @pedrobaeza, again i didn’t see your last message before posting this. in the previous paragraph i was replying to @MiquelRForgeFlow.

MiquelRForgeFlow (approved):

hbrunn (approved):

PR #5612, opened 2026-05-04
[19.0][MIG] hr_recruitment

hbrunn:
depends on:

- [x] https://github.com/OCA/openupgradelib/pull/446


openupgrade_scripts/scripts/hr_recruitment/19.0.1.1/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+    Prepare merging hr.candidate into hr.applicant:
+    Merge models
+    Rename hr_candidate to legacy table
+    Lift constraints on hr_candidate.id
+    """
+    openupgrade.merge_models(env.cr, "hr.candidate", "hr.applicant", "candidate_id")
+    openupgrade.rename_tables(env.cr, [("hr_candidate", None)])
+    openupgrade.lift_constraints(
+        env.cr, openupgrade.get_legacy_name("hr_candidate"), "id", cascade=True
+    )
```
MiquelRForgeFlow, line 18:
```suggestion
    openupgrade.remove_tables_fks(
        env.cr, [openupgrade.get_legacy_name("hr_candidate")]
    )
```
cleaner and safer

MiquelRForgeFlow (commented):
Please rebase.

openupgrade_scripts/scripts/hr_recruitment/19.0.1.1/pre-migration.py
hbrunn: they do different things, I think we need both actually

MiquelRForgeFlow: I thought all the things that `lift_constraints` does are included in `remove_tables_fks`. If not, then put both 🤷‍♂️

hbrunn: lift_constraints also removes primary key constraints, such that constraints on columns pointing to hr_candidate are lifted too when cascade=true is set

MiquelRForgeFlow (approved):

PR #5619, opened 2026-05-09
[13.0] [UPD] openupgrade_analysis

primes2h:


MiquelRForgeFlow:
@pedrobaeza should we update some scripts here, or in another PR after merging this one?

pedrobaeza:
Well, an analysis about possible extra works should be done, yes

addons/account/migrations/13.0.1.1/openupgrade_analysis.txt
```diff
@@ -358,11 +362,18 @@ l10n_fr_certification / account.move             / l10n_fr_hash (char)
 l10n_fr_certification / account.move             / l10n_fr_secure_sequence_number (integer): DEL
 l10n_fr_certification / res.company              / l10n_fr_secure_sequence_id (many2one): DEL relation: ir.sequence
 ---XML records in module 'account'---
+DEL account.account.tag: account.demo_capital_account (noupdate)
```
MiquelRForgeFlow: The analysis is done with demo data, and it should be done instead without demo data.

MiquelRForgeFlow (changes requested):

addons/account/migrations/13.0.1.1/openupgrade_analysis.txt
primes2h: > The analysis is done with demo data, and it should be done instead without demo data.

Right, sorry. It should be ok now.

addons/maintenance/migrations/13.0.1.0/openupgrade_analysis.txt
```diff
@@ -9,4 +9,8 @@ mail.message.subtype: maintenance.mt_cat_req_created (noupdate) (noupdate switch
 mail.message.subtype: maintenance.mt_mat_assign (noupdate) (noupdate switched)
 mail.message.subtype: maintenance.mt_req_created (noupdate) (noupdate switched)
 mail.message.subtype: maintenance.mt_req_status (noupdate) (noupdate switched)
+maintenance.stage: maintenance.stage_0 (noupdate) (noupdate switched)
+maintenance.stage: maintenance.stage_1 (noupdate) (noupdate switched)
+maintenance.stage: maintenance.stage_3 (noupdate) (noupdate switched)
+maintenance.stage: maintenance.stage_4 (noupdate) (noupdate switched)
```
MiquelRForgeFlow, line 15: needs scripts here

MiquelRForgeFlow (commented):
I have looked the changes, and only this seems to be needed.

addons/maintenance/migrations/13.0.1.0/openupgrade_analysis.txt
MiquelRForgeFlow: Ah, I see it's already done in https://github.com/OCA/OpenUpgrade/commit/8a105fa2e3c3953bf6e28c9e9993ace5f466825f.

MiquelRForgeFlow (approved):
No extra work needed.

PR #5626, opened 2026-05-13
[18.0][OU-IMP] point_of_sale: Set the access_token value in pos.config to prevent all records from having the same value

victoralmau:
Set the `access_token` value in `pos.config` to prevent all records from having the same value

@Tecnativa TT62316

### DIFF
--- a/openupgrade_scripts/scripts/point_of_sale/18.0.1.0.2/post-migration.py
+++ b/openupgrade_scripts/scripts/point_of_sale/18.0.1.0.2/post-migration.py
@@ -1,7 +1,17 @@
 # Copyright 2025 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from uuid import uuid4
+
 from openupgradelib import openupgrade
+
+
+def fill_pos_config_token(env):
+    """Set the access_token to an appropriate value, thereby preventing all records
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

openupgrade_scripts/scripts/point_of_sale/18.0.1.0.2/pre-migration.py
```diff
@@ -3,6 +3,24 @@
...
+    if openupgrade.column_exists(env.cr, "pos_config", "access_token"):
+        # If the column already exists (because pos_self_order was installed),
+        # no further action is required.
+        return
+    openupgrade.add_fields(
+        env,
+        [("access_token", "pos.config", "pos_config", "char", False, "point_of_sale")],
+    )
+    openupgrade.logged_query(
+        env.cr, "UPDATE pos_config SET access_token = gen_random_uuid();"
```
pedrobaeza: Putting this into a DB, I see the returned UUID is not the same format, including `-` in this SQL form, and being longer. I think you should do this in post and use Python function.

pedrobaeza (changes requested):

pedrobaeza:
```python
from uuid import uuid4

for config in env["pos.config"].search([]):
    config.access_token = uuid4().hex[:16]
```

pedrobaeza (approved):

PR #5646, opened 2026-05-20
[19.0][OU-ADD] l10n_es: Migration scripts

eduezerouali-tecnativa:
cc @Tecnativa TT58647
ping @pedrobaeza


hbrunn (approved):
can't judge this functionally, but technically looks correct

eduezerouali-tecnativa:
I was missing the entry in the coverage file. Done

hbrunn:
/ocabot migration l10n_es

<dependency></dependency>

openupgrade_scripts/scripts/l10n_es/19.0.5.4/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,121 @@
...
+---Fields in module 'l10n_es'---
+---XML records in module 'l10n_es'---
+DEL account.account.tag: l10n_es.account_tag_mod390_m104
+DEL account.account.tag: l10n_es.account_tag_mod390_m230
+DEL account.account.tag: l10n_es.account_tag_mod390_m232
+DEL account.account.tag: l10n_es.account_tag_mod390_m773
+DEL account.account.tag: l10n_es.account_tag_mod390_m774
+DEL account.account.tag: l10n_es.account_tag_mod390_m775
+DEL account.account.tag: l10n_es.account_tag_mod390_m776
+
```
pedrobaeza: Don't put empty line here (and in the rest in a similar way)

openupgrade_scripts/scripts/l10n_es/19.0.5.4/pre-migration.py
```diff
@@ -0,0 +1,211 @@
+# Copyright 2026 Tecnativa - Eduardo Ezerouali
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from openupgradelib import openupgrade
+
+_xmlid_renames = [
+    (
+        "l10n_es_modelo130.mod_130",
```
pedrobaeza: This is not needed, as in any case, this should be done merging the modules, where the XML-ID renaming will take place, but the module `l10n_es_modelo130` seems to be something that I'm not aware of. I don't see it on community. @hbrunn do you know where this comes from?

pedrobaeza (changes requested):

openupgrade_scripts/scripts/l10n_es/19.0.5.4/pre-migration.py
pedrobaeza: Oh, found it: https://github.com/odoo/odoo/tree/18.0/addons/l10n_es_modelo130. Then definitively nothing more to do: https://github.com/OCA/OpenUpgrade/blob/19.0/openupgrade_scripts/apriori.py#L26

eduezerouali-tecnativa: Done. I have removed the rename. I assume that _xmlids_to_delete is still required, as it is not covered by the merging process.

pedrobaeza (approved):

PR #5668, opened 2026-05-22
[16.0][FIX] fix pos.session cash register amounts

huguesdk:
ensure that the `account.bank.statement` record corresponding to the cash journal is used to set the value of the cash register fields, as there can be multiple records linked to the pos.session (one per payment method), and ensure that the values are not null.

### DIFF
--- a/openupgrade_scripts/scripts/point_of_sale/16.0.1.0.1/post-migration.py
+++ b/openupgrade_scripts/scripts/point_of_sale/16.0.1.0.1/post-migration.py
@@ -27,11 +27,16 @@ def copy_register_balance(cr):
     openupgrade.logged_query(
         cr,
         """
-        UPDATE pos_session session
-        SET cash_register_balance_end_real=statement.balance_end_real,
-            cash_register_balance_start=statement.balance_start
-        FROM account_bank_statement statement
-        WHERE statement.pos_session_id=session.id
+        update pos_session as ps
+        set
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Next time, please put the module name in the commit message.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

huguesdk:
@pedrobaeza thanks for the merge!

about the module name in the commit message, are you talking about the subject (summary) or the body of the message? because the documentation contradicts itself: it first says:

> Write a short commit summary without prefixing it. It should not be longer than 50 characters: This is a commit message
> 
> Then, in the message itself, specify the part of the code impacted by your changes (module name, lib, transversal object, ...) and a description of the changes. This part should be multiple lines no longer than 80 characters.

following this, there should be no tag or module name in the subject, and the module name should be in the body.

but then, the last bullet point says to use tags, and then the examples put the module name in the subject.

the problem with putting this in the subject is that it takes a lot of space and often doesn’t leave enough space to write something meaningful and not exceed 50 characters, especially with long module names.

what do you think? should we maybe allow for longer subjects (for example: 50 characters not including the tag and module name)? or putting the module name in the body is enough?

pedrobaeza:
Well, for me the examples are clear:

imagen

You can see the module name there. If other redaction is ambiguous, please propose a change in that document.

PR #5501, opened 2026-02-02
[19.0][MIG] portal

hbrunn:
depends:

- [x] https://github.com/OCA/OpenUpgrade/pull/5446
- [x] https://github.com/OCA/OpenUpgrade/pull/5466
- [x] https://github.com/OCA/openupgradelib/pull/437

PR #5673, opened 2026-05-26
[19.0][FIX] loyalty: incomplete migration

hbrunn:
previous attempt was [incomplete](https://github.com/OCA/OpenUpgrade/pull/5629#issuecomment-4541258827) and [wrong](https://github.com/OCA/OpenUpgrade/actions/runs/26442547247/job/77840404135)


hbrunn:
/ocabot migration loyalty

<dependency></dependency>

<dependency>Depends on : 

- product : #5477
- portal : #5501
- account : #5509</dependency>

dnplkndll:
@hbrunn is this out of scope like the backfill on dates since would hit a copy only?

https://github.com/hbrunn/OpenUpgrade/compare/19.0-fix-crm...ledoent:OpenUpgrade:19.0-fix-loyalty-suggestion

hbrunn:
it's not out of scope, the replacement should however be `_mail_get_customer()` instead, and the fields to update should come from `mail.template#_get_dynamic_field_names`. Then I'd rather see this be done in an end-migration script and using the ORM, as speed isn't much of an issue here.

People with custom reports will have to fix them themselves.

You can propose your fix after this one is merged, which will be faster if you actually take part in the review process.

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
> You can propose your fix after this one is merged, which will be faster if you actually take part in the review process.

@dnplkndll You can proceed now to propose new fixes.

pedrobaeza:
But if this is incomplete, it shouldn't put `Done` in the documentation.

PR #5675, opened 2026-05-26
[19.0][FIX] hr_expense: don't update product uom when product is invoiced

hbrunn:
we can't update product UOMs if the products in question are invoiced

Depends on

- [x] https://github.com/OCA/openupgradelib/pull/452


pedrobaeza:
Can you rebase for checking the CI?

pedrobaeza (approved):

PR #5676, opened 2026-05-29
[19.0][OU-ADD] google_account: Nothing to do

carlos-lopez-tecnativa:
TT58624


hbrunn (commented):
we usually leave those empty as no db layout changes is an implicit nothing to do, but I don't mind adding the explicitly

hbrunn (approved):

pedrobaeza (approved):

carlos-lopez-tecnativa:
> we usually leave those empty as no db layout changes is an implicit nothing to do, but I don't mind adding the explicitly

Sorry, but I wasn't aware of that criterion. I had in mind that it is always necessary to explicitly mark it as either `Done` or `Nothing to do`.

Is this criterion documented somewhere?

PR #5682, opened 2026-06-02
[19.0][IMP] base, uom: use xml transformation instead of work file

hbrunn:
with https://github.com/OCA/openupgradelib/pull/452 merged, we can get rid of all copies of noupdate changes and use the transformation mechanism exclusively


legalsylvain:
thanks a lot. That is very simpler to read.

legalsylvain (approved):

pedrobaeza (approved):
Indeed, this is cleaner.

PR #5692, opened 2026-06-03
[19.0][OU-ADD] account_edi: Nothing to do

eduezerouali-tecnativa:
cc @Tecnativa TT58563
@pedrobaeza


legalsylvain:
/ocabot migration account_edi


<dependency></dependency>

PR #5693, opened 2026-06-03
[19.0][OU-ADD] base_vat: Nothing to do

eduezerouali-tecnativa:
cc @Tecnativa TT58604
@pedrobaeza


legalsylvain:
/ocabot migration base_vat

<dependency></dependency>

PR #5566, opened 2026-03-23
[19.0][MIG] hr+dependencies

hbrunn:

Issue #5363, opened 2025-09-04
Migration to version 19.0


hbrunn:
I'm starting to work on initializing the 19.0 branch this week, with the usual `openupgrade_framework` and `base` migration script.

- [x] account - By @hbrunn - #5509
- [x] account_edi - By @eduezerouali-tecnativa - #5692
- [x] account_edi_ubl_cii - By @eduezerouali-tecnativa - #5685
- [x] account_payment - By @hbrunn - #5527
- [x] analytic - By @hbrunn - #5509
- [x] attachment_indexation - By @eduezerouali-tecnativa - #5699
- [ ] auth_oauth - By @carlos-lopez-tecnativa - #5759
- [x] auth_password_policy - By @carlos-lopez-tecnativa - #5738
- [x] auth_password_policy_signup - By @carlos-lopez-tecnativa - #5739
- [x] auth_signup - By @hbrunn - #5501
- [x] auth_totp - By @carlos-lopez-tecnativa - #5733
- [x] auth_totp_mail - By @carlos-lopez-tecnativa - #5734
- [x] auth_totp_portal - By @carlos-lopez-tecnativa - #5737
- [x] barcodes - By @hbrunn - #5547
- [x] barcodes_gs1_nomenclature - By @hbrunn - #5547
- [x] base - By @hbrunn - #5446
- [x] base_address_extended - By @eduezerouali-tecnativa - #5695
- [x] base_geolocalize - By @hbrunn - #5731
- [x] base_iban - By @eduezerouali-tecnativa - #5702
- [x] base_import - By @eduezerouali-tecnativa - #5694
- [x] base_setup - By @hbrunn - #5466
- [x] base_sparse_field - By @pilarvargas-tecnativa - #5689
- [x] base_vat - By @eduezerouali-tecnativa - #5693
- [x] board - By @eduezerouali-tecnativa - #5697
- [x] bus - By @hbrunn - #5466
- [x] calendar - By @MiquelRForgeFlow - #5613
- [x] contacts - By @eduezerouali-tecnativa - #5681
- [x] crm - By @eduezerouali-tecnativa - #5706
- [x] delivery - By @pilarvargas-tecnativa - #5717
- [x] digest - By @hbrunn - #5505
- [x] don't warn about changes in the base module in the analysis cronjob if only the general log file is changed (done in https://github.com/OCA/OpenUpgrade/pull/5389)
- [ ] event - By @hbrunn - #5748
[...]

PR #5705, opened 2026-06-09
[19.0][MIG] hr_attendance

hbrunn:


hbrunn:
/ocabot migration hr_attendance

<dependency>Depends on : 

- hr : #5566
- barcodes : #5547
- base_geolocalize : #5731</dependency>

hbrunn:
barcodes is nothing to do and base_geolocalize has no db changes which we treat as an implicit nothing to do

legalsylvain:
> barcodes is nothing to do and base_geolocalize has no db changes which we treat as an implicit nothing to do

- I changed my script to add the merged PR. so it is more explicit. (now hr is displayed) Is it better ?
- barcodes is a dependency : https://github.com/odoo/odoo/blob/19.0/addons/hr_attendance/__manifest__.py#L18. I don't know if it has nothing todo. I don't use hr_attendance module.
- base_geolocalize : has no db changes. Indeed.

[script ref](https://github.com/grap/openupgrade-extra-tools/blob/master/check_dependency.py).

hbrunn:
thanks @legalsylvain for the link - actually I mailed you about not finding this script a while ago but never received an answer. did you change mail addresses?

barcodes has been set to nothing to do in https://github.com/OCA/OpenUpgrade/pull/5547, I don't see at first glance right now why this isn't picked up by `_get_module_coverage`.

for base_geolocalize, I just added https://github.com/OCA/OpenUpgrade/pull/5731

specifics aside, I wonder if you agree to move get_workload.py to https://github.com/OCA/OpenUpgrade/tree/documentation/docsource/maintainer_scripts and make check_dependency.py a github action? Less manual work and the check happens instantly.

legalsylvain:
> barcodes has been set to nothing to do in https://github.com/OCA/OpenUpgrade/pull/5547, I don't see at first glance right now why this isn't picked up by _get_module_coverage.

_get_module_coverage get PR lists, parsing the "main" issue comment. (https://github.com/OCA/OpenUpgrade/issues/5363#issue-3384690512)

I suspect this "multi migration comment" failed. https://github.com/OCA/OpenUpgrade/pull/5547#issuecomment-4714924021
I recalled manually, module by module.

-> now, this comment is correctly updated. https://github.com/OCA/OpenUpgrade/pull/5705#issuecomment-4660271760

> specifics aside, I wonder if you agree to move get_workload.py to https://github.com/OCA/OpenUpgrade/tree/documentation/docsource/maintainer_scripts and make check_dependency.py a github action? Less manual work and the check happens instantly.

I'd like to, but I'm not an expert of github action. We could plan to do it in the next OCA days, if you planed to come ?

hbrunn:
I do plan to come there, but that's quite little effort, not really worth wasting the OCA days on. will do one of the coming weeks and ping you on the PR

legalsylvain:
> will do one of the coming weeks and ping you on the PR

Thanks. please ping me if you have any questions regarding current script.

hbrunn:
@legalsylvain done in https://github.com/OCA/OpenUpgrade/pull/5745, now setting our whole discussion as off-topic

legalsylvain:
> @legalsylvain done in https://github.com/OCA/OpenUpgrade/pull/5745, now setting our whole discussion as off-topic

thanks a lot for this Job. I manually run the script a last time for the V19. new 19 PR should be handled by your script.

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+    "hr_attendance.hr_attendance_overtime_rule_employee_company",
+    "hr_attendance.hr_attendance_rule_attendance_officer_overtime_restrict",
+    "hr_attendance.hr_attendance_rule_attendance_overtime_admin",
+    "hr_attendance.hr_attendance_rule_attendance_overtime_simple_user",
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "hr_attendance", "19.0.2.0/noupdate_changes.xml")
```
pedrobaeza, line 77: For letting the record as it's in 19, we should remove the group `hr_attendance.group_hr_attendance_manager` from `hr_attendance.hr_attendance_rule_attendance_admin` record.

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,175 @@
+---Models in module 'hr_attendance'---
+obsolete model hr.attendance.overtime (renamed to hr.attendance.overtime.line)
+new model hr.attendance.overtime.line (renamed from hr.attendance.overtime)
+
```
pedrobaeza, line 4: Why adding these empty lines that makes the reading harder? I think the comment should be together with the analysis lines block.

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,175 @@
...
+hr_attendance / hr.attendance            / in_city (char)                : DEL
+
+# DONE: renamed to in_location
+
+hr_attendance / hr.attendance            / in_country_name (char)        : DEL
+
+# NOTHING TO DO
+
+hr_attendance / hr.attendance            / in_location (char)            : NEW
```
pedrobaeza, line 25: Reorganize this line into `in_city` DEL comment for seeing both together and to understand the change.

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,216 @@
...
+                "datetime",
+                None,
+                "hr_attendance",
+            ),
+        ],
+    )
+    # date is required in v19, fill with create_date if empty, possibly wrong
+    env.cr.execute(
+        f"""
+        UPDATE hr_attendance_overtime_line SET date=(
```
pedrobaeza: Why not get it the same from the nearest attendance record?

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,216 @@
...
+        "hr_attendance_overtime_line",
+        "duration_real",
+        "manual_duration",
+    ),
+]
+
+SQL_EMPLOYEE2TZ = """
+(
+    SELECT
+    hr_employee.id employee_id,
```
pedrobaeza: I find you new SQL style a bit difficult to read, and very vertical scroll intensive. Either indent the lines after the command (`SELECT`, `FROM`, etc), or put them in the same line as always.

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,216 @@
...
+        hr_attendance_overtime_line
+        SET
+        time_start=hr_attendance.check_in,
+        time_stop=hr_attendance.check_out
+        FROM
+        hr_attendance
+        WHERE
+        hr_attendance.employee_id=hr_attendance_overtime_line.employee_id
+        AND
+        hr_attendance.date=hr_attendance_overtime_line.date
```
pedrobaeza: This part may not fit according current code, and even more, we may have more than one record for the same date.

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,216 @@
...
+    hr_employee.id
+) employee2zone
+"""
+
+
+def hr_attendance_date(env):
+    """
+    Pre-fill hr.attendance#date
+    """
+    openupgrade.add_fields(
```
pedrobaeza: Why using `add_fields` instead of `add_columns` or a simply `ALTER TABLE`?

pedrobaeza (dismissed):

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/post-migration.py
hbrunn: done

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
hbrunn: It's not so [new](https://docs.telemetry.mozilla.org/concepts/sql_style.html), but indeed was lacking indentation

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
hbrunn: good question, changed to add_columns

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
hbrunn: how do you propose to find the nearest record? create_date? that's just as prone to filling in a wrong date here.

I consider overtime lines without date garbage data which people should clean up, and do the update only to satisfy the non-null constraint

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
hbrunn: which code do you mean? and indeed this can lead to updating the same record multiple times, but how does this matter?

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/upgrade_analysis_work.txt
hbrunn: well I think the empty lines improve readability. If that's so important to you, add it to https://oca.github.io/OpenUpgrade/090_contribute.html#how-to-contribute-new-migration-scripts and I'll bulk update the whole branch, as I've been doing this for a long time

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/upgrade_analysis_work.txt
hbrunn: hard disagree about reordering lines in the work file. I want to be able to say `diff upgrade_analysis*` to easily verify all lines from the generated file are in the work file (=only whitespace or comments in the diff)

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
carlos-lopez-tecnativa: Based on the link, this formatting does not follow the recommended style https://docs.telemetry.mozilla.org/concepts/sql_style.html#join-conditions
Image

```suggestion
    FROM
        hr_employee
    JOIN
        resource_resource
        ON hr_employee.resource_id=resource_resource.id
    LEFT JOIN
        resource_calendar
        ON resource_resource.calendar_id=resource_calendar.id
    JOIN
        res_company
        ON hr_employee.company_id=res_company.id
    LEFT JOIN
        resource_calendar company_resource_calendar
        ON res_company.resource_calendar_id=company_resource_calendar.id
```

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,215 @@
...
+        [
+            (
+                "hr.attendance",
+                "validated_overtime_hours",
+                "float",
+                None,
+                "hr_attendance",
+            )
+        ],
+    )
```
carlos-lopez-tecnativa: This field already exists in v18 and is computed and stored, so why is it necessary to add the column?
The logic used to compute the field has changed slightly, but I understand that adding the column should not be necessary.

https://github.com/odoo/odoo/blob/e88fb3a5f5c2239f2a3f434a4ac3052e3e156ef1/addons/hr_attendance/models/hr_attendance.py#L49

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,175 @@
...
+hr_attendance / hr.attendance.overtime.ruleset / rule_ids (one2many)           : NEW relation: hr.attendance.overtime.rule
+hr_attendance / hr.employee              / overtime_ids (one2many)       : relation is now 'hr.attendance.overtime.line' ('hr.attendance.overtime') [nothing to do]
+hr_attendance / hr.version               / ruleset_id (many2one)         : NEW relation: hr.attendance.overtime.ruleset, hasdefault: default
+hr_attendance / res.company              / attendance_device_tracking (boolean): NEW hasdefault: default
+
+# NOTHING TO DO
+
+---XML records in module 'hr_attendance'---
+NEW hr.attendance.overtime.rule: hr_attendance.hr_attendance_overtime_employee_schedule_rule (noupdate)
+NEW hr.attendance.overtime.rule: hr_attendance.hr_attendance_overtime_non_working_days_rule (noupdate)
```
carlos-lopez-tecnativa, line 114: I think we should force the creation of these records to preserve the behavior from v18

In v18, there are two company-level fields used to configure tolerance for both the employee and the company. Odoo did not remove these fields([see](https://github.com/odoo/odoo/blob/9c20c8b2d1864d357e06d0dd27c455cef5b26b1e/addons/hr_attendance/models/res_company.py#L16-L19)), but the implementation is now handled by the new `hr.attendance.overtime.ruleset` model.

Image

Image

If these fields are set in my database, attendance records are taken into account when computing extra hours. However, in v19, if the corresponding rules are missing and the `ruleset_id` field is not set on `hr.version`, **extra hours are not computed correctly**.

Image

These records are [not created](https://github.com/odoo/odoo/blob/9c20c8b2d1864d357e06d0dd27c455cef5b26b1e/odoo/tools/convert.py#L374-L376) because they are inside a `noupdate` block with `forcecreate="0"` see [this file](https://github.com/odoo/odoo/blob/b867b64e339d833c325713307203e31e3f5a1b1e/addons/hr_attendance/data/hr_attendance_overtime_rule_data.xml#L4). In addition, we need to transfer the values from the company fields to the corresponding rule.

carlos-lopez-tecnativa (changes requested):
TT58628

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/pre-migration.py
hbrunn: thanks, indeed, we shouldn't touch that at all

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/upgrade_analysis_work.txt
hbrunn: good catch, added handling that and will look into upgrade_analysis flagging this kind of record

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/post-migration.py
```diff
@@ -26,8 +26,53 @@ def cleanup_hr_attendance_rule_attendance_admin(env):
...
+                            "company_id": company.id,
+                            "rule_ids": [
+                                Command.link(other_rule.copy().id)
+                                for other_rule in default_rule.ruleset_id.rule_ids
+                                if other_rule != default_rule
+                            ],
+                        }
+                    ).id,
+                }
+            )
```
carlos-lopez-tecnativa, line 69: I don't agree with this logic.

Why are you setting the fields on the default rule and then copying that rule to every company?

I don't think the fields should be set on the default rule.

If the default rule does not have a company_id set, why should it take the values from the main company?

carlos-lopez-tecnativa (changes requested):

openupgrade_scripts/scripts/hr_attendance/19.0.2.0/post-migration.py
hbrunn: this results in the default rule without company applying to all companies that have the same settings as the main company, which probably is the most common case. you only get specific rules for companies that have different settings than the main company.

what would your alternative do?

carlos-lopez-tecnativa: > this results in the default rule without company applying to all companies that have the same settings as the main company, which probably is the most common case. you only get specific rules for companies that have different settings than the main company.
> 
> what would your alternative do?

**Option 1:** If you create or duplicate the rule for the `main_company`, then don't set these fields on the default rule. Otherwise, we assume that the values from the main company are the same for all other companies. In other words, **don't modify the default rule; instead, create a new rule for each company that has these fields set.**

**Option 2:** If the default rule without a `company_id` is intended to preserve the field values from the `main_company`, then don't create or duplicate another rule for the `main_company`.

Example:

My Company has these fields set.
C2 has these fields set, but with values different from My Company.
C3 does not have these fields set.

**With Option 1, I would expect the following:**

Don't modify the default rule, so it can be used by C3.
Don't create a rule for C3, because the default rule is shared.
Create a dedicated rule for My Company and C2, with the specific values configured for each company.

**What I see as incorrect behavior:**

A rule is created for C3, even though this company does not have these fields set.
The default rule receives the same values as My Company, but another rule is also created for My Company. Instead, it should use the default rule rather than creating a new one.

image

I prefer Option 1. What do you think?

hbrunn: i believe i implemented option 2, and i think that's the good one. would expect no new rule for main company in your example

carlos-lopez-tecnativa: > i believe i implemented option 2, and i think that's the good one. would expect no new rule for main company in your example

OK, if you want to use Option 2, that's fine. However, in this case a new rule is still created for My Company. Please check my screenshot; this is a database migrated from v18.

The rule is created because, in the if statement, you compare values that have been divided by 60 when assigned to the default rule, while the values on the company fields are not divided by 60. As a result, they are never equal, and the code always enters the block that copies the rule.

`default_rule.employer_tolerance != company.overtime_company_threshold`

It should be:

`default_rule.employer_tolerance != (company.overtime_company_threshold / 60.0)`

carlos-lopez-tecnativa:
```suggestion
    new_employer_tolerance = main_company.overtime_company_threshold / 60.0
    new_employee_tolerance = main_company.overtime_employee_threshold / 60.0
    default_rule.employer_tolerance = new_employer_tolerance
    default_rule.employee_tolerance = new_employee_tolerance
    for company in env["res.company"].search([("id", "!=", main_company.id)]):
        rule = default_rule
        if (
            default_rule.employer_tolerance != new_employer_tolerance
            or default_rule.employee_tolerance != new_employee_tolerance
        ):
            rule = default_rule.copy(
                {
                    "employer_tolerance": new_employer_tolerance,
                    "employee_tolerance": new_employee_tolerance,
                    "ruleset_id": default_rule.ruleset_id.copy(
                        {
                            "company_id": company.id,
                            "rule_ids": [
                                Command.link(other_rule.copy().id)
                                for other_rule in default_rule.ruleset_id.rule_ids
                                if other_rule != default_rule
                            ],
                        }
                    ).id,
                }
            )
```

hbrunn: thanks, your code would never create a new rule though. I think d698ab8a119c gets it right

carlos-lopez-tecnativa: Sorry, but after testing the latest changes, I'm not convinced by the results.

Following my example:

- My Company has these fields set.
- C2 has these fields set, but with values different from My Company.
- C3 does not have these fields set.

Currently, **a new rule is created for C3**, but I would expect no rule to be created for this company.

The default rule (without company_id set) takes the values from main_company, but this rule is shared. As a result, other companies inherit values from main_company incorrectly.

I made these changes locally, and the behavior is correct according to my expectations:

- No rule is created for C3.
- An explicit rule is created for every company that has these fields set.
- The default rule remains untouched.

I think this behavior is more correct because the default rule remains shared and is used by companies that do not have specific values configured, while companies with customized values get their own dedicated rules.

Image

Image

hbrunn: well, I disagree this is the most natural expectation. I think we should treat everything that's different from the main company the exception, and whatever the main company does should be the default for new companies

carlos-lopez-tecnativa: Okay, while another person can review this and decide whether the values configured on main_company can be used by other companies.

Please avoid creating rules for companies that do not have these fields set, because the default rule without a company_id should be used instead.

In my example, no rule should be created for C3.

Could you apply this change?
image

hbrunn: we need this rule because otherwise the default one will be active. remember the goal here is to get v19 to behave in a similar way it was in v18

carlos-lopez-tecnativa (approved):
This PR is working, so I don't think this should block it. If we later find a case where creating rules for companies that do not have these fields set is unnecessary, we can improve this behavior. For now, let's go ahead.

PR #5681, opened 2026-06-02
[19.0][OU-ADD] contacts: Nothing to do

eduezerouali-tecnativa:
cc @Tecnativa TT58610
ping @pedrobaeza @hbrunn

PR #5613, opened 2026-05-06
[19.0][MIG] calendar

MiquelRForgeFlow:
.

PR #5706, opened 2026-06-09
[19.0][OU-ADD] crm: Migration scripts

eduezerouali-tecnativa:
cc @Tecnativa TT58614
@pedrobaeza


legalsylvain:
/ocabot migration crm

<dependency>Depends on : 

- base_setup : #5466
- sales_team : #5528
- mail : #5466
- calendar : #5613
- resource : #5504
- utm : #5580
- web_tour : #5466
- contacts : #5681
- digest : #5505
- phone_validation : #5566</dependency>

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+                THEN 'lost'
+            ELSE 'pending'
+        END
+    """,
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "crm", "19.0.1.9/noupdate_changes.xml")
```
hbrunn: I think all those values are problematic. The writes on stage records will break if the user has deleted them in v18, and `crm_pls_fields_param` might have been customized.

This is one of the very few cases where I'd suggest to not load the file, but write code to do the updates conditionally. For the param something like  `','.join(set(param.split(',')) | set('new', 'fields'))`

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _won_stage_field(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE crm_lead cl
+        SET won_status = CASE
+            WHEN cl.probability = 100
+                 AND EXISTS (
```
hbrunn: this should be a left join

hbrunn (changes requested):

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
eduezerouali-tecnativa: Done

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
eduezerouali-tecnativa: Done

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
```diff
@@ -0,0 +1,72 @@
...
+        "stage_lead4": 10,
+    }
+    for xml_id, color in colors.items():
+        record = env.ref(f"crm.{xml_id}", raise_if_not_found=False)
+        if record:
+            record.color = color
+
+
+@openupgrade.migrate()
+def migrate(env, version):
```
hbrunn, line 46:
```suggestion
def migrate(env, version):
    # do not load noupdate_changes.xml, because all entries there are problematic
    # the necessary changes are done manually in the _load_data_* functions below
```

hbrunn (changes requested):

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
eduezerouali-tecnativa: Done

hbrunn (approved):

openupgrade_scripts/scripts/crm/19.0.1.9/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,37 @@
+---Models in module 'crm'---
+---Fields in module 'crm'---
+crm          / crm.lead                 / commercial_partner_id (many2one): NEW relation: res.partner, hasdefault: compute, stored: False
+# NOTHING TO DO
+
+crm          / crm.lead                 / mobile (char)                 : DEL
+# DONE: if phone is empty and mobile not set phone with mobile
```
pedrobaeza: I think we should concatenate the mobile in the pone field in case both exist, or write it on the internal notes to not lose data. And put if done in pre-migration or post-migration.

openupgrade_scripts/scripts/crm/19.0.1.9/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,37 @@
...
+
+crm          / crm.lead                 / title (many2one)              : DEL relation: res.partner.title
+# NOTHING TO DO
+
+crm          / crm.lead                 / won_status (selection)        : NEW selection_keys: ['lost', 'pending', 'won'], isfunction: function, stored
+#DONE: set to prevent massive compute
```
pedrobaeza:
```suggestion
# DONE: pre-migration: pre-created column and set it to prevent massive computation
```

openupgrade_scripts/scripts/crm/19.0.1.9/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,37 @@
...
+crm          / crm.lead.scoring.frequency.field / color (integer)               : NEW hasdefault: default
+crm          / crm.stage                / color (integer)               : NEW
+# NOTHING TO DO
+
+crm          / crm.stage                / rotting_threshold_days (integer): NEW hasdefault: default
+# NOTHING TO DO: default is set to 0 to keep old behavoir
+
+crm          / crm.stage                / team_id (many2one)            : DEL relation: crm.team
+crm          / crm.stage                / team_ids (many2many)          : NEW relation: crm.team
+# Done: Transform m2o to m2m
```
pedrobaeza:
```suggestion
# DONE: post-migration: Transform m2o to m2m
```

openupgrade_scripts/scripts/crm/19.0.1.9/upgrade_analysis_work.txt
pedrobaeza: And I see you have done it on post-migration, so this is not preventing anything.

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
```diff
@@ -0,0 +1,75 @@
...
+    # do not load noupdate_changes.xml, because all entries are problematic
+    # the necessary changes are done manually in the _load_data_* functions below
```
pedrobaeza: Why the color loading in the stages is problematic? cc @hbrunn

pedrobaeza (changes requested):

openupgrade_scripts/scripts/crm/19.0.1.9/upgrade_analysis_work.txt
eduezerouali-tecnativa: I had the feeling more case can appear so I tag it as case 1. If we concatenate that, can this cause trouble with `phone_validation` ?

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
eduezerouali-tecnativa: I assume that is because what happens if those records were eliminated by the user, but in this can be manged by the https://github.com/OCA/openupgradelib/blob/ac165a1dccd1fba4f09fb15c194420cdaa4672af/openupgradelib/openupgrade.py#L434 
Am i wrong?

hbrunn: it's indeed problematic when the stages are deleted. for this hack to be active, you need to pass `mode='init_no_create'`, and then also pass an xml transformation to remove overwriting the config parameter. also a way to go, but not really simpler than what happens now. would be future proof for subsequent changes to the file though

pedrobaeza: OK, the `forcecreate="False"` parameter is the defining one. Then please:

- Specify the real reasons, not only to mention "conflicting".
- Put the `load_data` instruction, but commented, to give context.

openupgrade_scripts/scripts/crm/19.0.1.9/upgrade_analysis_work.txt
pedrobaeza: Well, not really, as that module only format the numbers, not block them, and they are already formatted.

pedrobaeza:
ping @eduezerouali-tecnativa

eduezerouali-tecnativa:
Changes Done. 
- Move `won_status` calculation to pre-migration
- Add case to `_migrate_lead_mobile` where phone and mobile are set
- Add more comments for better explanations

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
```diff
@@ -0,0 +1,67 @@
...
+    # Case 2: both phone and mobile exist and differ → concatenate them
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE crm_lead
+        SET phone = phone || ' / ' || mobile
+        WHERE phone IS NOT NULL AND phone != ''
+          AND mobile IS NOT NULL AND mobile != ''
+          AND btrim(phone) != btrim(mobile)
+        """,
```
pedrobaeza, line 13: It can be done in one shot:

```suggestion
    openupgrade.logged_query(
        env.cr,
        """
        UPDATE crm_lead
        SET phone = CONCAT(COALESCE(phone, ''), mobile)
        WHERE COELESCE(mobile, '') != ''
        """,
    )
```

openupgrade_scripts/scripts/crm/19.0.1.9/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+from openupgradelib import openupgrade
+
+
+def _won_status_set(env):
+    # Precreate won_status and set with sql
+    openupgrade.add_columns(
+        env,
+        [
+            ("crm.lead", "won_status", "selection", None, "crm_lead"),
+        ],
```
pedrobaeza, line 12:
```suggestion
        [("crm.lead", "won_status", "selection", "pending", "crm_lead")],
```

openupgrade_scripts/scripts/crm/19.0.1.9/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+                 AND cs.is_won = TRUE
+                THEN 'won'
+            WHEN cl.active = FALSE
+                 AND cl.probability = 0
+                THEN 'lost'
+            ELSE 'pending'
+        END
+        FROM crm_lead cl2
+        LEFT JOIN crm_stage cs ON cs.id = cl2.stage_id
+        WHERE cl2.id = cl.id
```
pedrobaeza: As we are already initializing all of them to "pending", just do 2 updates for the rest (note the changes in the boolean conditions):

```
       UPDATE crm_lead cl
       SET won_status = 'won'
       FROM crm_stage cs
       WHERE cs.id = cl.stage_id
       AND cs.is_won
       AND cl.probability = 100
```

```
       UPDATE crm_lead cl
       SET won_status = 'lost'
       WHERE NOT cl.active
       AND cl.probability = 0
```

pedrobaeza (approved):

PR #5712, opened 2026-06-11
[18.0][OU-IMP] website: fix migrated snippets

pilarvargas-tecnativa:
Add missing data-vcss="001" and data-vxml="001" attributes
required by Odoo 18 on migrated snippets.

@Tecnativa TT63052

@pedrobaeza please review

### DIFF
--- a/openupgrade_scripts/scripts/website/18.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/website/18.0.1.0/post-migration.py
@@ -1,0 +1,36 @@
+# Copyright 2026 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def migrate_snippet_attribute(env, snippet_class, attribute, value):
+    openupgrade.logged_query(
+        env.cr,
+        rf"""
+        UPDATE ir_ui_view v
+           SET arch_db = (
+               SELECT jsonb_object_agg(
+                          lang,
+                          regexp_replace(
+                              arch,
+                              $$<section(?=[^>]*class="[^"]*{snippet_class}[^"]*")(?![^>]*{attribute}=)([^>]*)>$$,
+                              $$<section\1 {attribute}="{value}">$$,
+                              'g'
+                          )
+                      )
+                 FROM jsonb_each_text(v.arch_db) AS translation(lang, arch)
+           )
+         WHERE v.website_id IS NOT NULL
+           AND v.arch_db::text LIKE '%{snippet_class}%'
+        """,
+    )
+
+
+@openupgrade.migrate()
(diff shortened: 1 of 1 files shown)


openupgrade_scripts/scripts/website/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2026 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def migrate_s_carousel_snippets(env):
+    env.cr.execute(
```
pedrobaeza: Use `openupgrade.logged_query` instead, and check the pre-commit. You may need to add `r` as prefix.

pilarvargas-tecnativa: thanks

pedrobaeza (commented):
As the SQL is mostly the same, you can put it once, and just pass arguments for the differences.

openupgrade_scripts/scripts/website/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,73 @@
...
+               WHERE view.website_id IS NOT NULL
+                 AND view.arch_db::text LIKE '%{snippet_class}%'
+               GROUP BY view.id
+          ) data
+         WHERE v.id = data.id
+        """,
+    )
+
+
+def migrate_s_comparisons_snippets(env):
```
pedrobaeza: Why not refactoring this one?

openupgrade_scripts/scripts/website/18.0.1.0/post-migration.py
```diff
@@ -0,0 +1,73 @@
...
+                         lang,
+                         regexp_replace(
+                             arch,
+                             $$<section(?=[^>]*class="[^"]*{snippet_class}[^"]*")(?![^>]*{attribute}=)([^>]*)>$$,
+                             $$<section\1 {attribute}="{value}">$$,
+                             'g'
+                         )
+                     ) AS arch_db
+                FROM ir_ui_view view
+                CROSS JOIN LATERAL jsonb_each_text(view.arch_db) AS value(lang, arch)
```
pedrobaeza: I don't get the need of this join. You can do it without it I think.

pilarvargas-tecnativa: done

openupgrade_scripts/scripts/website/18.0.1.0/post-migration.py
pilarvargas-tecnativa: done

pedrobaeza (approved):

PR #5714, opened 2026-06-12
[16.0][IMP] account: migrate bank statement chatter attachments to attachment_ids

CarlosRoca13:
In Odoo 16, account.bank.statement gains a new attachment_ids many2many field to ir.attachment. Add a post-migration query that links the existing chatter attachments (ir.attachment with res_model = 'account.bank.statement') into this field.

cc @Tecnativa TT63091

ping @pedrobaeza @sergio-teruel

### DIFF
--- a/openupgrade_scripts/scripts/account/16.0.1.2/post-migration.py
+++ b/openupgrade_scripts/scripts/account/16.0.1.2/post-migration.py
@@ -32,6 +32,23 @@ _deleted_xml_records = [
 ]
 
 
+def _fill_account_bank_statement_attachment_ids(env):
+    """The chatter attachments of the bank statements are now linked through
+    the new `attachment_ids` many2many field, so we fill it with the existing ones.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO account_bank_statement_ir_attachment_rel
+            (account_bank_statement_id, ir_attachment_id)
+        SELECT ia.res_id, ia.id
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
But where these attachments come from? It's an OCA module?

sergio-teruel:
Before version 16.0, the `account.bank.statement` model inherited from `mail.thread`, so users could attach files directly to the statement record through the chatter.

Starting from 16.0, this model no longer inherits from `mail.thread`, so the previous attachments are no longer visible through the chatter. However, Odoo added the `attachment_ids` field directly on the model here:

https://github.com/odoo/odoo/blob/f95bcc097fd7e70af8d28a66c010ae9b9490f439/addons/account/models/account_bank_statement.py#L100

In addition, `account_statement_base` already exposes this field in the form view using the `many2many_binary` widget:

https://github.com/OCA/account-reconcile/blob/18.0/account_statement_base/views/account_bank_statement.xml#L83

So I think there are two possible approaches:

1. Restore the chatter by inheriting from `mail.thread` again and adding the chatter back to the view.
2. Do not restore the chatter, but migrate/recreate the relation between the bank statements and their existing attachments so they are available through the current `attachment_ids` field.

Personally, I would prefer option 2, because it keeps the current Odoo design and avoids reintroducing the chatter only for attachment visibility.

pedrobaeza (approved):
Understood. Thanks for the clarifications. They may be in the commit message, but let's not block it.

PR #5716, opened 2026-06-13
[19.0][OU-FIX] account_edi_ubl_cii: guard map_values when source column is absent

mvalee:
## Problem

The `account_edi_ubl_cii` 19.0 pre-migration script calls `openupgrade.map_values()`
on `account_tax.ubl_cii_tax_exemption_reason_code`, but that column is **not present
on every 18.0 database** (e.g. Community installs that never stored it). On such a DB
the migration aborts:

```
psycopg2.errors.UndefinedColumn: column "ubl_cii_tax_exemption_reason_code" does not exist
ERROR ... OpenUpgrade: account_edi_ubl_cii: error in migration script
         .../account_edi_ubl_cii/19.0.1.0/pre-migration.py
CRITICAL ... Failed to initialize database `<db>`.
```

Because it runs during `-u all`, the whole 18.0 -> 19.0 upgrade fails, not just this module.

Hit while migrating a real Belgian Community 18.0 -> 19.0 database.

## Fix

Guard the value remap with `openupgrade.column_exists()`. When the column is absent
there is nothing to remap, and the 19.0 module update creates it afterwards, so
skipping is correct (not a workaround).

## Verification

With the guard, a full `-u all` 18.0 -> 19.0 upgrade completes on a Community database
that lacks the column (176 modules loaded), with data and accounting balances intact.


openupgrade_scripts/scripts/account_edi_ubl_cii/19.0.1.0/pre-migration.py
```diff
@@ -19,10 +19,18 @@
...
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.map_values(
-        env.cr,
-        "ubl_cii_tax_exemption_reason_code",
-        "ubl_cii_tax_exemption_reason_code",
-        _reason_code_map,
-        table="account_tax",
-    )
+    # The ``ubl_cii_tax_exemption_reason_code`` column is not present on every
```
hbrunn: how can this happen? that's more interesting than the generated bullshit below

hbrunn (changes requested):

openupgrade_scripts/scripts/account_edi_ubl_cii/19.0.1.0/pre-migration.py
aurelien-tempo: Hello,
I just had the case during my migration from 16 to 19.
In 18, the account_edi_ubl_cii module is auto-installed if you have the account module, while the account_edi_ubl_cii_tax_extension isn't.
In 19, the 2nd module was merged into the 1st one.
So if you had account_edi_ubl_cii but not account_edi_ubl_cii_tax_extension in 18, the column does not exist yet.

hbrunn: so please write a condensed version of this in the comment, and delete the rest

aurelien-tempo: @mvalee could you please update the comment so that it can be merged ?

mvalee: have adapted

pedrobaeza (approved):

hbrunn (approved):

PR #5719, opened 2026-06-15
[18.0][OU-FIX] l10n_it: remove xmlid for taxes as they would be partially duplicated with possible critics errors during migration

sergiocorato:
During migration from 17.0 of l10n_it the records with option `auto_sequence="1"` re-create theirs xmlids and sometimes they conflict with the existing records.

This PR remove completely them, to let Odoo re-create them safely.

This PR is for Italian module, maybe it could be a need of other localization modules too.

### DIFF
--- a/openupgrade_scripts/scripts/l10n_it/18.0.0.9/pre-migration.py
+++ b/openupgrade_scripts/scripts/l10n_it/18.0.0.9/pre-migration.py
@@ -1,0 +1,27 @@
+from openupgradelib import openupgrade
+
+# List of XML IDs of tax reports defined in l10n_it
+# These include the generic VAT report and the annual reports
+ITALIAN_REPORT_XML_IDS = [
+    "l10n_it.tax_report_vat",
+    "l10n_it.tax_annual_report_vat_va",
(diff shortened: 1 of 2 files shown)


hbrunn (commented):
won't you delete manual customizations this way?

sergiocorato:
> won't you delete manual customizations this way?

These reports are recreated from templates, so these manual customizations will be lost anyway.

These reports are only visible in EE but I don't use it so I don't know how they work, but how do you do customizations? They're only reports for what I can see.

hbrunn:
yes right, people will have to live with that.

it's still weird to me that `delete_records_safely_by_xml_id` does not do the job here, did you investigate this? Note in any case that `delete_childs` does not do what you think it does: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L3638-L3642, is this possibly the problem? In my opinion, this is simply fixable by replacing `"ir.ui.view"` with `record._name`

sergiocorato:
No more needed from commit https://github.com/OCA/OCB/commit/30bbf03637de2858b4bcc0ca2869ce46174271e1

PR #5732, opened 2026-06-16
[19.0][MIG] hr_timesheet

hbrunn:


hbrunn:
/ocabot migration hr_timesheet

<dependency>Depends on : 

- hr : #5566
- hr_hourly_cost : #5687
- analytic : #5509
- project : #5572
- uom : #5472</dependency>

carlos-lopez-tecnativa (commented):
LGTM.

Just a technical note: in this commit, Odoo removed `default_user` in favor of `default_user_group`:
https://github.com/odoo/odoo/commit/6451f5cce5b1fca770a95f766fe1fb99a9f56bd3#diff-b221f962f9047cb9f77630022f0d5689469db17e677854bc38f2653837791d75R9

In a fresh v19 database created **without demo data**, this new group does not have any privileges assigned.

Image

However, in a database migrated from v18 (**also without demo data**), I can see that this group does have the privileges assigned.

Image


I'm not sure whether each module should handle this individually or if it is acceptable for the `base` migration script to **copy all groups from default_user to default_user_group**, **considering that Odoo only performs this assignment when demo data is installed:**
https://github.com/OCA/OpenUpgrade/blob/c7915dca6f29f4fb6dc2e577fb93a8da0a78e176/openupgrade_scripts/scripts/base/19.0.1.3/post-migration.py#L85

TT58641

pedrobaeza (approved):
About the default groups for new users, I think it can be done in `base` module, yes.

PR #5736, opened 2026-06-16
[19.0][MIG] hr_skills

hbrunn:


hbrunn:
/ocabot migration hr_skills

<dependency>Depends on : 

- hr : #5566</dependency>

openupgrade_scripts/scripts/hr_skills/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,60 @@
...
+    env.cr.execute(
+        """
+        UPDATE hr_employee_skill
+        SET valid_from = create_date
+        WHERE valid_from IS NULL AND create_date IS NOT NULL
+        """
+    )
+
+
+def hr_job_job_skill_ids(env):
```
pedrobaeza, line 37:
```suggestion
def hr_job_skill_ids(env):
```

openupgrade_scripts/scripts/hr_skills/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def hr_employee_skill_validity_dates(env):
+    """
+    Set hr.employee.skill#valid_from from hr.employee.skill.log
+    """
+    env.cr.execute(
+        """
+        UPDATE hr_employee_skill hes
+        SET valid_from = hesl.date
```
pedrobaeza, line 14: What about `valid_to`?

pedrobaeza (changes requested):
And please check the comments about the style of other PRs.

openupgrade_scripts/scripts/hr_skills/19.0.1.0/post-migration.py
hbrunn: we don't know that. and hr_employee_skill_log only contains the last skill, so no way to deduce end times from start times of the next skill level

openupgrade_scripts/scripts/hr_skills/19.0.1.0/post-migration.py
hbrunn: the field is called job_skill_ids

openupgrade_scripts/scripts/hr_skills/19.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,125 @@
...
+hr_skills    / res.users                / employee_skill_ids (one2many) : module is now 'project_hr_skills' ('hr_skills')
+hr_skills_survey / hr.resume.line           / department_id (many2one)      : not stored anymore
+
+# NOTHING TO DO
+
+---XML records in module 'hr_skills'---
+NEW hr.resume.line.type: hr_skills.resume_type_training [renamed from hr_skills_slides module]
+DEL hr.resume.line.type: hr_skills.resume_type_social_media
+NEW ir.actions.act_window: hr_skills.action_hr_employee_skill_certification
+NEW ir.actions.act_window: hr_skills.hr_employee_certification_report_action [renamed from hr_skills_survey module] (deleted domain)
```
MiquelRForgeFlow, line 86: In case `hr_skills_survey` was installed, you have to remove the `domain`.

hbrunn: thanks, I'm doing this unconditionally now

MiquelRForgeFlow (approved):

openupgrade_scripts/scripts/hr_skills/19.0.1.0/post-migration.py
pedrobaeza: That can be arguable, specially for progressive skills, example language level, you are L1 until you are L2, and then up to L3, and so on. But let's continue for now, and I'll dig more on that one.

pedrobaeza (approved):

PR #5748, opened 2026-06-23
[19.0][MIG] event

hbrunn:


hbrunn:
/ocabot migration event

<dependency>Depends on : 

- barcodes : #5547
- base_setup : #5466
- mail : #5466
- phone_validation : #5566
- portal : #5501
- utm : #5580</dependency>

hbrunn:
tests fixed in https://github.com/OCA/openupgradelib/pull/456

openupgrade_scripts/scripts/event/19.0.1.9/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,126 @@
+---Models in module 'event'---
+new model event.mail.slot
+new model event.slot
+
+# NOTHING TO DO: new funcitonality
+
+---Fields in module 'event'---
+event        / event.event              / badge_format (selection)      : selection_keys removed: [96x134, 96x82]
+
+# DONE: mapped keys to four_per_sheet
```
pedrobaeza: Apart from the same comment of the other PR about letting an empty line in between, this comment misses if it's done in pre or post-migration, as we have done in other PRs.

openupgrade_scripts/scripts/event/19.0.1.9/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+        env["event.question"],
+        "event_question",
+        "event_type_ids",
+        "event_type_id",
+    )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(env, "event", "19.0.1.9/noupdate_changes.xml")
```
pedrobaeza: I think the switch of the cron to 24 should be mitigated, as we may change the cron period or unit previously.

And I miss the changes for `event_question_name` and company for the new fields `is_default` and `is_reusable` that are in https://github.com/odoo/odoo/blob/19.0/addons/event/data/event_question_data.xml

pedrobaeza (changes requested):

openupgrade_scripts/scripts/event/19.0.1.9/upgrade_analysis_work.txt
hbrunn: added where appropriate

openupgrade_scripts/scripts/event/19.0.1.9/post-migration.py
hbrunn: removed the cron record and added code that adjusts the interval if default settings were kept

the whole concept of default questions is new, before they were produced in [code](https://github.com/odoo/odoo/blob/18.0/addons/event/models/event_event.py#L55-L60), so I don't think there's anything to do about that

pedrobaeza (approved):
Although this is still not the exact expected format in work file, it's not reasonable to block it. Let's move on.

PR #5749, opened 2026-06-24
[17.0][FIX] stock: picked not synchronized between stock.move and stock.move.line

CarlosRoca13:
When the stock.move has stock.move.line related, the value of the stock.move was not set to False but it can have stock.move.line with the falsy value, so check if the value of the lines are False so set the value of the move.

cc @Tecnativa TT63355

ping @pedrobaeza @carlosdauden


carlosdauden (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

pedrobaeza (approved):

PR #5751, opened 2026-06-24
[19.0][OU-ADD] maintenance,stock_maintenance: Migration scripts

carlos-lopez-tecnativa:
TT58667


legalsylvain:
/ocabot migration maintenance


<dependency>Depends on : 

- stock : #5547
- maintenance : #5751</dependency>

openupgrade_scripts/scripts/maintenance/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,52 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE maintenance_request mr
+        SET schedule_end = schedule_date + INTERVAL '1 hour'
+        WHERE mr.schedule_date IS NOT NULL AND COALESCE(mr.duration, 0) = 0
+        """,
+    )
```
pedrobaeza, line 36: You can do it in one shot:

```suggestion
        UPDATE maintenance_request mr
        SET schedule_end = schedule_date + INTERVAL '1 hour' * COALESCE(mr.duration, 1)
        WHERE mr.schedule_date IS NOT NULL
        """,
    )
```

openupgrade_scripts/scripts/maintenance/19.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,33 @@
...
+
+maintenance  / maintenance.equipment    / match_serial (boolean)        : module is now 'stock_maintenance' ('maintenance')
+# NOTHING TO DO: It was simply moved to the stock_maintenance module.
+
+maintenance  / maintenance.equipment.category / alias_id (many2one)           : DEL relation: mail.alias, required
+# DONE: post-migration, delete the mail.alias records related to maintenance.equipment.category
+
+maintenance  / maintenance.equipment.category / _inherits                     : DEL _inherits: {'mail.alias': 'alias_id'}, stored: False
+maintenance  / maintenance.equipment.category / message_follower_ids (one2many): DEL relation: mail.followers
+maintenance  / maintenance.equipment.category / message_ids (one2many)        : DEL relation: mail.message
```
pedrobaeza, line 14: This can go with the previous `DONE` line.

pedrobaeza (changes requested):

openupgrade_scripts/scripts/maintenance/19.0.1.0/pre-migration.py
carlos-lopez-tecnativa: I adjusted the query, but `COALESCE` only takes effect when the value is `NULL`. In the database, the value can also be 0, so that case was not being handled correctly. I have now taken that into account and consolidated the logic into a single query.

Please let me know if this works for you.

pedrobaeza (approved):

PR #5754, opened 2026-06-26
[17.0][FIX] don't set install_filename when loading views from openupgrade_scripts

aurelien-tempo:
Backport changes from https://github.com/OCA/OpenUpgrade/pull/5445 on 18.0 to OpenUpgrade 17.0 through a cherry-pick of the commit from 18.0.
As commented in https://github.com/OCA/OpenUpgrade/issues/5437, I got the issue when migrating the portal module from 16.0 to 17.0.
Cherry-picking the commit on local fixed the issue for me.

### DIFF
--- a/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
+++ b/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
@@ -5,6 +5,7 @@ import logging
 
 from odoo import api
 from odoo.exceptions import ValidationError
+from odoo.modules.module import get_resource_from_path
 from odoo.tools import mute_logger
 
 from odoo.addons.base.models.ir_ui_view import NameManager, View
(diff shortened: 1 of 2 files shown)


openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
```diff
@@ -85,6 +86,21 @@ def _check_field_paths(self, node, field_paths, model_name, use):
...
+    """
+    Remove install_filename from context if it's from openupgrade_scripts.
+    Without this, arch_fs will point to openupgrade_scripts' file which most likely
+    won't exist when the migrated database is deployed, which breaks resetting views
+    """
+    if "install_filename" in self._context:
+        path_info = get_resource_from_path(self._context["install_filename"])
+        if path_info[0] == "openupgrade_scripts":
+            self = self.with_context(
+                **{k: v for k, v in self._context.items() if k != "install_filename"}
```
hbrunn: this makes this monkey patch basically a no-op. remove this commit, then with what you did just now you should be fine

hbrunn (changes requested):

openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_ui_view.py
aurelien-tempo: Thank  you, I didn't understand that exploding it like suggested by the Pylint warning would change the behavior when removing elements from the context. I reverted the change and disabled the warning.

aurelien-tempo:
Changes done. I can squash before or after approval, as you prefer @hbrunn .

hbrunn:
please do the squashing

hbrunn (approved):

pedrobaeza (approved):

pedrobaeza:
This contains a bug fixed in #5764

aurelien-tempo:
I didn't see the fix that was already in the 18.0 and 19.0 branches, sorry

PR #5708, opened 2026-06-10
[19.0][OU-ADD] gamification: Migration scripts

pilarvargas-tecnativa:
@Tecnativa TT58621

PR #5760, opened 2026-06-26
[19.0][OU-ADD] survey: Migration scripts

carlos-lopez-tecnativa:
TT63304


legalsylvain:
/ocabot migration survey

<dependency>Depends on : 

- auth_signup : #5501
- mail : #5466
- web_tour : #5466
- gamification : #5708
- http_routing : TODO</dependency>

eduezerouali-tecnativa (approved):

openupgrade_scripts/scripts/survey/19.0.3.7/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Models in module 'survey'---
+---Fields in module 'survey'---
+survey       / survey.survey            / lang_ids (many2many)          : NEW relation: res.lang, hasdefault: default
+survey       / survey.user_input        / lang_id (many2one)            : NEW relation: res.lang
+# NOTHING TO DO: Leave lang_ids empty to use all installed languages.
```
pedrobaeza: This is not what the default does, as it assigns in `lang_ids` the first installed language found:

https://github.com/odoo/odoo/blob/161715c850496d3683baa5d1600380470d0b5ff5/addons/survey/models/survey_survey.py#L49

In post-migration, just delete whole m2m table for getting the expected result.

pedrobaeza (changes requested):
@eduezerouali-tecnativa please perform the changes, as Carlos is away these days.

openupgrade_scripts/scripts/survey/19.0.3.7/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from openupgradelib import openupgrade
+
+from odoo import Command
+
+
+def _empty_lang_ids(env):
+    surveys = env["survey.survey"].search([("lang_ids", "!=", False)])
```
pedrobaeza: Do it by SQL, please. Always SQL by default unless it's not possible (or very difficult) to recreate it by SQL.

hbrunn: shouldn't we be able to avoid this being filled in the first place by precreating the field?

eduezerouali-tecnativa: Done

pedrobaeza: @hbrunn it's not a field, but a m2m relation table, and pre-create it is not direct. I remember to call an internal "initialize" method of the field to get the same, but I don't have it right now. If you know that code, then I prefer of course to do it on pre.

hbrunn: ah, I thought I remembered openupgradelib creates the table, but it doesn't. also not trivial to just call some function, but I think we can just copy the [statement](https://github.com/odoo/odoo/blob/19.0/odoo/orm/fields_relational.py#L1339-L1349) and call it in pre-migration.

pedrobaeza: And there's still the update of the FK, so I think it's better to leave it this way.

hbrunn (approved):

openupgrade_scripts/scripts/survey/19.0.3.7/post-migration.py
eduezerouali-tecnativa: Are we 100% sure default is trigger on m2m?

hbrunn: you're working on it, so try it?

pedrobaeza (approved):

openupgrade_scripts/scripts/survey/19.0.3.7/post-migration.py
eduezerouali-tecnativa: Can confirm that the M2M default isn't being called, at least in update_db. I did a quick test by adding this field to an existing model, and the field showed "No records" for all existing records, indicating the default wasn't applied. It is, however, triggered when creating a new record, which is the expected behavior.

pedrobaeza: Ok, thanks for confirming. I would say that can be considered a fault in the ORM that someday may be fixed, so it's not bad to keep it this cautious deleting.

PR #5762, opened 2026-06-29
[19.0][OU-IMP] base: Concat phone and mobile

christian-ramos-tecnativa:
`mobile` field no longer exist in `res.partner`, let's concatenate the `mobile` value with the `phone` for those users who use `mobile` instead. In case there is no phone avoid adding the " | " in the middle.

@Tecnativa
@pedrobaeza @eduezerouali-tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/base/19.0.1.3/post-migration.py
+++ b/openupgrade_scripts/scripts/base/19.0.1.3/post-migration.py
@@ -96,6 +96,23 @@ def _init_default_user_group(env):
     )
 
 
+def _migrate_partner_mobile(env):
+    openupgrade.copy_columns(env.cr, {"res_partner": [("phone", None, None)]})
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE res_partner
+        SET phone = CONCAT(
+            COALESCE(phone, ''),
+            CASE WHEN COALESCE(phone, '') != '' THEN ' // ' ELSE '' END,
+            mobile
+        )
+        WHERE COALESCE(mobile, '') != ''
+        AND COALESCE(phone, '') != COALESCE(mobile, '')
+        """,
+    )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.load_data(
@@ -114,3 +131,4 @@ def migrate(env, version):
     _ir_actions_server_html_value(env)
     _ir_filters_user_ids(env)
     _res_lang(env)
+    _migrate_partner_mobile(env)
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Make a copy of the column before doing the concatenation to keep the original contents.

openupgrade_scripts/scripts/base/19.0.1.3/post-migration.py
```diff
@@ -96,6 +96,21 @@ def _init_default_user_group(env):
...
 
 
+def _migrate_partner_mobile(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE res_partner
+        SET phone = CONCAT(
+            COALESCE(phone, ''),
+            CASE WHEN COALESCE(phone, '') != '' THEN ' | ' ELSE '' END,
```
MiquelRForgeFlow:
```suggestion
            CASE WHEN COALESCE(phone, '') != '' THEN ', ' ELSE '' END,
```

MiquelRForgeFlow (changes requested):
You should check if both mobile and phone are the same. If that's the case, then the concatenation is not needed.

By the way, I am not sure if the concatenation is the right way to proceed. If `phone` is missing, and `mobile` is filled, then it's fine to to fill `phone` with `mobile`. But if both are filled, then the field `phone` filled with the concatenation will no longer be operational. If someone has a custom module that uses the `phone` field to automatically call or send sms, then it will fail getting that number.

What about `phone_validation` module?

christian-ramos-tecnativa:
Thinking about this, as @MiquelRForgeFlow pointed out the concatenation result will never be a valid phone number, so Im proposing the next solution:
- If no mobile: nothing to do
- If no phone and mobile -> copy mobile into phone field
- If phone and mobile -> create a child contact address with same name plus "(mobile)" but mobile number
WDYT? @pedrobaeza @MiquelRForgeFlow

pedrobaeza:
I prefer the concatenation. That field is usually a reference one for viewing them, not for automation like VoIP. If so, extra handling will be needed, but can be done later in custom. Put a separator like double `//` for being easily extracted.

MiquelRForgeFlow:
Ok, but remember to update `upgrade_analysis_work.txt` correspondingly.

christian-ramos-tecnativa:
Added `copy_columns`, make sure concatenation is done when `phone` and `mobile` are different, changed separator to `//` and updated `upgrade_analysis_work.txt`

eduezerouali-tecnativa (approved):

pedrobaeza (approved):

MiquelRForgeFlow (approved):

PR #5763, opened 2026-06-29
[19.0][OU-IMP] crm: Add separator to phone and mobile concatenation

christian-ramos-tecnativa:
Depending on how the user wrote down the phone and mobile fields, after the migration the phone can get ambiguous. Added a separator to avoid it.

@Tecnativa 
@pedrobaeza @eduezerouali-tecnativa

### DIFF
--- a/openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
+++ b/openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
@@ -4,12 +4,18 @@ from openupgradelib import openupgrade
 
 
 def _migrate_lead_mobile(env):
+    openupgrade.copy_columns(env.cr, {"crm_lead": [("phone", None, None)]})
     openupgrade.logged_query(
         env.cr,
         """
         UPDATE crm_lead
-        SET phone = CONCAT(COALESCE(phone, ''), mobile)
+        SET phone = CONCAT(
+            COALESCE(phone, ''),
+            CASE WHEN COALESCE(phone, '') != '' THEN ' // ' ELSE '' END,
+            mobile
+        )
         WHERE COALESCE(mobile, '') != ''
+        AND COALESCE(phone, '') != COALESCE(mobile, '')
         """,
     )
 


pedrobaeza (commented):
As you are here, please make a copy of the column first for preserving original information.

christian-ramos-tecnativa:
Do you mean creating an extra column like `original_phone` with the info in `phone` column?

pedrobaeza:
Use `copy_columns` method, and pass `None` to get an automatic backup column name.

MiquelRForgeFlow (changes requested):
Same comments as https://github.com/OCA/OpenUpgrade/pull/5762. What about `phone_validation` module?

pedrobaeza:
`phone_validation` has nothing to do with this, as the formatting was already done.

MiquelRForgeFlow:
I meant that what we finally do in https://github.com/OCA/OpenUpgrade/pull/5762, we have to it here too.

pedrobaeza:
Yes, but the problem you state with `phone_validation` in both PR is not a problem.

christian-ramos-tecnativa:
Added `copy_columns`, make sure concatenation is done when `phone` and `mobile` are different and changed separator to `//`

openupgrade_scripts/scripts/crm/19.0.1.9/post-migration.py
```diff
@@ -4,12 +4,18 @@
 
 
 def _migrate_lead_mobile(env):
+    openupgrade.copy_columns(env.cr, {"crm_lead": [("mobile", None, None)]})
```
pedrobaeza: The one to copy is `phone`, not `mobile`.

christian-ramos-tecnativa: True! changed

eduezerouali-tecnativa (approved):

pedrobaeza (approved):

MiquelRForgeFlow (approved):

PR #5765, opened 2026-07-01
[19.0][IMP] base: Load additional renames/merges from env.

david-banon-tecnativa:
With this change, merging and renaming private modules no longer needs to be done in a custom module, and now can happen at the same time as OCA modules.
To use this feature:

When running openupgrade, set the OPENUPGRADE_MERGED_MODULES and OPENUPGRADE_RENAMED_MODULES environment variables to raw JSON data containing the private renames/merges in the same format as in apriori.py .
Example OPENUPGRADE_MERGED_MODULES to merge "custom module" into "helpdesk_mgmt".
`{"custom_module":"helpdesk_mgmt"}`

No changes should be made on current workflows if this feature is not used.

Once this PR gets approved i'll backport to all versions of OU

@pedrobaeza @victoralmau 
@tecnativa TT63118

### DIFF
--- a/openupgrade_scripts/scripts/base/19.0.1.3/pre-migration.py
+++ b/openupgrade_scripts/scripts/base/19.0.1.3/pre-migration.py
@@ -170,8 +170,12 @@ def migrate(env, version):
         } AS (SELECT name, state FROM ir_module_module);
         """,
     )
-    openupgrade.update_module_names(env.cr, renamed_modules.items())
-    openupgrade.update_module_names(env.cr, merged_modules.items(), merge_modules=True)
+    openupgrade.update_module_names(
+        env.cr, renamed_modules.items(), environment_namespec=True
+    )
+    openupgrade.update_module_names(
+        env.cr, merged_modules.items(), merge_modules=True, environment_namespec=True
+    )
     openupgrade.clean_transient_models(env.cr)
     openupgrade.rename_xmlids(env.cr, _renamed_xmlids)
     openupgrade.rename_xmlids(env.cr, _merged_xmlids, allow_merge=True)


pedrobaeza (approved):

hbrunn (changes requested):
I would prefer to see this in openupgradelib, otherwise we have to reimplement this for every version

pedrobaeza:
@hbrunn that's not possible, as you may call `update_module_names` multiple times over whole OpenUpgrade migration scripts, and add there all the injected environment variables is not correct IMO.

hbrunn:
so add a parameter enabling this which defaults to False

pedrobaeza:
> so add a parameter enabling this which defaults to False

And we will need to modify the same each OpenUpgrade version to add here such parameter...

hbrunn:
true, but we don't need to copy&paste a bunch of logic

hbrunn:
another option would be to have a function in openupgrade_scripts itself that does the initial renaming and merging, which the base migration just calls, and your code happens there. that then also goes automatically to all new versions

pedrobaeza:
OK, let's do it that way. @david-banon-tecnativa do you have clear what to do? The keyword argument can be `use_environment_arg=False`, and when True, and depending on `merge_modules` variable, read one or the other.

pedrobaeza:
> another option would be to have a function in openupgrade_scripts itself that does the initial renaming and merging, which the base migration just calls, and your code happens there. that then also goes automatically to all new versions

You wrote meanwhile I was writing the other. The problem with this suggestion is that there's no apriori.py file in openupgralib to import, so I don't think we can move such code.

openupgrade_scripts/scripts/base/19.0.1.3/pre-migration.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2025 Hunki Enterprises BV
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+
```
pedrobaeza: Don't add here extra line.

hbrunn (approved):
/ocabot merge nobump

hbrunn:
@david-banon-tecnativa please document the env vars somewhere in https://github.com/OCA/OpenUpgrade/tree/documentation

hbrunn:
https://github.com/OCA/OpenUpgrade/blob/documentation/docsource/040_run_migration.rst#configuration-options probably is the right place

david-banon-tecnativa:
Ok, OMW

PR #5782, opened 2026-07-07
[19.0][MIG] im_livechat

hbrunn:


hbrunn:
/ocabot migration im_livechat

<dependency>Depends on : 

- mail : #5466
- rating : #5572
- digest : #5505
- utm : #5580</dependency>

openupgrade_scripts/scripts/im_livechat/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def chatbot_script_step_message(env):
+    """
+    Convert chatbot.script.step#message to html
+    """
+    openupgrade.copy_columns(env.cr, {"chatbot_script_step": [("message", None, None)]})
+    openupgrade.convert_field_to_html(
+        env.cr, "chatbot_script_step", "message", "message", translate=True
+    )
```
MiquelRForgeFlow, line 13: I think the `copy_columns` part should be done in pre-migration, and then here you should do
```suggestion
    openupgrade.convert_field_to_html(
        env.cr, "chatbot_script_step", "message", openupgrade.get_legacy_name("message"), translate=True
    )
```

Right?

hbrunn: why does the copying need to go to pre-migration? no other code touches this.
and [convert_field_to_html](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2751) works fine inplace, no need to work on the copy

MiquelRForgeFlow (approved):

PR #5821, opened 2026-07-15
[19.0][MIG] base_install_request

MiquelRForgeFlow:
.

PR #5822, opened 2026-07-15
[19.0][MIG] snailmail_account: nothing to do

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration snailmail_account

<dependency>Depends on : 

- account : #5509
- snailmail : #5821</dependency>

PR #5772, opened 2026-07-07
[19.0][MIG] sms

hbrunn:

PR #5529, opened 2026-02-23
[19.0][MIG] sale

hbrunn:
depends:

- [x] https://github.com/OCA/OpenUpgrade/pull/5527
- [x] https://github.com/OCA/OpenUpgrade/pull/5528

PR #5824, opened 2026-07-15
[19.0][MIG] pos_sale

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration pos_sale

<dependency>Depends on : 

- sale : #5529
- sms : #5772</dependency>

PR #5827, opened 2026-07-15
 [19.0][OU-ADD] membership: Migration scripts

carlos-lopez-tecnativa:
This module was removed from Odoo in https://github.com/odoo/odoo/pull/197228, but it has now been backported to OCA in https://github.com/OCA/vertical-association/pull/224. Therefore, it should be removed from the `renamed_modules` variable.

TT63407
@Tecnativa @pedrobaeza @pilarvargas-tecnativa @MiquelRForgeFlow @hbrunn could you please review this?


pedrobaeza (approved):

MiquelRForgeFlow:
It seems to me a bad practice merging this before merging https://github.com/OCA/vertical-association/pull/224.

pedrobaeza:
It's not the same case, as the renaming is simply wrong.

PR #5829, opened 2026-07-16
[19.0][MIG] hr_calendar

mihien:


MiquelRForgeFlow:
/ocabot migration hr_calendar

<dependency>Depends on : 

- hr : #5566
- calendar : #5613</dependency>

PR #5833, opened 2026-07-17
fixup! [MIG] hr_calendar

mihien:
Missed the deletion of two records


legalsylvain:
/ocabot migration hr_calendar

<dependency>Depends on : 

- hr : #5566
- calendar : #5613</dependency>

pedrobaeza:
@legalsylvain please don't mark PRs that are not the main migration ones.

@mihien this is not correct. The views with noupdate=0 are autoremoved by the ORM, so closing this.

PR #5836, opened 2026-07-21
[19.0][MIG] fleet

hbrunn:


hbrunn:
/ocabot migration fleet

<dependency>Depends on : 

- base : #5446
- mail : #5466</dependency>

mihien (approved):
Reviewed and tested, everything was correct.

Also learned a good lesson during my testing: don't rely on demo data otherwise you end up with false positive on missing data after migration, if a demo partner is not present in 19 anymore for example.

pedrobaeza (approved):

PR #5842, opened 2026-07-22
[DOCS][IMP] run_migration: Recommended command

SirPyTech:
The `--upgrade-path` option is automatically added in https://github.com/OCA/OpenUpgrade/blob/439f5a570b06d8486328a88ace98c992ba12b9b6/openupgrade_framework/__init__.py#L9-L16.

I have always been using the `--upgrade-path` option because it is the "Recommended command", and then a few days ago @HekkiMelody points me out this piece of code :eyes:
Should we change the "Recommended command"?


hbrunn (approved):

legalsylvain (approved):
Thanks !

PR #5845, opened 2026-07-24
[18.0][OU-IMP] account: create batch payment sequence

aurelien-tempo:
When no batch payment sequence exists for a company, an error occurs when trying to do a batch payment.
With https://github.com/odoo/odoo/pull/273905, Odoo now creates the sequence on account module installation.
But it isn't created when migrating from a 17.0 to a 18.0 database.
This change would now create the missing sequences during the migration.

### DIFF
--- a/openupgrade_scripts/scripts/account/18.0.1.3/post-migration.py
+++ b/openupgrade_scripts/scripts/account/18.0.1.3/post-migration.py
@@ -208,12 +208,29 @@ def _handle_outstanding_accounts(env):
             )
 
 
+def _create_batch_payment_sequence(env):
+    """Creates a Batch Payment Number Sequence for every company that does not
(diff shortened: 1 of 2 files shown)


hbrunn (approved):

hbrunn (changes requested):
please add a remark about this in https://github.com/aurelien-tempo/OpenUpgrade/blob/6ce95dc4b9e64fad3292e01ccfc39428ee596e3d/openupgrade_scripts/scripts/account/18.0.1.3/upgrade_analysis_work.txt#L218

aurelien-tempo:
Remark added, thanks for the review
Edit : fixed the typo in the remark

pedrobaeza (approved):

PR #5847, opened 2026-07-28
[19.0][MIG] hr_recruitment_skills

hbrunn:


hbrunn:
/ocabot migration hr_recruitment_skills

<dependency>Depends on : 

- hr_skills : #5736
- hr_recruitment : #5612</dependency>

BhaveshHeliconia (approved):

openupgrade_scripts/scripts/hr_recruitment_skills/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, copy_columns)
+    openupgrade.rename_models(env.cr, renamed_models)
+    openupgrade.rename_tables(env.cr, renamed_tables)
+    openupgrade.rename_fields(env, renamed_fields)
+    hr_applicant_skill(env)
+    hr_applicant_skill_ids(env)
```
MiquelRForgeFlow, line 79: In v17, it already existed `hr_applicant_skill`. I mean, the v19 is like v17. See https://github.com/OCA/OpenUpgrade/commit/a169af6c2946450e3e75c3ef98738c641d0f771e. My point is: I think that the code should be adapted in a way that people coming from v17 using OU don't get problems, or have a more performance-wise flow, if they have not deleted yet old table.

MiquelRForgeFlow (changes requested):

openupgrade_scripts/scripts/hr_recruitment_skills/19.0.1.0/pre-migration.py
hbrunn: good catch, I'm checking for this table existing in 820aed28998c now (will squash when approved). But I don't see how we can differentiate in a straightforward way between a database that is directly migrating from v17, and one that has been migrated to v18, used for a while, then migrated to v19. Not really worth the trouble I think.

MiquelRForgeFlow (approved):

PR #5854, opened 2026-07-29
[19.0][OU-IMP] hr: Set salary_distribution

christian-ramos-tecnativa:
After migration if you try to enter in a employee with `bank_account_ids` it will raise the next error:
```
File "/opt/odoo/auto/addons/hr/models/hr_employee.py", line 1828, in _compute_primary_bank_account_id
primary_account = min(
                                  ^^^^
File "/opt/odoo/auto/addons/hr/models/hr_employee.py", line 1830, in <lambda>
 key=lambda acc: employee.salary_distribution.get(str(acc.id), {}).get("sequence", float("inf")),
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'bool' object has no attribute 'get'
```
Force the compute of `salary_distribution` to avoid it

@Tecnativa TT62941

### DIFF
--- a/openupgrade_scripts/scripts/hr/19.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/hr/19.0.1.1/post-migration.py
@@ -101,6 +101,21 @@ def hr_employee_bank_account_ids(env):
     openupgrade.m2o_to_x2m(
         env.cr, env["hr.employee"], "hr_employee", "bank_account_ids", "bank_account_id"
     )
+    # Set salary distribution to avoid error on _compute_primary_bank_account_id
+    env.cr.execute(
+        """
+        UPDATE hr_employee
+        SET salary_distribution = json_build_object(
+            bank_account_id::text,
+            json_build_object(
+                'amount', 100.0,
+                'sequence', 1,
+                'amount_is_percentage', true
+            )
+        )
+        WHERE bank_account_id IS NOT NULL
+        """
+    )
 
 
 def hr_employee_notes(env):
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
This process seems heavy. Can't you do it by SQL?

christian-ramos-tecnativa:
Done!

openupgrade_scripts/scripts/hr/19.0.1.1/post-migration.py
```diff
@@ -101,6 +101,21 @@ def hr_employee_bank_account_ids(env):
...
+        UPDATE hr_employee
+        SET salary_distribution = json_build_object(
+            bank_account_id::text,
+            json_build_object(
+                'amount', 100.0,
+                'sequence', 1,
+                'amount_is_percentage', true
+            )
+        )
+        WHERE bank_account_id IS NOT NULL
```
pedrobaeza, line 116: Should the rest be filled with an empty JSON dictionary?

christian-ramos-tecnativa: AFAIK there should be no problem for the NULL values as it will get filled when the bank_account_ids is updated

pedrobaeza (approved):

pedrobaeza (commented):
One more thing: shouldn't you update the analysis work file for that field?

pedrobaeza (approved):
/ocabot merge nobump

christian-ramos-tecnativa:
Done. I'll keep it in mind for next time 😅​

PR #5820, opened 2026-07-15
[19.0][MIG] auth_passkey

MiquelRForgeFlow:
- [x] Depends on https://github.com/OCA/OpenUpgrade/pull/5828.

PR #5858, opened 2026-07-30
[19.0][MIG] auth_passkey_portal: nothing to do

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration auth_passkey_portal

<dependency>Depends on : 

- auth_passkey : #5820
- portal : #5501</dependency>

PR #5860, opened 2026-07-30
[19.0][MIG] hr_skills_event: nothing to do

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration hr_skills_event

<dependency>Depends on : 

- hr_skills : #5736
- event : #5748</dependency>

PR #5783, opened 2026-07-08
[19.0][OU-ADD] website_slides: Migration scripts

pilarvargas-tecnativa:
@Tecnativa TT63428

PR #5585, opened 2026-04-13
[19.0][MIG] website_sale + dependencies

hbrunn:

PR #5863, opened 2026-07-30
[19.0][MIG] website_sale_slides: nothing to do

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration website_sale_slides

<dependency>Depends on : 

- website_slides : #5783
- website_sale : #5585</dependency>

PR #5866, opened 2026-07-31
[19.0][OU-ADD] crm_sms: Nothing to do

BhaveshHeliconia:


legalsylvain:
/ocabot migration crm_sms

<dependency>Depends on : 

- crm : #5706
- sms : #5772</dependency>

PR #5868, opened 2026-07-31
[19.0][OU-ADD] pos_self_order: Migration scripts

BhaveshHeliconia:


openupgrade_scripts/scripts/pos_self_order/19.0.1.0/upgrade_analysis_work.txt
```diff
@@ -0,0 +1,33 @@
...
+# NOTHING TO DO
+
+pos_self_order / pos.category             / pos_config_ids (many2many)    : NEW relation: pos.config
+# NOTHING TO DO
+
+pos_self_order / pos.config               / self_ordering_image_background_ids (many2many): NEW relation: ir.attachment
+# NOTHING TO DO
+
+pos_self_order / pos.config               / self_ordering_takeaway (boolean): DEL
+# DONE: preserved in pre-migration.py and used in post-migration.py to conditionally link takeout preset.
```
hbrunn: I don't see why you need to copy this, you can just keep the original name

openupgrade_scripts/scripts/pos_self_order/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,84 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def link_presets_conditionally(env):
+    """
+    Link default presets to existing POS configurations
+    based on legacy self_ordering_takeaway setting.
+    """
```
hbrunn: I think most of this should go to the migration script of pos_restaurant

hbrunn (changes requested):

hbrunn:
/ocabot migration pos_self_order
/ocabot migration pos_restaurant

<dependency>Depends on : 

- http_routing : #5501
- link_tracker : #5696
- pos_restaurant : #5868</dependency>

BhaveshHeliconia:
@hbrunn Thanks for the review!

I have addressed both comments. I removed the unnecessary pre-migration.py from pos_self_order and kept the original self_ordering_takeaway column name for the post-migration.

I also moved the preset linking logic to pos_restaurant, where pos.preset is defined. I have kept the pos_restaurant changes in a separate commit within this PR.

I also tested the migration with different POS configurations and verified that the presets and fiscal positions are correctly assigned.

Please let me know if you would prefer the pos_restaurant changes to be moved to a separate PR.

MiquelRForgeFlow:
@BhaveshHeliconia  please, rebase. I made an error when force pushing to 19.0 (it's already amended)

BhaveshHeliconia:
> @BhaveshHeliconia please, rebase. I made an error when force pushing to 19.0 (it's already amended)

Done. Thanks!

pedrobaeza:
@hbrunn @MiquelRForgeFlow is it OK for you?

PR #5879, opened 2026-08-04
[19.0][OU-ADD] product_expiry: Migration scripts

BhaveshHeliconia:


openupgrade_scripts/scripts/product_expiry/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,43 @@
...
+    old_type_id = row_old[0] if row_old else None
+
+    env.cr.execute(
+        """
+        SELECT res_id FROM ir_model_data
+        WHERE module = 'mail' AND name = 'mail_activity_data_todo'
+        """
+    )
+    row_new = env.cr.fetchone()
+    new_type_id = row_new[0] if row_new else None
```
hbrunn: use `env.ref(..., False)`

MiquelRForgeFlow:
/ocabot migration product_expiry

<dependency>Depends on : 

- stock : #5547</dependency>

openupgrade_scripts/scripts/product_expiry/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,52 @@
...
+    env.cr.execute(
+        """
+        UPDATE stock_move_line sml
+        SET removal_date = CASE
+            WHEN lot.removal_date IS NOT NULL
+            THEN lot.removal_date
+
+            WHEN COALESCE(spt.use_create_lots, FALSE) IS TRUE
+                 AND COALESCE(pt.use_expiration_date, FALSE) IS TRUE
+                 AND sml2.expiration_date IS NOT NULL
```
hbrunn: as you already have this in the conditions below, I think this can be the else branch, as the current else branch shouldn't ever be reached

hbrunn (changes requested):

openupgrade_scripts/scripts/product_expiry/19.0.1.0/post-migration.py
hbrunn: btw I'm not sure if this conditional is faster than two queries, did you test this?

openupgrade_scripts/scripts/product_expiry/19.0.1.0/pre-migration.py
BhaveshHeliconia: Done, thanks for pointing this out and for the review!

openupgrade_scripts/scripts/product_expiry/19.0.1.0/post-migration.py
BhaveshHeliconia: Tested both approaches with a 10k-row dataset using EXPLAIN ANALYZE, and the performance is basically the same. The single query avoids a second scan of stock_move_line, while the split queries process fewer rows individually, so they end up being almost identical.

I am happy to split them into two queries if you prefer it for readability, but from a performance perspective, there is not much difference.

hbrunn (approved):

openupgrade_scripts/scripts/product_expiry/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,47 @@
...
+    column empty to avoid a MemoryError on large tables. Mirrors
+    _compute_removal_date (lot priority, use_create_lots / use_expiration_date
+    gating, picking_id -> move_id fallback for picking type)."""
+    env.cr.execute(
+        """
+        UPDATE stock_move_line sml
+        SET removal_date = CASE
+            WHEN lot.removal_date IS NOT NULL
+            THEN lot.removal_date
+
```
MiquelRForgeFlow:
```suggestion
```

openupgrade_scripts/scripts/product_expiry/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
...
+def migrate(env, version):
+    reassign_alert_date_activities(env)
+
+
+def reassign_alert_date_activities(env):
+    """Reassign alert-date activities to the standard To-do type
+    before the old activity type record is removed by the data sync."""
+    old_type = env.ref("product_expiry.mail_activity_type_alert_date_reached", False)
+    new_type = env.ref("mail.mail_activity_data_todo", False)
+
```
MiquelRForgeFlow:
```suggestion
```

openupgrade_scripts/scripts/product_expiry/19.0.1.0/post-migration.py
BhaveshHeliconia: Done. Thanks!

MiquelRForgeFlow (approved):

PR #5881, opened 2026-08-04
[19.0][FIX] stock_account: product.value and stock.valuation.layer have different semantics

hbrunn:
this deals with the issue hinted at in #5855

### DIFF
--- a/openupgrade_scripts/scripts/stock_account/19.0.1.1/post-migration.py
+++ b/openupgrade_scripts/scripts/stock_account/19.0.1.1/post-migration.py
@@ -49,13 +49,14 @@ def stock_location_valuation_account_id(env):
     Set stock.location#valuation_account_id from valuation_in_account_id and
     valuation_out_account_id if they are the same
     """
-    env.cr.execute(
+    openupgrade.logged_query(
+        env.cr,
         """
         UPDATE stock_location
         SET valuation_account_id=valuation_in_account_id
         WHERE
         valuation_in_account_id=valuation_out_account_id
-        """
+        """,
     )
 
 
@@ -63,7 +64,8 @@ def stock_move_value(env):
     """
     Set stock.move#value to sum of product.value#value for this move
     """
-    env.cr.execute(
+    openupgrade.logged_query(
+        env.cr,
         """
         UPDATE stock_move
         SET value=aggregated_values.agg_value
@@ -75,7 +77,7 @@ def stock_move_value(env):
             GROUP BY stock_move_id
         ) aggregated_values
         WHERE aggregated_values.stock_move_id=stock_move.id
-        """
+        """,
     )
 
 
@@ -84,16 +86,14 @@ def product_value(env):
     Fill product.value with stock valuations not assigned to a move
     (=manual valuations)
     """
-    openupgrade.logged_query(
-        env.cr,
+    env.cr.execute(
         """
         ALTER TABLE product_value
         ADD COLUMN IF NOT EXISTS stock_valuation_layer_id int
         """,
     )
     # simple case: the valuation layer has unit_cost set
-    openupgrade.logged_query(
-        env.cr,
+    env.cr.execute(
         """
         INSERT INTO product_value
         (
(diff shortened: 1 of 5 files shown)


openupgrade_scripts/scripts/stock_account/19.0.1.1/post-migration.py
```diff
@@ -84,20 +69,44 @@ def stock_move_value(env):
...
+        """
+    )
+
+
+def product_value(env):
+    """
+    Fill product.value with stock valuations not assigned to a move
+    (=manual valuations)
+    """
+    env.cr.execute(
```
MiquelRForgeFlow, line 87:
```suggestion
    openupgrade.logged_query(env,
```
It would be nice if you could do this everywhere.

hbrunn: better pass a cursor there. I don't find printing the queries in the log helpful, but if it makes the reviewer happy, I'll oblige

MiquelRForgeFlow: In my opinion, SELECT queries should be called with `env.cr.execute` and UPDATE queries should be called with `openupgrade.logged_query`, but really I don't have a strong opinion for INSERT INTO queries. If you think logging INSERT INTO queries adds noise, then use `env.cr.execute`, ok for me.

MiquelRForgeFlow:
BTW, is this PR ready to review and merge?

openupgrade_scripts/scripts/stock_account/19.0.1.1/post-migration.py
MiquelRForgeFlow: logging

MiquelRForgeFlow: logging

openupgrade_scripts/scripts/stock_account/19.0.1.1/post-migration.py
```diff
@@ -84,20 +69,85 @@ def stock_move_value(env):
...
+    )
+
+
+def product_value(env):
+    """
+    Fill product.value with stock valuations not assigned to a move
+    (=manual valuations)
+    """
+    # simple case: the valuation layer has unit_cost set
+    openupgrade.logged_query(
```
MiquelRForgeFlow: yeah, better env.cr.execute here, sorry

hbrunn:
I forgot some details like lifting constraints on the svl table and linking product_value records to the svl they come from, but by and large, that's what I think should happen

MiquelRForgeFlow:
@hbrunn  please, rebase. I made an error when force pushing to 19.0 (it's already amended)

hbrunn:
why is anyone allowed to force push in the first place?

MiquelRForgeFlow:
I have just removed the force push permissions.

hbrunn:
https://github.com/OCA/OpenUpgrade/settings/branch_protection_rules/5942394 looked like this:

image

I've changed that to

image

MiquelRForgeFlow:
Nice.

MiquelRForgeFlow:
The option was marked in "Specify who can force push". I only removed the people there. So I suppose that when the list of people there is empty, github changes it to Everyone.

hbrunn:
> I forgot some details like lifting constraints on the svl table and linking product_value records to the svl they come from, but by and large, that's what I think should happen

that's done now

MiquelRForgeFlow:
Could you attend my comments regarding logging/not logging queries? :)

hbrunn:
done

MiquelRForgeFlow (approved):

cuongnmtm:
My commit inside the PR fixes the wrong product cost issue in my DB 

https://github.com/OCA/OpenUpgrade/pull/5951

cuongnmtm:
If you think it is ok, please cherry-pick so I can close my PR

hbrunn:
the comments your bullshit generator adds keep referring to v17. but we don't care about v17, we care about v18. also see my comment in your PR, which fails the tests added

hbrunn:
@OCA/openupgrade-maintainers please review/merge this as the migration as it is, is very wrong, and there hasn't been any helpful contribution by the original reporter of the issue

pedrobaeza (approved):
OK, let's go with the new approach.

PR #5882, opened 2026-08-05
[19.0][OU-ADD] partnership: Migration scripts

BhaveshHeliconia:


MiquelRForgeFlow:
/ocabot migration partnership

<dependency>Depends on : 

- crm : #5706
- sale : #5529</dependency>

openupgrade_scripts/apriori.py
```diff
@@ -38,6 +38,7 @@
     "l10n_pe_website_sale": "l10n_pe",
     "l10n_pl_taxable_supply_date": "l10n_pl",
     "l10n_ro_efactura_synchronize": "l10n_ro_edi",
+    "membership": "partnership",
```
MiquelRForgeFlow: Don't do this, see: https://github.com/OCA/OpenUpgrade/commit/62e06a458b27eb906137e25dab1c444b5a5fe5d9. Maybe what you have to do instead is in `base` pre-migration, check if membership was installed. If it was installed, mark `partnership` to be installed too. Use queries to do this, in `base` pre-migration you can use `env`.

BhaveshHeliconia: Done. Thanks!

openupgrade_scripts/scripts/partnership/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,65 @@
...
+            (new_mod, new_name),
+        )
+        if cr.fetchone():
+            openupgrade.logged_query(
+                cr,
+                "DELETE FROM ir_model_data WHERE module = %s AND name = %s",
+                (old_mod, old_name),
+            )
+        else:
+            openupgrade.rename_xmlids(cr, [(old_xmlid, new_xmlid)])
```
MiquelRForgeFlow:
```suggestion
    openupgrade.rename_xmlids(cr, xmlid_renames)
```

openupgrade_scripts/scripts/partnership/19.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,65 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+column_renames = {
+    "product_template": [
+        ("membership", "openupgrade_legacy_membership"),
+    ],
+    "res_partner": [
+        ("membership_state", "openupgrade_legacy_membership_state"),
```
MiquelRForgeFlow:
```suggestion
        ("membership", None),
    ],
    "res_partner": [
        ("membership_state", None),
```

openupgrade_scripts/scripts/partnership/19.0.1.0/pre-migration.py
BhaveshHeliconia: Done. Thanks for the suggestion!

openupgrade_scripts/scripts/partnership/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    cr = env.cr
+    if openupgrade.column_exists(
+        cr, "product_template", "openupgrade_legacy_membership"
+    ):
```
MiquelRForgeFlow:
```suggestion
    if openupgrade.column_exists(
        cr, "product_template", openupgrade.get_legacy_name(membership)
    ):
```

openupgrade_scripts/scripts/partnership/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+        openupgrade.logged_query(
+            cr,
+            """
+            UPDATE product_template
+            SET service_tracking = 'partnership'
+            WHERE openupgrade_legacy_membership IS TRUE
```
MiquelRForgeFlow:
```suggestion
            f"""
            UPDATE product_template
            SET service_tracking = 'partnership'
            WHERE {openupgrade.get_legacy_name(membership)}
```

openupgrade_scripts/scripts/partnership/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+            """,
+        )
+        bronze_grade = env.ref("partnership.res_partner_grade_data_bronze", False)
+        if bronze_grade:
+            openupgrade.logged_query(
+                cr,
+                """
+                UPDATE product_template
+                SET grade_id = %s
+                WHERE openupgrade_legacy_membership IS TRUE
```
MiquelRForgeFlow:
```suggestion
                f"""
                UPDATE product_template
                SET grade_id = %s
                WHERE {openupgrade.get_legacy_name(membership)}
```

pedrobaeza (changes requested):
We don't want `partnership` module as replacement of `membership`. They are different things. `membership` has been moved to OCA, so please don't do any of this.

openupgrade_scripts/scripts/partnership/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+                  AND grade_id IS NULL
+                """,
+                (bronze_grade.id,),
+            )
+
+    if openupgrade.table_exists(cr, "membership_membership_line"):
```
MiquelRForgeFlow:
```suggestion
```
not needed, because if the first check passed, then this one passes too.

openupgrade_scripts/scripts/partnership/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+        openupgrade.logged_query(
+            cr,
+            """
+            UPDATE res_partner p
+            SET grade_id = sub.grade_id
+            FROM (
+                SELECT DISTINCT ON (mml.partner)
+                    mml.partner,
```
MiquelRForgeFlow: you should consider the possibility OCA's `membership` was installed before, which renamed the field to `partner_id`.

openupgrade_scripts/scripts/partnership/19.0.1.0/post-migration.py
```diff
@@ -0,0 +1,69 @@
...
+    ):
+        openupgrade.drop_columns(
+            cr, [("product_template", "openupgrade_legacy_membership")]
+        )
+    if openupgrade.column_exists(
+        cr, "res_partner", "openupgrade_legacy_membership_state"
+    ):
+        openupgrade.drop_columns(
+            cr, [("res_partner", "openupgrade_legacy_membership_state")]
+        )
```
MiquelRForgeFlow:
```suggestion
```
not needed. If someone don't want these columns, they can use data clean-up modules.

MiquelRForgeFlow:
> We don't want partnership module as replacement of membership. They are different things. membership has been moved to OCA, so please don't do any of this.

Wrong. This PR is not a replacement. `partnership` can be a dependency of a module you have installed, like `website_crm_partner_assign`. Thus, you must need `partnership` and this PR handles this.

BhaveshHeliconia:
> We don't want `partnership` module as replacement of `membership`. They are different things. `membership` has been moved to OCA, so please don't do any of this.

Thanks for the clarification! @pedrobaeza, I've removed all membership-related migration logic. The PR now only handles the XML ID renames from website_crm_partner_assign to partnership, which is needed since partnership is a dependency of website_crm_partner_assign in 19.0.

MiquelRForgeFlow (approved):

BhaveshHeliconia:
@MiquelRForgeFlow Thanks for the suggestion and guidance!

pedrobaeza (approved):

PR #5877, opened 2026-08-03
[MIG] website_crm_partner_assign

mihien:

PR #5883, opened 2026-08-05
[19.0][OU-ADD] website_crm_partner_assign: Migration scripts

BhaveshHeliconia:


openupgrade_scripts/scripts/website_crm_partner_assign/19.0.1.2/pre-migration.py
```diff
@@ -0,0 +1,55 @@
...
+        "website_crm_partner_assign.res_partner_grade_data_silver",
+        "partnership.res_partner_grade_data_silver",
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    cr = env.cr
+    for old_xmlid, new_xmlid in xmlid_renames:
```
MiquelRForgeFlow, line 41: should be done instead in `partnership` module pre-migration.

MiquelRForgeFlow:
Giving priority to https://github.com/OCA/OpenUpgrade/pull/5877 for being older.

PR #5884, opened 2026-08-05
[FIX] hr_recruitment: fix lift_constraints() constraint drop order for PostgreSQL 17+

quoc-pn:
## Bug

On PostgreSQL 17+, `NOT NULL` is catalogued as a named constraint (`<table>_<column>_not_null`) instead of an implicit attribute flag on `pg_attribute`.

`openupgrade.lift_constraints()` drops all constraints on a column in a single `ALTER TABLE ... DROP CONSTRAINT a, DROP CONSTRAINT b` statement. When the column has both a not-null constraint and a primary key, and the not-null constraint's drop clause is listed before the primary key's, PostgreSQL rejects the statement:

```
ERROR: column "id" is in a primary key
```

PostgreSQL validates each `ALTER TABLE` subcommand against the catalog state *before* the statement runs, not against the pending effects of earlier subcommands in the same statement — so even though the very next clause would drop the primary key, the not-null drop is rejected first.

This surfaced in `hr_recruitment/19.0.1.1/pre-migration.py` when merging `hr.candidate` into `hr.applicant`: after renaming `hr_candidate` to its legacy table, `lift_constraints(cr, legacy_table, "id", cascade=True)` fails on PostgreSQL 17/18 with the error above.

## Fix

Drop the primary key constraint on the legacy table in its own, separate `ALTER TABLE ... DROP CONSTRAINT ... CASCADE` statement *before* calling `lift_constraints()`. By the time `lift_constraints()` runs, only the not-null constraint remains on the column, so there's no ordering conflict.

## Test

Reproduced the failure on a database running PostgreSQL 18.4 during an Odoo 19 upgrade, confirmed the fix resolves it.

### DIFF
--- a/openupgrade_scripts/scripts/hr_recruitment/19.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/hr_recruitment/19.0.1.1/pre-migration.py
@@ -13,10 +13,20 @@ def hr_candidate2hr_applicant(env):
     """
     openupgrade.merge_models(env.cr, "hr.candidate", "hr.applicant", "candidate_id")
     openupgrade.rename_tables(env.cr, [("hr_candidate", None)])
-    openupgrade.lift_constraints(
-        env.cr, openupgrade.get_legacy_name("hr_candidate"), "id", cascade=True
+    legacy_table = openupgrade.get_legacy_name("hr_candidate")
+    # PostgreSQL >= 17 catalogs NOT NULL as a named constraint and refuses to
+    # drop it while the column is still part of a primary key, even when the
+    # pkey drop is queued in the same ALTER TABLE statement (as
+    # openupgrade.lift_constraints does). Drop the pkey first, on its own, so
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't this be fixed in openupgradelib instead?

quoc-pn:
> Shouldn't this be fixed in openupgradelib instead?

@pedrobaeza Thanks for the suggestion. I'll create a PR in **openupgradelib**.

This PR will be closed once the corresponding PR in **openupgradelib** has been created.

quoc-pn:
Closing this in favor of fixing the root cause upstream in openupgradelib itself: lift_constraints() combined all constraint drops into a single ALTER TABLE statement, which PostgreSQL 17+ rejects when a not-null constraint is listed before the primary key it depends on. See the fix at https://github.com/OCA/openupgradelib/pull/462, which drops each constraint individually.

PR #5897, opened 2026-08-10
[19.0][OU-IMP] website: preserve footer color combination

pilarvargas-tecnativa:
Odoo changed the default footer color combination from 5 to 2 in odoo/odoo@62cad846a5c8adf52fc0450bc3f5cfd0492d0622.

Explicitly set combination 5 for existing websites to preserve their previous appearance after migration.

@Tecnativa TT63973

@pedrobaeza please review

### DIFF
--- a/openupgrade_scripts/scripts/website/19.0.1.0/end-migration.py
+++ b/openupgrade_scripts/scripts/website/19.0.1.0/end-migration.py
@@ -1,0 +1,19 @@
+# Copyright 2026 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def preserve_footer_color_combination(env):
+    for website in env["website"].search([]):
+        env["website.assets"].with_context(
+            website_id=website.id
+        ).make_scss_customization(
+            "/website/static/src/scss/options/colors/user_color_palette.scss",
+            {"footer": 5},
+        )
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    preserve_footer_color_combination(env)


pedrobaeza:
Check the CI

MiquelRForgeFlow:
@pilarvargas-tecnativa  please, rebase. I made an error when force pushing to 19.0 (it's already amended)

pilarvargas-tecnativa:
> @pilarvargas-tecnativa please, rebase. I made an error when force pushing to 19.0 (it's already amended)

Done. Thanks!

pilarvargas-tecnativa:
I think the problem stems from the fact that the migration of `website_mail_group` is missing. I’d checked to make sure the script was included, but I haven’t had time to submit a pull request.

MiquelRForgeFlow:
migrating `website_mail_group` in https://github.com/OCA/OpenUpgrade/pull/5920, but `mail_group` is missing.

MiquelRForgeFlow:
done `mail_group` in https://github.com/OCA/OpenUpgrade/pull/5924

pedrobaeza:
Why is that one needed?

pilarvargas-tecnativa:
> Why is that one needed?

The “final” phase runs once all the modules included in the migration diagram have been updated. The error relating to “website_mail_group” indicates that this module has not been included in the update. When resolving web.assets_frontend, its ir.asset appears, but its files are not authorised as belonging to a loaded addon.

I think that’s the reason

MiquelRForgeFlow (commented):
BTW, why this is done in end-migration? Will this change be reverted if after migrating you do an odoo update all?

pilarvargas-tecnativa:
> BTW, why this is done in end-migration? Will this change be reverted if after migrating you do an odoo update all?

It is done in `end-migration` because `make_scss_customization()` eventually
calls `IrAsset._get_related_bundle()`, which resolves `web.assets_frontend`.
During `post-migration`, `website` and other addons contributing to that bundle
may not yet be available to the asset resolver.

A later `-u all` will not revert it. `save_asset()` persists the customization
as a website-specific `ir.attachment` and creates an `ir.asset` with the
`replace` directive. These are user-created records without XML IDs, so a
module update may invalidate and rebuild the compiled bundles, but it does not
overwrite or remove the source customization.

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #5905, opened 2026-08-12
[19.0][OU-IMP] sale_product_configurator_widget_product_label: Merged into sale

carlos-lopez-tecnativa:
This module is no longer necessary because Odoo now uses the widget created in the `account` module. This change was introduced by https://github.com/odoo/odoo/pull/170610, so this extension is no longer needed.

TT64131
@Tecnativa @pedrobaeza @pilarvargas-tecnativa could you please review this?


pedrobaeza:
Shouldn't this target 18.0 instead?

carlos-lopez-tecnativa:
Superseded by https://github.com/OCA/OpenUpgrade/pull/5906

PR #5909, opened 2026-08-13
[19.0][OU-FIX] hr_expense: Compute expense state by SQL

pedrobaeza:
Doing it by ORM is not advised due to performance, but also because some business constraints can be triggered, like in this case `_check_can_approve`, provoking a crash in the migration.

There's also a problem in executing it in post-migration instead of end-migration: if the extra modules changing
`_get_invoice_in_payment_state` are not yet loaded, you won't get `in_payment` state, so the logic has been converted to SQL, and move to end-migration.

@Tecnativa TT61992


pedrobaeza:
@hbrunn I have moved the script to end-migration, but the test is performed in post. Is there a way to perform the test at the end instead?

pedrobaeza:
@hbrunn I have discovered that the `tagged` decorator serves for that, and the tests have served for debugging the SQL.

PR #5910, opened 2026-08-14
[19.0][OU-FIX] account: Remove error logs

pedrobaeza:
- OpenUpgrade sample DB contains companies for a lot of localization CoAs, but they may not be yet loaded because the migration scripts are not done. And anyway, that message is better passed to channel admin for being aware of the problem.
- WIP: locate the origin of the "Error while loading the localization..."

@Tecnativa


hbrunn:
the error message is written by the [account module](https://github.com/odoo/odoo/blob/19.0/addons/account/models/chart_template.py#L1285) while constructing the data for the localization.

I agree with moving this to a message in the admin channel, but suggest to log an error in case the channel doesn't exist. Sooner or later I think we should provide a function in openupgradelib that can be used for that.

About the issue itself: I plan to change the way the demo db is created for v20 by caching an install of (n-1) for the modules marked as migrated in the main branch, and then installing the module(s) the PR is about on that copy. Should be comparably fast and have much less side effects, such as demo data from uninstalled modules lingering.

pedrobaeza:
Thanks for the insights. I dug into the same. The current situation is that I don't know which origin code raise that, as it seems not done by any migration script. Do you have any clue?

hbrunn:
I think this happens in _get_chart_template_data, where it iterates through all tagged functions to build the data dict. We *could* squelch those messages when calling this function because we only care about the new fields on the company record, but then I'm hesitating because this might also dump helpful error messages.

Ultimately, this is only a problem for our demo db, right? All other dbs should have their COAs set up correctly and it's a good thing to warn if that's not the case (or raise as proposed in the other PR). If so, won't we be better off to disable companies with an xmlid of a localization that's not yet marked as migrated? (we mark not-yet-migrated modules as [uninstalled](https://github.com/OCA/OpenUpgrade/blob/19.0/.github/workflows/test-migration.yml#L138) without uninstalling them, so all their data still sits in the db).

pedrobaeza:
But I have commented the end-migration scripts that do the `_load_data`, and they still appears, so they should be generated by other thing. Not too much time to investigate yet. It can happen in any desynchronized CoA AFAIK.

MiquelRForgeFlow:
@pedrobaeza  please, rebase. I made an error when force pushing to 19.0 (it's already amended)

PR #5911, opened 2026-08-14
[17.0][OU-IMP] point_of_sale: Set the appropriate value for point_of_sale.limited_product_count

victoralmau:
Set the appropriate value for `point_of_sale.limited_product_count` (related to https://github.com/odoo/odoo/blob/88423973629f04e0f6c4d0c11c7633227df14156/addons/point_of_sale/models/pos_config.py#L833)

@Tecnativa TT64083

### DIFF
--- a/openupgrade_scripts/scripts/point_of_sale/17.0.1.0.1/post-migration.py
+++ b/openupgrade_scripts/scripts/point_of_sale/17.0.1.0.1/post-migration.py
@@ -30,6 +30,37 @@ def fill_pos_order_shipping_date(env):
     )
 
 
+def fill_limited_product_count(env):
+    """In v16, you could use the limited_partners_loading and limited_products_amount
+    fields in pos.config to specify whether to perform a limited load and, if so,
+    the number of products to load; it is important to maintain this behavior in v17.
+    """
+    env.cr.execute(
+        """
+        SELECT CASE
+            WHEN EXISTS (
+                SELECT 1
+                FROM pos_config
+                WHERE limited_partners_loading
+            )
+            THEN COALESCE(
+                MIN(limited_products_amount) FILTER (
+                    WHERE limited_partners_loading
+                      AND limited_products_amount > 0
+                ),
+                0
+            )
+            ELSE 0
+        END
+        FROM pos_config
+        """
+    )
+    limited_product_count = env.cr.fetchone()[0]
+    env["ir.config_parameter"].set_param(
+        "point_of_sale.limited_product_count", str(limited_product_count)
+    )
+
+
 def product_template_convert_pos_categ_id_m2o_to_m2m(env):
     openupgrade.m2o_to_x2m(
         env.cr,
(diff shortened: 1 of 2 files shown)


openupgrade_scripts/scripts/point_of_sale/17.0.1.0.1/post-migration.py
```diff
@@ -30,6 +30,37 @@ def fill_pos_order_shipping_date(env):
...
+    fields in pos.config to specify whether to perform a limited load and, if so,
+    the number of products to load; it is important to maintain this behavior in v17.
+    """
+    env.cr.execute(
+        """
+        SELECT CASE
+            WHEN EXISTS (
+                SELECT 1
+                FROM pos_config
+                WHERE limited_partners_loading = TRUE
```
pedrobaeza:
```suggestion
                WHERE limited_partners_loading
```

openupgrade_scripts/scripts/point_of_sale/17.0.1.0.1/post-migration.py
```diff
@@ -30,6 +30,37 @@ def fill_pos_order_shipping_date(env):
...
+                ),
+                0
+            )
+            ELSE 0
+        END
+        FROM pos_config
+        """
+    )
+    limited_product_count = env.cr.fetchone()[0]
+    env["ir.config_parameter"].set_param(
```
pedrobaeza, line 59: We should always write this value? What happens if it's 0? And if not exists?

victoralmau: The https://github.com/odoo/odoo/blob/88423973629f04e0f6c4d0c11c7633227df14156/addons/point_of_sale/models/pos_config.py#L833 method retrieves the value; otherwise, it returns a default value; we are not interested in that behavior; we want to define the value that “existed in v16” (regardless of whether it was 300, 10,000, or 0).

victoralmau: This is important in v18, according to the method https://github.com/odoo/odoo/blob/9c58663fa5cea09f1a930442ee85f718ea3c306b/addons/point_of_sale/models/product.py#L95 and https://github.com/odoo/odoo/blob/9c58663fa5cea09f1a930442ee85f718ea3c306b/addons/point_of_sale/models/product.py#L100 products will be displayed in a “different” order: https://github.com/odoo/odoo/blob/9c58663fa5cea09f1a930442ee85f718ea3c306b/addons/point_of_sale/models/product.py#L124

pedrobaeza: But 0 works as unlimited?

victoralmau: Yes, it displays all records (just like in v16).

pedrobaeza (approved):

PR #5913, opened 2026-08-17
[18.0][OU-ADD] website_twitter: Nothing to do

carlos-lopez-tecnativa:
TT64173
@Tecnativa @pedrobaeza @pilarvargas-tecnativa could you please review this?


pedrobaeza:
This is not correct, as the module simply disappears, but at OU level, or you merge it with something, or it remains as it was, but not "nothing to do". As we are not putting any alternative, let's simply close it.

PR #5917, opened 2026-08-19
[19.0][OU-IMP] hr_expense: hr_expense_sheet compatibility

victoralmau:
`hr_expense_sheet` compatibility

Related to https://github.com/OCA/hr-expense/pull/378

@Tecnativa TT58631

### DIFF
--- a/openupgrade_scripts/scripts/hr_expense/19.0.2.1/post-migration.py
+++ b/openupgrade_scripts/scripts/hr_expense/19.0.2.1/post-migration.py
@@ -47,6 +47,60 @@ def hr_expense_approval_fields(env):
     )
 
 
+def hr_expense_sheet_compatibility(env):
+    """
+    We update the statuses of the hr.expense.sheet records to match those of
+    hr_expense_sheet (to ensure consistency with the statuses of hr.expense)
+    """
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("state"),
+        "state",
+        [
+            ("submit", "submitted"),
+            ("approve", "approved"),
+            ("cancel", "refused"),
+        ],
+        table="hr_expense_sheet",
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE hr_expense_sheet
+        SET state = 'posted'
+        WHERE state = 'post' AND payment_state = 'not_paid'
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE hr_expense_sheet
+        SET state = 'in_payment'
+        WHERE state = 'post' AND payment_state = 'partial'
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE hr_expense_sheet
+        SET state = 'paid'
+        WHERE state = 'post' AND payment_state IN ('paid', 'reversed')
+        """,
+    )
+    openupgrade.map_values(
+        env.cr,
+        openupgrade.get_legacy_name("approval_state"),
+        "approval_state",
+        [
+            ("submit", "submitted"),
+            ("approve", "approved"),
+            ("post", "posted"),
+            ("cancel", "refused"),
+        ],
+        table="hr_expense_sheet",
+    )
+
+
 def update_product_uoms(env):
     """
     Set updated product uoms if possible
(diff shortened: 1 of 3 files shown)


MiquelRForgeFlow:
@victoralmau please, rebase. I made an error when force pushing to 19.0 (it's already amended)

openupgrade_scripts/scripts/hr_expense/19.0.2.1/post-migration.py
```diff
@@ -47,6 +47,80 @@ def hr_expense_approval_fields(env):
...
+    We update the statuses of the hr.expense.sheet records to match those of
+    hr_expense_sheet (to ensure consistency with the statuses of hr.expense)
+    """
+    # Change states
+    openupgrade.logged_query(
+        env.cr, "UPDATE hr_expense_sheet SET state = 'submitted' WHERE state = 'submit'"
+    )
+    openupgrade.logged_query(
+        env.cr, "UPDATE hr_expense_sheet SET state = 'approved' WHERE state = 'approve'"
+    )
```
carlos-lopez-tecnativa: I think you should copy the columns first to preserve the old values as a backup. Additionally, you can use `map_values` instead of SQL.

openupgrade_scripts/scripts/hr_expense/19.0.2.1/post-migration.py
```diff
@@ -47,6 +47,80 @@ def hr_expense_approval_fields(env):
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE hr_expense_sheet
+        SET approval_state = 'refused'
+        WHERE approval_state = 'cancel'
+        """,
+    )
```
carlos-lopez-tecnativa: The same as above.

carlos-lopez-tecnativa (changes requested):

openupgrade_scripts/scripts/hr_expense/19.0.2.1/upgrade_analysis_work.txt
```diff
@@ -46,6 +46,8 @@ hr_expense   / hr.expense               / state (selection)             : select
 # DONE: recompute state for records in states 'done', 'reported'
 
 hr_expense   / hr.expense               / untaxed_amount (float)        : NEW isfunction: function, stored
+
+# DONE: hr_expense_sheet compatibility
```
hbrunn, line 50: what has this column to do with the above?

hbrunn (changes requested):
why does any of this happen in hr_expense's migration, and not in the pre-init script of hr_expense_sheet?

openupgrade_scripts/scripts/hr_expense/19.0.2.1/upgrade_analysis_work.txt
victoralmau: I am referring to everything that comes after that: `hr.expense.sheet`

victoralmau:
> why does any of this happen in hr_expense's migration, and not in the pre-init script of hr_expense_sheet?

Why can't it be in OpenUpgrade?

There are several reasons why I believe it SHOULD be here:
- It is “difficult” for someone who has used Odoo and `hr.expense.sheet` before 19.0 to stop using that functionality; reporting ‘grouped’ expenses (and reviewing them) without that option is “_very difficult_.”
- Modifying data in tables that will no longer be used in Odoo core has no side effects; in my opinion, it’s unnecessary to add logic to `hr_expense_sheet` to perform that renaming when OpenUpgrade already does it in `hr_expense`.

openupgrade_scripts/scripts/hr_expense/19.0.2.1/upgrade_analysis_work.txt
hbrunn: so put it after what you're referring to

hbrunn:
I think it just doesn't belong to this migration script because people won't necessary use hr_expense_sheet. And technically, it's not difficult to detect on installation of that module if you're coming from an ex-v18 version, so I don't see a reason to have unrelated code in the migration script of hr_expense

victoralmau:
@pedrobaeza and @MiquelRForgeFlow What do you think about this?

pedrobaeza (commented):
It's a fact that both OpenUpgrade and OCA `hr_expense_sheet` initialization are depending each other, as if you migrate with enterprise, you won't find the expected columns (it removes them), but it's true that copying some columns, processing them or renaming them without knowing if you are going to install `hr_expense_sheet` is useless and time/space consuming (although we plan to install it in most installations), so let's move it to pre_init_hook as stated by Holger, and you can detect if `former_sheet_id` exists and then trigger in that case the rest of the code. In the module, you should also include a migration script without `@openupgrade.migrate` decorator importing and calling pre_init_hook.

victoralmau:
Moved to https://github.com/OCA/hr-expense/pull/378

PR #5926, opened 2026-08-19
[16.0][OU-FIX] stock: set lot groups to portal user following new defaults

### LINKED ISSUE #5918: Migration to 16.0 deactivates group_stock_production_lot
## Module
stock, maybe base 

## Describe the bug
group_stock_production_lot is no more active after migration from a v. 14 to a v. 16

## To Reproduce
**Affected versions**:
16.0

1. enable group_stock_production_lot in v. 14.0
2. migrate to v. 16.0

**Expected behavior**
group_stock_production_lot is active

**Additional context**
This field hadn't a default in v. 14.0 of group, now it's `group="base.group_user,base.group_portal"`. If in the instance only `base.group_user` have the group `group_production_lot` enabled, the field `group_stock_production_lot` is inactive by default.
I don't know how these cases are managed in openupgrade process.

sergiocorato:
Fix #5918

### DIFF
--- a/openupgrade_scripts/scripts/stock/16.0.1.1/pre-migration.py
+++ b/openupgrade_scripts/scripts/stock/16.0.1.1/pre-migration.py
@@ -180,6 +180,20 @@ def _prefill_stock_move_quantity_done(env):
     )
 
 
+def _set_lot_groups_to_portal_user(env):
+    # If base.group_user has implied_group stock.group_production_lot, we consider it as
+    # the default, so set it to the base.group_portal too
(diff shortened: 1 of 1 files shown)


pedrobaeza:
AFAIK, groups on portal users are not relevant (they are hidden in the users UI). Why this matters?

sergiocorato:
> AFAIK, groups on portal users are not relevant (they are hidden in the users UI). Why this matters?

Because group_stock_production_lot is set to False if portal group haven't lot enabled.

Issue #5575, opened 2026-03-31
[18.0] website: Snippets custom failed


flotho:
## Module
website

## Describe the bug
When you come from a v16 the custom snippets are declared with an xpath referencving a div.

## To Reproduce
Personnalise a snippet in < v18

Migrate and try to open the website editor.

ping @DorianMAG

pedrobaeza:
Any more detail?

DorianMAG:
Hi !

The older version of custom snippet and like this

```xml
<data inherit_id="website.snippets">
    <xpath expr="//div[@id='snippet_custom']" position="attributes">
        <attribute name="class" remove="d-none" separator=" "/>
    </xpath>
    <xpath expr="//div[@id='snippet_custom_body']" position="inside">
        <t t-snippet="website.s_features_0a5b8fc1ef6040c097729e97bf24c755" t-thumbnail="/website/static/src/img/snippets_thumbs/s_features.svg"/>
    </xpath>
</data>
```

and the new

```xml
<data inherit_id="website.snippets">
    <xpath expr="//snippets[@id='snippet_custom']" position="inside">
        <t t-snippet="website.s_text_image_9d9c83b5a5a84e0f92fe382ee21f250d" t-thumbnail="oe-thumbnail"/>
    </xpath>
</data>
```

Any ideas on how to migrate, and is it even possible?

Best regards

pedrobaeza:
A migration script should be added that search for these fragments (are that views?), and do an XML manipulation to convert one thing to the other. You can get inspiration for example in https://github.com/OCA/OpenUpgrade/pull/5569

DorianMAG:
> A migration script should be added that search for these fragments (are that views?), and do an XML manipulation to convert one thing to the other. You can get inspiration for example in [#5569](https://github.com/OCA/OpenUpgrade/pull/5569)

Thanks for the feedback.
It's very clear.
Regards

pedrobaeza:
#5940 should fix it.

PR #5940, opened 2026-08-25
[18.0][FIX] web_editor: Run custom snippets migration at end

pilarvargas-tecnativa:
Related to #5575

@Tecnativa 

@etobella @pedrobaeza please review


pedrobaeza:
I think the reference number is not correct.

pilarvargas-tecnativa:
> I think the reference number is not correct.

#5575 is the issue reporting the custom snippets migration problem. Or do you mean I should reference #5435, where the original migration script was added?

pedrobaeza (approved):
Yes, I was looking for your conversation with @etobella

PR #5942, opened 2026-08-25
[18.0][OU-ADD] product_matrix: Nothing to do

carlos-lopez-tecnativa:
TT64115


carlos-lopez-tecnativa:
/ocabot migration product_matrix

<dependency>Depends on : 

PR #5945, opened 2026-08-25
[14.0][OU-FIX] sale_timesheet: Add timesheet_product_id column only if it does not exist

celm1990:
In the Enterprise edition, this field is added by the [industry_fsm](https://github.com/odoo/enterprise/blob/b27b98153f966c36700d154bdad4603eb85ecb32/industry_fsm/models/project.py#L28) module. To prevent an SQL error when the column already exists, add the IF NOT EXISTS clause.

@pedrobaeza @MiquelRForgeFlow @hbrunn could you please review this?

### DIFF
--- a/openupgrade_scripts/scripts/sale_timesheet/14.0.1.0/post-migration.py
+++ b/openupgrade_scripts/scripts/sale_timesheet/14.0.1.0/post-migration.py
@@ -22,7 +22,8 @@ def _fill_project_timesheet_product(env):
         env.cr,
         """UPDATE project_project pp
         SET timesheet_product_id = %s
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
The link doesn't work, as it's private. I think this is something that should be handled outside OpenUpgrade, as if the column exists, we can't control which value it has and why is there, so adding defensive code in case they exist is not the mission of OpenUpgrade (except the column exists by the own Odoo community in a previous version), but something to be prepared previously.

Imagine you have the column because there's an exotic module installed in your DB adding `timesheet_product_id` with another value. It would be more tolerable if the `UPDATE` query still overwrites all the values.

In this case, I would remove the column before passing OU.

PR #5946, opened 2026-08-26
[17.0][IMP] website: activate default header for websites using deprecated header layout

hbrunn:


pedrobaeza:
I'm not sure about this, as this won't crash, but silently will change totally the format of website layout.

hbrunn:
the layout is totally changed because the views in question are deleted during the migration. and then after migration some js crashes because the expected header node doesn't exist, leaving users with a defunct website editor, and they'll have to know that they need to go to the backend and activate one of the existing headers

pedrobaeza:
Where are they deleted? And you are searching for views matching the `website_id` field, but they are not the generic or neither exist, isn't it?

hbrunn:
> Where are they deleted?

ie [here](https://github.com/OCA/OpenUpgrade/actions/runs/32955158190/job/98135113019?pr=5946#step:13:24953)

> And you are searching for views matching the `website_id` field, but they are not the generic or neither exist, isn't it?

there will be a website-specific copy of the view because that's what Odoo does when you activate some other header layout, or any other view for that matter

pedrobaeza (approved):
OK, I'm not sure about the possible effects, but let's see it in action.

PR #5948, opened 2026-08-26
[15.0][OU-FIX] base: fix mail.default.from conflict during migration

celm1990:
Prevent `DETAIL: Key (key)=(mail.default.from) already exists.`
@pedrobaeza @MiquelRForgeFlow @hbrunn could you please review this?


pedrobaeza:
This should go into `mail`, not here, but I'm also reluctanct to add this kind of defensive code if we don't know if the contents of the parameter is ok according the expected one. If you have pre-create it, don't do it, and set it afterwards.

PR #5951, opened 2026-08-28
[FIX] stock_account: a zero unit_cost is not a unit cost

cuongnmtm:


openupgrade_scripts/scripts/stock_account/tests/data_stock_account_migration.py
```diff
@@ -82,3 +82,33 @@ def buy(product, **kwargs):
...
+        "standard_price": 10,
+        "is_storable": True,
+    }
+)
+buy(product_reval, price_unit=20, product_uom_qty=4)
+assert product_reval.standard_price == 20
+env.cr.commit()
+
+# 17.0 writes unit_cost 0 -- not NULL -- for this kind of adjustment, and the ORM
+# offers no way to ask for that shape, so the layer is inserted directly.
```
hbrunn, line 103: if there's no way to create this kind of record via Odoo functions, how are they relevant?

hbrunn (changes requested):

PR #5684, opened 2026-06-02
[19.0][OU-ADD] website_mass_mailing: Nothing to do

eduezerouali-tecnativa:
cc @Tecnativa TT58752
ping @pedrobaeza @hbrunn

PR #5961, opened 2026-09-03
[19.0] More website_* modules

MiquelRForgeFlow:
.


MiquelRForgeFlow:
/ocabot migration website_crm_livechat

<dependency>Depends on : 

- website_sale : #5585
- website_mass_mailing : #5684</dependency>

PR #5973, opened 2026-09-07
[19.0][OU-ADD] website_login_page_editable: Merged into web

cristina-hidalgo-tecnativa:
Merged into `web`

Related to https://github.com/odoo/odoo/commit/91de83460b0708794196bd8fba3d7ed5efece11c

@Tecnativa TT64155
@pedrobaeza @pilarvargas-tecnativa

### DIFF
--- a/openupgrade_scripts/apriori.py
+++ b/openupgrade_scripts/apriori.py
@@ -71,6 +71,8 @@ merged_modules = {
     "stock_picking_mass_action": "stock",
     # OCA/vertical-association
     "membership_extension": "membership",
+    # OCA/website
+    "website_login_page_editable": "web",
 }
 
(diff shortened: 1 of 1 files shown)


hbrunn (changes requested):
I only see this addon in v17, but the functionality was added to core in v19

hbrunn (approved):
I only see this addon in v17, but the functionality was added to core in v19

pedrobaeza (approved):
This is because potentially, the module is applicable to 18, so if migrating to 18, the merge shouldn't apply, and you will still need the module, but you have to migrate it. If you migrate from 17 to 19 (as our case), then we don't need it anymore and it's merged in this version, as included in core.

PR #5976, opened 2026-09-08
[IMP] point to the v20 crowdfunding

hbrunn:


legalsylvain (approved):

pedrobaeza (approved):
OK, but the lots are not OK in my opinion. Lot 5 should be `sale` (missing) and `sale_management`, but to not include `point_of_sale`, which is by far less used than `purchase` or `hr`.

hbrunn:
I must admit I never found the lots to be very relevant, as we migrate all core modules anyways, and at least in the beginning, dependencies dictate the order

PR #5977, opened 2026-09-09
[18.0][FIX] mrp: location_dest_id with the MRP module installed.

angelgarciadelachica:
## **Summary**

This PR fixes the pre-migration script for the MRP module in version 18 by adjusting the calculation of `location_dest_id` to consider inheritance and the custom logic applied by the Manufacturing (MRP) module.

## Problem Description

The current pre-migration script for the stock module to v18 renames `location_dest_id` to `location_final_id` and recalculates `location_dest_id` using the standard `stock.move` queries (based on `_compute_location_dest_id()`).

However, this logic is incomplete when the mrp module is installed, since this module overrides the `_compute_location_dest_id()` method. Currently, there is no mechanism in place to handle this scenario, resulting in an incorrect `location_dest_id` in stock movements generated from a production order.

## To Reproduce

1. Perform a database migration from v17 to v18 with MRP installed.
2. Examine an existing manufacturing order (MO) in the source version:
Selección_430

4. Verify the stock movements generated by that MO:
Selección_429

5. In v18, review the stock movement history. You will notice that `location_dest_id` has not been migrated correctly. The expected behavior, following the logic for creating an MO in v18, should be different:
Selección_432

@HaraldPanten @ValentinVinagre 
I-8773

### DIFF
--- a/openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
+++ b/openupgrade_scripts/scripts/mrp/18.0.2.0/pre-migration.py
@@ -29,8 +29,32 @@ def mrp_workorder_sequence(env):
     )
 
 
+def fill_mrp_stock_move_location_dest_id(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move sm
+        SET location_dest_id = mp.location_dest_id
+        FROM mrp_production mp
+        WHERE sm.production_id = mp.id
+            AND mp.location_dest_id IS NOT NULL
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move sm
+        SET location_dest_id = mp.production_location_id
+        FROM mrp_production mp
+        WHERE sm.raw_material_production_id = mp.id
+            AND mp.production_location_id IS NOT NULL;
+        """,
+    )
+
+
 @openupgrade.migrate()
 def migrate(env, version):
     openupgrade.add_columns(env, add_columns)
     openupgrade.copy_columns(env.cr, copy_columns)
     mrp_workorder_sequence(env)
+    fill_mrp_stock_move_location_dest_id(env)


pedrobaeza:
If the moves were generated in 17, they should contain already the correct destination location. Can you point where in OpenUpgrade they are being changed back to WH/Stock?

angelgarciadelachica:
Thanks for your reply! :-)

In the Stock pre-migration script:
1. [Rename](https://github.com/OCA/OpenUpgrade/blob/068c6309b8f1ff71c8212a870424abbf12048389/openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py#L10) the old `location_dest_id` to `location_final_id`
2. Then create a new `location_dest_id`, which is populated by [`fill_stock_move_location_dest_id()`](https://github.com/OCA/OpenUpgrade/blob/068c6309b8f1ff71c8212a870424abbf12048389/openupgrade_scripts/scripts/stock/18.0.1.1/pre-migration.py#L42) using the logic from [`_compute_location_dest_id()`](https://github.com/odoo/odoo/blob/db1315502cfa31d62305948524160b211c5e50c0/addons/stock/models/stock_move.py#L220).
The problem is that the migration script does not take into account the inherited logic from the mrp module -> [`_compute_location_dest_id()`](https://github.com/odoo/odoo/blob/81d91393f481f2454b9df76387de195f1312e14d/addons/mrp/models/stock_move.py#L287-L297).

angelgarciadelachica:
Hi! Could you review this PR? @HviorForgeFlow @hbrunn
Thank you very much!!

HaraldPanten (approved):

pedrobaeza:
OK, understood. Thanks for the explanations. Then, isn't better to simply restore the previous value got in `location_final_id`?

HviorForgeFlow:
CC @MiquelRForgeFlow

pedrobaeza:
ping @angelgarciadelachica

angelgarciadelachica:
@pedrobaeza Yes, you're right. It would also be possible to obtain the `location_dest_id` from the `location_final_id` for movements that have a `production_id` or `raw_material_production_id`.

The reason I suggested doing this using the two queries is mainly to adhere to the logic of Odoo 18 and maintain the same behavior as during a normal creation of these movements.

Furthermore, these two queries have the advantage that they can also be used to correct instances that have already been migrated to v18 and are experiencing this issue.

In any case, there would be no problem changing it and doing it the way you suggest if you consider that to be a more appropriate solution.

Issue #5474, opened 2025-12-13
[17.0] repair: not-null constraint conflict


FranciscoBigWings:
The migration is failing due to a conflict between OpenUpgrade's data cleanup logic. The failure occurs because a temporary action by the migration script—deleting a placeholder location (ID 26 in this case)—triggers a ON DELETE SET NULL, but the field being updated (location_dest_id) is simultaneously required to be non-empty (NOT NULL).

## Module
Repair

## Describe the bug
OpenUpgrade runs ALTER TABLE commands to add the new Odoo 17 stock location fields.
ALTER TABLE "repair_order" ADD COLUMN "location_dest_id" int4

Later the scripts does this:
JOIN stock_location sld ON sld.id = rl.location_dest_id

And this:
UPDATE repair_order ro SET ... location_dest_id = spt.default_location_dest_id ...

The OpenUpgrade script reaches its final cleanup stage and attempts to delete the temporary location ID 26:
OpenUpgrade Action: DELETE FROM "stock_location" WHERE id IN (26). Because one or more constraints exist between stock_location and repair_order (likely set to ON DELETE SET NULL), the database automatically attempts to run the cascade query: UPDATE ONLY "public"."repair_order" SET "location_dest_id" = NULL WHERE ... location_dest_id = 26

The Conflict: null value in column "location_dest_id" violates not-null constraint

**Additional context**
Python 3.12

The uncommented log for the reference
```
2025-12-13 17:27:16,238 15469 DEBUG odoo17 OpenUpgrade: -1 rows affected after 0:00:00.001639 running ALTER TABLE "stock_warehouse" ADD COLUMN "repair_type_id" int4 
2025-12-13 17:27:16,240 15469 DEBUG odoo17 OpenUpgrade: -1 rows affected after 0:00:00.000961 running ALTER TABLE "repair_order" ADD COLUMN "picking_type_id" int4 
2025-12-13 17:27:16,241 15469 DEBUG odoo17 OpenUpgrade: -1 rows affected after 0:00:00.000792 running ALTER TABLE "repair_order" ADD COLUMN "location_dest_id" int4 
2025-12-13 17:27:16,242 15469 DEBUG odoo17 OpenUpgrade: -1 rows affected after 0:00:00.000492 running ALTER TABLE "repair_order" ADD COLUMN "parts_location_id" int4 
[...]

PR #5985, opened 2026-09-15
[FIX] repair: Virtual Locations have no Warehouse

SirPyTech:
Fix https://github.com/OCA/OpenUpgrade/issues/5474.

When the Location of a Repair Order is a Virtual Location (no Warehouse), the temporary Location/Type could not be deleted resulting in the error:
> psycopg2.errors.NotNullViolation: null value in column "location_dest_id" of relation "repair_order" violates not-null
> constraint CONTEXT:  SQL statement "UPDATE ONLY "public"."repair_order" SET "location_dest_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "location_dest_id""

Archive and keep the temporary Location/Type if that happens.


SirPyTech:
@FranciscoBigWings @pleirb you faced the linked issue, would you like to have a look? Thanks!

HekkiMelody (approved):
Code review, LGTM.

Also, tested on a migration that had this problem.

PR #5987, opened 2026-09-18
[18.0][OU-ADD] account_reports: migration scripts

SparkDrago05:
This adds migration support for renamed Profit and Loss XML IDs in account_reports.

Changes:
- account_financial_report_totalincome0
  -> account_financial_report_operating_income0
- account_financial_report_totalincome0_balance
  -> account_financial_report_operating_income0_balance

Without these XML-ID renames, existing Odoo 17 databases can fail during the
Odoo 18 module update with duplicate-key errors on:

- account_report_line_code_uniq
- account_report_expression_line_label_uniq

The migration preserves the existing account.report.line and
account.report.expression records and lets Odoo 18 update them in place.


pedrobaeza:
`account_reports` is an enterprise module, so this can't be hosted here, because it can disclose private information about that module, which is private, and can't be tested in the CIs.

PR #5988, opened 2026-09-18
[18.0][FIX] analytic: migrate default_applicability to company-dependent jsonb

SparkDrago05:
Fixes the migration of account.analytic.plan.default_applicability.

In 17.0 the field is stored as a regular selection value.
In 18.0 it is company-dependent and stored as a JSONB mapping keyed by company.

The pre-migration converts existing values before Odoo initializes the
18.0 field definition.

This addresses the migration gap already noted in upgrade_analysis_work.txt.


pedrobaeza:
You can use this openupgradelib method: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_180.py#L7

hbrunn (changes requested):
this PR is bullshit: there is no gap as this is done in post migration, and there is no column in v17. see failing tests

PR #5989, opened 2026-09-18
[13.0][OU-FIX] hr_contract: fix value of hr_employee.contract_id

huguesdk:
fix the query setting the values of `hr_employee.contract_id`: have one row per employee instead of just one row in the sub-query.

another way to write it while keeping the `limit 1` (using [a `lateral` sub-query](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL)) could be:

```sql
        UPDATE hr_employee he
        SET contract_id = sub.contract_id
        FROM hr_employee he2, LATERAL (
            SELECT hc.id as contract_id
            FROM hr_contract hc
            WHERE he2.id = hc.employee_id AND hc.state = 'open'
            ORDER BY hc.date_start DESC
            LIMIT 1
        ) sub
        WHERE he2.id = he.id AND he.contract_id IS NULL
```

what do you prefer?


pedrobaeza (approved):

huguesdk:
thanks @pedrobaeza!

